PR #1787, opened 2022-12-19
Add view_identities permission

### LINKED ISSUE #1773: Have read-only permission for 'users'
We have groups that will give 'View Project' access to our members. However, that only gives access to the features, not the 'users' section. We would like to have View Project include viewing the users, or have a different config that'll control read/write access to users.

gz#219

kyle-ssg:

## Changes

Assuming the **view_identities_permission** feature is enabled:

- Users with VIEW_IDENTITIES can view the users aside nav
- Users with MANAGE_IDENTITIES can delete an identity, it will be disabled for VIEW_IDENTITIES permissions
- Users with VIEW_IDENTITIES can see a read only view of the identity detail page
- Users with MANAGE_IDENTITIES can edit a users feature states, this used to be based on UPDATE_FEATURE_STATE


matthewelwell (dismissed):

matthewelwell:
It seems like this has had an effect on users with VIEW_ENVIRONMENT permissions from viewing Change Requests? 

image

To reproduce: 

 * Create a user with VIEW_PROJECT and VIEW_ENVIRONMENT permissions
 * Login with that user and try to view the Change Requests page

This _might_ be unrelated to this PR but I've tested and it doesn't happen in production or staging.

matthewelwell:
The above is unrelated to this PR - it's caused by the fact that the preview environment is built without the change request views (as they are contained in a separate private repository and bundled into the application at build time for SaaS / private cloud build workflows).

matthewelwell:
In testing this, I've noticed a few issues:

- [x] Tooltips for delete identity and 'Add New Trait' buttons show that the user needs the 'update feature state' permission, rather than 'manage identities' (screenshots below)
- [x] Tooltip for 'Create Users' button shows that the user needs 'Admin' for the environment but should show 'manage identities' (screenshot below) 
- [x] When an identity has no traits, I am able to click on the 'Add New Trait' button (note: this is not the case when the identity already has traits). It looks like this works too because the FE is using the SDK endpoint rather than the admin endpoint. 
- [x] Delete trait button is clickable and tries to delete the trait but fails silently with a 403 error on the API (which is correct) 

image

image

image

matthewelwell (approved):
Approved but I've created #1819 to track issue related to point 3 above.

PR #1793, opened 2022-12-20
Support float value

### LINKED ISSUE #1790: User traits list UI doesn't like large floats
I've noticed large floats (maybe all floats) don't show up on the users trait list view in the UI

image

I've worked around this some by saving them as strings.  In the actual database and from the SDK the values are correct.

image

gz#227

kyle-ssg:

## Changes
This fixes the display of traits, where float_value is never displayed.

### DIFF
--- a/frontend/common/utils/utils.js
+++ b/frontend/common/utils/utils.js
@@ -275,7 +275,6 @@ module.exports = Object.assign({}, require('./base/_utils'), {
                 type: 'bool',
                 boolean_value: val,
                 integer_value: null,
-                float_value: null,
                 string_value: null,
             };
         }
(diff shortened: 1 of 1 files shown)


matthewelwell:
@kyle-ssg Unless the FE is re-using code across Traits and FeatureStates, I don't think this will resolve the linked issue? The linked issue relates to traits only and the code in this PR seems to refer only to FeatureStates.

kyle-ssg:
@matthewelwell it's just the name of the parameter that's missleading, both traits and feature states run through this since they have the same properties (string_value etc)

Here's the before and after: 

image

matthewelwell (approved):

PR #1807, opened 2022-12-21
Refactor projectOverrides

### LINKED ISSUE #1803: Segments redirecting to projects in self hosted version
Self hosted version. Clicking on the segment will redirect to project page. The router is problematic

gz#236

kyle-ssg:

## Changes

The way that project overrides were loaded had a race condition issue that was highlighted by recent Redux additions.

This changes the way that ``project.js`` is loaded so that it always includes overrides since they are loaded in the head.


matthewelwell (approved):
This seems ok to me but it's way above my JS pay grade!

PR #1812, opened 2022-12-29
feat(metadata): Add a generic way to add  metadata to different objects

### LINKED ISSUE #1610: Add meta data fields to core entities
# Overview 

Larger teams often require the ability to add their own meta-data to the following entities:

- Flags
- Segments
- Environments

Examples of the meta data that they might add:

- JIRA ticket code (e.g. `FLAGSMITH-152`) to a Flag
- URL for Infra overview for an Environment
- Link to Amplitude for a Segment

These data could be optional or required. For example, engineers should be required to add a JIRA ticket against a new Flag when the flag is created

# Implementation

Administrators should be able to define fields that they want associated with an entity, and whether it is required or optional.

To begin with the administrator should be able to define with the field is single line or multi-line. For v1, just allow any text. 

gz#40

gagantrivedi:

## Changes

Add, a metadata app that can be used to add metadata to different objects
As a working proof of that, it adds the metadata to environment 

### New APIs
`​/api​/v1​/metadata​/fields​/`
`​/api​/v1​/metadata​/fields​/{ID}`
`​/api​/v1​/organisations​/{organisation_pk}​/metadata-model-fields​/`
`​/api​/v1​/organisations​/{organisation_pk}​/metadata-model-fields​/{ID}`
`​/api​/v1​/organisations​/{organisation_pk}​/metadata-model-fields​/supported-content-types​/`
`​/api​/v1​/organisations​/{organisation_pk}​/metadata-model-fields​/supported-required-for-models​/`

Covered by unit test cases


api/metadata/models.py
```diff
@@ -0,0 +1,44 @@
...
+    is_required = models.BooleanField(default=False)
+
+    class Meta:
+        unique_together = ("field", "content_type")
+
+
+class Metadata(AbstractBaseExportableModel):
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    object_id = models.PositiveIntegerField()
+    content_object = GenericForeignKey("content_type", "object_id")
```
matthewelwell, line 117: Oh man, I didn't know this existed - that's super cool!

api/metadata/models.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class FieldType(models.TextChoices):
+    INTEGER = "int"
+    STRING = "str"
+    BOOLEAN = "bool"
+    FLOAT = "float"
+
+
+class MetadataField(AbstractBaseExportableModel):
```
matthewelwell, line 36: So, as I understand it: 

This is a class for creating generic fields in an organisation?

...

api/metadata/models.py
```diff
@@ -0,0 +1,44 @@
...
+    name = models.CharField(max_length=255)
+    type = models.CharField(
+        max_length=255, choices=FieldType.choices, default=FieldType.STRING
+    )
+    description = models.TextField(blank=True, null=True)
+    organisation = models.ForeignKey(Organisation, on_delete=models.CASCADE)
+
+
+class MetadataModelField(AbstractBaseExportableModel):
```
matthewelwell, line 80: ... this is a class for linking a metadata field to another model class and specifying if it's required or not? 

...

api/metadata/models.py
```diff
@@ -0,0 +1,44 @@
...
+    field = models.ForeignKey(MetadataField, on_delete=models.CASCADE)
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    is_required = models.BooleanField(default=False)
+
+    class Meta:
+        unique_together = ("field", "content_type")
+
+
+class Metadata(AbstractBaseExportableModel):
```
matthewelwell, line 109: ... and this is a class which actually contains the data for a specific field against a specific model object? 

...

api/metadata/models.py
```diff
@@ -0,0 +1,44 @@
...
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
```
matthewelwell, line 115: ... what I don't understand is why we need content_type here when it's also defined against the MetadataModelField class (which we also link to below)?

api/metadata/models.py
gagantrivedi: Yep

api/metadata/models.py
gagantrivedi: Yep

api/metadata/models.py
gagantrivedi: Yep

api/metadata/models.py
gagantrivedi: Yeah, it's for creating the `GenericForeignKey` which is used for creating a reverse relation here:
https://github.com/Flagsmith/flagsmith/pull/1812/files#diff-a0b057e41b454fcf5879b6c8ddbb1cb10a4c446e671ef37ff3b242efa1e7200bR84

api/environments/serializers.py
```diff
@@ -29,7 +30,9 @@ class Meta:
         )
 
 
-class EnvironmentSerializerLight(serializers.ModelSerializer):
+class EnvironmentSerializerLight(MetadataSerializerMixin, serializers.ModelSerializer):
```
matthewelwell: Hmm... the point of the 'Light' serializer is to keep it light (i.e. not have to make extra queries, etc.). We shouldn't add the metadata here, we should perhaps create a totally new serializer.

api/environments/models.py
```diff
@@ -91,6 +93,7 @@ class Environment(
     banner_colour = models.CharField(
         null=True, blank=True, max_length=7, help_text="hex code for the banner colour"
     )
+    metadata = GenericRelation(Metadata)
```
matthewelwell, line 101: Man, this stuff is so cool!

api/import_export/export.py
```diff
@@ -83,6 +85,18 @@ def export_organisation(organisation_id: int) -> typing.List[dict]:
...
 
 
+def export_metadata(organisation_id: int) -> typing.List[dict]:
+    return _export_entities(
+        _EntityExportConfig(MetadataField, Q(organisation__id=organisation_id)),
+        _EntityExportConfig(
+            MetadataModelField, Q(field__organisation__id=organisation_id)
+        ),
+        _EntityExportConfig(
+            Metadata, Q(model_field__field__organisation__id=organisation_id)
```
matthewelwell, line 104: Will this work without using the UUID to handle the relationship?

api/metadata/mixins.py
```diff
@@ -0,0 +1,92 @@
...
+
+from .models import MetadataModelField
+from .serializers import MetaDataModelFieldSerializer
+
+
+class MetaDataViewMixin:
+    app_label = None
+    model_name = None
+
+    def get_base_metadata_fields_queryset(self):
```
matthewelwell: I think this method could be marked as 'private', right?

api/metadata/mixins.py
```diff
@@ -0,0 +1,92 @@
...
+
+    @swagger_auto_schema(method="GET", responses={200: MetaDataModelFieldSerializer()})
+    @action(
+        detail=False,
+        methods=["GET"],
+        url_path="metadata-fields",
+        serializer_class=MetaDataModelFieldSerializer,
+    )
+    def get_model_metadata_fields(self, request):
+        content_type = ContentType.objects.get(
```
matthewelwell: Unless I'm yet to reach a point where app_label and model_name are used elsewhere, I think we can change this to `ContentType.objects.get_for_model(self.model_class)` and just define the model class against the view. Or perhaps we can even infer that from the queryset or something? ... but that's probably overkill.

api/metadata/mixins.py
```diff
@@ -0,0 +1,92 @@
...
+        responses={200: MetaDataModelFieldSerializer()},
+        request_body=MetaDataModelFieldSerializer(),
+    )
+    @action(
+        detail=False,
+        methods=["PUT"],
+        url_path=r"metadata-field/(?P<metadata_pk>\d+)",
+    )
+    def update_model_metadata_field(self, request, metadata_pk=None):
+        queryset = self.get_base_metadata_fields_queryset()
```
matthewelwell: Is there a reason we shouldn't filter based on the ContentType here? Just to save an extra query I guess?

api/metadata/mixins.py
```diff
@@ -0,0 +1,92 @@
...
+        method="POST",
+        responses={201: MetaDataModelFieldSerializer()},
+        request_body=MetaDataModelFieldSerializer(),
+    )
+    @action(detail=False, methods=["POST"], url_path="metadata-field")
+    def create_model_metadata_field(self, request):
+        serializer = MetaDataModelFieldSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        if not self.request.user.organisations.filter(
```
matthewelwell: Just playing devil's advocate here, is there a reason we wouldn't just create a viewset for MetadataModelFields nested under an organisation? The FE can then send up the content type as part of the payload?

api/metadata/mixins.py
```diff
@@ -0,0 +1,92 @@
...
+
+        serializer = MetaDataModelFieldSerializer(
+            model_metadata_field, data=request.data, partial=True
+        )
+        serializer.is_valid(raise_exception=True)
+
+        serializer.save()
+        return Response(serializer.data)
+
+    @update_model_metadata_field.mapping.delete
```
matthewelwell: Ooh, what does this do?

api/metadata/models.py
matthewelwell: Is there any way around it? Can the reverse relation be created based on related fields for example? Something like `model_field__content_type`. Feels weird that we're having to define the same data twice.

api/metadata/permissions.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class MetadataFieldPermissions(IsAuthenticated):
+    def has_permission(self, request, view):
+        if not super().has_permission(request, view):
+            return False
+
+        organisation = request.data.get("organisation")
+        if view.action == "create" and request.user.belongs_to(organisation):
+            return request.user.is_organisation_admin(request.data.get("organisation"))
```
matthewelwell: I think you can replace `request.data.get("organisation")` with just `organisation` since you've already stored it in a variable above. 

Also, it might be worth creating an organisation level permission for this (although perhaps that's something that we can do in a separate PR).

api/metadata/serializers.py
```diff
@@ -0,0 +1,133 @@
...
+        model = MetadataModelField
+        fields = ("id", "field", "is_required")
+
+
+class FieldDataField(serializers.Field):
+    """
+    Custom field to handle the different types of data supported by the metadata field
+    """
+
+    def to_internal_value(self, data):
```
matthewelwell: Do we need to do this? Shouldn't we just get the client to send up the type and we can always just store the data as a string?

api/metadata/serializers.py
```diff
@@ -0,0 +1,133 @@
...
+            raise serializers.ValidationError(
+                f"Value string is too long. Must be less than {FIELD_VALUE_MAX_LENGTH} character"
+            )
+        return data
+
+    def to_representation(self, value):
+        return value
+
+
+class MetadataListSerializer(serializers.ListSerializer):
```
matthewelwell: Should we update the name of this serializer since we're (seemingly) using it for more than just listing.

api/metadata/serializers.py
```diff
@@ -0,0 +1,133 @@
...
+    def save(self, **kwargs):
```
matthewelwell: I'm a little confused what we're doing here - is this some kind of bulk create functionality?

api/metadata/serializers.py
```diff
@@ -0,0 +1,133 @@
...
+        fields = ("model_field", "field_value")
+        list_serializer_class = MetadataListSerializer
+
+    def validate(self, data):
+        if type(data["field_value"]).__name__ != data["model_field"].field.type:
+            raise serializers.ValidationError("Invalid type")
+        return data
+
+    def to_representation(self, value):
+        # Convert field_value to its appropriate type(from string)
```
matthewelwell: Why though?

api/metadata/serializers.py
```diff
@@ -0,0 +1,133 @@
...
+                raise serializers.ValidationError(
+                    f"Missing required metadata field {metadata_field.field.name}"
+                )
+
+    def save(self, **kwargs):
+        self.validate_required_metadata()
+
+        instance = super().save(**kwargs)
+
+        metadata = self.get_initial().get("metadata", None)
```
matthewelwell: Can we use a WritableNestedSerializer for this instead? or perhaps not because of all the GenericRelation stuff?

matthewelwell (commented):
I've been through the logic (but not the test code yet) and added a few comments. Once we've nailed down all the questions on the logic then I can go through the tests but I feel like there might end up being a few changes to the tests based on the comments on the logic. 

I also noticed that there is an additional (probably erroneous) file that's been added at `api/output.txt` (annoyingly you can't leave comments on empty files in GH).

api/metadata/models.py
gagantrivedi: Yeah, it does feel weird, but I don't think there is any other way to create a reverse relation

api/metadata/serializers.py
gagantrivedi: Hmm, I think there was some reason (Ah no it was related to the old design) good shout

api/metadata/serializers.py
gagantrivedi: But we already have the expected type for this, right? This is used to verify its type

api/import_export/export.py
gagantrivedi: You mean with MetadataModelField ?

api/metadata/mixins.py
gagantrivedi: I don't think that helps much, though, right? We still need to perform these checks, right?

api/metadata/mixins.py
gagantrivedi: It basically provides the same `url_path` as the update method

api/metadata/serializers.py
gagantrivedi: Let  front-end to do the type casting?

matthewelwell: Yep, I think for any new fields that can be multiple types, we should not do any type casting on the BE - it makes things very difficult! The FE can then use the string value and the returned type value to display it how they want.

api/import_export/export.py
matthewelwell: Yes, sorry, I think I put the comment in the wrong place.

api/metadata/mixins.py
matthewelwell: Well, it just means we don't have to dive into the request body to do permission checks which for some reason feels slightly wrong to me (although I'm aware we already have other scenarios that we do do this.

api/metadata/mixins.py
matthewelwell: Ah, ok, so it's basically just shorthand for: 

```python
@action(
        detail=False,
        methods=["DELETE"],
        url_path=r"metadata-field/(?P<metadata_pk>\d+)",
    )
```

api/metadata/serializers.py
matthewelwell: Hmm... I think we should just get the FE to send a string (so we don't have to hack the API docs) and then do something more like: 

```python
casting_functions = {"int": int, "str": str, ...}

try:
    casting_function = casting_functions.get(field_type)  # we might need to use context to get the field type here? 
    casting_function(data)
except ValueError:
    raise ValidationError(...)

if len(str(data)) > FIELD_VALUE_MAX_LENGTH:
    raise ValidationError(...)

return data
```

api/import_export/export.py
gagantrivedi: It does have UUID, though?

matthewelwell: But it's not a traditional foreign key, right? So I'm not sure how it will link to the relevant entities when we export it (I can't quite remember how the export works tbh!)

api/metadata/mixins.py
matthewelwell: Did this get removed instead?

gagantrivedi: Yep

api/metadata/mixins.py
matthewelwell: Nice, just saw that we were able to use `Meta.model` 👌

api/metadata/serializers.py
```diff
@@ -0,0 +1,81 @@
...
+                "str": str,
+                "bool": str_to_bool,
+            }[expected_type]
+            casting_function(data["field_value"])
+        except ValueError:
+            raise serializers.ValidationError(
+                f"Invalid data type. Must be the string representation of {expected_type}"
+            )
+
+        if expected_type == str and len(data["field_value"]) > FIELD_VALUE_MAX_LENGTH:
```
matthewelwell: I think this should be `"str"` ?

api/metadata/serializers.py
```diff
@@ -0,0 +1,81 @@
...
+    def validate(self, data):
+        data = super().validate(data)
+        expected_type = data["model_field"].field.type
+        try:
+            casting_function = {
+                "int": int,
+                "float": float,
+                "str": str,
+                "bool": str_to_bool,
+            }[expected_type]
```
matthewelwell: Should we add a separate 'except KeyError` ? I guess the type can't be anything other than one of these but might be best to catch it just in case?

api/metadata/views.py
```diff
@@ -0,0 +1,93 @@
...
+
+@method_decorator(
+    name="list",
+    decorator=swagger_auto_schema(
+        manual_parameters=[
+            openapi.Parameter(
+                "organisation",
+                openapi.IN_QUERY,
+                "ID of the organisation to filter by.",
+                required=False,
```
matthewelwell: I think we should probably make this required? I can't see any reason why we'd want to show fields outside of the context of an organisation?

api/metadata/permissions.py
```diff
@@ -0,0 +1,58 @@
...
+from metadata.models import MetadataField
+
+
+class MetadataFieldPermissions(IsAuthenticated):
+    def has_permission(self, request, view):
+        if not super().has_permission(request, view):
+            return False
+
+        organisation = request.data.get("organisation")
+        if view.action == "create" and request.user.belongs_to(organisation):
```
matthewelwell: I'm sure why we need the `request.user.belongs_to(organisation)` clause? From the looks of it, that's covered by `request.user.is_organisation_admin(organisation)`

api/metadata/permissions.py
```diff
@@ -0,0 +1,58 @@
...
+
+        # list is handled by the view
+        if view.action == "list":
+            return True
+
+        # move on to object specific permissions
+        return view.detail
+
+    def has_object_permission(self, request, view, obj):
+        if view.action in ("update", "destroy") and request.user.is_organisation_admin(
```
matthewelwell: What about `partial_update` and `retrieve` ? I think the ViewSet is a ModelViewSet so these actions will exist.

api/metadata/permissions.py
```diff
@@ -0,0 +1,58 @@
...
+
+        return False
+
+
+class MetadataModelFieldPermissions(IsAuthenticated):
+    def has_permission(self, request, view):
+        if not super().has_permission(request, view):
+            return False
+
+        organisation_pk = int(view.kwargs.get("organisation_pk"))
```
matthewelwell: If the FE accidentally tries to call `/organisations/undefined/metadata-...` this will throw a hard exception, right? Obviously we'd expect it to error in that case but it would be nice if it failed more gracefully than raising a `TypeError`.

api/metadata/permissions.py
```diff
@@ -0,0 +1,58 @@
...
+
+        if request.user.belongs_to(organisation_pk):
+            if view.action in ["list", "get_supported_content_types"] or view.detail:
+                return True
+
+            if view.action == "create":
+                field = MetadataField.objects.get(id=request.data.get("field"))
```
matthewelwell: Similar to above, this could throw `MetadataField.DoesNotExist`, right?

api/metadata/permissions.py
```diff
@@ -0,0 +1,58 @@
...
+
+                return (
+                    request.user.is_organisation_admin(organisation_pk)
+                    and organisation_pk == field.organisation.id
+                )
+
+        return False
+
+    def has_object_permission(self, request, view, obj):
+        if view.action in ("update", "destroy") and request.user.is_organisation_admin(
```
matthewelwell: As above, what about the other actions?

api/metadata/serializers.py
```diff
@@ -0,0 +1,81 @@
...
+        return data
+
+
+class MetadataSerializerMixin:
+    def validate_required_metadata(self, data):
+        metadata = data.get("metadata", [])
+
+        content_type = ContentType.objects.get_for_model(self.Meta.model)
+
+        required_metadata_fields = MetadataModelField.objects.filter(
```
matthewelwell: Shouldn't we be filtering this by organisation too?

api/metadata/tests/conftest.py
```diff
@@ -0,0 +1,23 @@
+import pytest
```
matthewelwell: Using this line to comment as it's the first line in a file (and Github still doesn't allow commenting on files as a whole!) 

Unless there's a specific reason not to, I think these test files should all be in `api/tests/metadata/...`?

api/tests/unit/import_export/test_unit_import_export_export.py
```diff
@@ -125,6 +130,50 @@ def test_export_environments(project):
...
+    metadata_field = MetadataField.objects.create(
+        name="test_field", type="int", organisation=organisation
+    )
+
+    environment_metadata_field = MetadataModelField.objects.create(
+        field=metadata_field, content_type=environment_type
+    )
+
+    metadata = Metadata.objects.create(
+        object_id=environment.id,
```
matthewelwell, line 154: One thing I'm keen to check here is whether this link to the environment will still work post import? Or if we could see some unexpected results here?

api/metadata/serializers.py
gagantrivedi: Getting key Error here will mean that something terrible has happened(with the db?) and it should fail as loud as possible in that case, no?

matthewelwell: Yes, you're probably right. Let's leave it as is.

api/metadata/serializers.py
```diff
@@ -0,0 +1,90 @@
...
+            )
+
+        if expected_type == "str" and len(data["field_value"]) > FIELD_VALUE_MAX_LENGTH:
+            raise serializers.ValidationError(
+                f"Value string is too long. Must be less than {FIELD_VALUE_MAX_LENGTH} character"
+            )
+        return data
+
+
+class MetadataSerializerMixin:
```
gagantrivedi: I am not sure if we should still call it a mixin. Thoughts?

api/metadata/serializers.py
gagantrivedi: Yep, good catch

api/metadata/models.py
```diff
@@ -0,0 +1,49 @@
...
+    organisation = models.ForeignKey(Organisation, on_delete=models.CASCADE)
+
+    class Meta:
+        unique_together = ("name", "organisation")
+
+
+class MetadataModelField(AbstractBaseExportableModel):
+    field = models.ForeignKey(MetadataField, on_delete=models.CASCADE)
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    is_required = models.BooleanField(default=False)
```
matthewelwell: Just chatted to @dabeeeenster and @kyle-ssg about this. Ben made a good point that users may not want metadata to be required across the entire organisation. So, for example, on features, you may not want to make JIRA ticket number required across all the projects in your organisation. We need to think about a way to make things 'required with respect to'. 

Can you have a think about this @gagantrivedi ?

api/metadata/models.py
```diff
@@ -0,0 +1,49 @@
...
+from django.db import models
+
+from organisations.models import Organisation
+
+from .fields import GenericObjectID
+
+FIELD_VALUE_MAX_LENGTH = 2000
+
+
+class FieldType(models.TextChoices):
```
matthewelwell, line 28: We probably want to add more data types here, and make them less closely coupled to python data types. So for example, we probably want to split STRING into SINGLE_LINE_STRING and MULTI_LINE_STRING (primarily for the UI's benefit). 

I haven't looked through the rest of the PR to see how valid this might be but I'm imagining we would define custom validation functions against the field types.

dabeeeenster: Also we should have a `URL` data type

api/metadata/models.py
gagantrivedi: Yeah, good point

api/import_export/export.py
matthewelwell: I'm guessing this is resolved by the code in `json_serializers_with_metadata_support` ?

api/metadata/fields.py
```diff
@@ -0,0 +1,29 @@
+from django.core import exceptions
+from django.db import models
+
+
+class GenericObjectID(models.PositiveIntegerField):
```
matthewelwell, line 5: I might need you to run through this in a bit more detail with me @gagantrivedi.

api/metadata/models.py
```diff
@@ -0,0 +1,126 @@
...
+from organisations.models import Organisation
+from projects.models import Project
+
+from .fields import GenericObjectID
+
+FIELD_VALUE_MAX_LENGTH = 2000
+
+METADATA_SUPPORTED_MODELS = ["environment"]
+
+# A map of model name to a function that takes the object id and returns the organisation_id
```
matthewelwell, line 17: I'm not sure why we have two levels here? What relevance does the `"environment"` top level key have?

api/metadata/models.py
```diff
@@ -0,0 +1,126 @@
...
+        ).organisation.id,
+    }
+}
+
+
+class FieldType(models.TextChoices):
+    INTEGER = "int"
+    STRING = "str"
+    BOOLEAN = "bool"
+    FLOAT = "float"
```
matthewelwell: We should get rid of float I think

api/metadata/models.py
```diff
@@ -0,0 +1,126 @@
...
+        "project": lambda project_id: Project.objects.get(
+            id=project_id
+        ).organisation.id,
+    }
+}
+
+
+class FieldType(models.TextChoices):
+    INTEGER = "int"
+    STRING = "str"
```
matthewelwell, line 30: We should split `STRING` into `MULTILINE_STRING` and `SINGLE_LINE_STRING`. Or perhaps just `STRING` and `MULTILINE_STRING` would be sufficient?

api/metadata/models.py
```diff
@@ -0,0 +1,126 @@
...
+
+    name = models.CharField(max_length=255)
+    type = models.CharField(
+        max_length=255, choices=FieldType.choices, default=FieldType.STRING
+    )
+    description = models.TextField(blank=True, null=True)
+    organisation = models.ForeignKey(Organisation, on_delete=models.CASCADE)
+
+    def is_field_value_valid(self, field_value: str) -> bool:
+        if len(field_value) > FIELD_VALUE_MAX_LENGTH:
```
matthewelwell, line 47: If we have multiline strings, we may need to have different max length values depending on the type?

api/metadata/models.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def validate_url(self, field_value: str):
+        try:
+            result = urlparse(field_value)
+            return all([result.scheme, result.netloc])
+        except ValueError:
+            return False
+
+    def validate_str(self, field_value: str):
+        return True
```
matthewelwell, line 71: Do we need to make sure we convert the value to `str` here (and in the other methods?) ?

api/metadata/views.py
```diff
@@ -0,0 +1,124 @@
...
+        ]
+    ),
+)
+class MetadataFieldViewSet(viewsets.ModelViewSet):
+    permission_classes = [MetadataFieldPermissions]
+    serializer_class = MetadataFieldSerializer
+
+    def get_queryset(self):
+        queryset = MetadataField.objects.filter(
+            organisation__in=self.request.user.organisations.all()
```
matthewelwell: Couldn't this be: `organisation__user=self.request.user` ?

api/metadata/permissions.py
```diff
@@ -0,0 +1,77 @@
...
+from organisations.models import Organisation
+
+
+class MetadataFieldPermissions(IsAuthenticated):
+    def has_permission(self, request, view):
+        if not super().has_permission(request, view):
+            return False
+
+        # list is handled by the view
+        if view.action == "list" or view.detail:
```
matthewelwell, line 15: Perhaps not super important but I think we _might_ want to implement a permission check for the `list` action: it would be better for the user to see a 'you do not have permission' vs an empty list, right?

api/metadata/permissions.py
```diff
@@ -0,0 +1,77 @@
...
+            return True
+
+        if view.action == "create":
+            with suppress(Organisation.DoesNotExist):
+                organisation_id = request.data.get("organisation")
```
matthewelwell, line 20: Shouldn't we get this from `self.kwargs` rather than the request data? Can we remove `organisation` from the request data?

api/metadata/serializers.py
```diff
@@ -0,0 +1,134 @@
...
+                    data["content_type"].model
+                ][requirement["content_type"].model]
+            except KeyError:
+                raise serializers.ValidationError(
+                    "Invalid requirement for model {}".format(
+                        data["content_type"].model
+                    )
+                )
+
+                if (
```
matthewelwell: I think this code is unreachable? It should be one level higher, right?

api/metadata/models.py
matthewelwell: I guess this is helping somehow when we want to say that 'metadata field x is required for this project / organisation' ?

api/metadata/models.py
```diff
@@ -0,0 +1,126 @@
...
+
+METADATA_SUPPORTED_MODELS = ["environment"]
+
+# A map of model name to a function that takes the object id and returns the organisation_id
+SUPPORTED_REQUIREMENTS_MAPPING = {
+    "environment": {
+        "organisation": lambda org_id: org_id,
+        "project": lambda project_id: Project.objects.get(
+            id=project_id
+        ).organisation.id,
```
matthewelwell:
```suggestion
        ).organisation_id,
```

api/metadata/serializers.py
matthewelwell: Why not?

api/metadata/serializers.py
```diff
@@ -0,0 +1,134 @@
...
+        raise NotImplementedError()
+
+    def get_required_for_object(
+        self, requirement: MetadataModelFieldRequirement, data: dict
+    ) -> Model:
+        model_name = requirement.content_type.model
+        try:
+            return getattr(self, f"get_{model_name}_from_validated_data")(data)
+        except AttributeError:
+            raise ValueError(f"`get_{model_name}_from_validated_data` does not exists")
```
matthewelwell:
```suggestion
            raise ValueError(f"`get_{model_name}_from_validated_data` method does not exist")
```

matthewelwell: Should this be a `NotImplementedError` ? 🤔

api/metadata/views.py
```diff
@@ -0,0 +1,124 @@
...
+            if organisation_id is None:
+                raise ValidationError("organisation parameter is required")
+            queryset = queryset.filter(organisation__id=organisation_id)
+
+        return queryset
+
+
+@method_decorator(
+    name="list",
+    decorator=swagger_auto_schema(
```
matthewelwell: Couldn't we use a query_serializer for this instead?

api/metadata/views.py
```diff
@@ -0,0 +1,124 @@
...
+        ]
+    ),
+)
+class MetaDataModelFieldViewSet(viewsets.ModelViewSet):
+    permission_classes = [MetadataModelFieldPermissions]
+    serializer_class = MetaDataModelFieldSerializer
+
+    def get_queryset(self):
+        queryset = MetadataModelField.objects.filter(
+            field__organisation_id=self.kwargs.get("organisation_pk")
```
matthewelwell, line 61: Are we doing a permission check for this somewhere else?

api/metadata/serializers.py
matthewelwell: Yep, ok, perhaps this should inherit from `serializers.BaseSerializer` and be named something like: `SerializerWithMetadata`

api/metadata/models.py
gagantrivedi: That's not needed since it's a text field and anything can be converted to string, right?

api/metadata/models.py
gagantrivedi: I don't think people want to go above 2K?

api/metadata/permissions.py
gagantrivedi: As discussed - this view is not nested under organisation

api/metadata/models.py
matthewelwell: Perhaps not, I guess we can deal with this later if we need to.

matthewelwell (approved):
I've approved this now but let's set up a meeting next week (preferably with @dabeeeenster and @kyle-ssg) to review the functionality and make sure that it fulfils the requirements correctly.

PR #1824, opened 2023-01-05
Add hide disabled flags to environments

gagantrivedi:

## Changes

Add `hide_disabled_flags` to environments. 

Covered by unit test cases


matthewelwell:
@gagantrivedi @matthewelwell we need to check if this prevents disabled flags being written to the environment document.

api/environments/identities/models.py
```diff
@@ -107,9 +107,11 @@ def get_all_feature_states(self, traits: typing.List[Trait] = None):
                 current_flag = identity_flags[flag.feature_id]
                 if flag > current_flag:
                     identity_flags[flag.feature_id] = flag
-
-        if self.environment.project.hide_disabled_flags:
-            # filter out any flags that are disabled if configured on the project
+        if (
+            self.environment.hide_disabled_flags
```
matthewelwell: I think I'd like to see this logic as a method on the environment model so that we can unit test it with a parametrised test to fully describe the functionality? 

So the test would look something like: 

```python
@pytest.mark.parametrize(
  "environment_value, project_value, expected_result",
  (
    (True, True, True),
    (True, False, False),
    (False, True, False),
    (False, False, False),
  )
)
def test_get_hide_disabled_flags(environment_value, project_value, expected_result, organisation):
    project = Project.objects.create(name="test", organisation=organisation, hide_disabled_flags=project_value)
    environment = Environment.objects.create(name="test", project=project, hide_disabled_flags=environment_value)
    assert environment.get_hide_disabled_flags() is expected_result
```

matthewelwell: In writing this (which was the reason I actually wrote out the test), I actually think the logic might be slightly wrong here. I think the project value should be a 'default' that can be overridden by the environment. I think we talked about allowing null values for the environment level value to signify 'not set'. 

The reason we're adding this is primarily for a customer to be able to test enabling it in a single environment so they don't have to enable it at the project level for all of their environments. 

So, I think the test cases should actually be: 

```python
@pytest.mark.parametrize(
  "environment_value, project_value, expected_result",
  (
    (True, True, True),
    (True, False, True),
    (False, True, False),
    (False, False, False),
    (None, True, True),
    (None, False, False),
  )
)
```

Then the UI just has a drop down instead of a switcher that has the following options: 

 - use project value
 - hide disabled flags
 - show disabled flags

If you think that using an enum would be more appropriate based on this then I'm fine with that too - although I think my preference would be the null / boolean approach.

api/environments/migrations/0027_add_hide_disabled_flags.py
```diff
@@ -0,0 +1,45 @@
+# Generated by Django 3.2.16 on 2023-01-05 03:01
+
+from django.db.models import Subquery, OuterRef
+from django.db import migrations, models
+
+
+def copy_hide_disabled_flags_config_from_project(apps, schema_editor):
```
matthewelwell: I think my comment above would negate the need for this too?

api/environments/identities/models.py
```diff
@@ -107,9 +107,8 @@ def get_all_feature_states(self, traits: typing.List[Trait] = None):
                 current_flag = identity_flags[flag.feature_id]
                 if flag > current_flag:
                     identity_flags[flag.feature_id] = flag
-
-        if self.environment.project.hide_disabled_flags:
-            # filter out any flags that are disabled if configured on the project
+        if self.environment.get_hide_disabled_flags() is True:
+            # filter out any flags that are disabled
```
gagantrivedi, line 112: I am not sure what CH1245 is?

api/environments/migrations/0027_add_hide_disabled_flags.py
gagantrivedi: Yep

gagantrivedi:
> @gagantrivedi @matthewelwell we need to check if this prevents disabled flags being written to the environment document.

That should be done in the engine, right?

api/environments/identities/models.py
matthewelwell: Oh wow, that's a ticket from a previous project management solution - Clubhouse! We won't have access to that anymore so we should probably just remove this comment. It would be harder to achieve in the database now anyway.

api/features/tests/test_views.py
```diff
@@ -609,43 +609,45 @@ def test_get_flags(self):
...
-        for flag in response_json:
-            assert flag["feature"]["id"] != disabled_flag.id
+    # When
+    api_client.credentials(HTTP_X_ENVIRONMENT_KEY=environment.api_key)
+    response = api_client.get(url)
+
+    # Then
+    assert response.status_code == status.HTTP_200_OK
 
-        # but enabled ones are
```
matthewelwell, line 647: Hmm... I'd like to ensure that the test still includes this logic to check that enabled flags are still returned.

gagantrivedi: Yep, done

matthewelwell (approved):

PR #1825, opened 2023-01-05
improvement/update-python-tooling-for-audit

dabeeeenster:
Fixes https://github.com/Flagsmith/flagsmith/actions/runs/3844899207/jobs/6548429720

### DIFF
--- a/.github/workflows/api-audit.yml
+++ b/.github/workflows/api-audit.yml
@@ -25,6 +25,7 @@ jobs:
               run: |
                 cd api
                 python -m pip install --upgrade pip
+                python -m pip install --upgrade setuptools
                 pip install -r requirements-dev.txt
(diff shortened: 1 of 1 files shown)


.github/workflows/api-audit.yml
```diff
@@ -25,6 +25,7 @@ jobs:
               run: |
                 cd api
                 python -m pip install --upgrade pip
+                python -m pip install --upgrade setuptools
```
matthewelwell, line 31: This seems to indicate then that the issue is in the base install of python that we set up in actions/setup-python above?

dabeeeenster: It is - hence why we need to update it before running the security audit

matthewelwell (approved):

PR #1832, opened 2023-01-10
improvement/date-range-dashboard-traffic

dabeeeenster:
can now go back up to 1 year in graph view

### DIFF
--- a/api/sales_dashboard/views.py
+++ b/api/sales_dashboard/views.py
@@ -137,11 +137,8 @@ def organisation_info(request, organisation_id):
 
     # If self-hosted and running without an Influx DB data store, we don't want to/cant show usage
     if settings.INFLUXDB_TOKEN:
-        date_range = "30d"
-
-        if request.GET.get("date_range"):
-            date_range = request.GET["date_range"]
-            context["date_range"] = date_range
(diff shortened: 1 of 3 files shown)


api/sales_dashboard/views.py
```diff
@@ -137,7 +137,15 @@ def organisation_info(request, organisation_id):
 
     # If self-hosted and running without an Influx DB data store, we don't want to/cant show usage
     if settings.INFLUXDB_TOKEN:
-        event_list, labels = get_event_list_for_organisation(organisation_id)
+        date_range = "30d"
```
matthewelwell: Couldn't all of this be simplified to: 

```python
date_range = request.GET.get("date_range", "30d")
context["date_range"] = date_range
```

The only thing that you lose with this approach I guess is that "date_range" is always set in the context, vs your approach which only sets it if it was included as a GET parameter. But, do we care?

dabeeeenster: haha didnt even know you could do `request.GET.get("date_range", "30d")` so will just do that.

dabeeeenster: OK - updated

matthewelwell (approved):

PR #1849, opened 2023-01-16
feat(analytics/usage-data): Add ability to store usage data in Postgres

gagantrivedi:

#### Add support for running recurring tasks to task processor:
Implementation details: We have created different model classes for Recurring Tasks as well as Recurring Task runs.
Similar to ` run_task` we have created another function called `run_recurring_tasks` that gets called by the `TaskRunner` right after `run_task`, the difference being that we select (for update) all the tasks since, and then we calculate if we should run them or not (captured by the method named `should_execute`) and the result being either success or failure is captured by recurring task runs just like task runs.

To create a recurring task, we use the decorator named `register_recurring_task`  if it's run by processor (`RUN_BY_PROCESSOR`) and the task is deleted by `run_recurring_tasks` if they are not part of the task registry

### API Changes

* API `​/api​/v1​/projects​/{project_pk}​/features​/{id}​/influx-data​/` is deprecated please instead use 
`​/api​/v1​/projects​/{project_pk}​/features​/{id}​/evaluation-data​/`

* API ​`/api​/v1​/organisations​/{id}​/influx-data​/` is deprecated, please use `/api​/v1​/organisations​/{organisation_pk}​/usage-data​/`

* API `​/api​/v1​/organisations​/{id}​/usage​/` is deprecated, please use `/api​/v1​/organisations​/{organisation_pk}​/usage-data​/total-count​/`

Adds unit test cases


api/app/settings/common.py
```diff
@@ -194,6 +202,18 @@
             "CONN_MAX_AGE": DJANGO_DB_CONN_MAX_AGE,
         },
     }
+    if "DJANGO_DB_NAME_ANALYTICS" in os.environ:
```
dabeeeenster, line 206: Really like this idea of being able to specify a separate db

dabeeeenster:
Are there any features/time series aspects of Postgres that we are employing? We also need to write the [downsampling](https://www.influxdata.com/blog/downsampling-influxdb-v2-0/) code!

gagantrivedi:
> Are there any features/time series aspects of Postgres that we are employing? We also need to write the [downsampling](https://www.influxdata.com/blog/downsampling-influxdb-v2-0/) code!

Nah, we are planning to do downsampling manually

dabeeeenster:
> > Are there any features/time series aspects of Postgres that we are employing? We also need to write the [downsampling](https://www.influxdata.com/blog/downsampling-influxdb-v2-0/) code!
> 
> Nah, we are planning to do downsampling manually

Yep sounds good - can make this an async task

api/app_analytics/models.py
```diff
@@ -0,0 +1,93 @@
...
+    resource = models.IntegerField(choices=Resource.choices)
+    total_count = models.PositiveIntegerField()
+    created_at = models.DateTimeField(auto_now_add=True)
+    bucket_size = models.PositiveIntegerField(help_text="Bucket size in minutes")
+
+
+# TODO: move to tasks
+def populate_bucket(bucket_size: int):
+    now = timezone.now()
+    process_from = now - timezone.timedelta(minutes=bucket_size + 1)
```
matthewelwell: I think this would need to be: 

```python
process_from = (now - timezone.timedelta(minutes=bucket_size + 1)).replace(second=0, microsecond=0)
```

Maybe we should also use the scheduled time of the task rather than `now` as well? But obviously that's something that will need the refactor to use a task.

matthewelwell: I'd say we should probably store the bucket sizes in seconds too, just in case we need the flexibility in the future (although I very much doubt it)

api/app_analytics/models.py
```diff
@@ -0,0 +1,93 @@
...
+        APIUsageBucket.objects.create(
+            environment_id=row["environment_id"],
+            resource=row["resource"],
+            total_count=row["count"],
+            bucket_size=bucket_size,
+            created_at=now,
+        )
+
+
+class APIUsageByDay(models.Model):
```
matthewelwell: I wonder if this is a good time to use Proxy models?

api/app_analytics/models.py
matthewelwell: Hmm... actually, I think this is slightly wrong still. To replicate what @dabeeeenster shared, we'd need something like:

```python
current_minute = timezone.now().replace(second=0, microsecond=0)
process_from = current_minute - timezone.timedelta(minutes=31 * bucket_size)
```

Note that this is purely replicating the 30m job that Ben shared - we'd need some generic way of defining how many buckets each task should process (which is obviously also a product of how often we run the task).

api/app_analytics/models.py
```diff
@@ -0,0 +1,93 @@
...
+    created_at = models.DateTimeField(auto_now_add=True)
+    bucket_size = models.PositiveIntegerField(help_text="Bucket size in minutes")
+
+
+# TODO: move to tasks
+def populate_bucket(bucket_size: int):
+    now = timezone.now()
+    process_from = now - timezone.timedelta(minutes=bucket_size + 1)
+    data = (
+        APIUsageRaw.objects.filter(created_at__gt=process_from)
```
matthewelwell: Here we'd need to aggregate by date somehow as we want to process more than one time window each time the task runs - this is only creating the data for a single time window. 

Could we use the `Extract` or `Trunc` functions of the Django ORM perhaps? 

https://docs.djangoproject.com/en/3.2/ref/models/database-functions/#extract
https://docs.djangoproject.com/en/3.2/ref/models/database-functions/#trunc

api/tests/unit/app_analytics/test_tasks.py
```diff
@@ -0,0 +1,71 @@
...
+
+    # Then
+    assert APIUsageBucket.objects.filter(bucket_size=15).count() == 3
+
+    # all three buckets are 15 minutes apart
+    buckets = APIUsageBucket.objects.all().order_by("created_at")
+
+    buckets[0].created_at == buckets[1].created_at + timezone.timedelta(
+        minutes=15
+    ) == buckets[0].created_at + timezone.timedelta(minutes=30)
```
matthewelwell: I think this should be `buckets[2].created_at + ...` ?

gagantrivedi: Yep

api/app/settings/common.py
```diff
@@ -260,6 +280,8 @@
 
 if INFLUXDB_TOKEN:
     MIDDLEWARE.append("app_analytics.middleware.InfluxDBMiddleware")
+if USE_POSTGRES_FOR_ANALYTICS:
```
matthewelwell, line 285: Do we want to prevent people from using Influx and Postgres here?

api/app_analytics/analytics_db_service.py
```diff
@@ -0,0 +1,132 @@
...
+        return get_usage_data_from_local_db(
+            organisation, environment_id=environment_id, project_id=project_id
+        )
+    return get_usage_data_from_influxdb(
+        organisation, environment_id=environment_id, project_id=project_id
+    )
+
+
+def _get_environment_ids_for_org(organisation) -> List[int]:
+    environment_ids = Environment.objects.filter(
```
matthewelwell: Not sure if the comment below is relevant here but wouldn't this be simpler written as something like: 

```python
return [e.id for e in Environment.objects.filter(project__organisation=organisation)]
```

api/app_analytics/analytics_db_service.py
```diff
@@ -0,0 +1,132 @@
...
+    )
+
+
+def get_feature_evaluation_data_from_local_db(
+    feature: Feature, environment_id: int, period: int = 30
+) -> List[FeatureEvaluationData]:
+    feature_evaluation_data = (
+        FeatureEvaluationBucket.objects.filter(
+            environment_id=environment_id,
+            bucket_size=settings.ANALYTICS_READ_BUCKET_SIZE,
```
matthewelwell: We may want to expose this value later down the line but I don't think there's any need to do it now I guess.

api/app_analytics/influxdb_schema.py
```diff
@@ -0,0 +1,43 @@
...
+
+from .dataclasses import FeatureEvaluationData, UsageData
+
+
+class FeatureEvaluationDataSchema(Schema):
+    count = fields.Integer(allow_none=True)
+    day = fields.Date(allow_none=True)
+
+    class Meta:
+        unknown = INCLUDE
```
matthewelwell: Is there a reason we're including unknown here?

api/app_analytics/influxdb_schema.py
```diff
@@ -0,0 +1,43 @@
...
+
+    @pre_load
+    def preprocess(self, data, **kwargs):
+        # the data returned by influx db looks like this:
```
matthewelwell, line 15: 🤦

api/app_analytics/influxdb_schema.py
```diff
@@ -0,0 +1,43 @@
...
+        # {
+        #     "datetime": "2021-01-01",
+        #     "some_feature_name": 10
+        # }
+        valid_data = {"day": data.pop("datetime")}
+        valid_data["count"] = data.popitem()[1]
```
matthewelwell: Couldn't this be rewritten as: 

```python
return {"day": data.pop("datetime"), **data}
```

It feels to me like that would be more readable / less error prone.

api/app_analytics/influxdb_schema.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class UsageDataSchema(Schema):
+    flags = fields.Integer(data_key="Flags", allow_none=True)
+    traits = fields.Integer(data_key="Traits", allow_none=True)
+    identities = fields.Integer(data_key="Identities", allow_none=True)
+    environment_document = fields.Integer(
+        data_key="Environment-document", allow_none=True
+    )
+    day = fields.Date(data_key="name", allow_none=True)
```
matthewelwell, line 38: `data-key="name"` feels wrong here?

api/app_analytics/influxdb_wrapper.py
```diff
@@ -197,6 +200,15 @@ def get_multiple_event_list_for_organisation(
     return dataset
 
 
+def get_usage_data(
```
matthewelwell, line 203: I'm surprised this is a totally new function rather than replacing / updating an existing one?

api/app_analytics/models.py
```diff
@@ -0,0 +1,63 @@
...
+    @classmethod
+    def get_lowercased_name(cls, resource: int) -> str:
+        member = next(filter(lambda member: member.value == resource, cls), None)
+        if not member:
+            raise ValueError("Invalid resource: {resource}")
+
+        return member.name.lower()
+
+    @classmethod
+    def get_from_resource_name(cls, resource: str) -> int:
```
matthewelwell, line 23: I guess here you could do something more like: 

```python
try:
    return getattr(cls, resource.upper().replace("-", "_")).value
except (KeyError, AttributeError):
    # default case (maybe just throw the exception)
```

api/app_analytics/models.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class APIUsageRaw(models.Model):
+    environment_id = models.PositiveIntegerField()
+    created_at = models.DateTimeField(auto_now_add=True)
+    host = models.CharField(max_length=255)
+    resource = models.IntegerField(choices=Resource.choices)
+
+    class Meta:
+        index_together = (("environment_id", "created_at"),)
```
matthewelwell, line 37: Are there alternative index options that would be better here? I'm not sure about the efficiency of a generic index on a datetime field?

api/app_analytics/permissions.py
```diff
@@ -0,0 +1,21 @@
...
+        ):
+            return True
+
+        raise PermissionDenied(
+            "User does not have sufficient privileges to perform this action"
+        )
+
+    def has_object_permission(self, request, view, obj):
+        # Object level permissions are not required for the usage data API
+        return NotImplementedError
```
matthewelwell: shouldn't this `raise NotImplementedError()` rather than return the class?

api/app_analytics/tasks.py
```diff
@@ -0,0 +1,151 @@
...
+    if environment is None:
+        return
+    APIUsageRaw.objects.create(
+        environment_id=environment.id,
+        resource=resource,
+        host=host,
+    )
+
+
+def _get_api_usage_source_data(
```
matthewelwell, line 127: Being picky but I prefer to see the important functions in the module at the top, and any private / helper functions beneath. So, I'd rather see the `populate_bucket_size_task` defined here and these private / helper functions defined below.

api/app_analytics/tasks.py
```diff
@@ -0,0 +1,151 @@
...
+            .annotate(count=Sum("total_count"))
+        )
+    return (
+        FeatureEvaluationRaw.objects.filter(filters, bucket_size=source_bucket_size)
+        .values("environment_id", "feature_name")
+        .annotate(count=Sum("evaluation_count"))
+    )
+
+
+def get_start_of_current_bucket(bucket_size: int) -> datetime:
```
matthewelwell: I think this logic is technically restricted to a maximum bucket size of 60 minutes, right? Maybe that's ok, but we should probably raise an exception if we try to use a larger bucket than that for future prevention?

api/app_analytics/tasks.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def populate_api_usage_bucket(
+    bucket_size: int, run_every: int, source_bucket_size: int = None
+):
+    start_of_current_bucket = get_start_of_current_bucket(
+        bucket_size,
+    )
+
+    for i in range(1, (run_every // bucket_size) + 1):
```
matthewelwell: Think it's probably worth adding some comments here - quite a lot of maths going on! 😄

api/app_analytics/views.py
```diff
@@ -77,3 +95,38 @@ class SelfHostedTelemetryAPIView(CreateAPIView):
     permission_classes = ()
     authentication_classes = ()
     serializer_class = TelemetrySerializer
+
+
+class UsageDataAPIViewSet(viewsets.GenericViewSet):
```
matthewelwell: What's the reasoning for using a viewset here without any of the mixins?

api/organisations/views.py
```diff
@@ -135,10 +135,16 @@ def remove_users(self, request, pk):
         serializer.save()
         return Response(serializer.data, status=200)
 
-    @action(detail=True, methods=["GET"])
+    @swagger_auto_schema(
+        deprecated=True,
+        operation_description="Please use ​​/api​/v1​/organisations​/{organisation_pk}​/usage-data​/total_count​/",
```
matthewelwell: can we standardise to use hyphens rather than underscores in URL paths?

api/task_processor/models.py
```diff
@@ -53,14 +56,42 @@ def schedule_task(
...
+                task_identifier=self.task_identifier,
+                run_every=self.run_every,
+                scheduled_for=self.scheduled_for + self.run_every,
+            )
+            .exclude(id=self.id)
+            .exists()
+        )
+
+    def schedule_next_run(self) -> "Task":
+        next_task_schedule = self.scheduled_for + self.run_every
```
matthewelwell: We should perhaps error if self.run_every is None?

api/app_analytics/tasks.py
```diff
@@ -0,0 +1,151 @@
...
+if settings.USE_POSTGRES_FOR_ANALYTICS:
+
+    @register_task_handler(run_every=60)
+    def populate_bucket_size_task(
+        bucket_size: int, run_every: int, source_bucket_size: int = None
+    ):
+        populate_api_usage_bucket(bucket_size, run_every, source_bucket_size)
+        populate_feature_evaluation_bucket(bucket_size, run_every, source_bucket_size)
+
+    populate_bucket_size_task.delay(settings.ANALYTICS_READ_BUCKET_SIZE, 60)
```
matthewelwell: Is there a way that on startup we can get the task processor to start the first task here instead of having to do it manually? It should be able to check the registry, right?

api/task_processor/processor.py
```diff
@@ -25,6 +25,10 @@ def run_tasks(num_tasks: int = 1) -> typing.List[TaskRun]:
         task_runs = []
 
         for task in tasks:
+            if task.run_every is not None and not task.is_next_run_scheduled:
```
matthewelwell: I guess the alternative here would involve having a shared registry, then we could check each recurring task in the registry to see when it was last run and subsequently run it if the time period has passed. Although this would again be subject to transactional issues. 

One thing that I was thinking when we were chatting yesterday be worth investigating is having a separate daemon thread that's started by the management command. 

So you'd have to update the registry to include information about whether it's recurring or not but then you'd have a thread that looks something like: 

```python
class RecurringScheduler(threading.Thread):
    def run(self, *args, **kwargs):
        while True:
            recurring_tasks = get_recurring_tasks_from_registry()  # returns a tuple (perhaps) of (str) task name and (bool) whether it's recurring or not
            for recurring_task in recurring_tasks:
                if recurring_task[1]:
                    schedule_future_tasks(name=recurring_task[1], num_tasks_to_schedule=10)  # note that I guess we'd need to know how many are already scheduled in the future
```

On first glance, however, it feels like this would be less robust that your approach tbh. There would likely be a race condition where this thread schedules an extra task while one is already being processed. 

Perhaps some food for thought though.

matthewelwell: The other slight concern is that the code at the top of the `_run_task()` function is double checking that the task hasn't already been run. Here, if that is the case, we'd end up with 2 scheduled tasks for the future, right?

api/tests/unit/app_analytics/test_tasks.py
```diff
@@ -0,0 +1,322 @@
...
+
+    assert len(buckets) == runs_every // bucket_size
+    # and
+    start_time = timezone.now().replace(minute=0, second=0, microsecond=0)
+    for bucket in buckets:
+        start_time = start_time - timezone.timedelta(minutes=bucket_size)
+        assert bucket.created_at == start_time
+
+
+# @pytest.mark.freeze_time("2023-01-19T09:00:00+00:00")
```
matthewelwell: Is there a reason this is commented out?

matthewelwell (commented):
Think I've been through everything here now (although I've only given the tests a very cursory glance!). I've added a few comments.

api/app_analytics/influxdb_schema.py
gagantrivedi: No, that would be incorrect, right? That will return something like: `{'day': '2021-01-01', 'some_feature_name': 10}`

api/app/settings/common.py
gagantrivedi: Yep

api/app_analytics/models.py
gagantrivedi: That's the most common access pattern for this though, maybe more patterns will emerge once we do benchmarking

api/app_analytics/views.py
gagantrivedi: Ah, it's because we are not using any of the functionality they add

api/task_processor/models.py
gagantrivedi: yep

api/app_analytics/influxdb_schema.py
gagantrivedi: No, that's how influx returns the data https://github.com/Flagsmith/flagsmith/pull/1849/files#diff-2e5003f9f8f9c9021686c11d021382fd42a786b9f7ed8ce95dea6ed403b14cf9L196

api/app_analytics/influxdb_wrapper.py
gagantrivedi: Yeah, because the current API depends on the structure of the data returned by `get_multiple_event_list_for_organisation`

api/app_analytics/models.py
gagantrivedi: Yes

api/task_processor/processor.py
gagantrivedi: > The other slight concern is that the code at the top of the `_run_task()` function is double checking that the task hasn't already been run. Here, if that is the case, we'd end up with 2 scheduled tasks for the future, right?

Are you talking about the current solution?

api/app_analytics/tasks.py
gagantrivedi: But we still need to pass the args/kwargs that the function needs, another way would be to use a different decorator that does that

api/tests/unit/app_analytics/test_tasks.py
gagantrivedi: Yeah, it was redundant

api/app_analytics/influxdb_schema.py
matthewelwell: Ugh, that link doesn't work when you have files marked as 'viewed' it just takes me to the PR - GH really doesn't handle that very well! Can you link to the main branch perhaps? (assuming this hasn't changed in this PR?)

api/app_analytics/influxdb_schema.py
```diff
@@ -0,0 +1,41 @@
...
+        unknown = EXCLUDE
+
+    @pre_load
+    def preprocess(self, data, **kwargs):
+        # the data returned by influx db looks like this:
+        # {
+        #     "datetime": "2021-01-01",
+        #     "some_feature_name": 10
+        # }
+        return {"day": data.pop("datetime"), "count": data.popitem()[1]}
```
matthewelwell: I don't quite understand the benefit here of using `data.popitem()[1]` vs just regular dictionary access. Like, since we're returning a new dictionary, couldn't this just be: 

```python
return {"day": data["datetime"], "count": data["count"]}
```

I don't quite understand why we need to empty the `data` dict.

api/app_analytics/tasks.py
matthewelwell: I guess we're relying on the validation in the function below for this but, technically, this function also won't work for a `bucket_size` value greater than 60, right?

api/app_analytics/tasks.py
matthewelwell: Can we add these?

api/app_analytics/influxdb_schema.py
gagantrivedi: Yeah, because here: https://github.com/Flagsmith/flagsmith/blob/main/api/app_analytics/influxdb_wrapper.py#L196

api/app_analytics/influxdb_schema.py
gagantrivedi: No, because the key for `count`(in the data returned by influx) is `feature_name` and we don't have access to feature name here

.github/workflows/api-pull-request.yml
```diff
@@ -62,6 +62,7 @@ jobs:
             - name: Run Tests
               env:
                   DATABASE_URL: "postgres://postgres:postgres@localhost:${{ job.services.postgres.ports[5432] }}/postgres"
+                  ANALYTICS_DATABASE_URL: "postgres://postgres:postgres@localhost:${{ job.services.postgres.ports[5432] }}/postgres"
```
matthewelwell: A couple of points here: 

 * Should we at least use a different database here? 
 * Does the migration process by default migrate all databases?

api/scripts/run-docker.sh
```diff
@@ -60,6 +63,8 @@ elif [ "$1" == "dump-organisation-to-s3" ]; then
     dump_organisation_to_s3 "$2" "$3" "$4"
 elif [ "$1" == "dump-organisation-to-local-fs" ]; then
     dump_organisation_to_local_fs "$2" "$3"
+elif [ "$1" == "migrate_analytics_db" ]; then
```
matthewelwell: Bit of Sophie's choice here! I'd rather all of these were consistent when it comes to '_' or '-' but we've used '_' for migrate_identities... 🤔 

How difficult would it be to change from migrate_identities to migrate-identities ? It's only our SaaS platform that uses it, right, so it should be easy enough.

.github/workflows/api-pull-request.yml
gagantrivedi: > Does the migration process by default migrate all databases?


Pyest does

api/scripts/run-docker.sh
gagantrivedi: Yep, done

api/app_analytics/influxdb_schema.py
matthewelwell: Sorry, yes, of course! Maybe, just for clarity then though, we should do: 

```python
feature_name, count = data.popitem()[1]
return {"day": data.pop("datetime"), "count": count}
```

flake8 may complain about the unused variable, in which case we could just do: 

```python
_, count = data.popitem()[1]
return {"day": data.pop("datetime"), "count": count}
```

(but then I'm not convinced that's _that_ much more readable than what you had.

api/app_analytics/influxdb_schema.py
matthewelwell: Ugh, yeah, I have no idea why that is `"name"` how ridiculous 😠

api/app_analytics/views.py
matthewelwell: But then why use a viewset at all in that case? Why not use a GenericAPIView for example? I'm not sure what using the `Viewset` class adds here?

api/app_analytics/views.py
```diff
@@ -77,3 +95,38 @@ class SelfHostedTelemetryAPIView(CreateAPIView):
...
+    http_method_names = ["get", "head", "options", "trace"]
+
+    def _get_organisation(self):
+        organisation_pk = self.kwargs.get("organisation_pk")
+        return Organisation.objects.get(pk=organisation_pk)
+
+    @swagger_auto_schema(
+        responses={200: UsageTotalCountSerializer()},
+    )
+    @action(detail=False, methods=["GET"])
```
matthewelwell: I think (as per my comment [here](https://github.com/Flagsmith/flagsmith/pull/1849#discussion_r1093403925)) we need to add the `url_path` argument here?

api/app_analytics/tasks.py
matthewelwell: Hmm... yes, interesting. I like the idea of using the decorator instead. 

I think actually, in reviewing this again, this call to `delay` is incorrect? I think it needs to `args` and/or `kwargs` as iterable / dict arguments rather than just passing the arguments directly like this.

api/task_processor/processor.py
matthewelwell: Yes, because you're scheduling the next task before calling `_run_task` so the check to verify we haven't already run the task (and subsequently scheduled a future task) isn't happening until you've already scheduled the future task. 

... ah, but I guess that the check at `Task.is_next_run_scheduled` should cover us for that case?

gagantrivedi: > ... ah, but I guess that the check at Task.is_next_run_scheduled should cover us for that case?

Yep, exactly

api/app_analytics/views.py
gagantrivedi: Oh, it's because we have to nest it under `organisations_router` and router only supports view sets

api/app_analytics/views.py
gagantrivedi: Ah yes, good catch

api/app_analytics/tasks.py
gagantrivedi: Yep, moved it to a decoartor

api/task_processor/management/commands/runprocessor.py
```diff
@@ -22,6 +23,10 @@ class Command(BaseCommand):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
+        # Used by the `register_recurring_task` decorator
+        # to decide wether it should register the task or not.
+        os.environ["RUN_BY_PROCESSOR"] = "True"
```
gagantrivedi, line 32: We can move it to environment variables, but I am kinda using it as a replacement of https://stackoverflow.com/questions/24827748/django-significance-of-run-main-environment-variable , what do you think?

api/task_processor/decorators.py
```diff
@@ -68,3 +67,33 @@ def run_in_thread(*, args: typing.Tuple = (), kwargs: typing.Dict = None):
...
+        task_identifier = f"{task_module}.{task_name}"
+        register_task(task_identifier, f)
+        task = Task.schedule_task(
+            schedule_for=timezone.now(),
+            task_identifier=task_identifier,
+            run_every=run_every,
+            args=args,
+            kwargs=kwargs,
+        )
+        task.save()
```
gagantrivedi, line 96: #TODO: do update or create

api/app_analytics/views.py
matthewelwell: Well, we don't have to do we? We can just add something like the following, right? 

```
urlpatterns = [
    ...
    path("/organisations/<int:organisation_id>/usage/", UsageDataAPIView),
    ...
]
```

I'm not saying we definitely _should_ but I think it might be the neater option?

api/task_processor/decorators.py
```diff
@@ -68,3 +67,42 @@ def run_in_thread(*, args: typing.Tuple = (), kwargs: typing.Dict = None):
...
+        task_identifier = f"{task_module}.{task_name}"
+        register_task(task_identifier, f)
+        task = Task.schedule_task(
+            schedule_for=timezone.now(),
+            task_identifier=task_identifier,
+            run_every=run_every,
+            args=args,
+            kwargs=kwargs,
+        )
+        task, _ = Task.objects.update_or_create(
```
matthewelwell: I'm not sure why we're doing `task = Task.schedule_task(...)` immediately followed by `Task.objects.update_or_create(...)` here?

api/task_processor/management/commands/runprocessor.py
matthewelwell: I don't fully understand the use case - can't we use the existing `TASK_RUN_METHOD` setting?

api/tests/unit/app_analytics/test_tasks.py
```diff
@@ -0,0 +1,318 @@
...
+        resource=Resource.FLAGS,
+    )
+    # update created_at
+    event.created_at = when
+    event.save()
+
+    return event
+
+
+@pytest.mark.freeze_time("2023-01-19T09:09:47.325132+00:00")
```
matthewelwell, line 39: Ah, this is cool!

api/tests/unit/app_analytics/test_tasks.py
```diff
@@ -0,0 +1,318 @@
...
+        _create_api_usage_event(environment_id, now - timezone.timedelta(minutes=1 * i))
+        # create events in some other environments as well - just to make sure
+        # we don't aggregate them in the same environment
+        _create_api_usage_event(999, now - timezone.timedelta(minutes=1 * i))
+
+    # Next, let's go 1 hr back in the past and run this
+    freezer.move_to(timezone.now() - timezone.timedelta(hours=1))
+    populate_api_usage_bucket(bucket_size, run_every=timedelta(minutes=60))
+
+    # Then - we should have four buckets
```
matthewelwell, line 58: Shouldn't we have 5 buckets? I think we agreed that we would do `(run_every // bucket size) + 1` ?

api/tests/unit/app_analytics/test_tasks.py
```diff
@@ -0,0 +1,318 @@
...
+    assert buckets[7].total_count == 15
+
+
+@pytest.mark.parametrize(
+    "bucket_size, runs_every_minutes",
+    [(15, 60), (10, 60), (10, 30), (30, 30), (60, 60), (10, 10), (60, 60 * 4)],
+)
+@pytest.mark.freeze_time("2023-01-19T09:09:47.325132+00:00")
+@pytest.mark.django_db(databases=["analytics"])
+def test_populate_api_usage_bucket(freezer, bucket_size, runs_every_minutes):
```
matthewelwell: I'm unsure why we have these 2 separate tests (although I'm not necessarily complaining) but one is called `test_populate_api_usage_bucket_15m_bucket` which seems to test more specifics about the buckets and this one is just called `test_populate_api_usage_bucket` which just tests that the correct number of buckets are created with the correct start times. Should we add some more documentation here about what each test's purpose is?

api/tests/unit/app_analytics/test_tasks.py
```diff
@@ -0,0 +1,318 @@
...
+    )
+    assert (
+        FeatureEvaluationRaw.objects.filter(
+            environment_id=environment_id, feature_name="feature2", evaluation_count=20
+        ).count()
+        == 1
+    )
+
+
+def _create_feature_evaluation_event(environment_id, feature_name, count, when):
```
matthewelwell: If possible, I'd prefer to see this at the bottom of the module.

api/tests/unit/app_analytics/test_tasks.py
```diff
@@ -0,0 +1,318 @@
...
+    assert buckets[3].total_count == 15
+
+    # Now, let's move forward 1hr and run this again
+    freezer.move_to(timezone.now() + timezone.timedelta(hours=1))
+    populate_feature_evaluation_bucket(bucket_size, run_every=timedelta(minutes=60))
+
+    # Then - we should have another four buckets created by the second run
+    buckets = (
+        FeatureEvaluationBucket.objects.filter(
+            bucket_size=15, environment_id=environment_id, feature_name=feature_name
```
matthewelwell:
```suggestion
            bucket_size=bucket_size, environment_id=environment_id, feature_name=feature_name
```

api/tests/unit/app_analytics/test_tasks.py
gagantrivedi: No, we only do `run_every//bucket_size` the fifth bucket is still open, right?

api/tests/unit/app_analytics/test_tasks.py
gagantrivedi: Yeah, I have updated the name

api/task_processor/management/commands/runprocessor.py
matthewelwell: Let's be a bit more explicit with this comment as to why it's necessary. Something along the lines of: 

"The task decorator runs in the global scope so will be executed whenever other e.g. manage.py commands are run and will try to save the task to the database which we want to avoid for non DB dependent commands."

api/task_processor/decorators.py
matthewelwell: Let's remove the call to `Task.schedule_task` and only have the `update_or_create` call. Let's also add a comment to explain why (because of the chance of multiple processor instances running).

api/tests/unit/app_analytics/test_tasks.py
matthewelwell: My concern is that we could miss events on the join between processor intervals. So, for example if the task to populate the buckets runs at say 10:15:01, it will start processing (backwards) from the bucket that is bounded by 10 and 10:15. If there is a raw event that happens at 10:14:59 but doesn't get written to the RawEvent table until 10:15:01 then it could be missed. 

As discuss, it sounds like this is not an issue as the event would be written with a date of 10:15:01 and hence will be counted in the next bucket.

The downside to this is that we are not necessarily accurately tracking the request time but I don't think that this matters in our use case. Do you agree @dabeeeenster ?

api/task_processor/migrations/0007_auto_20230217_1409.py
```diff
@@ -14,21 +13,24 @@ class Migration(migrations.Migration):
...
                 ('scheduled_for', models.DateTimeField(blank=True, default=django.utils.timezone.now, null=True)),
-                ('run_every', models.DurationField()),
                 ('task_identifier', models.CharField(max_length=200)),
                 ('serialized_args', models.TextField(blank=True, null=True)),
                 ('serialized_kwargs', models.TextField(blank=True, null=True)),
+                ('run_every', models.DurationField()),
             ],
-            bases=(django_lifecycle.mixins.LifecycleModelMixin, models.Model),
+        ),
+        migrations.RemoveField(
```
matthewelwell: This can probably just be removed from a previous migration, right? We haven't deployed this code anywhere yet.

matthewelwell:
Some things we should consider here: 

 1. How do we use the RecurringTask functionality to schedule more 'cron' like tasks? e.g. Run the first hour of the first day of each month. 
 2. What about tasks that take longer than run_every? I think this would result in tasks backing up as the RecurringTask row would be locked so nothing else would pick it up and 'overlap'. This is probably fine but we may want to note this down in a comment somewhere. 
 3. Can ANALYTICS_DATABASE_URL be exactly the same as DATABASE_URL?
 4. Should we create our own migrate command to check for the env var? So, rather than having to run `python manage.py migrate --database analytics` in only situations that we know the analytics db exists, we could have a new management command like `python manage.py migrateanalyticsdb` that just subclasses `migrate` and checks for the existence of `ANALYTICS_DATABASE_URL` and does nothing if it doesn't exist. 

Answers (to populate as necessary): 

1. (see 2)
2. 1 and 2 are more or less related to each other, both require a resource free to run them as when they are needed to be run, the solution that I can think of is maybe to create a thread for each cron if they are time sensitive (although that still depends on the scheduler executing our thread) but all of that will depend on the number of recurring tasks that we have and how sensitive they are
3. Yes. 
4. Done

api/task_processor/migrations/0007_auto_20230217_1409.py
gagantrivedi: Yeah, this already has been removed. This is some old version :confused:

api/requirements-dev.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with python 3.10
+# This file is autogenerated by pip-compile with python 3.9
```
matthewelwell: This should always be generated with 3.10.

matthewelwell (approved):

PR #1885, opened 2023-02-02
feat(mv_v2_evaluation): Add use_mv_v2_evaluation

gagantrivedi:
Add use_mv_v2_evaluation in order to support consistent mv evaluation for identities

## Changes
Add `use_mv_v2_evaluation` to environment in order to provide consistent MV evaluation in local vs server mode

Add unit test cases


gagantrivedi:
Depends on https://github.com/Flagsmith/flagsmith-engine/pull/152

gagantrivedi:
Depends on https://github.com/Flagsmith/flagsmith-engine/pull/152

api/environments/identities/models.py
```diff
@@ -38,6 +38,13 @@ class Meta:
     def natural_key(self):
         return self.identifier, self.environment.api_key
 
+    @property
+    def composite_key(self):
+        return f"{self.environment.api_key}_{self.identifier}"
+
+    def get_hash_key(self, use_mv_v2_evaluation: bool = False) -> str:
```
matthewelwell, line 45: I guess here we don't technically need to pass in the `use_mv_v2_evaluation` argument, right? We can just do: 

```python
return self.composite_key if self.environment.use_mv_v2_evaluation else str(self.id)
```

api/environments/identities/serializers.py
```diff
@@ -106,16 +107,15 @@ def get_feature_state_value(
...
     ) -> typing.Union[str, int, bool]:
         identity = self.context["identity"]
+        environment_api_key = self.context["environment_api_key"]
 
-        if isinstance(identity, Identity):
-            identity_id = identity.id
-        else:
-            identity_id = identity.django_id or identity.identity_uuid
+        environment = Environment.get_from_cache(environment_api_key)
+        hash_key = identity.get_hash_key(environment.use_mv_v2_evaluation)
```
matthewelwell, line 113: Ah, but I guess (based on my previous question) - we wouldn't get it from the cache then. BUT wouldn't we have the environment `select_related` on the identity already?

api/environments/migrations/0029_add_default_value_for_use_mv_v2_evaluation.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("environments", "0028_add_use_mv_v2_evaluation"),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
matthewelwell: Can we combine this with the previous migration? I guess we want a different default to be used on existing environments versus new ones, but we could still just have all the operations in the same migration, right? Or is there a way we can use the `preserve_default` argument? (I think not but worth checking).

api/environments/models.py
```diff
@@ -95,8 +95,17 @@ class Environment(
...
+        help_text=(
+            "If true will exclude flags from SDK which are disabled. NOTE: If set, this"
+            " will override the project `hide_disabled_flags`"
+        ),
+    )
+    use_mv_v2_evaluation = models.BooleanField(
+        default=True,
+        help_text=(
+            "Enable this to have consistent multivariate evaluations across all SDKs(on"
+            " local and sever side mode)"
```
matthewelwell:
```suggestion
            " local and server side mode)"
```

api/environments/models.py
```diff
@@ -95,8 +95,17 @@ class Environment(
...
-        "disabled. NOTE: If set, this will override the project `hide_disabled_flags`",
+        help_text=(
+            "If true will exclude flags from SDK which are disabled. NOTE: If set, this"
+            " will override the project `hide_disabled_flags`"
+        ),
+    )
+    use_mv_v2_evaluation = models.BooleanField(
+        default=True,
+        help_text=(
+            "Enable this to have consistent multivariate evaluations across all SDKs(on"
```
matthewelwell:
```suggestion
            "Enable this to have consistent multivariate evaluations across all SDKs (in"
```

api/environments/identities/serializers.py
gagantrivedi: The major reason was to have the same interface as the engine though, does that make sense?

matthewelwell: Yeah, I see that. It's quite interesting really - I guess there are pros/cons of both options. For some reason, I am still leaning towards the simpler interface here (removing the argument) but I don't think my opinion is strong enough to require a change here.

api/environments/migrations/0029_add_default_value_for_use_mv_v2_evaluation.py
gagantrivedi: Yeah, it's not possible to do in one operation

matthewelwell (approved):

PR #1903, opened 2023-02-06
Add soft delete to segment

matthewelwell:

## Changes

Fixes race condition creating audit log record for FeatureSegment when Segment is deleted. 

Note: I had to update the base class for a bunch of models that were combining the base exportable model and soft delete object since it caused unexpected effects as both were overriding the manager class. 

Manually ran the API, FE and Task Processor locally to verify that the records were correctly written.


matthewelwell:
Superseded by #1909

PR #1908, opened 2023-02-07
Add archived filter to compare environments

kyle-ssg:

## Changes

Adds an archived filter to the compare page.

image

Used the compare feature in production.


matthewelwell:
@kyle-ssg this doesn't seem to be working too well for me in the Uffizzi preview environment. I've noticed a couple of issues so far: 

 1. In a project where there are 2 flags and 2 environments, one archived and one not, when there is a difference between the enabled state of the non-archived flag, I just get a blank screen when I try to compare the 2 environments. 
 2. In the same project, if there are no differences between the features in each of the environments, I don't see the archived button. 

I'm going to try to verify this in a second project to confirm it happens every time but I wanted to write down the reproduction steps first.

matthewelwell:
Yep, confirmed that this is reproducible in a fresh project following the exact steps as above. 

The compare UI without the archived button (when there aren't any changes): 

image

Blank screen (when there _are_ changes): 

image

kyle-ssg:
This looked to be a pre-existing issue attempting to compare to exactly matching environments. Should be fixed @matthewelwell

matthewelwell (approved):

PR #1917, opened 2023-02-10
Prevent AuditLog records with no changes

matthewelwell:

## Changes

Re-add logic to prevent AuditLog records being written if no changes have occurred on the model. This also involved making a couple of other changes: 

 1. I've added AuditLog functionality to the FeatureStateValue model
 2. I've written some custom logic to handle the AuditLog record when updating the priorities for FeatureSegments 

Added unit tests and tested many things manually.


api/audit/constants.py
```diff
@@ -27,6 +26,11 @@
 IDENTITY_FEATURE_STATE_UPDATED_MESSAGE = (
     "Flag state / Remote config value updated for feature '%s' and identity '%s'"
 )
+# fmt: off
```
matthewelwell, line 29: I had to add this due to some line length mismatch between black and flake8.

If a string is longer than 88 characters even on its own line, black will just put it on a single line, which then exceeds the 120 line limit imposed by flake8. 

We should sanitise the different line length limits that we have but I didn't want to dilute this PR with formatting changes.

api/audit/tasks.py
```diff
@@ -73,3 +79,52 @@ def create_audit_log_from_historical_record(
...
+    write to the database). segment_a will be moved from 0 -> 1 using bulk update. This seems fine as we
+    then know b was moved and don't really care that a was moved out of the way. If the user, however moves
+    segment_a to priority 1, the single save will be done on segment_b since it'll be first in the list and
+    segment_a will be bulk updated.
+
+    This is a very simple case, and it likely gets more complicated / difficult to follow if there are more
+    than 2 overrides.
+    """
+
+    # TODO: use previous priorities to show what changed.
```
matthewelwell, line 114: I will do this in a separate PR alongside the rest of the changes to include information about what changed.

api/features/feature_segments/serializers.py
```diff
@@ -63,21 +64,65 @@ def get_is_feature_specific(self, instance: FeatureSegment) -> bool:
...
+    def create(self, validated_data):
+        id_priority_pairs = FeatureSegment.to_id_priority_tuple_pairs(validated_data)
+        (
+            changed,
+            previous_id_priority_pairs,
+            feature_segments,
+        ) = FeatureSegment.update_priorities(id_priority_pairs)
+
+        if changed:
+            create_segment_priorities_changed_audit_log.delay(
```
matthewelwell: I don't love that this is happening in the serializer, however, I put it here for 2 reasons: 

1. I need access to the user / master api key
2. putting it in `models` caused a circular import

I thought about raising it another level to the view, however, that would mean either: (a) calling the `update_priorities` method in the view and just using the serializer for validation; or (b) returning more information from this `create` method. 

Interested to know what your thoughts are on this @gagantrivedi .

api/features/feature_segments/tests/test_views.py
```diff
@@ -134,49 +132,32 @@ def test_delete_feature_segment(segment, feature, environment, client):
     assert not FeatureSegment.objects.filter(id=feature_segment.id).exists()
 
 
-@pytest.mark.parametrize(
-    "client", [lazy_fixture("master_api_key_client"), lazy_fixture("admin_client")]
-)
-def test_audit_log_created_when_feature_segment_created(
```
matthewelwell, line 140: I removed these as I don't really see that we need an audit log created based on FeatureSegment creation. It's only important (I think) to audit when a segment override feature state is created.

api/features/feature_segments/tests/test_views.py
```diff
@@ -194,6 +175,24 @@ def test_priority_of_multiple_feature_segments(
     assert json_response[0]["id"] == feature_segment.id
     assert json_response[1]["id"] == another_feature_segment.id
 
+    expected_user_id = (
+        None if "HTTP_AUTHORIZATION" in client._credentials else admin_user.id
```
matthewelwell: I don't love that we have to access `client._credentials` here but (a) I'm not really sure why they're 'private' in a fixture like this and (b) the alternative is splitting this out into separate tests which seems unnecessary.

api/features/feature_segments/serializers.py
gagantrivedi: I think Ideally it should be in `models` ?, but out of the two I am leaning slightly towards doing it in serializer (because it does not directly affect the data being returned, in the view layer sense)

matthewelwell: Yeah, the issue with putting it in models was the circular import. I _could_ use a local import but that makes testing pretty tough. I'll have another look at it this morning but I'm not sure if I can get it working. The other option I considered was creating a custom signal for when priorities are changed?

api/features/feature_segments/tests/test_views.py
gagantrivedi: Another option would be to pass a boolean to the function, something like:

```
@pytest.mark.parametrize(
    "client, expect_user_id", [(lazy_fixture("master_api_key_client"), False), (True, lazy_fixture("admin_client"))]
)

```

api/features/feature_segments/serializers.py
gagantrivedi: Hmm yeah, tough choice. But I think even with that, the logic for audit logs will still be at multiple places. Considering that, I don't think it's too right now

matthewelwell: Yeah, I have tried to chase down the circular imports and tidy it up but I'm not sure if it's better or worse? 

See here: https://github.com/Flagsmith/flagsmith/pull/1917/commits/3962504bc8de76fd1b2643e09443e1627343f0a3

api/features/feature_segments/tests/test_views.py
matthewelwell: Yeah, that's probably slightly cleaner. Will do that.

matthewelwell: (assuming I rollback the change I made to move the task call to models)

api/features/feature_segments/serializers.py
gagantrivedi: Ah yeah, but I think prefer this because it keeps the audit log logic in models at least, no?

gagantrivedi (approved):

api/features/feature_segments/serializers.py
matthewelwell: Yep, I think on the whole it is a little neater.

PR #1942, opened 2023-02-20
Read only segment override mv value

### LINKED ISSUE #1941: Multivariate values reset when editing for a segment override
When adding adding variants to a feature, if you try to add a segment override, it will allow you to edit the values, however it will immediately reset the values after you click out of the input field. Screen recording below to illustrate the issue. 

https://user-images.githubusercontent.com/14089968/220169534-bd6ca001-5116-4a0b-9634-85ce466b1488.mov

gz#435

kyle-ssg:

## Changes

For segment overrides of a multivariate flag, the value needs to be read only.

Created / edited a segment override.

### DIFF
--- a/frontend/web/components/Highlight.js
+++ b/frontend/web/components/Highlight.js
@@ -127,7 +127,6 @@ class Highlight extends React.Component {
             return <Element {...props}>{children}</Element>;
         }
 
-        debugger
         const html = this.props.preventEscape ? this.state.focus ? this.state.value : this.props.children ? { ...this.state.value } : disabled? defaultDisabledValue : defaultValue
             : escapeHtml(this.state.focus ? this.state.value : this.props.children ? { ...this.state.value } : disabled? defaultDisabledValue : defaultValue);
         return (
--- a/frontend/web/components/Highlight.js
+++ b/frontend/web/components/Highlight.js
@@ -15,6 +15,7 @@ function escapeHtml(unsafe) {
     };
 }
 const defaultValue = { __html: 'Enter a value...' };
+const defaultDisabledValue = { __html: ' ' };
 const collapsedHeight = 110;
 class Highlight extends React.Component {
   state = {
@@ -111,7 +112,7 @@ class Highlight extends React.Component {
     }
 
     render() {
-        const { children, className, element: Element, innerHTML } = this.props;
+        const { children, className, element: Element, innerHTML, disabled } = this.props;
         const props = { ref: this.setEl, className };
 
         if (innerHTML) {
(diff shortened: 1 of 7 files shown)


matthewelwell:
@kyle-ssg should we prevent them from updating the control value too? 🤔 

Also, is there a way to make the disabled state more obvious? It's not really obvious until you click it which then activates the reordering functionality. Not essential but thought I'd check here.

kyle-ssg:
Adjusted styles for read only.


 I think read only control makes sense, and have adjusted but In theory, it does currently let people specify a different value to the environment as the control.

 I wonder if some people would find that useful rather than having to add a variation option just to cater for a particular segment needing a different value.

kyle-ssg:
@dabeeeenster wanted to raise this, have you heard / do you know whether people are overriding the control value for segments? We technically allow it, and are discussing whether we should disable it.

I'm wondering whether an option could be to rename the control label for segment overrides on MV features and let them edit it.

image

dabeeeenster:
I don't know for sure but I would think some people for sure would be doing this

dabeeeenster:
This looks good. I wonder whether you should be able to filter on > 1 environment tho? At the moment selecting a different env deselects the currently selected one. That removes the ability to filter on > 1...

matthewelwell:
Figured I would get the empirical data for us to make a decision here. The following query returns the data for MV features which have been overridden for a segment with a different control value. There are 277 such records across 162 unique features. 

Seems like it's therefore not an insignificant amount of people that are using this functionality (rightly or wrongly) so perhaps we should leave it as editable? As you mentioned though @kyle-ssg perhaps renaming the input label would be a good idea. 

```sql
select
	f.name,
	env_fs.environment_id,
	env_fs.enabled,
	env_fsv.string_value,
	segment_fs.environment_id,
	segment_fs.enabled,
	segment_fsv.string_value
from
	features_feature f
left outer join features_featurestate env_fs on
	env_fs.feature_id = f.id
	and env_fs.feature_segment_id is null
	and env_fs.identity_id is null
inner join features_featurestatevalue env_fsv on
	env_fs.id = env_fsv.feature_state_id
left outer join features_featurestate segment_fs on
	segment_fs.feature_id = f.id
	and segment_fs.feature_segment_id is not null
	and env_fs.environment_id = segment_fs.environment_id
inner join features_featurestatevalue segment_fsv on
	segment_fs.id = segment_fsv.feature_state_id
where
	f."type" = 'MULTIVARIATE'
	and (segment_fsv.string_value != env_fsv.string_value
		or segment_fsv.integer_value != env_fsv.integer_value
		or segment_fsv.boolean_value != env_fsv.boolean_value);
```

kyle-ssg:
Ok, updated 
image

matthewelwell (approved):

PR #1943, opened 2023-02-21
Update vulnerable packages

matthewelwell:
## Changes

The purpose of this PR is to update vulnerable packages as found in the Security Audit [here](https://github.com/Flagsmith/flagsmith/actions/runs/4230464844). 

As a part of this, it was necessary to upgrade many of the parent packages and to bring requirements.txt and requirements-dev.txt in line. 

Note: there are a few packages that have been major versioned, however they are all sub dependencies of libraries that have only been minor versioned so we shouldn't have any issues. 

Unit test coverage should be sufficient.


dabeeeenster (approved):
Presume the formatting updates is because new black/flake8?

matthewelwell:
> Presume the formatting updates is because new black/flake8?

Yeah, new black.

PR #1951, opened 2023-02-23
Return identity / segment override numbers on the feature list endpoint

### LINKED ISSUE #1619: Add UI indication that a feature has at least one segment / identity override on overview page
Example in the form of a mock up: 

Screenshot 2022-11-07 at 15 42 21

matthewelwell:
## Changes

Adds 2 new parameters to the list features endpoint to provide the number of identity / segment overrides for that feature in a given environment. 

Added a unit test.

### DIFF
--- a/api/features/views.py
+++ b/api/features/views.py
@@ -7,7 +7,7 @@ from core.constants import FLAGSMITH_UPDATED_AT_HEADER
 from core.permissions import HasMasterAPIKey
 from django.conf import settings
 from django.core.cache import caches
-from django.db.models import Q, QuerySet
+from django.db.models import Count, Q, QuerySet
 from django.utils.decorators import method_decorator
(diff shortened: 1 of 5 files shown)


api/features/views.py
```diff
@@ -155,6 +177,7 @@ def get_serializer_context(self):
                 ),
                 user=self.request.user,
             )
+        context["environment_id"] = self.request.query_params.get("environment")
```
gagantrivedi: why are we doing this?

matthewelwell: Good question - we don't need it. At first I thought about doing the logic in the serializer but then I realised that was ridiculous and posed an N+1 issue 🤦 

Will remove this line now.

gagantrivedi (approved):

PR #1957, opened 2023-02-25
Add feature validation

### LINKED ISSUE #1915: create feature fails with wrong error
Go to create feature.
Enter an id
Paste a block of text larger than 32k or about into the value
click save
you get an error that the feature id already exists
I believe the correct error should be that the data is too large in the value field.
If you delete parts of the value and keep shrinking it eventually it saves fine.

gz#400

kyle-ssg:

## Changes

Errors when creating/saving a feature will now get surfaced in the UI.

image

image


matthewelwell:
@kyle-ssg do we still need the toast message as well as the field error? I guess we may want the toast message for other errors and it's painful to remove it only for these errors? 

image

kyle-ssg:
Yep, sure. Done!

matthewelwell:
The latest change seems to have removed the new error messages again? 

https://user-images.githubusercontent.com/14089968/221823339-1f5d15d2-9868-4fd6-9760-f662c6b22e3e.mov

kyle-ssg:
Ah, was due to a change in main. Should be added again now.

matthewelwell (approved):

PR #1958, opened 2023-02-25
Add last logged in

### LINKED ISSUE #1854: See when a user was last logged in
As an organization administrator, I want to be able to see when a user was last logged in in order to find stale user accounts that could potentially be removed.

gz#321

kyle-ssg:

## Changes

Adds last login

image

image

Viewed users on staging

### DIFF
--- a/frontend/web/components/pages/OrganisationSettingsPage.js
+++ b/frontend/web/components/pages/OrganisationSettingsPage.js
@@ -204,6 +204,20 @@ const OrganisationSettingsPage = class extends Component {
         />);
     }
 
+    formatLastLoggedIn = (last_login)=> {
+        if(!last_login) return "Never"
+
+        const diff = moment().diff(moment(last_login), "days")
+        if (diff>=30) {
+            return <>
+            {`${diff} days ago`}
+                <br/>
+                <span className="text-small text-muted">{moment(last_login).format("Do MMM YYYY")}</span>
+            </>
+        }
+        return "Within 30 days"
+    }
+
     render() {
         const { name, webhook_notification_email } = this.state;
         const { props: { webhooks, webhooksLoading } } = this;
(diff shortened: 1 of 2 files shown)


kyle-ssg:
@matthewelwell If I'm not mistaken, this timestamp is when the login endpoint was hit. Since user's tokens don't expire, this figure may end up being very misleading.

matthewelwell:
Right, yes, you are correct @kyle-ssg. IIRC, we implemented a change recently on the FE, however, that means the cookies expire after (I think) 30 days, right? So they'd need to login again at least every 30 days? 

Assuming I'm correct here, I'd say that we have a couple of options here: 

1. Leave it as per the functionality that we have but we'd need to change the header here from 'Last Active' to 'Last Login' (maybe with a tool tip that states "users are only required to login every 30 days, this value may be 30 days out of date". 
2. @gagantrivedi, we could look at using the task processor to update this value whenever a user makes a request to the Core API.

IMO (1) would be sufficient, (2) seems like overkill for now and would result in a not insignificant amount of extra load on the processor / database.

kyle-ssg:
I think 30 days is a long time for that figure to be out by if it's used when considering whether to remove seats, I feel like we wouldn't want to give people the impression a user could be inactive when they are in fact using the dashboard.

Though @dabeeeenster might have a different opinion.

dabeeeenster:
We definitely want to show when their token was last used to make an API request, whether via the dashboard or REST or whatever. Could we add an endpoint that the dashboard hits which updates the `last_logged_in` timestamp when the dashboard loads? If the copy is then "last logged in" it is semantically correct. 

It would be nice to have this also show when the API was used but its not that important for now.

dabeeeenster:
@kyle-ssg we're happy to just say "Last logged in" either:

- Within 30 days
- XXX days ago (DD-MMM-YYYY)

kyle-ssg:
Updated

image


image

image

matthewelwell (approved):

PR #1973, opened 2023-02-28
Prevent all model entities being deleted when OneToOneField parent is soft deleted

matthewelwell:
## Changes

Add SoftDelete base class to all models which have a OneToOneField to a model which itself subclasses SoftDeleteObject. 

Fixes issues created by bug in django-softdelete here: https://github.com/scoursen/django-softdelete/issues/99

An alternative option here would be to add a mixin / base model that overrides the delete method and just does nothing / or hard deletes the entity if the `changeset` argument is provided. I feel like it's pretty similar to this solution but perhaps a bit more hacky. The benefit would be that we wouldn't have to make any database changes in this PR.

Added a unit test to verify that subscriptions aren't all deleted when any organisation is deleted.


api/integrations/common/models.py
```diff
@@ -28,7 +27,6 @@ class Meta:
         abstract = True
 
     @hook(AFTER_SAVE)
-    @hook(AFTER_DELETE)
```
gagantrivedi, line 31: Why did we remove this? Is it because `save` gets called anyway? 
yeah, looks like that's why

matthewelwell: Yeah, exactly. Now that it's soft deleted, `delete` performs a `save`. Otherwise it gets called twice (because it still triggers the `post_delete` hook as well).

gagantrivedi:
> ## Changes
> 
> Add SoftDelete base class to all models which have a OneToOneField to a model which itself subclasses SoftDeleteObject.
> 
> Fixes issues created by bug in django-softdelete here: [scoursen/django-softdelete#99](https://github.com/scoursen/django-softdelete/issues/99)
> 
> An alternative option here would be to add a mixin / base model that overrides the delete method and just does nothing / or hard deletes the entity if the `changeset` argument is provided. I feel like it's pretty similar to this solution but perhaps a bit more hacky. The benefit would be that we wouldn't have to make any database changes in this PR.
> ## How did you test this code?
> 
> Added a unit test to verify that subscriptions aren't all deleted when any organisation is deleted.

I think it's okay considering that fact that either this will get fixed in the upstream or we will patch this

matthewelwell:
@gagantrivedi I've updated this to just tell the models to DO_NOTHING when the parent is deleted which I think is slightly neater?

matthewelwell:
> @gagantrivedi I've updated this to just tell the models to DO_NOTHING when the parent is deleted which I think is slightly neater?

This wouldn't work without some additional hacking to get the post_delete signal to trigger when related objects are 'deleted'. I couldn't find anything obvious as to how to get the signal to trigger in this case so I've reverted that change. 

The logic can be seen [here](https://github.com/Flagsmith/flagsmith/pull/1973/commits/4b7ed28e54d8aebe093afe1fd28f8ca2543dbfe6) though for reference.

gagantrivedi (approved):

PR #1975, opened 2023-02-28
Soft delete webhook configuration

matthewelwell:
## Changes

Adds SoftDelete logic to WebhookConfiguration model. 

This model didn't see to be affected by the issues with parent model deletions, however, it could be in the future so I'm adding this to it now to ensure we don't see this. 

Logic is tested in other models.


api/webhooks/models.py
```diff
@@ -1,10 +1,24 @@
...
+
+
+class AbstractBaseExportableWebhookModel(
+    AbstractBaseWebhookModel, AbstractBaseExportableModel
+):
+    class Meta:
+        abstract = True
+
+
+class AbstractBaseSoftDeleteExportableWebhookModel(
```
gagantrivedi, line 20: Where is this being used?

matthewelwell: Here: https://github.com/Flagsmith/flagsmith/pull/1975/files#diff-c2f900907b02278b20496d66a689e7b7b21abb20fc46c850417652e7a8eddaf8L13

gagantrivedi (approved):

PR #1978, opened 2023-03-01
Chores/allow scheduled change requests

kyle-ssg:

## Changes
This splits change requests from scheduling, allowing people to view them separately.

image

image

image

How did you test this code?
Testing with a enterprise and startup org.

### DIFF
--- a/frontend/common/stores/change-requests-store.js
+++ b/frontend/common/stores/change-requests-store.js
@@ -18,7 +18,7 @@ const controller = {
             }).catch(e => API.ajaxHandler(store, e));
     },
     getChangeRequests: (envId, { committed, live_from_after }, page) => {
-        const has4EyesPermission = Utils.getPlansPermission('4_EYES');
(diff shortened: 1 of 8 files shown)


matthewelwell:
I'm getting console errors testing this in Uffizzi. The 'Schedule update' button does nothing and I see the following error in the console. 

image

matthewelwell:
Ah, I have a feeling it's because the workflows endpoints don't exist in Uffizzi... I'll test it locally.

matthewelwell:
This seems to work quite nicely functionally, but some of the UI is a bit confusing. I've highlighted some of the key details here: 

Screenshot 2023-03-01 at 14 58 12

The navigation on the left is also highlighting the 'Change Requests' option which may be confusing for customers that don't technically have access to Change Requests. 

I appreciate that we're trying to reuse the functionality as much as possible but is there any way we can tweak some of these things?

kyle-ssg:
@matthewelwell done

matthewelwell (approved):

PR #1979, opened 2023-03-01
Remove deleted / old versions from overrides data

matthewelwell:
## Changes

Fixes an issue with the logic for including segment / identity override information in the request to list features. Previously, the code did not account for deleted feature states or old versions (although the latter is unnecessary since we don't version segment / identity overrides yet). 

Existing unit test + new test.

### DIFF
--- a/api/features/dataclasses.py
+++ b/api/features/dataclasses.py
@@ -1,0 +1,19 @@
+import typing
+from dataclasses import dataclass
+
+
+@dataclass
+class EnvironmentFeatureOverridesData:
+    """
+    Dataclass to hold the information about the number of overrides for a given feature in a given
+    environment.
+    """
+
+    num_segment_overrides: int = 0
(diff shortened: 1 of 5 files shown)


api/features/dataclasses.py
```diff
@@ -0,0 +1,19 @@
...
+class EnvironmentFeatureOverridesData:
+    """
+    Dataclass to hold the information about the number of overrides for a given feature in a given
+    environment.
+    """
+
+    num_segment_overrides: int = 0
+    num_identity_overrides: typing.Optional[int] = None
+
+    def add_identity_override(self):
```
gagantrivedi, line 18: Why do we not have `add_segment_override` . Ah, okay because it can be None. Maybe we should add a comment here stating why?

matthewelwell: Yep, good point, will do.

gagantrivedi (approved):

PR #1995, opened 2023-03-06
Improve audit page loading

### LINKED ISSUE #1993: Audit log paging is slow and feels broken
1. The audit log generally on first load is slow - maybe an API issue?
2. The paging feels broken due to 1. But we should maybe show a spinner when waiting for a page of results
3. I'm not sure if when turning off/on a filter the list is showing the first page or not? Seems like its not? 

gz#490

kyle-ssg:

## Changes

Page request will revert back to 1 when the environment changes, audit list will now indicate it's loading a new page.

image

### DIFF
--- a/frontend/common/stores/config-store.js
+++ b/frontend/common/stores/config-store.js
@@ -44,7 +44,7 @@ flagsmith.init({
     onChange: controller.loaded,
     api: Project.flagsmithClientAPI,
     cacheFlags: true,
-    realtime: Project.flagsmithRealtime,
+    realtime: true,
     AsyncStorage,
     enableAnalytics: Project.flagsmithAnalytics,
 }).catch(() => {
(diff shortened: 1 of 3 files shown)


dabeeeenster:
@kyle-ssg can we include page number somewhere too? Would be helpful

kyle-ssg:
I'll take a look at keeping that state. Edge paging doesn't bring back that info.

kyle-ssg:
Added

matthewelwell:
> I'll take a look at keeping that state. Edge paging doesn't bring back that info.

This isn't touching edge though so you shouldn't have to keep the state here? On this endpoint you should be able to display how many pages, etc. there are as all the data you need is available to you. The only endpoint that it's not available on is the `/environments/:key/edge-identities/` endpoint. 

This is fine for now, but it would probably be nice to have a view of how many pages there are.

kyle-ssg:
Ok yeah I can see a count have adjusted to use that to calculate the current page

matthewelwell (approved):

PR #2017, opened 2023-03-14
feat(sdk/hide-sensitive-data): Add option to hide sensitive data

gagantrivedi:
Add option to hide "description" from the response of sdk endpoints

## Changes

Add option to hide "description" from the response of sdk endpoints

Adds unit test cases


api/features/serializers.py
```diff
@@ -222,6 +225,15 @@ class Meta:
         writeonly_fields = ("initial_value", "default_enabled")
 
 
+class SDKFeatureSerializer(HideSensitiveFieldsSerializerMixin, FeatureSerializer):
+    sensitive_fields = (
```
gagantrivedi, line 229: All of them are not really sensitive fields, but I can't think of a better name

api/environments/identities/views.py
```diff
@@ -247,7 +249,7 @@ def post(self, request):
         # trait values are serialized correctly
         response_serializer = IdentifyWithTraitsSerializer(
             instance=instance,
-            context={"identity": instance.get("identity")},  # todo: improve this
+            context=self._get_serializer_context(instance.get("identity")),
```
matthewelwell: I guess the alternative here is to do something like: 

```python
self.identity = instance.get("identity") 
```

in the code above and then just use `self.identity` in the `get_serializer_context()` method so we don't have to create this, slightly confusing, `_get_serializer_context` method. What do you think?

matthewelwell (commented):
Some things that I've considered while reviewing this PR: 

 1. Should we just expose a field that allows people to define the fields that they want to hide? Maybe this is dangerous as it would allow them to break things? 
 2. Should we remove them from the response or just null / blank them out? (to avoid breaking changes)

gagantrivedi:
> Some things that I've considered while reviewing this PR:
> 
>     1. Should we just expose a field that allows people to define the fields that they want to hide? Maybe this is dangerous as it would allow them to break things?
> 
>     2. Should we remove them from the response or just null / blank them out? (to avoid breaking changes)

Re 1 yeah, I don't really think that's a good idea, maybe we should have the button that says something like `use edge like response`?

matthewelwell:
> Re 1 yeah, I don't really think that's a good idea, maybe we should have the button that says something like use edge like response?

Yeah, I guess we can be careful with the wording the UI. Perhaps make sure that we provide a big warning saying 'THIS WILL CAUSE A BREAKING CHANGE IN THE API (but won't break any Flagsmith SDKs)'

matthewelwell (approved):

PR #2019, opened 2023-03-14
Add markdown support for butter_bar

### LINKED ISSUE #2013: XSS issue using butter bar feature
Using the `butter_bar` flag available as a flag for the Flagsmith dashboard, it is possible to exercise an XSS vulnerability. 

By giving the flag a value of, e.g. `">.txt`, the dashboard will run the malicious code. 

We can resolve this in the flagsmith dashboard where it renders the value of the butter bar, but it begs the question whether there should be a generic solution available in the flagsmith JS client to sanitise the value of the flag.

kyle-ssg:

## Changes

butter_bar remote config will now be rendered as markdown.

image

- Tested with md links et
- Tested xss style tags, this is handled by https://github.com/remarkjs/react-markdown#security


matthewelwell (approved):
Looks good. Do we need to update any documentation somewhere? 

Side note, mostly for @dabeeeenster: This is one of the things that we'll need to include in some release notes before it gets released. I think we should consider having release branches. I'm not sure how this would work with wanting to deploy to production every time we merge FE code though.

dabeeeenster:
OK how about we label PRs that are relevant for release notes?

PR #2028, opened 2023-03-15
Use postgres endpoints for analytics

### LINKED ISSUE #780: Use Postgres to store Influx Flag Analytics and API traffic data
Main concern is around how we would do the downsampling in Postgres.

kyle-ssg:

## Changes

Replaces influx endpoint usage with new intelligent endpoints that point to the configured datastore automatically.

- Compared Flagmsmith's feature flag evaluations (Which can be seen on the preview URL
- **Note** - I haven't been able to test organisation usage analytics as it's returning empty, we should only merge this once that's been tested / validated.


frontend/api/index.js
```diff
@@ -65,7 +65,10 @@ app.get('/config/project-overrides', (req, res) => {
         { name: 'api', value: process.env.FLAGSMITH_PROXY_API_URL ? '/api/v1/' : process.env.FLAGSMITH_API_URL },
         { name: 'maintenance', value: envToBool('ENABLE_MAINTENANCE_MODE', false) },
         { name: 'flagsmithClientAPI', value: process.env.FLAGSMITH_ON_FLAGSMITH_API_URL },
-        { name: 'disableInflux', value: !envToBool('ENABLE_INFLUXDB_FEATURES', true) },
+        { name: 'disableAnalytics', value: envToBool('DISABLE_ANALYTICS_FEATURES', false) },
+        { name: 'disableInflux', value: !envToBool('ENABLE_INFLUXDB_FEATURES', true) ||
+                envToBool('DISABLE_INFLUXDB_FEATURES', false)
+        },
```
kyle-ssg, line 71: Note, this adds DISABLE_INFLUXDB_FEATURES. It was noted that the docs/frontend use ENABLE_INFLUXDB_FEATURES but the backend uses DISABLE, since we can't just remove the existing env var this will allow us to deprecate it/update the docs.

dabeeeenster (approved):

dabeeeenster:
This is working in staging but I think the data might not be quite right? 

Here's the FE:

image

And here's the DB table:

image

Seems to me like the total for the 16th should be 14 and not 12? Any idea why it's 12 @gagantrivedi ? Is the code SUMing all the records for that day?

dabeeeenster (changes requested):
Waiting on data fix

dabeeeenster (approved):

PR #2034, opened 2023-03-16
Alias preview urls

kyle-ssg:

## Changes

Branches now get a url alias for staging / production preview deployments. Prior to this the url would change per commit.

See new preview urls

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -119,6 +119,7 @@
     "suppress-exit-code": "^1.0.0",
     "terser-webpack-plugin": "^5.3.6",
     "toml": "^3.0.0",
+    "vercel": "^28.16.15",
     "webpack": "5.52.1",
     "webpack-bundle-analyzer": "^4.5.0",
     "webpack-cli": "^4.10.0",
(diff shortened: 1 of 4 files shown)


dabeeeenster:
What's this good for?!

kyle-ssg:
It means that URLs are consistent / don't change per commit.

dabeeeenster:
> It means that URLs are consistent / don't change per commit.

ah great

frontend/package.json
```diff
@@ -119,6 +119,7 @@
     "suppress-exit-code": "^1.0.0",
     "terser-webpack-plugin": "^5.3.6",
     "toml": "^3.0.0",
+    "vercel": "^28.16.15",
```
dabeeeenster, line 122: Should this not be in devDependencies?

kyle-ssg: No, because our actions do npm ci

kyle-ssg: It's the same reason why webpack etc is there too.

dabeeeenster (approved):

matthewelwell (approved):

PR #2038, opened 2023-03-16
Prevent duplicate leads in pipedrive

matthewelwell:
## Changes

This PR includes 2 main changes: 

 * Only create leads in Pipedrive when a user is first created, not whenever the user is saved
 * If there is an existing person, don't create a new one

Added unit tests

### DIFF
--- a/api/integrations/lead_tracking/pipedrive/client.py
+++ b/api/integrations/lead_tracking/pipedrive/client.py
@@ -47,6 +47,17 @@ class PipedriveAPIClient:
             for org in api_response_data["items"]
         ]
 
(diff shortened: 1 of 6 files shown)


api/integrations/lead_tracking/pipedrive/client.py
```diff
@@ -127,4 +138,6 @@ def _make_request(
                 f"Response code was {response.status_code}. Expected {expected_status_code}"
             )
 
+        print(response.json())
```
gagantrivedi: Ah, you missed a print

gagantrivedi (changes requested):

gagantrivedi (approved):

PR #2049, opened 2023-03-21
feat(permission/roles): Add roles to enable better permission management 

gagantrivedi:

## Changes

Create a role that can hold environment/project/organisation permission and can be attached to user or groups

New API Endpoints: 

`​/api​/v1​/organisations​/{organisation_pk}​/roles​/`
`​/api​/v1​/organisations​/{organisation_pk}​/roles​/{role_pk}​/environments-permissions​/`
`​/api​/v1​/organisations​/{organisation_pk}​/roles​/{role_pk}​/groups​/`
`​/api​/v1​/organisations​/{organisation_pk}​/roles​/{role_pk}​/projects-permissions​/
`​/api​/v1​/organisations​/{organisation_pk}​/roles​/{role_pk}​/users​/`

Adds unit test cases


api/organisations/roles/permissions.py
```diff
@@ -0,0 +1,23 @@
...
+from organisations.roles.models import Role
+
+
+class NestedRolePermission(NestedIsOrganisationAdminPermission):
+    def has_permission(self, request, view):
+        if super().has_permission(request, view):
+            roles_pk = view.kwargs.get("role_pk")
+            with suppress(ObjectDoesNotExist):
+                role = Role.objects.get(pk=roles_pk)
+                return role.organisation.id == int(view.kwargs.get("organisation_pk"))
```
matthewelwell, line 19: Shouldn't this do some kind of check that the user also belongs to the organisation?

api/organisations/roles/permissions.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class NestedRolePermission(NestedIsOrganisationAdminPermission):
+    def has_permission(self, request, view):
+        if super().has_permission(request, view):
+            roles_pk = view.kwargs.get("role_pk")
+            with suppress(ObjectDoesNotExist):
+                role = Role.objects.get(pk=roles_pk)
+                if role.organisation.id == int(view.kwargs.get("organisation_pk")):
+                    if view.action == "create":
```
matthewelwell: I'm not entirely sure what's going on here? I think there's a mistake somewhere because we have both paths here returning `True`

api/organisations/roles/views.py
```diff
@@ -0,0 +1,145 @@
...
+        if self.action == "list":
+            project_id = self.request.query_params.get("project")
+            if not project_id:
+                raise ValidationError("'project' GET parameter is required.")
+            q &= Q(project__id=project_id)
+
+        return RoleProjectPermission.objects.filter(q)
+
+    def perform_create(self, serializer):
+        role = Role.objects.get(id=self.kwargs["role_pk"])
```
matthewelwell: Can we just use the pk without the database query here as we have above in the RoleEnvironmentPermissionsViewSet? 

Also, should we have a `perform_update` method here too?

api/organisations/views.py
```diff
@@ -241,13 +244,16 @@ def permissions(self, request):
...
-                "permissions": permission_keys,
-                "admin": request.user.is_organisation_admin(org),
-            }
+
+        organisation_permissions_calculator = OrganisationPermissionsCalculator(
+            pk=org.id
+        )
+        permission_data = organisation_permissions_calculator.get_permission_data(
+            user_id=request.user.id,
+            is_organisation_admin=request.user.is_organisation_admin(org),
```
matthewelwell: Perhaps this will become clear soon but it's somewhat odd to me that we pass in whether the user is an organisation admin to the method on the calculator. It seems like the calculator should handle this?

api/permissions/permissions_calculator.py
```diff
@@ -0,0 +1,271 @@
...
+    def admin(self) -> bool:
+        return (
+            self.is_organisation_admin
+            or self.user.admin
+            or any(group.admin for group in self.groups)
+            or any(role.admin for role in self.roles)
+        )
+
+    @property
+    def permissions(self) -> typing.Set[str]:
```
matthewelwell, line 110: I think this property would benefit from some comments / a docstring - it's not immediately obvious what it's doing.

api/permissions/permissions_calculator.py
```diff
@@ -0,0 +1,271 @@
...
+
+@dataclass
+class PermissionData:
+    user: UserPermissionData
+    groups: typing.List[GroupPermissionData]
+    roles: typing.List[RolePermissionData]
+    is_organisation_admin: bool = False
+
+    @property
+    def admin(self) -> bool:
```
matthewelwell, line 101: I'm a little unclear what the context is for this 'admin' status.

api/permissions/permissions_calculator.py
```diff
@@ -0,0 +1,271 @@
...
+                set(),
+            )
+        )
+
+
+class BasePermissionCalculator:
+    def __init__(self, pk: int):
+        self.pk = pk
+
+    def _get_user_permission_qs(self, user_id: int) -> UserPermissionQs:
```
matthewelwell: I think it would be useful to add some docstrings to these methods.

api/permissions/permissions_calculator.py
```diff
@@ -0,0 +1,271 @@
...
+                lambda a, b: a.union(b),
+                [role.permissions for role in self.roles],
+                set(),
+            )
+        )
+
+
+class BasePermissionCalculator:
+    def __init__(self, pk: int):
+        self.pk = pk
```
matthewelwell: Is the pk we provide here the one for the object? e.g. the project id.

api/permissions/permissions_calculator.py
```diff
@@ -0,0 +1,271 @@
...
+            )
+
+            group_permission_data_objects.append(group_permission_data_object)
+
+        return group_permission_data_objects
+
+    def _get_roles_permission_data(
+        self, user_id: int
+    ) -> typing.List[RolePermissionData]:
+        pass
```
matthewelwell: This seems like an error here?

api/permissions/permissions_calculator.py
```diff
@@ -0,0 +1,271 @@
...
+    def _get_user_permission_qs(self, user_id: int) -> UserPermissionQs:
+        raise NotImplementedError()
+
+    def _get_group_permission_qs(self, user_id: int) -> GroupPermissionQs:
+        raise NotImplementedError()
+
+    def _get_role_permission_qs(self, user_id: int) -> RolePermissionQs:
+        raise NotImplementedError()
+
+    def get_permission_data(
```
matthewelwell: It's interesting that each of the methods from here could be static methods, right? Should we change them to static methods or perhaps just convert them to functions?

api/projects/tags/views.py
```diff
@@ -12,7 +12,7 @@ class TagViewSet(viewsets.ModelViewSet):
 
     def get_queryset(self):
         project = get_object_or_404(
-            self.request.user.get_permitted_projects(["VIEW_PROJECT"]),
+            self.request.user.get_permitted_projects("VIEW_PROJECT"),
```
matthewelwell: This could use the `VIEW_PROJECT` constant now, right?

matthewelwell: In fact, there are a lot of places that we're still using `"VIEW_PROJECT"`.

api/projects/views.py
```diff
@@ -222,9 +220,7 @@ def get_user_project_permissions(request, **kwargs):
     user_id = kwargs["user_pk"]
 
     project_permissions_calculator = ProjectPermissionsCalculator(kwargs["project_pk"])
-    user_permissions_data = (
-        project_permissions_calculator.get_user_project_permission_data(user_id)
-    )
+    user_permissions_data = project_permissions_calculator.get_permission_data(user_id)
```
matthewelwell: It's _slightly_ misleading that this is named `user_permissions_data` when the `get_permission_data` method is returning a `PermissionData` object, not a `UserPermissionData` object.

api/users/models.py
```diff
@@ -212,134 +197,44 @@ def get_organisation_role(self, organisation):
...
     ) -> bool:
-        permission_query = Q(permissions__key=permission_key) | Q(admin=True)
-        return (
-            UserEnvironmentPermission.objects.filter(
-                Q(environment=environment, user=self) & permission_query
-            ).exists()
-            or UserPermissionGroupEnvironmentPermission.objects.filter(
-                Q(environment=environment, group__users=self) & permission_query
-            ).exists()
+        return environment in get_permitted_environments_for_user(
```
matthewelwell: This could use `self.get_permitted_environments`, right?

matthewelwell (commented):
I've been through most of this PR and left a few comments. There are a few test files I still need to get through but probably best to discuss / address these comments first.

api/permissions/permissions_calculator.py
gagantrivedi: Yep

api/organisations/roles/permissions.py
gagantrivedi: that happens here: https://github.com/Flagsmith/flagsmith/pull/2049/files#diff-f03378d8df1afe7feaf70e473e92d9ed78246e341cd8dea664037f9a25f9220eR13

api/organisations/roles/permissions.py
gagantrivedi: Yeah, my bad, I forgot to remove it

api/organisations/roles/views.py
gagantrivedi: Yep

api/permissions/permissions_calculator.py
gagantrivedi: It is used like this: https://github.com/Flagsmith/flagsmith/blob/main/api/projects/views.py#L232

api/organisations/views.py
gagantrivedi: Yeah, this is sort of a weird one since you can be a project/environment admin using the usual `UserPermissionGroupProjectPermission` or `UserProjectPermission` but you can be organisation permission only using `UserOrganisation`, it will make more sense if you look at the refactored permisson_calculator now I think

api/permissions/permissions_calculator.py
gagantrivedi: Yep, done

api/organisations/roles/permissions.py
matthewelwell: Should we add a comment to confirm that? I'm keen to minimise comments but maybe here it would be useful.

api/features/permissions.py
```diff
@@ -11,16 +11,17 @@
 )
 from features.models import Feature, FeatureState
 from projects.models import Project
+from projects.permissions import CREATE_FEATURE, VIEW_PROJECT
 
 ACTION_PERMISSIONS_MAP = {
-    "retrieve": "VIEW_PROJECT",
+    "retrieve": VIEW_PROJECT,
     "destroy": "DELETE_FEATURE",
```
matthewelwell: Is there a reason that DELETE_FEATURE has been left as a string?

api/permissions/permissions_calculator.py
matthewelwell: But what is the context? A user can only be an admin in the context of an object, right? I think some description / comment here would be useful.

api/permissions/permissions_calculator.py
matthewelwell: As per the above, this has to be in the context of some object, right? Like a project or an environment? Maybe the issue I'm struggling to understand is that this whole class is in the context of an entity?

api/permissions/permissions_calculator.py
```diff
@@ -0,0 +1,308 @@
...
+    environment_permission_qs = _EnvironmentPermissionQs(environment_id, user_id)
+    return PermissionData(
+        user=get_user_permission_data(environment_permission_qs.user_qs),
+        groups=get_groups_permission_data(environment_permission_qs.group_qs),
+        roles=get_roles_permission_data(environment_permission_qs.role_qs),
+    )
+
+
+def _is_organisation_admin(organisation_id, user_id) -> bool:
+    # # TODO: implement
```
matthewelwell: This still needs doing?

api/permissions/permissions_calculator.py
```diff
@@ -0,0 +1,308 @@
...
+            reduce(
+                lambda a, b: a.union(b),
+                [role.permissions for role in self.roles],
+                set(),
+            )
+        )
+
+
+@dataclass
+class _ProjectPermissionQs:
```
matthewelwell: I think we might need to name these classes slightly better. I got confused for a while how we were iterating over a dataclass (e.g. [here](https://github.com/Flagsmith/flagsmith/pull/2049/files#diff-c77369100711cd5d00b41280f9d72531d7e4d31b749ad5e8b287cf068001177aR190)) because I assumed that `UserPermissionQs` was also one of these dataclasses. 

I think we can just be more verbose with the naming, right? Something like `_ProjectPermissionQuerysetDataclass` or `_ProjectPermissionQuerysets`

api/tests/unit/permissions/permission_service/test_get_permitted_environments_for_user.py
```diff
@@ -0,0 +1,177 @@
...
+    # First, let's assert that the user does not have access to any environment
+    for permission in EnvironmentPermissionModel.objects.all().values_list(
+        "key", flat=True
+    ):
+        assert (
+            get_permitted_environments_for_user(test_user, project, permission).count()
+            == 0
+        )
+    # Next, let's give user some permissions using `user_permission`
+    environment_permission_using_user_permission.permissions.add(VIEW_ENVIRONMENT)
```
matthewelwell: **Minor**

I think we could perhaps simplify this as follows: 

```
permissions_as_user = [VIEW_ENVIRONMENT, UPDATE_FEATURE_STATE]
environment_permission_using_user_permission.permissions.add(*permissions_as_user)
```

Then we can reuse this list below.

api/users/tests/test_models.py
```diff
@@ -173,47 +166,6 @@ def test_has_organisation_permission_is_false_when_user_does_not_have_permission
...
-@pytest.mark.parametrize(
-    "user_permission_keys, group_permission_keys, expected_keys",
-    (
-        ([], [], set()),
-        ([CREATE_PROJECT], [], {CREATE_PROJECT}),
-        ([], [CREATE_PROJECT], {CREATE_PROJECT}),
-        ([CREATE_PROJECT], [CREATE_PROJECT], {CREATE_PROJECT}),
-    ),
-)
-def test_get_permission_keys_for_organisation(
```
matthewelwell, line 186: I'm not sure why this test was removed?

gagantrivedi: Because the method `get_permission_keys_for_organisation` was removed in favour of the permission calculator module

api/permissions/permissions_calculator.py
gagantrivedi: I have added the description to the class itself - does that makes sense?

gagantrivedi:
Superseded by https://github.com/Flagsmith/flagsmith/pull/2232

PR #2083, opened 2023-03-30
chore: Static flags

kyle-ssg:

- [ ] I have run [`pre-commit`](https://docs.flagsmith.com/deployment/locally-api#pre-commit) to check linting

## Changes

- ``npm run static:generate`` will generate a ``flagsmith.json`` with the Staging flags from **Flagsmith Website** on app.flagsmith.com
- ``npm run static:copy`` will copy the ``flagsmith.json`` file to the project.
- ``npm run static:remove`` will remove the ``flagsmith.json`` file from the project.
- If the project has a ``flagsmith.json`` file the frontend will run in offline mode, it will never fetch flags from our API, it will always use the feature states defined in ``flagsmith.json``. Due to the above, this will also not use SSE/realtime.

_Please describe._


kyle-ssg:
@matthewelwell @dabeeeenster, all that needs to happen in addition to this PR is to call ``npm run static:copy`` wherever relevant. It'll result in the SDK running completely offline with ``flagsmith.json``.

If I'm not mistaken, I think there might also be a need to not call it (or to call ``npm run static:remove``) when FLAGSMITH_ON_FLAGSMITH_API_KEY doesn't match flagsmith.json

dabeeeenster:
We need to write a github action that runs regularly (suggest once a day) that:

- Gets the latest version of the code
- Generates the static json file
- Diff's that file with the one committed in the code
- If they are the same, do nothing
- If they are different, create a PR with the new static file

kyle-ssg:
AFAIK this is no longer valid as we'll be using feature import / export to achieve this.

Issue #1523, opened 2022-10-10
Add task to prune processed tasks table


dabeeeenster:
The number of days before pruning record should be configurable by an env var, and default to 7 days.

matthewelwell:
Analysing the impact from the job that ran on staging to clear out ~1M tasks, we can see a small spike in RDS CPU usage. 

![image](https://github.com/Flagsmith/flagsmith/assets/14089968/8f20fb8b-36e8-496e-99dc-cb0d49b6421a)

This is with a batch size of 2000 records.

gagantrivedi:
Implemented here: https://github.com/Flagsmith/flagsmith/pull/2095

PR #2095, opened 2023-04-05
chore: clean up old tasks

matthewelwell:
## Changes

Adds a recurring task to clean up old, completed tasks. 

Some considerations here that it would be good to get some extra feedback on: 

 1. The code here so far just relies on the first scheduled run of the task to clear up what's already there. It feels like this is perhaps safer than writing a custom migration which will inevitably cause the build to timeout. The concern, however, is that we have slightly less control over _when_ it runs. As a sub thought on this one, the first time it runs will take a significant amount of time. The monitoring will think the task is unhealthy but it will do nothing about it (I'm 99% sure).
 3. I've assumed for now that we don't want to delete failed tasks (in case we need to retry them). I'm not sure how many of these there are at the moment so it might be that we want to start deleting these in the future (since we've got the retention period to rely on here anyway)
 4. I'm deleting the tasks in smaller batches to try to reduce the impact on the database deleting huge numbers of records but this will draw out the time it takes to delete the records. 

Added unit tests


api/task_processor/models.py
```diff
@@ -131,6 +132,10 @@ class Meta:
 
     @property
     def should_execute(self) -> bool:
+        now = timezone.now()
+        if self.first_run_time and self.first_run_time > now.time():
```
gagantrivedi: Should we add a comment here? It took me more than ideal time to understand this, but when I finally did, it sort of became obvious. What do you think?

gagantrivedi (approved):
Minor comment

matthewelwell:
Fixes #1523

api/task_processor/models.py
matthewelwell: I've added a comment and slightly tweaked the logic, as I realised that the logic that was there was susceptible to issues if the task start time slipped (over a long period of time) such that it drifted beyond midnight it would never run again.

infrastructure/aws/production/ecs-task-definition-task-processor.json
```diff
@@ -126,6 +126,10 @@
                 {
                     "name": "PIPEDRIVE_LEAD_LABEL_EXISTING_CUSTOMER_ID",
                     "value": "d828a3a0-bf34-11ed-b08d-b12d3497f5c4"
+                },
+                {
+                    "name": "TASK_RETENTION_DAYS",
+                    "value": "280"
```
matthewelwell, line 128: We originally released the task processor on 22/9/2023 meaning that we should have around 295 days of tasks in the database. Setting the retention period will allow us stop the amount of tasks increasing at least while slowly allowing us to chip away at clearing up old tasks. The plan will be to slowly reduce this retention period down to the 30 day default.

gagantrivedi (approved):

PR #2097, opened 2023-04-06
Group Admins

### LINKED ISSUE #1886: Add the ability to configure admins within a group who can manage the membership of other users in that group
Organisation admins should be able to define multiple admins in a user group that can then control the membership of other users in that group.

Along with this, there should be a setting that allows all members of the group to add other members to the group.

kyle-ssg:

## Changes

- Migrate GET,Delete groups to TypeScript
- Add a delete group admin / make group admin endpoint
- Allow all users to see the manage groups page (org admins will see the standard manage org page instead)

image

image


matthewelwell:
@kyle-ssg a few things I've noticed: 

 1. It doesn't seem to work going from Group Admin -> User. I can't see that it ever makes a request to `.../remove-admin`? (see screen recording attached)
 2. The UI is quite confusing. I think perhaps the best step would be to only show the users here that are already members of the group. Then we can show a separate form to add new users or something. I appreciate that's perhaps a larger chunk of work but could we perhaps tidy up what we have a little bit? Feels to me like the admin status should be a checkbox rather than a drop down since it can only take a value of on or off. 

https://user-images.githubusercontent.com/14089968/230393561-a1c3c2f2-2f53-48eb-aaa5-82e0550b2b5d.mov

kyle-ssg:
RE remove admin, this is now sorted.

Have adjusted the UI of this, the initial consideration was like organisation users where perhaps in future there could be other roles but yeah this seems unlikely.

image

matthewelwell:
Screenshot 2023-04-11 at 10 37 50

@kyle-ssg I'm not seeing the admin switcher? 

Note: this is using the vercel preview against production.

kyle-ssg:
@matthewelwell, this was due to the group_admins flag being disabled in production - I've adjusted the PR to also hide the "Admin" heading in this case.

It can be tested against the staging deployment above or by enabling the group_admins flag.

matthewelwell:
@kyle-ssg this looks better, and the removal works now. The new 'Add user' workflow is slightly confusing though as you need to add them and then hit 'Update Group' but there's no indication that you need to do that. Could we add an indication that users haven't been added until they hit 'Update Group'?

frontend/web/components/modals/CreateGroup.js
```diff
@@ -162,11 +169,15 @@ const CreateGroup = class extends Component {
                     }}
                     value={name}
                     onChange={(e) =>
-                      this.setState({ name: Utils.safeParseEventValue(e) })
+                      this.setState({
+                        emailEdited: true,
```
matthewelwell: This should be `nameEdited` or `groupNameEdited`, right? We're not editing any emails here unless I'm mistaken?

frontend/web/components/modals/CreateGroup.js
```diff
@@ -318,7 +346,10 @@ const CreateGroup = class extends Component {
                                 type='button'
                                 disabled={!(isAdmin || email !== yourEmail)}
                                 id='remove-feature'
-                                onClick={() => this.toggleUser(id)}
+                                onClick={() => {
+                                  this.toggleUser(id)
+                                  this.setState({ isDeleted: true })
```
matthewelwell: Probably `this.setState({ usersRemoved: true })` would be better here, right?

matthewelwell (commented):
This looks great. I love that we've factored in adding / removing users and toggling the admin. 

I've added a couple of minor comments about some variable naming.

frontend/web/components/modals/CreateGroup.js
novakzaballa: My mistake, done.

frontend/web/components/modals/CreateGroup.js
novakzaballa: Done

novakzaballa:
@matthewelwell @kyle-ssg A new commit has been added for displaying a confirmation dialog whenever the user has changed any value(s), please review again.

kyle-ssg:
@novakzaballa can't approve as it's my PR but latest changes look good from a code perspective 👍

matthewelwell:
@kyle-ssg @novakzaballa the functionality here looks good but I encountered a bug on the API when I was testing it all the way through. I've put a fix for this up here: #2139

matthewelwell:
@novakzaballa sorry - just found one more bug on the frontend. The interstitial modal also displays when you successfully update the group. 

https://user-images.githubusercontent.com/14089968/233934166-84a3d941-074f-4a5d-98c3-0d76e7b6087c.mov

novakzaballa:
> @novakzaballa sorry - just found one more bug on the frontend. The interstitial modal also displays when you successfully update the group.
> 
>  Screen.Recording.2023-04-24.at.08.54.48.mov

@matthewelwell thanks for testing it. Corrected

matthewelwell (approved):

PR #2106, opened 2023-04-10
Archive flags and remove beta page

kyle-ssg:

## Changes

We aren't making use of the beta features page, this PR removes it. This PR will also allow us to archive the following flags:

- auto_seats
- beta_features
- case_sensitive_flags
- compare_environments
- delete_account (Was put up due to an API issue)
- delete_environment (Was put up due to an API issue)
- delete_organisation (Was put up due to an API issue)
- delete_project (Was put up due to an API issue)
- feature_name_regex
- flag_analytics
- scaleup_audit
- update_feature_state_permission
- usage_chart (Based on ENV var anyway)
- usage_filter
- use_admin_identity_featurestates
- view_identities_permission

E2E /  checked each removal


frontend/common/utils/utils.tsx
```diff
@@ -294,7 +266,7 @@ const Utils = Object.assign({}, require('./base/_utils'), {
         break
       }
       case 'AUTO_SEATS': {
-        valid = isScaleupOrGreater && Utils.getFlagsmithHasFeature('auto_seats')
```
matthewelwell, line 297: Appreciate this isn't really part of this PR but I'm not sure this logic is actually 100% correct? We don't allow enterprise SaaS customers to auto upgrade their seats (for commercial reasons). The code here: https://github.com/Flagsmith/flagsmith/blob/main/frontend/common/utils/utils.tsx#L289 seems to suggest that this would allow them to do so?

kyle-ssg: Ah, ok yep that makes sense - will adjust as part of this 👍

frontend/common/utils/utils.tsx
```diff
@@ -294,7 +268,7 @@ const Utils = Object.assign({}, require('./base/_utils'), {
...
+        valid = isScaleupOrGreater && !isEnterprise
```
matthewelwell, line 271: Interestingly, this is logically equivalent to `isScaleUp` because there's nothing above scale up except enterprise, but maybe this allows us to handle additional plans in the future (although I'm not sure we will need to).

matthewelwell (approved):
approved with a minor comment

PR #2118, opened 2023-04-12
feat: Update static flags

dabeeeenster:
Waiting on https://github.com/Flagsmith/flagsmith/pull/2083 to be merged


.github/workflows/frontend-update-static-flags.yml
```diff
@@ -0,0 +1,53 @@
...
+
+      - name: Create Pull Request for new flags if required
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_PRIVATE_ACCESS_TOKEN }}
+        if: steps.get_changes.outputs.changed != 0
+        run: |
+          git config user.email "ben.rometsch@flagsmith.com"
+          git config user.name "Ben Rometsch"
+          git stash
+          git branch static-flags-updated
```
matthewelwell: I think you can remove this line and change the one below to `get checkout -B static-flags-updated`.

matthewelwell: Also, what will happen here if the workflow runs again while the existing PR is still open, e.g. if it's opened on a Friday or something?

.github/workflows/frontend-update-static-flags.yml
```diff
@@ -0,0 +1,53 @@
...
+          echo "changed=$(git status --porcelain | wc -l)" >> $GITHUB_OUTPUT
+
+      - name: Create Pull Request for new flags if required
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_PRIVATE_ACCESS_TOKEN }}
+        if: steps.get_changes.outputs.changed != 0
+        run: |
+          git config user.email "ben.rometsch@flagsmith.com"
+          git config user.name "Ben Rometsch"
+          git stash
```
matthewelwell: Do we need the `stash` and `stash pop` ? As far as I can tell, `git checkout -B <branch name>` doesn't overwrite local changes.

.github/workflows/frontend-update-static-flags.yml
```diff
@@ -0,0 +1,53 @@
...
+          npm ci
+          npm run static:generate
+          echo "changed=$(git status --porcelain | wc -l)" >> $GITHUB_OUTPUT
+
+      - name: Create Pull Request for new flags if required
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_PRIVATE_ACCESS_TOKEN }}
+        if: steps.get_changes.outputs.changed != 0
+        run: |
+          git config user.email "ben.rometsch@flagsmith.com"
```
matthewelwell: Does this need to be your user? Can we have some bot like user?

.github/workflows/frontend-update-static-flags.yml
dabeeeenster: Hmmm yeah that's a good point. Maybe dont run it at weekends? Doing the work to see if a PR already exists sounds problematic?

matthewelwell: Yeah, ok, we can go with that for now and see. I guess if it does occur, all that will happen is it will fail when it tries to push the commit to the branch since we're using the same branch name and it will fail because of conflicts.

matthewelwell (approved):
Approving to merge into another branch to consolidate PRs.

PR #2119, opened 2023-04-12
Fix issue generating task to write AuditLog for Edge Identity Overrides

matthewelwell:
## Changes

Fix issue writing task to create AuditLog record for Edge Identity Overrides when the value is numeric. 

Added a new unit test.


api/requirements.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with Python 3.11
+# This file is autogenerated by pip-compile with Python 3.10
```
khvn26: Noticed this and wanted to highlight it – you might want to check your local Python env.

khvn26 (approved):

api/requirements.txt
matthewelwell: Ah yep, good catch. Thanks!

matthewelwell: I've updated my environment and re run `pip-compile` just in case there were any differences.

gagantrivedi (approved):

PR #2120, opened 2023-04-12
Support domain override from environment

khvn26:

## Changes

Flagsmith API has two new settings:
- `DOMAIN_OVERRIDE` (`FLAGSMITH_DOMAIN` env var): forces a domain in emails regardless of `django.contrib.sites` content.
- `DEFAULT_DOMAIN`: a fallback value for cases when neither `DOMAIN_OVERRIDE` nor `SITE_ID` is specified.

- Covered with unit tests.
- Sent a couple invites manually and checked correctness of the URLs in the resulting emails.

### DIFF
--- a/api/app/settings/common.py
+++ b/api/app/settings/common.py
@@ -870,3 +870,9 @@ SOFTDELETE_CASCADE_ALLOW_DELETE_ALL = False
 
 # Used for serializing and deserializing GenericForeignKey(used in metadata) using the natural key of the object
 SERIALIZATION_MODULES = {"json": "import_export.json_serializers_with_metadata_support"}
+
+# Controls the app domain used in emails (currently invites and change requests).
+# If set, domain stored with `django.contrib.sites` is disregarded.
+DOMAIN_OVERRIDE = env.str("FLAGSMITH_DOMAIN", "")
+# Used when no Django site is specified.
+DEFAULT_DOMAIN = "app.flagsmith.com"
--- a/api/organisations/invites/models.py
+++ b/api/organisations/invites/models.py
@@ -61,9 +61,9 @@ class Invite(AbstractBaseInviteModel):
         db_table = "users_invite"
 
     def save(self, *args, **kwargs):
-        # send email invite before saving invite
+        super().save(*args, **kwargs)
+        # send email invite after saving invite
         self.send_invite_mail()
-        super(Invite, self).save(*args, **kwargs)
 
     def get_invite_uri(self):
         return f"{get_current_site_url()}/invite/{str(self.hash)}"
--- a/api/organisations/invites/models.py
+++ b/api/organisations/invites/models.py
@@ -5,7 +5,12 @@ from django.core.mail import EmailMultiAlternatives
 from django.db import models
 from django.template.loader import get_template
 from django.utils import timezone
-from django_lifecycle import BEFORE_CREATE, LifecycleModelMixin, hook
+from django_lifecycle import (
+    AFTER_CREATE,
+    BEFORE_CREATE,
+    LifecycleModelMixin,
+    hook,
+)
 
 from app.utils import create_hash
 from organisations.invites.exceptions import InviteLinksDisabledError
@@ -47,7 +52,7 @@ class InviteLink(
             raise InviteLinksDisabledError()
 
 
-class Invite(AbstractBaseInviteModel):
+class Invite(LifecycleModelMixin, AbstractBaseInviteModel):
     email = models.EmailField()
     invited_by = models.ForeignKey(
         FFAdminUser, related_name="sent_invites", null=True, on_delete=models.CASCADE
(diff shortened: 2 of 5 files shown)


.github/workflows/api-pull-request.yml
```diff
@@ -70,6 +70,5 @@ jobs:
           ANALYTICS_DATABASE_URL:
             postgres://postgres:postgres@localhost:${{ job.services.postgres.ports[5432] }}/analytics
           DJANGO_SETTINGS_MODULE: app.settings.test
-          LOG_LEVEL: INFO
```
matthewelwell: We should move this (and the later change to the test) to another PR as well as the init_data.yml removal as well.

khvn26:
Docs PR [here](https://github.com/Flagsmith/flagsmith-docs/pull/339).

api/app/settings/common.py
```diff
@@ -870,3 +870,9 @@
 
 # Used for serializing and deserializing GenericForeignKey(used in metadata) using the natural key of the object
 SERIALIZATION_MODULES = {"json": "import_export.json_serializers_with_metadata_support"}
+
+# Controls the app domain used in emails (currently invites and change requests).
+# If set, domain stored with `django.contrib.sites` is disregarded.
+DOMAIN_OVERRIDE = env.str("FLAGSMITH_DOMAIN", "")
+# Used when no Django site is specified.
+DEFAULT_DOMAIN = "app.flagsmith.com"
```
matthewelwell, line 878: Was there a reason to use `app.flagsmith.com` here vs just `flagsmith.com` ?

api/tests/unit/core/test_helpers.py
```diff
@@ -1,27 +1,75 @@
+from typing import TYPE_CHECKING
+
+import pytest
 from core.helpers import get_current_site_url
 from django.contrib.sites.models import Site
 
+if TYPE_CHECKING:
+    from pytest_django.fixtures import SettingsWrapper
+else:
+    SettingsWrapper = None
```
matthewelwell: Interesting. I've not seen this before, historically, I've done: 

```
if TYPE_CHECKING:

def test_something(settings: "SettingsWrapper"):
    ...
```

Is there a benefit to having this `else` clause here instead?

matthewelwell:
Note: we also need to make sure this is included in the release notes for the next version. Improving the release notes process is something I'm still working on...!

api/tests/unit/core/test_helpers.py
khvn26: Both are valid, I just prefer less quotes. UPD: changed to quotes-style for consistency, will use it in the future.

api/app/settings/common.py
khvn26: It was the default argument value – see [here](https://github.com/Flagsmith/flagsmith/blob/e009e58e0ef4bbc9d645b8b12098abafb7b41178/api/core/helpers.py#L5).

Do the cloud invites/workflows point to `flagsmith.com` instead? Is there a working codepath for this setting at all?

.github/workflows/api-pull-request.yml
khvn26: Done, see [PR](https://github.com/Flagsmith/flagsmith/pull/2121).

api/app/settings/common.py
matthewelwell: >Do the cloud invites/workflows point to flagsmith.com instead? 

Yes, because the domain is set in the default Site object in our SaaS deployment. 

I actually don't think that default will ever be used unless you explicitly delete the default Site object that is created when you run the migrations.

khvn26: What’s the use case for overriding the domain with having a Site?

Do you think we should rather allow the users to define a fallback value and recommend them to clear the Sites/the SITE_ID setting?

khvn26: Resolved this in DMs with @matthewelwell.

For the time being we want to keep the config options as flexible as possible. Further down the road we might want to deprecate django_sites usage, but this will require us to revisit the [installation flow](https://docs.flagsmith.com/deployment/hosting/locally-api#environments-with-no-direct-console-access-eg-heroku-ecs) for some environments.

The docs have been updated to reflect the change ([PR](https://github.com/Flagsmith/flagsmith-docs/pull/340)).

matthewelwell (approved):
Approving but, as discussed this morning, it would be good to manually test this to verify the email templates are correctly generated.

matthewelwell (approved):

api/organisations/invites/models.py
```diff
@@ -60,14 +65,10 @@ class Meta:
...
 
-    def save(self, *args, **kwargs):
-        # send email invite before saving invite
-        self.send_invite_mail()
-        super(Invite, self).save(*args, **kwargs)
-
     def get_invite_uri(self):
         return f"{get_current_site_url()}/invite/{str(self.hash)}"
 
+    @hook(AFTER_CREATE)
```
gagantrivedi, line 71: Do we have a way for the user to send the invite email again (if the first one fails for some reason)?

khvn26: We do, see the [resend action](https://github.com/Flagsmith/flagsmith/blob/b1c0ffca9e237635bde0e6bbe8ba06ea26a8d855/api/organisations/invites/views.py#L122-L123).

gagantrivedi: Gotcha

PR #2125, opened 2023-04-14
logging(task/processor): Add logging to identify task causing DB error

gagantrivedi:

## Changes

Wrap run_tasks in try/except to avoid thread from dying 
Catch and log DatabaseError caused by tasks to identify it

Running task processor manually on a task that raises an Integrity Error

### DIFF
--- a/api/task_processor/processor.py
+++ b/api/task_processor/processor.py
@@ -3,6 +3,7 @@ import traceback
 import typing
 
 from django.db import transaction
+from django.db.utils import DatabaseError
 from django.utils import timezone
 
 from task_processor.models import (
@@ -89,6 +90,13 @@ def _run_task(task: Task) -> typing.Optional[typing.Tuple[Task, TaskRun]]:
 
         task_run.finished_at = timezone.now()
         task.mark_success()
+    except DatabaseError as e:
+        logger.error(
+            "Database error while running task %s:  %s",
+            task.id,
+            e,
+            exc_info=True,
+        )
     except Exception:
         task.mark_failure()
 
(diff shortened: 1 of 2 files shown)


api/task_processor/processor.py
```diff
@@ -89,6 +90,13 @@ def _run_task(task: Task) -> typing.Optional[typing.Tuple[Task, TaskRun]]:
 
         task_run.finished_at = timezone.now()
         task.mark_success()
+    except DatabaseError as e:
+        logger.error(
+            "Database error while running task %s:  %s",
+            task.id,
+            e,
+            exc_info=True,
+        )
```
khvn26, line 99: I think we have to add `task.mark_falure()` here?

api/task_processor/threads.py
```diff
@@ -24,7 +27,10 @@ def __init__(
     def run(self) -> None:
         while not self._stopped:
             self.last_checked_for_tasks = timezone.now()
-            run_tasks(self.queue_pop_size)
+            try:
+                run_tasks(self.queue_pop_size)
+            except Exception as e:
+                logger.exception(e)
             run_recurring_tasks(self.queue_pop_size)
```
khvn26, line 34: Do we want to wrap this line too?

gagantrivedi: No, that's fine, this whole try/catch will go away once we figure out the task that is causing the issue

api/task_processor/processor.py
gagantrivedi: No, there is no point doing that since we don't have access to the database any more here

matthewelwell (approved):

PR #2126, opened 2023-04-15
Improvement/Bring docs into monorepo

dabeeeenster:
This PR will bring in the docs to the main repo for many benefits!

Once the PR is merged we can set up the vercel integration to auto build docs.flagsmith.com.

Some notes:

- Removed husky pre-commit as the linting is now managed by the mono-repo with https://pre-commit.com/


docs/docusaurus.config.js
```diff
@@ -0,0 +1,174 @@
...
+    presets: [
+        [
+            'docusaurus-preset-openapi',
+            {
+                docs: {
+                    sidebarPath: require.resolve('./sidebars.js'),
+                    routeBasePath: '/',
+                    // Please change this to your repo.
+                    editUrl: 'https://github.com/flagsmith/flagsmith-docs/tree/main',
+                    lastVersion: 'current',
```
khvn26: We'll need to change this, right?
```suggestion
                    editUrl: 'https://github.com/flagsmith/flagsmith/tree/main/docs/',
                    lastVersion: 'current',
```

.github/workflows/docs-pull-request.yml
```diff
@@ -0,0 +1,38 @@
+name: Docs Lint/Link Checks
+
+on:
+  push:
+    branches:
+      - main
```
matthewelwell: This workflow yaml file is called `docs-pull-request.yml` but we're running it on pushes to the main branch? Something needs to change here.

.github/workflows/docs-pull-request.yml
```diff
@@ -0,0 +1,38 @@
...
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          cache: 'npm'
+          cache-dependency-path: 'docs/package-lock.json'
+
+      - name: npm install
+        run: npm ci --include=dev --ignore-scripts
+      - name: Run docusuarus build to check for broken links etc
+        if: github.event_name == 'pull_request'
```
matthewelwell: Related to the above, I think this means that the workflow won't actually do anything other than `npm ci` on pushes to main.

matthewelwell (changes requested):

.github/workflows/docs-pull-request.yml
dabeeeenster: Updated

.github/workflows/docs-pull-request.yml
dabeeeenster: Updated

matthewelwell (approved):

PR #2127, opened 2023-04-17
fix(task-processor/deadlock): use Postgres function to fetch task for processing 

gagantrivedi:

## Changes
Remove transaction in favour of the pl/pgsql function used to lock and return the tasks to process.

Add/update unit test cases


api/tests/unit/task_processor/test_unit_task_processor_processor.py
```diff
@@ -172,8 +174,35 @@ def test_run_task_runs_task_and_creates_task_run_object_when_failure(db):
...
+
+    # Then
+    for task_run in task_runs:
+        assert task_run.result == TaskResult.FAILURE
+        assert task_run.started_at
+        assert task_run.finished_at is None
+        assert task_run.error_details is not None
+
+    task.refresh_from_db()
+    assert not task.completed
```
matthewelwell: Should we check here that `is_locked` is False too?

matthewelwell (commented):
This looks good to me. It would definitely be good to have some more documentation somewhere around the SQL query though. I wonder whether we should pull the SQL out to separate `.sql` files and add some comments that can be read easily. I wonder if the RunSQL base command that PostgresOnlyRunSQL inherits from allows you to pass a path to a sql file instead of a string?

gagantrivedi:
> This looks good to me. It would definitely be good to have some more documentation somewhere around the SQL query though. I wonder whether we should pull the SQL out to separate `.sql` files and add some comments that can be read easily. I wonder if the RunSQL base command that PostgresOnlyRunSQL inherits from allows you to pass a path to a sql file instead of a string?

Yeah, good point

api/task_processor/migrations/sql/get_recurring_tasks_to_process.sql
```diff
@@ -0,0 +1,27 @@
...
+    row_to_return task_processor_recurringtask;
+BEGIN
+    -- Select the tasks that needs to be processed
+    FOR row_to_return IN
+        SELECT *
+        FROM task_processor_recurringtask
+        WHERE is_locked = FALSE
+        ORDER BY id
+        LIMIT num_tasks
+        FOR UPDATE SKIP LOCKED
```
matthewelwell, line 14: Do we still need this?

gagantrivedi: Yep, else we risk  two(or more) processor getting the same task

matthewelwell: I thought that was the point of the `is_locked` attribute though? I'm not sure why we need both solutions here.

api/core/migration_helpers.py
```diff
@@ -7,6 +7,14 @@
 
 
 class PostgresOnlyRunSQL(migrations.RunSQL):
+    def __init__(self, *args, sql_file_path: str = None, **kwargs):
+        if sql_file_path:
+            with open(sql_file_path) as f:
+                sql = f.read()
+                super().__init__(sql, *args, **kwargs)
```
matthewelwell: I think this might do something odd if the user passes `sql` as an argument as well (I _think_ it would put it in the reverse_sql?). We should ensure that only one of `sql` or `sql_file_path` is passed? Or maybe just create a classmethod? Something like: 

```python
@classmethod
def from_sql_file(cls, file_path: str, reverse_sql: str) -> "PostgresOnlyRunSQL":
    with open(file_path) as f:
        return cls(f.read(), reverse_sql=reverse_sql)
```

api/task_processor/migrations/0008_add_get_task_to_process_function.py
```diff
@@ -0,0 +1,31 @@
...
+    ]
+
+    operations = [
+        PostgresOnlyRunSQL.from_sql_file(
+            os.path.join(
+                os.path.dirname(__file__),
+                "sql",
+                "get_tasks_to_process.sql",
+            ),
+            reverse_sql="DROP FUNCTION IF EXISTS get_tasks_to_process()",
```
matthewelwell: Having attempted to run this, I think the correct syntax is actually: 

```suggestion
            reverse_sql="DROP FUNCTION IF EXISTS get_tasks_to_process;",
```

api/task_processor/migrations/0008_add_get_task_to_process_function.py
```diff
@@ -0,0 +1,31 @@
...
+        ),
+        PostgresOnlyRunSQL.from_sql_file(
+            os.path.join(
+                os.path.dirname(__file__),
+                "sql",
+                "get_recurring_tasks_to_process.sql",
+            ),
+            reverse_sql="DROP FUNCTION IF EXISTS get_recurringtasks_to_process()",
```
matthewelwell:
```suggestion
            reverse_sql="DROP FUNCTION IF EXISTS get_recurringtasks_to_process;",
```

gagantrivedi: Ah, yes, I forgot to update this

matthewelwell (approved):

PR #2130, opened 2023-04-18
feat: run-task-processor Docker entrypoint command

khvn26:
* Add ability to wait for migrations via waitdb command

* Add run-task-processor entrypoint command

## Changes

* Added ability to wait for migrations via `waitdb` command.
* Added the `run-task-processor` entrypoint command

Built a Docker image locally and tested the new entrypoint it with the flagsmith/self-hosted docker-compose file.

### DIFF
--- a/api/core/management/commands/waitfordb.py
+++ b/api/core/management/commands/waitfordb.py
@@ -1,43 +1,86 @@
 import logging
 import time
 from argparse import ArgumentParser
-from datetime import datetime
+from typing import Any
 
 from django.core.management import BaseCommand
 from django.db import OperationalError, connections
+from django.db.migrations.executor import MigrationExecutor
 
 logger = logging.getLogger(__name__)
 
 
 class Command(BaseCommand):
-    def add_arguments(self, parser: ArgumentParser):
+    def add_arguments(self, parser: ArgumentParser) -> None:
         parser.add_argument(
             "--waitfor",
             type=int,
+            dest="wait_for",
             help="Number of seconds to wait for db to become available.",
             default=5,
         )
+        parser.add_argument(
+            "--migrations",
+            action="store_true",
+            dest="should_wait_for_migrations",
+            help="Whether to wait until all migrations are applied.",
+            default=False,
+        )
+        parser.add_argument(
+            "--database",
+            type=str,
+            dest="database",
+            help=(
+                'The database to wait for ("default", "analytics").'
+                'Defaults to the "default" database.'
+            ),
+            default="default",
+        )
 
-    def handle(self, *args, **options):
-        wait_for = options["waitfor"]
-        start = datetime.now()
+    def handle(
+        self,
+        *args: Any,
+        wait_for: int,
+        should_wait_for_migrations: bool,
+        database: str,
+        **options: Any,
+    ) -> None:
+        start = time.monotonic()
         wait_between_checks = 0.25
 
         logger.info("Checking if database is ready for connections.")
 
         while True:
-            if (datetime.now() - start).total_seconds() > wait_for:
+            if time.monotonic() - start > wait_for:
                 msg = f"Failed to connect to DB within {wait_for} seconds."
                 logger.error(msg)
                 exit(msg)
 
-            conn = connections.create_connection("default")
+            conn = connections.create_connection(database)
             try:
                 with conn.temporary_connection() as cursor:
                     cursor.execute("SELECT 1")
                 logger.info("Successfully connected to the database.")
-                return
+                break
             except OperationalError:
                 logger.warning("Database not yet ready for connections.")
 
             time.sleep(wait_between_checks)
+
+        if should_wait_for_migrations:
+            logger.info("Checking for applied migrations.")
+
+            while True:
+                if time.monotonic() - start > wait_for:
+                    msg = f"Didn't detect applied migrations for {wait_for} seconds."
+                    logger.error(msg)
+                    exit(msg)
+
+                conn = connections[database]
+                executor = MigrationExecutor(conn)
+                if not executor.migration_plan(executor.loader.graph.leaf_nodes()):
+                    logger.info("No pending migrations detected, good to go.")
+                    return
+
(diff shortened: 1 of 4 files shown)


api/core/management/commands/waitfordb.py
```diff
@@ -36,8 +50,26 @@ def handle(self, *args, **options):
...
+        if should_wait_for_migrations:
+            logger.info("Checking for applied migrations.")
+
+            while True:
+                if time.monotonic() - start > wait_for:
+                    msg = f"Didn't detect applied migrations for {wait_for} seconds."
+                    logger.error(msg)
+                    exit(msg)
+
+                conn = connections["default"]
```
gagantrivedi: This will not work for analytics db migrations, right?

gagantrivedi (changes requested):

api/core/management/commands/waitfordb.py
khvn26: Yes, good catch. Do we want to cover the replicas as well?

matthewelwell: I don't think we need to cover the replicas, no. We can assume that the replication process will handle that for us.

api/scripts/run-docker.sh
```diff
@@ -59,6 +62,9 @@ if [ "$1" == "migrate" ]; then
 elif [ "$1" == "serve" ]; then
     if [ $# -eq 2 ]; then go_to_sleep "$2"; fi
     serve
+elif [ "$1" == "run-tasks" ]; then
```
matthewelwell: I'd say that run-processor or run-task-processor would be better here. run-tasks sounds like an atomic operation rather than a long running process.

khvn26: Agreed. Fixed.

api/core/management/commands/waitfordb.py
khvn26: Added a `database` option for the command like in Django's `migrate` command.

api/scripts/run-docker.sh
```diff
@@ -28,6 +28,9 @@ function serve() {
              ${STATSD_HOST:+--statsd-prefix $STATSD_PREFIX} \
              app.wsgi
 }
+function run_task_processor() {
+    python manage.py waitfordb --waitfor 30 --migrations && python manage.py runprocessor --sleepintervalms 500
```
matthewelwell: Do we need to also check the analytics db is ready here? If analytics are enabled we will need to wait for the analytics db I think?

api/scripts/run-docker.sh
```diff
@@ -28,6 +28,9 @@ function serve() {
              ${STATSD_HOST:+--statsd-prefix $STATSD_PREFIX} \
              app.wsgi
 }
+function run_task_processor() {
```
gagantrivedi, line 31: It feels a little odd that this sits in `run-docker.sh` and is not used by any docker file? Maybe a makefile is a better place for this?

dabeeeenster:
Yeah i think we should consider adding makefiles to the core project...

api/scripts/run-docker.sh
khvn26: I plan to update flagsmith/self-hosted to use this entrypoint. I actually encountered the need for this trying to run the Compose file.

api/scripts/run-docker.sh
khvn26: True. Added the wait for analytics database.

api/scripts/run-docker.sh
matthewelwell: We should also update the `docker-compose.yml` file in the root of this repository.

matthewelwell (approved):
LGTM but worth getting the ✅ from @gagantrivedi as well.

api/scripts/run-docker.sh
khvn26: Fair point, updated.

api/scripts/run-docker.sh
```diff
@@ -59,6 +66,9 @@ if [ "$1" == "migrate" ]; then
 elif [ "$1" == "serve" ]; then
     if [ $# -eq 2 ]; then go_to_sleep "$2"; fi
     serve
+elif [ "$1" == "run-task-processor" ]; then
+    if [ $# -eq 2 ]; then go_to_sleep "$2"; fi
```
gagantrivedi: nit: should we get rid of  `go_to_sleep` here? @matthewelwell  what do you think?

matthewelwell: Yeah, you're right, we definitely _should_, but I don't think it's necessary here. It doesn't look like that functionality is used anywhere but it's technically part of the public API of our docker image and I don't want to have to major version our docker image(s).

gagantrivedi: > Yeah, you're right, we definitely _should_, but I don't think it's necessary here. It doesn't look like that functionality is used anywhere but it's technically part of the public API of our docker image and I don't want to have to major version our docker image(s).

But we can choose to not add that here, right? Since that's not part of the public API yet?

matthewelwell: > > Yeah, you're right, we definitely _should_, but I don't think it's necessary here. It doesn't look like that functionality is used anywhere but it's technically part of the public API of our docker image and I don't want to have to major version our docker image(s).
> 
> But we can choose to not add that here, right? Since that's not part of the public API yet?

Sorry, that's true, yes!

gagantrivedi: > > > Yeah, you're right, we definitely _should_, but I don't think it's necessary here. It doesn't look like that functionality is used anywhere but it's technically part of the public API of our docker image and I don't want to have to major version our docker image(s).
> > 
> > 
> > But we can choose to not add that here, right? Since that's not part of the public API yet?
> 
> Sorry, that's true, yes!

And maybe add a comment there somewhere that it is deprecated and Will be removed

khvn26: Done.

gagantrivedi (approved):

PR #2131, opened 2023-04-18
Update requirements needed for 3.11.3

matthewelwell:
## Changes

I had to recreate my local development environment and in the process couldn't install some of the requirements. I think because some older package versions do not have wheels to support 3.11, but I'm not entirely sure why the builds are therefore succeeding. Either way, it seems like a good thing to upgrade these older packages anyway. 

Unit tests.


khvn26:
`google-auth` and `slack-sdk` bumps are huge. We've got to make sure that nothing's broken.

PR #2132, opened 2023-04-18
Update requirements needed for 3.11

matthewelwell:
## Changes

I had to recreate my local development environment and in the process couldn't install some of the requirements. I think because some older package versions do not have wheels to support 3.11, but I'm not entirely sure why the builds are therefore succeeding. Either way, it seems like a good thing to upgrade these older packages anyway. 

Unit tests.


khvn26:
google-auth and slack-sdk bumps are huge. We've got to make sure that nothing's broken. I don't suppose e2e covers the relevant scenarios?

matthewelwell:
Yeah, good point. I don't believe the E2E tests cover either of these so we'll need to test these manually. I will try and do this at some point. This PR is not urgent anyway, I'd like to look into why I had to upgrade them in my local environment anyway.

gagantrivedi:
> Yeah, good point. I don't believe the E2E tests cover either of these so we'll need to test these manually. I will try and do this at some point. This PR is not urgent anyway, I'd like to look into why I had to upgrade them in my local environment anyway.

I think because wheel are available for the runner(linux) and not mac OS?

khvn26:
@gagantrivedi I have Python 3.11.3 locally and didn't have to bump any of the requirements.

```sh
➜  ~ sw_vers
ProductName:		macOS
ProductVersion:		13.3.1
BuildVersion:		22E261
```

dabeeeenster:
I've never had any issues with 3.11 either?

matthewelwell:
Yeah, I'm not sure - I'll try creating a new clean environment again to see if I can replicate it. It could possibly have been related to the fact I was using terrible train WiFi at the time...

matthewelwell:
Yep, works fine trying again on stable wifi. I'll close this PR.

PR #2135, opened 2023-04-18
feat: remove deprecated DISABLE_INFLUXDB_FEATURES override

khvn26:

## Changes

Confusing `DISABLE_INFLUXDB_FEATURES` environment variable is removed. All relevant code moved to `DISABLE_ANALYTICS_FEATURES`. 

Via CI:
- E2E
- Uffizzi deployment


api/app/settings/common.py
```diff
@@ -731,7 +731,6 @@
 PREVENT_SIGNUP = env.bool("PREVENT_SIGNUP", default=False)
 PREVENT_EMAIL_PASSWORD = env.bool("PREVENT_EMAIL_PASSWORD", default=False)
 DISABLE_ANALYTICS_FEATURES = env.bool("DISABLE_ANALYTICS_FEATURES", default=False)
-DISABLE_INFLUXDB_FEATURES = env.bool("DISABLE_INFLUXDB_FEATURES", default=True)
```
matthewelwell, line 738: Hmm... removing environment variables like this could be construed as a breaking change. We may have self hosted customers using this environment variable.

We probably have a couple of options here: 

 1. Remove it anyway and make a big point of it in the release notes
 2. Just update the code / setting but still use the same environment variable

I think (2) is probably the way forward but happy to discuss.

docker-compose.uffizzi.yml
```diff
@@ -38,14 +38,14 @@ services:
...
+      DISABLE_ANALYTICS_FEATURES: 'true' # set to false to enable InfluxDB
       PREVENT_SIGNUP: 'false'
       ENABLE_ADMIN_ACCESS_USER_PASS: 'true'
       DJANGO_SECRET_KEY: '**SoMeTeMpOrArYsEcReT!!'
 
       ENV:
-        dev # set to "prod" in production.
-
+        dev
+        # set to "prod" in production.
```
matthewelwell: Is there a reason this was pushed down onto the next line? Some kind of conflicting linting?

frontend/api/index.js
```diff
@@ -136,7 +130,7 @@ if (process.env.FLAGSMITH_PROXY_API_URL) {
 if (isDev) {
   // Serve files from src directory and use webpack-dev-server
   // eslint-disable-next-line
-    console.log('Enabled Webpack Hot Reloading');
+  console.log('Enabled Webpack Hot Reloading')
```
matthewelwell: The indentation changes here are good, but I'm not sure we should be removing the semi colon?

docker-compose.uffizzi.yml
khvn26: Red Hat's YAML extension does this. Reverted.

api/app/settings/common.py
khvn26: Agree on the option 2, thanks! Added reading from `DISABLE_INFLUXDB_FEATURES`.

frontend/api/index.js
khvn26: Looks like pre-commit's prettier hook ignores the `eslint-disable-next-line` annotation. Reverted.

.github/actions/e2e-tests/action.yml
```diff
@@ -40,7 +40,6 @@ runs:
         SLACK_TOKEN: ${{ inputs.slack_token }}
         STATIC_ASSET_CDN_URL: /
         ENV: ${{ inputs.environment }}
-        ENABLE_INFLUXDB_FEATURES: ${{ inputs.enable_influxdb_features }}
```
matthewelwell, line 43: So, I guess we're removing this from here as the environment variable is no longer used? We rely on the DISABLE_ANALYTICS_FEATURES from the API, is that right? 

If so, we should remove the input from the list of `inputs` above and remove the value wherever we're calling this action.

matthewelwell (commented):
@kyle-ssg could you also review this from a FE code perspective.

.github/actions/e2e-tests/action.yml
khvn26: Great catch, fixed. Also fixed the docs.

matthewelwell: I don't quite follow what the benefit of the change is here? You've updated the input but we're not using the input anywhere in this action as far as I can see? Should `disable_analytics_features` be an input to the `run-local-api` action?

api/Makefile
```diff
@@ -0,0 +1,26 @@
+.EXPORT_ALL_VARIABLES:
```
matthewelwell: I think this has snuck in accidentally?

.github/actions/e2e-tests/action.yml
khvn26: The following workflows do use this input:

- `.github/workflows/platform-pull-request.yml`
- `.github/workflows/platform-test-merge-to-main.yml`

Or am I missing something here?

api/Makefile
khvn26: Yes :( Got rid of it.

.github/actions/e2e-tests/action.yml
khvn26: You're right, it's meaningless now. Removed both. the input and the references.

khvn26: Agree, let's control this from the backend. Moved `disable_analytics_features` input to `run-local-api`.

matthewelwell (approved):
This looks good to me now.

PR #2138, opened 2023-04-19
Feature: Allow users to change their email address

### LINKED ISSUE #887: Allow users to change their email address
Users should be able to change their email address. Suggested flow:

- Log in
- Enter new email address and confirm with password
- Email sent to previous email address to notify that the email on their Flagsmith account has been updated and they should contact Flagsmith support if there are any issues

novakzaballa:

## Changes

- Add ChangeEmailAddress modal
- Add an API client code for updating the user email address
- Send a notification when the email address has been changed

- Click 'Account' at the top of the application.
- Click 'Change Email Address' button.
- Set a new email and your current password.
- Click 'Save Changes' button.


api/users/models.py
```diff
@@ -121,6 +122,28 @@ def __str__(self):
...
 
+    @hook(AFTER_UPDATE, when="email", has_changed=True)
+    def send_warning_email(self):
+        message = inspect.cleandoc(
+            f"""
+            Hi there,
+
+            Your new email address is {self}.
+
+            If you haven't request this change please contact us at support team.
```
khvn26: I think we should link to `support@flagsmith.com` here.

Won't be viable for on-prem installs but we'll figure out how to store the local support contacts later.

novakzaballa: Good catch Kim, I missed the email address. Corrected.

api/users/models.py
```diff
@@ -121,6 +122,28 @@ def __str__(self):
     def subscribe_to_mailing_list(self):
         mailer_lite.subscribe(self)
 
+    @hook(AFTER_UPDATE, when="email", has_changed=True)
+    def send_warning_email(self):
+        message = inspect.cleandoc(
+            f"""
```
matthewelwell: Can we move this to use a template instead of the string here? Also, let's use the user's `first_name` attribute instead of 'there' (since we should have it here).

And let's update the content slightly. Instead of saying 'Your new email address is {self}.' let's say: 

```
Hi {first_name},

A change was just made to update the email associated with your Flagsmith account. If you did not request this change, please contact support@flagsmith.com as soon as possible. 

The Flagsmith Team
```

api/users/models.py
```diff
@@ -121,6 +122,28 @@ def __str__(self):
...
+
+            Your new email address is {self}.
+
+            If you haven't request this change please contact us at support@flagsmith.com
+
+            The Flagsmith Team
+            """
+        )
+
+        send_mail(
```
matthewelwell: We should delegate the sending of this email to the asynchronous processor. You can see an example of this being used [here](https://github.com/Flagsmith/flagsmith/blob/main/api/users/tasks.py#L9) but we can also go through this together next week.

frontend/common/services/useUserEmail.ts
```diff
@@ -0,0 +1,52 @@
...
+  data: Req['updateUserEmail'],
+  options?: Parameters<
+    typeof userEmailService.endpoints.updateUserEmail.initiate
+  >[1],
+) {
+  store.dispatch(
+    userEmailService.endpoints.updateUserEmail.initiate(data, options),
+  )
+  return Promise.all(
+    store.dispatch(userEmailService.util.getRunningQueriesThunk()),
```
matthewelwell, line 38: ... this might be a silly question but what's a 'Thunk' 😄 ? `getRunningQueriesThunk`

frontend/web/components/modals/ChangeEmailAddress.tsx
```diff
@@ -0,0 +1,101 @@
...
+              setEmail(Utils.safeParseEventValue(event))
+            }}
+            isValid={email && email.length}
+            type='email'
+            name='email'
+            id='email'
+            placeholder='E.g. email123@email.com'
+          />
+          <InputGroup
+            title='Re-enter Password'
```
matthewelwell: We should title this 'Confirm Password'

matthewelwell (commented):
Added a few comments on specific lines of code. I'm also wondering whether we should include this in the E2E tests? @kyle-ssg what do you think?

frontend/common/services/useUserEmail.ts
kyle-ssg: FE stuff 😄 this is CLI output, it's necessary to await asynchronous tasks on the FE

kyle-ssg:
> Added a few comments on specific lines of code. I'm also wondering whether we should include this in the E2E tests? @kyle-ssg what do you think?

Yep discussed with @novakzaballa actually, agreed on an approach and will do this as part of this PR. It'll be added to the end of the invite test so

- Invite user
- Register with user
- Change email
- Login with new email
- Delete account (upcoming pr)

frontend/common/services/useUserEmail.ts
kyle-ssg: Redux state works in a single direction in an immutable way, you do an action, it updates state which updates the ui

"Thunks" are basically side effects to actions, with thunks it's like
- Trigger action (e.g. update email)
- Thunk takes the action, does api stuff and generates an action to say it's loaded with the data
- Reducers take loaded action and update state

matthewelwell: Got it. FE tech and their stupid names! 🙄

api/users/models.py
```diff
@@ -121,6 +121,18 @@ def __str__(self):
     def subscribe_to_mailing_list(self):
         mailer_lite.subscribe(self)
 
+    @hook(AFTER_UPDATE, when="email", has_changed=True)
+    def send_warning_email(self):
+        from users.tasks import send_email_changed_notification_email
```
matthewelwell: We should avoid using local imports like this as much as possible but I can't see an easy way to avoid it here!

novakzaballa: Yes, I will try to avoid it as much as possible.

novakzaballa:
@matthewelwell the call to logout user has been added after the user changes its email address.

api/users/models.py
```diff
@@ -121,6 +121,18 @@ def __str__(self):
     def subscribe_to_mailing_list(self):
         mailer_lite.subscribe(self)
 
+    @hook(AFTER_UPDATE, when="email", has_changed=True)
+    def send_warning_email(self):
```
matthewelwell: This method (and the task function) should have unit tests. Let's go through how / where to implement these later.

novakzaballa:
> > Added a few comments on specific lines of code. I'm also wondering whether we should include this in the E2E tests? @kyle-ssg what do you think?
> 
> Yep discussed with @novakzaballa actually, agreed on an approach, and will do this as part of this PR. It'll be added to the end of the invite test so
> 
> * Invite user
> * Register with user
> * Change email
> * Login with new email
> * Delete account (upcoming pr)

@kyle-ssg let's do the E2E testing of this in a new ticket please since it will include the `delete account` feature which is not yet deployed. I created a draft task in our Kanban for this: https://github.com/orgs/Flagsmith/projects/2/views/7?pane=issue&itemId=27359177

api/users/models.py
novakzaballa: @matthewelwell tests have been added. Please Re-review

api/users/tests/test_models.py
```diff
@@ -318,3 +323,55 @@ def test_user_create_does_not_call_pipedrive_tracking_if_ignored_domain(
...
+    data = {"new_email": "test_user1@test.com", "current_password": "password"}
+
+    url = "/api/v1/auth/users/set_email/"
+
+    # When
+    response = client.post(url, data=json.dumps(data), content_type="application/json")
+
+    # Then
+    assert response.status_code == status.HTTP_204_NO_CONTENT
+    assert user.email == "test_user1@test.com"
```
matthewelwell: we should extract this as a variable (or perhaps just `assert user.email == data["new_email"]`)

api/users/tests/test_models.py
```diff
@@ -318,3 +323,55 @@ def test_user_create_does_not_call_pipedrive_tracking_if_ignored_domain(
 
 def test_user_email_domain_property():
     assert FFAdminUser(email="test@example.com").email_domain == "example.com"
+
+
+@pytest.mark.django_db
+def test_change_email_address_api(mocker):
```
matthewelwell: Since this test is no longer a model test only, it should be located elsewhere. Probably it should live in `api/tests/unit/users/test_unit_users_views.py`.

api/users/tests/test_models.py
```diff
@@ -318,3 +323,55 @@ def test_user_create_does_not_call_pipedrive_tracking_if_ignored_domain(
...
+    # Then
+    assert response.status_code == status.HTTP_204_NO_CONTENT
+    assert user.email == "test_user1@test.com"
+
+    args, kwargs = mocked_send_mail.call_args
+
+    assert len(args) == 0
+    assert len(kwargs) == 5
+    assert kwargs["subject"] == "Your Flagsmith email address has been changed"
+    assert kwargs["from_email"] == "noreply@flagsmith.com"
```
matthewelwell: This might give unexpected results in the case that the tests are run with the DEFAULT_FROM_EMAIL environment variable set, right?

api/users/tests/test_models.py
```diff
@@ -318,3 +323,55 @@ def test_user_create_does_not_call_pipedrive_tracking_if_ignored_domain(
...
+    assert kwargs["recipient_list"] == ["test_user@test.com"]
```
matthewelwell: again, can we extract the email address to a variable?

api/users/tests/test_models.py
matthewelwell: It's also a little confusing that this test relies on the TASK_RUN_METHOD environment variable not being set to TASK_PROCESSOR. In that case, I believe this test would fail, right? 

Should we, in that case, use this end to end test to verify that the task itself is called with the correct arguments and then unit test the task function independently?

api/users/tests/test_models.py
```diff
@@ -318,3 +323,55 @@ def test_user_create_does_not_call_pipedrive_tracking_if_ignored_domain(
...
+    assert kwargs["fail_silently"] is True
+
+
+def test_send_email_changed_notification():
```
matthewelwell: This should live in `api/tests/users/test_unit_users_tasks.py`.

matthewelwell: This should live in `api/tests/users/test_unit_users_tasks.py`.

frontend/web/components/modals/ChangeEmailAddress.tsx
```diff
@@ -0,0 +1,101 @@
...
+            value={password}
+            onChange={(event: React.ChangeEvent<HTMLInputElement>) => {
+              setPassword(Utils.safeParseEventValue(event))
+            }}
+            type='password'
+            name='password'
+          />
+          {isError && (
+            <div className='alert alert-danger'>
+              Error Updating Email Address, please ensure you have entered your
```
matthewelwell: The only way I can think this might fail is if they try to set it to an email address that is not unique. Perhaps we should indicate that here? Something like: 

`An error occurred attempting to update your email address. Please verify that the password you entered is correct and that the email address you have provided is not already being used.`

We could perhaps even inspect the API responses to see if the error messages returned indicate what occurred? Although actually, since we allow users to sign up for free, we wouldn't want to confirm that it was an issue with the email address being unique as that would allow malicious actors to enumerate email addresses. An alternative would be to guard this endpoint with rate limiting using the built in DRF rate limiting.

frontend/web/components/pages/AccountSettingsPage.js
```diff
@@ -140,24 +141,47 @@ class TheComponent extends Component {
...
-                          value={email}
-                          onChange={(e) =>
-                            this.setState({
-                              first_name: Utils.safeParseEventValue(e),
-                            })
-                          }
-                          type='text'
-                          name='Email Address'
-                        />
+                        <div className='md-8'>
```
matthewelwell, line 144: I would like to see this wrapped in a feature flag again.

matthewelwell (commented):
I've also tested this manually in staging and all seems to work nicely but I didn't receive the email to the old email address. This could be an issue with the email sending in our staging environment, however, so I'll check this out once I'm on more stable wifi!

matthewelwell (commented):
I've also tested this manually in staging and all seems to work nicely but I didn't receive the email to the old email address. This could be an issue with the email sending in our staging environment, however, so I'll check this out once I'm on more stable wifi!

api/users/tests/test_models.py
novakzaballa: Corrected.

api/users/tests/test_models.py
novakzaballa: It has been designed to work exactly in that way, am I missing something?

api/users/tests/test_models.py
novakzaballa: Corrected

api/users/tests/test_models.py
novakzaballa: Done

api/users/tests/test_models.py
novakzaballa: Moved

novakzaballa: Duplicated

frontend/web/components/modals/ChangeEmailAddress.tsx
novakzaballa: Done. Error captured and different messages are displayed in each case.

frontend/web/components/pages/AccountSettingsPage.js
novakzaballa: We had a feature flag for the new "Change Email" button. I added a new feature flag for showing the email information, is that what you referred to?

api/users/tests/test_models.py
matthewelwell: Yes, the issue is that, if you run this test like this, it will fail: 

```bash
TASK_RUN_METHOD=TASK_PROCESSOR pytest -k test_change_email_address_api 
```

The best option here would be to mock the `send_email_changed_notification_email` task function, not the `send_mail` function. So we can change the mock to: 

```python
mocked_task = mocker.patch("users.models.send_email_changed_notification_email")

...

mocked_task.delay.assert_called_once_with(...)
```

Then, since we have a unit test on the `send_email_changed_notification_email`, we've made sure that we're covering all of the code path. 

The general rule here is that we should never end to end test anything that expects a function decorated with `@register_task_handler` to be called as it could mean we get different behaviour in different test environments. We want to try to make our test suite replicable in all environments.

frontend/web/components/pages/AccountSettingsPage.js
matthewelwell: Sorry, that's my fault! I didn't see the flag below. We only need a single flag here. You can remove the `show_email_address` flag.

matthewelwell (approved):

PR #2139, opened 2023-04-20
Update permissions for group admins

matthewelwell:
## Changes

Updates permission checks to correctly account for Group admins. Note that as part of this, a request to interact with a particular user permission group from a user that does not have permission to do so will now result in a 404, not a 403 since it no longer appears in the queryset for that user. This is _technically_ a breaking change but it feels like the right behaviour so I would say is a necessary fix.

Added / updated tests.

### DIFF
--- a/api/users/tests/test_views.py
+++ b/api/users/tests/test_views.py
@@ -488,3 +488,25 @@ def test_retrieve_user_permission_group_includes_group_admin(
         next(filter(lambda u: u["id"] == group_admin_user.id, users))["group_admin"]
         is True
     )
+
+
+def test_group_admin_can_retrieve_group(organisation, django_user_model, api_client):
+    # Given
+    user = django_user_model.objects.create(email="test@example.com")
+    user.add_organisation(organisation)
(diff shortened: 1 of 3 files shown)


khvn26 (approved):
LGTM 👍

api/users/tests/test_views.py
```diff
@@ -486,3 +488,25 @@ def test_retrieve_user_permission_group_includes_group_admin(
         next(filter(lambda u: u["id"] == group_admin_user.id, users))["group_admin"]
         is True
     )
+
+
+def test_group_admin_can_retrieve_group(organisation, django_user_model, api_client):
```
khvn26: Nit: we don't have a stance re: typing yet, but I try to type-annotate every code line I touch hoping we'll be able to leverage mypy sooner than later.

novakzaballa (approved):
LGTM

api/users/tests/test_views.py
matthewelwell: Yeah, good point. I haven't used it for fixtures before but I guess there is no reason not to... will update.

PR #2141, opened 2023-04-20
Add user name to confirmation dialog when removing users

### LINKED ISSUE #2137: Add user name to confirmation dialog when removing users
When removing a user, you get presented a confirmation dialog. However, that dialog does not state any information about the user you are about to remove. Hence if you clicked the wrong trash bin button (the buttons are to the far right and the names are to the far left) you may end up with removing the wrong user.

Adding information about the user in the confirmation dialog would help preventing such mistakes..

novakzaballa:

## Changes

- Add user name to confirmation dialog when removing users

- Click 'Manage' at the top of the application.
- Click 'Members' tab.
- Click remove button for any user.
- Check if the full name is displayed in the dialog.


frontend/web/components/pages/OrganisationSettingsPage.js
```diff
@@ -984,7 +984,10 @@ const OrganisationSettingsPage = class extends Component {
                                                         id='delete-invite'
                                                         type='button'
                                                         onClick={() =>
-                                                          this.deleteUser(id)
+                                                          this.deleteUser(
+                                                            id,
+                                                            `${first_name} ${last_name}`,
```
matthewelwell: There are certain scenarios where a user might not have a first / last name. We should try to cater for this. @kyle-ssg are there other places in the FE code that we're doing this already? One option would be to use their email address here instead. 

Another thing (out of the original scope of this PR) is that the wording in this modal is wrong. This is not 'deleting' a user - it's just removing them from the organisation. We should update the wording to reflect this.

novakzaballa: What about just using this message template?
`Are you sure you want to remove the user ${fullname} with email address ${email}`

matthewelwell: That would still end up looking a bit odd if `fullname` is missing or undefined, right?

matthewelwell: That would still end up looking a bit odd if `fullname` is missing or undefined, right?

novakzaballa: I don't think so since (handling the scapes) it would read:
`Are you sure you want to remove the user with email address ${email}`
Which looks proper to me, but ... you know it's not my first language :)

kyle-ssg: I think this is a problem app-wide, I'd suggest we create a format function 

```
const getUserDisplayName = (user?:ProfileSummary)=> {}
```

As for this PR let's use the above for that but create a follow-up ticket investigating where else this can be used.

matthewelwell: Sure, sounds good!

frontend/common/utils/format.js
```diff
@@ -90,6 +90,10 @@ const Format = {
       : Format.camelCase(sn)
   },
 
+  userDisplayName(person) {
```
matthewelwell, line 93: What's the reasoning behind creating a new function here vs just using Format.fullName directly in the OrganisationSettingsPage.js file?

novakzaballa: Corrected.

matthewelwell (approved):

PR #2142, opened 2023-04-21
fix(settings): use smtp.EmailBacked as default if `SENDGRID_API_KEY` is not defined

gagantrivedi:

## Changes

The problem with using `SendGridBackend` is that
it does not fail silently if `SENDGRID_API_KEY` is not set

We did consider using `dummy.emailbackend` as the default, but that does not make much sense for `DEBUG=False`(i.e: non development) kind of environments since we still want to send_mail with `fail_silently=False` to raise error

Tested manually by causing an exception with `DEBUG=False`

### DIFF
--- a/api/app/settings/common.py
+++ b/api/app/settings/common.py
@@ -11,7 +11,6 @@ https://docs.djangoproject.com/en/1.9/ref/settings/
 """
 import importlib
 import json
-import logging
 import os
 import sys
 import warnings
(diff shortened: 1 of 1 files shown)


matthewelwell (commented):
Thinking about it, this is technically a breaking change. For example, we need to update our task definitions to explicitly use the sendgrid backend. Other users will need to do the same. Given how painful this can be, I am quite keen to do it. So, we can either: 

 1. Do it and make sure it's clear in the release notes
 2. Add it to the V2 API issue we have that's collecting all the breaking changes we need to make.

gagantrivedi:
> Thinking about it, this is technically a breaking change. For example, we need to update our task definitions to explicitly use the sendgrid backend. Other users will need to do the same. Given how painful this can be, I am quite keen to do it. So, we can either:
> 
>     1. Do it and make sure it's clear in the release notes
> 
>     2. Add it to the V2 API issue we have that's collecting all the breaking changes we need to make.

Another option would be to fall back to send grid as default if `SENDGRID_API_KEY` is set? That way, we don't need to update our task processor definition? (but it would be good to still add it to the v2 issue to remove this behaviour)

matthewelwell:
Ah yes, good idea!

matthewelwell (approved):

PR #2157, opened 2023-04-25
Feature: Delete user account

### LINKED ISSUE #2099: Add the ability to completely delete your own user account
Users should have the ability to completely delete their user account and all of their data. 

This option should live in the settings accessed via the 'Account' option in the top navigation. Below 'Show JSON references'. See screenshot below. 

Image

novakzaballa:

## Changes

- Add ConfirmDeleteAccount modal
- Add an API client code for deleting the user account
- Organisations are delete when the last user is deleted
- Merge improve_update_account branch

- Click 'Account' at the top of the application.
- Click delete button in the Delete Account section.
- Set your current password in the ConfirmDeleteAccount.
- Click 'Delete Account' button.
- Try login with the deleted account.


api/users/models.py
```diff
@@ -121,6 +121,18 @@ def __str__(self):
     def subscribe_to_mailing_list(self):
         mailer_lite.subscribe(self)
 
+    @hook(BEFORE_DELETE)
+    def delete_organisation_with_a_last_user(self):
```
khvn26: I'd love to see this covered with unit tests.

api/users/models.py
```diff
@@ -121,6 +121,18 @@ def __str__(self):
...
+        user_organisations = UserOrganisation.objects.filter(
+            user_id=self.initial_value("id")
+        )
+        for user_organisation in user_organisations:
+            users_count = UserOrganisation.objects.filter(
```
khvn26: This can be optimized, e.g. something like:

```python
organisation_ids_to_delete = self.organisations.\
    .annotate(num_users=Count("users"))\
    .filter(num_users=1)\
    .values_list("id", flat=True)
```

khvn26 (changes requested):

frontend/web/components/modals/ConfirmDeleteAccount.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  const ModalBody: FC<ConfirmDeleteAccountType> = ({
+    lastUserOrganisations,
+    hasOrganisations,
+  }) => {
+    if (lastUserOrganisations.length >= 1) {
+      return (
+        <p>
+          You are the last user from:
+          {lastUserOrganisations.map((o) => {
```
kyle-ssg: Can be ```<strong>${lastUserOrganisations.join(", ")}</strong>```

api/users/models.py
novakzaballa: The proposed solution can't be implemented as is since organizations is a many_to_many field. However the code has been simplified in a similar way.

api/users/models.py
novakzaballa: Unit tests added.

api/users/models.py
khvn26: Awesome!

api/users/models.py
khvn26: Great, thank you!

api/users/models.py
```diff
@@ -121,6 +121,13 @@ def __str__(self):
     def subscribe_to_mailing_list(self):
         mailer_lite.subscribe(self)
 
+    @hook(BEFORE_DELETE)
+    def delete_organisation_with_a_last_user(self):
```
gagantrivedi: Extreme nitpick:  what do you think about renaming the function to `..with_the_last...` instead of `..with_a_last.`?

novakzaballa: Thanks @gagantrivedi! what about `delete_single_user_organisations` or `delete_only_user_organisations`?

novakzaballa: Or it can be also `delete_last_user_organisations` to be aligned with the frontend.

khvn26 (dismissed):

api/users/tests/test_models.py
```diff
@@ -289,6 +289,51 @@ def test_user_remove_organisation_removes_user_from_the_user_permission_group(
...
+    assert Organisation.objects.filter(name="org2").count() == 1
+
+    # organization org3 is deleted since its only user is user1
+    user1.delete_orphan_organisations()
+    assert Organisation.objects.filter(name="org3").count() == 0
+    # org1 and org2 remain
+    assert Organisation.objects.filter(name="org1").count() == 1
+    assert Organisation.objects.filter(name="org2").count() == 1
+
+    # delete user 1 and user 2
```
matthewelwell: This feels like it's not testing the method here in isolation. If we change the lifecycle hook on the method, this test might fail. I feel like we should either commit to a more integration style test and _only_ use `user.delete()` to test the functionality, or we should _only_ use `user.delete_orphan_organisation()`. (In which case, we can use `user.organisations.remove(org)` I believe to test the logic as needed. 

I don't think I have a preference as to which one we want to do here.

frontend/common/types/requests.ts
```diff
@@ -65,6 +65,8 @@ export type Req = {
   }
   createTag: { projectId: string; tag: Omit<Tag, 'id'> }
   getSegment: { projectId: string; id: string }
+  updateAccount: Account
+  deleteAccount: Account
```
matthewelwell: I'm not entirely sure what this mapping is representing but it feels slightly odd to me that `deleteAccount` is mapped to an `Account` object. Is this mapping what the input should be to the `deleteAccount` method or perhaps the output?

frontend/web/components/modals/ConfirmDeleteAccount.tsx
```diff
@@ -0,0 +1,101 @@
...
+        </p>
+        {lastUserOrganisations.length >= 1 && (
+          <p>
+            You are the last user of
+            <strong>
+              {` ${lastUserOrganisations
+                .map((o: Organisation) => o.name)
+                .join(',')}.`}
+            </strong>{' '}
+            If you continue those organisations will be also deleted.
```
matthewelwell:
```suggestion
            If you continue, those organisations will be also deleted.
```

frontend/web/components/pages/AccountSettingsPage.js
```diff
@@ -221,6 +243,21 @@ class TheComponent extends Component {
                           checked={flagsmith.getTrait('json_inspect')}
                         />
                       </Row>
+                      <Row className='mt-4' space>
```
matthewelwell: I'd love to see this wrapped in a feature flag!

api/users/models.py
```diff
@@ -121,6 +121,12 @@ def __str__(self):
     def subscribe_to_mailing_list(self):
         mailer_lite.subscribe(self)
 
+    @hook(BEFORE_DELETE)
+    def delete_orphan_organisations(self):
```
matthewelwell, line 125: Having thought about this, I'm not sure that this is definitely the functionality that we want. There are circumstances in which users can be deleted where they will not be notified that they are also deleting organisations. Let's discuss.

api/custom_auth/views.py
```diff
@@ -51,3 +54,22 @@ def delete_token(request):
         sender=request.user.__class__, request=request, user=request.user
     )
     return Response(status=status.HTTP_204_NO_CONTENT)
+
+
+class CustomUserDeleteView(UserViewSet):
+    def destroy(self, request, *args, **kwargs):
+        if "delete_orphan_organizations" in request.query_params:
```
matthewelwell: We can use the walrus operator here to simplify things. 

https://docs.python.org/3/whatsnew/3.8.html#assignment-expressions

matthewelwell: We can actually also offload a lot of this logic to a query serializer. There are a few examples of this throughout the code. You basically just need to define a serializer like: 

```python
class UserDeleteQuerySerializer(serializers.Serializer):
    delete_orphan_organisations = serializers.BooleanField(default=False)
```

then in the view code, you can just do: 

```python
serializer = UserDeleteQuerySerializer(data=request.query_params)
if serializer.data["delete_orphan_organisations"]:
     # handle the organisation deletes
super().destroy(...)
```

api/custom_auth/views.py
```diff
@@ -51,3 +54,22 @@ def delete_token(request):
...
+            delete_orphan_organizations = request.query_params[
+                "delete_orphan_organizations"
+            ]
+            if delete_orphan_organizations.lower() == "true":
+                FFAdminUser.delete_orphan_organisations(request.user)
```
matthewelwell: For some reason, we're calling this as if it's a static method but it's not... this needs revisiting.

api/custom_auth/views.py
novakzaballa: Thank you @matthewelwell, I'll try using the serializer to take care of the param validation. I think it should inherit from UserDeleteSerializer since that one already inherits from CurrentPasswordSerializer.

api/users/models.py
novakzaballa: The corrected version implements the delete orphan organisations in a view.

api/users/tests/test_models.py
novakzaballa: Right. I  commit to a more integration-style test calling the delete endpoint in the test.

frontend/web/components/modals/ConfirmDeleteAccount.tsx
novakzaballa: Corrected

frontend/common/types/requests.ts
novakzaballa: In this case, it is the input to the endpoint, it includes the updateAccount as well because I merged the branch that @kyle-ssg prepared previously, since we will be using this new RTK approach to call the endpoints we took the opportunity to move the existing update account logic to this pattern. The `deleteAccount` has been corrected to reflect the current request shape.

frontend/web/components/pages/AccountSettingsPage.js
novakzaballa: The feature flag `delete_account` has been created and used here to display the Delete Account feature. It is set to false for production, and true for staging.

api/custom_auth/views.py
novakzaballa: Corrected

api/custom_auth/views.py
novakzaballa: Custom view optimized. I made it part of an existing custom view called ThrottledUserViewSet and renamed it to `FFAdminUserViewSet` to reflect our custom view and contain future customizations.

api/custom_auth/urls.py
```diff
@@ -4,16 +4,17 @@
...
+    FFAdminUserViewSet,
     delete_token,
 )
 
 app_name = "custom_auth"
 
-throttled_user_router = DefaultRouter()
-throttled_user_router.register(r"users", ThrottledUserViewSet)
+ffadmin_user_router = DefaultRouter()
+ffadmin_user_router.register(r"users", FFAdminUserViewSet)
```
matthewelwell, line 14: why has this changed from ThrottledUserViewSet to FFAdminUserViewSet?

api/custom_auth/views.py
```diff
@@ -30,7 +32,21 @@ class CustomAuthTokenLoginWithMFACode(AuthTokenLoginWithMFACode):
     throttle_scope = "mfa_code"
 
 
-class ThrottledUserViewSet(UserViewSet):
+@api_view(["DELETE"])
+@permission_classes([IsAuthenticated])
+def delete_token(request):
```
matthewelwell, line 40: I'm not too sure why we need this endpoint here?

api/custom_auth/views.py
```diff
@@ -30,7 +32,21 @@ class CustomAuthTokenLoginWithMFACode(AuthTokenLoginWithMFACode):
...
+        sender=request.user.__class__, request=request, user=request.user
+    )
+    return Response(status=status.HTTP_204_NO_CONTENT)
+
+
+class UserDeleteQuerySerializer(UserDeleteSerializer):
+    delete_orphan_organisations = serializers.BooleanField(default=False)
+
+
+class FFAdminUserViewSet(UserViewSet):
```
matthewelwell, line 52: I think this could all be simplified to just rewrite the `perform_destroy` method. Something like the following: 

```python
@method_decorator(
    decorator=swagger_auto_schema(query_serializer=DeleteCurrentUserQuerySerializer()),
    name="destroy",
)
class ThrottledUserViewSet(UserViewSet):
    throttle_scope = "signup"

    def get_throttles(self):
        """
        Used for throttling create(signup) action
        """
        throttles = []
        if self.action == "create":
            throttles = [ScopedRateThrottle()]
        return throttles

    def perform_destroy(self, instance):
        if self.request.query_params:
            serializer = DeleteCurrentUserQuerySerializer(
                data=self.request.query_params
            )

            if serializer.validated_data.get("delete_orphan_organisations"):
                self.request.user.delete_orpan_organisations()

        super().perform_destroy(instance)
```

matthewelwell, line 49: I think this could all be simplified to just rewrite the `perform_destroy` method. Something like the following: 

```python
@method_decorator(
    decorator=swagger_auto_schema(query_serializer=DeleteCurrentUserQuerySerializer()),
    name="destroy",
)
class ThrottledUserViewSet(UserViewSet):
    throttle_scope = "signup"

    def get_throttles(self):
        """
        Used for throttling create(signup) action
        """
        throttles = []
        if self.action == "create":
            throttles = [ScopedRateThrottle()]
        return throttles

    def perform_destroy(self, instance):
        if self.request.query_params:
            serializer = DeleteCurrentUserQuerySerializer(
                data=self.request.query_params
            )

            if serializer.validated_data.get("delete_orphan_organisations"):
                self.request.user.delete_orpan_organisations()

        super().perform_destroy(instance)
```

matthewelwell (changes requested):

api/custom_auth/urls.py
novakzaballa: As discussed, a generic name is better for this class since it should be used to override any behavior related with FFAdminUser view

api/custom_auth/views.py
novakzaballa: By renaming the class ThrottledUserViewSet to FFAdminUserViewSet the endpoint has been moved also.

api/custom_auth/views.py
novakzaballa: `perform_destroy` overridden. Also, as discussed the main business logic has been moved to the user model in the method delete().

api/custom_auth/views.py
matthewelwell: I still don't quite understand why it needed to be moved as it complicates the PR but it's fine for now.

api/custom_auth/views.py
```diff
@@ -30,7 +34,25 @@ class CustomAuthTokenLoginWithMFACode(AuthTokenLoginWithMFACode):
...
+@permission_classes([IsAuthenticated])
+def delete_token(request):
+    Token.objects.filter(user=request.user).delete()
+    user_logged_out.send(
+        sender=request.user.__class__, request=request, user=request.user
+    )
+    return Response(status=status.HTTP_204_NO_CONTENT)
+
+
+class UserDeleteQuerySerializer(serializers.Serializer):
```
matthewelwell: I think we can remove this now?

api/tests/unit/users/test_unit_users_views.py
```diff
@@ -0,0 +1,84 @@
...
+        return client.delete(
+            url, data=json.dumps(data), content_type="application/json"
+        )
+
+    # create a couple of users
+    email1 = "test1@example.com"
+    email2 = "test2@example.com"
+    email3 = "test3@example.com"
+    password = "password"
+    user1 = FFAdminUser.objects.create(email=email1)
```
matthewelwell: We can replace each of these with `FFAdminUser.objects.create_user(email=email1|2|3, password=password)`

frontend/common/services/useAccount.ts
```diff
@@ -0,0 +1,72 @@
...
+    endpoints: (builder) => ({
+      deleteAccount: builder.mutation<Res['account'], Req['deleteAccount']>({
+        invalidatesTags: [{ id: 'LIST', type: 'Account' }],
+        query: (query: Req['deleteAccount']) => ({
+          body: {
+            current_password: query.current_password,
+            delete_orphan_organisations: query.delete_orphan_organisations,
+          },
+          method: 'DELETE',
+          url: `auth/users/me`,
```
matthewelwell: Ah, this is an annoying one - we need to be specific about the trailing slashes here. 

```suggestion
          url: `auth/users/me/`,
```

frontend/common/services/useAccount.ts
```diff
@@ -0,0 +1,72 @@
...
+      }),
+      updateAccount: builder.mutation<Res['account'], Req['updateAccount']>({
+        invalidatesTags: (res) => [
+          { id: 'LIST', type: 'Account' },
+          { id: res?.id, type: 'Account' },
+        ],
+        query: (query: Req['updateAccount']) => ({
+          body: query,
+          method: 'PUT',
+          url: `auth/users/me`,
```
matthewelwell:
```suggestion
          url: `auth/users/me/`,
```

matthewelwell (changes requested):

api/custom_auth/views.py
novakzaballa: Yes, removed

api/tests/unit/users/test_unit_users_views.py
novakzaballa: Right. Replaced

frontend/common/services/useAccount.ts
novakzaballa: Done

matthewelwell (approved):

PR #2170, opened 2023-05-02
chore/remove-simply-from-docs

dabeeeenster:
As per https://justsimply.dev/


docs/docs/advanced-use/edge-proxy.md
```diff
@@ -162,10 +162,10 @@ Note that the Edge Proxy will currently _not_ send the Trait data back to the Co
 
 ## Architecture
 
-The standard Flagsmith architecture is very simple:
+The standard Flagsmith architecture:
 
 ![Image](/img/edge-proxy-existing.png)
 
-With the proxy added to the mix, things are still simple; they are just lower latency!
+With the proxy added to the mix they are just lower latency!
```
matthewelwell: This sentence no longer really makes sense.

docs/docs/deployment/hosting/kubernetes.md
```diff
@@ -20,8 +20,8 @@ Then view `http://localhost:8080` in a browser. This will install using default
 
 ### Ingress configuration
 
-The above is a quick and simple way of gaining access to Flagsmith, but in many cases will need to configure ingress to
-work with an ingress controller.
+The above is a quick way of gaining access to Flagsmith, but in many cases will need to configure ingress to work with
```
matthewelwell:
```suggestion
The above is a quick way of gaining access to Flagsmith, but in many cases you will need to configure ingress to work with
```

docs/docs/quickstart.md
```diff
@@ -32,20 +32,19 @@ Flagsmith manages Flags with Projects, so let's create one now:
...
-first Flag. This flag is simply going to control whether a button shows on our super simple web page.
+first Flag. This flag will control whether a button shows on our web page.
 
 <div style={{textAlign: 'center'}}><img width="75%" src="/img/quickstart/demo_create_3.png"/></div>
 
 Flags within Flagsmith are a combination of both: A Boolean value - the `Flag State` and then optionally: A
-String/Integer/Float value - the `Flag Value`. For now, we're just going to use the `Boolean` value of the flag to
-control whether the button shows. Let's go ahead and create a flag called `show_demo_button`. We're going to leave it as
-Disabled by default:
+String/Integer/Float value - the `Flag Value`. For now, we're going to use the `Boolean` value of the flag to control
```
matthewelwell: Feels like this changes the meaning of the sentence. Perhaps we should change `just` to `only` instead? 

```suggestion
String/Integer/Float value - the `Flag Value`. For now, we're only going to use the `Boolean` value of the flag to control
```

matthewelwell (changes requested):

matthewelwell (approved):

PR #2172, opened 2023-05-02
improvement/mixpanel-webhooks-cohorts

dabeeeenster:
Explain how to get Mixpanel Cohorts into Flagsmith

### DIFF
--- a/docs/docs/integrations/mixpanel.md
+++ b/docs/docs/integrations/mixpanel.md
@@ -75,8 +75,12 @@ Every time an `Identity` requests their flags from the Flagsmith API, Flagsmith 
 
 ## Getting Mixpanel Cohorts into Flagsmith
 
-You need to send data in the other direction, and push cohorts from MixPanel into Flagsmith, using
-[Mixpanel Webhooks](https://developer.mixpanel.com/docs/cohort-webhooks).
+If you want to control Flagsmith Flags based on Cohorts in Mixpanel, you will need to send cohort data from Mixpanel
+into Flagsmith. There is no way currently of automating this process from the Mixpanel side, but you can send cohorts
+from MixPanel into Flagsmith, using [Mixpanel Webhooks](https://developer.mixpanel.com/docs/cohort-webhooks).
+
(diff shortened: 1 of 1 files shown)


docs/docs/integrations/mixpanel.md
```diff
@@ -72,3 +72,39 @@ Every time an `Identity` requests their flags from the Flagsmith API, Flagsmith
  }
 }
 ```
+
+## Getting Mixpanel Cohorts into Flagsmith
```
matthewelwell, line 76: Ok, I think I understand now. So the idea is: 

Identity cohort changes in Mixpanel -> Triggers Webhook -> Hits endpoint on customer infrastructure -> Customer triggers request to Flagsmith to set traits. 

I don't think we've mentioned here anywhere that there will need to be that 'middleman' service in the client's infrastructure. Perhaps it's obvious but I think it's best to be explicit.  

We should probably make also sure people are aware that they can remove traits by setting the `trait_value` as `null`

docs/docs/integrations/mixpanel.md
```diff
@@ -72,3 +72,39 @@ Every time an `Identity` requests their flags from the Flagsmith API, Flagsmith
...
+curl -X "PUT" "https://edge.api.flagsmith.com/api/v1/traits/bulk/" \
+     -H 'X-Environment-Key: XXXXXX' \
+     -H 'Content-Type: application/json' \
+     -d $'[
+  {
+    "trait_value": 3.14,
+    "identity": {
+      "identifier": "bennyr"
+    },
+    "trait_key": "floaty"
```
matthewelwell: Can we give a useful example here? Perhaps one of the following:

```json
{
    "trait_value": "cohort_a,cohort_b",
    "identity": {
      "identifier": "bennyr"
    },
    "trait_key": "mixpanel_cohorts"
}
```

```json
{
    "trait_value": true,
    "identity": {
      "identifier": "bennyr"
    },
    "trait_key": "mp_cohort_a"
}
```

(Slight preference for option 1)

matthewelwell (approved):

PR #2179, opened 2023-05-03
fix(audit/models): trigger sse messages correctly

gagantrivedi:

## Changes

Update `update_environments_updated_at` method to trigger SSE messages for every type of audit log apart from change request

Adds unit test cases


api/audit/models.py
```diff
@@ -72,15 +76,15 @@ def get_history_record_model_class(
         module = import_module(module_path)
         return getattr(module, class_name)
 
+    @property
+    def can_related_object_type_change_feature_value_for_sdk(self) -> bool:
```
matthewelwell: I'd say, since this is a property, it can be simplified to `affects_sdk_feature_value` ? Or we can be more explicit (if it's only SSE based) and say `should_trigger_sse_event`?

api/audit/models.py
```diff
@@ -72,15 +76,15 @@ def get_history_record_model_class(
...
+        # Returns true if the audit log have an object type that can change the
+        # feature value for the sdk endpoints
+        return self.related_object_type != RelatedObjectType.CHANGE_REQUEST.name
+
     @hook(AFTER_SAVE)
     def update_environments_updated_at(self):
```
matthewelwell: Since we're now actually triggering the SSE messages, the name of this method should probably change?

api/audit/signals.py
```diff
@@ -127,15 +123,6 @@ def send_environments_to_dynamodb(sender, instance, **kwargs):
     )
 
 
-@receiver(post_save, sender=AuditLog)
-@handle_skipped_signals
-def trigger_environment_update_messages(sender, instance, **kwargs):
```
matthewelwell, line 132: What's the logic behind removing this and moving it to the lifecycle hook method? Just simplification? Seems odd that we decorated it with `handle_skipped_signals` but there are no side effects to not having it?

api/audit/models.py
```diff
@@ -93,6 +97,14 @@ def update_environments_updated_at(self):
             updated_at=self.created_date
         )
 
+        if self.environment_id:
+            environment = self.environment
+            # update manually to save a db call
```
matthewelwell: We should probably just add a little more detail to this comment? Does the environment's value actually get updated elsewhere in the flow?

gagantrivedi: No, it does not

api/audit/signals.py
gagantrivedi: Yeah, just simplification, the skip signal was never configured for this - we only have skip signal for sending environment to dynamo

api/audit/models.py
matthewelwell: Then it never gets written to the database?

gagantrivedi: Oh, no, that happens here: https://github.com/Flagsmith/flagsmith/pull/2179/files#diff-72a451e6a3921d402b47d41355d1cc5822092dd5cb4e2c85d13b38391033e879L92

it is to save refresh_from_db

matthewelwell: Ah, yep, got it!

api/audit/signals.py
gagantrivedi: Another fun fact, signals are called before hooks

api/audit/models.py
```diff
@@ -72,17 +82,13 @@ def get_history_record_model_class(
         module = import_module(module_path)
         return getattr(module, class_name)
 
-    @hook(AFTER_SAVE)
+    @hook(
+        AFTER_SAVE,
+        priority=priority.HIGHEST_PRIORITY,
```
matthewelwell, line 104: Oh, cool - does this define the order in which hooks are run?

api/features/workflows/core/models.py
```diff
@@ -181,15 +183,25 @@ def email_subject(self):
...
+            if self.committed_at < feature_state.live_from:
+                create_feature_state_went_live_audit_log.delay(
+                    delay_until=feature_state.live_from, args=(feature_state.id,)
+                )
+            else:
+                message = FEATURE_STATE_UPDATED_BY_CHANGE_REQUEST_MESSAGE % (
+                    feature_state.feature.name,
+                    feature_state.change_request.title,
+                )
+                AuditLog.objects.create(
```
matthewelwell: We should defer this to a task still for simplicity I think?

api/features/workflows/core/models.py
```diff
@@ -181,15 +183,25 @@ def email_subject(self):
...
+                    related_object_id=feature_state.id,
+                    related_object_type=RelatedObjectType.FEATURE_STATE.name,
+                    environment=feature_state.environment,
+                    project=feature_state.environment.project,
+                    log=message,
+                    is_system_event=True,
```
matthewelwell: Hmm... I guess this is an interesting question. Is it a system event or an action performed by the change request creator or the CR committer?

api/tests/unit/features/workflows/core/test_unit_workflows_models.py
```diff
@@ -444,31 +446,35 @@ def test_change_request_email_subject(change_request_no_required_approvals):
...
+    change_request_no_required_approvals.commit(committed_by=admin_user)
 
     # Then
     mock_create_feature_state_went_live_audit_log.delay.assert_not_called()
+    for feature_state in change_request_no_required_approvals.feature_states.all():
+        log = FEATURE_STATE_UPDATED_BY_CHANGE_REQUEST_MESSAGE % (
+            feature_state.feature.name,
+            feature_state.change_request.title,
+        )
+        AuditLog.objects.get(
```
matthewelwell: probably best to do something like: 

```python
assert AuditLog.objects.filter(...).count() == 1
```

api/tests/unit/organisations/invites/test_unit_invites_models.py
```diff
@@ -12,6 +12,7 @@
     from pytest_django.fixtures import SettingsWrapper
 
 
+@pytest.mark.django_db
```
matthewelwell, line 15: Odd, I'm not sure how this test was ever passing?

api/audit/constants.py
```diff
@@ -12,6 +12,9 @@
     "Flag state / Remote Config value update scheduled for %s for feature: %s"
 )
 FEATURE_STATE_UPDATED_MESSAGE = "Flag state updated for feature: %s"
+FEATURE_STATE_UPDATED_BY_CHANGE_REQUEST_MESSAGE = (
+    "Flag state / Remote config updated for feature: %s by" " Change Request: %s"
```
matthewelwell: Ah, looks like black has reformatted his badly!

api/audit/models.py
gagantrivedi: Yep

api/tests/unit/organisations/invites/test_unit_invites_models.py
gagantrivedi: It is caused by upgrading the lifecycle hook - they have "refactored" some bits

matthewelwell: Ah, ok! 👍

api/features/workflows/core/models.py
gagantrivedi: Yeah, good question, I am leaning towards the request creator? 
Note: we also need to change this for feature state going live (it uses system event)

api/audit/models.py
```diff
@@ -44,10 +54,10 @@ class AuditLog(LifecycleModel):
     related_object_type = models.CharField(max_length=20, null=True)
     related_object_uuid = models.CharField(max_length=36, null=True)
 
-    skip_signals = models.CharField(
+    skip_subscribers = models.CharField(
```
gagantrivedi: I am still a bit uncertain about this name (and the way it is used) Is this a case of YAGNI(maybe we just need a boolean called `environment_updated`?) what do you think?

api/audit/migrations/0012_rename_skip_signals_auditlog_skip_subscriber.py
```diff
@@ -0,0 +1,27 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("audit", "0011_auditlog_related_object_uuid"),
+    ]
+
+    operations = [
+        migrations.RenameField(
```
matthewelwell: Erk, we can't do this! It will bring the whole API down while it rewrites the whole AuditLog table (which is huge!). Should we just change the field name in the code and use the existing database column?

api/audit/models.py
```diff
@@ -72,17 +91,13 @@ def get_history_record_model_class(
         module = import_module(module_path)
         return getattr(module, class_name)
 
-    @hook(AFTER_SAVE)
+    @hook(
+        AFTER_CREATE,
+        priority=priority.HIGHEST_PRIORITY,
+        when="environment_document_updated",
```
matthewelwell, line 105: Ah, I didn't realise you could use properties here, that's quite cool.

api/audit/models.py
matthewelwell: We could just be more explicit and call it `skip_signals_and_hooks` or something (with the above caveat that the column name has to remain the same)? I think it gives us a good amount of flexibility so I'm keen to keep it how it is.

api/audit/migrations/0012_rename_skip_signals_auditlog_skip_subscriber.py
gagantrivedi: Nope, I don't think that's true - you can even change the size of varchar column without rewrite https://wiki.postgresql.org/wiki/What%27s_new_in_PostgreSQL_9.2#Reduce_ALTER_TABLE_rewrites

api/audit/models.py
gagantrivedi: Yeah, it's pretty cool

api/audit/migrations/0012_rename_skip_signals_auditlog_skip_subscriber.py
matthewelwell: Sorry, maybe my wording was incorrect here. The issue is that we'll have a schema mismatch between the old instances and the new ones so the API will be offline while the new deployment is rolled out.

api/audit/models.py
```diff
@@ -93,6 +111,34 @@ def update_environments_updated_at(self):
...
+        Environment.write_environments_to_dynamodb(
+            environment_id=self.environment_id, project_id=self.project_id
+        )
+
+    @hook(
+        AFTER_CREATE,
+        when="environment_document_updated",
+        is_now=True,
+    )
+    def send_environment_update_message(self):
```
matthewelwell, line 131: So, as I understand it, this is all happening in the same thread when an AuditLog gets written. So a task runner, when it creates an AuditLog will have to: 

 1. Create the AuditLogRecord
 2. Update any associated environments in the db with the new `updated_at` value
 3. Write the updated document to dynamodb
 4. Send the update message to fastly for real time

I'm not sure if this is an issue or not, but it is a _little_ difficult to follow. Perhaps it's something that we can refactor later?

gagantrivedi: > So, as I understand it, this is all happening in the same thread when an AuditLog gets written. So a task runner, when it creates an AuditLog will have to:
> 
>     1. Create the AuditLogRecord
> 
>     2. Update any associated environments in the db with the new `updated_at` value
> 
>     3. Write the updated document to dynamodb
> 
>     4. Send the update message to fastly for real time
> 
> 
> I'm not sure if this is an issue or not, but it is a _little_ difficult to follow. Perhaps it's something that we can refactor later?

It's mostly correct except the 4th point - it creates a task to send the SSE message

api/audit/models.py
```diff
@@ -94,12 +94,22 @@ def get_history_record_model_class(
...
+        if self.environment and self.project is None:
+            self.project = self.environment.project
+
     @hook(
         AFTER_CREATE,
         priority=priority.HIGHEST_PRIORITY,
         when="environment_document_updated",
         is_now=True,
     )
+    def process_environment_update(self):
```
matthewelwell, line 108: Love this - so much cleaner!

matthewelwell (approved):

PR #2181, opened 2023-05-04
Minor updates to Uffizzi GHA workflows to bring in line with other projects

axisofentropy:
## Changes

The most important is the error handling when GitHub's Artifact Registry hiccups and does not deliver. This causes a terribly ambiguous error you can see here: https://github.com/Flagsmith/flagsmith/actions/runs/4829275209/jobs/8604115305#step:2:31

These updates are borrowed from other open-source projects that use Uffizzi. Unfortunately these workflows only execute after they're merged to the project's default branch.


dabeeeenster:
LGTM

matthewelwell (approved):

PR #2182, opened 2023-05-04
Prevent deleted API keys from being returned

matthewelwell:
## Changes

Ensures that deleted MasterAPIKeys are not returned by the viewset after being deleted. 

Added a new integration test.


api/tests/integration/api_keys/test_viewset.py
```diff
@@ -125,3 +125,42 @@ def test_create_master_api_key_ignores_organisation_in_body(admin_client, organi
 
     assert list_response_json["results"][0]["name"] == name
     assert key.startswith(list_response_json["results"][0]["prefix"])
+
+
+def test_deleted_api_key_is_not_returned_in_list_and_cannot_be_used(
+    admin_client, organisation, master_api_key, master_api_key_client
```
khvn26: nit: typing

khvn26 (approved):

api/tests/integration/api_keys/test_viewset.py
matthewelwell: Good point - have added.

api/tests/integration/api_keys/test_viewset.py
```diff
@@ -125,3 +126,45 @@ def test_create_master_api_key_ignores_organisation_in_body(admin_client, organi
 
     assert list_response_json["results"][0]["name"] == name
     assert key.startswith(list_response_json["results"][0]["prefix"])
+
+
+def test_deleted_api_key_is_not_returned_in_list_and_cannot_be_used(
+    admin_client: APIClient,
+    organisation: int,
+    master_api_key: dict,
```
khvn26:
```suggestion
    master_api_key: typing.Tuple["AbstractAPIKey", str],
```

api/tests/integration/api_keys/test_viewset.py
```diff
@@ -125,3 +126,45 @@ def test_create_master_api_key_ignores_organisation_in_body(admin_client, organi
...
+    master_api_key_client: APIClient,
+):
```
khvn26:
```suggestion
) -> None:
```

khvn26 (changes requested):

api/tests/integration/api_keys/test_viewset.py
matthewelwell: I'm not sure I agree that we should add the return type to all of our tests? I think that `None` is assumed to be the default return type, and we can be fairly confident that all test functions / methods have no return type. Happy to discuss this further though if you have a strong opinion on this.

khvn26: I would agree with you if omitting a return type would mean an implicit `None`. Unfortunately, it means implicit `Any` , so even obvious cases like tests and `__init__` methods have to be annotated if we want to come to a fully typed project in the future.

khvn26 (approved):

Issue #2151, opened 2023-04-24
Add Makefile for `api` + integrate dotenv


khvn26:
Provide a one stop shop for all development needs by writing a `Makefile` and integrating dotenv/direnv.

This is just for backend for now.

## Targets

- [ ] poetry + dependencies setup
- [ ] tests (including e2e?)
- [x] linters
- [x] docker-compose up/down/logs
- [x] docker build

## Environment

- [x] `DJANGO_SETTINGS_MODULE`
- [x] `DATABASE_URL`

PR #2185, opened 2023-05-04
Add Makefile

khvn26:

## Changes

Contributes to #2151.

This adds a Makefile along with the default dotenv file to the `api` project and modifies Python tools' configurations to accommodate for those.

The Makefile has been tested and refined for some time. It has been my personal daily driver for a couple of weeks. I went ahead and migrated the API PR workflow to Makefile.

The following targets are added:

- `docker-up`, `docker-logs` and `docker-down`: manage the Docker Compose project to bring up dependencies for Flagsmith API (currently just Postgres defined in `docker/db.yaml`). Configurable via Compose's default environment variables.
- `docker-build`: build the API image and tag it with the `flagsmith-api:local` tag. Tag name configurable.
- `test`: Run Pytest. Pytest configuration resides in `pyproject.toml` now and establishes sane defaults: verbosity level to include full test names and the Django settings module. Additional options may be provided with an `opts` variable.
- `django-migrations`: Autogenerate migrations. Specify appts with an `opts` variable.
- `django-migrate`: Identical to `run-docker.sh`'s `migrate` command.
- `django-serve`: Spin up a gunicorn instance with live-reload enabled.
- `lint`: Runs linters (`lint-black`, `lint-isort` and `lint-flake8`).
- `install`: Installs dependencies (currently just `install-packages`, which installs dev dependencies too).

- Ran every Makefile target and observerd the results:

  1. Linter behaviour is identical to `pre-commit`.
  2. `pytest` runs with correct Django settings module.
  3. Docker Compose brings up the database with correct credentials and applies the default project name.
  4. `make docker-build` applies the configured tag.
  5. `make django-migrate` and `make django-migrations` wait for the database so i.e. `make docker-up django-migrate` won't crash.
  6. `make django-serve` spins up a working API server and the live-reload works.

- `install`, `lint` and `test` targets run in CI!

### DIFF
--- a/api/Makefile
+++ b/api/Makefile
@@ -10,12 +10,16 @@ DOTENV_OVERRIDE_FILE ?= .env
 -include .env-local
 -include $(DOTENV_OVERRIDE_FILE)
 
+.PHONY: install-pip
+install-pip:
+	python -m pip install --upgrade pip
+
 .PHONY: install-packages
 install-packages:
 	pip install -r requirements-all.txt
 
 .PHONY: install
-install: install-packages
+install: install-pip install-packages
 
 .PHONY: lint-black
 lint-black:
--- a/api/Makefile
+++ b/api/Makefile
@@ -60,15 +60,19 @@ docker-build:
 test:
 	pytest $(opts)
 
-.PHONY: migrations
-migrations:
+.PHONY: django-migrations
+django-make-migrations:
 	python manage.py waitfordb
 	python manage.py makemigrations $(opts)
 
-.PHONY: migrate
-migrate:
+.PHONY: django-migrate
+django-migrate:
 	python manage.py waitfordb
 	python manage.py migrate
+
+.PHONY: django-collect-static
+django-collect-static:
+	python manage.py collectstatic --noinput
 
 .PHONY: serve
 serve:
--- a/api/Makefile
+++ b/api/Makefile
@@ -1,0 +1,71 @@
+.EXPORT_ALL_VARIABLES:
+
+DOCKER_TAG ?= flagsmith/flagsmith-api:local
+
+COMPOSE_FILE ?= ../docker/db.yaml
+COMPOSE_PROJECT_NAME ?= flagsmith
+
+DOTENV_OVERRIDE_FILE ?= .env
+
+-include .env-local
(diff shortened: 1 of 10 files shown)


matthewelwell:
image

dabeeeenster:
Yep really like this - we need to do the same for the frontend and docs I think!

khvn26:
> we need to do the same for the frontend and docs I think!

Agreed. I suggest we merge this first and roll with it for a while. When the target list is settled, we'll be able to write a root Makefile too and document its usage for contributors.

dabeeeenster:
Yep agree. Not sure if its best to have 1 makefile for the monorepo or individual ones in subfolders?

api/Makefile
```diff
@@ -0,0 +1,75 @@
...
+		--tag=$(DOCKER_TAG) \
+		--file=Dockerfile \
+		.
+
+.PHONY: test
+test:
+	pytest $(opts)
+
+.PHONY: migrations
+migrations:
```
dabeeeenster: maybe call this `django-migrations`

api/Makefile
```diff
@@ -0,0 +1,75 @@
...
+	python manage.py waitfordb
+	python manage.py makemigrations $(opts)
+
+.PHONY: migrate
```
dabeeeenster: and this `django-migrate`

api/Makefile
```diff
@@ -0,0 +1,75 @@
...
+migrate:
+	python manage.py waitfordb
+	python manage.py migrate
+
+.PHONY: serve
+serve:
```
dabeeeenster, line 78: and this `django-serve`

api/Makefile
khvn26: Sure. Done.

api/Makefile
khvn26: Done.

api/Makefile
khvn26: Done.

khvn26:
> Not sure if its best to have 1 makefile for the monorepo or individual ones in subfolders?

I prefer individual ones with sensible common targets in the root (to be able to e.g. `collectstatic` after bundling the frontend).

dabeeeenster (approved):

PR #2190, opened 2023-05-08
Add specific permission to view audit logs

### LINKED ISSUE #2174: Add specific permission to view audit logs

Users want to be able to give permissions to users other than organisation admins to view the audit log records. 

A specific permission (for now, at the organisation level) to give users permission to view audit log records. 

The only alternative is giving users organisation admin permissions which is not a viable solution in a lot of cases. 

**Additional context**

N/a

matthewelwell:
## Changes

Adds a specific permission at the organisation level for viewing Audit Log records. 

Note that as part of this, I've had to add a new endpoint since the current endpoint gets all of the audit log records for all organisations that a user belongs to. 

The new endpoint is nested beneath the organisation so that it can be filtered accordingly and permissions can be applied. This will not affect the UI since the audit log is _always_ currently shown in the context of a project (which 
always belongs to a single organisation). 

Added unit tests


api/audit/permissions.py
```diff
@@ -0,0 +1,15 @@
+from rest_framework.permissions import BasePermission
+
+from organisations.models import Organisation
+from organisations.permissions.permissions import VIEW_AUDIT_LOG
+
+
+class OrganisationAuditLogPermissions(BasePermission):
+    def has_permission(self, request, view):
```
khvn26:
```suggestion
    def has_permission(self, request: HTTPRequest, view: APIView) -> bool:
```

api/audit/tests/test_views.py
```diff
@@ -87,3 +88,40 @@ def test_audit_log_can_be_filtered_by_is_system_event(
     assert response.json()["count"] == 1
     assert response.status_code == status.HTTP_200_OK
     assert response.json()["results"][0]["is_system_event"] is True
+
+
+def test_regular_user_cannot_list_audit_log(
+    project, environment, organisation, django_user_model, api_client
```
khvn26: nit: typing here and all tests below

api/audit/views.py
```diff
@@ -1,53 +1,60 @@
...
-    ),
+    decorator=swagger_auto_schema(query_serializer=AuditLogsQueryParamSerializer()),
 )
-class AuditLogViewSet(mixins.ListModelMixin, viewsets.GenericViewSet):
+class _BaseAuditLogViewSet(mixins.ListModelMixin, viewsets.GenericViewSet):
     serializer_class = AuditLogSerializer
     pagination_class = CustomPagination
     filterset_fields = ["is_system_event"]
 
     def get_queryset(self):
```
khvn26:
```suggestion
    def get_queryset(self) -> QuerySet[AuditLog]:
```

khvn26 (approved):

matthewelwell:
@khvn26 I added all the typing as per your comments.

PR #2205, opened 2023-05-15
ci: Add functionality to create defaults in docker compose

### LINKED ISSUE #2601: Add self-hosted docker environment variables for user login credentials to support automated testing

I am trying to incorporate flagsmith into my integration tests using testcontainers and I would like to be able to programmatically login so that I can obtain an App key to run various REST requests and SDK functions.

The docs suggest I need to manually create a user everytime i run a new docker container instance. I cant rely on volumes for persisting user data. https://docs.flagsmith.com/deployment/hosting/docker

I can see there are a variety of env vars supported: https://github.com/Flagsmith/self-hosted/blob/main/docker-compose.yml, would it be possible to add FLAGSMITH_USER, FLAGSMITH_PASSWORD in a similar fashion to postgres to give a predictable user to login with?

Include FLAGSMITH_USER and FLAGSMITH_PASSWORD environment variables on the Flagsmith container to create a default user without having to manually register on startup.

Doing a postgres DB dump and pulling out the API key. This is obviously less than ideal.

**Additional context**

matthewelwell:
(updated by @khvn26)

## Changes

Adds `ALLOW_ADMIN_INITIATION_VIA_CLI` setting, set to `False` by default, that enables the `python manage.py bootstrap` management command.

The command detects an empty Flagsmith instance and initializes a superuser, an organisation, and a project. It does NOT do anything if the Flagsmith instance is not empty.

Additionally, the PR introduces individual commands for creating the default user, org, and project separately, with overridable defaults.

Ran locally against a fresh database and one from a well-used Flagsmith instance, and observed expected behaviour.


api/core/management/commands/initselfhosted.py
```diff
@@ -0,0 +1,79 @@
...
+        organisation = None
+        project = None
+
+        if should_create_user():
+            user = user_model.objects.create_user(
+                email=settings.DEFAULT_USER_EMAIL,
+                password=settings.DEFAULT_USER_PASSWORD,
+                is_superuser=settings.DEFAULT_USER_IS_SUPERUSER,
+            )
+            logger.info(
```
khvn26: This and the following logs look like they should be command output via `self.stdout.write`.

docker-compose.yml
```diff
@@ -10,6 +10,16 @@ services:
       POSTGRES_DB: flagsmith
     container_name: flagsmith_postgres
 
+  migrator:
```
khvn26: Init containers are cool but for the compose file I'd prefer too keep things as lean as possible to ease the cognitive load for the users.

Also `depends_on` isn't reliable without `condition` and throwing in `condition` makes us open a can of worms in regards to Compose's versioning.

khvn26 (dismissed):

api/core/management/commands/initselfhosted.py
matthewelwell: Yep, makes sense, will change.

docker-compose.yml
matthewelwell: Yeah, agreed. Will update.

docker-compose.yml
```diff
@@ -68,3 +80,21 @@ services:
       - postgres
     links:
       - postgres
+
+  # Use this container to initialise a default user / organisation / project in your installation.
+  # Any of the DEFAULT_* environment variables can be removed to only create certain entities as required.
+  #  init:
```
khvn26: I don't at all love the idea of shipping commented code for the users to uncomment to enable something... just yuck.

dabeeeenster: I cant really think of a neater way of doing this. It's also pretty common in docker-compose files to uncomment snippets to enable and disable services.

laurencefass:
Im coming here from https://github.com/Flagsmith/flagsmith/issues/2601#issuecomment-1670893352. 

Comment says "fixed".

Is this now available in self hosted docker image?

Any idea when this will be available for deployment?

Thanks.

matthewelwell:
> Im coming here from [#2601 (comment)](https://github.com/Flagsmith/flagsmith/issues/2601#issuecomment-1670893352).
> 
> Is this now available in self hosted docker image?
> 
> Any idea when this will be available for deployment?
> 
> Thanks.

Hi @laurencefass, this is not yet available anywhere. There is some additional work that needs to be done for this PR to be ready. We're hoping to finish it up in the next couple of weeks but other priorities have taken over a little. You can follow this PR and, once it's merged, you'll get a notification.

khvn26 (commented):
Un-requesting changes, not approving as it's my code now that needs someone else's approval.

api/projects/management/commands/createinitialproject.py
```diff
@@ -0,0 +1,54 @@
...
+    ) -> None:
+        organisation = Organisation.objects.filter(
+            name=organisation_name or settings.ORGANISATION_NAME
+        ).first()
+        if (
+            not settings.ALLOW_ADMIN_INITIATION_VIA_CLI
+            or not organisation
+            or Project.objects.count()
+        ):
+            logger.debug("Skipping initial organisation creation.")
```
khvn26:
```suggestion
            logger.debug("Skipping initial project creation.")
```

api/organisations/management/commands/createinitialorganisation.py
```diff
@@ -0,0 +1,50 @@
...
+            dest="organisation_name",
+            help="Organisation name",
+            default=None,
+        )
+
+    def handle(
+        self,
+        *args: Any,
+        admin_email: str | None,
+        organisation_name: str | None,
```
zachaysan: It's strange to me that we allow no email and default to admin@example.com I take it this command is for people that want to spin something up really fast and use it right away?

api/tests/integration/ core/test_commands.py
```diff
@@ -0,0 +1,138 @@
...
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    # Given
+    expected_user_email = "test@example.com"
+    expected_user_password = "foobar"
+    expected_organisation_name = "Test Org"
+    expected_project_name = "Test Project"
+
+    # simulate pipe
+    monkeypatch.setattr(sys, "stdin", io.StringIO(expected_user_password + "\n"))
```
zachaysan: This is cool.

api/users/forms.py
```diff
@@ -29,11 +29,9 @@ class InitConfigForm(forms.Form):
     )
 
     def create_admin(self):
```
zachaysan: Can `create_admin` be run multiple times? If so, the `create_initial_superuser` function name isn't quite right since it could be used for multiple superusers.

zachaysan (approved):
I have some questions, but you probably have good answers to them. Approved.

api/users/forms.py
khvn26: > Can `create_admin` be run multiple times?

It's not supposed to be as the view using the form gets closed off after the user is created.

api/tests/integration/ core/test_commands.py
khvn26: I don't quite like monkeypatching (and ORM calls) in my integration tests... but thanks!

api/organisations/management/commands/createinitialorganisation.py
khvn26: Yes. The PR does add a `bootstrap` call to the entrypoint so the folks spinning up a new instance are going to get the superuser, org and project right away, unless `ALLOW_ADMIN_INITIATION_VIA_CLI` is set to `false`.

api/organisations/views.py
```diff
@@ -187,9 +187,9 @@ def get_subscription_metadata(self, request, pk):
     @action(detail=True, methods=["GET"], url_path="portal-url")
     def get_portal_url(self, request, pk):
         organisation = self.get_object()
-        if not organisation.has_paid_subscription():
+        if not organisation.has_subscription():
```
matthewelwell: This looks like a bad merge?

api/organisations/management/commands/createinitialorganisation.py
```diff
@@ -0,0 +1,50 @@
...
+        **options: Any,
+    ) -> None:
+        if not settings.ALLOW_ADMIN_INITIATION_VIA_CLI or Organisation.objects.exists():
+            logger.debug("Skipping initial organisation creation.")
+            return
+        organisation = Organisation.objects.create(
+            name=organisation_name or settings.ORGANISATION_NAME
```
matthewelwell: What's the benefit to this approach vs setting the default in the parser argument above?

api/tests/integration/ core/test_commands.py
```diff
@@ -0,0 +1,161 @@
...
+    project = Project.objects.first()
+    assert project.name == expected_project_name
+    assert project.organisation == organisation
+
+    # Bootstrap logged as expected.
+    capture_result = capsys.readouterr()
+    assert not capture_result.err
+    assert 'Superuser "test@example.com" created successfully.' in capture_result.out
+    assert 'Organisation "Test Org" created successfully.' in capture_result.out
+    assert 'Project "Test Project" created successfully.' in capture_result.out
```
matthewelwell: Shouldn't these use the `expected_<entity>` arguments?

api/tests/integration/ core/test_commands.py
```diff
@@ -0,0 +1,161 @@
...
+def test_bootstrap__empty_instance__cli_overrides__creates_expected(
+    settings: SettingsWrapper,
+    capsys: pytest.CaptureFixture,
+) -> None:
+    # Given
+    settings.ALLOW_ADMIN_INITIATION_VIA_CLI = True
+
+    expected_user_email = "test@example.com"
+    expected_organisation_name = "Test Org"
+    expected_project_name = "Test Project"
```
matthewelwell, line 109: These seem to be the same as above (which is why the string asserts above are passing still).

api/tests/integration/ core/test_commands.py
```diff
@@ -0,0 +1,161 @@
...
+    expected_organisations = Organisation.objects.all()
+    expected_projects = Project.objects.all()
+
+    # When
+    call_command("bootstrap")
+
+    # Then
+    FFAdminUser.objects.all() == expected_users
+    Organisation.objects.all() == expected_organisations
+    Project.objects.all() == expected_projects
```
matthewelwell: I'm dubious as to whether these are valid tests. I don't quite know enough about the internals of the django ORM but I would be concerned that actually all we're doing here is e.g. `FFAdminUser.objects.all() == FFAdminUser.objects.all()` because there is no reason for django to evaluate the queryset until this point.

api/organisations/views.py
khvn26: Fixed! Thanks for catching this.

api/organisations/management/commands/createinitialorganisation.py
khvn26: Good point, updated.

api/tests/integration/ core/test_commands.py
khvn26: Agreed, done.

api/tests/integration/ core/test_commands.py
khvn26: Code above updated so this is resolved.

api/tests/integration/ core/test_commands.py
khvn26: This is a valid concern. Added the explicit list coercion.

api/organisations/management/commands/createinitialorganisation.py
matthewelwell: We can now change this to the following though, right? 

```python
organisation = Organisation.objects.create(name=organisation_name)
```

khvn26: Oops! Done.

matthewelwell (commented):
As it's my original PR, I can't approve but ✅

PR #2207, opened 2023-05-15
Fix view / manage identity

kyle-ssg:

## Changes

 - Viewing a user page was based on manage_identities rather than view_identities
 - Editing a feature state for a user was based on manage_feature_states rather than manage_identities

A note whilst testing this, I had to remember this but manage identities does not include the ability to edit their feature states.

### DIFF
--- a/frontend/web/components/pages/UserPage.js
+++ b/frontend/web/components/pages/UserPage.js
@@ -272,595 +272,545 @@ const UserPage = class extends Component {
         id={environmentId}
       >
         {({ permission: manageUserPermission }) => (
-          <div className='app-container'>
-            <IdentityProvider onSave={this.onSave}>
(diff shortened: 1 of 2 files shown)


matthewelwell:
@kyle-ssg so, just to clarify, the permissions from the FE perspective should now be as follows: 

VIEW_IDENTITIES - users can list identities and click through to a read only view of an identity
MANAGE_IDENTITIES - users can edit an identity's traits
UPDATE_FEATURE_STATE - users can edit an identity's feature states (as well as all other feature states in the environment)

is that correct?

kyle-ssg:
Yep @matthewelwell with the addition that manage identities can also create new ones too

matthewelwell (approved):

PR #2211, opened 2023-05-17
feat/add-org-id-to-settings

dabeeeenster:
Shows the organisation ID which can be helpful durng support etc.

Also tidied up some styling inconsistencies

image


novakzaballa (approved):
Looks good to me. The only question I have is why we are stop using the `<Row>` and `<Column>` components?

PR #2214, opened 2023-05-17
feat: Migrate to poetry

### LINKED ISSUE #2150: Migrate to Poetry
Python packaging [isn't solved yet](https://github.com/mitsuhiko/rye/discussions/6) but [Poetry](https://python-poetry.org/docs/) is the most popular closest-to-the-solution thing out there.

With Poetry, we'll get:

- Dependency locks.
- Automatic venvs.
- Python version pinning.
- Runtime and dev dependencies managed in `pyproject.toml`.
- Neat build targets in a future `Makefile`.

## Migration checklist

- [ ] `api` project is Poetry-installable.
- [ ] Build isn't broken for both `Dockerfile`s.
- [ ] All `api-` and `platform-` workflows are not broken.
- [ ] Update documentation

tushar5526:
Closes #2150


api/pyproject.toml
```diff
@@ -35,3 +35,121 @@ skip = ['migrations','.venv','.direnv']
 
 [tool.pytest.ini_options]
 addopts = ['--ds=app.settings.test', '-vvvv', '-p', 'no:warnings', '-n', 'auto']
+
+[tool.poetry]
+name = "y"
```
tushar5526: Can you share the values for the following fields?

tushar5526:
Do you also want Dockerfile.redhat-ubi to be migrated as well ?

dabeeeenster:
> Do you also want Dockerfile.redhat-ubi to be migrated as well ?

Yes please! Also the Dockerfile in the root of the repo which is our main file

dabeeeenster:
Great thanks! cc @matthewelwell there are also a few ECS related Github actions that we need to update

dabeeeenster:
Also the makefile

api/pyproject.toml
```diff
@@ -35,3 +35,121 @@ skip = ['migrations','.venv','.direnv']
...
+version = "0.1.0"
+description = ""
+authors = ["Your Name <you@example.com>"]
+readme = "README.md"
+
+[tool.poetry.dependencies]
```
khvn26, line 46: We're missing dev dependencies. Please add them as an [optional group](https://python-poetry.org/docs/managing-dependencies/#optional-groups).

api/Dockerfile
```diff
@@ -4,20 +4,14 @@ WORKDIR /app
...
 RUN pip install --upgrade setuptools
+RUN pip install poetry
+# Set up venv
+RUN python -m venv /opt/venv
 
-# Install our python dependencies
-RUN pip install -r requirements.txt
-
+COPY pyproject.toml poetry.lock ./
+RUN . /opt/venv/bin/activate && poetry install --no-dev --no-root
```
khvn26: I see multiple problems here:

1. `--no-dev` should be `--without dev` or `--only main`, see [docs](https://python-poetry.org/docs/cli/#install).
2. Python venv in a Docker image is redundant and not easy to manage. How do we make sure the entrypoint is run in the correct venv? I suggest you consider either using `POETRY_VIRTUALENVS_CREATE=false` or [exporting to a requirements.txt file](https://python-poetry.org/docs/cli/#export).

tushar5526:
Hey @dabeeeenster the PR is ready for review and I would need the following information as well. 

[tool.poetry]
name = "y"
version = "0.1.0"
description = ""
authors = ["Your Name <you@example.com>"]
readme = "README.md"
License

api/Dockerfile
```diff
@@ -4,20 +4,14 @@ WORKDIR /app
...
-RUN python -m venv /opt/venv
-# Make sure we use the virtualenv:
-ENV PATH="/opt/venv/bin:$PATH"
-
-COPY requirements.txt .
-
-# Make sure we are running latest pip and setuptools to avoid potential security warnings
 RUN pip install --upgrade pip
 RUN pip install --upgrade setuptools
+RUN pip install poetry
```
khvn26: Please update `api/Makefile` to acommodate for new dependency management.

I'd also prefer to have Poetry version pinned there. To me, the most sensible way to do this would be to set a `POETRY_VERSION` env var, as it transparently gets picked up by the [official Poetry installer](https://python-poetry.org/docs/#installing-with-the-official-installer) — which IMO should be used locally and likely in the Docker build too. The reason for this is the fact that the official installer places Poetry in its isolated venv automatically, which eliminates the need for us to maintain a venv for the application (see [comment](https://github.com/Flagsmith/flagsmith/pull/2214#discussion_r1196993606)).

api/pyproject.toml
```diff
@@ -35,3 +35,194 @@ skip = ['migrations','.venv','.direnv']
...
+appdirs = "1.4.4"
+asgiref = "3.5.0"
+azure-core = "1.23.1"
+azure-identity = "1.12.0"
+boto3 = "1.26.75"
+botocore = "1.29.75"
+cachetools = "4.1.1"
+certifi = "2022.12.7"
+cffi = "1.15.1"
+chargebee = "2.7.7"
```
khvn26: I'd use more loose version constraints, e.g. the [tilde requirements](https://python-poetry.org/docs/dependency-specification/#tilde-requirements) or even the caret requirements. I certainly don't see value in pinning the patch versions. @matthewelwell thoughts?

api/pyproject.toml
```diff
@@ -35,3 +35,194 @@ skip = ['migrations','.venv','.direnv']
...
+
+[tool.poetry]
+name = "y"
+version = "0.1.0"
+description = ""
+authors = ["Your Name <you@example.com>"]
+readme = "README.md"
+
+[tool.poetry.dependencies]
+python = "^3.11"
```
khvn26:
```suggestion
python = "~3.10,~3.11"
```

khvn26 (changes requested):

api/pyproject.toml
tushar5526: Hi, I have added them here. https://github.com/Flagsmith/flagsmith/pull/2214/files#diff-40f8f985c548e2b57b803d17f42833f8a4d94603ca5b5e0bf48de51872809400R158

api/pyproject.toml
tushar5526: I was earlier planning to specify dependencies in the `.in` file one by one, starting with django<4, but it seems the projects work only with 3.2.x django versions. Manually adding main dependencies in `.in` using `poetry add` led to some conflicts towards the end, So I went ahead and added all the dependencies from the `.txt` files with their exact version.

api/pyproject.toml
tushar5526: Is 3.10 officially supported, the docker build failed after adding 3.10?

api/pyproject.toml
matthewelwell: This is one of my gripes about Poetry, it is not very well suited for large projects such as Flagsmith. It is tailored towards standalone packages rather than large web projects. We can add some of these items but I would much rather that this data was not manually defined here. For example, our current list of authors is 42... 

Here are my thoughts currently, but I will discuss further with the team to determine whether poetry is the right solution here. 

```toml
name = "flagsmith-api"
version = "2.52.0"
description = "The API component for the Flagsmith application."
authors = ["Flagsmith <support@flagsmith.com>"]
readme = "README.md"
```

For the record, I love poetry for the dependency management and for simplifying publishing packages to e.g. pypi.

api/Dockerfile
tushar5526: The application stage of docker build is the same as before. I don't think this would affect the entrypoint. Please correct me if I am missing something.

api/Dockerfile
tushar5526: Resolved the comments.

api/pyproject.toml
tushar5526: I have updated the dependencies to include only main dependencies as mentioned in `.in` files

tushar5526:
Fixed the falling tests. It seems 2 GHAs need approvals before they can run.

gagantrivedi:
> Fixed the falling tests. It seems 2 GHAs need approvals before they can run.

Hi @tushar5526 those two action are deployment related - we can skip them for now

api/Makefile
```diff
@@ -7,19 +7,25 @@ COMPOSE_PROJECT_NAME ?= flagsmith
...
 install-pip:
 	python -m pip install --upgrade pip
 
+.PHONY: install-poetry
+install-poetry:
+	pip install poetry==${POETRY_VERSION}
+
 .PHONY: install-packages
 install-packages:
 	pip install -r requirements-all.txt
```
khvn26: The `install-packages` target shoud use Poetry now.

Additionally, there should be a way to specify/omit optional dependency groups so the target could be used in Docker builds. I'd suggest using an `opts` environment variable like the `test` and `django-migrations` targets do.

api/Makefile
```diff
@@ -7,19 +7,25 @@ COMPOSE_PROJECT_NAME ?= flagsmith
...
 -include .env-local
 -include $(DOTENV_OVERRIDE_FILE)
 
 .PHONY: install-pip
 install-pip:
 	python -m pip install --upgrade pip
 
+.PHONY: install-poetry
+install-poetry:
+	pip install poetry==${POETRY_VERSION}
```
khvn26: I'm not very keen about using `pip` to install Poetry as it's very ambiguous in regards to which virtual environment is used.

If the implication is that the target installs Poetry to the application's venv, this should not happen as Poetry has a lot of dependencies of its own which may introduce conflicts and unnecessarily bloat the site-packages. We can help the user to avoid this, though; [the official installer](https://python-poetry.org/docs/#installing-with-the-official-installer) sets Poetry up in a dedicated venv without any additional hoops to go through.

Dockerfile
```diff
@@ -17,22 +17,19 @@ WORKDIR /app
...
 RUN pip install --upgrade setuptools
+RUN pip install "poetry==$POETRY_VERSION"
 
-# Install our python dependencies
-RUN pip install -r requirements.txt
+# Set up venv
+RUN python -m venv /opt/venv
 
+COPY api/pyproject.toml api/poetry.lock ./
+RUN . /opt/venv/bin/activate && poetry install --only main
```
khvn26: 1. I'd love to see the Makefile target used here for Poetry installation.
2. There's no need to use a venv at all in a Docker build. `POETRY_VIRTUALENVS_CREATE` environment variable controls this, see [docs](https://python-poetry.org/docs/configuration/#virtualenvscreate).

Dockerfile
```diff
@@ -17,22 +17,19 @@ WORKDIR /app
 
 RUN apt-get update && apt-get install -y gcc build-essential libpq-dev musl-dev python3-dev
 
-# Set up venv
-RUN python -m venv /opt/venv
-# Make sure we use the virtualenv:
-ENV PATH="/opt/venv/bin:$PATH"
+ENV POETRY_VERSION=1.5.0
```
khvn26: If we use the Make target to install Poetry, we can remove this line. The version could be overriden by mounting a dotenv file. We could keep `POETRY_VERSION` as a build argument but I don't think that's necessary.

khvn26 (changes requested):

api/Dockerfile
khvn26: The venv-in-Docker thread is over [here](https://github.com/Flagsmith/flagsmith/pull/2214#discussion_r1231666787) now. The `--no-dev` is addressed [here](https://github.com/Flagsmith/flagsmith/pull/2214#discussion_r1231661483) (if we use the Make target in a Docker build, which I strongly insist on).

Dockerfile
tushar5526: Do we need multi-stage builds? Something like the following should work perfectly. We only have to cache the package installation step as suggested here - https://python-poetry.org/docs/faq/#poetry-busts-my-docker-cache-because-it-requires-me-to-copy-my-source-files-in-before-installing-3rd-party-dependencies

```
FROM python
COPY pyproject.toml poetry.lock .
RUN pip install poetry && poetry install --no-root --no-directory
COPY src/ ./src
RUN poetry install --no-dev
```

api/Makefile
tushar5526: Thanks for the suggestion, using official installer now.

Dockerfile
tushar5526: Thanks for the suggestion. Implemented in new commits.

Dockerfile
tushar5526: Yes, I have removed it from here. It is only present in Makefile now.

tushar5526:
@khvn26 Thank you for the detailed feedback.

Few things to notice in recent commits.
- `RUN apt-get update && apt-get install -y gcc build-essential libpq-dev musl-dev python3-dev` services still seems to build and run fine without this. 

- We are also copying `bin` folder from build-stage to copy executables now. Earlier `venv` used to contain all the `site-packages` and `executables` which were copied to `application`. https://github.com/Flagsmith/flagsmith/pull/2214/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R39

- https://github.com/Flagsmith/flagsmith/pull/2214/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L39, moved `COPY` further down in build process to cache `apt-get update and installs`

tushar5526:
I will update the Redhat Dockerfile, if this gets approved.

api/Dockerfile.redhat-ubi
```diff
@@ -14,8 +14,11 @@ USER root
 RUN yum -y update-minimal --security --sec-severity=Important --sec-severity=Critical
 USER 1001
 
-ADD --chown=1001:0 requirements.txt .
-RUN pip install -r requirements.txt
+ENV POETRY_VERSION=1.5.0
```
matthewelwell: Do we need this, since we've defined a default version in the Makefile?

matthewelwell (changes requested):
Some thoughts on this still: 

 * We should standardise to use requirements.txt (which should be generated by poetry). This will allow us to keep a lot of the current process and, most importantly, prevents us from breaking downstream modules (see [this workflow](https://github.com/Flagsmith/flagsmith/blob/main/.github/workflows/platform-docker-publish-all-features-image.yml) for reference). To be clear, this process of generating the requirements.txt file should be added to all workflows that rely on it, requirements.txt should not be committed to the repository. 
 * This PR should remove the existing requirements.* files from the repository. 
 * The Makefile should make use of `poetry run` for existing commands
 * We should add a makefile command for generating requirements.txt

api/Dockerfile.redhat-ubi
tushar5526: I will update this Dockerfile, after the current Dockerfile is reviewed and approved.

tushar5526:
> * We should standardise to use requirements.txt (which should be generated by poetry). This will allow us to keep a lot of the current process and, most importantly, prevents us from breaking downstream modules (see [this workflow](https://github.com/Flagsmith/flagsmith/blob/main/.github/workflows/platform-docker-publish-all-features-image.yml) for reference). To be clear, this process of generating the requirements.txt file should be added to all workflows that rely on it, requirements.txt should not be committed to the repository.

- Do we also want to generate a requirements file first and then install dependencies while building the image or is the current setup fine?
- Should I `add requirements*` to `.gitignore` to prevent accidental commits.

matthewelwell:
> > * We should standardise to use requirements.txt (which should be generated by poetry). This will allow us to keep a lot of the current process and, most importantly, prevents us from breaking downstream modules (see [this workflow](https://github.com/Flagsmith/flagsmith/blob/main/.github/workflows/platform-docker-publish-all-features-image.yml) for reference). To be clear, this process of generating the requirements.txt file should be added to all workflows that rely on it, requirements.txt should not be committed to the repository.
> 
> * Do we also want to generate a requirements file first and then install dependencies while building the image or is the current setup fine?

Yes, we should use the requirements.txt wherever possible. 

> * Should I `add requirements*` to `.gitignore` to prevent accidental commits.

No, I think it's fine to leave it out of gitignore.

matthewelwell:
@tushar5526 can you fix the unit tests and I will re-review?

tushar5526:
> @tushar5526 can you fix the unit tests and I will re-review?

Can you re-run the tests once you get a chance ?

matthewelwell:
> > @tushar5526 can you fix the unit tests and I will re-review?
> 
> Can you re-run the tests once you get a chance ?

It looks like they've already run and are still failing?

tushar5526:
> It looks like they've already run and are still failing?

I missed to add the new dependencies after rebasing on `main`. Added them, please re-run the workflows when you get a chance.

matthewelwell:
@tushar5526 looks like the actions are still failing - I think because of the migration issue you can see [here](https://github.com/Flagsmith/flagsmith/actions/runs/5435480140/jobs/9885768554?pr=2214). Probably it's caused by an unexpected change in the version of the `django-simple-history` package. This makes me slightly nervous that we're getting unexpected changes in package version elsewhere. I'd like to keep the versions the same in this PR and only migrate from pip-tools -> poetry.

api/poetry.lock
```diff
@@ -0,0 +1,3703 @@
...
+django = ">=2.2"
+pytz = ">=2016.10"
+
+[package.extras]
+bounce = ["cryptography (>=36.0.2)", "requests (>=2.27.1)"]
+events = ["cryptography (>=36.0.2)", "requests (>=2.27.1)"]
+
+[[package]]
+name = "django-simple-history"
+version = "3.3.0"
```
matthewelwell: We've accidentally(?) performed a major version upgrade here. See [here](https://github.com/Flagsmith/flagsmith/blob/main/api/requirements.txt#L110), we should be using v2.12.0, or at least v2.x.x

Ideally we wouldn't perform ANY version upgrades as part of this upgrade. Is there a way to generate a `poetry.lock` file from an existing `requirements.txt` perhaps? Perhaps we also need to define the restrictions in `pyproject.toml` better?

tushar5526: Yes, I see that now. I will re-verify that once again. How I generated the `pyproject.toml.dependencies` was to go through all the base dependencies in `requirements.in` and then do a poetry add by using the exact version as specified in `requirements.txt`. I think I missed this base dependency. I will re-verify all the dependencies in `requirements.in` and their versions, once again. Thanks

tushar5526:
> @tushar5526 looks like the actions are still failing - I think because of the migration issue you can see [here](https://github.com/Flagsmith/flagsmith/actions/runs/5435480140/jobs/9885768554?pr=2214). Probably it's caused by an unexpected change in the version of the `django-simple-history` package. This makes me slightly nervous that we're getting unexpected changes in package version elsewhere. I'd like to keep the versions the same in this PR and only migrate from pip-tools -> poetry.

So should I use the exact versions here or still use `^` or `~` for some flexibility ?

matthewelwell:
> > @tushar5526 looks like the actions are still failing - I think because of the migration issue you can see [here](https://github.com/Flagsmith/flagsmith/actions/runs/5435480140/jobs/9885768554?pr=2214). Probably it's caused by an unexpected change in the version of the `django-simple-history` package. This makes me slightly nervous that we're getting unexpected changes in package version elsewhere. I'd like to keep the versions the same in this PR and only migrate from pip-tools -> poetry.
> 
> So should I use the exact versions here or still use `^` or `~` for some flexibility ?

I guess the ideal would be to define everything as `^x.x.x` in pyproject.toml with the exact versions that we're currently using (in `requirements.txt` provided in `poetry.lock`. If that's too much effort, it would be fine (I think) to define them as `~x.x.x` and let poetry find the latest patch version.

tushar5526:
I have updated the base dependencies manually to match it with `requirements.txt` file. Is there any way to run tests locally. GHAs do not run automatically for new contributors probably.

matthewelwell:
> I have updated the base dependencies manually to match it with `requirements.txt` file. Is there any way to run tests locally. GHAs do not run automatically for new contributors probably.

Yes, you can run them via the makefile: 

```
> make test
```

matthewelwell:
@tushar5526, I'm not sure why the tests haven't been re-run here but it could be related to the conflicts. Could you rebase onto the main branch?

tushar5526:
> @tushar5526, I'm not sure why the tests haven't been re-run here but it could be related to the conflicts. Could you rebase onto the main branch?

Did that, I think GitHub restricts running actions for new contributors by default - https://github.blog/2021-04-22-github-actions-update-helping-maintainers-combat-bad-actors/

tushar5526:
Few things to notice
- Even with `^` requirements, the automatic upgrade of dependencies is failing tests. For example `pyotp` related tests fails with `^`
- Locally I still see `1` test failing, but I am skeptical there would be more undefined behavior that might pop up later. 

What we can do is use `~` initially and then move to `^` eventually?

tushar5526:
Wrote a small script to find the version differences between the `requirements.txt` exported from `pyproject.toml` and with the main branch's `requirements.txt`

| Package                 | Version in File 1 | Version in File 2 |
|-------------------------|-------------------|-------------------|
| analytics-python        | 1.4.post1         | 1.4.0             |
| asgiref                 | 3.7.2             | 3.5.0             |
| azure-core              | 1.27.1            | 1.23.1            |
| azure-identity          | 1.13.0            | 1.12.0            |
| boto3                   | 1.27.0            | 1.26.75           |
| botocore                | 1.30.0            | 1.29.75           |
| cachetools              | 5.3.1             | 4.1.1             |
| certifi                 | 2023.5.7          | 2022.12.7         |
| chargebee               | 2.28.0            | 2.7.7             |
| charset-normalizer      | 3.1.0             | 3.0.1             |
| cryptography            | 41.0.1            | 39.0.1            |
| django-axes             | 5.41.1            | 5.32.0            |
| django-cors-headers     | 3.14.0            | 3.5.0             |
| django-debug-toolbar    | 3.8.1             | 3.2.1             |
| django-health-check     | 3.17.0            | 3.14.3            |
| django-ipware           | 5.0.0             | 3.0.2             |
| django-ordered-model    | 3.7.4             | 3.4.1             |
| django-storages         | 1.13.2            | 1.10.1            |
| django                  | 3.2.20            | 3.2.19            |
| djangorestframework-api-key | 2.3.0         | 2.2.0             |
| djangorestframework     | 3.12.4            | 3.12.1            |
| drf-nested-routers      | 0.92.5            | 0.92.1            |
| drf-writable-nested     | 0.6.4             | 0.6.2             |
| drf-yasg2               | 1.19.4            | 1.19.3            |
| environs                | 9.5.0             | 9.2.0             |
| flagsmith-flag-engine   | 4.0.2             | 4.0.1             |
| google-api-core         | 2.11.1            | 1.23.0            |
| google-api-python-client| 1.12.11           | 1.12.5            |
| google-auth-httplib2    | 0.1.0             | 0.0.4             |
| google-auth             | 2.21.0            | 1.23.0            |
| googleapis-common-protos| 1.59.1            | 1.52.0            |
| httplib2                | 0.22.0            | 0.19.0            |
| influxdb-client         | 1.36.1            | 1.28.0            |
| markupsafe              | 2.1.3             | 2.1.2             |
| marshmallow             | 3.19.0            | 3.14.1            |
| msal                    | 1.22.0            | 1.21.0            |
| opencensus-context      | 0.1.3             | 0.1.2             |
| opencensus-ext-azure    | 1.1.9             | 1.1.4             |
| opencensus              | 0.11.2            | 0.9.0             |
| packaging               | 23.1              | 23.0              |
| portalocker             | 2.7.0             | 2.4.0             |
| protobuf                | 4.23.3            | 3.18.3            |
| psutil                  | 5.9.5             | 5.9.0             |
| psycopg2-binary         | 2.9.6             | 2.9.5             |
| pyasn1-modules          | 0.3.0             | 0.2.8             |
| pyasn1                  | 0.5.0             | 0.4.8             |
| pydantic                | 1.10.11           | 1.10.9            |
| pyjwt[crypto]           | 2.7.0             | 2.6.0             |
| python-dotenv           | 1.0.0             | 0.15.0            |
| pytz                    | 2023.3            | 2020.4            |
| rsa                     | 4.9               | 4.7               |
| ruamel-yaml             | 0.17.32           | 0.17.20           |
| s3transfer              | 0.6.1             | 0.6.0             |
| sentry-sdk              | 1.27.0            | 1.17.0            |
| shortuuid               | 1.0.11            | 1.0.1             |
| slack-sdk               | 3.21.3            | 3.9.0             |
| smsapi-client           | 2.7.0             | 2.4.3             |
| twilio                  | 8.4.0             | 7.9.1             |
| urllib3                 | 1.26.16           | 1.26.14           |
| urlman                  | 2.0.1             | 1.4.0             |
| wheel                   | 0.40.0            | 0.38.4            |
| whitenoise              | 6.5.0             | 6.0.0             |

tushar5526:
I tested with `~` requirements for base dependencies and tests ran fine locally. Let me know what suits you better.

tushar5526:
I have switched to `~` requirements. We can keep switching to caret requirements moving forward.

khvn26 (approved):
This looks good to me. Great job, @tushar5526 🚀

matthewelwell:
Reviewing the list [here](https://github.com/Flagsmith/flagsmith/pull/2214#issuecomment-1622445112), it seems there are a few major version upgrades still (I'm making the assumption that pyproject.toml generated versions are File 1). Of those, I think `google-auth` is the only one that's concerning as I think we're requiring that directly?

tushar5526:
> Reviewing the list [here](https://github.com/Flagsmith/flagsmith/pull/2214#issuecomment-1622445112), it seems there are a few major version upgrades still (I'm making the assumption that pyproject.toml generated versions are File 1). Of those, I think `google-auth` is the only one that's concerning as I think we're requiring that directly?

Yes your assumption is right. 

```
google-api-python-client==1.12.5
    # via -r requirements.in
google-auth==1.23.0
    # via
    #   google-api-core
    #   google-api-python-client
    #   google-auth-httplib2
```

`google-api-python-client` is constrained using `~` and google-auth is not listed as a core dependency in `requirements.in` file, hence the version bump 🤔

Here is the dependency tree for `google-api-python-client` if it helps
Screenshot 2023-08-02 at 8 32 38 PM

matthewelwell (approved):

matthewelwell:
@tushar5526 can you rebase and let's get this merged 🙌

tushar5526:
> @tushar5526 can you rebase and let's get this merged 🙌

Squash the commits as well?

matthewelwell:
> Squash the commits as well?

Up to you, we will squash and merge the PR so it's not necessary to squash commits. You can just merge the main branch in rather than rebase if that's easier.

tushar5526:
> > Squash the commits as well?
> 
> Up to you, we will squash and merge the PR so it's not necessary to squash commits. You can just merge the main branch in rather than rebase if that's easier.

Yes, you can squash them then! I have merged the main branch.

api/webhooks/webhooks.py
```diff
@@ -3,6 +3,8 @@
 import typing
 
 import requests
+import backoff
```
khvn26: Did that sneak in on accident? 🤔

khvn26 (changes requested):

api/webhooks/webhooks.py
tushar5526: Oh, my bad. I merged a wrong branch. Thanks, let me fix that quickly.

tushar5526: It should be fixed now. Please take a look.

khvn26: Looking good, thanks!

khvn26 (approved):

PR #2217, opened 2023-05-18
Handle soft deleted integrations

### LINKED ISSUE #2129: Soft deleted Integration key can not be added again
Seems like there is a bug here caused by the fact that we have added SoftDelete to the integrations model but not updated the unique constraint. So, since there are entities in the SegmentConfiguration model table still (although they've been deleted), it is not possible to add any new ones. We need to update the unique constraint to only check for entities that do not have a value for deleted_at.

matthewelwell:
## Changes

Refactors all integration models to upsert upsert if there is a soft deleted integration that already exists. 

Added unit tests.


api/integrations/slack/views.py
```diff
@@ -81,8 +81,8 @@ def slack_oauth_callback(self, request, environment_api_key):
             code, self._get_slack_callback_url(environment_api_key)
         )
 
-        SlackConfiguration.objects.update_or_create(
-            project=env.project, defaults={"api_token": bot_token}
+        SlackConfiguration.objects.all_with_deleted().update_or_create(
```
gagantrivedi: Should we create a test case that covers this?

api/integrations/common/serializers.py
```diff
@@ -0,0 +1,23 @@
+from rest_framework.serializers import ModelSerializer
+
+
+class _BaseIntegrationModelSerializer(ModelSerializer):
+    one_to_one_field_name = None
+
+    def create(self, validated_data):
+        if (related_obj := validated_data.get(self.one_to_one_field_name)) and (
```
gagantrivedi: This if condition is doing a bit too much I think?  Maybe we can add a comment? Or move this if check to a different method? Named something like `get_existing_object`? What do you think?

gagantrivedi (changes requested):

api/integrations/slack/views.py
matthewelwell: I have fought with this for a long time to write a test that makes sense but, without refactoring all this code into a service layer, there is no easy (not hacky) way to test it. In doing this though, I realised that there is currently no way to delete a SlackConfiguration object anyway so, the way I see, the pragmatic solutions are: 

 1. Keep the change I made _just in case_ it happens in the future, but accept there is no test case to cover it
 2. Just revert this change so we are at least alerted via Sentry if it does happen

api/integrations/common/serializers.py
matthewelwell: Good point - I have done a combination of the 2 things you suggested. Let me know what you think.

api/integrations/slack/views.py
gagantrivedi: I think I prefer 2 because it should not happen under normal circumstances

api/integrations/common/serializers.py
gagantrivedi: Yeah, it looks perfect

api/integrations/slack/views.py
matthewelwell: Yep, ok, I've done that.

gagantrivedi (approved):

PR #2221, opened 2023-05-18
Add group assignment model for Change Requests

### LINKED ISSUE #2204: Able to assign a change request to a user without permission to view change requests
Currently, the list of assignees when creating a change request includes all members of an organisation. This can include users that do not have permission to view change requests in a given environment. 

This causes an issue in that, when a user without permission, is assigned to a change request, they are sent an email notification but are unable to view the change request as they do not have permission, resulting in a slightly strange behaviour for that user. 

There are a couple of solutions that we can develop here to fix the immediate issue: 

 1. The list of possible assignees should only include those that have permission to approve that change request. This would be very difficult to achieve with the current permissions model. 
 2. When the change request is created, it should review the list of assignees and make sure that they each have permission to approve the change request and only assign / send notifications to those that do have permission. 

Note: an improvement was added to alleviate these issues in a PR [here](https://github.com/Flagsmith/flagsmith/pull/2221). This added the ability to assign CRs to groups vs having to scroll through a list of users manually.

matthewelwell:
## Changes

Adds the model needed to assign change requests to a group. 

Added unit test for model method. Other tests will be added to flagsmith-workflows repository.


novakzaballa (approved):
LGTM

PR #2222, opened 2023-05-18
fix/better-k8s-docs-values

dabeeeenster:
Be explicit about where the k8s config changes go.

### DIFF
--- a/docs/docs/deployment/hosting/kubernetes.md
+++ b/docs/docs/deployment/hosting/kubernetes.md
@@ -29,6 +29,9 @@ wget https://raw.githubusercontent.com/Flagsmith/flagsmith-charts/main/charts/fl
 helm install -n flagsmith --create-namespace flagsmith flagsmith/flagsmith -f values.yaml
 ```
 
+We recommend reading the [Helm docs](https://helm.sh/docs/intro/using_helm/#helm-install-installing-a-package) for
+further information.
+
 ### Ingress configuration
 
 The above is a quick and simple way of gaining access to Flagsmith, but in many cases will need to configure ingress to
--- a/docs/docs/deployment/hosting/kubernetes.md
+++ b/docs/docs/deployment/hosting/kubernetes.md
@@ -16,7 +16,8 @@ helm install -n flagsmith --create-namespace flagsmith flagsmith/flagsmith
 kubectl -n flagsmith port-forward svc/flagsmith-frontend 8080:8080
 ```
 
-Then view `http://localhost:8080` in a browser. This will install using default options, in a new namespace `flagsmith`.
+Then view `http://localhost:8080` in a browser. This will install the chart using default options, in a new namespace
+`flagsmith`.
 
 If you want to make changes to the default setup you can make a local copy of the
 [`charts/flagsmith/values.yaml`](https://github.com/Flagsmith/flagsmith-charts/blob/main/charts/flagsmith/values.yaml)file,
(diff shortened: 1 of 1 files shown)


dabeeeenster:
We need to add instructions to this PR in terms of how/where you store `values.yaml` changes? One for @khvn26 maybe?

matthewelwell (commented):
I'd argue that this is not our responsibility to document. And a lot of this isn't really correct. You shouldn't modify the values.yml in our repository, you should add flagsmith to your chart.yml and then create your own values.yml file (as far as I understand it). 

One thing I've noticed that others do (e.g. https://github.com/jenkinsci/helm-charts), which I think we should also if we don't already (I cant see that we do), is link to the helm docs.

dabeeeenster:
> I'd argue that this is not our responsibility to document. And a lot of this isn't really correct. You shouldn't modify the values.yml in our repository, you should add flagsmith to your chart.yml and then create your own values.yml file (as far as I understand it).
> 
> One thing I've noticed that others do (e.g. https://github.com/jenkinsci/helm-charts), which I think we should also if we don't already (I cant see that we do), is link to the helm docs.

Maybe we can link out to some resources on the solution to that. I think we shouldnt assume people are helm/k8s experts and if we can explain exactly how/where they modify this data that would help them

docs/docs/deployment/hosting/kubernetes.md
```diff
@@ -16,7 +16,24 @@ helm install -n flagsmith --create-namespace flagsmith flagsmith/flagsmith
 kubectl -n flagsmith port-forward svc/flagsmith-frontend 8080:8080
 ```
 
-Then view `http://localhost:8080` in a browser. This will install using default options, in a new namespace `flagsmith`.
+Then view `http://localhost:8080` in a browser. This will install the chart using default options, in a new namespace
+`flagsmith`.
+
+If you want to make changes to the default setup you can make a local copy of the
```
khvn26: They _can_, but they don't _have to_. I'll very much prefer to emphasize that the default values file should be used for reference, not for input.

I wonder if we offered a [json schema](https://www.arthurkoziel.com/validate-helm-chart-values-with-json-schemas/) we could avoid this bit of docs altogether.

dabeeeenster: Can you provide a suggestion?

dabeeeenster:
TBH I feel like we are really over-thinking this PR...

docs/docs/deployment/hosting/kubernetes.md
```diff
@@ -16,7 +16,24 @@ helm install -n flagsmith --create-namespace flagsmith flagsmith/flagsmith
...
+[`charts/flagsmith/values.yaml`](https://github.com/Flagsmith/flagsmith-charts/blob/main/charts/flagsmith/values.yaml)file,
+modify it and then install with the `-f` flag:
```
khvn26:
```suggestion
Refer to the chart's default [`values.yaml`](https://github.com/Flagsmith/flagsmith-charts/blob/main/charts/flagsmith/values.yaml) file to learn which values are expected by the chart. You can use it as a reference for building your own values file:
```

docs/docs/deployment/hosting/kubernetes.md
khvn26: See https://github.com/Flagsmith/flagsmith/pull/2222/files#r1222919381

khvn26 (approved):

dabeeeenster:
WOHOO!

PR #2223, opened 2023-05-19
improvement/dont-run-some-actions-for-docs-changes

dabeeeenster:


khvn26:
LGTM. Can we also add an [ignore script for Vercel](https://vercel.com/guides/how-do-i-use-the-ignored-build-step-field-on-vercel#with-folders-and-workspaces) to skip docs deployments when there are no changes in `docs/`?

khvn26 (approved):

Issue #430, opened 2021-10-21
Add ability to hide flags from SDKs


dabeeeenster:
## Overview

There is a common requirement to be able to send a subset of flags to an SDK client. As an example use case, if there is a single Flagsmith project that is being consumed by both a server side and client side SDK. Its probable that some of the server side SDK flags are neither relevant to the client side SDK, and in addition its probably that they would want top be private from the client side SDK. 

## Potential Approach

We could provide an API key per `Tag`. This would then filter the flags returned to the SDK to the subset of flags that are tagged with the particular tag. 

In the use case above, this would mean having two tags: `Server Side` and `Client Side`. Each SDK would then be configured with the API key specific to each Tag.

khvn26:
I came up with three approaches to this. My initial direction was to explore potential token<->tag relationships, but it quickly became more complex than I'd anticipated due to env key/server-side key disparity.
Keep in mind that all three approaches involve adding feature tags to the environment document for the filtering — this is an engine change.
As you can see I'm very much in favour of Proposal 3. What do you think?

## Proposal 1: API key per Tag

One-to-many Token<->(tag, env) relationship. Ability to add tokens to a tag in the UI. When authenticated with a tag token, get a tag by token, filter the resulting flags by tag.

Cons:

- Too many auth codepaths for my taste — env key/server-side key is enough pain to manage.
- Only one tag can be leveraged for filtering features.
- One more DynamoDB table for the Edge, which affects latency.

Pros:

- Latency is only affected for tag tokens — no filtering done when not needed.

## Proposal 2: Taggable API keys

Many-to-many Token<->(tag, env) relationship. Ability to add tags to existing tokens. When authenticated with any token, grab its tags along with the environment.
Same tradeoffs as above minus the latency one — here the latency will be affected for *all* requests.

## Proposal 3: `tags` parameter for SDK endpoints.

Skip the auth modifications altogether. Add the ability to filter output by user-provided tag or tags. Make these configurable in SDKs.

Pros:

- Much simpler implementation — no new models or relations, just the engine part and the filtering part need to be done.
- No latency cost for Edge API.
- No UI changes, only SDK API.

Cons:

- It's a bit more explicit? (Is that a con?)

khvn26:
I've hit an interesting roadblock where we're not really able to easily extend the feature schema in the document. Additionally, we found that:

- for the use case in question we only need to care about filtering for remote evaluation
[...]

Issue #2079, opened 2023-03-28
Move environment document serialization out of engine to Core API


matthewelwell:
Currently a lot of the serialization for environment documents lives in the django engine. This should be moved to the Core API to make testing / optimisations easier.

khvn26:
The refactoring steps will be as follows:

### 1. Breaking `flag_engine` release

Draft PR here: https://github.com/Flagsmith/flagsmith-engine/pull/167

- Remove `flag_engine.api` module
- Move dataclasses to Pydantic, export them to library's `__init__`/`__all__`
- Remove marshmallow schemas and document 

### 2. Edge API release

- Bump `flag_engine`
- Decouple Edge API schemas from `flag_engine`'s schemas

### 3. Core API release

- Bump `flag_engine`
- Add logic from `flag_engine.api` module to `map_environment_to_document` interface

PR #2226, opened 2023-05-19
feat: protected features

khvn26:

## Changes

This PR adds backend code for #430:

- Add `Feature.is_server_key_only` denoting whether the feature should be omitted from SDK calls made with client-side environment keys
- Enrich environment document with `project.server_key_only_feature_ids` for Edge API filtering
- FIlter out features with `is_server_key_only` set to `True` in SDK environment, identity views

Covered with unit tests.


api/environments/identities/views.py
```diff
@@ -254,12 +257,22 @@ def post(self, request):
             },
         )
 
+    def _get_additional_filters(self) -> Q | None:
+        if self.request.originated_from is RequestOrigin.CLIENT:
```
matthewelwell, line 261: We should perhaps be defensive here and use `getattr(self.request, "originated_from", None)` ?

api/environments/mappers.py
```diff
@@ -0,0 +1,15 @@
+from typing import TYPE_CHECKING, Any
+
+from flag_engine.api.document_builders import build_environment_document
+
+if TYPE_CHECKING:
+    from environments.models import Environment
+
+
+def map_environment_to_document(environment: "Environment") -> dict[str, Any]:
```
matthewelwell, line 9: I think it would be good to include some documentation here to explain why we're doing this here and not in a schema?

matthewelwell (commented):
Added a couple of minor comments.

api/environments/identities/views.py
khvn26: I'd actually prefer this part broken if we switch/modify the authentication class for some reason. Unit tests cover this so we should be safe.

api/environments/mappers.py
khvn26: Yeah, guess I'm being a bit sneaky with this :)

Ultimately this interface will become part of https://github.com/Flagsmith/flagsmith/issues/2079. Is it okay to reference the issue in the function docstring or a comment?

khvn26: Done.

api/environments/identities/views.py
matthewelwell: Yep, ok, sounds good 👍

matthewelwell (approved):

matthewelwell (approved):

PR #2227, opened 2023-05-19
Allow groups as change request approvers

novakzaballa:

## Changes

Now users can assign groups as reviewers/approvers for change requests.

Manually,

Create an environment and a feature, create some users and groups.

- Configure the enviroment to use change requests.
- change the feature and assign some users and groups as reviewers.
- Edit the change request.

For this feature a couple of flags have been created:

- `enable_groups_as_reviewers` : If this feature flag is set to `true`, this flag will show the Assigned groups section in the Change Requests and in the Create Change Request modal in the Features page.
- `disable_users_as_reviewers`: This flag has been named and used in its negative form for backward compatibility. If it is set to `false` and `enable_groups_as_reviewers` is also false, the app will behave as before these changes. If the flag is set to true,  the Assigned users section in the Change Requests and in the Create Change Request modal in the Features page will be hidden.


frontend/web/components/groupSelect.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const [filter, setFilter] = useState<string>('')
+  const grouplist =
+    groups &&
+    groups.filter((v) => {
+      const search = filter.toLowerCase()
+      if (!search) return true
+      return `${v.name}`.toLowerCase().includes(search)
+    })
+  return (
+    <InlineModal
```
kyle-ssg: Upper case filename GroupSelect.tsx

matthewelwell:
Seems like the feature flags `groups_as_reviewers` is not working as expected - can still see groups option when flag is enabled. Seems like it just prevents the 'Add groups' link being clicked, vs hiding the option entirely.

frontend/web/components/groupSelect.tsx
novakzaballa: Done

matthewelwell:
image

Slight UI issue here - don't need to add `Groups: ` before the list of assigned groups.

dabeeeenster:
I cant get these cogs to work? 
image

novakzaballa:
> I cant get these cogs to work?
> image

@dabeeeenster these cogs have been replaced by the `Add user` and `Add group` buttons to provide a more consistent UX with the Create Change Request modal.

matthewelwell (dismissed):
This looks good to me now!

matthewelwell:
@novakzaballa as mentioned in Slack, adding the `users_as_reviewers` flag will mean that all existing self hosted customers will have to enable the flag to maintain the current functionality. This will be a bit of a nightmare to manage. I appreciate there was reasoning behind adding that flag but I think that after all, we should not add the flag and leave the users as reviewers as functionality that is always available.

An alternative would be to change the flag name to `disable_users_as_reviewers` and invert all of the conditional checks. This is probably easier?

matthewelwell (approved):

PR #2232, opened 2023-05-23
feat: Add roles

gagantrivedi:

## Changes

Create a role that can hold environment/project/organisation permission and can be attached to user or groups

New API Endpoints: 

`​/api​/v1​/organisations​/{organisation_pk}​/roles​/`
`​/api​/v1​/organisations​/{organisation_pk}​/roles​/{role_pk}​/environments-permissions​/`
`​/api​/v1​/organisations​/{organisation_pk}​/roles​/{role_pk}​/groups​/`
`​/api​/v1​/organisations​/{organisation_pk}​/roles​/{role_pk}​/projects-permissions​/`
`​/api​/v1​/organisations​/{organisation_pk}​/roles​/{role_pk}​/users​/`

Adds unit test cases


gagantrivedi:
depends on https://github.com/Flagsmith/flagsmith-rbac/pull/1

api/environments/views.py
```diff
@@ -185,46 +185,12 @@ def user_permissions(self, request, *args, **kwargs):
                 },
             )
 
-        # TODO: tidy this mess up
```
matthewelwell, line 188: Nice to finally get rid of this TODO!

api/organisations/permissions/migrations/0002_auto_20230410_0940.py
```diff
@@ -0,0 +1,25 @@
+# Generated by Django 3.2.18 on 2023-04-10 09:40
```
matthewelwell: Can we give this migration a name rather than just the auto generated name from django?

api/permissions/permission_service.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+def user_has_organisation_permission(
+    user: "FFAdminUser", organisation: Organisation, permission_key: str
+) -> bool:
+    if is_user_organisation_admin(user, organisation):
+        return True
+
+    base_filter = get_base_permission_filter(
+        user, Organisation, permission_key, allow_admin=False
```
matthewelwell, line 95: It's not immediately obvious why `allow_admin` is false here - I guess that's something to do with the nuance that the organisation admin permission is handled slightly different?

api/permissions/permission_service.py
```diff
@@ -0,0 +1,150 @@
...
+
+    return Organisation.objects.filter(filter_).exists()
+
+
+def _is_user_object_admin(
+    user: "FFAdminUser", object_: Union[Project, Environment]
+) -> bool:
+    base_filter = get_base_permission_filter(user, type(object_))
+    filter_ = base_filter & Q(id=object_.id)
+    return type(object_).objects.filter(filter_).exists()
```
matthewelwell: Bit nitpicky here but we're calling `type(object_)` twice in this function - we could extract the class to a variable.

api/permissions/permissions_calculator.py
```diff
@@ -0,0 +1,253 @@
...
+
+if settings.IS_RBAC_INSTALLED:
+    from rbac.permissions_calculator import (
+        RolePermissionData,
+        get_roles_permission_data_for_environment,
+        get_roles_permission_data_for_organisation,
+        get_roles_permission_data_for_project,
+    )
+else:
+    RolePermissionData = []
```
matthewelwell: I wonder if we'd benefit here from using a base class that we override in the RBAC module? Then (I think) this logic would be more like: 

```python
role_permission_data_manager = BaseRolePermissionDataManager()
if settings.IS_RBAC_INSTALLED:
    role_permission_data_manager = RolePermissionDataManager()
```

Alternatively, we could just move this stub functions to another module. I just think I don't love having the stub functions defined here.

api/permissions/permissions_calculator.py
```diff
@@ -0,0 +1,253 @@
...
+        roles=get_roles_permission_data_for_environment(environment_id, user_id),
+    )
+
+
+def get_user_permission_data(
+    user_permission_qs: UserPermissionQs,
+) -> UserPermissionData:
+    user_permission_data = UserPermissionData()
+
+    for user_permission in user_permission_qs.prefetch_related("permissions"):
```
matthewelwell: I'm sure I'm missing something here but I'm not sure under which circumstances we'll be dealing with more than one user_permission here? Wouldn't this always be evaluated in the context of a project / environment / organisation? In which case, I think that there is a unique constraint on the user & project / environment / organisation combination, right?

api/tests/unit/features/test_unit_features_views.py
```diff
@@ -55,7 +55,7 @@ def test_list_feature_states_from_simple_view_set(
     )
 
     # When
-    with django_assert_num_queries(7):
+    with django_assert_num_queries(8):
```
matthewelwell, line 58: What is the extra query that we've added here?

api/tests/unit/permissions/permission_service/test_is_user_environment_admin.py
```diff
@@ -0,0 +1,77 @@
...
+):
+    assert (
+        is_user_environment_admin(
+            admin_user, organisation_two_project_one_environment_one
+        )
+        is False
+    )
+
+
+def test_is_user_environment_admin_returns_false_for_user_with_admin_permission_of_other_environment(
```
matthewelwell, line 58: I think this test is unfinished?

api/tests/unit/permissions/test_unit_permissions_calculator.py
```diff
@@ -0,0 +1,313 @@
...
+        ),
+        (
+            set(),
+            False,
+            {VIEW_PROJECT},
+            False,
+            {VIEW_PROJECT},
+            False,
+        ),
+        (
```
matthewelwell: I think this is a duplicate of the first test case?

api/tests/unit/permissions/test_unit_permissions_calculator.py
```diff
@@ -0,0 +1,313 @@
...
+            set(),
+            False,
+            {VIEW_ENVIRONMENT},
+            False,
+            {VIEW_ENVIRONMENT},
+            False,
+        ),
+        (set(), False, set(), False, set(), False),
```
matthewelwell: I think this is a duplicate fo the first test case?

api/tests/unit/permissions/test_unit_permissions_calculator.py
```diff
@@ -0,0 +1,313 @@
...
+            False,
+        ),
+        (
+            set(),
+            False,
+            {CREATE_PROJECT},
+            {CREATE_PROJECT},
+            False,
+        ),
+        (
```
matthewelwell: I think this is a duplicate of the first test case?

api/tests/unit/permissions/test_unit_permissions_calculator.py
```diff
@@ -0,0 +1,313 @@
...
+    user_permission_data = get_environment_permission_data(
+        environment.id, user_id=user.id
+    )
+
+    # Then
+    assert user_permission_data.admin == expected_admin
+    assert user_permission_data.permissions == expected_permissions
+
+
+@pytest.mark.parametrize(
```
matthewelwell, line 206: I don't think we have a test case here to cover when a user has permissions directly and from a group?

api/permissions/permissions_calculator.py
gagantrivedi: Ah, yes - that's a good point, but we don't have the unique constraint

api/tests/unit/features/test_unit_features_views.py
gagantrivedi: It's because of the query here: `https://github.com/Flagsmith/flagsmith/pull/2232/files#diff-7904b170017aa6b6a88dae1dc0905d9308c5b7b9475020895a44427433d7a60cR77` earlier it was using this function here: https://github.com/Flagsmith/flagsmith/pull/2232/files#diff-c2ac9247f7c75c1415fd7b42bc9ea9b8e62b83abfedc455585a807fe14f5c09bL303

matthewelwell: Hmm... but that logic already existed here, right? 

https://github.com/Flagsmith/flagsmith/pull/2232/files#diff-c2ac9247f7c75c1415fd7b42bc9ea9b8e62b83abfedc455585a807fe14f5c09bL340-L341

gagantrivedi: Correct, but it was not using that function - it was using this function here: https://github.com/Flagsmith/flagsmith/pull/2232/files#diff-c2ac9247f7c75c1415fd7b42bc9ea9b8e62b83abfedc455585a807fe14f5c09bL297

api/permissions/permissions_calculator.py
matthewelwell: So, here is the data on the size of those tables and the number of duplicate combinations that break the unique constraint we want to add: 

|table|# records|# duplicate combinations|
|------|--|--|
|environments_userenvironmentpermission|~16500|2|
|projects_userprojectpermission|~7500|1|
|organisation_permissions_userorganisationpermission|~250|0|
|environments_userpermissiongroupenvironmentpermission|~500|0|
|projects_userpermissiongroupprojectpermission|~250|1|
|organisation_permissions_userpermissiongrouporganisationper792d|~50|0|

api/tests/unit/permissions/test_unit_permissions_calculator.py
gagantrivedi: Done

api/tests/unit/permissions/permission_service/test_is_user_environment_admin.py
gagantrivedi: Done

api/permissions/permission_service.py
gagantrivedi: Yeah, that's true. e.g: `UserOrganisationPermission` and `UserPermissionGroupOrganisationPermission` don't have admin attributes but environment and project does

api/environments/permissions/migrations/0006_merge_duplicate_permissions.py
```diff
@@ -0,0 +1,31 @@
+# Generated by Django 3.2.18 on 2023-06-26 05:05
+
+from django.db import migrations
+
+from util.migrations import merge_duplicate_permissions
+
+
+def merge_duplicate_project_permissions(apps, schema_editor):
```
matthewelwell:
```suggestion
def merge_duplicate_environment_permissions(apps, schema_editor):
```

api/organisations/permissions/models.py
```diff
@@ -8,14 +8,38 @@
 
 
 class UserOrganisationPermission(AbstractBasePermissionModel):
-    organisation = models.ForeignKey(Organisation, on_delete=models.CASCADE)
+    organisation = models.ForeignKey(
+        Organisation,
+        on_delete=models.CASCADE,
+        related_query_name="userpermission",
```
matthewelwell, line 14: Why `related_query_name` and not just `related_name` ?

api/permissions/permission_service.py
matthewelwell: Just to be explicit - should we add a comment here?

api/permissions/permissions_calculator.py
matthewelwell: I see that you've just extracted it to a wrapper module for now - I'd say this is ok for now and we can probably improve it later.

api/organisations/permissions/migrations/0003_merge_duplicate_permissions.py
```diff
@@ -0,0 +1,31 @@
+# Generated by Django 3.2.18 on 2023-06-26 12:15
+
+from django.db import migrations
+
+from util.migrations import merge_duplicate_permissions
+
+
+def merge_duplicate_project_permissions(apps, schema_editor):
```
matthewelwell:
```suggestion
def merge_duplicate_organisation_permissions(apps, schema_editor):
```

api/organisations/permissions/models.py
gagantrivedi: It's to allow filter on the field in the query set. It is used here: https://github.com/Flagsmith/flagsmith/pull/2232/files#diff-7904b170017aa6b6a88dae1dc0905d9308c5b7b9475020895a44427433d7a60cR127

matthewelwell: Yep, I'm just asking because I'm not sure why we'd use `related_query_name` explicitly instead of just `related_name`. I guess right now we have no need to do `organisation.userpermission` but it would be consistent with the rest of our ForeignKey implementations. 

Also, should it be `userpermissions` rather than `userpermission`?

api/permissions/permission_service.py
gagantrivedi: Added here: https://github.com/Flagsmith/flagsmith/pull/2232/files#diff-7904b170017aa6b6a88dae1dc0905d9308c5b7b9475020895a44427433d7a60cR91

api/organisations/permissions/models.py
gagantrivedi: I think they kind of have different use cases, e.g: for permissions we use related_query_name because we need to filter on those fields, e.g: https://github.com/Flagsmith/flagsmith/blob/main/api/projects/models.py#L149 

> Also, should it be userpermissions rather than userpermission?

No, they are supposed to be singular, right? Because in the query they point to one row.
e.g: https://docs.djangoproject.com/en/4.2/ref/models/fields/#django.db.models.ForeignKey.related_query_name

matthewelwell (approved):
Approved - we should come up with a release plan here though before merging.

PR #2239, opened 2023-05-23
Add button to sales dashboard to update Chargebee cache

matthewelwell:
## Changes

Add button to Sales Dashboard to update Chargebee cache

Added a unit test and ran the application manually to verify the button calls the endpoint correctly.


api/organisations/chargebee/tasks.py
```diff
@@ -0,0 +1,8 @@
+from organisations.chargebee.cache import ChargebeeCache
+from task_processor.decorators import register_task_handler
+
+
+@register_task_handler()
+def update_chargebee_cache():
+    chargebee_cache = ChargebeeCache()
+    chargebee_cache.refresh()
```
khvn26, line 8: Two questions:

1. How is the cache not being updated in the first place? Some webhook getting skipped by Chargebee? Any errors on our side?
2. Can/should this also be a cron job?

khvn26 (approved):

api/organisations/chargebee/tasks.py
matthewelwell: It sets the cache with a 12h expiry then automatically refreshes it the next time data from the cache is requested. This was a quick and dirty solution to allow us to clear it if needs be when a new plan / add-on is added to chargebee, without having to handle a new webhook (which is obviously the correct solution).

khvn26: Understood. Want to create the webhook issue while you're at it?

matthewelwell: Done https://github.com/Flagsmith/flagsmith/issues/2246

PR #2253, opened 2023-05-30
fix(sensitive-data): use Null/[] as the value

gagantrivedi:

## Changes

Since some of the clients use strict schema and will break if the key is not part of the response - this change adds the sensitive fields back to the response but makes them Null/[]

### DIFF
--- a/api/environments/sdk/serializers_mixins.py
+++ b/api/environments/sdk/serializers_mixins.py
@@ -4,5 +4,5 @@ class HideSensitiveFieldsSerializerMixin:
         environment = self.context["request"].environment
         if environment.hide_sensitive_data:
             for field in self.sensitive_fields:
-                data.pop(field)
+                data[field] = [] if isinstance(data[field], list) else None
         return data
--- a/api/features/tests/test_views.py
+++ b/api/features/tests/test_views.py
@@ -656,13 +656,13 @@ def test_get_flags_hide_sensitive_data(api_client, environment, feature):
     # When
     api_client.credentials(HTTP_X_ENVIRONMENT_KEY=environment.api_key)
     response = api_client.get(url)
-    fs_sensitive_fields = [
+    feature_sensitive_fields = [
         "created_date",
         "description",
         "initial_value",
         "default_enabled",
     ]
-    feature_sensitive_fields = ["id", "environment", "identity", "feature_segment"]
+    fs_sensitive_fields = ["id", "environment", "identity", "feature_segment"]
 
     # Then
     assert response.status_code == status.HTTP_200_OK
--- a/api/features/tests/test_views.py
+++ b/api/features/tests/test_views.py
@@ -25,6 +25,10 @@ from features.models import (
     FeatureState,
     FeatureStateValue,
 )
+from features.serializers import (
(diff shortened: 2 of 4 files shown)


api/environments/sdk/serializers_mixins.py
```diff
@@ -4,5 +4,5 @@ def to_representation(self, instance):
         environment = self.context["request"].environment
         if environment.hide_sensitive_data:
             for field in self.sensitive_fields:
-                data.pop(field)
+                data[field] = [] if isinstance(data[field], list) else None
```
matthewelwell, line 7: Is `None` an acceptable value for the feature description field? It should ideally be an empty string, right?

gagantrivedi: No, it is actually None
```
    description = models.TextField(null=True, blank=True)

```

api/features/tests/test_views.py
```diff
@@ -656,15 +660,18 @@ def test_get_flags_hide_sensitive_data(api_client, environment, feature):
     # When
     api_client.credentials(HTTP_X_ENVIRONMENT_KEY=environment.api_key)
     response = api_client.get(url)
+    fs_sensitive_fields = SDKFeatureStateSerializer.sensitive_fields
+    feature_sensitive_fields = SDKFeatureSerializer.sensitive_fields
```
khvn26: I'd prefer those fields explicitly tested here. E.g. if we change `SDKFeatureStateSerializer.sensitive_fields` to `()` the feature will be broken but the test won't.

api/pyproject.toml
```diff
@@ -34,4 +34,4 @@ known_third_party=['_pytest','apiclient','app_analytics','axes','chargebee','cor
 skip = ['migrations','.venv','.direnv']
 
 [tool.pytest.ini_options]
-addopts = ['--ds=app.settings.test', '-vvvv', '-p', 'no:warnings', '-n', 'auto']
+addopts = ['--ds=app.settings.test', '-vvvv', '-p', 'no:warnings' ]
```
matthewelwell: Heh, I think this was a mistake? (but one that I've probably made too because Pycharm debugging doesn't seem to work with multiple processes).

matthewelwell (changes requested):

api/pyproject.toml
khvn26: We can move `xdist` options over to the `Makefile`. We'll have to do so for coverage too as `pytest-cov` disrupts the debugger, might as well move `xdist` there too.

matthewelwell: @khvn26 do you want to create a separate PR for this?

gagantrivedi: > 



> Heh, I think this was a mistake? (but one that I've probably made too because Pycharm debugging doesn't seem to work with multiple processes).

Yep, it was a mistake - but yeah we should do it

matthewelwell (approved):

matthewelwell:
I guess we need to do this in the Edge API repo too @gagantrivedi ?

gagantrivedi:
Yeah, it's merged in edge

PR #2264, opened 2023-06-02
Show pricing plan and limits

### LINKED ISSUE #631: Show users their pricing plan & limits
We should show our users that they are API call limited. Get the total API calls made in the last 30 days for the organisation (as per the Org usage graph) and show it in the top banner if the calls are within 30% of the limit, so:

"45k out of 50k API calls" with a CTA to upgrade

pls it should be different from other buttons

image

novakzaballa:

## Changes

- Updated the "upgrade" button to show Api calls in 30 days/max api  calls in the subscription, when only 30% left

_Please describe._

### DIFF
--- a/frontend/common/utils/format.js
+++ b/frontend/common/utils/format.js
@@ -112,6 +112,17 @@ const Format = {
     return value
   },
 
+  formatNumber(number) {
+    const suffixes = ['', 'K', 'M', 'B', 'T']
+    const numDigits = Math.floor(Math.log10(number)) + 1
+    const suffixIndex = Math.floor((numDigits - 1) / 3)
+
+    let shortValue = number / Math.pow(1000, suffixIndex)
+    shortValue = Math.floor(shortValue)
+
+    return shortValue + suffixes[suffixIndex]
+  },
+
   moment(value, format) {
     // DATE, hh:mm > 23:00
     if (value) {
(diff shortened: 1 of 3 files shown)


frontend/common/utils/format.js
```diff
@@ -112,6 +112,17 @@ const Format = {
     return value
   },
 
+  formatNumber(number) {
```
gagantrivedi: What does this do? Can we add a comment or something?

gagantrivedi (dismissed):

frontend/common/utils/format.js
novakzaballa: The function was renamed, and I added a comment. Also, I changed the behavior a bit.

kyle-ssg:
@dabeeeenster @matthewelwell  a couple of product questions based on this PR

- The PR adjusts the upgrade nav to show regardless of their plan, I'm not sure we'd want to do this? Especially with auto seats this seems more relevant to startup / free?
- The PR adds a usage message based on <30% remaining, I guess maybe this is down the line but I'd have thought that the velocity of usage is more important than the amount remaining. E.g. 50% usage in the first 2 months

dabeeeenster:
@kyle-ssg 

point 1: I think we would want to show if people are near their limit regardless of their plan. 

point 2: Given that the data is from the last 30 days, I think current data is fine.

kyle-ssg (approved):

PR #2279, opened 2023-06-08
feat: Use get-metadata-subscription to get max_api_calls

### LINKED ISSUE #2268: Ensure that the data about max API calls and actual API calls is on-sync and up to date

There are discrepancies in the max API calls number between the sales dashboard and the Django admin views for some subscriptions. Also, we need to ensure that the number of API calls is kept up to date for every organisation.

1. Go to the sales dashboard and pick an organisation with a start-up plan. See that the max number of API calls is 1M.
2. Go to the Django admin page and search for the same organisation. See that the max number of API calls is 50K.

The organisation subscription information must be kept on-sync. The same number of max API calls should be shown in the sales dashboard and the Django admin pages.

novakzaballa:

## Changes

Get `value of max_api_calls` from `get-metadata-subscription` endpoint. 

In order to test locally
- enable `max_api_calls_alert` for your identity
- change `FLAGSMITH_API_URL` to production
- Locally, you can hardcode values of `max_api_calls` or `total_api_calls` returned by the cooresponding endpoint. You can also change the hardcoded condition in `common/utils/utiils.tx` L58, that currently triggers the alert on 70% of usage and above in order to force the display of the alert message.


frontend/web/components/App.js
```diff
@@ -254,6 +258,17 @@ const App = class extends Component {
         this.setState({ totalApiCalls: res[0]?.data?.totals.total })
       })
     }
+    const subscriptionModel = OrganisationStore.getSubscriptionMeta()
```
matthewelwell: One thing that I'm concerned about here is that for each request the client makes to `get-subscription-metadata`, we are making a call to Chargebee's API. What that means is that: 

 a. This endpoint is not the most performant endpoint we have (as we're reliant on the performance of an external API). We should therefore ensure that we're not blocking anything waiting on the response from this endpoint. Maybe that's already happening here, but I'd like to confirm that. 
 b. For each request to this endpoint on our side, we trigger a request to Chargebee's API which has rate limits, etc. so we should be careful about abusing it. 

Here are the CB API limits. We are on the 'Rise' plan, as per the highlighted section. 

Screenshot 2023-06-19 at 15 07 08

Based on the above, is there anything we can do to optimise the requests from the frontend? Or, perhaps we should optimise this on the BE? Store some kind of additional cache of a customer's subscription information after a successful request from CB? 

I'm guessing that by adding this here, we're triggering the request on every successful load of the Flagsmith dashboard. I don't know where our current traffic is right now, but I don't imagine this number is beyond the realms of possibility. 

cc @kyle-ssg @dabeeeenster

dabeeeenster: Can we just store this with a 24 hour TTL in our DB? I thought we were already storing this data in our database? Isn't that what `organisations_organisationsubscriptioninformationcache` is for?

matthewelwell: > Can we just store this with a 24 hour TTL in our DB?

Yes, I think that's what I would suggest. 

>I thought we were already storing this data in our database? Isn't that what organisations_organisationsubscriptioninformationcache is for?

Currently this table is only used for the sales dashboard. We could, however, expose this to the frontend for sure. We should make sure we implement the recurring task before we do that though.

novakzaballa: @matthewelwell do we have in place a job to execute the update of this data? That is what you are talking about when you say "recurring task" right?

matthewelwell: Yep, you can see the functionality [here](https://github.com/Flagsmith/flagsmith/blob/main/api/task_processor/models.py#L126).

matthewelwell: TODO: 

 * Use CB webhook to keep OrganisationSubscriptionInformationCache up to date with max_seats and max_api_calls
 * Update the get-subscription-metadata endpoint to use the OrganisationSubscriptionInformationCache object 
 * Update the current task which updates the OrganisationSubscriptionInformationCache models to ONLY update influx data

To clarify, we should continue using live influx data in the frontend dashboard.

novakzaballa: Done

api/organisations/views.py
```diff
@@ -278,11 +283,37 @@ def chargebee_webhook(request):
             )
             logger.error(error_message)
             return Response(status=status.HTTP_200_OK)
+        existing_organisation_subscription_information_cache = (
+            OrganisationSubscriptionInformationCache.objects.filter(
+                organisation_id=existing_subscription.id
```
matthewelwell:
```suggestion
                organisation_id=existing_subscription.organisation_id
```

matthewelwell: This could have been really bad! We'd have ended up updating the data for the wrong organisations because we were using the id of the subscription.

api/organisations/views.py
```diff
@@ -278,11 +283,37 @@ def chargebee_webhook(request):
...
+            ).exists()
+        )
 
         subscription_status = subscription_data.get("status")
         if subscription_status == "active":
             if subscription_data.get("plan_id") != existing_subscription.plan:
                 existing_subscription.update_plan(subscription_data.get("plan_id"))
+            if existing_organisation_subscription_information_cache:
```
matthewelwell: This (including the above) could all be simplified by using `OrganisationSubscriptionInformationCache.objects.update_or_create()` I believe?

api/organisations/subscriptions/metadata.py
```diff
@@ -42,7 +42,9 @@ def __str__(self):
         return (
             "%s Subscription Metadata (seats: %d, api_calls: %d, projects: %s, chargebee_email: %s)"
             % (
-                self.payment_source.title(),
+                self.payment_source.title()
+                if self.payment_source is not None
+                else "no title",
```
matthewelwell: It's not a title don't forget, the `.title()` is converting the string to title case, not getting a title attribute!

We should use something like `"unknown payment source"`

api/organisations/subscription_info_cache.py
```diff
@@ -13,7 +13,7 @@
 ]
 
 
-def update_caches():
+def update_caches(update_influx=False):
```
matthewelwell: typehinting please! 

Also, this is a bit messy as we're implicitly updating chargebee if update influx is false but _only_ updating influx if it's true. We should either just create 2 completely separate functions, or have one function that takes an attribute of `update_from_chargebee: bool = False` and then _always_ update influx, but only update chargebee if the attribute is set.

matthewelwell: I think my preference is for having 2 separate functions and just get rid of this one entirely, unless there's a reason to keep it.

api/organisations/views.py
```diff
@@ -278,11 +283,37 @@ def chargebee_webhook(request):
...
+                metadata = get_subscription_metadata(
```
matthewelwell: Do we need to do this? Isn't the data already available in the chargebee webhook payload? Maybe we still need to check the plan / addons against the chargebee cache, but I think that, by calling this function, we're making another outbound request to chargebee unnecessarily, right?

matthewelwell (changes requested):
Some general feedback is that we should try and extract logic to well named functions / methods here too. We're cramming a lot into the chargebee webhook view for example. I know the code for that view is already a bit messy, but we shouldn't add to the mess, we should try to improve it!

api/organisations/views.py
novakzaballa: Yes, horrible. Corrected

api/organisations/views.py
novakzaballa: Right. Done

api/organisations/subscriptions/metadata.py
novakzaballa: Done

api/organisations/subscription_info_cache.py
novakzaballa: The code for those 2 functions would be almost exactly the same. I kept one function but clarified the parameter name and type.

api/organisations/views.py
novakzaballa: Improved as per our conversation in this regard.

api/organisations/chargebee/chargebee.py
```diff
@@ -128,6 +128,22 @@ def get_subscription_metadata(
         return subscription_metadata
 
 
+def get_max_seats_and_max_api_calls(
```
matthewelwell: We should be able to refactor `get_subscription_metadata` to use this as well, right? So we're not duplicating code.

api/organisations/models.py
```diff
@@ -203,14 +203,17 @@ def get_portal_url(self, redirect_url):
...
         metadata = None
-
         if self.subscription_id == TRIAL_SUBSCRIPTION_ID:
             metadata = BaseSubscriptionMetadata(
                 seats=self.max_seats, api_calls=self.max_api_calls
             )
 
         if self.payment_method == CHARGEBEE and self.subscription_id:
-            metadata = get_subscription_metadata(self.subscription_id)
+            metadata = ChargebeeObjMetadata(
```
matthewelwell: We should probably add a comment here as to why we're now just grabbing the data from the subscription information cache directly.

api/organisations/chargebee/chargebee.py
```diff
@@ -128,6 +128,22 @@ def get_subscription_metadata(
...
+    suscription_data,
+) -> typing.Optional[ChargebeeObjMetadata]:
```
matthewelwell: This is the wrong typehint

api/organisations/chargebee/chargebee.py
```diff
@@ -128,6 +128,22 @@ def get_subscription_metadata(
         return subscription_metadata
 
 
+def get_max_seats_and_max_api_calls(
+    suscription_data,
```
matthewelwell:
```suggestion
    subscription_data,
```

matthewelwell: This also needs a typehint

api/organisations/chargebee/chargebee.py
matthewelwell: We also discussed previously that functions / methods should never really have `and` in their name, it implies they're doing 2 separate tasks. Is there a way that we can change the behaviour of this function to remove the 'and'? 

Shouldn't we just change the 2 functions to be: `get_subscription_metadata_from_id` and `get_subscription_metadata_from_subscription`? Something like that? 

One possible improvement here might be to create `classmethod`s on the `ChargebeeObjMetadata` class, right?

api/organisations/chargebee/chargebee.py
```diff
@@ -128,6 +128,22 @@ def get_subscription_metadata(
...
+) -> typing.Optional[ChargebeeObjMetadata]:
+    if not (suscription_data.get("id") and suscription_data.get("id").strip() != ""):
+        logger.warning("Subscription id is empty or None")
+        return None
+
+    with suppress(ChargebeeAPIError):
+        plan_metadata = get_plan_meta_data(suscription_data.get("plan_id"))
+        subscription_metadata = plan_metadata
```
matthewelwell: I'm not sure why we're doing a reassignment of the variable here?

api/organisations/subscription_info_cache.py
```diff
@@ -6,16 +6,16 @@
...
-from .subscriptions.constants import CHARGEBEE
+from .subscriptions.constants import CHARGEBEE, SubscriptionCacheEntity
 
 OrganisationSubscriptionInformationCacheDict = typing.Dict[
     int, OrganisationSubscriptionInformationCache
 ]
 
 
-def update_caches():
+def update_caches(update_cache_entity: SubscriptionCacheEntity):
```
matthewelwell: Let's just use INFLUX and CHARGEBEE, not INFLUX_AND_CHARGEBEE and pass a tuple here.

api/organisations/subscription_info_cache.py
```diff
@@ -30,8 +30,13 @@ def update_caches():
         for org in organisations
     }
 
-    _update_caches_with_influx_data(organisation_info_cache_dict)
-    _update_caches_with_chargebee_data(organisations, organisation_info_cache_dict)
+    if update_cache_entity.value == "INFLUX":
```
matthewelwell:
```suggestion
    if update_cache_entity == SubscriptionCacheEntity.INFLUX:
```

matthewelwell (changes requested):

api/organisations/models.py
novakzaballa: Done.

api/organisations/chargebee/chargebee.py
novakzaballa: Corrected.

api/organisations/subscription_info_cache.py
novakzaballa: Done.

api/organisations/subscription_info_cache.py
novakzaballa: Corrected.

api/organisations/chargebee/chargebee.py
novakzaballa: Deleted.

api/organisations/chargebee/chargebee.py
novakzaballa: Corrected. A new function `get_subscription_metadata_from_id` has been added, and `get_subscription_metadata` has been refactored.

api/organisations/chargebee/chargebee.py
novakzaballa: Corrected.

api/organisations/chargebee/chargebee.py
```diff
@@ -104,6 +106,28 @@ def get_hosted_page_url_for_subscription_upgrade(
 
 
 def get_subscription_metadata(
+    chargebee_subscription_dict: typing.Union[dict, chargebee.Subscription],
+    customer_email: str,
+):
```
matthewelwell: Let's add a typehint to the return

api/organisations/chargebee/chargebee.py
```diff
@@ -104,6 +106,28 @@ def get_hosted_page_url_for_subscription_upgrade(
 
 
 def get_subscription_metadata(
```
matthewelwell: I wonder if `extract_subscription_metadata` would be a better name here?

api/organisations/chargebee/chargebee.py
```diff
@@ -104,6 +106,28 @@ def get_hosted_page_url_for_subscription_upgrade(
...
+    chargebee_subscription_dict: typing.Union[dict, chargebee.Subscription],
```
matthewelwell: `chargebee_subscription_dict` is not a good argument name if it can be either a `dict` or `Subscription`.

api/organisations/subscription_info_cache.py
```diff
@@ -4,18 +4,18 @@
...
-from .subscriptions.constants import CHARGEBEE
+from .subscriptions.constants import CHARGEBEE, SubscriptionCacheEntity
 
 OrganisationSubscriptionInformationCacheDict = typing.Dict[
     int, OrganisationSubscriptionInformationCache
 ]
 
 
-def update_caches():
+def update_caches(update_cache_entity: typing.Tuple[SubscriptionCacheEntity, ...]):
```
matthewelwell:
```suggestion
def update_caches(update_cache_entities: typing.Tuple[SubscriptionCacheEntity, ...]):
```

api/organisations/subscriptions/subscription_service.py
```diff
@@ -1,5 +1,5 @@
 from organisations.chargebee import (
-    get_subscription_metadata as get_chargebee_subscription_metadata,
+    get_subscription_metadata_from_id as get_chargebee_subscription_metadata,
```
matthewelwell: Why use a named import here rather than just updating the function calls?

api/organisations/tasks.py
```diff
@@ -30,5 +32,12 @@ def send_org_over_limit_alert(organisation_id):
 
 
 @register_task_handler()
-def update_organisation_subscription_information_caches():
-    subscription_info_cache.update_caches()
+def update_organisation_subscription_information_influx_caches():
+    influx_cache = (SubscriptionCacheEntity.INFLUX,)
```
matthewelwell: Do we need to separate this out? Can't we just do: 

```python
subscription_info_cache.update_caches((SubscriptionCacheEntity.INFLUX,))
```

If we want to call it out to it's own line, we should update the variable name as it's not `influx_cache`. I'd say we should just reuse the name of the argument in the function? 

```python
update_cache_entities=(SubscriptionCacheEntity.INFLUX,)
```

Same comment applies to the function below as well.

api/organisations/tests/test_views.py
```diff
@@ -42,6 +37,14 @@
 User = get_user_model()
 
 
+class MockSubscriptionMetadata:
```
matthewelwell: I don't think we need this class do we? We can just use the actual `ChargebeeObjMetadata` class in the tests.

api/organisations/tests/test_views.py
```diff
@@ -803,7 +847,7 @@ def test_get_subscription_metadata(
     expected_chargebee_email = "test@example.com"
 
     get_subscription_metadata = mocker.patch(
-        "organisations.models.get_subscription_metadata",
+        "organisations.models.Subscription.get_subscription_metadata",
```
matthewelwell: I prefer not to patch methods on concrete objects unless we really have to, is there any way to avoid it here?

api/organisations/utils.py
```diff
@@ -0,0 +1,20 @@
+def dict_to_class(dictionary):
```
matthewelwell: This function is super ugly (and untested!), we cannot abuse python like this! Can we rethink this? I assume it's because we're trying to coerce a dictionary to a chargebee.Subscription model. 

If so, we should just create our own dataclass perhaps which we can create from a chargebee.Subscription model or a dictionary.

matthewelwell (changes requested):

api/organisations/chargebee/chargebee.py
novakzaballa: Renamed to `chargebee_subscription`, and the type was updated to `dict`.

api/organisations/chargebee/chargebee.py
novakzaballa: Done.

api/organisations/subscription_info_cache.py
novakzaballa: Done.

api/organisations/tasks.py
novakzaballa: Done.

api/organisations/tests/test_views.py
novakzaballa: Deleted.

api/organisations/utils.py
novakzaballa: Function removed. The approach changed to use `dict` instead of a class as the type of the `subscription` param.

api/organisations/tests/test_views.py
novakzaballa: Not sure if I understood this observation. The prior version of this test was using almost the same approach but the `get_subscription_metadata` was imported from chargebee and now we are using the method with the same name in the `organisations.models.Subscription` model.

api/organisations/chargebee/chargebee.py
novakzaballa: Renamed.

api/organisations/tests/test_views.py
matthewelwell: Well, no - before we were mocking a function that is imported into the `organisations.models` module. Now we are mocking a method on the `Subscription` class which lives in the `organisations.models` module. It's very different and can cause confusion! It will work and is not the _end_ of the world, but it should be a last resort. Can you look into whether you can revert this to just mocking an imported function?

api/organisations/chargebee/chargebee.py
```diff
@@ -112,20 +133,15 @@ def get_subscription_metadata(
...
-            quantity = getattr(addon, "quantity", None) or 1
-            addon_metadata = chargebee_cache.addons[addon.id] * quantity
-            subscription_metadata = subscription_metadata + addon_metadata
-
-        return subscription_metadata
+        subscription = vars(chargebee_result.subscription)
+        # convert the addons into a list of dictionaries since vars don't do it recursively
+        subscription["addons"] = [
+            vars(addon) for addon in chargebee_result.subscription.addons
+        ]
```
matthewelwell: Let's move this behaviour to another function. `_convert_chargebee_subscription_to_dictionary` perhaps?

novakzaballa: Created.

api/organisations/tests/test_views.py
novakzaballa: Modify as discussed.

api/organisations/tests/test_views.py
```diff
@@ -608,15 +623,69 @@ def test_when_subscription_plan_is_changed_max_seats_and_max_api_calls_are_updat
...
         assert self.subscription.plan == new_plan_id
         assert self.subscription.max_seats == new_max_seats
         assert self.subscription.max_api_calls == new_max_api_calls
 
+        assert subscription_information_cache.allowed_seats == new_max_seats
+        assert subscription_information_cache.allowed_30d_api_calls == new_max_api_calls
+        assert subscription_information_cache.allowed_projects == new_max_projects
+        assert subscription_information_cache.chargebee_email == self.cb_user.email
+
+        # Update plan
```
matthewelwell: Should this be a new test?

api/organisations/tests/test_views.py
```diff
@@ -673,20 +744,38 @@ def test_when_subscription_is_cancelled_then_cancellation_date_set_and_alert_sen
...
-                }
+                    "addons": [
+                        {
+                            "id": "addon-id",
+                            "quantity": 2,
+                            "unit_price": 0,
+                            "amount": 0,
+                            "object": "addon",
+                        }
+                    ],
```
matthewelwell: I'm not sure why this needed adding here?

api/organisations/tests/test_views.py
```diff
@@ -851,18 +953,26 @@ def test_get_subscription_metadata_returns_404_if_the_organisation_have_no_subsc
...
+        "max_api_calls": expected_api_calls,
+        "payment_source": CHARGEBEE,
+        "chargebee_email": expected_chargebee_email,
+    }
+    get_subscription_metadata.assert_called_once_with(
+        chargebee_subscription.subscription_id
+    )
 
 
-def test_get_subscription_metadata_returns_defaults_if_chargebee_error(
```
matthewelwell, line 858: Is this test no longer valid?

api/organisations/views.py
```diff
@@ -269,27 +274,43 @@ def chargebee_webhook(request):
...
+                    "organisation_id": existing_subscription.organisation_id,
+                    "allowed_projects": subscription_metadata.projects,
+                    "chargebee_email": subscription_metadata.chargebee_email,
+                },
+            )
+
         elif subscription_status in ("non_renewing", "cancelled"):
             existing_subscription.cancel(
-                datetime.fromtimestamp(subscription_data.get("current_term_end")),
+                datetime.fromtimestamp(subscription_data.get("current_term_end", "")),
```
matthewelwell: I'm not sure why this was added, I think this would just error in the same way...

```
>>> import datetime
>>> datetime.datetime.fromtimestamp("")
  File "<stdin>", line 1, in <module>
TypeError: 'str' object cannot be interpreted as an integer
>>> datetime.datetime.fromtimestamp(None)
  File "<stdin>", line 1, in <module>
TypeError: 'NoneType' object cannot be interpreted as an integer
```

api/sales_dashboard/templates/sales_dashboard/home.html
```diff
@@ -18,9 +18,13 @@ <h1 class="h2">Organisations</h1>
...
           <div class='float-left'>
-            <form method="POST" action="{% url 'sales_dashboard:update-organisation-subscription-information-caches' %}">
+            <form method="POST" action="{% url 'sales_dashboard:update-organisation-subscription-information-influx-caches' %}">
               {% csrf_token %}
-              <input type="submit" value="Update subscription information caches" class="btn btn-primary m-1">
+              <input type="submit" value="Update subscription information caches from influx" class="btn btn-primary m-1">
+            </form>
+            <form method="POST" action="{% url 'sales_dashboard:update-organisation-subscription-information-chargebee-caches' %}">
+              {% csrf_token %}
+              <input type="submit" value="Update subscription information caches from chargebee" class="btn btn-primary m-1">
```
matthewelwell: I guess the ideal here would be to have a separate form (perhaps in a modal) that has 2 check boxes and allows you to define which caches to update. My _slight_ concern here is that I'm not too sure what will happen if someone presses both buttons in quick succession.

api/sales_dashboard/templates/sales_dashboard/home.html
```diff
@@ -18,9 +18,13 @@ <h1 class="h2">Organisations</h1>
...
             </form>
             <small class="p-1">Last updated at: {{ subscription_information_caches_updated_at|default:"never" }}</small>
```
matthewelwell: I think we need to consider how we display this data now? This will not be accurate anymore.

api/tests/unit/organisations/chargebee/test_unit_chargebee_chargebee.py
```diff
@@ -260,8 +261,9 @@ def test_get_hosted_page_url_for_subscription_upgrade(self):
         )
 
 
-@pytest.mark.parametrize("addon_quantity", (None, 1))
-def test_get_subscription_metadata(mocker, chargebee_object_metadata, addon_quantity):
+def create_mocked_subscription_metadata(
+    mocker, chargebee_object_metadata, addon_quantity
+):
```
matthewelwell: Can we make this a fixture that returns a callable? Also, we need to add type hints to each of the fixtures and the return type.

api/tests/unit/organisations/chargebee/test_unit_chargebee_chargebee.py
```diff
@@ -287,8 +289,40 @@ def test_get_subscription_metadata(mocker, chargebee_object_metadata, addon_quan
...
 
+    return {
+        "customer_email": customer_email,
+        "mocked_chargebee_cache": mocked_chargebee_cache,
+        "mock_subscription_response": mock_subscription_response,
+    }
+
+
+@pytest.mark.parametrize("addon_quantity", (None, 1))
+def test_get_subscription_metadata(mocker, chargebee_object_metadata, addon_quantity):
```
matthewelwell:
```suggestion
def test_extract_subscription_metadata(mocker: MockerFixture, chargebee_object_metadata: ChargebeeObjMetadata, addon_quantity: typing.Optional[int]) -> None:
```

api/tests/unit/organisations/chargebee/test_unit_chargebee_chargebee.py
```diff
@@ -287,8 +289,40 @@ def test_get_subscription_metadata(mocker, chargebee_object_metadata, addon_quan
...
+        "plan_id": plan_id,
+        "addons": [
+            {
+                "id": addon_id,
+                "quantity": 2,
+                "unit_price": 0,
+                "amount": 0,
+            }
+        ],
+    }
```
matthewelwell, line 287: This should probably be in the `Given` section.

api/organisations/tests/test_views.py
novakzaballa: I could be, however, we are using the output of the `create` part as input for the `update `part. I am aware that looks like too large, maybe I can split it somehow and try to reuse code?

api/sales_dashboard/templates/sales_dashboard/home.html
novakzaballa: Maybe we could keep it only for Influx (and probably add a new `updated_at` field for Chargebee) and add a note somewhere specifying that it is referred only to INflux and that ChargeBee is synced automatically?

api/sales_dashboard/templates/sales_dashboard/home.html
novakzaballa: Do we really need to keep both options? I men the update Chagebee should be necessary only once right? hen we could disable it, maybe using a flag?

api/sales_dashboard/templates/sales_dashboard/home.html
matthewelwell: Yep, makes sense!

matthewelwell: Although I'm not sure if you'll be able to use the same logic since it just grabs the first entity and uses the `updated_at` from there - this could have been updated by the chargebee automated process rather than the influx bulk update.

novakzaballa: > - this could have been updated by the chargebee automated process rather than the influx bulk update.
If we have two `updated_at` fields, the one for Influx will be correct since we always update Influx data in bulk, right?
Regarding Chargebee, we could store the `chargebee_updated_at` but I don't think it is worth it to show it in the dashboard right?

api/tests/unit/organisations/chargebee/test_unit_chargebee_chargebee.py
novakzaballa: Moved.

api/tests/unit/organisations/chargebee/test_unit_chargebee_chargebee.py
novakzaballa: Done.

api/organisations/views.py
novakzaballa: Removed.

api/organisations/tests/test_views.py
novakzaballa: Got it, removed.

api/organisations/tests/test_views.py
novakzaballa: Yes, it's valid, add it again.

api/organisations/tests/test_views.py
novakzaballa: Splitted.

api/sales_dashboard/templates/sales_dashboard/home.html
novakzaballa: Two new fields, 'influx_updated_at' and 'chargebee_updated_at', were added to the 'organisations_organisationsubscriptioninformationcache' table. The date is obtained from 'influx_updated_at'.

api/sales_dashboard/templates/sales_dashboard/home.html
novakzaballa: Only 2 buttons were left, one that updates Influx data, and another that updates both(Influx and Chargebee).

api/conftest.py
```diff
@@ -439,3 +445,40 @@ def create_project_permission(db):
 @pytest.fixture
 def manage_user_group_permission(db):
     return OrganisationPermissionModel.objects.get(key=MANAGE_USER_GROUPS)
+
+
+@pytest.fixture
+def mocked_subscription_metadata(
```
matthewelwell: I'm not sure why this needs to be in the root dir? Feels like it could just be in `api/tests/unit/organisations/chargebee/conftest.py` ?

api/conftest.py
```diff
@@ -439,3 +445,40 @@ def create_project_permission(db):
...
+        "organisations.chargebee.chargebee.ChargebeeCache", autospec=True
+    )
+    mocked_chargebee_cache.return_value.plans = {plan_id: chargebee_object_metadata}
+    mocked_chargebee_cache.return_value.addons = {addon_id: chargebee_object_metadata}
+
+    return {
+        "customer_email": customer_email,
+        "chargebee_object_metadata": chargebee_object_metadata,
+        "mock_subscription_response": mock_subscription_response,
+    }
```
matthewelwell: It's better I think to separate these into 3 separate fixtures.

api/organisations/migrations/0044_auto_20230719_2157.py
```diff
@@ -0,0 +1,27 @@
...
+
+    dependencies = [
+        ('organisations', '0043_organisationsubscriptioninformationcache_allowed_projects'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name='organisationsubscriptioninformationcache',
+            name='updated_at',
+        ),
```
matthewelwell: Hmm... we can't really remove fields as we'll end up with a schema mismatch. The issue is that, when we perform a release, the new release runs the migration on the database but the old application servers are still running until the new ones are healthy. The old servers will expect a field to exist that doesn't and we'll end up with those old servers throwing unhandled exceptions. 

We should just leave the `updated_at` field and add the 2 new ones.

api/organisations/models.py
```diff
@@ -95,6 +96,11 @@ def num_seats(self):
     def has_subscription(self) -> bool:
         return hasattr(self, "subscription") and bool(self.subscription.subscription_id)
 
+    def has_subscription_information_cache(self) -> bool:
+        return hasattr(self, "subscription_information_cache") and bool(
+            self.subscription_information_cache.id
```
matthewelwell: I'm not certain if this will work, but it would be preferable, if possible, to do this:

```suggestion
            self.subscription_information_cache_id
```

api/organisations/subscription_info_cache.py
```diff
@@ -30,14 +30,18 @@ def update_caches():
...
+    if SubscriptionCacheEntity.CHARGEBEE in update_cache_entities:
+        _update_caches_with_chargebee_data(organisations, organisation_info_cache_dict)
 
     to_update = []
     to_create = []
 
     for subscription_info_cache in organisation_info_cache_dict.values():
-        subscription_info_cache.updated_at = timezone.now()
+        subscription_info_cache.chargebee_updated_at = timezone.now()
+        subscription_info_cache.influx_updated_at = timezone.now()
```
matthewelwell: I don't think these should both be updated here, right? We should only update them if the relevant `SubscriptionCacheEntity` was included.

api/sales_dashboard/views.py
```diff
@@ -90,17 +91,19 @@ def get_context_data(self, **kwargs):
         # Use the most recent OrganisationSubscriptionInformationCache object to determine when the caches
         # were last updated.
```
matthewelwell: This comment should be updated.

api/organisations/tests/test_views.py
```diff
@@ -581,42 +583,132 @@ def setUp(self) -> None:
...
-    ):
-        # Given
-        new_plan_id = "new-plan-id"
-        new_max_seats = 3
-        new_max_api_calls = 100
+    @mock.patch("organisations.views.extract_subscription_metadata")
+    def mocked_subscription_plan_data(
+        self,
+        mock_get_subscription_metadata: ChargebeeObjMetadata,
+        mock_get_plan_meta_data: dict[str, str],
```
matthewelwell: These types are not correct - these objects are `MagicMock`s.

api/organisations/tests/test_views.py
```diff
@@ -581,42 +583,132 @@ def setUp(self) -> None:
...
     @mock.patch("organisations.models.get_plan_meta_data")
-    def test_when_subscription_plan_is_changed_max_seats_and_max_api_calls_are_updated(
-        self, mock_get_plan_meta_data
-    ):
-        # Given
-        new_plan_id = "new-plan-id"
-        new_max_seats = 3
-        new_max_api_calls = 100
+    @mock.patch("organisations.views.extract_subscription_metadata")
+    def mocked_subscription_plan_data(
```
matthewelwell: This should be a private method and probably called `_get_mocked_subscription_plan_data`

api/organisations/tests/test_views.py
```diff
@@ -581,42 +583,132 @@ def setUp(self) -> None:
...
+        subscription_information_cache = (
+            OrganisationSubscriptionInformationCache.objects.get(
+                organisation=self.organisation
+            )
+        )
+
+        self.subscription.refresh_from_db()
+        subscription_information_cache.refresh_from_db()
+
+        return [res, subscription_information_cache]
```
matthewelwell: If we have to return multiple entities from this method (I'd rather that we didn't), we should use a tuple. In fact, you can just do: 

```python
return res, subscription_information_cache
```

api/organisations/tests/test_views.py
matthewelwell: Also, a docstring would be helpful as this method is quite complex.

api/organisations/tests/test_views.py
```diff
@@ -581,42 +583,132 @@ def setUp(self) -> None:
...
+        new_max_api_calls = 100
+        new_max_projects = 3
+
+        # When
+        res, subscription_information_cache = self.mocked_subscription_plan_data(
+            plan_id=new_plan_id,
+            max_seats=new_max_seats,
+            max_api_calls=new_max_api_calls,
+            max_projects=new_max_projects,
+        )
```
matthewelwell: It's really hard to see what we're testing here. We need to improve the readability of each of these tests. Can we use better named variables / methods? Should we use comments?

api/organisations/tests/test_views.py
```diff
@@ -793,7 +899,47 @@ def test_trigger_sample_webhook_calls_trigger_sample_webhook_method_with_correct
...
+    subscription_information_cache = (
+        OrganisationSubscriptionInformationCache.objects.create(
+            organisation=organisation,
+            allowed_seats=expected_seats,
+            allowed_projects=expected_projects,
+            allowed_30d_api_calls=expected_api_calls,
+            chargebee_email=expected_chargebee_email,
+        )
+    )
+    subscription_information_cache.refresh_from_db()
```
matthewelwell: I don't know that we need this here?

api/organisations/tests/test_views.py
matthewelwell: But the tests have been swapped around which is slightly confusing. Can we swap them back?

api/tests/unit/organisations/chargebee/test_unit_chargebee_chargebee.py
```diff
@@ -260,35 +261,56 @@ def test_get_hosted_page_url_for_subscription_upgrade(self):
...
-    mock_subscription_response = MockChargeBeeSubscriptionResponse(
-        subscription_id=subscription_id,
-        plan_id=plan_id,
-        customer_email=customer_email,
-        addons=[MockChargeBeeAddOn(addon_id=addon_id, quantity=addon_quantity)],
+    subscription = {
+        "status": status,
+        "id": subscription_id,
+        "plan_id": plan_id,
+        "addons": [
```
matthewelwell, line 279: What about if addons is an empty list? Are there any other edge cases we need to cater for?

api/conftest.py
```diff
@@ -1,4 +1,4 @@
-import typing
+from typing import Tuple
```
matthewelwell, line 1: Not really an issue, but I'm not sure why these changes have snuck into this PR.

api/conftest.py
novakzaballa: Moved.

api/conftest.py
novakzaballa: Done.

api/organisations/migrations/0044_auto_20230719_2157.py
novakzaballa: Done.

api/organisations/subscription_info_cache.py
novakzaballa: Corrected.

api/sales_dashboard/views.py
novakzaballa: Updated.

api/organisations/tests/test_views.py
novakzaballa: Corrected, and docstring added.

api/organisations/tests/test_views.py
```diff
@@ -581,42 +583,161 @@ def setUp(self) -> None:
...
+    @mock.patch("organisations.views.extract_subscription_metadata")
+    def _get_mocked_subscription_plan_data(
+        self,
+        mock_get_subscription_metadata: mock.MagicMock,
+        mock_get_plan_meta_data: mock.MagicMock,
+        plan_id: str,
+        max_seats: int,
+        max_api_calls: int,
+        max_projects: int,
+        chargebee_updated_at: datetime = datetime.now(),
```
matthewelwell: While this is probably ok in tests, it's bad practice to use function returns as default arguments as they are evaluated when the module is loaded. As such, the `chargebee_updated_at` will _always_ take the value that python evaluates when the module is loaded.

api/organisations/tests/test_views.py
novakzaballa: Comments added.

api/organisations/tests/test_views.py
novakzaballa: It was not necessary, deleted.

api/tests/unit/organisations/chargebee/test_unit_chargebee_chargebee.py
novakzaballa: A new test added without addons.

api/organisations/tests/test_views.py
novakzaballa: Thanks, Done.

api/organisations/tests/test_views.py
novakzaballa: Corrected.

api/organisations/models.py
novakzaballa: this does not work:
`subscription_information_cache_id`
but this works:
`subscription_information_cache`

api/organisations/tests/test_views.py
```diff
@@ -581,42 +583,161 @@ def setUp(self) -> None:
...
-                    "plan_id": new_plan_id,
-                }
+                    "plan_id": plan_id,
+                },
+                "customer": {"email": self.cb_user.email},
             }
         }
 
-        # When
+        if plan_id != "updated-plan-id":
```
matthewelwell: Hmm... I don't like that we're basing some logic here on a hard coded string that might get passed in. Can't we just pass a boolean as an argument to this method?

matthewelwell: If we had moved these tests to be pytest based function tests, then we could have just used parameterisation and this could all have been much neater.

api/organisations/tests/test_views.py
```diff
@@ -581,42 +583,161 @@ def setUp(self) -> None:
...
+        # Given
+        plan_id = "plan-id"  # The ID of the current plan
+        max_seats = 3  # The maximum number of seats allowed in the current plan
+        max_api_calls = (
+            100  # The maximum number of API calls allowed in the current plan
+        )
+        max_projects = 3  # The maximum number of projects allowed in the current plan
+
+        # When
+        res, subscription_information_cache = self._get_mocked_subscription_plan_data(
```
matthewelwell: My concern here (and why I think we should have used parameterisation instead) is that, when you look at the test in isolation, it looks like we're testing some mocked data which isn't very useful.

novakzaballa: @matthewelwell  Tests changed as suggested. Can you please take a look at them?

api/organisations/tests/test_views.py
```diff
@@ -952,6 +960,95 @@ def test_organisation_get_influx_data(
...
+    "plan_id, max_seats, max_api_calls, max_projects, is_updated",
+    [
+        ("plan-id", 3, 100, 3, False),
+        ("updated-plan-id", 5, 500, 10, True),
+    ],
+)
+@mock.patch("organisations.models.get_plan_meta_data")
+@mock.patch("organisations.views.extract_subscription_metadata")
+def test_when_plan_is_changed_max_seats_and_max_api_calls_are_updated(
+    mock_get_subscription_metadata,
```
matthewelwell:
```suggestion
    mock_extract_subscription_metadata,
```

api/organisations/tests/test_views.py
```diff
@@ -952,6 +960,95 @@ def test_organisation_get_influx_data(
...
+@pytest.mark.parametrize(
+    "plan_id, max_seats, max_api_calls, max_projects, is_updated",
+    [
+        ("plan-id", 3, 100, 3, False),
+        ("updated-plan-id", 5, 500, 10, True),
+    ],
+)
+@mock.patch("organisations.models.get_plan_meta_data")
+@mock.patch("organisations.views.extract_subscription_metadata")
+def test_when_plan_is_changed_max_seats_and_max_api_calls_are_updated(
```
matthewelwell, line 974: This is much clearer now!

api/organisations/tests/test_views.py
```diff
@@ -952,6 +960,95 @@ def test_organisation_get_influx_data(
...
+            OrganisationSubscriptionInformationCache.objects.create(
+                organisation=organisation,
+                allowed_seats=1,
+                allowed_30d_api_calls=10,
+                allowed_projects=1,
+                chargebee_email=chargebee_email,
+                chargebee_updated_at=updated_at,
+                influx_updated_at=None,
+            )
+        )
```
matthewelwell, line 1026: In general, I don't love controlling test flow like this, based on parameters in the test, but I think it makes sense here to reduce code duplication.

api/organisations/tests/test_views.py
```diff
@@ -952,6 +960,95 @@ def test_organisation_get_influx_data(
...
+    assert subscription.max_api_calls == max_api_calls
+
+    assert subscription_information_cache.allowed_seats == max_seats
+    assert subscription_information_cache.allowed_30d_api_calls == max_api_calls
+    assert subscription_information_cache.allowed_projects == max_projects
+    assert subscription_information_cache.chargebee_email == chargebee_email
+    assert subscription_information_cache.chargebee_updated_at
+    assert subscription_information_cache.influx_updated_at is None
+    if is_updated:
+        assert subscription_information_cache.chargebee_updated_at != updated_at
```
matthewelwell: This should be `>`, no?

matthewelwell: We could also use `pytest-freezegun` if we want to be accurate with these timestamps.

api/organisations/tests/test_views.py
```diff
@@ -952,6 +960,95 @@ def test_organisation_get_influx_data(
...
+
+    # When
+    res = admin_client.post(url, data=json.dumps(data), content_type="application/json")
+
+    subscription_information_cache = (
+        OrganisationSubscriptionInformationCache.objects.get(organisation=organisation)
+    )
+
+    subscription.refresh_from_db()
+    subscription_information_cache.refresh_from_db()
```
matthewelwell: This is unnecessary since we're getting it from the database directly above.

novakzaballa: `subscription_information_cache.refresh_from_db()`, deleted.

api/organisations/tests/test_views.py
novakzaballa: > This should be `>`, no?

Corrected.


> We could also use `pytest-freezegun` if we want to be accurate with these timestamps.

Added and used

api/organisations/tests/test_views.py
novakzaballa: Done.

api/organisations/tests/test_views.py
novakzaballa: Swapped.

matthewelwell (approved):

PR #2282, opened 2023-06-08
improvement/add robots.txt

### LINKED ISSUE #2281: Self hosted installs should default to noindex in robots.txt
The default `robots.txt` on the front end when self hosted should probably be:

```
User-agent: *
Disallow: /
```

I cant think of a good reason why anyone would not want this. 

For our SaaS platform we should have a different `robots.txt` that includes indexing.

dabeeeenster:
Stop google indexing self-hosted installs. 

~~I havent applied this to the dedicated front end builds as we want google to index our SaaS version. This will need https://github.com/Flagsmith/flagsmith/issues/2287~~

The Front end image already does this here https://github.com/Flagsmith/flagsmith/blob/d5cd937036ee28473548da358c7f11288f267cc2/frontend/api/index.js#L158

### DIFF
--- a/api/app/templates/robots.txt
+++ b/api/app/templates/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Disallow: /
--- a/api/app/urls.py
+++ b/api/app/urls.py
@@ -4,6 +4,7 @@ from django.conf import settings
 from django.conf.urls import include, url
 from django.contrib import admin
 from django.urls import path
+from django.views.generic.base import TemplateView
 
 from users.views import password_reset_redirect
 
@@ -32,6 +33,10 @@ urlpatterns = [
         name="project_overrides",
     ),
     path("processor/", include("task_processor.urls")),
+    path(
+        "robots.txt",
+        TemplateView.as_view(template_name="robots.txt", content_type="text/plain"),
+    ),
 ]
 
 if settings.DEBUG:


novakzaballa (approved):
LGTM after re-running E2E tests

matthewelwell:
@dabeeeenster a couple of thoughts on this: 

 1. This will only work in a single container install as the route is defined on the API
 2. I'm not sure what you mean by 'This will need #2287' ? What does this PR actually do then? What more needs doing? Can we merge this (once point 1 is resolved)?

dabeeeenster:
> 



> @dabeeeenster a couple of thoughts on this:
> 
> 1. This will only work in a single container install as the route is defined on the API

This PR is for the unified container. Not sure what you mean by the above statement? 

> 2. I'm not sure what you mean by 'This will need [Define robots.txt in front end only docker image #2287](https://github.com/Flagsmith/flagsmith/issues/2287)' ? What does this PR actually do then? What more needs doing? Can we merge this (once point 1 is resolved)?

We need to also define the robots.txt in the Front End containerso that google doesnt index installs with 1 FE and 1 BE container.

matthewelwell:
So, the description should actually read: 'Stop google indexing single-container self-hosted installs.'

matthewelwell (approved):

PR #2283, opened 2023-06-08
Limit invite users according with org max seats

### LINKED ISSUE #1969: Invite URL allows free saas users to break the 1 user limit
**Describe the bug**
Create a free saas org, send the invite URL to someone else, they can sign in and join the org.

**To Reproduce**
Create a free saas org, send the invite URL to someone else, they can sign in and join the org.

**Expected behavior**
They are told they are at the seat limit and need to upgrade

- [ ] Self Hosted with Docker
- [x] SaaS at flagsmith.com
- [ ] Some other way

**Additional context**

gz#461

novakzaballa:

### Backend:
Validates the organisation seats used vs max seats on the `GET api/v1/organisations/{pk}/invite-links` and `POST api/v1/users/join/#` endpoints. Throws HTTP Error 403.

### Frontend
- Handles `organisation reached max seats` error on API `get-invite-links` endpoint.
- Show max seats reached message.
Screenshot 2023-06-09 at 01 10 00

- On sign-up from invite link, catches API error when organisation reached max seats 
Screen Shot 2023-06-22 at 15 34 09

Manual Locally
- In the local database choose an organisation in `organisations_subscription` table, and set `max_seat` value to N, being N the max number of seats you want to test.
- Using the invite link create N users. Copy the invite link to re-use it.
- See that when the number of users is N the invite link is not displayed and a warning message is shown instead. 
- See that for the n + 1 attempt you are not able to sign-up using the existing invite link.

E2E tests should be considered for the scenarios covered by the new behavior implemented.


api/organisations/models.py
```diff
@@ -99,12 +99,12 @@ def has_subscription(self) -> bool:
     def is_paid(self):
         return self.has_subscription() and self.subscription.cancellation_date is None
 
-    def over_plan_seats_limit(self):
+    def over_plan_seats_limit(self, shift=0):
```
matthewelwell: Let's use `additional_seats` as the argument.

api/organisations/invites/views.py
```diff
@@ -59,6 +59,12 @@ def join_organisation_from_link(request, hash):
 
     invite = get_object_or_404(InviteLink, hash=hash)
 
+    organisation = Organisation.objects.get(id=invite.organisation.id)
+    if organisation.over_plan_seats_limit(shift=1):
+        raise PermissionDenied(
+            "The seats limit has been reached. Please contact your organisation."
+        )
```
matthewelwell: Should reuse the logic [here](https://github.com/Flagsmith/flagsmith/blob/main/api/organisations/invites/views.py#L131-L136)

api/organisations/invites/views.py
```diff
@@ -85,6 +91,12 @@ class InviteLinkViewSet(
     def get_queryset(self):
         organisation_pk = self.kwargs.get("organisation_pk")
         user = self.request.user
+        organisation = Organisation.objects.get(id=organisation_pk)
+        if organisation.over_plan_seats_limit(shift=1):
+            raise PermissionDenied(
+                "The seats limit has been reached in your organisation."
+            )
```
matthewelwell: Should reuse the logic from [here](https://github.com/Flagsmith/flagsmith/blob/main/api/organisations/invites/views.py#L131-L136)

api/organisations/models.py
matthewelwell:
```suggestion
    def over_plan_seats_limit(self, additional_seats: int = 0):
```

novakzaballa: This is no longer necessary applying your suggestions below

api/organisations/invites/views.py
novakzaballa: done

api/organisations/invites/views.py
novakzaballa: done

api/organisations/invites/views.py
```diff
@@ -58,6 +58,15 @@ def join_organisation_from_link(request, hash):
         raise PermissionDenied("Invite links are disabled.")
 
     invite = get_object_or_404(InviteLink, hash=hash)
+    organisation = Organisation.objects.get(id=invite.organisation.id)
+    subscription_metadata = organisation.subscription.get_subscription_metadata()
```
matthewelwell: Why are you doing another explicit lookup in the database here? You can just do: 

```python
invite.organisation.subscription.get_subscription_metadata()
```

if you need the organisation explicitly, just do `organisation = invite.organisation`. This will do a lookup in the database if it needs to (i.e. if a join wasn't done on the initial query).

api/organisations/invites/views.py
```diff
@@ -85,6 +94,16 @@ class InviteLinkViewSet(
...
         user = self.request.user
+        organisation = Organisation.objects.get(id=organisation_pk)
+        subscription_metadata = organisation.subscription.get_subscription_metadata()
+
+        if (
+            len(settings.AUTO_SEAT_UPGRADE_PLANS) > 0
+            and organisation.num_seats >= subscription_metadata.seats
+            and not organisation.subscription.can_auto_upgrade_seats
+        ):
+            raise SubscriptionDoesNotSupportSeatUpgrade()
```
matthewelwell: DRY! Can you extract this somewhere else?

matthewelwell (changes requested):

api/organisations/invites/views.py
novakzaballa: Done. Sorry, it seems like I left behind work in progress.

api/e2etests/permissions.py
```diff
@@ -5,6 +5,8 @@
 
 class E2ETestPermission(BasePermission):
     def has_permission(self, request, view):
+        os.environ.setdefault("E2E_TEST_AUTH_TOKEN", "some-token")
```
matthewelwell: What was the reasoning for adding this? I think we probably shouldn't have it because, if we do, then people could exploit the E2E teardown on self hosted installs?

api/organisations/invites/tests/test_views.py
```diff
@@ -62,6 +69,22 @@ def test_get_invite_links_for_organisation(self):
         for invite_link in response_json:
             assert all(attr in invite_link for attr in expected_attributes)
 
+    def test_get_invite_links_for_organisation_returns_400(self):
```
matthewelwell: Can we be more descriptive with this test? I'm guessing the reason this is returning 400 is because there is no paid subscription, and hence they can't add more seats?

api/organisations/invites/tests/test_views.py
```diff
@@ -70,12 +93,29 @@ def test_delete_invite_link_for_organisation(self):
...
+        # update subscription to add another seat
+        self.set_subscription_max_seats(2)
+
         # When
         response = self.client.delete(url)
 
         # Then
         assert response.status_code == status.HTTP_204_NO_CONTENT
 
+    def test_delete_invite_link_for_organisation_return_400(self):
```
matthewelwell: Why should the delete endpoint return 400? I guess it makes sense but I don't think we particularly care about this do we?

api/organisations/models.py
```diff
@@ -110,6 +110,16 @@ def reset_alert_status(self):
         self.alerted_over_plan_limit = False
         self.save()
 
+    def seats_at_limit_and_cannot_autoupgrade(self):
```
matthewelwell: This method is probably doing too much (the clue is in the fact that you're having to use `and` in the name). 

Why not just have a method called `is_auto_seat_upgrade_possible()` and another one to check if they're at or over their seats limit? So, you'd end up with something like: 


```python
def is_auto_seat_upgrade_available(self) -> bool:
    return len(settings.AUTO_SEAT_UPGRADE_PLANS) > 0 and self.subscription.can_auto_upgrade_seats

def has_reached_seats_limit(self) -> bool:
    if self.has_subscription():
        subscription_metadata = self.subscription.get_subscription_metadata()
        return self.num_seats >= subscription_metadata.seats
    return self.num_seats >= getattr(self.subscription, "max_seats", 1)
```

Then when evaluating whether to raise an error, you can just do: 

```python
if organisation.has_reached_seats_limit() and not organisation.is_auto_seat_upgrade_available():
    raise SubscriptionDoesNotSupportSeatUpgrade()
```

I think that's much cleaner / clearer?

matthewelwell: Also, please bear in mind that I would expect any new methods to have unit tests covering them (not just the view tests you've added above).

api/e2etests/permissions.py
novakzaballa: Was inserted to manage an error during testing, left behind by mistake. Removed.

api/organisations/invites/tests/test_views.py
novakzaballa: Done

api/organisations/models.py
novakzaballa: Method split and tests added.

api/organisations/invites/tests/test_views.py
novakzaballa: Delete is calling the get view_set and therefore it is returning 400 when the seat limit has been reached. We could refactor the delete endpoint, however, in practice, it is not expected that customers who have reached their seat limit to be able to see the invite link and therefore they should not be able to delete it or re-create it.

api/organisations/tests/test_models.py
```diff
@@ -114,7 +114,11 @@ def test_organisation_over_plan_seats_limit_returns_true_if_over_plan_seats_limi
 
 def test_organisation_over_plan_seats_no_subscription(organisation, mocker, admin_user):
     # Given
-    mocker.patch("organisations.models.MAX_SEATS_IN_FREE_PLAN", 0)
+    mocker.patch(
```
matthewelwell: We don't need to mock this - `max_seats` is a concrete field on the subscription class, we can just do 

```python
organisation.subscription.max_seats = 0
organisation.subscription.save()
```

matthewelwell (changes requested):

api/organisations/tests/test_models.py
novakzaballa: Ok thanks, done.

matthewelwell (dismissed):

matthewelwell:
@novakzaballa seems like the E2E test failures here are genuine - are you able to reproduce locally?

novakzaballa:
> @novakzaballa seems like the E2E test failures here are genuine - are you able to reproduce locally?

Nope, I am not able to reproduce the error locally, I am currently struggling with that.

novakzaballa:
> @novakzaballa seems like the E2E test failures here are genuine - are you able to reproduce locally?

I was finally able to reproduce the error and fix it. It seems like I broke my E2E configuration and one of the E2E tests when I merged the latest `main` changes into this branch. Now all tests passed.

matthewelwell:
I just tested this using the Uffizzi preview environment ([here](https://pr-2283-deployment-28007-flagsmith.app.uffizzi.com/)) and I was able to invite and register a new user in an organisation on the free plan using an email invite. This created 2 questions in my head: 

 1. Surely, based on the logic we've implemented here - that shouldn't be possible, right? Interestingly, the `/invite-links` endpoint returned a 400 as I would expect based on the implementation. 
 2. How will this affect people self hosting? As per [these docs](https://docs.flagsmith.com/version-comparison), we want to make sure that there is no limit on users for people that are self hosting.

matthewelwell:
Ok, so I've looked into this a bit further and I have a feeling we're going about this slightly wrong. Mostly based on the second point above. 

To clarify, the issue we have is that in **SaaS**, users on the free or startup plans should not be able to use the invite links to break their plan limits. Users of the self hosted should still be able to use the invite links as there are no user limits. 

It looks like the frontend is using the `payments_enabled` flag to determine whether it is running in SaaS or not. You can see the logic defined [here](https://github.com/Flagsmith/flagsmith/blob/main/frontend/web/components/pages/OrganisationSettingsPage.js#L286-L292). Can we use this to drive the behaviour from the frontend perspective? 

Perhaps we could also implement restrictions on the API side purely based on the existence of the `ENABLE_CHARGEBEE` which would realistically give the same result as the `payments_enabled` flag the frontend is using. 

Based on this, I would suggest that the changes to this PR should be as follows: 

 1. Hide the Invite Links from the Organisation Settings page using the variables already defined [here](https://github.com/Flagsmith/flagsmith/blob/main/frontend/web/components/pages/OrganisationSettingsPage.js#L286-L292)
 2. On the API, in the `FFAdminUser.join_organisation_from_invite()` method, we should extend the logic to check to see if the `ENABLE_CHARGEBEE` setting is enabled and, if it is, reject the user if they would breach their plan and auto seats upgrade is not possible. Something like the following should work: 

```python
    def join_organisation_from_invite(self, invite: "AbstractBaseInviteModel"):
        ...

        organisation = invite.organisation
        if settings.ENABLE_CHARGEBEE and organisation.over_plan_seats_limit(additional_seats=1):
            if organisation.is_auto_seat_upgrade_available():
                organisation.subscription.add_single_seat()
            else:
                raise SubscriptionDoesNotSupportSeatUpgrade()
```

novakzaballa:
> Ok, so I've looked into this a bit further and I have a feeling we're going about this slightly wrong. Mostly based on the second point above.
> 
> To clarify, the issue we have is that in **SaaS**, users on the free or startup plans should not be able to use the invite links to break their plan limits. Users of the self hosted should still be able to use the invite links as there are no user limits.
> 
> It looks like the frontend is using the `payments_enabled` flag to determine whether it is running in SaaS or not. You can see the logic defined [here](https://github.com/Flagsmith/flagsmith/blob/main/frontend/web/components/pages/OrganisationSettingsPage.js#L286-L292). Can we use this to drive the behaviour from the frontend perspective?
> 
> Perhaps we could also implement restrictions on the API side purely based on the existence of the `ENABLE_CHARGEBEE` which would realistically give the same result as the `payments_enabled` flag the frontend is using.
> 
> Based on this, I would suggest that the changes to this PR should be as follows:
> 
> 1. Hide the Invite Links from the Organisation Settings page using the variables already defined [here](https://github.com/Flagsmith/flagsmith/blob/main/frontend/web/components/pages/OrganisationSettingsPage.js#L286-L292)
> 2. On the API, in the `FFAdminUser.join_organisation_from_invite()` method, we should extend the logic to check to see if the `ENABLE_CHARGEBEE` setting is enabled and, if it is, reject the user if they would breach their plan and auto seats upgrade is not possible. Something like the following should work:
> 
> ```python
>     def join_organisation_from_invite(self, invite: "AbstractBaseInviteModel"):
>         ...
> 
>         organisation = invite.organisation
>         if settings.ENABLE_CHARGEBEE and organisation.over_plan_seats_limit(additional_seats=1):
>             if organisation.is_auto_seat_upgrade_available():
>                 organisation.subscription.add_single_seat()
>             else:
>                 raise SubscriptionDoesNotSupportSeatUpgrade()
> ```

Regarding point 1 referring to the frontend, the logic is already being used, by using the variable `needsUpgradeForAdditionalSeats` in this line: https://github.com/Flagsmith/flagsmith/blob/8afbb9e0029bc56e0f8e04653b20319841a6cec5/frontend/web/components/pages/OrganisationSettingsPage.js#L671C43-L671C43

novakzaballa:
> Ok, so I've looked into this a bit further and I have a feeling we're going about this slightly wrong. Mostly based on the second point above.
> 
> To clarify, the issue we have is that in **SaaS**, users on the free or startup plans should not be able to use the invite links to break their plan limits. Users of the self hosted should still be able to use the invite links as there are no user limits.
> 
> It looks like the frontend is using the `payments_enabled` flag to determine whether it is running in SaaS or not. You can see the logic defined [here](https://github.com/Flagsmith/flagsmith/blob/main/frontend/web/components/pages/OrganisationSettingsPage.js#L286-L292). Can we use this to drive the behaviour from the frontend perspective?
> 
> Perhaps we could also implement restrictions on the API side purely based on the existence of the `ENABLE_CHARGEBEE` which would realistically give the same result as the `payments_enabled` flag the frontend is using.
> 
> Based on this, I would suggest that the changes to this PR should be as follows:
> 
> 1. Hide the Invite Links from the Organisation Settings page using the variables already defined [here](https://github.com/Flagsmith/flagsmith/blob/main/frontend/web/components/pages/OrganisationSettingsPage.js#L286-L292)
> 2. On the API, in the `FFAdminUser.join_organisation_from_invite()` method, we should extend the logic to check to see if the `ENABLE_CHARGEBEE` setting is enabled and, if it is, reject the user if they would breach their plan and auto seats upgrade is not possible. Something like the following should work:
> 
> ```python
>     def join_organisation_from_invite(self, invite: "AbstractBaseInviteModel"):
>         ...
> 
>         organisation = invite.organisation
>         if settings.ENABLE_CHARGEBEE and organisation.over_plan_seats_limit(additional_seats=1):
>             if organisation.is_auto_seat_upgrade_available():
>                 organisation.subscription.add_single_seat()
>             else:
>                 raise SubscriptionDoesNotSupportSeatUpgrade()
> ```

Done

api/organisations/invites/views.py
```diff
@@ -59,6 +59,12 @@ def join_organisation_from_link(request, hash):
 
     invite = get_object_or_404(InviteLink, hash=hash)
 
+    if (
```
matthewelwell: Now that we've updated the logic in the `join_organisation_from_invite` function we should remove this validation (and from the `InviteLinkViewSet.get_queryset` below) or update it to also verify if payments are enabled, before we throw the exception.

matthewelwell (changes requested):

matthewelwell (approved):

PR #2284, opened 2023-06-09
Fix intermitent issue toggling segment overrides

### LINKED ISSUE #2067: Intermittent issue toggling segment overrides
It seems that there is an intermittent issue where a segment override can get into a state where it is not possible to update it's state. 

The underlying cause for this seems to be that there is a FeatureSegment entity without a corresponding FeatureState entity. My guess is that this is because these objects are created separately in 2 separate API requests, so there is a chance that the FeatureSegment is created but the creation of the FeatureState fails. 

Some thoughts about how we might resolve this: 

 1. To resolve the immediate issue, we can remove any orphan FeatureSegment entities. This shouldn't impact the behaviour of the API but will remove these 'broken' overrides from users' dashboards (there are currently ~25 instances in the production database). 
 2. Implement some logic in the frontend such that it cleans up the FeatureSegment if the call to create the FeatureState fails (this is likely not foolproof). 
 3. Have a recurring task clean up any orphaned FeatureSegment entities
 4. Create a new endpoint to create a segment override which will create both the FeatureSegment and the FeatureState at once (so we get transactional consistency)

My feeling is that option 4 is the only 'foolproof' option here.

novakzaballa:

### Backend
Add previously existing attributes to the response of the `create-segment-overrides` endpoint. 

### Frontend
Use the new endpoint

- Create a feature flag
- Create a Segment
- Create a Segment Override through the segments page
- Create a Segment override through the Edit Feature modal.


frontend/common/stores/feature-list-store.js
```diff
@@ -196,6 +197,7 @@ const controller = {
     onComplete,
   ) => {
     let prom
+    const featureFlagId = flag.id || projectFlag.id || environmentFlag.feature
```
novakzaballa: @kyle-ssg  Is this is a good way to determine the feature flag ID?

kyle-ssg: projectFlag.id

novakzaballa: Thanks @kyle-ssg , I am using now the `v.feature` value which was the one used before my changes. I am using the same as the future inside the `feature_segment_value` structure.

frontend/common/stores/feature-list-store.js
```diff
@@ -293,46 +308,49 @@ const controller = {
...
-                            multivariate_feature_option:
-                              environmentFlag.multivariate_feature_state_values[
-                                o.multivariate_feature_option_index
-                              ].multivariate_feature_option,
-                            percentage_allocation: o.percentage_allocation,
-                          }
-                        }),
-                    },
-                  ),
+      const segmentOverridesRequest = console.log(
```
matthewelwell: This looks odd - we're assigning the `segmentOverridesRequest` variable to the return from a `console.log` statement?

novakzaballa: Sorry, work in progress left behind. Deleted.

frontend/common/types/responses.ts
```diff
@@ -92,6 +93,12 @@ export type UserPermission = {
 export type GroupPermission = Omit<UserPermission, 'user'> & {
   group: UserGroup
 }
+export type FeatureStateValue = {
+  boolean_value: boolean
```
kyle-ssg: I think this is an incorrect duplicate of FeatureStateValue on line 165

kyle-ssg (approved):

frontend/common/types/responses.ts
kyle-ssg: Actually if I'm not mistaken, both are wrong and the fields should be nullable rather than allow for undefined (perhaps best to check on BE)

novakzaballa: The duplicated type was removed, and the fields were updated by verifying the BE.

api/features/serializers.py
```diff
@@ -395,9 +395,19 @@ class CreateSegmentOverrideFeatureStateSerializer(WritableNestedModelSerializer)
...
             "feature_segment",
+            "deleted_at",
+            "uuid",
+            "created_at",
+            "updated_at",
+            "version",
+            "live_from",
+            "environment",
+            "identity",
+            "change_request",
```
matthewelwell: Do we _need_ all these fields? I don't think, for example, the frontend is using the `version` field? Certainly I think we should be making a lot of these `read_only_fields`.

matthewelwell (changes requested):

api/features/serializers.py
novakzaballa: It looks like  `version` is the only field not used in the FE, I'll remove it from the response. 
Given that we only use this data structure in this new end_point and it is an **agregate**, should we mark all the fields as `read_only_fields`?

matthewelwell: Yes, I think we should mark as many as possible as read only.

matthewelwell: Just saw your latest changes - I don't think that they can _all_ be read only - only the new ones that you added probably?

matthewelwell:
@novakzaballa can we remove the TODO section at the bottom of the PR description?

api/features/serializers.py
matthewelwell: We need to remove `"version"` from `read_only_fields` [here](https://github.com/Flagsmith/flagsmith/pull/2284/files#diff-882e5a880410fc240a0f5909e834923bd492b4eda74b4fa87e1ba931cff7438eR418).

novakzaballa:
> @novakzaballa can we remove the TODO section at the bottom of the PR description?

Removed.

api/features/serializers.py
novakzaballa: Removed.

matthewelwell (approved):

PR #2289, opened 2023-06-10
Design System Setup - Update bootstrap + buttons

kyle-ssg:

## Changes 

- Sass and Boostrap are at latest versions
- Button uses a theme property
- Button css is less custom and more uses built in variables 
- Creates a _variables-new file to setup colours, fonts etc from bootstrap

The main thing to test is that there is no major visual regression in the app as this PR introduces minimal design changes. The follow up pull requests to this will start introducing new components.


matthewelwell:
There seems to be some inconsistencies with the buttons

image

image

I'm not sure why the 'Create Project' and 'Create Organisation' have a '+' character but the create feature (among others, e.g. 'Create Segment') don't? Interestingly - it looks like 'Create Organisation' always had it but it seems to have been added to 'Create Project' as part of this PR. I don't particularly mind which we go with here, but it would be good to standardise. 

Also, the '+' seems a little misaligned on the 'Create Project' button to me?

kyle-ssg:
@matthewelwell have removed from all for now, this is a stop gap until the page designs are implemented in upcoming PRs.

frontend/web/components/modals/base/Modal.tsx
```diff
@@ -0,0 +1,112 @@
...
+      onClosed={onClose}
+      title={title}
+    >
+      {body}
+    </_ModalDefault>,
+    document.getElementById('modal'),
+  )
+})
+
+export const openModal2 = (global.openModal2 = (
```
matthewelwell, line 93: Can we add any context about what the difference is between `openModal` and `openModal2` ? I assume perhaps that the idea is that `openModal` will be removed eventually in favour of `openModal2` but I don't want this to drag on and the context to get lost.

frontend/web/components/pages/OrganisationSettingsPage.js
```diff
@@ -1201,9 +1197,12 @@ const OrganisationSettingsPage = class extends Component {
                               Audit webhooks let you know when audit logs occur,
                               you can configure 1 or more audit webhooks per
                               organisation.{' '}
-                              <ButtonLink href='https://docs.flagsmith.com/advanced-use/system-administration#audit-log-webhooks/'>
+                              <Button
+                                theme='text'
+                                href='https://docs.flagsmith.com/advanced-use/system-administration#audit-log-webhooks/'
```
matthewelwell, line 1202: @dabeeeenster I have a feeling this has changed / will change?

frontend/web/styles/_variables.scss
```diff
@@ -1,9 +1,11 @@
+@import "new/variables-new";
```
matthewelwell, line 1: is using `new` here the best option? Seems like the sort of thing that might get looked back on in a years time with annoyance...

matthewelwell (commented):
I've clicked around using the preview link and verified that nothing seems broken but also added a couple of general comments.

frontend/web/styles/_variables.scss
kyle-ssg: It'll eventually replace the whole variables file once the designs are complete - they both have to exist in parallel for a while

kyle-ssg: Each pr should delete from the old one and add to the new until we rename

frontend/web/components/modals/base/Modal.tsx
kyle-ssg: There are a few areas where there are modals on modals. This already existed prior to the PR.

kyle-ssg: Added comment

frontend/web/components/pages/OrganisationSettingsPage.js
matthewelwell: This should be updated to https://docs.flagsmith.com/system-administration/webhooks#audit-log-web-hooks as per https://github.com/Flagsmith/flagsmith/pull/2337 - but it obviously also relies on that PR being merged.

kyle-ssg: This seems a bit unrelated to this PR, though can update if the above gets merged soon.

matthewelwell: True - I just spotted it in the diff and had Ben's PR in mind. I guess, maybe the best thing to do is merge this one and then ensure this is updated in Ben's PR.

matthewelwell (approved):

PR #2290, opened 2023-06-12
feat: bump engine, mapper layer

khvn26:

## Changes

*NOTE*: This PR should be merged after github.com/flagsmith/flagsmith-engine/pull/167 is merged and released, with flag_engine's dependency updated accordingly.

- Bumped `flag_engine` to the new version.
- Added ﻿`mapper.engine` and `﻿mapper.dynamodb` modules that rely exclusively on engine's models for validation and use pure Python code for serialization. This approach provides more control over data gathering, better coupling/cohesion with ﻿flag_engine, and potentially better performance. The increments will look nicer, too.
- Updated view code to catch pydantic's validation errors where marshmallow's were used before. `pyngo` is used to convert error messages to DRF.
- `api.util.renderers.PydanticJSONRenderer` added to json-dump datetimes and UUIDs to conform with Pydantic's `dict()` behaviour.
- Added missing `webhook_config` to the big join when gathering data for an environment document, resulting in one less database query.

- Existing unit and E2E tests.
- Covered 100% of `mapper` modules layer with new unit tests.


api/edge_api/identities/serializers.py
```diff
@@ -2,33 +2,34 @@
 import typing
 
 from django.utils import timezone
+from flag_engine.features.models import FeatureModel as EngineFeatureModel
```
matthewelwell, line 5: Seems like we've got a bunch of imports from the same module here all defined individually, is this a linting / formatting thing? Can we combine all the imports into a single statement?

api/edge_api/identities/serializers.py
```diff
@@ -2,33 +2,34 @@
...
+from pydantic import ValidationError as PydanticValidationError
+from pyngo import drf_error_details
 from rest_framework import serializers
 from rest_framework.exceptions import ValidationError
 
 from environments.models import Environment
 from features.models import Feature, FeatureState, FeatureStateValue
 from features.multivariate.models import MultivariateFeatureOption
 from features.serializers import FeatureStateValueSerializer
+from mappers import map_feature_to_engine, map_mv_option_to_engine
```
matthewelwell: Should mappers live at the root level in the repo? It feels slightly wrong but I'm not sure what the alternative is.

api/edge_api/identities/models.py
```diff
@@ -4,18 +4,15 @@
...
-from flag_engine.utils.collections import IdentityFeaturesList
+from flag_engine.identities.builders import build_identity_model
+from flag_engine.identities.models import IdentityFeaturesList, IdentityModel
 
 from api_keys.models import MasterAPIKey
 from environments.dynamodb import DynamoIdentityWrapper
 from environments.models import Environment
 from features.models import FeatureState
 from features.multivariate.models import MultivariateFeatureStateValue
+from mappers import map_engine_identity_to_identity_document
```
matthewelwell: One that might be answered later but I have a feeling we're now defining this logic in the core API and the Edge API, is that correct?

api/edge_api/identities/serializers.py
```diff
@@ -49,9 +50,9 @@ def save(self, **kwargs):
 
 
 class EdgeMultivariateFeatureOptionField(serializers.IntegerField):
-    def to_internal_value(self, data):
+    def to_internal_value(self, data) -> EngineMultivariateFeatureOptionModel:
```
matthewelwell: Can we add a type hint for the `data` argument here too?

api/edge_api/identities/views.py
```diff
@@ -159,8 +158,8 @@ def perform_destroy(self, instance):
     )
     @action(detail=True, methods=["get"], url_path="list-traits")
     def get_traits(self, request, *args, **kwargs):
-        identity = self.get_object()
-        data = trait_schema.dump(identity["identity_traits"], many=True)
+        identity = build_identity_model(self.get_object())
```
matthewelwell, line 164: Maybe I'm a bit lost here, but should this still be using the `build_environment_model` function or should it be using a mapper function now?

api/mappers/engine.py
```diff
@@ -0,0 +1,404 @@
...
+) -> Optional[IntegrationModel]:
+    if not integration:
+        return None
+    return IntegrationModel(
+        api_key=integration.api_key,
+        base_url=integration.base_url,
+    )
+
+
+def map_dynatrace_integation_to_engine(
```
matthewelwell:
```suggestion
def map_dynatrace_integration_to_engine(
```

matthewelwell: I guess the DynatraceIntegration is handled differently because of the presence of the `entity_selector` attribute? Is there a reason that we can't just include it in the `map_integration_to_engine` function above by using something like `getattr(integration, "entity_selector", None)` ?

api/mappers/engine.py
```diff
@@ -0,0 +1,404 @@
...
+    mv_option: "MultivariateFeatureOption",
+) -> MultivariateFeatureOptionModel:
+    return MultivariateFeatureOptionModel(value=mv_option.value, id=mv_option.id)
+
+
+def map_environment_to_engine(
+    environment: "Environment",
+) -> EnvironmentModel:
+    """
+    Maps Core API's `environments.models.Environment` mode instance to the
```
matthewelwell:
```suggestion
    Maps Core API's `environments.models.Environment` model instance to the
```

api/mappers/dynamodb.py
```diff
@@ -0,0 +1,114 @@
...
+DocumentValue: TypeAlias = Union[Dict[str, "DocumentValue"], str, bool, None, Decimal]
+Document: TypeAlias = Dict[str, DocumentValue]
+
+
+def map_environment_to_environment_document(
+    environment: "Environment",
+) -> Document:
+    return {
+        field_name: _map_value_to_document_value(value)
+        for field_name, value in map_environment_to_engine(
```
matthewelwell: I'm not too sure how this is supposed to work. As far as I can tell, `map_environment_to_engine` returns an `EnvironmentModel` instance which is a dataclass (and hence not iterable).

api/mappers/dynamodb.py
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+
+def map_identity_to_identity_document(
+    identity: "Identity",
+) -> Document:
+    return map_engine_identity_to_identity_document(map_identity_to_engine(identity))
+
+
+T = TypeVar("T")
```
matthewelwell: Is there a reason to define this here vs with the other type definitions at the top of the module?

api/mappers/engine.py
```diff
@@ -0,0 +1,404 @@
...
+        feature_segment_model = FeatureSegmentModel(
+            priority=feature_segment.priority,
+        )
+    else:
+        feature_segment_model = None
+
+    return FeatureStateModel(
+        enabled=feature_state.enabled,
+        django_id=feature_state.pk,
+        feature_state_value=feature_state.get_feature_state_value(),
```
matthewelwell: Do we need to consider the identity hash key here?

api/mappers/engine.py
```diff
@@ -0,0 +1,404 @@
...
+
+    :param Environment environment: the environment to map
+    :rtype EnvironmentModel
+    """
+    project: "Project" = environment.project
+    organisation: "Organisation" = project.organisation
+
+    # Read relationships.
+    project_segments: List["Segment"] = project.segments.all()
+    project_segment_rules_by_segment_id: Dict[
```
matthewelwell: I don't quite follow why we need this dictionary?

api/mappers/engine.py
```diff
@@ -0,0 +1,404 @@
...
+    feature versions.
+
+    :param Environment environment: the environment to map
+    :rtype EnvironmentModel
+    """
+    project: "Project" = environment.project
+    organisation: "Organisation" = project.organisation
+
+    # Read relationships.
+    project_segments: List["Segment"] = project.segments.all()
```
matthewelwell: We can optimise this query I think? If we're getting the rules (and conditions) for all the segments here, we should use `prefetch_related`, right? 

Maybe this is already done by the `EnvironmentManager.filter_for_document_builder()` method?

api/mappers/engine.py
```diff
@@ -0,0 +1,404 @@
...
+        project_segments,
+        environment.pk,
+    )
+    environment_feature_states: List["FeatureState"] = [
+        feature_state
+        for feature_state in environment.feature_states.all()
+        if feature_state.feature_segment_id is None
+        and feature_state.identity_id is None
+    ]
+    all_feature_states = (
```
matthewelwell: It's slightly confusing to me that we're using `all_feature_states` since it doesn't include identity overrides? I guess this is because we don't care about them in the environment document? 

Probably not a major issue, but I wonder if there's a better name for this variable?

api/tests/integration/edge_api/identities/test_edge_identity_featurestates_viewset.py
```diff
@@ -694,10 +694,9 @@ def test_edge_identities_post_returns_400_for_invalid_mvfs_allocation(
         identity_uuid
     )
     assert response.status_code == status.HTTP_400_BAD_REQUEST
-    assert (
-        response.json()["multivariate_feature_state_values"]
-        == "Total percentage allocation for feature must be less than 100 percent"
-    )
+    assert response.json()["multivariate_feature_state_values"] == [
+        "Total percentage allocation for feature must be less than 100 percent"
```
matthewelwell: I guess this is technically a breaking API change but it seemingly fits the DRF error models better so maybe we just go with it.

matthewelwell (commented):
I've reviewed up to the tests for now to provide some feedback. 

My general thought is that the mappers code is quite complex / long but perhaps necessary. It feels like it will be easier to change in the future but it's quite a change up front.

api/mappers/dynamodb.py
khvn26: In the new engine version, models are subclasses of Pydantic's `BaseModel`, which support this way of iterating over fields and values.

api/mappers/dynamodb.py
khvn26: None except to emphasize that it's used exclusively by `_noop_encoder`.

api/mappers/engine.py
khvn26: This mapper is only used for environment models, so no.

api/mappers/engine.py
khvn26: Yes, it's expected that all prefetches and joins are defined before the mapper is called.

api/mappers/engine.py
khvn26: I guess we could annotate segment instances with `SegmentRule` lists, but I decided it's a bit cleaner to store those separately.

api/mappers/engine.py
khvn26: It just seemed like a more clear way to describe that only Dynatrace requires this.

api/mappers/engine.py
khvn26: Changed to `all_environment_feature_states`.

api/mappers/engine.py
matthewelwell: I don't think that's true is it? 

https://github.com/Flagsmith/flagsmith/pull/2290/files/3506c14a0e2272589416669eefd8e021bb025021#diff-967f7cb2f63496c9174f36c8a54cdbaca392764a77f891ae4b27efcc6bf612eaR349

api/mappers/engine.py
matthewelwell: I think having a single function here would make it slightly easier to extend in the future if, for example, we want to add another field to all integration models?

api/mappers/engine.py
```diff
@@ -0,0 +1,404 @@
...
+        stop_serving_flags=organisation.stop_serving_flags,
+        persist_trait_data=organisation.persist_trait_data,
+    )
+    project_segment_models = [
+        SegmentModel(
+            id=segment.pk,
+            name=segment.name,
+            rules=[
+                map_segment_rule_to_engine(segment_rule)
+                for segment_rule in project_segment_rules_by_segment_id.pop(segment.pk)
```
matthewelwell: Related to the [above thread](https://github.com/Flagsmith/flagsmith/pull/2290/files/3506c14a0e2272589416669eefd8e021bb025021#r1236843797), yeah - I'm not sure why we don't just do 

```python
map_segment_rule_to_engine(segment_rule) for segment_rule in segment.rules.all()
```

I don't quite see the benefit to having to create a separate dictionary to hold the data?

khvn26: That's the point of the

```
    # No reading from ORM past this point!
```

comment — I really wanted to concentrate all queryset evaluations in one place so they're not scattered across the code.

Maybe it shouldn't matter that much? What do you think?

matthewelwell: Interesting! Yes, I do quite like that idea, but I think it will be very hard to enforce!

A possible improvement here would be to be more explicit with the current `# Read relationships.` comment (currently on line 176). Something like `# Read relationships - grab all the data needed from the ORM here.`

api/mappers/engine.py
khvn26: Yes, this is actually not true — this mapper is also used for identities.

However, this implementation completely mirrors the behaviour of current engine's schemas (see https://github.com/Flagsmith/flagsmith-engine/blob/625a22cb29ed42c5a17b62889d10cf276f55476b/flag_engine/api/schemas.py#L55-L56). This is probably intentional because multivariate feature states are meant to be resolved later in the API business logic.

Please correct me if I didn't get it right.

api/edge_api/identities/models.py
khvn26: This is true.

api/edge_api/identities/serializers.py
khvn26: Yeah, that's how `isort` deals with multiple aliased imports (the `as` keyword).

api/edge_api/identities/views.py
khvn26: I feel it'll be better to swap them out for Pydantic's `parse_obj` method later in a separate PR so we're not overwhelmed with the changes. The engine's model builders are still there for now, for compatibility.

api/edge_api/identities/serializers.py
khvn26: Done.

api/edge_api/identities/serializers.py
khvn26: Yeah, not quite the Django way where every app has its own utils but, given they're used throughout the project, it seems reasonable to me.

api/mappers/engine.py
khvn26: Done.

api/tests/integration/edge_api/identities/test_edge_identity_featurestates_viewset.py
khvn26: Yeah, I think we should.

api/edge_api/identities/serializers.py
matthewelwell: I guess the alternative would be to create a mappers module within each app, right? Is there a reason that a single top-level module is better?

api/edge_api/identities/models.py
matthewelwell: Are there any options to resolve this? I guess the only option would be to put it back in the engine ... or have another package responsible for it? Maybe defining it twice is the best option.

api/mappers/engine.py
```diff
@@ -0,0 +1,393 @@
...
+def map_integration_to_engine(
+    integration: Optional["EnvironmentIntegrationModel"],
+) -> Optional[IntegrationModel]:
+    if not integration:
+        return None
+    kwargs = {
+        "api_key": integration.api_key,
+        "base_url": integration.base_url,
+    }
+    if isinstance(integration, DynatraceConfiguration):
```
matthewelwell: This is fine, just curious why you chose this over `getattr(integration, "entity_selector", None)`?

api/mappers/dynamodb.py
matthewelwell: Ah, yes, of course. Nice, that's cool!

api/mappers/engine.py
matthewelwell: Yep, that makes sense!

api/tests/unit/environments/test_environments_views_sdk_environment.py
```diff
@@ -47,7 +47,7 @@ def test_get_environment_document(
     url = reverse("api-v1:environment-document")
 
     # When
-    with django_assert_num_queries(12):
+    with django_assert_num_queries(11):
```
matthewelwell, line 50: This seems like a good thing... but makes me nervous - do we know why we've dropped a query here?

api/tests/unit/environments/test_unit_environments_models.py
```diff
@@ -180,7 +180,7 @@ def test_environment_get_environment_document(environment, django_assert_num_que
     # Given
 
     # When
-    with django_assert_num_queries(4):
+    with django_assert_num_queries(3):
```
matthewelwell, line 183: As above, do we know why we've dropped a query here? (and below in `test_environment_get_environment_document_with_caching_when_document_not_in_cache`)

api/tests/unit/mappers/test_unit_mappers_dynamodb.py
```diff
@@ -0,0 +1,203 @@
...
+    return TestModel(field="foobar")
+
+
+def test_map_environment_to_environment_document__call_expected(
+    environment: "Environment",
+    mocker: MockerFixture,
+    mock_engine_model: BaseModel,
+) -> None:
+    # Given
+    map_environment_to_engine_mock = mocker.patch(
```
matthewelwell: I'm not too clear on why we need to rely on mocking so heavily here?

api/edge_api/identities/models.py
khvn26: The schemas are defined in engine (with Pydantic models) and now should be the only source of truth about what is stored in the database.

Every `map_engine_{entity}_to_{entity}_document` is a thin wrapper around

1. `BaseModel.dict(entity)`
2. Value type coercion: numeric to `Decimal` and the rest to `str`.

At the moment both Core API and Edge API solve this problem their own way (I personally prefer the value mapper written for Core to marshmallow schemas left in Edge). If we're to factor this logic to a common dependency this will be a very trivial utility. But it surely doesn't belong to the engine, IMO.

api/mappers/engine.py
khvn26: Done.

api/mappers/engine.py
khvn26: For better typing clarity/correctness. The base integration Django model does not have this field and I anticipate type checker errors when we come to enforcing it.

api/tests/unit/mappers/test_unit_mappers_dynamodb.py
```diff
@@ -0,0 +1,203 @@
...
+
+    # Then
+    assert result == document
+    map_identity_to_engine_mock.assert_called_once_with(identity)
+    map_engine_identity_to_identity_document_mock.assert_called_once_with(
+        mock_engine_model,
+    )
+
+
+def test_map_value_to_document_value__return_expected() -> None:
```
matthewelwell: This test is a bit of a mystery to me. I'm not entirely sure what all the set up data is supposed to represent? There seems to be a mix of actual engine models (e.g. MV models) and test models. Maybe this is all valid, but it would probably benefit from some better comments if so.

api/tests/unit/mappers/test_unit_mappers_dynamodb.py
```diff
@@ -0,0 +1,203 @@
...
+                "id": None,
+                "multivariate_feature_option": {"id": None, "value": "foobar"},
+                "mv_fs_value_uuid": str(mv_fs_value.mv_fs_value_uuid),
+                "percentage_allocation": Decimal("50.0"),
+            }
+        ],
+    }
+
+    # When
+    result = dynamodb._map_value_to_document_value(value)
```
matthewelwell: Is there any way that we can avoid testing a 'private' method directly?

api/tests/unit/mappers/test_unit_mappers_engine.py
```diff
@@ -0,0 +1,463 @@
...
+        name="segment",
+        project=project,
+        feature=multivariate_feature,
+    )
+    FeatureSegment.objects.create(
+        feature=multivariate_feature,
+        segment=segment,
+        environment=environment,
+    )
+    return multivariate_feature.feature_states.first()
```
matthewelwell: Unless I'm mistaken, I don't think this fixture is returning what you're expecting it to. There are 2 reasons why this is the case: 

 1. A `FeatureSegment` model instance on it's own does nothing, you need to create a `FeatureState` linked to the `FeatureSegment`
 2. `multivariate_feature.feature_states.first()` will always return the environment default, since the default ordering is by id and the env default feature state is always created first.

api/tests/unit/mappers/test_unit_mappers_engine.py
```diff
@@ -0,0 +1,463 @@
...
+    )
+    expected_result = FeatureStateModel(
+        feature=FeatureModel(
+            id=multivariate_feature.id,
+            name="feature",
+            type="MULTIVARIATE",
+        ),
+        enabled=False,
+        django_id=segment_multivariate_feature_state.id,
+        feature_segment=None,
```
matthewelwell: As per my comment above. I think there's something slightly off with this test. We seem to be trying to test some segment related behaviour (as well as MV?) but are then asserting that we get None for the feature segment.

api/mappers/engine.py
khvn26: UPD: never mind, the `isinstance` way is not feasible ATM because of circular imports. Reverted to `getattr`.

api/tests/unit/mappers/test_unit_mappers_engine.py
```diff
@@ -0,0 +1,463 @@
...
+    result = engine.map_feature_state_to_engine(
+        segment_multivariate_feature_state,
+        mv_fs_values,
+    )
+
+    # Then
+    assert result == expected_result
+
+
+@pytest.mark.xfail(reason="https://github.com/Flagsmith/flagsmith/issues/2295")
```
matthewelwell: This issue is now closed as resolved, can we remove this decorator?

api/tests/unit/mappers/test_unit_mappers_engine.py
```diff
@@ -0,0 +1,463 @@
...
+    environment: Environment,
+    identity: "Identity",
+    feature: "Feature",
+    trait: "Trait",
+    identity_featurestate: FeatureState,
+    mocker: MockerFixture,
+) -> None:
+    # Given
+    environment_api_key = environment.api_key
+    expected_result = IdentityModel.construct(
```
matthewelwell: what's the difference between `.construct()` and just calling the default `__init__()` method here?

api/mappers/engine.py
matthewelwell: Ah, ok - yep, this all makes sense.

api/tests/unit/environments/test_environments_views_sdk_environment.py
khvn26: Yes we do — we weren't joining all of the integrations, see [this change](https://github.com/Flagsmith/flagsmith/blame/6c3ecc65be20792f06d6cd752e4e5d8d403fafe2/api/environments/managers.py#L17-L18).

api/tests/unit/environments/test_unit_environments_models.py
khvn26: See [this comment](https://github.com/Flagsmith/flagsmith/pull/2290#discussion_r1242489443).

api/edge_api/identities/models.py
matthewelwell: Ok, I think this is fairly minor for now - let's see where we're at once all this is merged.

api/tests/unit/mappers/test_unit_mappers_dynamodb.py
khvn26: Fair enough. I'll address this comment along with [private method testing concerns](https://github.com/Flagsmith/flagsmith/pull/2290#discussion_r1242474066) and the [mocking data concern](https://github.com/Flagsmith/flagsmith/pull/2290#discussion_r1242468481) by rewriting these tests to use real inputs.

api/tests/unit/mappers/test_unit_mappers_dynamodb.py
khvn26: This has been fixed.

api/tests/unit/mappers/test_unit_mappers_dynamodb.py
khvn26: Done.

api/tests/unit/mappers/test_unit_mappers_dynamodb.py
khvn26: Done. This is now tested implicitly in the tests for public functions.

api/tests/unit/mappers/test_unit_mappers_engine.py
khvn26: This is used to be able to pass `mocker.ANY` as the `identity_uuid` attribute value for the expected result model. If we use `__init__`, Pydantic validation won't allow this.

api/tests/unit/mappers/test_unit_mappers_engine.py
khvn26: Fixed.

api/tests/unit/mappers/test_unit_mappers_engine.py
khvn26: Thanks! Fixed (and test name modified to better reflect the tested behaviour).

api/tests/unit/mappers/test_unit_mappers_engine.py
khvn26: Done.

api/edge_api/identities/serializers.py
khvn26: > I guess the alternative would be to create a mappers module within each app

However, some of the mappers are used by several apps, so a common mapper module is still inevitable.

api/tests/unit/mappers/test_unit_mappers_engine.py
matthewelwell: Maybe worth a comment?

api/tests/integration/edge_api/identities/test_edge_identity_featurestates_viewset.py
```diff
@@ -694,10 +694,7 @@ def test_edge_identities_post_returns_400_for_invalid_mvfs_allocation(
         identity_uuid
     )
     assert response.status_code == status.HTTP_400_BAD_REQUEST
-    assert (
-        response.json()["multivariate_feature_state_values"]
-        == "Total percentage allocation for feature must be less than 100 percent"
-    )
+    assert "multivariate_feature_state_values" in response.json()
```
matthewelwell, line 697: How come this was changed?

api/edge_api/identities/serializers.py
matthewelwell: @khvn26 I think this (slightly annoying) question is still outstanding? Be good to just get it wrapped up.

api/tests/integration/edge_api/identities/test_edge_identity_featurestates_viewset.py
khvn26: The message was changed in flagsmith-engine and this broke the test. I decided not to test the library code.

api/edge_api/identities/serializers.py
khvn26: Moved the module under the `util` module; it's `util.mappers` now.

matthewelwell (approved):

PR #2297, opened 2023-06-12
Feature/show chargebee email in org admin

### LINKED ISSUE #2288: Show the Chargebee Account login email address in the Org Settings screen
Often customers change accounts/owners of the billing in Chargebee but then lose track of who has the account login. We should surface this email address on the Organisation settings page.

We should be able to retrieve this data from the Chargebee API.

dabeeeenster:
Implements https://github.com/Flagsmith/flagsmith/issues/2288

Couldnt decide whether it was better to store this in the `OrganisationSubscriptionInformationCache` or the `Subscription` opted for the former.


dabeeeenster:
@matthewelwell this is almost complete - not sure how to really exercise things in the tests tho?

api/organisations/subscriptions/metadata.py
```diff
@@ -33,14 +35,19 @@ def __add__(self, other: "BaseSubscriptionMetadata"):
...
-            self.api_calls,
-            str(self.projects) if self.projects is not None else "no limit",
+        return (
+            "%s Subscription Metadata (seats: %d, api_calls: %d, projects: %s, chargebee_email: %s)"
+            % (
+                self.payment_source.title(),
+                self.seats,
+                self.api_calls,
+                self.chargebee_email,
+                str(self.projects) if self.projects is not None else "no limit",
```
matthewelwell:
```suggestion
                str(self.projects) if self.projects is not None else "no limit",
                self.chargebee_email,
```

api/sales_dashboard/templates/sales_dashboard/organisation.html
```diff
@@ -42,7 +42,7 @@ <h2>API Calls</h2>
                 </tbody>
             </table>
 
-            <h2>Subscription</h2>
+            <h2>Subscription - <a href="mailto:{{ chargebee_email }}">{{ chargebee_email }}</a></h2>
```
matthewelwell: Maybe not an issue because we should always have a value for `chargebee_email` but I'm not sure how this will handle it not existing. `mailto:` might bork.

matthewelwell (commented):
Re testing, you should just look for existing unit tests that cover `get_subscription_metadata` and update them to verify that the chargebee_email attribute is handled correctly. Looks like something is already broken from the changes you've made.

api/sales_dashboard/templates/sales_dashboard/organisation.html
dabeeeenster: Fixed

api/tests/unit/organisations/chargebee/test_unit_chargebee_chargebee.py
```diff
@@ -44,12 +44,14 @@ def __init__(
         plan_id="plan-id",
         created_at=datetime.utcnow(),
         customer_id="customer-id",
+        chargebee_email="chargebee-email",
```
matthewelwell: this argument should really be `customer_email`

matthewelwell (changes requested):

api/tests/unit/organisations/chargebee/test_unit_chargebee_chargebee.py
dabeeeenster: I'm following the `customer-id` and `plan-id` as above?

dabeeeenster: OK fixed

matthewelwell: Actually fixed now...

matthewelwell (approved):

PR #2302, opened 2023-06-14
chore/more tolerant webhook urls

### LINKED ISSUE #2262: Allow webhooks to be created even if they use a domain like https://benisskill/webhook
Currently you have to have a webhook service up and running in order to create the actual webhook on the Flagsmith side. We should change this to allow users to create the webhook, but warn them on creation that the service is currently not reachable on their side. 

image

dabeeeenster:
URLField has validators that only allow for public URLs. Some of our customers want to define webhooks with URLs like `https://webhook-processor/stream` or similar. This PR allows for that. 

URLFields are just `varchar(200)` with some django validation so the migrations here are NOOP on the actual DB.


dabeeeenster:
Weird not sure why these test are failing?

matthewelwell:
> Weird not sure why these test are failing?

Because you still have the migration file from your other branch. 

```
django.db.migrations.exceptions.NodeNotFoundError: Migration organisations.0041_alter_organisationwebhook_url dependencies reference nonexistent parent node ('organisations', '0040_organisationsubscriptioninformationcache_chargebee_email')
```

`0040_organisationsubscriptioninformationcache_chargebee_email` does not exist in this branch (or the main branch).

dabeeeenster:
OK this is green now @matthewelwell 

> > Weird not sure why these test are failing?
> 
> Because you still have the migration file from your other branch.
> 
> ```
> django.db.migrations.exceptions.NodeNotFoundError: Migration organisations.0041_alter_organisationwebhook_url dependencies reference nonexistent parent node ('organisations', '0040_organisationsubscriptioninformationcache_chargebee_email')
> ```
> 
> `0040_organisationsubscriptioninformationcache_chargebee_email` does not exist in this branch (or the main branch).

matthewelwell (approved):

PR #2331, opened 2023-06-23
Fix buttons styles

aliakseilatyp:

- [ ] I have run [`pre-commit`](https://docs.flagsmith.com/deployment/locally-api#pre-commit) to check linting

## Changes

⋅ added common button props to _new-variables
⋅ fixed modals

_Please describe._

### DIFF
--- a/frontend/web/components/AdminAPIKeys.js
+++ b/frontend/web/components/AdminAPIKeys.js
@@ -87,7 +87,8 @@ export class CreateAPIKey extends PureComponent {
                   <Button
                     disabled={!this.state.expiry_date}
                     onClick={() => this.setState({ expiry_date: null })}
-                    className='btn-danger'
+                    theme='secondary'
+                    size='small'
                   >
                     Clear
                   </Button>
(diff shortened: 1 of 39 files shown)


kyle-ssg:
Labels - old boostrap added a margin bottom of $spacer 

image


List Item Links - they seem to be using the button link component which has a lighter font weight, these are normally bold 
image

These were bold before, looks a bit strange with just text
image


Outline button is 50px in height, regular is 48 - probably need a border on all buttons and to reduce the height by 2px

image
Looks strange, perhaps make this btn outline

image

Edit segment looks a bit wrong here, maybe just make this the grey button

Change request layout seems to have broken
image

Change request modal has lost its slide in animation

Footer links are right aligned
image

These were centered 
image

Org settings > Keys - layout broke
image


Clear filters looks incorrect, maybe replace with btn-outline
image


Looking at designs, buttons inline with inputs are 40px height - I guess we need a btn-sm 

**Org Settings**
image
**Environment Settings**
image
**Org Settings**
image
**Account Settings > keys**
image
**Environment Settings > keys**
image
This one too but lets make this btn-secondary whilst we are at it, danger makes no sense
**Org Settings > keys > Create**
image

PR #2341, opened 2023-06-26
Fix modal close intercept

kyle-ssg:

## Changes

The bootstrap changes introduced a regression where confirmation modals would close the create feature modal regardless of the user's input and re-show the confirmation dialog, this happened for 2 reasons

- Closing a confirmation dialog would trigger the modal intercept as if the original modal was attempting to be closed.
- The modal would be dismissed regardless of the result of the close intercept, it should only close the modal if the intercept resolves with true.

image

Confirm / cancel are you sure dialog from the edit feature modal.

image


novakzaballa:
LGTM!

novakzaballa (approved):

PR #2347, opened 2023-06-27
feature/auto label prs

dabeeeenster:
Auto-labels prs based on changed files using https://github.com/actions/labeler


matthewelwell (approved):
I think it would be good to add `infrastructure` here but it looks fine otherwise.

PR #2356, opened 2023-06-28
Change feature type when multivariates are introduced / removed

kyle-ssg:

## Changes

When a feature changes to either having or not having multivariate_options, we need to update the type to either MULTIVARIATE or STANDARD. This will now introduce a put to the feature when we detect this change.

Added first option and saved, removed last option and saved.


kyle-ssg:
Closed as we are doing on the BE now

PR #2358, opened 2023-06-28
fix(feature-type): Update feature_type (if required) on mv options updates

gagantrivedi:

## Changes

Add hooks on MV options model to feature type if needed

Adds unit tests


api/tests/unit/features/multivariate/test_unit_multivariate_models.py
```diff
@@ -38,7 +39,7 @@ def test_multivariate_feature_option_get_delete_log_message_for_deleted_feature(
 ):
     # Given
     mvfo = MultivariateFeatureOption.objects.create(feature=feature, string_value="foo")
-    feature.delete()
+    mvfo.feature.delete()
```
matthewelwell: curious to know why this has changed?

api/tests/unit/features/test_migrations.py
```diff
@@ -192,3 +194,46 @@ def test_revert_feature_state_versioning_migrations(migrator):
     assert not NewFeatureState.objects.filter(id=v1.id).exists()
     assert not NewFeatureState.objects.filter(id=v3.id).exists()
     assert NewFeatureState.objects.filter(id=v2.id).exists()
+
+
+def test_fix_feature_type_migration(migrator):
```
matthewelwell, line 203: This should be decorated with the SKIP_MIGRATION_TESTS

api/tests/unit/features/multivariate/test_unit_multivariate_models.py
gagantrivedi: It's because this [hook](https://github.com/Flagsmith/flagsmith/pull/2358/files#diff-e48d5e2e19f8c9f0d2d868206fe36a2780573ee3e6828bad61cf0ee4c304125fR64) populates the `self.feature` which happens before the feature is deleted

api/tests/unit/features/test_migrations.py
gagantrivedi: Yep, good catch

api/tests/unit/features/multivariate/test_unit_multivariate_models.py
matthewelwell: I think we've done a bit of both here, right? We don't actually need the `mvfo.refresh_from_db()` if we're deleting the feature on the `mvfo` object directly. 

The code should either be: 

```python
# <comment explaining why we refresh from db>
mvfo.refresh_from_db()
feature.delete()
```

or

```python
# <comment explaining why we delete the feature from the mvfo object directly>
mvfo.feature.delete()
```

Seems like you've done a little bit of both here which adds a bit of extra confusion?

matthewelwell: (sorry, I think I was a bit too quick with my comment 😄 )

matthewelwell (approved):

api/tests/unit/features/multivariate/test_unit_multivariate_models.py
gagantrivedi: > I think we've done a bit of both here, right? We don't actually need the `mvfo.refresh_from_db()` if we're deleting the feature on the `mvfo` object directly.
> 
> The code should either be:
> 
> ```python
> # <comment explaining why we refresh from db>
> mvfo.refresh_from_db()
> feature.delete()
> ```
> 
> or
> 
> ```python
> # <comment explaining why we delete the feature from the mvfo object directly>
> mvfo.feature.delete()
> ```
> 
> Seems like you've done a little bit of both here which adds a bit of extra confusion?

:facepalm:

api/features/multivariate/models.py
```diff
@@ -59,6 +61,20 @@ def create_multivariate_feature_state_values(self):
                 percentage_allocation=self.default_percentage_allocation,
             )
 
+    @hook(AFTER_CREATE)
+    def make_feature_multivariate(self):
+        # We can't do this check using `when` and `is_not`
```
matthewelwell: This comment isn't _super_ clear. Is there some piece of documentation that we can link to? I assume it's related to triggering across the join? So, something in [these docs](https://rsinger86.github.io/django-lifecycle/fk_changes/) might be the issue?

api/tests/unit/features/multivariate/test_unit_multivariate_models.py
```diff
@@ -38,6 +39,9 @@ def test_multivariate_feature_option_get_delete_log_message_for_deleted_feature(
 ):
     # Given
     mvfo = MultivariateFeatureOption.objects.create(feature=feature, string_value="foo")
+    # clear MultivariateFeatureOption().feature populated by `AFTER_CREATE` hook
```
matthewelwell: I think we can actually be more explicit here: 

```python
# Since the `AFTER_CREATE` hook on MultivariateFeatureOption mutates 
# mvfo.feature directly, we need to make sure that we have the latest
# changes in order for the `AFTER_DELETE` hook to behave correctly. 
```

api/features/multivariate/models.py
gagantrivedi: It's not really an issue, though - I don't think. It's just the way that this works,i.e: capturing the state on __init__ 
e.g: if you create 4 instances of this, it will 4 calls

api/tests/unit/features/multivariate/test_unit_multivariate_models.py
gagantrivedi: Yeah, cool

api/features/multivariate/models.py
matthewelwell: But, surely that's the case with the code we've got here as well, right? It's always going to check across the join?

gagantrivedi: No, in our case it will only happen after_create

matthewelwell: Ah, interesting. Then yeah, django-lifecycle should definitely optimise that! I think the comment could be clearer then though to explicitly state _why_ we're doing it: 

```python
# Handle the check on feature.type in the method itself to ensure this is
# only performed after create. Using `when` and `is_not` means 
# LifecycleModel.__init__() queries for feature on every object init. 
```

matthewelwell (approved):

PR #2363, opened 2023-06-29
fix: don't use broken identity document builder

khvn26:

## Changes

This PR fixes the internal server error when invoking `/api/v1/environments/{environment_id}/edge-identities`

### DIFF
--- a/api/edge_api/identities/serializers.py
+++ b/api/edge_api/identities/serializers.py
@@ -12,7 +12,6 @@ from flag_engine.features.models import (
 from flag_engine.features.models import (
     MultivariateFeatureStateValueModel as EngineMultivariateFeatureStateValueModel,
 )
-from flag_engine.identities.builders import build_identity_dict
 from flag_engine.identities.models import IdentityModel as EngineIdentity
 from flag_engine.utils.exceptions import DuplicateFeatureState
 from pydantic import ValidationError as PydanticValidationError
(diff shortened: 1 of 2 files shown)


api/edge_api/identities/serializers.py
```diff
@@ -45,7 +48,9 @@ def save(self, **kwargs):
             raise ValidationError(
                 f"Identity with identifier: {identifier} already exists"
             )
-        EdgeIdentity.dynamo_wrapper.put_item(build_identity_dict(self.instance))
+        EdgeIdentity.dynamo_wrapper.put_item(
+            map_engine_identity_to_identity_document(self.instance)
```
matthewelwell, line 52: I'm keen to understand why we had to change this? My understanding was that we were keeping the `build_identity_dict` functions for backwards compatibility, but if we've broken them - are there any issues elsewhere? e.g. python client?

khvn26: > but if we've broken them

We have, sadly. However, I've checked all the Python code we have and `build_identity_dict` has only been used in Core API. Those two calls are the last ones. I suggest we deprecate `build_identity_dict` completely after we eliminate them.

matthewelwell: Hmm... do we know why it's broken? Is it an easy fix? Just conscious we already released the major version so I guess our options are: 

1. Fix it
2. Release another major version to remove it entirely
3. (Slightly skanky) Remove it from the current major version

khvn26: It's broken because Marshmallow dumps datetimes to ISO-formatted string, while Pydantic's `.dict()` method doesn't.

I'm not in favour of option 1 because the functionality is only needed for Core API. 

Option 3 is dodgy but we did promise to deprecate the `builders` modules in 4.0.0 release notes. Maybe we could get away with a minor release.

matthewelwell (approved):

api/edge_api/identities/serializers.py
khvn26: I suggest we move this over to https://github.com/Flagsmith/flagsmith-engine/issues/175

PR #2369, opened 2023-06-30
Fix swagger generation

matthewelwell:
## Changes

1. Replace unsupported [drf-yasg2](https://pypi.org/project/drf-yasg2/) with [drf-yasg](https://pypi.org/project/drf-yasg/)
2. Fix warnings and exceptions in swagger generation process

Note: there are some warnings spat to the logs still in the django-trench views. We can either submit a PR to their repo (although it's been dormant since 2022) or fork the repo (preferred). 

Ran the application locally and verified that swagger docs are generated successfully and without warnings (in our own views at least).

### DIFF
--- a/api/api/urls/deprecated.py
+++ b/api/api/urls/deprecated.py
@@ -8,9 +8,9 @@ app_name = "deprecated"
 
 urlpatterns = [
     url(
-        r"^identities/(?P<identifier>[-\w@%.]+)/traits/(?P<trait_key>[-\w.]+)",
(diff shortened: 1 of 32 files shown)


gagantrivedi (approved):
Maybe we can create an issue and ask if they would be open to a PR - we can fork if we don't get any response

matthewelwell:
> Maybe we can create an issue and ask if they would be open to a PR - we can fork if we don't get any response

Seems like it's fixed in the latest release but it's a breaking change and drops support for djoser so the upgrade will be painful. 

Relevant issues: 
 * https://github.com/merixstudio/django-trench/issues/133
 * https://github.com/merixstudio/django-trench/issues/95

PR #2378, opened 2023-07-04
Ensure versions are taken into account when building environment document

matthewelwell:
## Changes

Ensures that feature state versions are taken into account for environment feature states when building the environment document. 

Updating existing unit test.


khvn26 (approved):

matthewelwell:
Merging here despite E2E failures as they are deemed unrelated. The E2E passed for docker-compose but failed twice against local API. These were just related to Github performance issues affecting our tests rather than any change made in this PR.

PR #2382, opened 2023-07-05
feat: Feature Versioning V2

matthewelwell:
## Changes

Adds logic for a 'version 2' of feature versioning. The way it behaves is such that every feature, environment combination has a relationship to a new `EnvironmentFeatureVersion` entity. This new entity includes a `sha` attribute, much like a commit in git. Using this, it is possible to show a history of a given feature and to roll back as needed. 

Still TODO: 

 - ~Add parameter to automatically publish the version~
   - This is actually not the desired behaviour. We are sticking to multiple endpoints for now. 
 - ~Update change requests logic so that a change request can point to multiple environment versions~
   - To be done in a separate PR. 
 - ~Ensure that environment document rebuilds are triggered when a new version is published~
 - ~Add AuditLog records~
   - For now, audit log records will not be written for changes to a feature version - since the feature history will suffice. 
 - ~Multivariate?~
   - This needs discussion. The current logic in this PR matches that of the current change request logic. 
 - ~Exports?~
 - ~Webhooks~
   - To be done in a separate PR

TODO: 
 - Open separate PRs for: 
   - Webhooks
   - ~Change requests~

## How does it work?

Instead of directly making a change to a feature in a given environment, the expectation is that, when a first change is made, the FE makes a request to create a new version. It can include the initial change in the request to create the version so that it saves the change set as a 'draft'. Users* can then add further changes to the version (e.g. segment overrides) as they please. Once they're happy with the changes, they can publish it (TODO: handle change requests). 

By using this workflow, all users can then see a version history of a feature in a given environment. They can also trivially roll back to a previous version as needed. 

My current expectation is that users will also be able to add multiple unpublished versions to a change request. 

Current thinking is that this behaviour will be 'opt-in' for existing environments but default for new environments. This is so that we have a trigger for creating the initial version. This will need to be changed in the future so that new environments are defaulted to version 2, however, I imagine this will also require a substantial change to a lot of our tests. 

## Front-end Integration

These changes are currently restricted to environments which have `use_v2_feature_versioning: true` - this can be set with a PUT / PATCH request to the environment detail endpoint. 

The expectation is that the frontend will hit the following endpoints now to make a change to a flag. 

```
POST /api/v1/environments/:id/features/:id/versions
```

This will return a `sha` attribute and will ensure that all of the feature states from the previous version are transferred to the new version. To make a change to a given feature state, or to add a new segment override, the FE will need to make requests such as

```
POST /api/v1/environments/:id/features/:id/versions/:sha/feature-states/
PUT /api/v1/environments/:id/features/:id/versions/:sha/feature-states/:id/
```

Once the changes are complete, the FE will need to hit the following endpoint to publish the change: 

```
POST /api/v1/environments/:id/features/:id/versions/:sha/publish/
```

This endpoint does take an optional `live_from` parameter, but this will need more work to ensure that the current scheduling UI ties in correctly. 

Considerations: 

 1. Can we create an endpoint that takes the new state and value for an environment default and automatically creates and publishes a version (to simplify the simple workflow)
 2. Should we have an endpoint like: `GET /api/v1/environments/:id/features/:id/versions/latest/`

Added one long integration test and updated existing tests where necessary.

TODO: 
 - Test with Multivariate
 - Test with identity overrides
 - Test with multiple segment overrides


api/tests/unit/features/test_unit_features_views.py
```diff
@@ -55,7 +55,7 @@ def test_list_feature_states_from_simple_view_set(
     )
 
     # When
-    with django_assert_num_queries(8):
+    with django_assert_num_queries(9):
```
matthewelwell, line 952: This is due to the addition of the `get_object_or_404` function call here: https://github.com/Flagsmith/flagsmith/pull/2382/files#diff-9e42c40eece4b1da492acaad6b58480687eab073351781b6a0fec0a91435df9aR543

This was necessary to simplify the code in the `versioning` module.

api/features/versioning/signals.py
```diff
@@ -0,0 +1,57 @@
...
+
+@receiver(pre_save, sender=EnvironmentFeatureVersion)
+def update_live_from(instance: EnvironmentFeatureVersion, **kwargs):
+    if instance.published and not instance.live_from:
+        instance.live_from = timezone.now()
+
+
+@receiver(post_init, sender=EnvironmentFeatureVersion)
+def cache_fields(instance: EnvironmentFeatureVersion, **kwargs):
+    instance._init_fields = {"published": instance.published}
```
matthewelwell: We're essentially implementing our own version of LifecycleModel here, to avoid circular import issues. Right now, we only care about `published` but we can extend it as needed later on.

matthewelwell:
Note: tests are failing until we merge [this](https://github.com/Flagsmith/flagsmith/pull/2614) PR.

api/core/models.py
```diff
@@ -81,6 +81,9 @@ class _AbstractBaseAuditableModel(models.Model):
     class Meta:
         abstract = True
 
+    def get_skip_create_audit_log(self) -> bool:
```
gagantrivedi, line 85: Should this just be a property?

api/features/features_service.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def get_overrides_data(
+    environment: "Environment",
+) -> typing.Dict[int, EnvironmentFeatureOverridesData]:
+    """
+    Get the number of identity / segment overrides in a given environment for each feature in the
+    project.
+
+    :param environment_id: the id of the environment to get the overrides data for
```
gagantrivedi: This should be `environment`, right?

api/features/features_service.py
```diff
@@ -0,0 +1,33 @@
...
+        env_feature_overrides_data = all_overrides_data.setdefault(
+            feature_state.feature_id, EnvironmentFeatureOverridesData()
+        )
+        if feature_state.feature_segment_id:
+            env_feature_overrides_data.num_segment_overrides += 1
+        elif feature_state.identity_id:
+            env_feature_overrides_data.add_identity_override()
+        all_overrides_data[feature_state.feature_id] = env_feature_overrides_data
+
+    return all_overrides_data  # noqa
```
gagantrivedi: Why #noqa?  What weird rule does this violate?

api/features/versioning/signals.py
```diff
@@ -0,0 +1,57 @@
...
+    for feature_state in previous_environment_feature_version.feature_states.all():
+        feature_state.clone(
+            env=instance.environment, environment_feature_version=instance
+        )
+
+
+@receiver(pre_save, sender=EnvironmentFeatureVersion)
+def generate_sha(instance: EnvironmentFeatureVersion, **kwargs):
+    if not instance.sha:
+        instance.sha = hashlib.sha256(
```
gagantrivedi: Is there a reason why we are using sha instead of UUID?

api/features/versioning/views.py
```diff
@@ -0,0 +1,162 @@
...
+    def initial(self, request: Request, *args, **kwargs):
+        super().initial(request, *args, **kwargs)
+        environment_feature_version = get_object_or_404(
+            EnvironmentFeatureVersion, sha=self.kwargs["environment_feature_version_pk"]
+        )
+        if self.action != "list" and environment_feature_version.published is True:
+            raise FeatureVersionDeleteError("Cannot modify published version.")
+
+        # patch the objects onto the request
+        request.environment_feature_version = environment_feature_version
```
gagantrivedi: What are your thoughts on adding these objects to `self` instead of request if they are just being used by the view? Is patching request, a convention for sharing data between multiple methods in a view class? I think request having a large surface area that goes beyond the view makes this a bit hard to track?

gagantrivedi (changes requested):

api/core/models.py
matthewelwell: I wanted to avoid using a property here because it might include querying the database and I don't think that properties should do that.

api/features/features_service.py
matthewelwell: Yes, good spot.

api/features/features_service.py
matthewelwell: Good question - it's actually just a bug in pycharm: https://youtrack.jetbrains.com/issue/PY-39122/Django-Type-of-ForeignKeys-id-is-detected-as-Typeint-istead-of-int

I'll just remove the noqa and ignore my IDE warning for now.

api/features/versioning/signals.py
matthewelwell: Not really, more just for aesthetics I suppose. I wanted it to be analogous to a git commit history. Happy to change it if you'd prefer.

api/features/versioning/views.py
matthewelwell: Yeah, interesting, I switched from `self` to `request` because I think they were being used in the serializer too. The serializer does technically have access to the view as well via `self.context` but it just looks a bit uglier then: 

```python
self.context["view"].environment
```

Perhaps you're right though since the request is much wider reaching.

api/features/versioning/signals.py
gagantrivedi: Yeah, I think I'd prefer UUID, sha made me think there might be some spicy code coming :smile:

api/features/versioning/views.py
gagantrivedi: think we are adding them to the context using `  def get_serializer_context(self):`, no?

api/features/versioning/signals.py
matthewelwell: Cool, yep, done.

api/features/versioning/views.py
matthewelwell: Ah yes, this was why I changed it: 

image

I've moved them to the view itself but, to satisfy my IDE (and I guess python conventions), I have also overridden the `__init__` method to set them to `None`.

gagantrivedi: Oh, interesting...

api/features/versioning/versioning.py
```diff
@@ -0,0 +1,92 @@
+import typing
```
gagantrivedi: Should this module be renamed to versioning_service?

gagantrivedi (approved):
One minor comment

api/features/versioning/versioning.py
matthewelwell: Yeah, that's a fair comment, Have done so.

matthewelwell:
Closed and reopened to attempt to resolve issues deploying Uffizzi environment.

Issue #2030, opened 2023-03-15
Committing scheduled change requests in different order can affect scheduling


matthewelwell:
**Describe the bug**

Committing 2 change requests which are scheduled for the future in the wrong order (i.e. CR1 is scheduled for 1h from now, CR2 is scheduled 2h from now but CR2 is committed before CR1) causes the earlier scheduled feature state to be used instead of the later one. 

**To Reproduce**

1. Create 2 scheduled change requests, one for 2m from now (CR1), another for 3m from now (CR2)
2. Approve / commit CR2 then approve / commit CR1
3. Wait until the scheduled date/time has passed for both CRs
4. Notice that the value provided in CR1 is given when flags are retrieved for the environment.

**Expected behavior**

The value provided in CR2 is given when flags are retrieved for the environment. 

**Screenshots**

N/a

**How are you running Flagsmith?**

- [ ] Self Hosted with Docker
- [ ] Self Hosted with Kubernetes
- [ ] SaaS at flagsmith.com
- [ ] Some other way
- [x] N/A

**Additional context**

matthewelwell:
The fix that was released as part of this issue works if the feature states are created in a given order, but not in every order. More information to follow.

matthewelwell:
The fix applied as part of this issue (in [this PR](https://github.com/Flagsmith/flagsmith/pull/2036)) only worked in the scenario where the scheduled change requests were created in a different order than they were published. 

_example that was fixed by this PR_: 

Create CR (#CR1) scheduled for 2 days from now
Create CR (#CR2) scheduled for 1 day from now
Publish CR1
Publish CR2

_example that was not fixed by this PR_:

Create CR (#CR1) scheduled for 1 days from now
Create CR (#CR2) scheduled for 2 day from now
Publish CR2
Publish CR1

The example that was not fixed by this PR was not an issue at the time because of the way in which the feature states were filtered. When we released the change to use the mappers in the Core API repository, this was changed slightly. Previously the logic was (shortened for brevity): 

```python
existing_feature_state = features_map.get(fs.feature_id)
if not existing_feature_state or fs > existing_feature_state:
    features_map[fs.feature_id] = fs
```

but this was changed in the mappers to: 

```python
if existing_feature_state := prioritised_feature_state_by_feature_id.get(
    feature_state.feature_id
):
    if existing_feature_state > feature_state:
        continue
prioritised_feature_state_by_feature_id[
    feature_state.feature_id
] = feature_state
[...]

PR #2390, opened 2023-07-06
Fix environment document generation for scheduled CRs

matthewelwell:
## Changes

Fixes issue with evaluating versions of the same type. 

Description of the issue that this fixes is described in more detail in a comment [here](https://github.com/Flagsmith/flagsmith/issues/2030#issuecomment-1623629156). 

This change ensures that we rely on the version only if the `live_from` date is the same, otherwise, we rely on the `live_from` date. 

As per the [previous PR](https://github.com/Flagsmith/flagsmith/pull/2036) which attempted to fix this issue, we use `live_from` to determine which flag over the version in most scenarios.

Added a specific test for the issue as reported by a customer. Also updated unit tests where necessary. 

**Some manual tests that we should run in a live environment:** 

Test 1

Create & publish a change request (CR1), scheduled for some time in the future
Create & publish a change request (CR2), scheduled to go live immediately

_Observe that we receive the value from CR2 until the time that CR1 was scheduled for_

Test 2

Create a change request (CR1) scheduled for 5 minutes from now
Create a change request (CR2) scheduled for 2 minutes from now
Publish CR1
Publish CR2

_Observe that we receive the value from CR2 after 2 minutes have passed, until the time that CR1 was scheduled for_ 

Test 3

Create CR (#CR1) scheduled for 2 minutes from now
Create CR (#CR2) scheduled for 5 minutes from now
Publish CR2
Publish CR1

_Observe that we receive the value from CR1 after 2 minutes have passed, until the time that CR2 was scheduled for_

### DIFF
--- a/api/features/models.py
+++ b/api/features/models.py
@@ -485,10 +485,11 @@ class FeatureState(
             # further in the future can be lower than a feature state
             # whose live_from value is earlier.
             # See: https://github.com/Flagsmith/flagsmith/issues/2030
-            is_more_recent_live_from = self.is_more_recent_live_from(other)
-            is_more_recent_version = self._is_more_recent_version(other)
-            return self.version is not None and (
-                is_more_recent_live_from or is_more_recent_version
+            return (
+                self.is_live
+                and (not other.is_live or self.is_more_recent_live_from(other))
+                or self.live_from == other.live_from
+                and self._is_more_recent_version(other)
             )
 
         # if we've reached here, then self is just the environment default. In this case, other is higher priority if
@@ -685,12 +686,12 @@ class FeatureState(
             )
 
     @hook(BEFORE_CREATE)
-    def set_live_from_for_version_1(self):
+    def set_live_from(self):
         """
         Set the live_from date on newly created, version 1 feature states to maintain
         the previous behaviour.
         """
-        if self.version == 1 and not self.live_from:
+        if self.version is not None and self.live_from is None:
             self.live_from = timezone.now()
 
     @hook(AFTER_CREATE)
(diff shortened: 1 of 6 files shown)


khvn26 (approved):

api/features/models.py
```diff
@@ -485,10 +485,13 @@ def __gt__(self, other):
...
-            is_more_recent_version = self._is_more_recent_version(other)
-            return self.version is not None and (
-                is_more_recent_live_from or is_more_recent_version
+            return (
+                self.is_live
+                and (not other.is_live or self.is_more_recent_live_from(other))
+                or (
+                    self.live_from == other.live_from
+                    and self._is_more_recent_version(other)
+                )
```
khvn26: This is now a lot harder to read now, IMO. How about

```suggestion
            if self.is_live:
                return not other.is_live or self.is_more_recent_live_from(other)
            return self.live_from == other.live_from and self._is_more_recent_version(
                other
            )
```

matthewelwell: This is actually not the functionality that I was intending, but you're right that it matches what I had previously. 

The functionality I'm looking for is: 

```python
            return self.is_live and (
                (not other.is_live or self.is_more_recent_live_from(other))
                or (
                    self.live_from == other.live_from
                    and self._is_more_recent_version(other)
                )
            )
```

But perhaps the following is more readable? 

```python
if self.is_live:
    if not other.is_live or self.is_more_recent_live_from(other):
        return True
    elif (
       self.live_from == other.live_from
       and self._is_more_recent_version(other)
    ):
        return True
return False
```

khvn26: Looks good, let's use that.

PR #2397, opened 2023-07-07
refactor: remove husky

### LINKED ISSUE #2387: Review whether we need both husky and pre-commit
Currently the repository has configuration for both [husky](https://typicode.github.io/husky/) and [pre-commit](https://pre-commit.com/). We should review whether both of these are necessary. As far as I know, `pre-commit` can do everything that `husky` is currently doing but husky [seems to be a JS only package](https://github.com/typicode/husky/discussions/1094).

Some things to consider here: 

 * Would a global makefile be useful? 
 * Can we run a single version of this in a GH action to verify that the changes are conforming to the pre commit rules? e.g. `pre-commit --check`

dabeeeenster:
Moves everything from husky into pre-commit.


khvn26 (dismissed):

khvn26 (approved):

kyle-ssg:
Can we replace the pre commit before merging this please? It’s very useful

dabeeeenster:
> Can we replace the pre commit before merging this please? It’s very useful

Can you clarify exactly what you need/mean?

dabeeeenster:
@kyle-ssg it effectively just ran `npx lint-staged` - is that correct? What linters were running here?

kyle-ssg:
Husky + the following automatically lint / prettier fixes files about to be committed via the a pre-commit hook.

```
npx lint-staged --allow-empty
```

novakzaballa:
@dabeeeenster, @kyle-ssg @matthewelwell @khvn26 as mentioned in this thread in Slack https://flagsmith.slack.com/archives/CTF0THS2D/p1688633135062699, 
I think that we need to differentiate the coding time from the CI/CD check/deployment time, trying of course to not repeat ourselves and re-use as much code as possible. In dev time we need to run certain checks only on changed(staged) files, and we use `pre-commit` package for Python and Kyle uses Husky for the frontend. Husky uses a package.json file to work and Pre-commit is not well suited for monorepo as you can see here https://github.com/pre-commit/pre-commit/issues/466, it would force us to copy the linter/prettier config files from frontend to the root folder making our configuration a bit confusing and dirty. So my suggestion would be that for dev time checking we use the best/most-used tool in every case, I mean pre-commit for Python and Husky or another JS friendly hook, and in the root folder we can use the plain githooks as I did in this PR https://github.com/Flagsmith/flagsmith/pull/2403. Regarding the CI/CD checks you can use prebuilt actions, a make file, or run `pre-commit run --all-files` in the api folder and do something similar with the frontend.

dabeeeenster:
OK so now the current PR has all the config files in root of repo (I've merged in the ones that were in `frontend/`). I've also included eslint to the pre-commit setup. 

I think with eslint and prettier set up we are mimicing what lint-staged did but I couldnt find a definitive list of the linters it ran - any idea @kyle-ssg ?

I quite like this setup.

novakzaballa:
> OK so now the current PR has all the config files in root of repo (I've merged in the ones that were in `frontend/`). I've also included eslint to the pre-commit setup.
> 
> I think with eslint and prettier set up we are mimicing what lint-staged did but I couldnt find a definitive list of the linters it ran - any idea @kyle-ssg ?
> 
> I quite like this setup.

@dabeeeenster I can see that we are overlapping efforts, I will stop working on this PR https://github.com/Flagsmith/flagsmith/pull/2403/files since I don´t think it is compatible with what you are doing, since I am keeping Husky for the frontend, and the current `prettier:fix` script for docs for the dev time checks.

matthewelwell (dismissed):
This looks good to me but I'm keen to know what @kyle-ssg thinks. 

Also, we probably need to update the documentation somewhere for this?

matthewelwell:
@dabeeeenster I just ran `pre-commit run --all-files` and got 94 file changes...

dabeeeenster:
> @dabeeeenster I just ran `pre-commit run --all-files` and got 94 file changes...

Yep - this is all/mainly FE related tho? I don't think we've ever run full `prettier` etc over all the FE files. Is that right @kyle-ssg ?

dabeeeenster:
1 annoying aspect is that we need to do this for the `.flake8`:

image

as the linter is being run in the root folder.

PR #2403, opened 2023-07-10
ci: Move pre-commit to each sub-repo

novakzaballa:

## Changes

Use pure Githooks in root folder of the mono repo to run the hooks in api and frontend repos correspondingly.

Manually:

Run:

`git config core.hooksPath hooks`

Then make changes and commit to see both pre-commit tools running in api and frontend folders.

### DIFF
--- a/api/.pre-commit-config.yaml
+++ b/api/.pre-commit-config.yaml
@@ -1,0 +1,26 @@
+repos:
+  - repo: https://github.com/PyCQA/isort
+    rev: 5.12.0
+    hooks:
+      - id: isort
+        name: isort (python)
+  - repo: https://github.com/psf/black
+    rev: 23.1.0
+    hooks:
+      - id: black
+        language_version: python3
+        exclude: migrations
+  - repo: https://github.com/pycqa/flake8
+    rev: 6.0.0
+    hooks:
+      - id: flake8
+        name: flake8
+        args: [--config, api/.flake8]
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
+      - id: check-yaml
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: v3.0.0
+    hooks:
+      - id: prettier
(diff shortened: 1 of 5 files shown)


dabeeeenster:
Does this changes the docs in terms of how to set up pre-commit in the project?

novakzaballa:
> Does this changes the docs in terms of how to set up pre-commit in the project?

@dabeeeenster Sorry, I forgot to add the `docs` folder. Now I added the checks for the docs folder to the pre-commit script, running what I saw there in its package.json, is it ok?

```
npm run prettier:fix
```

matthewelwell:
I see the following: 

```
Starting pre-commit in 'api' folder:
isort (python).......................................(no files to check)Skipped
black................................................(no files to check)Skipped
flake8...............................................(no files to check)Skipped
check yaml...........................................(no files to check)Skipped
prettier.................................................................Passed
Starting pre-commit in 'frontend' folder:
.husky/pre-commit: line 2: .husky/_/husky.sh: No such file or directory
→ No staged files match any configured task.
```

Overall, this feels a bit hacky. It feels like there must be a better solution here, but I think I would prefer (over this), simply having to install the relevant pre-commit in each directory. This feels like it would become stale fast. 

Perhaps a global makefile could run each of the relevant pre-commit packages? [Relevant PR](https://github.com/Flagsmith/flagsmith/pull/2405)?

dabeeeenster:
I think we should close this in favour of https://github.com/Flagsmith/flagsmith/pull/2397 thoughts?

novakzaballa:
@dabeeeenster @matthewelwell, @kyle-ssg I think that there are a couple of differences:
- Here I added a Typescript check, I guess that could also easily added in pre-commit.
- Using this approach we don't need to copy or move the checks tooling configuration files to the root folder, which I think is cleaner. Some of those config files include `.eslintrsjs`,  `eslintignore`, `.prettierrc`, `tsconfig.json`.
- I guess, but I am not sure, that soon or later we will need or want another check to run in the context of the frontend folder, like E2E tests, or other checks.
-  I am afraid as well that pre-commit is mainly used for Python projects and it is not that used for JS projects, so it could not support everything we need now or in the future in the FE side.
Even If we want to use the `pre-commit` instead of Husky, I would suggest using this approach so we can run the tools in the context of each app folder so we don't need to have the configuration files at the root level.

dabeeeenster:
Closed in favour of a simpler life

PR #2409, opened 2023-07-11
Use larger runners for E2E

matthewelwell:
## Changes

Use 'larger runners' functionality for E2E tests.

Ran E2E workflows multiple times (TODO)


matthewelwell:
Despite my hopes that this would fix the issue, it hasn't... even with a runner that has 8 cores, 32GB RAM, we are still seeing the issue. Despite the obvious disappointment, this at least tells us that the issue is _not_ resource availability. This likely means there is something we can fix... we just need to find it.

PR #2417, opened 2023-07-12
refactor(docs): Subnav for integrations

dabeeeenster:
Splits out the integrations into subfolders as they were growing!

WARNING! When releasing this to prod we need to update the links in the `integration_data` production feature flag


kyle-ssg:
We could do this in a non-breaking way by adding a category field to the remote config items

dabeeeenster:
> We could do this in a non-breaking way by adding a category field to the remote config items

Might be nice to do this and then split out the integrations on the front end too - but simpler to just update the links regardless

kyle-ssg:
Ah, I see what you mean now - the "docs" properties, yep ok with you now

gagantrivedi (approved):

matthewelwell (approved):

PR #2422, opened 2023-07-12
feat: Use isEnterprise from endpoint version response to determine permissions

novakzaballa:

## Changes
Use `isEnterprise` attribute from API `/version` endpoint response to determine permissions.

Tested with both self-hosted version

### DIFF
--- a/frontend/common/utils/utils.tsx
+++ b/frontend/common/utils/utils.tsx
@@ -242,15 +242,15 @@ const Utils = Object.assign({}, require('./base/_utils'), {
     if (plan && plan.includes('start-up')) {
       return planNames.startup
     }
-    if (plan && plan.includes('enterprise')) {
+    if (
+      global.flagsmithVersion?.backend.is_enterprise ||
+      (plan && plan.includes('enterprise'))
+    ) {
       return planNames.enterprise
     }
     return planNames.free
   },
   getPlanPermission: (plan: string, permission: string) => {
-    if (global.flagsmithVersion?.backend.is_enterprise) {
-      return true
-    }
     let valid = true
     const planName = Utils.getPlanName(plan)
 
@@ -307,9 +307,6 @@ const Utils = Object.assign({}, require('./base/_utils'), {
   },
 
   getPlansPermission: (permission: string) => {
-    if (global.flagsmithVersion?.backend.is_enterprise) {
-      return true
-    }
     const isOrgPermission = permission !== '2FA'
     const plans = isOrgPermission
       ? AccountStore.getActiveOrgPlan()
--- a/frontend/common/utils/utils.tsx
+++ b/frontend/common/utils/utils.tsx
@@ -248,11 +248,12 @@ const Utils = Object.assign({}, require('./base/_utils'), {
     return planNames.free
   },
   getPlanPermission: (plan: string, permission: string) => {
-    let valid = true
-    const planName = Utils.getPlanName(plan)
-    if (!Utils.getFlagsmithHasFeature('plan_based_access')) {
+    if (global.flagsmithVersion?.backend.is_enterprise) {
       return true
     }
+    let valid = true
+    const planName = Utils.getPlanName(plan)
+
     if (!plan || planName === planNames.free) {
       return false
     }
@@ -306,7 +307,7 @@ const Utils = Object.assign({}, require('./base/_utils'), {
   },
 
   getPlansPermission: (permission: string) => {
-    if (!Utils.getFlagsmithHasFeature('plan_based_access')) {
+    if (global.flagsmithVersion?.backend.is_enterprise) {
       return true
     }
     const isOrgPermission = permission !== '2FA'


frontend/common/utils/utils.tsx
```diff
@@ -247,15 +247,22 @@ const Utils = Object.assign({}, require('./base/_utils'), {
...
     return planNames.free
   },
+
   getPlanPermission: (plan: string, permission: string) => {
-    let valid = true
-    const planName = Utils.getPlanName(plan)
-    if (!Utils.getFlagsmithHasFeature('plan_based_access')) {
+    if (
+      !Utils.getFlagsmithHasFeature('plan_based_access') &&
+      global.flagsmithVersion.is_enterprise
```
kyle-ssg: If I'm not mistaken this logic is incorrect, I think we allow features such as audit in open source

novakzaballa: Thank you @kyle-ssg. @matthewelwell, could you please confirm our business policy regarding the self-hosted free plan? If we have some features enabled for that case, I think we could return true for those in the same block of code.

matthewelwell: Audit shouldn't be allowed in open source (although it has been to this point). I think therefore this change is correct @kyle-ssg ?

matthewelwell:
@novakzaballa this is now released in v2.61.0 of the flagsmith images (flagsmith/flagsmith and flagsmith/flagsmith-private-cloud). You should be able to test it using those.

frontend/web/components/pages/OrganisationSettingsPage.js
```diff
@@ -268,6 +268,12 @@ const OrganisationSettingsPage = class extends Component {
     const verifySeatsLimit = Utils.getFlagsmithHasFeature(
       'verify_seats_limit_for_invite_links',
     )
+    const hasSelfHostedInvitePermission =
+      Utils.getPlansPermission('SELF_HOSTED_INVITE')
+    console.log(
+      'DEBUG: hasSelfHostedInvitePermission:',
+      hasSelfHostedInvitePermission,
+    )
```
matthewelwell: Can we remove this?

frontend/common/utils/utils.tsx
```diff
@@ -298,6 +306,10 @@ const Utils = Object.assign({}, require('./base/_utils'), {
         valid = isScaleupOrGreater
         break
       }
+      case 'SELF_HOSTED_INVITE': {
```
matthewelwell: Does the permission need to be `'SELF_HOSTED_INVITE'` here? Can't it just be `'INVITE'` ?

novakzaballa: Yes, done.

frontend/web/components/pages/OrganisationSettingsPage.js
novakzaballa: Sorry, I made the change but did not push it.

frontend/web/components/modals/InviteUsers.js
```diff
@@ -62,6 +62,7 @@ const InviteUsers = class extends Component {
   render() {
     const { invites } = this.state
     const hasRbacPermission = Utils.getPlansPermission('RBAC')
+    const hasSelfHostedInvitePermission = Utils.getPlansPermission('INVITE')
```
matthewelwell: This should also change to `hasInvitePermission` then, right?

Looks like it is defined in multiple places too.

novakzaballa: Yes, I'm doing some corrections that include this.

frontend/common/utils/utils.tsx
kyle-ssg: Ok, yeah after talking with Ben:

- Self hosted  = same as SaaS without a plan
- EE = same as enterprise plan

I think this means we can remove the plan_based_access flag entirely in this PR.

novakzaballa: @kyle-ssg @matthewelwell I have a couple of questions in this regard:

- In the case of self-hosted (free) AFAIK we should allow unlimited users (invites), while in SaaS free we only allow 1 user. Currently, in the self-hosted free version, we allow unlimited invites for new users, which can be admins or regular users since RBAC is enabled. After these changes, we will only allow admins, since RBAC will be disabled. Is that OK?
- If that’s the case should we change this description https://github.com/Flagsmith/self-hosted/blob/ed4a044ddf165e7ec5017cb0dc53e661c3e04b6e/README.md?plain=1#L27 ?
- Other behaviors, like limiting the number Projects to only 1, and in general all the non-free plan features that were enabled before for self-hosted free will no longer be available, I guess we are aware of all these changes right?

matthewelwell (approved):
This looks good to me. 

I'm keen to get @kyle-ssg's opinion here too. 

Also, one thing this has highlighted is that this means that all organisations in a given enterprise installation will have all functionality. Previously these controls were implemented on a per organisation basis via subscriptions that users created in their own installations. Keen to get @dabeeeenster's thoughts here?

dabeeeenster:
> This looks good to me.
> 
> I'm keen to get @kyle-ssg's opinion here too.
> 
> Also, one thing this has highlighted is that this means that all organisations in a given enterprise installation will have all functionality. Previously these controls were implemented on a per organisation basis via subscriptions that users created in their own installations. Keen to get @dabeeeenster's thoughts here?

That's fine

frontend/common/utils/utils.tsx
```diff
@@ -248,11 +248,12 @@ const Utils = Object.assign({}, require('./base/_utils'), {
     return planNames.free
   },
   getPlanPermission: (plan: string, permission: string) => {
-    let valid = true
-    const planName = Utils.getPlanName(plan)
-    if (!Utils.getFlagsmithHasFeature('plan_based_access')) {
+    if (global.flagsmithVersion?.backend.is_enterprise) {
```
kyle-ssg: I think it's more generic / less likely to have knock on effects if 

we return planNames.enterprise in getPlanName and remove the above conditional.

```
if (plan?.includes('enterprise') || global.flagsmithVersion?.backend.is_enterprise ) {
      return planNames.enterprise
    }
```

kyle-ssg: If we did that we could remove this if too.

kyle-ssg:
Added 1 comment to PR. In regards to controlling this per org, if that's important to the business I think ultimately we should just return plan data for those orgs, then the FE never has to distinguish between installations and it all comes down to the plan.

matthewelwell:
>I think ultimately we should just return plan data for those orgs

This is what we were doing previously though isn't it?

kyle-ssg:
Not sure, but if we were then that sounds better than detecting their installation type.

novakzaballa:
> Not sure, but if we were then that sounds better than detecting their installation type.
@kyle-ssg @matthewelwell If you take a look at the code changes this PR ended up only replacing `!Utils.getFlagsmithHasFeature('plan_based_access')` with `global.flagsmithVersion?.backend.is_enterprise` 
image

novakzaballa:
Now, in addition to the above comment, what I understand from @kyle-ssg comment is that we should remove these lines that return true if it is a self-hosted enterprise from the `getPlanPermission` and the `getPlansPermission` functions, and instead return 'enterprise' in the `getPlanName` function when `global.flagsmithVersion?.backend.is_enterprise` is true, is it correct?

matthewelwell (approved):

PR #2423, opened 2023-07-12
Ensure that project admins are considered as environment admins

matthewelwell:
## Changes

This is a somewhat temporary fix to ensure that project admins are evaluated as environment admins. It is considered to be temporary as it is not an efficient solution to the problem, likely resulting in more database queries than are strictly necessary. 

We (@gagantrivedi and I) need to discuss a more permanent solution for inherited admin permissions. 

Added a new unit test.


novakzaballa (approved):
LGTM

PR #2428, opened 2023-07-12
ci: fix labeller permissions

dabeeeenster:
Hit this error: https://github.com/Flagsmith/flagsmith/actions/runs/5535805708/jobs/10102641881?pr=2427
Fix as per https://github.com/actions/labeler#permissions

### DIFF
--- a/.github/workflows/github-labeler.yml
+++ b/.github/workflows/github-labeler.yml
@@ -1,7 +1,8 @@
 name: Github Labeler
 
 on:
-  - pull_request_target
+  pull_request_target:
+    types: [opened, synchronize, reopened, ready_for_review]
 
 jobs:
   apply-labels:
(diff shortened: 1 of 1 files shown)


.github/workflows/github-labeler.yml
```diff
@@ -1,8 +1,7 @@
 name: Github Labeler
 
 on:
-  pull_request:
-    types: [opened, synchronize, reopened, ready_for_review]
+  - pull_request_target
```
matthewelwell: I think this syntax is wrong though? The change here should just be to add `_target` to line 4 and the rest should stay the same? 

image

dabeeeenster:
The example given is:

image

I'm not actually sure why we are specifying those 4 particular PR types - any idea?

dabeeeenster:
Updated

matthewelwell (approved):

matthewelwell:
> The example given is:
> 
> image

Ah, sorry, YAML consistency is just baffling at times. 

> I'm not actually sure why we are specifying those 4 particular PR types - any idea?

Good question, I'm not sure. We probably should remove them and add the labels to all PRs I guess.

PR #2436, opened 2023-07-14
feat(master-api-key/roles): Add roles to master api key

### LINKED ISSUE #910: API Token Management
At the moment, to use the rest API, user's must use an API token associated with a user. We should add the ability to create API tokens scoped to allow access to the relevant resources.

gagantrivedi:

## Changes
- Adds authentication class for master API key
- Remove master API key middleware
- Remove master API key specific permissions 
- Add roles support to master API key 

New endpoint: 
```
[GET/PUT/DELETE/POST] /api/v1/organisations/{organisation_pk}/roles/{role_pk}/master-api-keys/
```

Covered by unit tests


api/features/feature_segments/serializers.py
```diff
@@ -110,6 +112,13 @@ def validate(self, attrs):
                 "All feature segments must belong to the same feature & environment."
             )
 
+        environment = environments.pop()
+
+        if not self.context["request"].user.has_environment_permission(
```
gagantrivedi, line 117: I don't really like that the permission check is in serializer instead of `permissions.py` but due to the structure of the request this feels a bit clean, what do you think?

api/api_keys/user.py
```diff
@@ -0,0 +1,62 @@
...
+)
+
+from .models import MasterAPIKey
+
+if typing.TYPE_CHECKING:
+    from environments.models import Environment
+    from projects.models import Project
+
+
+class APIKeyUser:
```
matthewelwell: Can we extract a common Abstract class here to ensure that the same methods are implemented on both this class and the user class?

api/core/models.py
```diff
@@ -132,13 +133,25 @@ def _get_project(self) -> typing.Optional["Project"]:
...
 
+def get_history_user(
+    instance: typing.Any, request: HttpRequest
+) -> typing.Optional["FFAdminUser"]:
+    from api_keys.user import APIKeyUser
+
+    try:
+        return None if isinstance(request.user, APIKeyUser) else request.user
+    except AttributeError:
+        return None
```
matthewelwell:
```suggestion
    return request.user if hasattr(request, "user") and not isinstance(request.user, APIKeyUser) else None
```

api/features/feature_segments/serializers.py
matthewelwell: Hmm... yeah, interesting I guess this is because we don't have any context on which objects are being accessed before this point. 

I think you're right, that keeping it here makes the most sense.

api/permissions/rbac_wrapper.py
```diff
@@ -23,3 +30,57 @@ def get_roles_permission_data_for_environment(*args, **kwargs):
 
     def get_role_permission_filter(*args, **kwargs) -> Q:
         return Q()
+
+
+def is_master_api_key_object_admin(
+    master_api_key: "MasterAPIKey", object_: Union[Project, Environment]
+) -> bool:
+    if not settings.IS_RBAC_INSTALLED:
+        return False
```
matthewelwell, line 39: Maybe I missed this somewhere already, but they can still be an admin via the org level admin role, right?

api/projects/views.py
```diff
@@ -97,12 +93,10 @@ def get_queryset(self):
...
-        if self.request.user.is_anonymous:
-            return
-
-        UserProjectPermission.objects.create(
-            user=self.request.user, project=project, admin=True
-        )
+        if isinstance(self.request.user, FFAdminUser):
+            UserProjectPermission.objects.create(
+                user=self.request.user, project=project, admin=True
+            )
```
matthewelwell: Interesting, this is going to mean that the key which creates the project will not have access to the project once its created it, right? (unless it's an admin key) Perhaps this is what you'd expect but it certainly feels slightly odd. 

I guess, to be fair, people are unlikely to be creating projects via these API keys anyway.

api/permissions/rbac_wrapper.py
gagantrivedi: Roles don't support organisation admin, a key can be organisation admin directly though and that is handled in permission_service

api/projects/views.py
gagantrivedi: Yeah, this is an interesting one, we can't really give it admin permission like we give it to the user because that will mean updating the assigned role

api/users/models.py
```diff
@@ -310,6 +311,9 @@ def remove_as_group_admin(self, group_id: int):
         ).update(group_admin=False)
 
 
+assert issubclass(FFAdminUser, UserABC)
```
gagantrivedi, line 331: @matthewelwell this looks a bit ugly, but we can't really inherit from the UserABC because of metaclass conflict

api/projects/views.py
matthewelwell: Yeah, I tend to agree.

api/permissions/rbac_wrapper.py
matthewelwell: Yep, thought so 👍

api/api_keys/user.py
```diff
@@ -18,11 +19,16 @@
...
         self.key = key
 
-    def is_authenticated(self):
+    @property
+    def is_authenticated(self) -> bool:
+        return True
+
+    @property
+    def is_master_api_key_user(self) -> bool:
         return True
```
matthewelwell: Does this need to be a property? Can we not set it directly in the `__init__` method?

api/core/models.py
```diff
@@ -136,12 +136,12 @@ def _get_project(self) -> typing.Optional["Project"]:
...
-    from api_keys.user import APIKeyUser
-
-    try:
-        return None if isinstance(request.user, APIKeyUser) else request.user
-    except AttributeError:
-        return None
+    return (
+        request.user
+        if hasattr(request, "user")
+        and not hasattr(request.user, "is_master_api_key_user")
```
matthewelwell: I think I'd rather see these as `getattr` instead of `hasattr`?

api/users/models.py
matthewelwell: I don't really follow what this is doing. In fact, I'm surprised it's not raising an AssertionError. If `FFAdminUser` doesn't inherit from `UserABC` this should raise an assertion error as far as I'm aware. 

The following snippet (which seems to be a simple representation of what you guys have here) will raise an AssertionError on the last line, so I'm not sure how your code isn't 🤔 

```python
class Base(ABC):
    @abstractmethod
    def foo(self):
        raise NotImplementedError()
    
class Sub(Base):
    def foo(self):
        return "sub-foo"
    
class NotSub:
    def foo(self):
        return "not-sub-foo"
        
assert issubclass(Sub, Base)  # this is fine
assert issubclass(NotSub, Base)  # this is not!
```

gagantrivedi: That is because of this:  `def __subclasshook__(cls, subclass):`

i.e: this code will work 


```python
from abc import ABC, abstractmethod

class Base(ABC):
    @abstractmethod
    def foo(self):
        raise NotImplementedError()

    @classmethod
    def __subclasshook__(cls, subclass):
        return "foo" in subclass.__dict__

class Sub(Base):
    def foo(self):
        return "sub-foo"

class NotSub:
    def foo(self):
        return "not-sub-foo"

assert issubclass(Sub, Base)  # this is fine
assert issubclass(NotSub, Base)  # this is not!



```

api/api_keys/user.py
gagantrivedi: It just mimics the style of django I think, no? 
e.g:

```
    @property
    def is_anonymous(self):
        return True

    @property
    def is_authenticated(self):
        return False


```

api/core/models.py
gagantrivedi: Do you mean something like the current version?

api/api_keys/user.py
matthewelwell: Ok, sure. I guess they did that though so that you could add run time logic to decide the outcome - I'm not sure we'll ever need it here. But I don't think it's an issue anyway.

.github/workflows/api-deploy-production-ecs.yml
```diff
@@ -39,6 +39,7 @@ jobs:
           flagsmith_saml_revision: v1.1.0
           flagsmith_workflows_revision: v1.2.5
           flagsmith_auth_controller_revision: v0.0.1
+          flagsmith_rbac_revision: v0.3.0
```
matthewelwell: I think this is a bad merge?

api/core/models.py
matthewelwell: Yep, looks good.

api/environments/views.py
```diff
@@ -189,7 +184,7 @@ def permissions(self, *args, **kwargs):
         url_name="my-permissions",
     )
     def user_permissions(self, request, *args, **kwargs):
-        if request.user.is_anonymous:
+        if hasattr(request.user, "is_master_api_key_user"):
```
matthewelwell: This should probably be a `getattr` too? There are a few more cases where we're doing this exact check too. I don't think it makes sense to be `hasattr`, right? Because (based on what we talked about earlier regarding the property), it could in some future scenario return `False`, right?

gagantrivedi: Yeah, done

api/conftest.py
```diff
@@ -292,19 +292,36 @@ def environment_api_key(environment):
...
 
 @pytest.fixture()
-def master_api_key_client(master_api_key):
+def master_api_key_client(master_api_key_and_obj):
+    key = master_api_key_and_obj[0]
     # Can not use `api_client` fixture here because:
     # https://docs.pytest.org/en/6.2.x/fixture.html#fixtures-can-be-requested-more-than-once-per-test-return-values-are-cached
     api_client = APIClient()
-    api_client.credentials(HTTP_AUTHORIZATION="Api-Key " + master_api_key[1])
+    api_client.credentials(HTTP_AUTHORIZATION="Api-Key " + key)
```
matthewelwell: Some thoughts here: 

 1. The arguments to the fixtures should have type hints. 
 2. `master_api_key_and_obj` is a little odd, it also doesn't indicate whether it's an admin key or not. It feels like all these fixtures should just return the object itself and the key? Then we'd only have a need for 2 fixtures, right?

gagantrivedi: You mean just two fixures for admin and non admin maseter api key  and both of them return the object and the key?I did think about doing that, but we do use just one of them at a lot of places

matthewelwell: I still think I prefer that - I really hate have the word 'and' in fixture / function names.

gagantrivedi: Yeah, that's a fair point - I have added another _object fixture

matthewelwell (approved):

PR #2446, opened 2023-07-14
ci: lint all the things

dabeeeenster:
This does a few things - has to be in one PR otherwise builds will fail

- Moves all CI linting out of individual actions and puts them in their own action
- Lints everything that is left


dabeeeenster:
@kyle-ssg I need some checks on this commit https://github.com/Flagsmith/flagsmith/pull/2446/commits/68740a2ef26845fdaf2f0d0b7eab4da325ec6ce7 also need to go through some of the files excluded:

```
frontend/web/components/tags/TagValues.tsx
frontend/web/components/pages/EnvironmentSettingsPage.js
frontend/common/utils/base/_utils.js
``` 

in `eslintignore`

.eslintignore
```diff
@@ -7,4 +7,7 @@ frontend/web/static
 frontend/web/project/libs.js
 frontend/e2e
 frontend/webpack
-frontend/api
\ No newline at end of file
+frontend/api
+frontend/web/components/tags/TagValues.tsx
+frontend/web/components/pages/EnvironmentSettingsPage.js
+frontend/common/utils/base/_utils.js
```
matthewelwell: is there any explanation why these are ignored?

matthewelwell (commented):
One thing that I am concerned about here (but maybe can be resolved in the makefile) is that it's kinda painful to actually apply linting changes. 

Previously, if I needed to make some linting changes on multiple files, I would just run `black . && isort .`. We now will be left with `make lint` which passes the `--check` (or similar) argument.

.eslintignore
dabeeeenster: My lack of JS. @kyle-ssg might be able to fix these?

dabeeeenster:
We can have `make lint-fix` or similar? 

`pre-commit run --all-files` or `pre-commit run --all-files <linter>` achieves this

PR #2480, opened 2023-07-20
feat(limits): Add limits to features, segments and segment overrides

### LINKED ISSUE #2477: Implement limits on the number of project and environment entities
We need to limit the number of features, segments and segment overrides that organisations can create in their projects and environments to prevent the environment document growing larger than the limits in dynamodb.

gagantrivedi:

## Changes

Limit number of segments, segment override and feature using environment variables 

Adds unit test cases


api/segments/serializers.py
```diff
@@ -52,19 +54,26 @@ def validate(self, attrs):
...
         rules_data = validated_data.pop("rules", [])
+
+        # create segment with nested rules and conditions
         segment = Segment.objects.create(**validated_data)
         self._update_or_create_segment_rules(
             rules_data, segment=segment, is_create=True
         )
         return segment
 
+    def validate_project_segment_limit(self, project: Project) -> None:
```
matthewelwell, line 68: This could just be part of the `validate` method couldn't it?

api/features/serializers.py
```diff
@@ -427,3 +441,21 @@ def _get_save_kwargs(self, field_name):
...
+    ) -> None:
+        if (
+            environment.feature_segments.count()
+            >= settings.MAX_SEGMENT_OVERRIDE_ALLOWED
+        ):
+            raise serializers.ValidationError(
+                {
+                    "environment": "The environment has reached the maximum allowed segments overrides limit."
+                }
+            )
```
matthewelwell: I have a feeling we may also need to add this to the `SimpleFeatureStateSerializer`?

matthewelwell:
> Check that the limit makes sense

A few things here: 

 - Do we want to optimise for the dynamodb limit? 
 - Do we want to optimise for memory usage in the task processor containers? 
 - What happens to the size of the document as we max out a 'regular' environment? 
 - We need to add validation to migrating to edge and possibly to even generating the environment document from the Core API - we have a customer that has ~1700 flags that is still using our core API. If they were to migrate to the Edge API, it may cause us some further issues.

Here are some statistics from some of our larger environments: 

image

... and the statistics from the top users of segments:

Screenshot 2023-07-20 at 09 42 55

... and the statistics from the top users of features: 

Screenshot 2023-07-20 at 09 44 48

... and the statistics from the top users of segment overrides

Screenshot 2023-07-20 at 10 08 15

Based on this, I would say that we can increase the limits you have mentioned here to perhaps: 

 * 100 segments. 
 * 300 features
 * 100 segment overrides

Obviously these statistics all depend heavily on what the values for these features are. Large JSON documents will of course cause an issue. 

Thoughts please @dabeeeenster @gagantrivedi

gagantrivedi:
> > Check that the limit makes sense
> 
> A few things here:
> 
>     * Do we want to optimise for the dynamodb limit?
> 
>     * Do we want to optimise for memory usage in the task processor containers?
> 
>     * What happens to the size of the document as we max out a 'regular' environment?
> 
>     * We need to add validation to migrating to edge and possibly to even generating the environment document from the Core API - we have a customer that has ~1700 flags that is still using our core API. If they were to migrate to the Edge API, it may cause us some further issues.
> 
> 
> Here are some statistics from some of our larger environments:
> image
> 
> ... and the statistics from the top users of segments:
> Screenshot 2023-07-20 at 09 42 55
> 
> ... and the statistics from the top users of features:
> Screenshot 2023-07-20 at 09 44 48
> 
> ... and the statistics from the top users of segment overrides
> Screenshot 2023-07-20 at 10 08 15
> 
> Based on this, I would say that we can increase the limits you have mentioned here to perhaps:
> 
>     * 100 segments.
> 
>     * 300 features
> 
>     * 100 segment overrides
> 
> 
> Obviously these statistics all depend heavily on what the values for these features are. Large JSON documents will of course cause an issue.
> 
> Thoughts please @dabeeeenster @gagantrivedi

I think in an ideal world I would want to build the document on every update (for edge project) and see if it still under the dynamo limit

matthewelwell:
> I think in an ideal world I would want to build the document on every update (for edge project) and see if it still under the dynamo limit

Hmm... yes, that's going to be way too resource intensive though so I think we need to just implement sensible limits here and deal with the edge cases as they happen. This should at least prevent anyone from generating an environment document that exceeds the memory limits of the task processor instances.

api/features/serializers.py
gagantrivedi: I don't think that deal with segment override, right?  Unless we want to limit identity overrides?

api/segments/serializers.py
gagantrivedi: But that would trigger this method for read, update or whenever the serializer is validated, right?

api/features/serializers.py
matthewelwell: It does, yeah - it's how the UI used to manage segment overrides before we added the endpoint to do it in a single call.

matthewelwell (approved):

api/features/serializers.py
matthewelwell: We can do this in another PR, I'd like to test this in staging and release it asap.

api/projects/models.py
```diff
@@ -61,6 +61,16 @@ class Project(LifecycleModelMixin, SoftDeleteExportableModel):
...
+    max_segments_allowed = models.IntegerField(
+        default=200, help_text="Max segments allowed for this project"
+    )
+    max_features_allowed = models.IntegerField(
+        default=400, help_text="Max features allowed for this project"
+    )
+    max_segments_overrides_allowed = models.IntegerField(
+        default=350,
+        help_text="Max segments overrides allowed for any (one) environment within this project",
+    )
```
matthewelwell: Would it be better to keep the environment variables and use those as the defaults? I'm not sure tbh, it might not be.

matthewelwell: I think we should also put these defaults back down to what we had previously (before I updated them yesterday), so 100 segments, 100 overrides, 400 features. We can then manually update these fields for the projects that currently have higher numbers. 

Could you verify here that these attributes all appear in the django admin?

gagantrivedi: Yeah, I have added them

matthewelwell (approved):

PR #2488, opened 2023-07-20
chore: update conventionalcommit types

dabeeeenster:


.github/workflows/platform-pull-request.yml
```diff
@@ -14,6 +14,19 @@ jobs:
       - uses: amannn/action-semantic-pull-request@v5
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          types: | # mirrors changelog-sections in the /release-please-config.json
+            fix
+            feat
+            infra
```
khvn26, line 21: Should we keep the `infra` type given that it's rather a scope (and @matthewelwell, who added the type, ultimately agrees that it should be a scope)?

I suggest we remove it from here and from `release-please-config.json`.

khvn26: As discussed internally, I guess it won't hurt to have a separate changelog section for SaaS-only infra changes.

If we'll want to hide an infra change from the changelog, we'll be able to use `chore(infra)` anyway.

khvn26 (approved):

PR #2489, opened 2023-07-20
docs: system limits

dabeeeenster:
Also renamed saml so its ordered correctly


docs/docs/system-administration/system-limits.md
```diff
@@ -0,0 +1,16 @@
+---
+title: System Limits
+---
+
+In order to ensure consistent performance, Flagsmith has the following limitations.
+
+## Entity Counts
+
```
gagantrivedi, line 8: we have increased it here: https://github.com/Flagsmith/flagsmith/pull/2480/commits/8fe5049ab7d12a3e666a17da405e4d6352108103

gagantrivedi (changes requested):

docs/docs/system-administration/system-limits.md
dabeeeenster: OK updated

gagantrivedi (approved):

PR #2495, opened 2023-07-21
fix(roles/org-permission): Add missing viewset

gagantrivedi:

## Changes

Bump rbac module to add support for organisation roles permission viewset 

Covered by rbac module


api/conftest.py
```diff
@@ -328,6 +328,11 @@ def view_project_permission():
     return PermissionModel.objects.get(key=VIEW_PROJECT)
 
 
+@pytest.fixture()
+def create_project_permission():
```
matthewelwell: I'm not sure why this was moved? It also no longer depends on the `db` fixture, which I think will break it (if it's used on its own at least).

gagantrivedi: To answer the first question: it was not being used. As for the second one: yes, added

matthewelwell (approved):

PR #2508, opened 2023-07-24
fix: ensure recurring tasks are unlocked after being picked up (but not executed)

matthewelwell:
## Changes

Ensures that recurring tasks are unlocked after being picked up (but not executed) by the task processor. 

Updated existing unit test.

### DIFF
--- a/api/tests/unit/task_processor/test_unit_task_processor_processor.py
+++ b/api/tests/unit/task_processor/test_unit_task_processor_processor.py
@@ -80,7 +80,7 @@ def test_run_recurring_tasks_multiple_runs(db, run_by_processor):
     task_identifier = "test_unit_task_processor_processor._create_organisation"
 
     @register_recurring_task(
-        run_every=timedelta(milliseconds=100), args=(organisation_name,)
+        run_every=timedelta(milliseconds=200), args=(organisation_name,)
     )
     def _create_organisation(organisation_name):
         Organisation.objects.create(name=organisation_name)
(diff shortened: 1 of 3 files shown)


api/tests/unit/task_processor/test_unit_task_processor_processor.py
```diff
@@ -89,17 +89,28 @@ def _create_organisation(organisation_name):
...
-    time.sleep(0.2)
 
-    second_task_runs = run_recurring_tasks()
+    # run the process again before the task is scheduled to run again to ensure
+    # that tasks are unlocked when they are picked up by the task processor but
+    # not executed.
+    no_task_runs = run_recurring_tasks()
 
-    task_runs = first_task_runs + second_task_runs
+    time.sleep(0.3)
```
khvn26, line 99: Is there a specific state we're waiting for here?

matthewelwell: Yeah, so the short answer is that we need to ensure that run_every (in this case 200ms) has passed.

The long answer is that this test should now ensure that the following happens:

 - first execution of `run_recurring_tasks()` - task has not been executed before so it should execute it
 - second execution of `run_recurring_tasks()` - task has been executed already and run_every has not passed since the first run
 - third execution of `run_recurring_tasks()` - task has been executed already and run_every has passed since the first run.

gagantrivedi (approved):

api/tests/unit/task_processor/test_unit_task_processor_processor.py
matthewelwell: It's a bit janky as we're writing tests that rely on time passing which I don't love as there could be a scenario where the tests are running so slowly that we get a false positive. 

The other option I considered was to write a unit test to verify that the task is unlocked after being picked up but not executed. I chose to go with this more 'integration' style approach as it gave me the opportunity to confirm that the exact bug I am fixing has been fixed. 

Perhaps though, I should do both...

khvn26: Both is good but just the one we have is pretty sufficient IMO.

gagantrivedi (approved):

PR #2526, opened 2023-07-26
feat: display warning and prevent creation on limit

novakzaballa:

## Changes

- Add limit Alerts If number of features, segments, and segments Overrides exceeds 90% of the limit or if the API calls exceed 70% of the limit.
- The buttons for creating features, segments, and segments overrides are disabled when the limit is reached.
- Return limits in the API response for projects and, environments.

Unit test added


frontend/web/components/pages/SegmentsPage.tsx
```diff
@@ -63,6 +65,12 @@ const SegmentsPage: FC<SegmentsPageType> = (props) => {
   })
   const [removeSegment] = useDeleteSegmentMutation()
   const hasHadResults = useRef(false)
+  const isLimitReached =
+    ProjectStore.getTotalSegments() >= ProjectStore.getMaxSegmentsAllowed()
```
kyle-ssg: We are referencing data in ProjectStore however we aren't listening to any changes (e.g. project updating or changing to a different project). Are we sure this is reactive?

frontend/web/components/pages/FeaturesPage.js
```diff
@@ -159,8 +160,21 @@ const FeaturesPage = class extends Component {
...
+              environmentFlags,
+              maxFeaturesAllowed,
+              projectFlags,
+              totalFeatures,
+            },
+            { removeFlag, toggleFlag },
+          ) => {
             const isLoading = FeatureListStore.isLoading
+            const isLimitReached = totalFeatures >= maxFeaturesAllowed
+            const showLimitAlert = Utils.calculaterRemainingLimitsPercentage(
```
kyle-ssg: typo: calculater

frontend/common/stores/project-store.js
```diff
@@ -161,6 +161,21 @@ const store = Object.assign({}, BaseStore, {
       return Promise.resolve(store.getEnvironmentIdFromKey(apiKey))
     })
   },
+  getMaxFeaturesAllowed: () => {
+    return store.model && store.model.max_features_allowed
```
kyle-ssg, line 179: Can these be null? If so, what happens?

frontend/common/utils/utils.tsx
```diff
@@ -53,19 +53,16 @@ const Utils = Object.assign({}, require('./base/_utils'), {
     return 100 - total
   },
 
-  calculaterRemainingCallsPercentage(value, total) {
-    const minRemainingPercentage = 30
+  calculaterRemainingLimitsPercentage(total: number, max: number, limit = 90) {
```
kyle-ssg, line 56: Is there a reason why the limit is defaulted to 90?

frontend/web/components/LimitAlert.tsx
```diff
@@ -0,0 +1,27 @@
...
+  percentage: number
+}
+
+const LimitAlert: FC<LimitAlertType> = (props) => {
+  const { limitType, percentage } = props
+  return (
+    <>
+      <div
+        className='alert flex-1 align-items-center'
+        style={{ background: '#FDFDDD' }}
```
kyle-ssg: Can we stick to the design system rather than have a static background

frontend/common/utils/utils.tsx
kyle-ssg: We are specifying types here that I don't believe are always correct, e.g. 
image

Can we guarantee these will always be defined?

frontend/web/components/LimitAlert.tsx
novakzaballa: Added `alert_warning` in the design system.

frontend/common/stores/project-store.js
novakzaballa: This should not be `null` since, by default, when creating a project a max is created for these fields:
```
max_segments_allowed
max_features_allowed
max_segment_overrides_allowed
```
Also, they should not be nullable as reflected in the OpenAPI doc.

image

frontend/web/components/pages/FeaturesPage.js
novakzaballa: Corrected.

frontend/web/components/pages/SegmentsPage.tsx
novakzaballa: The data gets updated every time the `Segmentpage` is loaded.

api/environments/models.py
```diff
@@ -121,6 +121,10 @@ class Environment(
         help_text="If true, will hide sensitive data(e.g: traits, description etc) from the SDK endpoints",
     )
 
+    @property
+    def total_segment_overrides(self) -> int:
```
matthewelwell: Do we need this here or can we annotate the queryset for the view we need?

api/projects/models.py
```diff
@@ -136,6 +136,14 @@ def is_edge_project_by_default(self) -> bool:
             and self.created_date >= settings.EDGE_RELEASE_DATETIME
         )
 
+    @property
+    def total_features(self) -> int:
+        return self.features.count()
+
+    @property
+    def total_segments(self) -> int:
+        return self.segments.count()
```
matthewelwell: As above, can we use `annotate` instead?

api/projects/tests/test_serializers.py
```diff
@@ -78,9 +81,32 @@ def test_ProjectSerializer_get_migration_status_calls_migrator_with_correct_argu
...
+        (ProjectIdentityMigrationStatus.NOT_APPLICABLE.value, False),
+    ],
+)
+def test_ProjectRetrieveSerializer_get_use_edge_identities(
+    project, migration_status, expected
+):
+    # Given
+    serializer = ProjectRetrieveSerializer(
+        context={"migration_status": migration_status}
+    )
```
matthewelwell: I'm not sure why Github thinks this is all new code?

api/projects/tests/test_views.py
```diff
@@ -157,6 +157,67 @@ def setUp(self):
     def _get_detail_url(self, project_id):
         return reverse("api-v1:projects:project-detail", args=[project_id])
 
+    def test_get_project_data(self):
```
matthewelwell: This test could be named better, we're listing projects - why not use the word `list` in the test? 

Also, for new tests, we should not add them to existing class based tests unless we really have to. Can we move this to `api/tests/unit/projects/test_unit_projects_views.py` instead?

api/projects/views.py
```diff
@@ -70,7 +71,14 @@
...
 )
 class ProjectViewSet(viewsets.ModelViewSet):
-    serializer_class = ProjectSerializer
+    def get_serializer_class(self):
+        if self.action == "list":
+            return ProjectListSerializer
+        elif self.action == "retrieve":
+            return ProjectRetrieveSerializer
+
+        return ProjectListSerializer
```
matthewelwell: This can be simplified to: 

```python
if self.action == "retrieve":
    return ProjectRetrieveSerializer
return ProjectListSerializer
```

matthewelwell (changes requested):

api/projects/tests/test_serializers.py
novakzaballa: This is another test

api/projects/tests/test_views.py
novakzaballa: Thanks! moved and renamed properly

api/projects/views.py
novakzaballa: Thanks, Done

api/projects/serializers.py
```diff
@@ -58,6 +56,34 @@ def get_use_edge_identities(self, obj: Project) -> bool:
         )
 
 
+class ProjectRetrieveSerializer(ProjectListSerializer):
+    total_features = serializers.SerializerMethodField()
+    total_segments = serializers.SerializerMethodField()
```
matthewelwell: These could just be `IntegerField`s.

novakzaballa (commented):
I put this `getEnv` function here because I saw that we also have createEnv, editEnv, and deleteEnv here, and making a change would require a major change in the code affecting not just this file but a lot files. Maybe we need to plan an update of these API functions and create a ticket for that.

frontend/web/components/LimitAlert.tsx
```diff
@@ -0,0 +1,27 @@
...
+import Button from 'components/base/forms/Button'
+
+type LimitAlertType = {
+  limitType: string
+}
+
+const LimitAlert: FC<LimitAlertType> = (props) => {
+  const { limitType } = props
+  return (
+    <div className='alert alert-danger flex-1 align-items-center'>
```
kyle-ssg: This should probably use the componentised <ErrorMessage

matthewelwell:
Some issues here: 

**API:** 

 1. Total counts do not exclude deleted rows

**FE:**

 1. There is a typo in the message, it should read 'Your project ...'
 2. We need to round the percentage to the nearest integer value
 3. The toast message moves from the top of the screen to the 'notifications' feed in the top right whenever another notification is sent there (when a feature is created for example) (see image 2 below)
 4. When the viewport width is reduced, the message shows over the top of everything (maybe this isn't an issue since you can close it) 
 5. Once I've exceeded the limit, I am still able to create a feature and just receive a generic error message saying that 'We could not create this feature, please check the name is not in use.' - my understanding was that the button would be greyed out with a warning displayed at the top of the feature modal. 
 6. There are 2 separate warnings that appear on refresh (see second screenshot below)
 7. We might want to update the styling to match the latest changes
 9. GET request to `/environments/:key` is missing a trailing slash, so it's failing in the Uffizzi preview environment.

**General thoughts:**

I wonder if, in general, we should only show these warnings at the top of the relevant screens. So, for the feature limits, only show it at the top of the list of features. That way we could get away with not making it a dismissible toast message. 

So, for example on the segments screen, we could just add a permanent warning message here: 

Screenshot 2023-08-04 at 16 27 28

I feel like this will be easier since it will be less prone to the UI issues that I've listed above and will be more informative to the users anyway. The only concern would be regarding the segment overrides. We would need to add this both to the feature modal, on the segment overrides tab and the segment modal on the features tab. 

**Some other illustrative screenshots:**

image

image

Screenshot 2023-08-04 at 16 12 00

**Appendix**

gist [here](https://gist.github.com/matthewelwell/498f167a85053e2b1e3f01ed368d1460) for testing purposes.

kyle-ssg (dismissed):

frontend/common/stores/project-store.js
```diff
@@ -162,6 +169,24 @@ const store = Object.assign({}, BaseStore, {
     })
   },
   getEnvs: () => store.model && store.model.environments,
+  getMaxFeaturesAllowed: () => {
+    return store.model && store.model.max_features_allowed
+  },
+  getMaxSegmentOverridesAllowed: () => {
+    return store.model && store.model.max_segment_overrides_allowed
```
kyle-ssg, line 176: - Can we update responses.ts to include these new types?
- It looks like this is quite a big breaking change, right now e2e is failing because all of these fields are undefined. 
- Should these be controlled via feature flag or assume undefined means Number.MAX_SAFE_INTEGER?

frontend/common/stores/project-store.js
```diff
@@ -162,6 +169,24 @@ const store = Object.assign({}, BaseStore, {
     })
   },
   getEnvs: () => store.model && store.model.environments,
+  getMaxFeaturesAllowed: () => {
+    return store.model && store.model.max_features_allowed
+  },
+  getMaxSegmentOverridesAllowed: () => {
```
kyle-ssg, line 181: - Can we update responses.ts to include these new types?
- It looks like this is quite a big breaking change, right now e2e is failing because all of these fields are undefined. 
- Should these be controlled via feature flag and assume undefined means Number.MAX_SAFE_INTEGER?

kyle-ssg (changes requested):

frontend/common/stores/project-store.js
novakzaballa: types and Number.MAX_SAFE_INTEGER added.

matthewelwell: These should never be undefined. It's perhaps just an issue that the store is getting updated from multiple places? I don't think it's wise to default to `Number.MAX_SAFE_INTEGER` when that is likely not the actual value. Can we please determine _why_ we're seeing undefined values and look at how we resolve them?

matthewelwell:
@novakzaballa this is looking much better, visually. Some outstanding issues: 

 1. When I create a feature / segment, the percentage in the warning doesn't change (I have to refresh the page)
 2. The alert to confirm that the limit has been reached isn't very clear. Instead of `Your project reached the limit of features|segments totals.` we should say `Your project reached the limit of allowed features|segments, please contact support to discuss increasing this limit.`
 3. I have not found able to test the alerts for when the limit on segment overrides has been reached. I would expect to see an alert in the modal for the following screenshot, and that the button to create a new segment override is greyed out. 

Screenshot 2023-08-07 at 16 59 57

novakzaballa:
> 

1. When I create a feature / segment, the percentage in the warning doesn't change (I have to refresh the page)
That was discussed previously and I mentioned that it was the current behavior, in general, these kind of values are updated only on a page load/reload. Should we make a new call to the corresponding endpoints every time a new feature/segment/segment-override is created/deleted?
2. The alert to confirm that the limit has been reached... should say Your project reached the limit of allowed features|segments, please contact support to discuss increasing this limit.
Corrected
3.
Screen Shot 2023-08-07 at 12 37 50
Screen Shot 2023-08-07 at 14 31 38

I can see the alert, how did you test it?

frontend/common/stores/project-store.js
novakzaballa: Right, as mentioned in a comment above these values should never be undefined, as the API is currently designed. Does it hurt if we have those Max Int values as a fallback? The issue was the logic to disable the creation of new segment overrides, that logic was corrected as part of the prior commit.

matthewelwell:
> That was discussed previously and I mentioned that it was the current behavior, in general, these kind of values are updated only on a page load/reload. Should we make a new call to the corresponding endpoints every time a new feature/segment/segment-override is created/deleted?

Well, I think for the most part it's ok, and maybe we should merge this PR and work on this separately. The main issue is that it won't change to show that they have _breached_ the limit when they add the final allowed feature / segment etc. Maybe that's not too big of an issue for now though. 

> I can see the alert, how did you test it?

Using Uffizzi. See [here](https://pr-2526-deployment-32784-flagsmith.app.uffizzi.com) 
login: matthew.elwell@flagsmith.com / mattelwell123
I have set the limits on the one project that exists to 21 features, segments and segment overrides. There are 20 of each of those entities in the project / environment.

novakzaballa:
> > I can see the alert, how did you test it?
> 
> Using Uffizzi. See [here](https://pr-2526-deployment-32784-flagsmith.app.uffizzi.com) login: [matthew.elwell@flagsmith.com](mailto:matthew.elwell@flagsmith.com) / mattelwell123 I have set the limits on the one project that exists to 21 features, segments and segment overrides. There are 20 of each of those entities in the project / environment.

@matthewelwell , I found a problem with the total segment overrides, I corrected it, could you test it again?

matthewelwell:
@novakzaballa as discussed there is a remaining issue where it's not possible to see the list of segments when the warning is visible. 

image

Also, I think that, instead of hiding the list of segments, we should just grey it out once the limit has been reached.

matthewelwell (approved):
This looks good from a functionality point of view now 👍 

One thing that we need to consider here, however, is that the API code needs to be released before the UI code I believe? 

So, we can either: 

1. Wrap the FE logic in a feature flag (seems overkill to me)
2. Remove the API code from this PR and create a new one with just that code. Then we can approve and merge them in the correct order. 

I have a preference for the latter as I think it will be simpler. I don't think it's worth adding a feature flag here. 

This is a more general issue that we probably need to consider when doing full stack PRs, however.

novakzaballa:
> This looks good from a functionality point of view now 👍
> 
> One thing that we need to consider here, however, is that the API code needs to be released before the UI code I believe?
> 
> So, we can either:
> 
> 1. Wrap the FE logic in a feature flag (seems overkill to me)
> 2. Remove the API code from this PR and create a new one with just that code. Then we can approve and merge them in the correct order.
> 
> I have a preference for the latter as I think it will be simpler. I don't think it's worth adding a feature flag here.
> 
> This is a more general issue that we probably need to consider when doing full stack PRs, however.

Maybe one solution is adding the MaxSafeInteger as fallback values to these new attributes?
In the meantime, I'll try to split the code although it doesn't look like an easy task.

novakzaballa:
> MaxSafeInteger as fallback values to these new attributes?

@matthewelwell I added MaxSafeInteger as fallback values for the new limit attributes and 0 for the count attributes. I tested against the staging API endpoints which don't have the attributes added in this PR and it works as before the changes. I also executed the tests locally against a copy of the code with the code of the current main branch, and they executed correctly.

frontend/web/components/SegmentOverrides.js
```diff
@@ -539,15 +539,25 @@ class TheComponent extends Component {
 
     const visibleValues = value && value.filter((v) => !v.toRemove)
 
+    const THRESHOLD = 90
```
matthewelwell: This is 90 everywhere, I'd say we should just define it in the `calculateRemainingLimitsPercentage` ?

matthewelwell (changes requested):
I've tested this manually and all looks good to me. 

I've also merged main and removed the CHANGELOG styling change which had snuck in (and subsequently resolved that by updating `.prettierignore`).

I've added one fairly minor comment to reduce the duplication of a given argument.

frontend/common/utils/utils.tsx
```diff
@@ -54,25 +56,40 @@ const Utils = Object.assign({}, require('./base/_utils'), {
     return 100 - total
   },
 
-  calculaterRemainingCallsPercentage(value, total) {
-    const minRemainingPercentage = 30
+  calculateRemainingLimitsPercentage(
+    total: number | undefined,
+    max: number | undefined,
+    threshold = 90,
```
matthewelwell, line 62: this should still have a type, right?

frontend/web/components/App.js
```diff
@@ -346,6 +350,74 @@ const App = class extends Component {
...
+                                      'payments_enabled',
+                                    ) && (
+                                      <a
+                                        href='#'
+                                        className='cursor-pointer nav-link p-2'
+                                        style={
+                                          Utils.calculateRemainingLimitsPercentage(
+                                            this.state.totalApiCalls,
+                                            maxApiCalls,
+                                            30,
```
matthewelwell: I'm not sure why this is 30?

frontend/web/components/App.js
```diff
@@ -346,6 +350,74 @@ const App = class extends Component {
...
+                                          <>
+                                            <UpgradeIcon
+                                              width='20'
+                                              fill='#9DA4AE'
+                                            />
+                                            <span>Upgrade</span>
+                                          </>
+                                        )}
+                                      </a>
+                                    )}
```
matthewelwell: As mentioned on slack, I'm not sure why this code is in this PR, I understood that the API calls limits were all part of [this PR](https://github.com/Flagsmith/flagsmith/pull/2279).

frontend/common/utils/utils.tsx
novakzaballa: Assigning a type to it would be redundant, as it's already implied to be a number.
https://github.com/typescript-eslint/typescript-eslint/blob/v5.4.0/packages/eslint-plugin/docs/rules/no-inferrable-types.md

frontend/web/components/App.js
novakzaballa: Is an issue, 70 is the correct percentage to trigger the warning regarding the max API calls. Thank you.

frontend/web/components/App.js
novakzaballa: This is a conflict resolution, based on this [PR](https://github.com/Flagsmith/flagsmith/pull/2264) that showed a warning when reaching a certain percentage.

frontend/web/components/App.js
matthewelwell: This is still 30?

frontend/web/components/App.js
matthewelwell: Ok.

frontend/web/components/App.js
novakzaballa: Yes, as you can see here: https://github.com/Flagsmith/flagsmith/pull/2264

frontend/common/stores/project-store.js
novakzaballa: Marked as resolved as per the comments above

frontend/web/components/SegmentOverrides.js
novakzaballa: Done.

matthewelwell:
I've tested this on staging by hacking in some data to the API for org id 40 to simulate the organisation being within 30% of its allowance. I'm seeing 2 issues: 

 1. I'm not seeing the alert that should be displayed
 2. The 'Upgrade' menu item is shown but doesn't do anything when clicked on (console error for this is provided below).

```
App.js:381 Uncaught ReferenceError: PaymentModal is not defined
    at onClick (App.js:381:46)
    at Object.m (react-dom.production.min.js:15:205)
    at E (react-dom.production.min.js:15:359)
    at react-dom.production.min.js:16:35
    at C (react-dom.production.min.js:16:236)
    at T (react-dom.production.min.js:17:248)
    at O (react-dom.production.min.js:17:57)
    at P (react-dom.production.min.js:17:412)
    at jn (react-dom.production.min.js:85:403)
    at Bn (react-dom.production.min.js:87:257)
```

matthewelwell (approved):
This looks good to me now. A minor nitpick would be to ask if we could add a little more margin between the 'Request more API calls' button and the items in the list above it. Maybe even making that button slightly smaller.

frontend/common/dispatcher/app-actions.js
```diff
@@ -37,6 +37,12 @@ const AppActions = Object.assign({}, require('./base/_app-actions'), {
       projectId,
     })
   },
+  getEnv(environmentId) {
```
kyle-ssg: Is there a reason why we need to get the environment this way? It adds to tech debt, ideally it would be an action in useEnvironment.ts

frontend/web/components/Aside.js
```diff
@@ -77,6 +78,7 @@ const Aside = class extends Component {
               ? {}
               : { live_from_after: new Date().toISOString() },
           )
+          AppActions.getEnv(this.props.environmentId)
```
kyle-ssg: I'm confused by this, we're already getting the project's environments in AppActions.getProject

novakzaballa: This is triggered every time we change environments. It is done this way because we need to save the response to the store so we can use its data to determine the content of the alert and whether it will be displayed.

novakzaballa: We use the environment endpoint because the project one doesn't have the limits information we need, and Matt asked to keep it this way since it will be more efficient in terms of querying the DB

frontend/web/components/App.js
```diff
@@ -484,6 +507,31 @@ const App = class extends Component {
                         </div>
                       ) : (
                         <Fragment>
+                          <Row>
+                            {user &&
```
kyle-ssg: As mentioned on slack, I think this should be componentised so that you can use hooks to get the data.

frontend/common/stores/project-store.js
```diff
@@ -83,6 +84,14 @@ const controller = {
       store.saved()
     })
   },
+  getEnv: (envId) => {
+    data.get(`${Project.api}environments/${envId}/`).then((environment) => {
```
kyle-ssg: Is this used anymore?

frontend/web/components/Aside.js
```diff
@@ -50,6 +50,7 @@ const Aside = class extends Component {
...
```
kyle-ssg: Do we need to do this

frontend/web/components/App.js
```diff
@@ -288,6 +299,17 @@ const App = class extends Component {
     )
     const dismissed = flagsmith.getTrait('dismissed_announcement')
     const showBanner = !dismissed || dismissed !== announcementValue.id
+    const maxApiCallsPercentage = Utils.calculateRemainingLimitsPercentage(
```
kyle-ssg: Is this used anymore

frontend/web/components/App.js
```diff
@@ -70,6 +74,13 @@ const App = class extends Component {
       AccountStore.getOrganisation()?.id &&
       this.state.activeOrganisation !== AccountStore.getOrganisation().id
     ) {
+      getSubscriptionMetadata(getStore(), {
```
kyle-ssg: Is this used anymore

frontend/web/components/Aside.js
kyle-ssg: After our conversation, I think this can now be removed.

kyle-ssg (approved):

PR #2527, opened 2023-07-27
fix: change request audit logs

gagantrivedi:

## Changes
- Add audit log for change request deletion
- Avoid creating feature state audit logs for uncommitted change request

Tests: https://github.com/Flagsmith/flagsmith-workflows/pull/21/files


api/features/models.py
```diff
@@ -886,6 +880,15 @@ def get_delete_log_message(self, history_instance) -> typing.Optional[str]:
...
             return None
 
+    def get_audit_log_related_object_id(self, history_instance) -> int:
+        # Change requests can create, update, or delete feature states that may never go live,
+        # since we already include the change requests in the audit log
+        # we don't want to create separate audit logs for the associated
+        # feature states
+        if self.belongs_to_uncommited_change_request:
+            return None
+        return super().get_audit_log_related_object_id(history_instance)
```
matthewelwell, line 890: Since we're now shifting this logic, I wonder if we should reorder the logic in the `create_audit_log_from_historical_record` task function so that we get the related object id first? Just because (I think) in general the log messages require more database queries to generate. 

i.e. here: https://github.com/Flagsmith/flagsmith/blob/main/api/audit/tasks.py#L87-L94

matthewelwell (approved):
Approved with one minor comment.

api/features/models.py
gagantrivedi: Yeah, I was thinking about doing it

PR #2530, opened 2023-07-27
feat: Launchdarkly importer

### LINKED ISSUE #2348: LaunchDarkly importer
Implement a Launch Darkly importer:

- This would ideally be a local script/app that can be run with an LD API key and a FS API key. 
- Read core data from LD
- Write it to FS

It would be nice if we could add this to the FS front end so we just put in an LD key and the importer runs

## Things to figure out

- Mismatch between core entities between the 2 platforms
- LD has a different flag value model to FS - how to manage this? 
- What language to write this in? Maybe javascript?

## Entities to Migrate

- Projects
- Environments
- Flags

## For later
- Segments./Segment rulesets
- Identities
- Identity data

## UX

1. Create an LD API token in LD (not clear what creds are required here)
2. Go to the Flagsmith Import page
3. Paste in the LD key and hit next
4. UI shows a list of LD projects with checkboxes for the list
5. User selects the projects to import and hits next
6. Flagsmith creates an async task to import the selected projects

dabeeeenster:

- [ ] I have run [`pre-commit`](https://docs.flagsmith.com/platform/contributing#pre-commit) to check linting

## Changes

Implements https://github.com/Flagsmith/flagsmith/issues/2348

_Please describe._

## Left to do

- [x] Add an API interface that takes an LD API key and an LD project ID and then runs the importer code
- [ ] Front End UI in front of the API


api/integrations/launch_darkly/client.py
```diff
@@ -0,0 +1,43 @@
...
+            case status.HTTP_401_UNAUTHORIZED:
+                pass
+            case status.HTTP_403_FORBIDDEN:
+                pass
+            case status.HTTP_404_NOT_FOUND:
+                pass
+            case status.HTTP_429_TOO_MANY_REQUESTS:
+                pass
+            case _:
+                return response.json()
```
matthewelwell: Would be good to have each of the methods define the return typehint as well.

matthewelwell: I guess the ideal (but probably overkill) would be to have some dataclasses (or pydantic models) instead of returning raw dictionaries.

api/integrations/launch_darkly/launch_darkly.py
```diff
@@ -0,0 +1,239 @@
...
+        user: FFAdminUser,
+        request: LaunchDarklyImportSerializer,
+        client: LaunchDarklyClient = None,
+    ):
+        self.client = client or LaunchDarklyClient(api_key)
+        self.request = request
+        self.logger = LaunchDarklyImport.objects.create(
+            user=user,
+            organisation=self.request["organisation_id"],
+            project=self.request["project_id"],
```
matthewelwell: Looks like `project_id` isn't required so there might be the possibility of a keyerror here?

api/integrations/launch_darkly/launch_darkly.py
```diff
@@ -0,0 +1,239 @@
...
+            self.logger.info(
+                "Successfully created new Flagsmith project for Launch Darklys' imported features"
+            )
+
+        self.logger.info(
+            f"Attempting to retrieve Launch Darkly environments under project id: {ld_project_id}"
+        )
+        ld_environments = self.client.get_environments(ld_project_id)
+        if not ld_environments:
+            # todo do we want to stop the script if no environments can be found
```
matthewelwell: Hmm... good question - I guess that as long as we can still create the features, there's no issue.

api/integrations/launch_darkly/launch_darkly.py
```diff
@@ -0,0 +1,239 @@
...
+                description=ld_flag.get("description"),
+                default_enabled=None,  # todo
+                type=STANDARD,
+                tags=None,  # todo
+                is_archived=ld_flag.get("archived", False),
+                owners=None,  # todo
+                is_server_key_only=None,  # todo
+            )
+        except IntegrityError:
+            # Feature with this name already exists
```
matthewelwell: We may actually want to do this check manually - I think that (for example in oracle) we don't always have the db constraint to rely upon.

api/integrations/launch_darkly/launch_darkly.py
```diff
@@ -0,0 +1,239 @@
...
+            ld_environment = ld_flag.get("environments", {}).get(environment.name)
+            if ld_environment:
+                feature_state = FeatureState.objects.create(
+                    feature=feature,
+                    environment=environment,
+                    identity=None,  # todo
+                    feature_segment=None,  # todo
+                    enabled=ld_environment.get("_summary").get("on"),
+                )
+                FeatureStateValue.objects.create(feature_state=feature_state)
```
matthewelwell: I'm surprised this works tbh, by creating the feature above, it should create the FeatureState in all environments. We probably need to do something like: 

```python
            feature = Feature.objects.create(
                ...
                default_enabled=False,  # default to false and set per environment based on LD values
                ...
            )
            for environment in environments:
                 ld_environment = ld_flag.get("environments", {}).get(environment.name)
                 if ld_environment:
                      feature_state = FeatureState.objects.filter(
                          feature=feature,
                          environment=environment,
                          identity=None,  # todo
                          feature_segment=None,  # todo
                      ).update(enabled=ld_environment.get("_summary").get("on"))
```

api/integrations/launch_darkly/launch_darkly.py
```diff
@@ -0,0 +1,239 @@
...
+    ):
+        # todo try get tags
+        try:
+            feature = Feature.objects.create(
+                name=ld_flag.get("key"),
+                project=project,
+                initial_value=None,  # todo
+                description=ld_flag.get("description"),
+                default_enabled=None,  # todo
+                type=STANDARD,
```
matthewelwell: I guess we need to determine whether this should be `type=MULTIVARIATE` or `type=STANDARD` based on the number of variations that are included in the data from LD? If there is only one, should we create just a STANDARD flag with a feature state value?

api/integrations/launch_darkly/launch_darkly.py
```diff
@@ -0,0 +1,239 @@
...
+            if ld_environment:
+                env_feature_option = feature_options.get(ld_environment.get(""))
+                MultivariateFeatureStateValue.objects.create(
+                    feature=feature,
+                    environment=environment,
+                    identity=None,  # todo
+                    feature_segment=None,  # todo
+                    enabled=ld_environment.get("_summary").get("on"),
+                    multivariate_feature_option=env_feature_option,
+                    updated_at=ld_environment.get("_summary").get("lastModified"),
```
matthewelwell: I have a feeling that this won't work, I think the django ORM will overwrite this to the current date. Not really an issue though.

api/integrations/launch_darkly/launch_darkly.py
```diff
@@ -0,0 +1,239 @@
...
+                integer_value=variant_values[1],
+                string_value=variant_values[2],
+            )
+            feature_options[value] = feature_option
+
+        for environment in environments:
+            ld_environment = ld_flag.get("environments", {}).get(environment.name)
+            if ld_environment:
+                env_feature_option = feature_options.get(ld_environment.get(""))
+                MultivariateFeatureStateValue.objects.create(
```
matthewelwell: The key thing that seems to be missing here is the `percentage_allocation`.

api/integrations/launch_darkly/launch_darkly.py
```diff
@@ -0,0 +1,239 @@
...
+            return INTEGER
+        return STRING
+
+    def _get_multivariant_value(
+        self, value: any, type: str
+    ) -> (Optional[bool], Optional[int], Optional[str]):
+        if type == BOOLEAN:
+            return (value, None, None)
+        if type in [INTEGER, FLOAT]:
+            return (None, int(value), None)
```
matthewelwell: We should perhaps log a warning here that we're possibly losing data if the value is a float.

api/integrations/launch_darkly/models.py
```diff
@@ -0,0 +1,47 @@
+from core.models import AbstractBaseExportableModel
+from django.db import models
+
+from integrations.launch_darkly.enums import LogLevel
+from organisations.models import Organisation
+from projects.models import Project
+
+
+class LaunchDarklyImportLogs(AbstractBaseExportableModel):
```
matthewelwell:
```suggestion
class LaunchDarklyImportLog(AbstractBaseExportableModel):
```

api/integrations/launch_darkly/models.py
```diff
@@ -0,0 +1,47 @@
...
+
+class LaunchDarklyImport(AbstractBaseExportableModel):
+    created_by = models.ForeignKey("users.FFAdminUser", on_delete=models.CASCADE)
+    organisation = models.ForeignKey(Organisation, on_delete=models.CASCADE)
+    project = models.ForeignKey(
+        Project, null=True, blank=True, on_delete=models.CASCADE
+    )
+    started_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+    completed = models.BooleanField(default=False)
```
matthewelwell: This could probably just be a property based on whether `completed_at` is not null.

api/integrations/launch_darkly/tests/data/standard_environments_response.json
```diff
@@ -0,0 +1,102 @@
+[
```
matthewelwell: General comment that I'd rather move these tests to `api/tests/`

matthewelwell (commented):
On the whole, this is coming together nicely. I've added a few comments - some of which are probably too specific for now but I figured I'd add them since I was reviewing it already.

api/integrations/launch_darkly/launch_darkly.py
dabeeeenster: So LD can have features without any environments in a project?

api/integrations/launch_darkly/launch_darkly.py
dabeeeenster: yep 100% agree to check in code.

api/integrations/launch_darkly/launch_darkly.py
ajhelsby: Yeah I've noticed that as well, trying to find the best way to retrieve this without a KeyError

api/integrations/launch_darkly/launch_darkly.py
ajhelsby: In theory a user could start with 2 envs, create all their features and then delete all the environments and just want to import the raw flags. It would be a very strange edge case. Just wanted to check how we'd want to handle that

api/integrations/launch_darkly/launch_darkly.py
ajhelsby: yeah coming soon, trying to get the information from the LD flag dicts

ajhelsby: yeah coming soon, trying to get the information from the LD flag dicts

api/integrations/launch_darkly/launch_darkly.py
ajhelsby: Absolutely will add it to the logs

api/integrations/launch_darkly/models.py
ajhelsby: 👍🏻

api/integrations/launch_darkly/models.py
ajhelsby: Yeah agreed it's definitely not required

api/integrations/launch_darkly/tests/data/standard_environments_response.json
ajhelsby: ok sure, will do

ajhelsby:
# LD Importer for Flagsmith: Documentation Update

This documentation provides an overview of the **Launch Darkly (LD) Importer** developed for **Flagsmith**, detailing its functionalities, components, and future considerations.

## Overview

The LD Importer is designed to convert LD flags into Flagsmith features. It involves creating a `LaunchDarklyWrapper` to connect to the LD API and an `LaunchDarklyImport` model for logging. The import process involves fetching LD flags, mapping them to Flagsmith features, and creating associated environment-specific feature states.

## Components

1. **LaunchDarklyWrapper**:
   - Initializes a client to connect to the LD API using a provided API key.
   
2. **LaunchDarklyImport**:
   - Contains logger methods (`info`, `error`, etc.) to create `LaunchDarklyImportLog` models that capture import actions.
   
3. **import_data Method**:
   - Fetches the user's organization to determine the target environment.
   - Creates a new project or uses an existing one based on provided project ID or fetches project information from LD.
   - Fetches environments and flags from LD.
   - Maps LD flags to Flagsmith features based on the `kind` field to differentiate boolean or multivariate flags.
   
## Execution Flow

1. **Organization Determination**:
   - Retrieves the organization from the request serializer. Alternatively, the organization could be extracted from the user's token.
   
2. **Project Creation/Selection**:
   - Creates a new project if no project ID is provided.
   - Uses an existing Flagsmith project if a project ID is provided.
   - Fetches project information from LD if no ID is given.
   
3. **Environment and Feature Creation**:
   - Fetches environments from LD and creates corresponding environments in Flagsmith.
   - Fetches flags from LD and converts them into Flagsmith features.
   - Maps boolean flags to `Feature` objects, using the LD `key` as the `name`.
   - For multivariate flags, creates `MultivariateFeatureOption` instances for each variation in LD.
   
4. **Feature States**:
   - For boolean flags, creates `FeatureState` instances for boolean values.
   - Retrieves boolean values from the `_summary -> on` field of LD flags' response.
   - Corresponding `FeatureStateValue` models are created.
   - For multivariate flags, creates `MultivariateFeatureStateValue` instances.
   - Sets `MultivariateFeatureStateValue` to enabled if `_summary -> on` is set for that environment.

## Future Considerations

- Refining exceptions to provide more specific error information.
- Implementing the script execution mechanism and alignment with the frontend.
- Enhancing unit test coverage, with a framework that allows passing LD API requests using JSON files.
- Determining critical failure points during the import process.

khvn26:
Breakdown for some of the stuff I added to make reviewing a bit less of a pain: 

## General

1. The import service is no longer able to create a project. Frontend and API users will need to provide an existing project.

1. The dedicated import log model is gone. The main `LaunchDarklyImportRequest` model now emits audit logs. It also implements a `status` JSON field which now holds the current import status and error messages we want to show to the users. The field is exposed in the API endpoints and consumed by the audit logs. I feel this approach could be a good framework for other import integrations.

1. The import itself is now a background task. This means we store the LD tokens now. The tokens are encrypted with Django's `django.contrib.signing` module, which uses `settings.DEFAULT_HASHING_ALGORITHM` (sha-256) and `settings.SECRET_KEY` as key. The tokens are erased after an import job is completed no matter the outcome.

1. Multivariate percentage values are now imported correctly. Since LD allows percents with fractions, a simple algorithm is used to round them to integer values while keeping the 100% sum intact.

1. All LD flag values are treated as string values.

1. It's assumed that the provided project is empty. In case of collsions, environments, features, feature states and feature state values are overwritten.

## API endpoint

1. Users can now consume the import service by calling `api/v1/projects/{project_pk>/imports/launch_darkly` endpoint. Import requests are created with a POST request. Existing import requests can be viewed with a GET request, list and detail endpoints both available.

3. The POST action stores an import estimate of how many environments and features will be created. It's available to users in the `status` field.

4. The POST action requires the caller to have the `CREATE_ENVIRONMENT` permission. The GET actions require the `VIEW_PROJECT` permission. 

## LD REST API client

1. LD encourages REST API users to [always request a specific version](https://apidocs.launchdarkly.com/#section/Overview/Versioning) using the `LD-API-Version` header. Flagsmith will consume version 20220603, set up with as a constant in api/integrations/launch_darkly/client.py module.

1. LD REST API employs pagination for collection requests. The client now takes this into account, exhausting all available pages.

1. We take advantage of [LD's OpenAPI schema](https://apidocs.launchdarkly.com/#section/Overview/OpenAPI-(Swagger)-and-client-libraries) schema by parsing it into `TypedDict` types used with the client. It gives us more confidence when writing the client consuming code. The generation process is automated and tailored to avoid having all of the LD schemas in our codebase (see the `generate-ld-client-types` Make target).

api/Makefile
```diff
@@ -84,3 +84,17 @@ django-collect-static:
 .PHONY: serve
 serve:
 	poetry run gunicorn --bind 0.0.0.0:8000 app.wsgi --reload
+
+.PHONY: generate-ld-client-types
+generate-ld-client-types:
+	curl -sSL https://static.launchdarkly.com/app/s/openapi.2161bd87d.json | \
```
gagantrivedi: How often is this URL supposed to change?

api/integrations/launch_darkly/services.py
```diff
@@ -0,0 +1,292 @@
...
+
+
+def _create_tags_from_ld(
+    ld_tags: list[str],
+    project_id: int,
+) -> dict[str, Tag]:
+    tags_by_ld_tag = {
+        tag: Tag(
+            label=tag,
+            color="#3d4db6",
```
gagantrivedi: Tags imported from LD will have the same colour? Should we make a constant?

api/pyproject.toml
```diff
@@ -29,7 +29,6 @@ use_parentheses = true
 multi_line_output = 3
 include_trailing_comma = true
 line_length = 79
-known_first_party=['analytics','app','custom_auth','environments','integrations','organisations','projects','segments','users','webhooks','api','audit','e2etests','features','permissions','util']
```
gagantrivedi: Why did we remove this?

api/integrations/launch_darkly/services.py
```diff
@@ -0,0 +1,292 @@
...
+
+
+def _sign_ld_value(value: str) -> str:
+    return signing.dumps(value, salt="ldimport")
+
+
+def _unsign_ld_value(value: str) -> str:
+    return signing.loads(
+        value,
+        salt="ldimport",
```
gagantrivedi: Does user_id/token makes more sense as salt?

gagantrivedi (changes requested):

api/pyproject.toml
khvn26: No idea how, I certainly did not intend to. Restored.

api/integrations/launch_darkly/services.py
khvn26: Agreed, done.

api/Makefile
khvn26: Changed to the public one from the LD docs.

api/integrations/launch_darkly/services.py
khvn26: Done.

api/Makefile
```diff
@@ -84,3 +84,17 @@ django-collect-static:
...
+
+.PHONY: generate-ld-client-types
+generate-ld-client-types:
+	curl -sSL https://app.launchdarkly.com/api/v2/openapi.json | \
+	npx openapi-format /dev/fd/0 \
+		--filterFile ld-openapi-filter.yaml | \
+	datamodel-codegen \
+		--output integrations/launch_darkly/types.py \
+		--output-model-type typing.TypedDict \
+		--target-python-version 3.10 \
```
gagantrivedi, line 96: This should be 3.11, right? Because that's what we use in production? https://github.com/Flagsmith/flagsmith/blob/2c56526706ea48b48714f03103bbb4de1db7ba3a/api/Dockerfile#L3C3-L3C3

api/integrations/launch_darkly/services.py
```diff
@@ -0,0 +1,297 @@
...
+    LaunchDarklyImportStatus,
+)
+from projects.tags.models import Tag
+
+if TYPE_CHECKING:  # pragma: no cover
+    from projects.models import Project
+    from users.models import FFAdminUser
+
+
+def _sign_ld_value(value: str, user_id: object) -> str:
```
gagantrivedi: user_id will be int, right?

api/integrations/launch_darkly/services.py
```diff
@@ -0,0 +1,297 @@
...
+    return signing.dumps(value, salt=f"ld_import_{user_id}")
+
+
+def _unsign_ld_value(value: str, user_id: object) -> str:
+    return signing.loads(
+        value,
+        salt=f"ld_import_{user_id}",
+        max_age=settings.LAUNCH_DARKLY_IMPORT_SENSITIVE_DATA_MAX_AGE_SECONDS,
```
gagantrivedi: This will fail if the task processor can't load this in less than `LAUNCH_DARKLY_IMPORT_SENSITIVE_DATA_MAX_AGE_SECONDS`(i.e: 30 ) seconds, right? I am not sure if this is a good idea?

api/integrations/launch_darkly/views.py
```diff
@@ -0,0 +1,70 @@
...
+            data=request.data
+        )
+        request_serializer.is_valid(raise_exception=True)
+
+        project = get_object_or_404(
+            self.request.user.get_permitted_projects(CREATE_ENVIRONMENT),
+            pk=self.kwargs["project_pk"],
+        )
+        instance = create_import_request(
+            project=project,
```
gagantrivedi: We should really add a check here to make sure we don't have any outstanding import request here before creating another one

gagantrivedi (changes requested):

api/integrations/launch_darkly/tasks.py
```diff
@@ -0,0 +1,12 @@
+from django.db import transaction
+
+from integrations.launch_darkly.models import LaunchDarklyImportRequest
+from integrations.launch_darkly.services import process_import_request
+from task_processor.decorators import register_task_handler
+
+
+@register_task_handler()
+def process_launch_darkly_import_request(import_request_id: int) -> None:
+    with transaction.atomic():
```
gagantrivedi, line 10: I am a bit concerned about this long-running traction: is there any way we can avoid this? or reduce it?

api/Makefile
khvn26: Enterprise image uses 3.10, so it's the lowest common denominator.

3.10 uses `typing_extensions` import for `NotRequired` annotations, so tests actually fail on 3.10 if we use 3.11 argument here.

api/integrations/launch_darkly/services.py
khvn26: Yes, but the function should not care about having an `int` here. Django pks can be anything so I decided against using the `int` annotation.

api/integrations/launch_darkly/services.py
khvn26: What would be a sensible max age from your standpoint? Should we have a max age at all?

api/integrations/launch_darkly/tasks.py
khvn26: Maybe the question is should we have it at all or should we prefer having partial success imports?

To support partial imports better, we could extend the import request status field to have a `partial_success` status and `error_messages` array. Moreover, we could have a per-object status report so users could see which exact entity failed to import.

api/integrations/launch_darkly/services.py
gagantrivedi: I'd probably just get rid of the max age

api/integrations/launch_darkly/services.py
gagantrivedi: But it's a private function with argument named `user_id`. I'd want the type check to shout as loud as possible if it finds user_id assigned to anything but an int in our code base

khvn26: I've changed it, even though this is definitely not the place to guard against non-integer PKs.

api/integrations/launch_darkly/views.py
khvn26: Good call. This means we have to stop obfuscating LD project ids though.

api/integrations/launch_darkly/services.py
khvn26: Done.

api/integrations/launch_darkly/views.py
khvn26: Done.

gagantrivedi (approved):
Let's create an issue for removing transaction before we merge this

api/integrations/launch_darkly/tasks.py
khvn26: As discussed internally, keeping the transaction for now. Partial imports TBD in a separate PR.

khvn26:
https://github.com/Flagsmith/flagsmith/issues/2838

PR #2538, opened 2023-07-27
fix:  Sanitize HTML tooltips

novakzaballa:

## Changes

DIsable HTML tooltips since they could cause XSS injection vulnerability

Manually. 
Insert XSS code like "xss" as the label of a tag.
When hover over the tag, it should display simple text and not HTML.


frontend/web/components/Toolip.js
```diff
@@ -16,7 +16,7 @@ const Tooltip = class extends React.Component {
           <span className='ion ion-ios-help' data-for={this.id} data-tip />
         )}
         <ReactTooltip
-          html
+          html={false}
```
kyle-ssg: We only want to disable html for some usages, a lot of the times we use tooltips e.g. permissions uses html for tooltip content.

novakzaballa: Implementing DOMPurify to sanitize the tooltip content since it removes XSS scripts but not HTML content.

novakzaballa: As per our conversation, proceeded to HTML encode the tag labels to escape HTML and display content as plain text.
Also, converted Tooltip code to Typescript functional component.

frontend/web/components/tags/Tag.tsx
```diff
@@ -47,6 +47,7 @@ const Tag: FC<TagType> = ({
 
   return (
     <Tooltip
+      htmlEncode
```
kyle-ssg: we'd probably want to make the property naming of this a bit more clear.

I think plainText might be good since in this case you're wanting to make the tooltip content text rather than html.

novakzaballa: Sure, that was my first option but then I thought it was not descriptive enough :). Done

novakzaballa: BTW, also fixed the component filename from toolip to tooltip.

frontend/web/components/Tooltip.tsx
```diff
@@ -0,0 +1,72 @@
...
+    <span>{`${children}`}</span>
+  </div>
+)
+
+const tooltipStyler = (
+  plainText: boolean,
+  children: string | JSX.Element | JSX.Element[] | (() => JSX.Element),
+): string => {
+  const html = renderToStaticMarkup(
+    <StyledTooltip>{plainText ? children : '{{placeholder}}'}</StyledTooltip>,
```
kyle-ssg: Is {{placeholder}} supposed to be here?

novakzaballa: Replace with {{html}} for clarity, also, some issues with typescript were corrected

kyle-ssg (approved):

PR #2559, opened 2023-08-01
feat: Add backoff to webhooks, and call multiple webhooks async

### LINKED ISSUE #1654: Implement backoff-delay retries for failed webhooks

tushar5526:
Closes #1654 

## Changes

Significant changes include calling `call_environment_webhooks` and `call_organisation_webhooks` using the task processor and making `_call_webhook_email_on_error` async so that any unhealthy webhook in the list does not block sending payload to other webhooks. 

I have also added `backoff` and `jitter` to `_call_webhook` with `3` retries and using a `max_time` of `2s` so that threads are not marked unhealthy by the task processor's healthcheck. 

I ran unit tests and set up two endpoints as webhooks for testing. Next, I created a feature flag and added those webhooks to it. I then called those webhooks by toggling the feature and tested for various combinations.


api/features/tasks.py
```diff
@@ -38,19 +36,16 @@ def trigger_feature_state_change_webhooks(
...
     if previous_state:
         data.update(previous_state=previous_state)
-    Thread(
-        target=call_environment_webhooks,
-        args=(instance.environment, data, event_type),
-    ).start()
 
-    Thread(
-        target=call_organisation_webhooks,
+    call_environment_webhooks.delay(args=(instance.environment, data, event_type))
```
matthewelwell: As task arguments are serialised in the database using JSON, these arguments need to be JSON-serialisable. We will definitely need to change `instance.environment` to `instance.environment_id` and (I think) we will also need to change `event_type` to `event_type.value`. Obviously this will also need changes to the `call_environment_webhooks` (and `call_organisation_webhooks` functions). 

```python
call_environment_webhooks.delay(args=(instance.environment_id, data, event_type.value))
```

api/webhooks/webhooks.py
```diff
@@ -88,6 +92,12 @@ def trigger_sample_webhook(
     return _call_webhook(webhook, serializer.data)
 
 
+@backoff.on_exception(
+    backoff.expo,
+    requests.exceptions.RequestException,
+    max_tries=8,
```
matthewelwell: 8 seems excessive to me? I would say 3 or 5 would be sufficient?

matthewelwell (changes requested):
Thanks for submitting this @tushar5526. I have added a couple of specific comments, but I'd also love to see some tests written for these too. It seems like perhaps we don't have any currently (otherwise I would have expected them to fail), but if you are able to add some, that would be really great.

api/webhooks/webhooks.py
tushar5526: Yes, I was not sure what to use. While testing - I found 3 to be less, so I think 5 would be better.

tushar5526:
Hi @matthewelwell, I modified the logic to use `id` instead of `environment` or `organisation` model object. Still, it fails the test case where `organistaion` object is deleted and associated webhooks are called later, so doing a `Organisation.objects.get(id)` errors out. Should the organisation webhooks be referenced, even if the organisation model object is deleted? 

https://github.com/Flagsmith/flagsmith/pull/2559/files#diff-86f132934c5b75bd30cb37ce30cc88b07b738d2b09d772cbeea42e2186809cbaR73

matthewelwell:
> Hi @matthewelwell, I modified the logic to use `id` instead of `environment` or `organisation` model object. Still, it fails the test case where `organistaion` object is deleted and associated webhooks are called later, so doing a `Organisation.objects.get(id)` errors out. Should the organisation webhooks be referenced, even if the organisation model object is deleted?
> 
> https://github.com/Flagsmith/flagsmith/pull/2559/files#diff-86f132934c5b75bd30cb37ce30cc88b07b738d2b09d772cbeea42e2186809cbaR73

No, we can ignore the case where the organisation is deleted.

api/webhooks/tests/test_webhooks.py
```diff
@@ -179,34 +179,35 @@ def test_call_environment_webhooks__multiple_webhooks__failure__calls_expected(
...
     call_environment_webhooks(
-        environment=environment,
+        environment_id=environment.id,
         data=expected_data,
         event_type=expected_event_type,
     )
 
     # Then
     assert send_failure_email_mock.call_count == 2
+    print("ARG LIST", send_failure_email_mock.mock_calls)
```
matthewelwell: Can we get rid of this `print` call please?

api/webhooks/webhooks.py
```diff
@@ -119,12 +141,14 @@ def _call_webhook_email_on_error(
         send_failure_email(webhook, data, webhook_type, res.status_code)
 
 
-def _call_webhooks(webhooks, data, event_type, webhook_type):
-    webhook_data = {"event_type": event_type.value, "data": data}
+def _call_webhooks(webhooks_id, data, event_type, webhook_type):
```
matthewelwell: Can we add some typehinting here? Appreciate that it didn't exist before, but we're trying to slowly improve the type hinting across the code base.

api/webhooks/webhooks.py
```diff
@@ -101,9 +118,14 @@ def _call_webhook(
     return requests.post(str(webhook.url), data=json_data, headers=headers, timeout=10)
 
 
+@register_task_handler()
 def _call_webhook_email_on_error(
-    webhook: WebhookModels, data: typing.Mapping, webhook_type: WebhookType
+    webhook_id: int, data: typing.Mapping, webhook_type: WebhookType
```
matthewelwell: I think `webhook_type: WebhookType` might be wrong here? It should be `webhook_type: str` I think?

api/webhooks/webhooks.py
matthewelwell: Also the argument name should be `webhook_ids`, not `webhooks_id`.

matthewelwell (changes requested):

api/webhooks/tests/test_webhooks.py
tushar5526: Oops! Forgot to remove this.

api/webhooks/webhooks.py
tushar5526: Yes, are there any static linters configured as well? It should have thrown an error in `pre-commit`. Similarly, I was able to make a commit with `webhook_id: WebhookModels`, but `webhook_id` was passed in as an integer.

api/webhooks/webhooks.py
tushar5526: Done, thanks!

api/webhooks/webhooks.py
matthewelwell: There aren't _yet_ but it's in the plans so we're trying to ensure we're chipping away at the typing. You're right that it would be good to have something that validates the diff only for now.

tushar5526:
@matthewelwell I am not sure how to go about tests for webhooks using the task processor, should I write a test that forces the arguments to be JSON serializable, in the case when the function definition changes in future? It seems more apt to validate the Input to the Task processor at `task.schedule_task` level and then write tests.

Maybe setting up integration tests for each function that is using Task Processor would be a better way to test. Let me know your thoughts.

matthewelwell:
> @matthewelwell I am not sure how to go about tests for webhooks using the task processor, should I write a test that forces the arguments to be JSON serializable, in the case when the function definition changes in future? It seems more apt to validate the Input to the Task processor at `task.schedule_task` level and then write tests.
> 
> Maybe setting up integration tests for each function that is using Task Processor would be a better way to test. Let me know your thoughts.

Hey @tushar5526, yes, good question. I guess the _ideal_ would be to enforce typing and validate in the decorator that the types of all the arguments are JSON serializable. I had a quick look at that though and it would by no means be foolproof. 

I would suggest that the integration test approach for each function would be the better option for sure. We may also need to add verification that the arguments are JSON serializable to the `delay` function in the `register_task_handler` decorator though, regardless of what the TASK_RUN_METHOD setting is. I think it's sensible to do this - we should always ensure that arguments to the `delay` are JSON serializable so they will work regardless of what the TASK_RUN_METHOD setting is.

tushar5526:
> I would suggest that the integration test approach for each function would be the better option for sure. We may also need to add verification that the arguments are JSON serializable to the `delay` function in the `register_task_handler` decorator though, regardless of what the TASK_RUN_METHOD setting is. I think it's sensible to do this - we should always ensure that arguments to the `delay` are JSON serializable so they will work regardless of what the TASK_RUN_METHOD setting is.

Should I open a new issue? Also, it would be great if you can mention what else is needed to be covered in this PR :)

api/features/tasks.py
tushar5526: Made the arguments serializable.

matthewelwell:
I've created [this PR](https://github.com/Flagsmith/flagsmith/pull/2747) to validate the arguments that are passed to task processor functions both when called directly, or when executed synchronously. This should help to ensure that all new tasks adhere to the serialization rules. 

Based on this, in terms of this PR, I don't think there is any additional logic required. I will review for a final time and then, once my separate PR is merged, we can merge it here and re run the tests as a final check. 

Hopefully that all makes sense?

api/webhooks/webhooks.py
```diff
@@ -43,31 +47,45 @@ class WebhookType(enum.Enum):
...
     if settings.DISABLE_WEBHOOKS:
         return
-
+    try:
+        environment = Environment.objects.get(id=environment_id)
+    except Environment.DoesNotExist:
+        return
     _call_webhooks(
-        environment.webhooks.filter(enabled=True),
+        (webhook.id for webhook in environment.webhooks.filter(enabled=True)),
```
matthewelwell: I don't think we need to worry about the serialization here, right? Once we're at this point, everything is contained within the run time of the task processor. We can pass any objects between functions at this point.

api/webhooks/webhooks.py
```diff
@@ -88,6 +106,12 @@ def trigger_sample_webhook(
     return _call_webhook(webhook, serializer.data)
 
 
+@backoff.on_exception(
+    backoff.expo,
+    requests.exceptions.RequestException,
+    max_tries=3,
+    jitter=backoff.full_jitter,
```
matthewelwell, line 113: One thing that has just cropped up in my head is that I'm not sure what the _maximum_ execution time is now for one of these tasks (assuming a reasonable response time from the webhook URL). My concern for this is that we have monitoring on each of the worker threads in the task processor. If a worker thread hasn't checked for new tasks in a given time period (defaults to 3s, but configured as 20s in our SaaS environment), that thread is deemed unhealthy and the health check will report as unhealthy. It seems perfectly reasonable that, with this logic, a single task to process a webhook could exceed 3s.

api/webhooks/webhooks.py
```diff
@@ -101,9 +125,14 @@ def _call_webhook(
     return requests.post(str(webhook.url), data=json_data, headers=headers, timeout=10)
 
 
+@register_task_handler()
 def _call_webhook_email_on_error(
```
matthewelwell: If this is a task, I'd rather it was a public method.

api/webhooks/webhooks.py
```diff
@@ -119,21 +148,33 @@ def _call_webhook_email_on_error(
         send_failure_email(webhook, data, webhook_type, res.status_code)
 
 
-def _call_webhooks(webhooks, data, event_type, webhook_type):
-    webhook_data = {"event_type": event_type.value, "data": data}
+def _call_webhooks(
+    webhook_ids: typing.Iterable[int],
```
matthewelwell: As above, I don't think we need to worry about the serialization here.

api/webhooks/webhooks.py
tushar5526: Yes, true. We don't need to serialize the arguments to `_call_webhooks`, but `_call_webhooks` calls `_call_webhook_email_on_error.delay`, which should be async so that unhealthy webhooks don't block the for loop. So even if I pass the `webhook` object completely, I would still have to get the serializable `webhook.id` to schedule `_call_webhook_email_on_error`. 


```python
def _call_webhooks(
    webhook_ids: typing.Iterable[int],
    data: typing.Mapping,
    event_type: str,
    webhook_type: WebhookType,
):
    webhook_data = {"event_type": event_type, "data": data}
    serializer = WebhookSerializer(data=webhook_data)
    serializer.is_valid(raise_exception=False)
    for webhook_id in webhook_ids:
        _call_webhook_email_on_error.delay(
            args=(webhook_id, serializer.data, webhook_type.value)
        )
```

I can also get the `webhook.id` in the `_call_webhooks` function, let me know what you think.

api/webhooks/webhooks.py
tushar5526: Very interesting, I will take a look into other `jitter` options.

api/webhooks/webhooks.py
matthewelwell: Yes, I think we should just change the call to `_call_webook_email_on_error.delay` to pass `webhook.id` and pass in the webhook objects to `_call_webhooks`.

api/webhooks/webhooks.py
tushar5526: Changed it to a public method. Thanks

api/webhooks/webhooks.py
tushar5526: Sure :)

api/webhooks/webhooks.py
tushar5526: I have found `max_time` under the `give_up` conditions of [Backoff](https://pypi.org/project/backoff/) project. We can set it to 2s and leave 1s as a buffer. It will solve the purpose.

matthewelwell (approved):
This looks good to me, but I'd appreciate another pair of eyes from @gagantrivedi or @khvn26 before merging.

api/tests/unit/audit/test_unit_audit_models.py
```diff
@@ -5,7 +5,7 @@
 
 def test_organisation_webhooks_are_called_when_audit_log_saved(project, mocker):
     # Given
-    mock_call_webhooks = mocker.patch("audit.signals.call_organisation_webhooks")
```
gagantrivedi, line 8: Think we should update this test to use `assert_called_once_with`(on line 16)

api/features/tests/test_tasks.py
```diff
@@ -51,13 +51,12 @@ def test_trigger_feature_state_change_webhooks(MockThread):
     event_type = environment_webhook_call_args[1]["args"][2]
     assert data["new_state"]["feature_state_value"] == new_value
     assert data["previous_state"]["feature_state_value"] == initial_value
-    assert event_type == WebhookEventType.FLAG_UPDATED
+    assert event_type == WebhookEventType.FLAG_UPDATED.value
```
gagantrivedi, line 57: Same as above, can we add a check here for environment_id/organisation_id

gagantrivedi (changes requested):
LGTM, apart from Minor test improvements

api/features/tests/test_tasks.py
tushar5526: Done, thanks!

gagantrivedi (approved):

PR #2564, opened 2023-08-02
chore: e2e tests with firefox

dabeeeenster:
Moving E2E tests from chrome to firefox. Pain to get the docker images working. 

Using GHCR to host a test runner image that reduces test time.

### DIFF
--- a/.github/workflows/platform-docker-build-e2e-image.yml
+++ b/.github/workflows/platform-docker-build-e2e-image.yml
@@ -23,5 +23,5 @@ jobs:
       - name: 'Build E2E Frontend Image'
         run: |
           cd frontend
-          docker build -f Dockerfile.e2e --tag ghcr.io/flagsmith/e2e-frontend:latest .
+          docker build -f Dockerfile-base.e2e --tag ghcr.io/flagsmith/e2e-frontend:latest .
           docker push ghcr.io/flagsmith/e2e-frontend:latest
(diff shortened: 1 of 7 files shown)


.github/workflows/platform-docker-build-e2e-image.yml
```diff
@@ -0,0 +1,28 @@
...
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{github.actor}}
+          password: ${{secrets.GITHUB_TOKEN}}
+
+      - name: Build E2E Frontend Image
+        run: |
+          cd frontend
+          docker build -f Dockerfile-base.e2e --tag ghcr.io/flagsmith/e2e-frontend:latest .
```
gagantrivedi: Should we rename the tag to something like `ghcr.io/flagsmith/e2e-frontend-base:latest` or something similar so that we don't get confused between the base image and the actual image? Thoughts?

gagantrivedi (changes requested):

.github/workflows/platform-docker-build-e2e-image.yml
dabeeeenster: OK yep changed

gagantrivedi (approved):

matthewelwell (approved):
Nice work @dabeeeenster

matthewelwell:
can we (perhaps in a separate PR) use the e2e docker image here too: https://github.com/Flagsmith/flagsmith/tree/main/.github/actions/e2e-tests ?

dabeeeenster:
> can we (perhaps in a separate PR) use the e2e docker image here too: https://github.com/Flagsmith/flagsmith/tree/main/.github/actions/e2e-tests ?

That test doesnt use docker - the Github runners have chrome and firefox pre installed so probably best to carry on using those...

PR #2583, opened 2023-08-04
fix: total api calls handling

kyle-ssg:

## Changes

**Current problems**

- The promise to retrieve organisation usage was incorrect, this is causing the following error in production.

Currently in production the total api call detection does not work, this is because we are incorrectly parsing the response.
image
- The action to get organisation usage is in the render function, this should never happen, side affects (especially API and set state calls) should never be there. This is causing the getOrganisation usage to be rapidly called in production. Luckily Redux RTK detects duplicate api call requests and only hits it once.
- The check to call get organisation usage included whether the total api calls state was 0, the problem with this is that it is infinitely called for new organisations and projects. The only reason why this is not breaking production is because of the first bullet point (It does call around 10 or so times but eventually fails and stops doing it).

**Changes**

- Fix the promise of getOrganisationUsage to return just the data of the API call
- Only call the organisation usage endpoint when we don't have the data for the active organisation (on mount, on organisation switch)

Tested retrieving the initial usage and then switching organisations.

### DIFF
--- a/frontend/common/services/useOrganisationUsage.ts
+++ b/frontend/common/services/useOrganisationUsage.ts
@@ -57,14 +57,11 @@ export async function getOrganisationUsage(
     typeof organisationUsageService.endpoints.getOrganisationUsage.initiate
   >[1],
 ) {
-  store.dispatch(
+  return store.dispatch(
     organisationUsageService.endpoints.getOrganisationUsage.initiate(
       data,
       options,
     ),
-  )
-  return Promise.all(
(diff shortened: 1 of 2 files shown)


matthewelwell:
@kyle-ssg is this separate to the graph in the organisation settings then because that seems to be working? 

image

kyle-ssg:
It is separate yes, this is for API usage warnings. Can see it erroring 

image


However I think there's a more elegant way of managing this so will mark the PR as draft and adjust later today.

matthewelwell (approved):

PR #2605, opened 2023-08-08
ci: cache api poetry

dabeeeenster:
OK Github python action has support for poetry

### DIFF
--- a/.github/workflows/api-pull-request.yml
+++ b/.github/workflows/api-pull-request.yml
@@ -47,13 +47,8 @@ jobs:
         run: pipx install poetry
       - uses: actions/setup-python@v4
         with:
-          python-version: '3.11'
-          cache: 'poetry'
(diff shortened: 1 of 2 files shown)


.github/workflows/api-pull-request.yml
```diff
@@ -47,10 +47,22 @@ jobs:
...
-          cache: pip
+
+      - name: Install Poetry
+        uses: snok/install-poetry@v1
+
+      - name: Load cached venv
+        id: cached-poetry-dependencies
+        uses: actions/cache@v2.1.6
+        with:
+          path: .venv
```
matthewelwell: I don't know that this will work? I don't think poetry uses `.venv` by default. I think it uses some `/usr/local/.poetry` or something?

matthewelwell (approved):

PR #2615, opened 2023-08-09
feat: add limits and totals to API responses

matthewelwell:
## Changes

Add feature, segment and segment override limits to API responses. 

Tests written by Novak as part of #2526.


novakzaballa (approved):
Approved

PR #2619, opened 2023-08-10
fix(password-reset): rate limit password reset emails

gagantrivedi:

## Changes

Rate limit password reset emails

New environment variables: 
```

# Define the cooldown duration, in seconds, for password reset emails
PASSWORD_RESET_EMAIL_COOLDOWN = env.int("PASSWORD_RESET_EMAIL_COOLDOWN", 60 * 60 * 24)

# Limit the count of password reset emails that can be dispatched within the `PASSWORD_RESET_EMAIL_COOLDOWN` timeframe.
MAX_PASSWORD_RESET_EMAILS = env.int("MAX_PASSWORD_RESET_EMAILS", 5)

```

Covered by unit tests


matthewelwell (commented):
Hmm... this feels like a bit overkill to me? I think we can just store the last password reset request time (although I'd argue we'd be better off storing this in a separate table than directly on the user model, as it will give us more flexibility), and then just define an interval between resets of, say, 30 or 60 seconds?

gagantrivedi:
> Hmm... this feels like a bit overkill to me? I think we can just store the last password reset request time (although I'd argue we'd be better off storing this in a separate table than directly on the user model, as it will give us more flexibility), and then just define an interval between resets of, say, 30 or 60 seconds?

I just wanted to completely get rid of this issue -  emails generated per 30 sec or 1 min are still a lot of emails?

api/users/models.py
```diff
@@ -158,6 +168,24 @@ def get_full_name(self):
...
+            if self.last_password_reset_email_at < timezone.now() - timezone.timedelta(
+                seconds=settings.PASSWORD_RESET_EMAIL_COOLDOWN
+            ):
+                return True
+            return (
+                self.num_of_password_reset_emails_sent
+                < settings.MAX_PASSWORD_RESET_EMAILS
+            )
+
+        return True
```
matthewelwell: In response to the main thread on the PR, I tend to agree that we should do slightly more to prevent it but I do think that if we use a separate model this logic will become slightly cleaner / more flexible? 

Something like: 

```python
class UserPasswordResetRequest(models.Model):
    user = models.ForeignKey("users.FFAdminUser", related_name="password_reset_requests")
    requested_at = models.DateTimeField(auto_now_add=True)
```

Then the code above becomes something like: 

```python
def can_send_password_reset_email(self) -> bool:
    limit = timezone.now() - timezone.timedelta(
        seconds=settings.PASSWORD_RESET_EMAIL_COOLDOWN
    )
    return (
        self.password_reset_requests.filter(requested_at__gte=limit).count() 
        > settings.MAX_PASSWORD_RESET_EMAILS
    )
```

Then on a successful password reset, we just clear the table for that user.

(sidenote: as per my comment on another PR, I'd rather not query the database in properties if we can avoid it)

gagantrivedi: Yeah, good point. Done

api/app/settings/common.py
```diff
@@ -903,3 +903,9 @@
 DOMAIN_OVERRIDE = env.str("FLAGSMITH_DOMAIN", "")
 # Used when no Django site is specified.
 DEFAULT_DOMAIN = "app.flagsmith.com"
+
+# Define the cooldown duration, in seconds, for password reset emails
+PASSWORD_RESET_EMAIL_COOLDOWN = env.int("PASSWORD_RESET_EMAIL_COOLDOWN", 60 * 60 * 24)
```
matthewelwell, line 908: I guess there's now a question whether we should add a recurring task to delete all UserPasswordResetRequest entities older than this time frame. Probably one we can worry about in the future but we will end up with stale data in there that likely never gets deleted / used.

api/users/models.py
```diff
@@ -158,6 +163,19 @@ def get_full_name(self):
...
+            seconds=settings.PASSWORD_RESET_EMAIL_COOLDOWN
+        )
+        return (
+            self.password_reset_requests.filter(requested_at__gte=limit).count()
+            < settings.MAX_PASSWORD_RESET_EMAILS
+        )
+
+    def password_reset_email_sent(self):
+        self.num_of_password_reset_emails_sent += 1
+        self.last_password_reset_email_at = timezone.now()
```
matthewelwell: I think this needs to be removed?

matthewelwell (changes requested):

api/users/models.py
gagantrivedi: Yeah, good spot

matthewelwell (approved):

api/app/settings/common.py
gagantrivedi: Yes, likely. I will create an issue for it once this is merged, sounds good?

PR #2641, opened 2023-08-11
refactor: remove tag hack for version number

### LINKED ISSUE #2636: Grab application version from release-please manifest
We don't need to hack tags in from Github any longer as we have https://github.com/Flagsmith/flagsmith/blob/main/.release-please-manifest.json

dabeeeenster:
Docker builds cant `COPY` from parent folders so all the Dockerfiles now need to be run in the root monorepo context so they can grab the `.release-please-manifest.json` file.

This also adds the suite of packages into the version json as can be seen in https://pr-2641-deployment-33281-flagsmith.app.uffizzi.com/version

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,7 +40,7 @@ COPY --from=build-python /usr/local/lib/python3.11/site-packages /usr/local/lib/
 COPY --from=build-python /usr/local/bin /usr/local/bin
 
 COPY api /app/
-COPY .release-please-manifest.json /app/.release-please-manifest.json
+COPY .release-please-manifest.json /app/.versions.json
 
 # Compile static Django assets
 RUN python /app/manage.py collectstatic --no-input
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,6 +40,7 @@ COPY --from=build-python /usr/local/lib/python3.11/site-packages /usr/local/lib/
 COPY --from=build-python /usr/local/bin /usr/local/bin
 
 COPY api /app/
+COPY .release-please-manifest.json /app/.release-please-manifest.json
 
 # Compile static Django assets
 RUN python /app/manage.py collectstatic --no-input
--- a/api/app/utils.py
+++ b/api/app/utils.py
@@ -12,7 +12,7 @@ def create_hash():
 
 def get_version_info() -> dict:
     """Reads the version info baked into src folder of the docker container"""
-    release_please_manifest_location = "./.release-please-manifest.json"
+    release_please_manifest_location = "./.versions.json"
     manifest_versions = None
 
     if os.path.isfile(release_please_manifest_location):
--- a/api/app/utils.py
+++ b/api/app/utils.py
@@ -1,3 +1,5 @@
+import json
+import os
 import pathlib
 
 import shortuuid
@@ -10,11 +12,25 @@ def create_hash():
 
 def get_version_info() -> dict:
     """Reads the version info baked into src folder of the docker container"""
+    release_please_manifest_location = "./.release-please-manifest.json"
+    manifest_versions = None
+
+    if os.path.isfile(release_please_manifest_location):
+        manifest_versions = json.loads(
+            _get_file_contents(release_please_manifest_location)
+        )
+        image_tag = manifest_versions["."]
+    else:
+        image_tag = "unknown"
+
     version_json = {
         "ci_commit_sha": _get_file_contents("./CI_COMMIT_SHA"),
-        "image_tag": _get_file_contents("./IMAGE_TAG"),
+        "image_tag": image_tag,
         "is_enterprise": pathlib.Path("./ENTERPRISE_VERSION").exists(),
     }
+
+    if manifest_versions:
+        version_json["package_versions"] = manifest_versions
 
     return version_json
 
--- a/api/app/utils.py
+++ b/api/app/utils.py
@@ -39,4 +39,4 @@ def _get_file_contents(file_path: str) -> str:
         with open(file_path) as f:
             return f.read().replace("\n", "")
     except FileNotFoundError:
-        return "unknown"
+        return UNKNOWN
--- a/api/app/utils.py
+++ b/api/app/utils.py
@@ -11,14 +11,14 @@ def create_hash():
 
 def get_version_info() -> dict:
     """Reads the version info baked into src folder of the docker container"""
-    release_please_manifest_location = "./.versions.json"
-    manifest_versions = None
+    release_please_manifest_location = ".versions.json"
     version_json = {}
     image_tag = "unknown"
 
-    manifest_versions = json.loads(_get_file_contents(release_please_manifest_location))
+    manifest_versions_string = _get_file_contents(release_please_manifest_location)
 
-    if manifest_versions:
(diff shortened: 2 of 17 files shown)


Dockerfile
```diff
@@ -40,6 +40,7 @@ COPY --from=build-python /usr/local/lib/python3.11/site-packages /usr/local/lib/
 COPY --from=build-python /usr/local/bin /usr/local/bin
 
 COPY api /app/
+COPY .release-please-manifest.json /app/.release-please-manifest.json
```
matthewelwell: Just wondering whether, for aesthetics this should be: 

```
COPY .release-please-manifest.json /app/.versions.json
```

Something just feels a bit ick using release-please-manifest in the actual code.

api/.coverage.Bens-14-M1-Pro.local.82081.235430
matthewelwell, line 1: Think this is a mistake?

api/app/utils.py
```diff
@@ -10,12 +12,26 @@ def create_hash():
 
 def get_version_info() -> dict:
     """Reads the version info baked into src folder of the docker container"""
+    release_please_manifest_location = "./.release-please-manifest.json"
+    manifest_versions = None
+
+    if os.path.isfile(release_please_manifest_location):
+        manifest_versions = json.loads(
+            _get_file_contents(release_please_manifest_location)
```
matthewelwell: We're doing the check twice here since `_get_file_contents` has a `try` / `except` for `FileNotFoundError` which results in `get_file_contents` returning `"unknown"`. With this approach, the path in `_get_file_contents` is redundant.

api/.coverage.Bens-14-M1-Pro.local.82081.235430
dabeeeenster: Oops - removed

Dockerfile
dabeeeenster: I know what you mean, but I think its better than grabbing the tag from Github. It's also the only place we are storing the individual package versions. Will change the filename

api/app/utils.py
dabeeeenster: Tweaked this

api/app/utils.py
```diff
@@ -10,9 +11,20 @@ def create_hash():
 
 def get_version_info() -> dict:
     """Reads the version info baked into src folder of the docker container"""
-    version_json = {
+    release_please_manifest_location = "./.versions.json"
+    manifest_versions = None
+    version_json = {}
+    image_tag = "unknown"
+
+    manifest_versions = json.loads(_get_file_contents(release_please_manifest_location))
```
matthewelwell: This is going to throw a JSONDecodeError if the file doesn't exist: 

```
>>> json.loads("unknown")
  File "/Applications/PyCharm.app/Contents/plugins/python/helpers/pydev/pydevconsole.py", line 364, in runcode
    coro = func()
           ^^^^^^
  File "<input>", line 1, in <module>
  File "/Users/matthewelwell/.asdf/installs/python/3.11.4/lib/python3.11/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/matthewelwell/.asdf/installs/python/3.11.4/lib/python3.11/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/matthewelwell/.asdf/installs/python/3.11.4/lib/python3.11/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

This highlights that we don't have a test which exercises this path!

matthewelwell: This is perhaps a bit neater:

```python
UNKNOWN = "unknown"
VERSIONS_INFO_FILE_LOCATION = "./.versions.json"

...

def get_version_info() -> dict:
    """Reads the version info baked into src folder of the docker container"""
    version_json: dict = {}
    image_tag: str = UNKNOWN

    versions_file_content: str = _get_file_contents(VERSIONS_INFO_FILE_LOCATION)

    if versions_file_content:
        versions_data: dict = json.loads(versions_file_content.replace("\n", ""))
        version_json["package_versions"] = versions_data
        image_tag = versions_data["."]

    return version_json | {
        "ci_commit_sha": _get_file_contents("./CI_COMMIT_SHA") or UNKNOWN,
        "image_tag": image_tag,
        "is_enterprise": pathlib.Path("./ENTERPRISE_VERSION").exists(),
    }


def _get_file_contents(file_path: str) -> typing.Optional[str]:
    """Attempts to read a file from the filesystem and return the contents"""
    try:
        with open(file_path) as f:
            return f.read()
    except FileNotFoundError:
        return None
```

matthewelwell (changes requested):

api/app/utils.py
dabeeeenster: OK refactored

matthewelwell: Well, you've _half_ refactored. I'd rather that the `_get_file_contents` function was so logic-aware, but whatever.

matthewelwell: I would really like to see a test that exercises this code if the file doesn't exist though!

dabeeeenster: OK test added

matthewelwell:
@dabeeeenster this probably needs changing again after the changes we made to the release please configuration. WE can just use `version.txt` now, right?

dabeeeenster:
We can still use `.release-please-manifest.json` - I've updated the tests to not expect the individual package versions.

api/app/utils.py
```diff
@@ -1,7 +1,11 @@
+import json
 import pathlib
 
 import shortuuid
 
+UNKNOWN = "unknown"
+VERSIONS_INFO_FILE_LOCATION = ".versions.json"
```
matthewelwell, line 7: Not really sure where this file path has come from? It should be `.release-please-manifest.json` or, reiterating my previous comment, `version.txt` which would some of the code here?

api/app/utils.py
```diff
@@ -1,7 +1,11 @@
+import json
 import pathlib
 
 import shortuuid
 
+UNKNOWN = "unknown"
```
matthewelwell, line 6: There's a hardcoded `"unknown"` string at the bottom of this file too. Can we update all usages to use this constant?

matthewelwell (changes requested):

api/app/utils.py
matthewelwell: Ah, just remembered it's part of the docker build I think? I still think it would be better to use the `version.txt` file though.

dabeeeenster: Yep the modified Dockerfiles do this. I think its better to stay as json as it's easier to parse across both JS and Python.

api/app/utils.py
dabeeeenster: OK that's done

matthewelwell (approved):

PR #2642, opened 2023-08-14
fix(user-create): duplicate email error message

gagantrivedi:

## Changes

Return a generic error message for user create if the email is not unique 

covered by unit test cases and verified manually by creating a user with duplicate email

### DIFF
--- a/api/custom_auth/serializers.py
+++ b/api/custom_auth/serializers.py
@@ -1,8 +1,10 @@
 from django.conf import settings
+from djoser.constants import Messages
 from djoser.serializers import UserCreateSerializer, UserDeleteSerializer
 from rest_framework import serializers
 from rest_framework.authtoken.models import Token
 from rest_framework.exceptions import PermissionDenied
+from rest_framework.validators import UniqueValidator
 
 from organisations.invites.models import Invite
 from users.constants import DEFAULT_DELETE_ORPHAN_ORGANISATIONS_VALUE
@@ -18,6 +20,8 @@ class CustomTokenSerializer(serializers.ModelSerializer):
 
 
 class CustomUserCreateSerializer(UserCreateSerializer):
+    key = serializers.SerializerMethodField()
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.fields["key"] = serializers.SerializerMethodField()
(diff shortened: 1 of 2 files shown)


api/custom_auth/serializers.py
```diff
@@ -18,6 +20,8 @@ class Meta:
 
 
 class CustomUserCreateSerializer(UserCreateSerializer):
+    key = serializers.SerializerMethodField()
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.fields["key"] = serializers.SerializerMethodField()
```
matthewelwell: I'm not sure why we were ever doing this in the `init` method, but it seems like we can probably remove this logic?

gagantrivedi: Yeah, I was going to ask you if you can remember

api/custom_auth/serializers.py
```diff
@@ -18,17 +20,30 @@ class Meta:
 
 
 class CustomUserCreateSerializer(UserCreateSerializer):
+    key = serializers.SerializerMethodField()
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
```
matthewelwell: I don't think this is doing anything anymore? We can get rid of the whole `__init__` method.

api/custom_auth/serializers.py
```diff
@@ -18,17 +20,30 @@ class Meta:
...
         )
         read_only_fields = ("is_active",)
         write_only_fields = ("sign_up_type",)
+        extra_kwargs = {
+            "email": {
+                "validators": [
+                    UniqueValidator(
+                        queryset=FFAdminUser.objects.all(),
+                        lookup="iexact",
+                        message=djoser_settings.CONSTANTS.messages.CANNOT_CREATE_USER_ERROR,
```
matthewelwell, line 39: How come this changed from `Messages.CANNOT_CREATE_USER_ERROR` ? Possibly so that we can tweak it in the settings object?

matthewelwell (changes requested):

api/custom_auth/serializers.py
gagantrivedi: Yeah, exactly

matthewelwell (approved):

PR #2646, opened 2023-08-14
feat: admin action to delete all segments for project

matthewelwell:
## Changes

Adds an admin action to delete all segments for a project.

This is mostly a support tool for a particular customer that breached the limit of segments and we need to clear them down. Since django handles cascading manually, it is challenging / risky to delete them manually in the database.

Added unit test and tested manually.


gagantrivedi (approved):

api/projects/admin.py
```diff
@@ -9,6 +11,10 @@
 from projects.tags.models import Tag
 from segments.models import Segment
 
+if typing.TYPE_CHECKING:
```
gagantrivedi: Why are we only importing them if TYPE_CHECKING?

gagantrivedi (approved):
one small comment

api/projects/admin.py
matthewelwell: Because we only need them for typing.

gagantrivedi: Is that the convention? Because both of them are going to be available at runtime anyway, and that `if` is a bit ugly? What do you think?

matthewelwell: Yeah, fair enough - I'm not too sure what the convention is. Seems like it should only really be used to avoid circular dependencies. I'll remove.

gagantrivedi (approved):

PR #2656, opened 2023-08-16
test: add more tests for feature versioning logic

matthewelwell:
## Changes

This PR is primarily for adding more tests to the feature versioning logic. It also adds one extra field to the FeatureSegment model to fix issues with uniqueness when creating multiple versions of segment overrides. 

NOTE: This PR is on top of #2382 - once approved, I will merge into there and merge them together. 

This PR is 90% tests 😄


api/features/models.py
```diff
@@ -220,12 +227,17 @@ class FeatureSegment(
...
 
     objects = FeatureSegmentManager()
 
     class Meta:
-        unique_together = ("feature", "environment", "segment")
+        unique_together = (
+            "feature",
+            "environment",
+            "segment",
+            "environment_feature_version",
```
matthewelwell, line 239: The `sqlmigrate` for the migration for this (and the new field) is as follows: 

```
BEGIN;
--
-- Add field environment_feature_version to featurestate
--
ALTER TABLE "features_featurestate" ADD COLUMN "environment_feature_version_id" uuid NULL CONSTRAINT "features_featurestat_environment_feature__0b08c62e_fk_feature_v" REFERENCES "feature_versioning_environmentfeatureversion"("uuid") DEFERRABLE INITIALLY DEFERRED; SET CONSTRAINTS "features_featurestat_environment_feature__0b08c62e_fk_feature_v" IMMEDIATE;
--
-- Add field environment_feature_version to historicalfeaturestate
--
ALTER TABLE "features_historicalfeaturestate" ADD COLUMN "environment_feature_version_id" uuid NULL;
CREATE INDEX "features_featurestate_environment_feature_version_id_0b08c62e" ON "features_featurestate" ("environment_feature_version_id");
CREATE INDEX "features_historicalfeature_environment_feature_versio_5d9a2014" ON "features_historicalfeaturestate" ("environment_feature_version_id");
COMMIT;
```

I think the modification of the index will cause the table to be locked for a very short amount of time, however, since we only have 17k records in the FeatureSegment table, I think this is an acceptable risk. Keen to know what you think here though @gagantrivedi @khvn26 ?

dabeeeenster:
> This PR is 90% tests 😄

Test the tests tho

api/features/models.py
gagantrivedi: I don't think any of them should lock the table? I don't think create index runs concurrently unless specified explicitly?

gagantrivedi (approved):

api/features/models.py
matthewelwell: Sorry, that's the wrong sql migrate 🤦 this is the right one: 

```sql
BEGIN;
--
-- Add field environment_feature_version to featuresegment
--
ALTER TABLE "features_featuresegment" ADD COLUMN "environment_feature_version_id" uuid NULL CONSTRAINT "features_featuresegm_environment_feature__11efe443_fk_feature_v" REFERENCES "feature_versioning_environmentfeatureversion"("uuid") DEFERRABLE INITIALLY DEFERRED; SET CONSTRAINTS "features_featuresegm_environment_feature__11efe443_fk_feature_v" IMMEDIATE;
--
-- Add field environment_feature_version to historicalfeaturesegment
--
ALTER TABLE "features_historicalfeaturesegment" ADD COLUMN "environment_feature_version_id" uuid NULL;
--
-- Alter unique_together for featuresegment (1 constraint(s))
--
ALTER TABLE "features_featuresegment" DROP CONSTRAINT "features_featuresegment_feature_id_environment_i_e5499ca3_uniq";
ALTER TABLE "features_featuresegment" ADD CONSTRAINT "features_featuresegment_feature_id_environment_i_f1fde686_uniq" UNIQUE ("feature_id", "environment_id", "segment_id", "environment_feature_version_id");
CREATE INDEX "features_featuresegment_environment_feature_version_id_11efe443" ON "features_featuresegment" ("environment_feature_version_id");
CREATE INDEX "features_historicalfeature_environment_feature_versio_88d80be3" ON "features_historicalfeaturesegment" ("environment_feature_version_id");
COMMIT;
```

Specifically, it's the `DROP CONSTRAINT` / `ADD CONSTRAINT` that I'm worried about.

matthewelwell: Ok, I've written some manual SQL for the migrations now instead. I think that the `ADD CONSTRAINT` operation will still require an exclusive lock on the table, but only for a very short time. See latest commit [here](https://github.com/Flagsmith/flagsmith/pull/2656/commits/7f7439c81b9ebf6c080b0f2aae2fffb5b800b605).

gagantrivedi (approved):

PR #2631, opened 2023-08-10
feat: re-add totals and limits

matthewelwell:
## Changes

Re-add totals and limits removed in #2625 due to performance issues. 

Uses separate queries instead of `annotate(Count(..., distinct=True))` to avoid further performance issues.

## How did you test this code?

Updated the test from the previous iteration to ensure its verifying that the numbers are correct. I've been unable to really reproduce the performance issues we saw in production, but there is a lot less black magic going on with this approach. The additional queries should have a negligible impact on performance.

PR #2663, opened 2023-08-17
test: add retrieve environment test

matthewelwell:
## Changes

Adds a new test that was missing (as discovered when responding to comments on #2631)

This PR consists of a single unit test only.

### DIFF
--- a/api/tests/unit/environments/test_unit_environments_views.py
+++ b/api/tests/unit/environments/test_unit_environments_views.py
@@ -1,0 +1,47 @@
+from django.urls import reverse
+from rest_framework import status
+from rest_framework.test import APIClient
+
+from environments.models import Environment
+
+
+def test_retrieve_environment(
+    admin_client: APIClient, environment: Environment
+) -> None:
+    # Given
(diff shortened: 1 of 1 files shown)


api/tests/unit/environments/test_unit_environments_views.py
```diff
@@ -0,0 +1,47 @@
+from django.urls import reverse
+from rest_framework import status
+from rest_framework.test import APIClient
+
+from environments.models import Environment
+
+
+def test_retrieve_environment(
```
novakzaballa, line 8: Should this test be named `test_retrieve_environment_details` for clarity?

novakzaballa (approved):
LGTM with a comment

api/tests/unit/environments/test_unit_environments_views.py
matthewelwell: I guess it could be, yes. I would argue that it's in a file called `test_unit_environments_views` and `retrieve` is the name of the view action we're testing, so it's not necessary to add `_details` here.

PR #2684, opened 2023-08-25
fix: create initial versions logic and correct filters for get environment flags

matthewelwell:
## Changes

This PR addresses 3 things: 

 1. The logic to enable v2 versioning was flawed in that there would be a short window while the application was creating the initial versions in which the application would likely return incorrect values. I have resolved this by creating a new endpoint to enable v2 versioning and handling the process in a single task. 
 2. The logic to create initial versions previously didn't factor in the previous versioning logic. I have updated this now to only add the initial versions to the latest versions of flags. 
 3. Corrected the filters to get environment flags. Although this shouldn't have any material effect, it is cleaner to use the same filters between environment flags and identity flags. 

Updated / added tests where necessary.

### DIFF
--- a/api/environments/views.py
+++ b/api/environments/views.py
@@ -6,11 +6,12 @@ import logging
 from django.db.models import Count
 from django.utils.decorators import method_decorator
 from drf_yasg import openapi
-from drf_yasg.utils import swagger_auto_schema
+from drf_yasg.utils import no_body, swagger_auto_schema
 from rest_framework import mixins, status, viewsets
 from rest_framework.decorators import action
 from rest_framework.exceptions import ValidationError
 from rest_framework.permissions import IsAuthenticated
+from rest_framework.request import Request
 from rest_framework.response import Response
 
 from environments.permissions.permissions import (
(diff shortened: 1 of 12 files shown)


api/environments/views.py
```diff
@@ -203,6 +205,13 @@ def user_permissions(self, request, *args, **kwargs):
     def get_document(self, request, api_key: str):
         return Response(Environment.get_environment_document(api_key))
 
+    @swagger_auto_schema(request_body=no_body, responses={202: ""})
+    @action(detail=True, methods=["POST"], url_path="enable-v2-versioning")
+    def enable_v2_versioning(self, request: Request, api_key: str) -> Response:
+        environment = self.get_object()
```
gagantrivedi, line 211: This method is not idempotent, right? We will have multiple initial feature versions if the user makes the call multiple times(I don't think it breaks the code - but it does, create extra data) ?

gagantrivedi (changes requested):

api/environments/views.py
matthewelwell: Yep, good catch! Resolved.

gagantrivedi (approved):

PR #2689, opened 2023-08-28
test: Split E2E test into different jobs

novakzaballa:

## Changes

Tests are now split into:
- Segment-part-1
- Segment-part-2
- Segment-part-3
- Flag
- Signup
- Invite
- Project

### Improvements

- Tests can now be executed in concurrent (multiple browsers) mode locally and in Github Actions, and in parallel mode (multiple jobs) in Github Actions.
- Since tests are now split, it is easier to find the error in the video and the logs.
- Now you need to re-run only the tests failing not the whole set of E2E tests
- The total time of execution is about half of the prior version.
- We are using Docker BuildX in every job and step and we are caching layers to speedup the process.

### Concurrency
By default, the E2E tests are executed in concurrent mode with **3** browsers.

#### Running all tests in the local dev environment:
In order to run all the E2E tests set the necessary environment variables:
For the API:
```
export E2E_TEST_AUTH_TOKEN='some-token'
export ENABLE_FE_E2E='True'
export FLAGSMITH_API_URL=http://127.0.0.1:8000/api/v1/
```
For the Frontend:
```
export FLAGSMITH_API_URL=http://127.0.0.1:8000/api/v1/
export E2E_TEST_TOKEN='some-token' # Only if it is a different environment from API
export ENV=local
npm run env (execute only once with ENV=local)
```
and execute:
```
npm run test:devBundle
```
#### Running only `Invite `and `Environment` E2E tests with two browsers, notice that you can use lowercased test names:
`ENV=local npx cross-env E2E_CONCURRENCY=2 npm run test:devBundle -- invite environment
`

### Running E2E tests in a container (no need to have node installed)
In the frontend folder:
```
docker buildx bake -f docker-compose-e2e-tests.yml
docker compose -f docker-compose-e2e-tests.yml run frontend npm run test
```
### Notes

- While the most common sources of flaky tests were solved, React node not being completely rendered, and button not being hit, and some events results not being verified before moving to the next test action, there could still be some flaky tests, however, now should be easier to detect them as well as simply re-run flaky tests.
- Segment-part-2, Segment-part-1, and Flags are long-lasting E2E tests, although the two segments take much more than Flags. Because of that, 3 browsers are configured by default, since Segments-part-2 last almost the same as Segment-part-1, which takes almost the same time as the rest of the tests.

Tested during Github Actions CI/CD execution

### DIFF
--- a/.github/actions/local-e2e-tests/action.yml
+++ b/.github/actions/local-e2e-tests/action.yml
@@ -1,0 +1,55 @@
+name: Run E2E tests in local environment
+description: Run the E2E tests against local environment
+
+inputs:
+  sentry_sdk_dsn:
+    description: The sentry SDK DSN
+    required: true
+  sendgrid_api_key:
+    description: The sendgrid API key to use for email sending
+    required: true
+  slack_token:
+    description: The slack authentication token.
+    required: true
+  tests:
+    description: The space separated list of E2E tests to be executed.
+    required: false
+    default: ''
+  concurrency:
+    description: The concurrent number of browsers to be used on testing.
+    required: false
+    default: 3
+
+runs:
+  using: composite
+
+  steps:
+    - name: Run Local API
+      id: run-local-api
+      uses: ./.github/actions/run-local-api
+      with:
+        e2e_test_token: some-token
+        # As per https://stackoverflow.com/q/65497331/421808 172.17.0.1 seems like the only way to resolve host DB
+        database_url: postgres://postgres:postgres@172.17.0.1:5432/flagsmith
+        sentry_sdk_dsn: ${{ inputs.sentry_sdk_dsn }}
+        sendgrid_api_key: ${{ inputs.sendgrid_api_key }}
+        disable_analytics_features: true
+
+    - name: Run E2E tests against local
+      uses: ./.github/actions/e2e-tests
+      env:
+        E2E_CONCURRENCY: ${{ inputs.concurrency }}
+      with:
+        e2e_test_token: some-token
+        slack_token: ${{ inputs.slack_token }}
+        environment: local
+        tests: ${{ inputs.tests }}
+
+    - name: Output API container status and logs
+      if: failure()
+      env:
+        API_CONTAINER_ID: ${{ steps.run-local-api.outputs.containerId }}
+      run: |
+        docker inspect $API_CONTAINER_ID | jq '.[0].State'
+        docker logs $API_CONTAINER_ID
+      shell: bash
--- a/.github/actions/local-e2e-tests/action.yml
+++ b/.github/actions/local-e2e-tests/action.yml
@@ -2,12 +2,6 @@ name: Run E2E tests in local environment
 description: Run the E2E tests against local environment
 
 inputs:
-  sentry_sdk_dsn:
-    description: The sentry SDK DSN
-    required: true
-  sendgrid_api_key:
-    description: The sendgrid API key to use for email sending
-    required: true
   slack_token:
     description: The slack authentication token.
     required: true
@@ -31,8 +25,6 @@ runs:
         e2e_test_token: some-token
         # As per https://stackoverflow.com/q/65497331/421808 172.17.0.1 seems like the only way to resolve host DB
         database_url: postgres://postgres:postgres@172.17.0.1:5432/flagsmith
-        sentry_sdk_dsn: ${{ inputs.sentry_sdk_dsn }}
-        sendgrid_api_key: ${{ inputs.sendgrid_api_key }}
         disable_analytics_features: true
 
     - name: Run E2E tests against local
--- a/.github/actions/unified-e2e-tests/action.yml
+++ b/.github/actions/unified-e2e-tests/action.yml
@@ -1,0 +1,40 @@
+name: Run E2E tests in docker unified environment
+description: Run the E2E tests against docker unified environment
+
+inputs:
+  github_actor:
+    description: Github actor
+    required: true
+  github_token:
+    description: Github token
+    required: true
+  e2e_test_token_staging:
+    description: The staging test token.
+    required: true
+
+runs:
+  using: composite
+
+  steps:
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v2
+
+    - name: Expose GitHub Runtime
+      uses: crazy-max/ghaction-github-runtime@v2
+
+    - name: Login to GitHub Container Registry
+      uses: docker/login-action@v2
+      with:
+        registry: ghcr.io
+        username: ${{ inputs.github_actor }}
+        password: ${{ inputs.github_token }}
+
+    - name: Build docker-compose with unified-image
+      working-directory: frontend
+      env:
+        E2E_TEST_TOKEN_STAGING: ${{ inputs.e2e_test_token_staging }}
+        ENV: staging
+        STATIC_ASSET_CDN_URL: /
+      run: |
+        docker buildx bake -f docker-compose-e2e-tests.yml --load
+      shell: bash
--- a/.github/workflows/platform-pull-request.yml
+++ b/.github/workflows/platform-pull-request.yml
@@ -29,9 +29,9 @@ jobs:
             style
             test
 
-  run-e2e-tests:
-    runs-on: e2e-runner-8c
-    name: Full E2E tests
+  run-e2e-tests-1:
+    runs-on: ubuntu-latest
+    name: E2E Segment Part 1 and environment tests
 
     services:
       postgres:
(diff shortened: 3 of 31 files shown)


.github/actions/local-e2e-tests/action.yml
```diff
@@ -0,0 +1,55 @@
...
+  steps:
+    - name: Run Local API
+      id: run-local-api
+      uses: ./.github/actions/run-local-api
+      with:
+        e2e_test_token: some-token
+        # As per https://stackoverflow.com/q/65497331/421808 172.17.0.1 seems like the only way to resolve host DB
+        database_url: postgres://postgres:postgres@172.17.0.1:5432/flagsmith
+        sentry_sdk_dsn: ${{ inputs.sentry_sdk_dsn }}
+        sendgrid_api_key: ${{ inputs.sendgrid_api_key }}
```
matthewelwell: Do we need these?

.github/actions/unified-e2e-tests/action.yml
```diff
@@ -0,0 +1,40 @@
...
+
+runs:
+  using: composite
+
+  steps:
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v2
+
+    - name: Expose GitHub Runtime
+      uses: crazy-max/ghaction-github-runtime@v2
```
matthewelwell, line 23: What does this step do / what do we need it for?

.github/workflows/platform-pull-request.yml
```diff
@@ -29,9 +29,9 @@ jobs:
             style
             test
 
-  run-e2e-tests:
+  run-e2e-tests-1:
     runs-on: ubuntu-latest
-    name: Full E2E tests
+    name: E2E Segment Part 1 and environment tests
```
matthewelwell: These names should all be standardised a little bit. Perhaps something like: 

E2E Local: Segment Part 1 & Environment
...
E2E Unified: Segment Part 3, Signup, Flags, Invite, Project

.github/workflows/platform-pull-request.yml
```diff
@@ -47,58 +47,137 @@ jobs:
...
+          sentry_sdk_dsn: ${{ secrets.SENTRY_SDK_DSN }}
+          sendgrid_api_key: ${{ secrets.SENDGRID_API_KEY }}
           slack_token: ${{ secrets.SLACK_TOKEN }}
-          environment: local
+          tests: segment-part-2
+          concurrency: 1
+
+  run-e2e-tests-3:
+    runs-on: ubuntu-latest
+    name: E2E segments-3, signup,flag, invite, project tests on local
```
matthewelwell: What is the reason for grouping these together?

frontend/Dockerfile.e2e
```diff
@@ -2,11 +2,18 @@ FROM ghcr.io/flagsmith/e2e-frontend-base:latest
...
 # Build Flagsmith
 WORKDIR /srv/flagsmith
-COPY . .
-RUN npm cache clean --force
-RUN npm ci
+COPY package.json package-lock.json .npmrc .nvmrc ./
+COPY bin/ ./bin/
+COPY env/ ./env/
 
+# Run npm ci before copying source code to improve caching of layers. Cache clean no needed?
```
matthewelwell: I don't love the idea of adding questions in comments like this - can we answer the question?

frontend/Dockerfile.e2e
```diff
@@ -2,11 +2,18 @@ FROM ghcr.io/flagsmith/e2e-frontend-base:latest
...
+# RUN npm cache clean --force
 ENV ENV=e2e
-RUN npm run env
+RUN npm ci
+
+
+COPY . .
+# Commented because it is already executed on the postinstall hook of npm ci
+# RUN npm run env
```
matthewelwell: Then let's get rid of it rather than commenting it out, right?

frontend/docker-compose-e2e-tests.yml
```diff
@@ -16,6 +16,13 @@ services:
     build:
       context: ../
       dockerfile: Dockerfile
+      x-bake:
+        cache-from:
+          - type=gha,scope=image-e2e-unified-api
+        cache-to:
+          - type=gha,mode=max,scope=image-e2e-unified-api
+        tags:
+          - frontend_flagsmith-api
```
matthewelwell, line 25: I think this will make this docker-compose file unusable locally, right? Should we actually create a brand new docker-compose file and store it in the `.github` directory or something?

frontend/docker-compose-e2e-tests.yml
```diff
@@ -43,7 +58,9 @@ services:
     ports:
       - 8080:8080
     depends_on:
-      - flagsmith-api
+      flagsmith-api:
+        condition: service_started
```
matthewelwell: Does this offer anything additional compared to how it was before?

api/e2etests/e2e_seed_data.py
```diff
@@ -0,0 +1,43 @@
...
+    password = "str0ngp4ssw0rd!"
+
+    def teardown(self):
+        # delete users created for e2e test by front end
+        if self.fe_e2e_test_user_email:
+            FFAdminUser.objects.filter(email=self.fe_e2e_test_user_email).delete()
+            FFAdminUser.objects.filter(
+                email=self.fe_e2e_test_user_email + ".io"
+            ).delete()
+            FFAdminUser.objects.filter(email="changeMail@test.com").delete()
```
matthewelwell: I think it would be better to build (and explain) the different emails in an __init__ method on this class?

api/e2etests/e2e_seed_data.py
```diff
@@ -0,0 +1,43 @@
...
+        else None
+    )
+    password = "str0ngp4ssw0rd!"
+
+    def teardown(self):
+        # delete users created for e2e test by front end
+        if self.fe_e2e_test_user_email:
+            FFAdminUser.objects.filter(email=self.fe_e2e_test_user_email).delete()
+            FFAdminUser.objects.filter(
+                email=self.fe_e2e_test_user_email + ".io"
```
matthewelwell: It's a bit odd to me that we're adding ".io" here, couldn't we just have [test@example.com](mailto:test@example.com) and [test2@example.com](mailto:test2@example.com) or something?

api/e2etests/views.py
```diff
@@ -19,31 +18,9 @@ class Teardown(APIView):
...
-        project = Project.objects.create(
-            name="My Test Project", organisation=organisation
-        )
-
-        Environment.objects.create(name="Development", project=project)
-
-        Environment.objects.create(name="Production", project=project)
-
+        e2e_seed_data = E2ESeedData()
+        e2e_seed_data.teardown()  # Llama al método teardown de E2ESeedData
```
matthewelwell: 🇪🇸 

Let's remove this comment as it's not adding anything (and try to keep comments in English! 😄 )

api/e2etests/views.py
```diff
@@ -19,31 +18,9 @@ class Teardown(APIView):
...
-
-        project = Project.objects.create(
-            name="My Test Project", organisation=organisation
-        )
-
-        Environment.objects.create(name="Development", project=project)
-
-        Environment.objects.create(name="Production", project=project)
-
+        e2e_seed_data = E2ESeedData()
```
matthewelwell: I think this can be instantiated globally, right? It doesn't need to be done at runtime?

matthewelwell (changes requested):

.github/actions/local-e2e-tests/action.yml
novakzaballa: THey are used here https://github.com/Flagsmith/flagsmith/blob/df6a7b799fca4de16811af3307d95bf75ecc0927/.github/actions/run-local-api/action.yml#L51-L52

.github/actions/unified-e2e-tests/action.yml
novakzaballa: It exposes $ACTIONS_CACHE_URL and $ACTIONS_RUNTIME_TOKEN needed by Docker buildx to authenticate in order to cache Docker layers. See https://docs.docker.com/build/cache/backends/gha/#authentication

.github/workflows/platform-pull-request.yml
novakzaballa: Its is just for efficiency. As explained in the description, segments-2 and segments-1 are the two tests that take the most, so an efficient way to arrange these tests is having 3 instances running in parallel one with the segment-2, another with segment-1 plus a small test, and a third one with all the remaining tests since they all take a bit less than segment-2 to run. Any way, they are independent tests, so will emit separated videos if they fail, actually only the failed tests will generate a video.

.github/workflows/platform-pull-request.yml
novakzaballa: Right, I will update that.

novakzaballa: Done

api/e2etests/views.py
novakzaballa: Oops, yes, left behind by mistake it seems, sorry

api/e2etests/views.py
novakzaballa: Yes

frontend/Dockerfile.e2e
novakzaballa: Yes

frontend/Dockerfile.e2e
novakzaballa: Right

frontend/docker-compose-e2e-tests.yml
novakzaballa: It specifies that docker should wait until the flagsmith-api service is started. In CI/CD doesn't seem to mpact, however testing locally it does.

frontend/docker-compose-e2e-tests.yml
novakzaballa: Nope, I tested all this changes locally, but we should use `docker compose` instead of `docker-compose` since I am not sure if the latter accepts this file format.

api/e2etests/e2e_seed_data.py
novakzaballa: Ok, I will do it

.github/actions/local-e2e-tests/action.yml
matthewelwell: Ok, I don't think they are necessary though. Can you remove them from here and let's see if the tests break? I would like to then (in a separate PR) remove them from everywhere else.

.github/workflows/platform-pull-request.yml
```diff
@@ -47,58 +47,137 @@ jobs:
...
+        with:
+          sentry_sdk_dsn: ${{ secrets.SENTRY_SDK_DSN }}
+          sendgrid_api_key: ${{ secrets.SENDGRID_API_KEY }}
+          slack_token: ${{ secrets.SLACK_TOKEN }}
+          tests: segment-part-3 signup flag invite project
+          concurrency: 2
+
+  run-e2e-segments-1-tests-docker-unified:
+    runs-on: ubuntu-latest
+    name: Unified E2E - Segments-1, Environment
```
matthewelwell:
```suggestion
    name: E2E Unified - Segments-1, Environment
```

.github/workflows/platform-pull-request.yml
```diff
@@ -47,58 +47,137 @@ jobs:
...
+        working-directory: frontend
+        env:
+          SLACK_TOKEN: ${{ inputs.slack_token }}
+          GITHUB_ACTION_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+        run: |
+          docker compose -f docker-compose-e2e-tests.yml run frontend npx cross-env E2E_CONCURRENCY=1 npm run test -- segment-part-1 environment
+
+  run-e2e-segments-2-tests-docker-unified:
+    runs-on: ubuntu-latest
+    name: Unified E2E - Segments-2
```
matthewelwell:
```suggestion
    name: E2E Unified - Segments-2
```

.github/workflows/platform-pull-request.yml
```diff
@@ -47,58 +47,137 @@ jobs:
...
         run: |
-          docker inspect $API_CONTAINER_ID | jq '.[0].State'
-          docker logs $API_CONTAINER_ID
+          docker compose -f docker-compose-e2e-tests.yml run frontend npx cross-env E2E_CONCURRENCY=1 npm run test -- segment-part-2
 
-  run-e2e-tests-docker-unified:
+  run-e2e-other-tests-docker-unified:
     runs-on: ubuntu-latest
-    name: Full E2E tests with unified image in Docker
+    name: Unified E2E - Segments-3, Signup, Flag, Invite, Project
```
matthewelwell:
```suggestion
    name: E2E Unified - Segments-3, Signup, Flag, Invite, Project
```

frontend/Dockerfile.e2e
```diff
@@ -2,11 +2,15 @@ FROM ghcr.io/flagsmith/e2e-frontend-base:latest
...
 # Build Flagsmith
 WORKDIR /srv/flagsmith
-COPY . .
-RUN npm cache clean --force
-RUN npm ci
+COPY package.json package-lock.json .npmrc .nvmrc ./
+COPY bin/ ./bin/
+COPY env/ ./env/
 
+# Run npm ci before copying source code to improve caching of layers. No need to clean cache.
```
matthewelwell:
```suggestion
# Run npm ci before copying source code to improve caching of layers.
```

Either we should explain why there's no need to clean the cache (if it's unusual to not do so), or we can just omit that statement from the comment.

frontend/Dockerfile.e2e
novakzaballa: Corrected

api/e2etests/e2e_seed_data.py
```diff
@@ -0,0 +1,53 @@
...
+        project: Project = Project.objects.create(
+            name="My Test Project", organisation=organisation
+        )
+        Environment.objects.create(name="Development", project=project)
+        Environment.objects.create(name="Production", project=project)
+
+        # Upgrade organisation seats
+        Subscription.objects.filter(
+            organisation__in=org_admin.organisations.all()
+        ).update(max_seats=2)
```
matthewelwell: IMO this shouldn't be a class since everything is static. We can change this to look like the following, and then just use the functions directly in other places. 


```suggestion
PASSWORD = "str0ngp4ssw0rd!"


def teardown(self) -> None:
      # delete users created for e2e test by front end
      FFAdminUser.objects.filter(email=E2E_SIGNUP_USER).delete()
      FFAdminUser.objects.filter(email=E2E_USER).delete()
      FFAdminUser.objects.filter(email=E2E_CHANGE_EMAIL_USER).delete()

def seed_data(self) -> None:
      # create user and organisation for e2e test by front end
      organisation: Organisation = Organisation.objects.create(
          name="Bullet Train Ltd"
      )
      org_admin: FFAdminUser = FFAdminUser.objects.create_user(
          email=E2E_USER,
          password=self.password,
          username=E2E_USER,
      )
      org_admin.add_organisation(organisation, OrganisationRole.ADMIN)

      project: Project = Project.objects.create(
          name="My Test Project", organisation=organisation
      )
      Environment.objects.create(name="Development", project=project)
      Environment.objects.create(name="Production", project=project)

      # Upgrade organisation seats
      Subscription.objects.filter(
          organisation__in=org_admin.organisations.all()
      ).update(max_seats=2)
```

Another thing to note is that we should probably clean the org data up as well, right? So, before we delete the users, we should delete any organisations that they belong to.

frontend/Dockerfile.e2e
novakzaballa: Done

api/e2etests/e2e_seed_data.py
novakzaballa: Right. Done

api/e2etests/e2e_seed_data.py
novakzaballa: Done

api/e2etests/views.py
novakzaballa: Done

api/e2etests/views.py
```diff
@@ -5,55 +5,20 @@
...
-
+from .e2e_seed_data import E2ESeedData
 from .permissions import E2ETestPermission
 
 
 class Teardown(APIView):
     schema = None
     permission_classes = (E2ETestPermission,)
     authentication_classes = (TokenAuthentication,)
+    e2e_seed_data = E2ESeedData()
```
matthewelwell: Based on a previous comment, I don't think this will be necessary but, for education purposes, this should either be at the module level, or in an `__init__` method. Right now, it's a possible issue because python will create a singleton `E2ESeedData` instance when it loads the module that will then be _shared_ between all instances of the `Teardown` API class. In this scenario, that's probably fine because the `E2ESeedData` class is only concerned with static data, but this could cause nasty issues in other scenarios.

frontend/docker-compose-e2e-tests.yml
matthewelwell: Isn't that what it did before though? What is the difference between these two:

```yaml
depends_on:
  - flagsmith-api
```

and 

```yaml
depends_on:
  flagsmith_api:
    condition: service_started
```

frontend/e2e/init.cafe.js
```diff
@@ -1,28 +1,37 @@
...
+const e2eTestApi = `${Project.api}e2etests/teardown/`
 const logger = getLogger()
 
+console.log(
+  '\n',
+  '\x1b[32m',
+  `E2E using API: ${e2eTestApi}. E2E URL: ${url}`,
+  '\x1b[0m',
+  '\n',
+)
```
matthewelwell, line 26: I'm not sure what this messy console.log is doing? Should it still be here?

.github/actions/local-e2e-tests/action.yml
novakzaballa: Ok

api/e2etests/views.py
novakzaballa: E2eSeeData is now a module with standalone functions.

api/e2etests/e2e_seed_data.py
novakzaballa: Yes. Done

.github/workflows/platform-pull-request.yml
novakzaballa: Done

.github/workflows/platform-pull-request.yml
novakzaballa: Done

.github/workflows/platform-pull-request.yml
novakzaballa: Done

.github/actions/local-e2e-tests/action.yml
novakzaballa: Removed, I don't think they are present anywhere else.

frontend/docker-compose-e2e-tests.yml
novakzaballa: Right, no difference is documented. I changed `service_started` with `service_healthy` and added `healthcheck` to `flagsmith-api` so we can test this locally. At least in my environments runs with this modification.

frontend/Dockerfile.e2e
novakzaballa: Done

frontend/e2e/init.cafe.js
novakzaballa: It is displaying some highlighted logs, I think we should keep it since remarks some important info in a large log.

api/e2etests/e2e_seed_data.py
```diff
@@ -0,0 +1,51 @@
...
+PASSWORD = "str0ngp4ssw0rd!"
+
+
+def delete_user_and_its_organisations(user_email: str) -> None:
+    user = FFAdminUser.objects.filter(email=user_email).first()
+    if user:
+        for organisation in Organisation.objects.filter(
+            id__in=user.organisations.all()
+        ):
+            organisation.delete()
```
matthewelwell:
```suggestion
    user.organisations.all().delete()
```

api/e2etests/e2e_seed_data.py
```diff
@@ -0,0 +1,51 @@
+from app.settings.common import (
+    E2E_CHANGE_EMAIL_USER,
+    E2E_SIGNUP_USER,
+    E2E_USER,
+)
```
matthewelwell: erk, NO!! settings must be imported like: 

```
```

then used as `settings.E2E_USER` for example.

api/e2etests/views.py
```diff
@@ -19,41 +15,9 @@ class Teardown(APIView):
...
-
-        Environment.objects.create(name="Production", project=project)
-
-        return Response(status=status.HTTP_204_NO_CONTENT)
-
-
-class UpdateSeats(APIView):
-    permission_classes = (E2ETestPermission,)
-    authentication_classes = (TokenAuthentication,)
+        if "ENABLE_FE_E2E" not in os.environ:
```
matthewelwell: This should be added as a setting in `app.settings.common`.

frontend/docker-compose-e2e-tests.yml
```diff
@@ -16,24 +16,45 @@ services:
...
     depends_on:
       - db
     links:
       - db:db
+    healthcheck:
+      test: "[ -e /dev/shm/log.txt ] && exit 0 || exit 1"
+      start_period: 60s
+      interval: 10s
+      timeout: 3s
+      retries: 30
```
matthewelwell, line 45: Sorry, I really don't understand what we're gaining here? What was wrong with the previous approach?

frontend/e2e/init.cafe.js
matthewelwell: Ok

frontend/docker-compose-e2e-tests.yml
novakzaballa: The previous approach, locally in my environment at least, doesn't wait for the API to be up and running, so it fails since migrations take some time and E2E tests start before the API service is ready.
image
https://docs.docker.com/compose/startup-order/

api/e2etests/e2e_seed_data.py
novakzaballa: Right, done

api/e2etests/e2e_seed_data.py
novakzaballa: Right, done.

api/e2etests/views.py
novakzaballa: Ok, Done

frontend/docker-compose-e2e-tests.yml
matthewelwell: Ok 👍

matthewelwell (approved):

PR #2700, opened 2023-08-29
chore(infra): re-add bad environments cache

matthewelwell:
## Changes

Re-adds the logic to cache bad environment keys in local memory to prevent 404 requests for flags hitting the database. 

Some context: 

This code was originally added in #1138 but subsequently removed in #1234 following an investigation into an unexplainable memory leak. 

Added unit tests.


novakzaballa (approved):
LGTM

PR #2708, opened 2023-08-30
fix(model/featurestate): make environment not null

gagantrivedi:

## Changes
Don't allow null value for environment field of feature state model.

Here is the sqlmigrate of the migration:

```
BEGIN;
--
-- Alter field environment on featurestate
--
SET CONSTRAINTS "features_featurestat_environment_id_c5882902_fk_environme" IMMEDIATE; ALTER TABLE "features_featurestate" DROP CONSTRAINT "features_featurestat_environment_id_c5882902_fk_environme";
ALTER TABLE "features_featurestate" ALTER COLUMN "environment_id" SET NOT NULL;
ALTER TABLE "features_featurestate" ADD CONSTRAINT "features_featurestat_environment_id_c5882902_fk_environme" FOREIGN KEY ("environment_id") REFERENCES "environments_environment" ("id") DEFERRABLE INITIALLY DEFERRED;
COMMIT;



```

Adds migration test


novakzaballa (approved):
LGTM

api/features/migrations/0060_set_featurestate_environment_not_null.py
```diff
@@ -0,0 +1,36 @@
+# Generated by Django 3.2.20 on 2023-08-30 13:14
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+def delete_feature_states_without_environment(apps, schema_editor):
+    FeatureState = apps.get_model("features", "FeatureState")
+    FeatureState.objects.filter(environment=None).delete()
```
matthewelwell: I don't know if it makes a difference, but the correct way to do this I think is: `filter(environment__isnull=True)`. I think it's perhaps because in different database engines, it handles `null` differently.

matthewelwell (changes requested):

api/features/migrations/0060_set_featurestate_environment_not_null.py
gagantrivedi: Yeah, done

matthewelwell (approved):

PR #2712, opened 2023-08-31
fix(featurestate-permissions): Add misc extra checks

gagantrivedi:

## Changes

Add misc checks to feature state view set

Adds unit test cases

### DIFF
--- a/api/environments/tests/test_views.py
+++ b/api/environments/tests/test_views.py
@@ -83,32 +83,6 @@ class EnvironmentTestCase(TestCase):
         assert response.data["results"][0]["identifier"] == identifier_one
         assert response.data["results"][1]["identifier"] == identifier_two
 
-    def test_should_update_value_of_feature_state(self):
-        # Given
-        feature = Feature.objects.create(name="feature", project=self.project)
-        environment = Environment.objects.create(name="test env", project=self.project)
-        feature_state = FeatureState.objects.get(
-            feature=feature, environment=environment
-        )
-        url = reverse(
-            "api-v1:environments:environment-featurestates-detail",
-            args=[environment.api_key, feature_state.id],
-        )
-
-        # When
-        response = self.client.put(
-            url,
-            data=self.fs_put_template % (feature_state.id, True, "This is a value"),
-            content_type="application/json",
-        )
-
-        # Then
-        feature_state.refresh_from_db()
-
-        assert response.status_code == status.HTTP_200_OK
-        assert feature_state.get_feature_state_value() == "This is a value"
-        assert feature_state.enabled
-
     def test_audit_log_entry_created_when_new_environment_created(self):
         # Given
         url = reverse("api-v1:environments:environment-list")
--- a/api/features/serializers.py
+++ b/api/features/serializers.py
@@ -313,6 +313,13 @@ class FeatureStateSerializerBasic(WritableNestedModelSerializer):
         except django.core.exceptions.ValidationError as e:
             raise serializers.ValidationError(e.message)
 
+    def validate_feature(self, feature):
+        if self.instance and self.instance.feature_id != feature.id:
+            raise serializers.ValidationError(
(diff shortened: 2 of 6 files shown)


api/features/views.py
```diff
@@ -394,10 +374,6 @@ def create(self, request, *args, **kwargs):
                 status=status.HTTP_201_CREATED,
                 headers=headers,
             )
-        else:
-            logger.error(serializer.errors)
```
gagantrivedi, line 398: We are changing the response structure for 400 here though, thoughts @matthewelwell

matthewelwell: I think it's fine 👍

api/environments/tests/test_views.py
```diff
@@ -83,32 +83,6 @@ def test_should_return_identities_for_an_environment(self):
...
-            data=self.fs_put_template % (feature_state.id, True, "This is a value"),
-            content_type="application/json",
-        )
-
-        # Then
-        feature_state.refresh_from_db()
-
-        assert response.status_code == status.HTTP_200_OK
-        assert feature_state.get_feature_state_value() == "This is a value"
-        assert feature_state.enabled
```
matthewelwell, line 110: I understand removing the test from here, but I can't see that we've reimplemented this test anywhere?

api/features/serializers.py
```diff
@@ -308,11 +308,33 @@ def save(self, **kwargs):
...
+            raise serializers.ValidationError(
+                "Cannot change the environment of a feature state"
+            )
+        return environment
+
     def validate(self, attrs):
-        environment = attrs.get("environment")
+        environment = (
+            attrs.get("environment")
+            or self.context["view"].get_environment_from_request()
```
matthewelwell: Should we just add the environment to the context instead?

api/environments/tests/test_views.py
gagantrivedi: It's implemented here: https://github.com/Flagsmith/flagsmith/blob/main/api/tests/integration/features/featurestate/test_environment_featurestate_viewset.py

api/features/serializers.py
gagantrivedi: That will trigger an extra query for actions that don't need this, list, for example

matthewelwell: Hmm... but it's a trivial query that would simplify the code. Alternatively, we could just check the action in the `get_serializer_context` ? Maybe this approach is best, but it does feel slightly weird calling methods on the view at this stage of the request.

gagantrivedi: Yeah, that's a good idea. have updated it

api/tests/unit/features/test_unit_features_views.py
```diff
@@ -864,3 +864,158 @@ def test_create_feature_reaching_max_limit(client, project, settings):
...
+    # Then
+    assert response.status_code == status.HTTP_201_CREATED
+    response.json()["feature_state_value"] == new_value
+
+
+@pytest.mark.parametrize(
+    "client",
+    [(lazy_fixture("admin_master_api_key_client")), (lazy_fixture("admin_client"))],
+)
+def test_cannon_create_feature_state_for_feature_from_different_project(
```
```suggestion
def test_cannot_create_feature_state_for_feature_from_different_project(
```

api/features/views.py
```diff
@@ -341,46 +347,26 @@ def get_environment_from_request(self):
...
         DEPRECATED: please use `/features/featurestates/` instead.
         Override create method to add environment and identity (if present) from URL parameters.
         """
         data = request.data
-        environment = self.get_environment_from_request()
-        if (
-            environment.project.organisation
-            not in self.request.user.organisations.all()
-        ):
-            return Response(status.HTTP_403_FORBIDDEN)
```
matthewelwell, line 362: I guess this is handled by the permission checks?

gagantrivedi: Yep, correct

matthewelwell (approved):

PR #2731, opened 2023-09-04
fix: allow registration via invite link if ALLOW_REGISTRATION_WITHOUT_INVITE is False

### LINKED ISSUE #1563: Unable to use OAuth login with invite links if `ALLOW_REGISTRATION_WITHOUT_INVITE=false`
The issue is here: https://github.com/Flagsmith/flagsmith/blob/main/api/custom_auth/serializers.py#L54

If `ALLOW_REGISTRATION_WITHOUT_INVITE` is disabled, we check to see if there is a valid invite, however, this only works for email invites. We need a way to determine if the user landed on the page from an invite link.

matthewelwell:
## Changes

Fixes #1563 

Added unit tests


novakzaballa (approved):
LGTM

PR #2738, opened 2023-09-05
fix(migrations): remove features/0060 set environment not null

gagantrivedi:

## Changes

Revert https://github.com/Flagsmith/flagsmith/commit/55a9ef7bcfc62b9752a7407fbfe545b2df6bb72c

N/A


matthewelwell:
So, I guess there are some considerations here: 

 1. Should we revert the change from the migration in the staging DB? 
 2. Should we delete the record in the django_migrations table in the production db? 

Since the migration hasn't materially changed the behaviour of the code we can probably get away with doing neither of these, but it might be slightly confusing to see. 

I would recommend that we do do both.

gagantrivedi:
> 



> So, I guess there are some considerations here:
> 
>     1. Should we revert the change from the migration in the staging DB?
> 
>     2. Should we delete the record in the django_migrations table in the production db?
> 
> 
> Since the migration hasn't materially changed the behaviour of the code we can probably get away with doing neither of these, but it might be slightly confusing to see.
> 
> I would recommend that we do do both.

Yeah, both would be ideal

matthewelwell (approved):

PR #2747, opened 2023-09-05
feat(task-processor): validate arguments passed to task processor functions

matthewelwell:
## Changes

Validates the inputs to tasks when running synchronously or in a separate thread. Primary use case for this is to catch errors with serialization before going into production. 

Added unit tests. Also, all existing calls to the task processor should subsequently be validated in the unit tests.


novakzaballa (approved):
LGTM

PR #2751, opened 2023-09-06
fix: Feature id in mv-option request is undefined

### LINKED ISSUE #2638: FE posts MV options to `/features/undefined/mv-options/` when creating a new feature
To reproduce: 

 1. In a browser with the network console open, create a new MV feature with at least one variation
 2. Observe that the FE makes a post request to `/features/undefined/mv-options/` when creating the variations

This highlights 2 issues: 

 1. The FE should be posting to `/features/:id/mv-options/` 
 2. The API should be validating the feature id given in the URL path

Item (1) will need to be resolved before item (2), or creating MV features will be broken.

novakzaballa:

## Changes

- The feature ID is no longer undefined.
- It is validated that the feature ID is in the request.

- Create a new MV feature.
- (API) Covered by unit tests.

### DIFF
--- a/api/features/multivariate/views.py
+++ b/api/features/multivariate/views.py
@@ -1,5 +1,5 @@
 from drf_yasg.utils import swagger_auto_schema
-from rest_framework import status, viewsets
+from rest_framework import viewsets
 from rest_framework.decorators import api_view
 from rest_framework.generics import get_object_or_404
 from rest_framework.response import Response
@@ -36,12 +36,7 @@ class MultivariateFeatureOptionViewSet(viewsets.ModelViewSet):
     def create(self, request, *args, **kwargs):
         feature_pk = self.kwargs.get("feature_pk")
         get_object_or_404(Feature, pk=feature_pk)
-
-        serializer = self.get_serializer(data=request.data)
-        if serializer.is_valid():
-            serializer.save()
-            return Response(serializer.data, status=status.HTTP_201_CREATED)
-        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+        return super().create(request, *args, **kwargs)
 
     def get_queryset(self):
         if getattr(self, "swagger_fake_view", False):
--- a/api/features/multivariate/views.py
+++ b/api/features/multivariate/views.py
@@ -1,3 +1,4 @@
+from django.core.exceptions import ValidationError
 from drf_yasg.utils import swagger_auto_schema
 from rest_framework import viewsets
 from rest_framework.decorators import api_view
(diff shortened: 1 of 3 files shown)


api/features/multivariate/views.py
```diff
@@ -37,8 +38,11 @@ def get_queryset(self):
         if getattr(self, "swagger_fake_view", False):
             return MultivariateFeatureOption.objects.none()
 
-        feature = get_object_or_404(Feature, pk=self.kwargs["feature_pk"])
-        return feature.multivariate_options.all()
+        if "feature_pk" in self.kwargs:
+            feature = get_object_or_404(Feature, pk=self.kwargs["feature_pk"])
+            return feature.multivariate_options.all()
+        else:
+            raise ValidationError("Feature Id cannot be undefined or is nonexistent.")
```
matthewelwell: This will have no effect here. The issue is when creating an MV feature option which doesn't use `get_queryset`. I recommend writing a test to cover the issue.

matthewelwell (changes requested):

api/features/multivariate/views.py
```diff
@@ -33,6 +33,21 @@ def get_permissions(self):
...
         ]
 
+    def create(self, request, *args, **kwargs):
+        feature_pk = self.kwargs.get("feature_pk")
+
+        if feature_pk == "undefined":
+            return Response(
+                status=status.HTTP_400_BAD_REQUEST,
+                data={"detail": "Feature Id cannot be 'undefined'."},
+            )
```
matthewelwell: We shouldn't only check for `"undefined"` here. What about other invalid ids? Or ids of features from projects the user doesn't belong to?

matthewelwell (changes requested):

api/features/multivariate/views.py
```diff
@@ -33,6 +33,16 @@ def get_permissions(self):
             )
         ]
 
+    def create(self, request, *args, **kwargs):
+        feature_pk = self.kwargs.get("feature_pk")
+        get_object_or_404(Feature, pk=feature_pk)
```
matthewelwell: This still isn't checking that it's a project they belong to.

api/features/multivariate/views.py
```diff
@@ -33,6 +33,16 @@ def get_permissions(self):
...
+
+        serializer = self.get_serializer(data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
matthewelwell:
```suggestion
        return super().create(request, *args, **kwargs
```

api/features/multivariate/views.py
novakzaballa: Is this being checked here, right?https://github.com/Flagsmith/flagsmith/blob/d2ff80e79ebbc653d5b398134d333c89f03c4042/api/features/multivariate/views.py#L21

api/features/multivariate/views.py
novakzaballa: Corrected.

api/features/multivariate/views.py
matthewelwell: No, looks like that is checking they have permissions on the project. So it should be sufficient to change this to:

```python
feature_pk = self.kwargs.get("feature_pk")
project_pk = self.kwargs.get("project_pk")
get_object_or_404(Feature.objects.filter(project__id=project_pk), pk=feature_pk)
```

novakzaballa: Now I understand what you meant, Thank you, corrected.

matthewelwell (approved):

PR #2761, opened 2023-09-12
test: e2e concurrency + split e2e tests

matthewelwell:

## Changes

Tests are now split into:
- Segment-part-1
- Segment-part-2
- Segment-part-3
- Flag
- Signup
- Invite
- Project

### Improvements

- Tests can now be executed in concurrent (multiple browsers) mode locally and in Github Actions, and in parallel mode (multiple jobs) in Github Actions.
- Since tests are now split, it is easier to find the error in the video and the logs.
- Now you need to re-run only the tests failing not the whole set of E2E tests
- The total time of execution is about half of the prior version.
- We are using Docker BuildX in every job and step and we are caching layers to speedup the process.

### Concurrency
By default, the E2E tests are executed in concurrent mode with **3** browsers.

#### Running all tests in the local dev environment:
In order to run all the E2E tests set the necessary environment variables:
For the API:
```
export E2E_TEST_AUTH_TOKEN='some-token'
export ENABLE_FE_E2E='True'
```
For the Frontend:
```
export E2E_TEST_TOKEN='some-token'
export FLAGSMITH_API_URL=http://127.0.0.1:8000/api/v1/
export ENV=local
npm run env (execute only once with ENV=local)
```
and execute:
```
npm run test:devBundle
```
#### Running only `Invite `and `Environment` E2E tests with two browsers, notice that you can use lowercased test names:
`ENV=local npx cross-env E2E_CONCURRENCY=2 npm run test:devBundle -- invite environment
`

### Running E2E tests in a container (no need to have node installed)
In the frontend folder:
```
docker buildx bake -f docker-compose-e2e-tests.yml --load
docker compose -f docker-compose-e2e-tests.yml run frontend npm run test
```
### Notes

- While the most common sources of flaky tests were solved, React node not being completely rendered, and button not being hit, and some events results not being verified before moving to the next test action, there could still be some flaky tests, however, now should be easier to detect them as well as simply re-run flaky tests.
- Segment-part-2, Segment-part-1, and Flags are long-lasting E2E tests, although the two segments take much more than Flags. Because of that, 3 browsers are configured by default, since Segments-part-2 last almost the same as Segment-part-1, which takes almost the same time as the rest of the tests.

Tested during Github Actions CI/CD execution


novakzaballa (approved):
Approved by Matt. Re-approving.

frontend/Dockerfile
```diff
@@ -7,10 +7,14 @@ USER node
...
 
-COPY --chown=node:node frontend .
-COPY .release-please-manifest.json .
+COPY --chown=node:node frontend/package.json frontend/package-lock.json frontend/.npmrc frontend/.nvmrc ./
+COPY --chown=node:node frontend/bin/ ./bin/
+COPY --chown=node:node frontend/env/ ./env/
 
 RUN npm ci --quiet --production
+
+COPY --chown=node:node frontend .
```
matthewelwell, line 16: I don't really understand why we had to change things here?

Dockerfile
```diff
@@ -1,11 +1,16 @@
...
 WORKDIR /app
-COPY . .
+COPY frontend/package.json frontend/package-lock.json frontend/.npmrc ./frontend/.nvmrc ./frontend/
+COPY frontend/bin/ ./frontend/bin/
+COPY frontend/env/ ./frontend/env/
 
 RUN cd frontend && npm ci --quiet --production
+
+# Copy the entire project - Webpack puts compiled assets into the Django folder
+COPY . .
```
matthewelwell, line 13: What's the reasoning behind splitting things up here?

frontend/e2e/helpers.cafe.ts
```diff
@@ -1,275 +1,375 @@
...
+      '\x1b[0m',
+      '\n',
+    )
+    lastTestGroup[testName] = groupName
+    lastTestName = testName
+  }
+  if (message) {
+    console.log(message)
+  }
+}
```
matthewelwell: Sorry, I just don't understand why this has become so much more complicated?

frontend/e2e/helpers.cafe.ts
```diff
@@ -1,275 +1,375 @@
...
+
+export const deleteTrait = async (index: number) => {
+  await click(byId(`delete-user-trait-${index}`))
+  await click('#confirm-btn-yes')
+  await waitForElementNotExist(byId(`user-trait-${index}`))
+}
+
+const lastTestGroup = {}
+let lastTestName = undefined
```
matthewelwell: This feels dangerous, defining global variables accessed by multiple tests? Especially as we're trying to improve the parallelisation.

frontend/Dockerfile
novakzaballa: To align with DOcker recommendations on how to optimize docker config files to take advantage of caching: https://docs.docker.com/build/cache/#order-your-layers

Dockerfile
novakzaballa: Same as above:
To align with Docker recommendations on how to optimize docker config files to take advantage of caching: https://docs.docker.com/build/cache/#order-your-layers. I enabled Docker layers caching but doesn't help if we copy the whole source code before installing the npm packages since any change to the source code triggers a rebuild of that layer.

frontend/e2e/helpers.cafe.ts
novakzaballa: It is actually a simple way to manage the logs in a concurrent environment. Tests running at the same time (in parallel) can call this function at any time, so I added a couple of things to make the logs more readable (like 3 dots ... when we display parts of a group that have already started before) and simplify the management.

frontend/e2e/helpers.cafe.ts
novakzaballa: That precisely allows managing the logs in an environment where several tests are being executed in parallel.

frontend/e2e/helpers.cafe.ts
```diff
@@ -1,275 +1,375 @@
...
+  await assertTextContent(byId(`user-trait-value-${index}`), expectedValue)
+}
+
+export const deleteTrait = async (index: number) => {
+  await click(byId(`delete-user-trait-${index}`))
+  await click('#confirm-btn-yes')
+  await waitForElementNotExist(byId(`user-trait-${index}`))
+}
+
+const lastTestGroup = {}
```
matthewelwell:
```suggestion
const lastTestSection = {}
```

novakzaballa: Sorry, I searched for the word group but I probably made a mistake. Correcting.

matthewelwell (commented):
LGTM (approving via comment as I created the PR)

PR #2763, opened 2023-09-12
feat: add webhooks to v2 feature versioning

matthewelwell:
## Changes

Adds webhook logic to v2 versioning. This is achieved by: 

 1. Ignoring current webhooks if the feature state is associated with an environment feature version object
 2. Creating a brand new webhook schema to be sent when a new version is published

Note: this PR is created on top of the current feature versioning PR #2382 

Added unit tests


novakzaballa (approved):
LGTM

PR #2772, opened 2023-09-14
ci: Run rbac tests with unit on merge to main

gagantrivedi:

## Changes

Create a workflow to run tests from private packages like rbac on merge to main 

here: https://github.com/Flagsmith/flagsmith/actions/runs/6794365166

### DIFF
--- a/.github/workflows/api-tests-with-private-packages.yml
+++ b/.github/workflows/api-tests-with-private-packages.yml
@@ -1,4 +1,4 @@
-name: Run unit tests for Rbac
+name: Run API Unit tests with private package
 
 on:
   push:
--- a/.github/workflows/api-tests-with-private-packages.yml
+++ b/.github/workflows/api-tests-with-private-packages.yml
@@ -67,12 +67,6 @@ jobs:
         shell: bash
         working-directory: api
 
-      - name: Check for missing migrations
-        env:
-          DOTENV_OVERRIDE_FILE: .env-ci
-          opts: --no-input --dry-run --check
-        run: make django-make-migrations
-
       - name: Run Tests
         env:
           DOTENV_OVERRIDE_FILE: .env-ci
--- a/.github/workflows/api-tests-with-private-packages.yml
+++ b/.github/workflows/api-tests-with-private-packages.yml
@@ -1,4 +1,4 @@
-name: Run API Unit tests with private package
+name: Run API tests with private package
 
 on:
   merge_group:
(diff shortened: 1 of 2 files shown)


gagantrivedi:
Depends on https://github.com/Flagsmith/flagsmith/pull/2685

gagantrivedi:
Docs fix is here: https://github.com/Flagsmith/flagsmith/pull/2938

khvn26:
I think this should modify the original `API Unit Tests` job, not add a new one.

gagantrivedi:
> I think this should modify the original `API Unit Tests` job, not add a new one.

No, we also want to test core without any of these private packages(and don't want it to run on pull request)

.github/workflows/api-tests-with-private-packages.yml
```diff
@@ -0,0 +1,84 @@
...
+      - name: integrate rbac tests
+        run: ln -s $GITHUB_WORKSPACE/flagsmith-rbac/tests tests/rbac_tests
+        shell: bash
+        working-directory: api
+
+      - name: Check for missing migrations
+        env:
+          DOTENV_OVERRIDE_FILE: .env-ci
+          opts: --no-input --dry-run --check
+        run: make django-make-migrations
```
matthewelwell: This should already have been done in the other repo, right?

.github/workflows/api-tests-with-private-packages.yml
```diff
@@ -0,0 +1,84 @@
...
+        python-version: ['3.10', '3.11']
+
+    steps:
+      - name: Cloning repo
+        uses: actions/checkout@v3
+
+      - name: Create analytics database
+        env:
+          PGPASSWORD: postgres
+        run: createdb -h localhost -U postgres -p 5432 analytics
```
matthewelwell: Do we need the analytics DB for these tests?

matthewelwell:
>>I think this should modify the original API Unit Tests job, not add a new one.
>
>No, we also want to test core without any of these private packages(and don't want it to run on pull request)

This is an interesting discussion, yeah. I guess theoretically, the unit tests should be isolated so we should be able to test them in a single workflow, right? The point regarding only running these on a merge to main (or ideally in the merge queue) is a good one though. 

@gagantrivedi can you update this to use the merge queue functionality?

gagantrivedi:
> This is an interesting discussion, yeah. I guess theoretically, the unit tests should be isolated so we should be able to test them in a single workflow, right?

I guess the problem here is that we call all python tests unit tests(in the workflow atleast) maybe we should rename this workflow?

matthewelwell (approved):

PR #2791, opened 2023-09-19
feat: Add or remove user and groups from roles

### LINKED ISSUE #2754: FE changes to add/remove User/Groups to/from Roles
Currently, the API supports adding users and groups to roles however in the frontend we don't have a way to do it.

In the management option of the top bar, when the admin goes to the members tab, we should display a roles sub-tab,
When an admin edits a role, a modal is shown, and in the first tab, where the name and description of the role are shown, we need to allow the admin to add and remove users and/or groups to the role.

**Additional context**
Initially, the ability to add and remove users/groups will reside only on the roles sub-tab of the members tab, in the screen shown when an admin clicks on the `Management` top menu option

novakzaballa:

## Changes

- Add users and roles in the roles modal.
- Add roles in the edit permission modal for users and groups.
- Edit permissions for projects and environments in their Settings tab.
- Added "show_role_management" flag to wrap this features.

### Manage roles
- Go to manage -> members.
- Click the "Roles" tab.
- Click the "Create Role" button.
- Create a new role.
- Select the role in the role list.
- Edit name and description. (If you want to exit without saving, a confirmation modal will appear.)
- Edit organisation, project, and environment permissions (If you want to exit or change the tab without saving, a confirmation modal will appear.)
- Click the trash button to delete the role.
### Add a role to a user.
- Go to Manage -> Members
- Click the "Edit permissions" button in a user.
- Click the "Add role" button.
- Select a role to add it.
- Click the role to remove it.
### Add a role to a Group.
- Click the "Groups" tab.
- Click the "Edit permissions" button in a group.
- Click the "Add role" button.
- Select a role to add it.
- Click the role to remove it.
### Add a User or Group to a Role.
- Click the "Roles" tab.
- Click a role.
- Click the "Members" tab.
- Click the "Assign role to User" or "Assign role to Group" button.
- Select a user or group to add it.
- Click the user or group to remove it.
### Edit role permissions for project settings and environment settings.
- Go to Project Settings.
- Click the "Members" tab.
- Click the "Roles" tab.
- Click a role.
- Edit Role.
- Click the "Save permission" button.
### Edit role permissions for environment settings.
- Go to Environment settings.
- Click the "Members" tab.
- Click the "Roles" tab.
- Click a role.
- Edit Role.
- Click the "Save permission" button.


frontend/common/dispatcher/action-constants.js
```diff
@@ -37,6 +37,7 @@ const Actions = Object.assign({}, require('./base/_action-constants'), {
   'GET_IDENTITY_SEGMENTS': 'GET_IDENTITY_SEGMENTS',
   'GET_ORGANISATION': 'GET_ORGANISATION',
   'GET_PROJECT': 'GET_PROJECT',
+  'GET_ROLES': 'GET_ROLES',
```
kyle-ssg: Can we use rtk here please?

frontend/common/stores/role-store.js
```diff
@@ -0,0 +1,46 @@
+const Dispatcher = require('../dispatcher/dispatcher')
+const BaseStore = require('./base/_store')
+const { getRoles } = require('../services/useRole')
+const { getStore } = require('../store')
+
+const controller = {
+  getRoles: (organisationId) => {
+    store.loading()
+    return getRoles(
```
kyle-ssg: I'm confused as to why we need any of this

frontend/common/dispatcher/action-constants.js
novakzaballa: Corrected.

frontend/common/stores/role-store.js
novakzaballa: Sorry, I was using the old way, I already replaced it.

frontend/web/components/GroupSelect.tsx
```diff
@@ -31,19 +33,23 @@ const GroupSelect: FC<GroupSelectType> = ({
       if (!search) return true
       return `${v.name}`.toLowerCase().includes(search)
     })
+  const modalClassName = isSmall
+    ? 'sm-inline-modal--tags'
```
kyle-ssg: Can we make this inline-modal-${size}

frontend/web/components/GroupSelect.tsx
```diff
@@ -18,6 +19,7 @@ const GroupSelect: FC<GroupSelectType> = ({
   disabled,
   groups,
   isOpen,
+  isSmall,
```
kyle-ssg: Can we make this size?: "sm"

frontend/web/components/CollapsibleNestedList.tsx
```diff
@@ -0,0 +1,183 @@
...
+          minHeight: '60px',
+        }}
+      >
+        {mainItems.map((mainItem, index) => (
+          <div key={index}>
+            <Row
+              key={index}
+              onClick={() => toggleExpand(mainItem.id)}
+              style={{
+                backgroundColor: '#fafafb',
```
kyle-ssg: Can we get rid of these static styles

frontend/web/components/CollapsibleNestedList.tsx
```diff
@@ -0,0 +1,183 @@
...
+        </div>
+      ) : (
+        <div>{isAdmin ? 'Administrator' : permissionsSummary}</div>
+      )
+    }
+
+    return (
+      <div
+        style={{
+          borderRadius: '10px',
```
kyle-ssg: Static styles, not using design system border radius either.

frontend/web/components/CollapsibleNestedList.tsx
```diff
@@ -0,0 +1,183 @@
...
+  id: string
+}
+
+type CollapsibleNestedListProps = {
+  mainItems: MainItem[]
+  role: Role
+  level: string
+}
+
+const CollapsibleNestedList: React.FC<CollapsibleNestedListProps> = forwardRef(
```
kyle-ssg: Not really sure if this is the right name for this, if it's coupled with projects and permissions. It's not obvious what this component does

frontend/web/components/CollapsibleNestedList.tsx
```diff
@@ -0,0 +1,183 @@
...
+          },
+          tabChanged() {
+            return unsavedProjects.length > 0
+          },
+        }
+      },
+      [unsavedProjects],
+    )
+
+    const PermissionsSummary = ({ levelId }) => {
```
kyle-ssg: Does this need to be defined within the render function of the parent component?

frontend/web/components/RolesSelect.tsx
```diff
@@ -0,0 +1,83 @@
...
+  value: number[] | undefined
+  isOpen: boolean
+  onAdd: (id: number, isUser: boolean) => void
+  onRemove: (id: number, isUser: boolean) => void
+  onToggle: () => void
+}
+const RoleSelect: FC<RoleSelectType> = ({
+  disabled,
+  isOpen,
+  onAdd,
```
kyle-ssg, line 20: Can we make this just onChange:(newValue)=>void, adding onAdd/onRemove doesn't seem very intuitive/necessary, it can be done within this component.

frontend/web/components/modals/ConfirmDeleteRole.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const [deleteRole, { isError, isSuccess: deleted }] = useDeleteRoleMutation()
+
+  useEffect(() => {
+    if (deleted) {
+      onComplete?.()
+      closeModal()
+    }
+  }, [deleted, onComplete])
+
+  const ModalBody: FC<ConfirmDeleteRoleType> = ({ role }) => {
```
kyle-ssg: Does this need to be defined in the render function? Also, why is it even a separate component ?

frontend/web/components/modals/ConfirmDeleteRole.tsx
```diff
@@ -0,0 +1,63 @@
...
+          </Button>
+        </div>
+      </form>
+    </div>
+  )
+}
+
+export default ConfirmDeleteRole
+
+module.exports = ConfirmDeleteRole
```
kyle-ssg: Does this need both module.exports and export default ?

frontend/web/components/GroupSelect.tsx
novakzaballa: Done.

frontend/web/components/GroupSelect.tsx
novakzaballa: Done.

frontend/web/components/CollapsibleNestedList.tsx
novakzaballa: Done.

frontend/web/components/CollapsibleNestedList.tsx
novakzaballa: Renamed.

frontend/web/components/CollapsibleNestedList.tsx
novakzaballa: No, Moved.

frontend/web/components/modals/ConfirmDeleteRole.tsx
novakzaballa: No, `module.exports` Deleted.

frontend/web/components/modals/ConfirmDeleteRole.tsx
novakzaballa: It is not necessary to have it there. I placed it in a separate component to keep it independent if I needed to make more significant changes, but that won't be the case. I already moved it into the return.

frontend/web/components/CollapsibleNestedList.tsx
novakzaballa: Done.

gagantrivedi:
Screenshot_20230928_163245

Should this be something like edit roles project permission? Because that's what you are editing here?

gagantrivedi:
Screenshot_20230928_163502

This should be `environment` instead of projects

gagantrivedi:
Should the modal close after it saves permission?

novakzaballa:
@gagantrivedi 
> Should this be something like edit roles project permission? Because that's what you are editing here?

Updated.

> This should be environment instead of projects

Thank you for catching this error, corrected.

> Should the modal close after it saves permission?


Yes, you're right, I think that's normal behavior, corrected.

frontend/web/components/pages/EnvironmentSettingsPage.js
```diff
@@ -690,8 +721,11 @@ const EnvironmentSettingsPage = class extends Component {
                           parentLevel='project'
                           parentSettingsLink={`/project/${this.props.match.params.projectId}/settings`}
                           id={this.props.match.params.environmentId}
+                          envId={env.id}
                           router={this.context.router}
                           level='environment'
+                          roleTabTitle='Environment Roles'
```
gagantrivedi: I think this is a bit misleading? It should rather be Role Environment permission? Because that's what it is, right? It is not a different role, but a permission attached to that role. Does that make sense?

frontend/web/components/pages/ProjectSettingsPage.js
```diff
@@ -461,6 +487,9 @@ const ProjectSettingsPage = class extends Component {
                         tabClassName='flat-panel'
                         id={this.props.match.params.projectId}
                         level='project'
+                        roleTabTitle='Project Roles'
```
gagantrivedi: Same as environment Roles

frontend/web/components/pages/EnvironmentSettingsPage.js
gagantrivedi: Also, do we need the text `Can edit roles project permission.` that written under that?

gagantrivedi: Should the user be allowed to add environment permissions to role here:?

gagantrivedi:
> Yes, you're right, I think that's normal behavior, corrected.

Think it only works on the organisation tab and not on the other tabs

gagantrivedi:
Edit project permission for the role on the creates this pop on save
Screenshot_20230929_131932

gagantrivedi:
> Edit project permission for the role on the creates this pop on save Screenshot_20230929_131932

Same behaviour on the environment panel as well

gagantrivedi (changes requested):

novakzaballa:
> Screenshot_20230928_163245
> 
> Should this be something like edit roles project permission? Because that's what you are editing here?

@gagantrivedi good catch I will remove that whole section since it is redundant. I probably forgot to remove it.

frontend/web/components/pages/ProjectSettingsPage.js
novakzaballa: Corrected.

frontend/web/components/pages/EnvironmentSettingsPage.js
novakzaballa: > I think this is a bit misleading? It should rather be Role Environment permission? Because that's what it is, right? It is not a different role, but a permission attached to that role. Does that make sense?

Corrected.

>Also, do we need the text Can edit roles project permission. that written under that?

removed.

novakzaballa:
> Edit project permission for the role on the creates this pop on save 
> Same behaviour on the environment panel as well

@gagantrivedi good catch, It was a bug introduced in the last commit, corrected.

novakzaballa:
In addition I'm making these changes suggested by @dabeeeenster in Projects and Environment Settings:
Screen Shot 2023-09-29 at 11 48 09

frontend/web/components/pages/ChangeRequestPage.js
```diff
@@ -27,7 +27,7 @@ const ChangeRequestsPage = class extends Component {
   static contextTypes = {
     router: propTypes.object.isRequired,
   }
-
+  // TODO: review this function users?.filter((v) => approvals?.includes(v.group))
```
kyle-ssg: Does this still need reviewing?  Am I right in saying this means if one person from a group approves then it treats everyone in the same group?

novakzaballa: This is not related to this PR, I think it is related to this issue https://github.com/Flagsmith/flagsmith/issues/2823,
Comment deleted.

frontend/common/types/responses.ts
```diff
@@ -295,6 +295,18 @@ export type Account = {
   is_superuser: boolean
 }
 
+export type Role = {
+  id?: number
```
kyle-ssg: Are all of these optional ?

kyle-ssg: Are all of these optional ?

frontend/web/components/EditPermissions.tsx
```diff
@@ -101,220 +162,690 @@ const _EditPermissionsModal: FC<EditPermissionModalType> = (props) => {
...
+      isEnvPermCreating,
+      isEnvPermUpdated,
+      isEnvPermCreated,
+      envUpdatedError,
+      envCreatedError,
+    ])
+
+    useEffect(() => {
+      if (isOrgPermUpdating || isOrgPermCreating) {
+        setSaving(true)
```
kyle-ssg: Can't this just be a variable? 

```
const isSaving = isOrgPermUpdating || isOrgPermCreating || isEnvPermUpdating || isEnvPermCreating
```

frontend/web/components/EditPermissions.tsx
```diff
@@ -56,38 +87,68 @@ type EntityPermissions = Omit<
...
+      push,
+      role,
+      roles,
+      user,
+    } = props
+    useImperativeHandle(
+      ref,
+      () => {
+        return {
+          onClosing() {
```
kyle-ssg, line 116: What is the purpose of these imperative refs?

frontend/web/components/EditPermissions.tsx
```diff
@@ -101,220 +162,690 @@ const _EditPermissionsModal: FC<EditPermissionModalType> = (props) => {
...
+        toast('Environment permissions Saved')
+        permissionChanged?.()
+        onSave?.()
+        setSaving(false)
+        if (editPermissionsFromSettings) {
+          close()
+        }
+      }
+      if (envUpdatedError || envCreatedError) {
+        toast('Failed to Save', 'danger')
```
kyle-ssg: Can't you just do all of this in .then of the actions checking for res.error? All of this useEffect looks unnecessary

frontend/web/components/EditPermissions.tsx
```diff
@@ -101,220 +162,690 @@ const _EditPermissionsModal: FC<EditPermissionModalType> = (props) => {
...
+      orgUpdatedError,
+      orgCreatedError,
+    ])
+
+    useEffect(() => {
+      if (isProjectPermUpdating || isProjectPermCreating) {
+        setSaving(true)
+      }
+      if (isProjectPermUpdated || isProjectPermCreated) {
+        refetchProjectPerm()
```
kyle-ssg: This shouldn't be necessary, you should define tags that are invalidated in the hook.

frontend/web/components/EditPermissions.tsx
```diff
@@ -101,220 +162,690 @@ const _EditPermissionsModal: FC<EditPermissionModalType> = (props) => {
...
+
+    const [
+      createRolePermissionGroup,
+      { data: groupsData, isSuccess: groupAdded },
+    ] = useCreateRolePermissionGroupMutation()
+
+    const [deleteRolePermissionGroup] = useDeleteRolePermissionGroupMutation()
+
+    const [
+      updateRoleEnvironmentPermissions,
```
kyle-ssg: This seems quite bloated, did you consider just having 1 delete/update role permissions with level:"environment"|"organisation"|"project"

frontend/web/components/EditPermissions.tsx
```diff
@@ -1,4 +1,10 @@
-import React, { FC, useEffect, useState } from 'react'
```
kyle-ssg, line 1: Overall I think this component needs a lot of tidying up, there's too much happening.

Unless there's a good reason I think we should
 - Cut down the hooks to just 2 with a project|org|env param
 - The useEffects / saving and error states don't need to be the result of useEffects
 - Manual refetching should never be necessary, providesTags/invalidateTags should be defined at the service level

frontend/web/components/EditPermissions.tsx
```diff
@@ -101,220 +162,690 @@ const _EditPermissionsModal: FC<EditPermissionModalType> = (props) => {
...
+            close()
+          })
+          .catch(() => {
+            setSaving(false)
+          })
+      } else {
+        if (entityId) {
+          switch (level) {
+            case 'organisation':
+              updateRoleOrganisationPermissions({
```
kyle-ssg: This adds to the above point, this could just be 1 action with a level param.

frontend/common/types/responses.ts
novakzaballa: According to the API yes
Screen Shot 2023-11-01 at 10 34 10
but I see that they should not be optional here.
Corrected.

frontend/web/components/pages/ChangeRequestPage.js
```diff
@@ -225,6 +224,12 @@ const ChangeRequestsPage = class extends Component {
         orgUsers,
         changeRequest.approvals.map((v) => v.user),
       )
+    console.log(
```
kyle-ssg: Can remove

novakzaballa: Removed.

frontend/web/components/EditPermissions.tsx
novakzaballa: Updated to use a level param.

frontend/web/components/EditPermissions.tsx
novakzaballa: This endpoint has no levels.

frontend/web/components/EditPermissions.tsx
novakzaballa: Done.

frontend/web/components/EditPermissions.tsx
novakzaballa: Done.

frontend/web/components/EditPermissions.tsx
novakzaballa: The intention was to know if there was something not saved in the initial tab.

frontend/web/components/EditPermissions.tsx
novakzaballa: Overall I think this component needs a lot of tidying up, there's too much happening.

Unless there's a good reason I think we should

>Cut down the hooks to just 2 with a project|org|env param

Done.

>The useEffects / saving and error states don't need to be the result of useEffects

Done.


>Manual refetching should never be necessary, providesTags/invalidateTags should be defined at the service level

Manual refetchings were deleted.

frontend/web/components/EditPermissions.tsx
novakzaballa: Deleted.

kyle-ssg (approved):

kyle-ssg:
Loads better

gagantrivedi (approved):
vercel is not working for me - approving to unblock @novakzaballa

PR #2802, opened 2023-09-22
fix: Last Influx data updated at never updates

### LINKED ISSUE #2760: In sales dashboard, Last Influx data updated at never updates
Hitting 'Update only influx data' doesnt store/display the last updated date. 

image

It should update with the date the action was run.

novakzaballa:

## Changes

- Added a filter for when the "influx update at" field is null.

### DIFF
--- a/api/sales_dashboard/views.py
+++ b/api/sales_dashboard/views.py
@@ -93,12 +93,14 @@ class OrganisationList(ListView):
         # when the caches were last updated.
         try:
             subscription_information_caches_influx_updated_at = (
-                OrganisationSubscriptionInformationCache.objects.order_by(
-                    "-influx_updated_at"
+                OrganisationSubscriptionInformationCache.objects.filter(
+                    influx_updated_at__isnull=False
                 )
+                .order_by("-influx_updated_at")
                 .first()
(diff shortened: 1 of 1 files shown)


api/sales_dashboard/views.py
```diff
@@ -93,12 +93,14 @@ def get_context_data(self, **kwargs):
         # when the caches were last updated.
         try:
             subscription_information_caches_influx_updated_at = (
-                OrganisationSubscriptionInformationCache.objects.order_by(
-                    "-influx_updated_at"
+                OrganisationSubscriptionInformationCache.objects.filter(
+                    influx_updated_at__isnull=False
                 )
+                .order_by("-influx_updated_at")
```
gagantrivedi: If we are only fetching the first element, we can do `desc(null_last=True)` /
docs: https://docs.djangoproject.com/en/3.1/ref/models/expressions/#django.db.models.Expression.desc

gagantrivedi (changes requested):

api/sales_dashboard/views.py
novakzaballa: Done.

gagantrivedi (approved):

PR #2805, opened 2023-09-25
feat: trial management in sales dashboard

dabeeeenster:
This is so the sales team can manage folk going into and coming out of trials. 

We need to clear up how we manage this data as we currently have two areas of the database that store this data, and we need to figure out which to honour based on the subscription state.

### DIFF
--- a/api/sales_dashboard/forms.py
+++ b/api/sales_dashboard/forms.py
@@ -7,7 +7,12 @@ from django.core.mail import send_mail
 
 from environments.models import Environment
 from features.models import Feature
-from organisations.models import Organisation
+from organisations.models import TRIAL_SUBSCRIPTION_ID, Organisation
+from organisations.subscriptions.constants import (
+    FREE_PLAN_ID,
+    MAX_API_CALLS_IN_FREE_PLAN,
+    MAX_SEATS_IN_FREE_PLAN,
+)
 from projects.models import Project
 from users.models import FFAdminUser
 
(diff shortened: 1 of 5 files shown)


novakzaballa (approved):
LGTM

api/sales_dashboard/forms.py
```diff
@@ -28,6 +33,29 @@ def save(self, organisation, commit=True):
         organisation.subscription.save()
 
 
+class StartTrialForm(forms.Form):
+    max_seats = forms.IntegerField()
+    max_api_calls = forms.IntegerField()
+
+    def save(self, organisation, commit=True):
```
matthewelwell: Typehints please!

```suggestion
    def save(self, organisation: Organisation, commit: bool = True):
```

api/sales_dashboard/forms.py
```diff
@@ -28,6 +33,29 @@ def save(self, organisation, commit=True):
...
+        organisation.subscription.max_seats = self.cleaned_data["max_seats"]
+        organisation.subscription.max_api_calls = self.cleaned_data["max_api_calls"]
+        organisation.subscription.subscription_id = TRIAL_SUBSCRIPTION_ID
+        organisation.subscription.customer_id = TRIAL_SUBSCRIPTION_ID
+        organisation.subscription.plan = "enterprise-saas-monthly-v2"
+        organisation.subscription.save()
+
+
+class EndTrialForm(forms.Form):
+    def save(self, organisation, commit=True):
```
matthewelwell: Typehints please!

```suggestion
    def save(self, organisation: Organisation, commit: bool = True):
```

api/sales_dashboard/forms.py
```diff
@@ -28,6 +33,29 @@ def save(self, organisation, commit=True):
...
+class StartTrialForm(forms.Form):
+    max_seats = forms.IntegerField()
+    max_api_calls = forms.IntegerField()
+
+    def save(self, organisation, commit=True):
+        organisation.subscription.max_seats = self.cleaned_data["max_seats"]
+        organisation.subscription.max_api_calls = self.cleaned_data["max_api_calls"]
+        organisation.subscription.subscription_id = TRIAL_SUBSCRIPTION_ID
+        organisation.subscription.customer_id = TRIAL_SUBSCRIPTION_ID
+        organisation.subscription.plan = "enterprise-saas-monthly-v2"
```
matthewelwell: Feels like we should accept a couple of different options here, right? Sometimes we want to give people a trial of scale up I think?

api/sales_dashboard/forms.py
matthewelwell: Also, this (and the similar save method on the EndTrialForm) should probably just call a new method on the `Subscription` class like `start|end_trial(max_seats: int, max_api_calls: int)`. That way, we can easily unit test the method on the subscription class.

matthewelwell (changes requested):

api/sales_dashboard/forms.py
dabeeeenster: I think it makes sense to offer them everything - it may lead to an upsell to Enteprise.

matthewelwell (approved):

PR #2807, opened 2023-09-27
fix(tasks): Create a different task to update environment document 

gagantrivedi:

## Changes
Create a different task for updating environment document and sending SSE message in order to reduce the transaction length (earlier it was doing all of the above as well in the same transaction)

Adds unit tests

### DIFF
--- a/api/audit/models.py
+++ b/api/audit/models.py
@@ -110,6 +110,8 @@ class AuditLog(LifecycleModel):
         is_now=True,
     )
     def update_environments_updated_at(self):
+        from environments.tasks import process_environment_update
+
         environments_filter = Q()
         if self.environment_id:
             environments_filter = Q(id=self.environment_id)
@@ -119,6 +121,5 @@ class AuditLog(LifecycleModel):
         self.project.environments.filter(environments_filter).update(
             updated_at=self.created_date
         )
-        from .tasks import process_environment_update
 
-        process_environment_update.delay(args=(self.id))
+        process_environment_update.delay(args=(self.id,))
--- a/api/audit/models.py
+++ b/api/audit/models.py
@@ -14,10 +14,6 @@ from django_lifecycle import (
 from api_keys.models import MasterAPIKey
 from audit.related_object_type import RelatedObjectType
 from projects.models import Project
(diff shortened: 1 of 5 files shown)


api/audit/models.py
```diff
@@ -113,12 +109,9 @@ def add_project(self):
         when="environment_document_updated",
         is_now=True,
     )
-    def process_environment_update(self):
-        self.update_environments_updated_at()
-        self.send_environments_to_dynamodb()
-        self.send_environment_update_message()
-
     def update_environments_updated_at(self):
```
matthewelwell: This method feels like it could be named better since it's triggering another task to handle the environment update logic. It might also benefit from a docstring I guess?

matthewelwell (commented):
One fairly minor comment, otherwise this looks good to me.

api/audit/models.py
```diff
@@ -129,19 +122,4 @@ def update_environments_updated_at(self):
...
-    def send_environment_update_message(self):
-        if self.environment_id:
-            environment = self.environment
-            # Because we updated the environment `updated_at` in the previous hook in bulk
-            # update it manually here to save a `refresh_from_db` call
-            environment.updated_at = self.created_date
-            send_environment_update_message_for_environment(environment)
-        else:
-            send_environment_update_message_for_project(self.project)
+        process_environment_update.delay(args=(self.id,))
```
matthewelwell, line 125: Are we certain that `self.id` is populated at this point in the flow? I think it probably is but I'm not _certain_ that any of the tests fully exercise this?

gagantrivedi: No, we have this: https://github.com/Flagsmith/flagsmith/pull/2807/files#diff-bf708a057d496be3362cd0687b441b41276bd6d62904e3346bbb1de29b09f497R175

api/audit/models.py
gagantrivedi: But it does update the `updated_at`

gagantrivedi: I guess I can change it back to `process_environment_update`

matthewelwell (approved):

PR #2816, opened 2023-09-30
docs: Update NextJS docs

### LINKED ISSUE #186: [Bug] - NextJS >= 13.0.0 SSR `useFlagsmith` state not matching the state that is passed to the provider.
I am using the NextJS SSR example for implementation and have ran into issues when working locally with the hook states not matching what is fetched in CustomApp.getInitialProps and then passed down to the FlagsmithProvider.

My expectation here is that all SSR components using the hook will leverage the state I have fetched in the initial CustomApp fetch (which is correct), what I am seeing however is that it seems to be just responding with an old version of the flag state.

Below (responses reduced so you can see the specific issue) is what comes back for my flags when I console out the `useFlagsmith` response in a component on the server

```js
{
  flagsmith: {
   ...
    flags: { editorial_plp: { id: 56935, enabled: true, value: 'on' } },
   ...
  }
}
```

and the same console log on the client console is:

```js
{
  flagsmith: {
    flags: {
      editorial_plp: { id: 56935, enabled: true, value: 'on' },
      shop_pay_button: { id: 58185, enabled: true, value: null }
    },
    ...
  }
}
```

and my flagsmithSSR log (server side)

```js
{
  flagsmithSSR: {
    flags: {
      editorial_plp: { id: 56935, enabled: true, value: 'on' },
      shop_pay_button: { id: 58185, enabled: true, value: null }
    },
    ...
  }
}
```

As you can see the first is not a reflection of all my feature flags

kyle-ssg:

## Changes
Updates docs based on https://github.com/Flagsmith/flagsmith-js-client/issues/186

### DIFF
--- a/docs/docs/clients/client-side/nextjs-and-ssr.md
+++ b/docs/docs/clients/client-side/nextjs-and-ssr.md
@@ -48,18 +48,17 @@ state.
 
 ```javascript
 import { FlagsmithProvider } from 'flagsmith/react';
-import flagsmith, { createFlagsmithInstance } from 'flagsmith/isomorphic';
+import { createFlagsmithInstance } from 'flagsmith/isomorphic';
 function MyApp({ Component, pageProps, flagsmithState }) {
+ const flagsmithRef = useRef(createFlagsmithInstance())
  return (
-  <FlagsmithProvider flagsmith={flagsmith} serverState={flagsmithState}>
+  <FlagsmithProvider flagsmith={flagsmithRef.current} serverState={flagsmithState}>
    <Component {...pageProps} />
   </FlagsmithProvider>
  );
 }
 
 MyApp.getInitialProps = async () => {
- // This could be getStaticProps too depending on your build flow.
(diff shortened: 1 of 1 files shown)


docs/docs/clients/client-side/nextjs-and-ssr.md
```diff
@@ -48,18 +48,17 @@ state.
 
 ```javascript
 import { FlagsmithProvider } from 'flagsmith/react';
-import flagsmith, { createFlagsmithInstance } from 'flagsmith/isomorphic';
+import { createFlagsmithInstance } from 'flagsmith/isomorphic';
 function MyApp({ Component, pageProps, flagsmithState }) {
+ const flagsmithRef = useRef(createFlagsmithInstance());
```
joelpierre, line 53: @kyle-ssg - Would I need to add the UUID of the user in here? or is it fine to use as is?

novakzaballa (approved):
LGTM

PR #2826, opened 2023-10-04
feat(tasks/queue-size): Implement queue_size

gagantrivedi:

## Changes

Implement queue size to limit low priority tasks from overwhelming the queue 

Plan and SQL:

```
explain analyze SELECT COUNT(*) AS "__count" FROM "task_processor_task" WHERE (NOT "task_processor_task"."completed" AND "task_processor_task"."num_failures" < 3 AND "task_processor_task"."task_identifier" = 'edge_request_forwarder.forward_identity_request');
                                                                       QUERY PLAN                                                                        
---------------------------------------------------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=69541.34..69541.35 rows=1 width=8) (actual time=5.017..5.018 rows=1 loops=1)
   ->  Index Scan using incomplete_tasks_idx on task_processor_task  (cost=0.38..68634.33 rows=362804 width=0) (actual time=0.018..5.013 rows=4 loops=1)
         Filter: ((task_identifier)::text = 'edge_request_forwarder.forward_identity_request'::text)
         Rows Removed by Filter: 1
 Planning Time: 0.305 ms
 Execution Time: 5.037 ms
(6 rows)

```

Adds unit test cases

### DIFF
--- a/api/task_processor/models.py
+++ b/api/task_processor/models.py
@@ -105,7 +105,7 @@ class Task(AbstractBaseTask):
         cls,
         schedule_for: datetime,
         task_identifier: str,
-        queue_size: int,
+        queue_size: typing.Optional[int],
         *,
         args: typing.Tuple[typing.Any] = None,
         kwargs: typing.Dict[str, typing.Any] = None,
--- a/api/task_processor/models.py
+++ b/api/task_processor/models.py
@@ -6,7 +6,7 @@ import simplejson as json
 from django.db import models
 from django.utils import timezone
 
-from task_processor.exceptions import TaskProcessingError
+from task_processor.exceptions import TaskProcessingError, TaskQueueFullError
 from task_processor.managers import RecurringTaskManager, TaskManager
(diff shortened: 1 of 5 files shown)


novakzaballa:
@gagantrivedi , can you please add to the description what will happen when the queue size limit is reached, as you explained in the Eng Meeting, just to have it documented here?

api/task_processor/models.py
```diff
@@ -105,10 +105,22 @@ def schedule_task(
         cls,
         schedule_for: datetime,
         task_identifier: str,
+        queue_size: int,
```
matthewelwell: Think this should be `queue_size: Optional[int]` ?

matthewelwell (commented):
Added one minor comment about the implementation but I do wonder if we might be better off implementing multiple queues (with prioritisation). I don't know if it would be much more work to do that or not?

matthewelwell (approved):
This should be a temporary measure until we can deprecate these tasks entirely.

PR #2837, opened 2023-10-06
feat: LaunchDarkly importer UI

### LINKED ISSUE #2348: LaunchDarkly importer

novakzaballa:

## Changes

- Added a tab within the project settings that allows you to import a project from Launchdarkly
- This feature is wrapped by the new "import_project" flag

- Go to the Project settings section
- Click the 'Import' tab
- Set the LD Key and click the 'next' button
- Select an LD project
- A modal will open, click on the confirm button


frontend/common/data/base/_data.js
```diff
@@ -19,7 +19,7 @@ module.exports = {
     if (method !== 'get')
       options.headers['Content-Type'] = 'application/json; charset=utf-8'
 
-    if (this.token) {
+    if (this.token && !isExternal) {
```
kyle-ssg: I think this is part of a different PR

frontend/web/components/pages/ImportPage.tsx
```diff
@@ -0,0 +1,181 @@
...
+      setAppIsLoading(true)
+      const intervalId = setInterval(async () => {
+        await refetch()
+
+        if (statusLoaded && status && status.status.result === 'success') {
+          clearInterval(intervalId)
+          setAppIsLoading(false)
+          window.location.reload()
+        }
+      }, 5000)
```
khvn26: Two questions/concerns here:

1. There's no need to block the UI during import as it's performed asynchronously. It'd be better to display the current import request status.
2. I believe polling interval could be less that 5 seconds as most imports should get completed within it.

novakzaballa: That's a good point @khvn26. However, precisely because it is taking a few seconds I think that it is a good/acceptable UX to show a blocking loading component during those few seconds, if the task would last around 1 minute or more instead, then I totally agree we should find another non-blocking UX behavior for this.

khvn26: Okay, good point. I guess it comes down to lessening the polling interval then. How about 1000 ms?

frontend/common/data/base/_data.js
novakzaballa: Yes, it was a common solution for this https://github.com/Flagsmith/flagsmith/pull/2842 too, which was just merged into main

frontend/web/components/pages/ImportPage.tsx
novakzaballa: Oh, right. I will lower that 1 sec

novakzaballa: Done.

khvn26 (approved):

frontend/common/data/base/_data.js
```diff
@@ -74,8 +74,8 @@ module.exports = {
     return this._request('delete', url, data, headers)
   },
 
-  get(url, data, headers) {
-    return this._request('get', url, data || null, headers)
+  get(url, data, headers, isExternal = false) {
```
kyle-ssg: I think I mentioned this in another PR, but we can just base this on if the url is Project.api in _request

novakzaballa: Done.

frontend/web/components/pages/ImportPage.tsx
```diff
@@ -0,0 +1,176 @@
...
+                            </div>
+                          </Row>
+                        </Button>
+                      </>
+                    )
+                  }}
+                  renderNoResults={
+                    <div>
+                      <Row>
+                        <div className='font-weight-medium'>No Prokects</div>
```
khvn26:
```suggestion
                        <div className='font-weight-medium'>No Projects</div>
```

PR #2847, opened 2023-10-13
feat(task-processor): Add priority support 

gagantrivedi:

## Changes

In order to allow high priority tasks to be executed first, this PR adds a priority column to the task table(with default null) 

Update unit test


api/task_processor/models.py
```diff
@@ -90,12 +101,14 @@ class Meta:
     def create(
         cls,
         task_identifier: str,
+        priority: TaskPriority = TaskPriority.NORMAL,
```
gagantrivedi, line 105: @matthewelwell think we can get rid of `schedule_task` and just use this method directly?

api/task_processor/migrations/sql/0011_get_tasks_to_process.sql
```diff
@@ -0,0 +1,30 @@
...
+RETURNS SETOF task_processor_task AS $$
+DECLARE
+    row_to_return task_processor_task;
+BEGIN
+    -- Select the tasks that needs to be processed
+    FOR row_to_return IN
+        SELECT *
+        FROM task_processor_task
+        WHERE num_failures < 3 AND scheduled_for < NOW() AND completed = FALSE AND is_locked = FALSE
+        ORDER BY priority ASC, scheduled_for ASC, created_at ASC
```
gagantrivedi, line 11: Only the above line is changed, but I had to create a different file because of versions

api/edge_api/identities/tasks.py
```diff
@@ -71,7 +72,7 @@ def call_environment_webhook_for_feature_state_change(
     call_environment_webhooks(environment, data, event_type=event_type)
 
 
-@register_task_handler()
+@register_task_handler(priority=TaskPriority.HIGHEST)
```
matthewelwell: Does this need to be highest? I think we basically want to reserve highest for those tasks involved in creating the environment document, right?

api/task_processor/models.py
matthewelwell: We _can_, but I quite like having the helper method to differentiate them. What's the benefit in removing it?

matthewelwell (commented):
I've added a couple of minor comments. 

This certainly seems like a ok approach but did we consider any other approaches here? For example, having distinct queues and the ability to define workers for those queues directly?  

I guess we could still define a number of workers that operate only on a given priority if we wanted to, right?

api/edge_api/identities/tasks.py
gagantrivedi: Think this is as import as environment document, right? We want the identity overrides to reflect as soon as possible

gagantrivedi:
> This certainly seems like a ok approach but did we consider any other approaches here? For example, having distinct queues and the ability to define workers for those queues directly?

Yes, but that approach was slightly more complex to maintain (according to my brief research). e.g:  do we create different table for different queue? Do we create different table for different priorities? Both approaches make adding a new task/priority level slightly more complex, without giving us much upside


> I guess we could still define a number of workers that operate only on a given priority if we wanted to, right?

Yes, with some code change

api/task_processor/models.py
gagantrivedi: I think the name is slightly odd, no? I'd expect it to schedule the task(i.e: put on some queue?) but it only creates the class instance. Does that make sense?

matthewelwell: I see, you mean that it doesn't actually persist it to the DB? But that's the same as the `create` method too, right? We could update the name to `create_scheduled_task` ? Maybe you're right though that we should consolidate the 2 methods.

api/edge_api/identities/tasks.py
matthewelwell: I'd say it's _slightly_ lower priority than environment document changes.

api/task_processor/models.py
gagantrivedi: > you mean that it doesn't actually persist it to the DB?

Yes, exactly

gagantrivedi: I've removed schedule_task method, but now I am not sure if it looks any better :confused: ?

api/core/migration_helpers.py
```diff
@@ -8,9 +8,14 @@
 
 class PostgresOnlyRunSQL(migrations.RunSQL):
     @classmethod
-    def from_sql_file(cls, file_path: str, reverse_sql: str) -> "PostgresOnlyRunSQL":
-        with open(file_path) as f:
-            return cls(f.read(), reverse_sql=reverse_sql)
+    def from_sql_file(
+        cls, file_path: str, reverse_sql: str = None, reverse_sql_file_path: str = None
```
matthewelwell: We should probably add validation here to ensure that `reverse_sql` and `reverse_sql_file_path` aren't used together? I guess an alternative would be to use a single argument and try to infer if it's a file path (e.g. `os.file.exists()`).

api/task_processor/migrations/0011_add_priority_to_get_tasks_to_process.py
```diff
@@ -0,0 +1,28 @@
...
+    ]
+
+    operations = [
+        PostgresOnlyRunSQL.from_sql_file(
+            os.path.join(
+                os.path.dirname(__file__),
+                "sql",
+                "0011_get_tasks_to_process.sql",
+            ),
+            # reverse_sql="DROP FUNCTION IF EXISTS get_tasks_to_process;",
```
matthewelwell: Can we remove this?

api/task_processor/models.py
matthewelwell: Well, there's less code so that's always a good thing. I think I would prefer to keep the logic regarding the queue in the Task model though. Can we just handle that in the `create` method now instead?

gagantrivedi: yeah, done

api/task_processor/models.py
```diff
@@ -90,44 +101,37 @@ class Meta:
...
         task_identifier: str,
+        scheduled_for: datetime,
+        priority: TaskPriority = TaskPriority.NORMAL,
+        queue_size: int = None,
         *,
         args: typing.Tuple[typing.Any] = None,
         kwargs: typing.Dict[str, typing.Any] = None,
     ) -> "Task":
+        if queue_size:
+            if cls.is_queue_full(task_identifier, queue_size):
```
matthewelwell: Nitpick: 

```suggestion
        if queue_size and cls.is_queue_full(task_identifier, queue_size):
```

api/task_processor/models.py
```diff
@@ -90,44 +101,37 @@ class Meta:
...
-            ):
-                raise TaskQueueFullError(
-                    f"Queue for task {task_identifier} is full. "
-                    f"Max queue size is {queue_size}"
-                )
-        task = cls.create(
-            task_identifier=task_identifier,
-            args=args,
-            kwargs=kwargs,
+    def is_queue_full(cls, task_identifier: str, queue_size: int) -> bool:
```
matthewelwell: This can probably be marked as private now, right? 

```suggestion
    def _is_queue_full(cls, task_identifier: str, queue_size: int) -> bool:
```

matthewelwell (approved):
Added a few minor comments but this looks good. It relies on us clearing down the `taskprocessor_task` table before we can release though.

PR #2852, opened 2023-10-16
chore(docs): Update minimum version for export / import

matthewelwell:
## Changes

Update documentation to represent correct minimum version for import / export.

Note that 2.28.2 is fine in 99% of situations, the fix that was added in 2.28.4 fixes a race condition regarding identity traits. 

N/a

### DIFF
--- a/docs/docs/deployment/configuration/importing-and-exporting.md
+++ b/docs/docs/deployment/configuration/importing-and-exporting.md
@@ -5,7 +5,7 @@ sidebar_position: 110
 
 :::tip
 
-You need to be running Flagsmith version `2.28.2` or greater to be able to export and import your Flagsmith data.
+You need to be running Flagsmith version `2.28.4` or greater to be able to export and import your Flagsmith data.
 
 :::
 


docs/docs/deployment/configuration/importing-and-exporting.md
```diff
@@ -5,7 +5,7 @@ sidebar_position: 110
 
 :::tip
 
-You need to be running Flagsmith version `2.28.2` or greater to be able to export and import your Flagsmith data.
+You need to be running Flagsmith version `2.28.4` or greater to be able to export and import your Flagsmith data.
```
khvn26, line 8: Should we add a note to upgrade Flagsmith to the latest version when migrating self-hosted->saas?

matthewelwell: Well, I think it might just work with 2.28.4, but I'm not sure. I would like to test this theory. If it doesn't work, we should add this, yes.

gnumoreno (approved):
Simple enough change, LGTM

PR #2863, opened 2023-10-19
feat(rate-limit/redis): Use redis to store throttling data for admin endpoints

gagantrivedi:

## Changes

Add support for using redis to store throttling data 

Adds unit tests


api/app/settings/common.py
```diff
@@ -645,6 +646,26 @@
...
 }
 
+REDIS_CACHE_NAME = "redis-cache"
+
+# Log exceptions generated by django-redis
+DJANGO_REDIS_LOG_IGNORED_EXCEPTIONS = env.bool(
+    "DJANGO_REDIS_LOG_IGNORED_EXCEPTIONS", True
+)
+
+# Uses redis-py native URL notation
```
matthewelwell: Can we add an e.g. here?

api/app/settings/common.py
```diff
@@ -645,6 +646,26 @@
     },
 }
 
+REDIS_CACHE_NAME = "redis-cache"
```
matthewelwell: Should the redis cache be used generically like this? Or should the name of the cache represent what is stored in it? (so we might end up with multiple caches using the same redis cache URL).

api/core/throttling.py
```diff
@@ -0,0 +1,23 @@
...
+from rest_framework.throttling import UserRateThrottle
+
+SDK_ENDPOINTS = [
+    "/api/v1/flags/",
+    "/api/v1/identities/",
+    "/api/v1/traits/",
+    "/api/v1/traits/bulk/",
+    "/api/v1/traits/increment-value/",
+    "/api/v1/environment-document/",
+]
```
matthewelwell: I have a feeling this is defined somewhere already? e.g. in the InfluxDBMiddleware or something? Also, for (very minor) performance benefits, should we make this a set?

api/tests/unit/core/test_throttling.py
```diff
@@ -0,0 +1,44 @@
...
+def test_admin_rate_throttle_allow_request_non_sdk_endpoint(rf, settings, test_user):
+    # Given
+    settings.REST_FRAMEWORK = {"DEFAULT_THROTTLE_RATES": {"admin": "1/minute"}}
+
+    # reload the throttling module to ensure the settings are reloaded
+    reload(rest_framework_throttling)
+    reload(core_throttling)
+
+    # When
+    request = rf.get("/api/v1/organizations/")
```
matthewelwell: Proper nitpick here but can we use `organisations` ?

🇬🇧

api/app/settings/common.py
gagantrivedi: Yes, good shout, I have removed it to allow user to decide which backend they want to use (limited to rate limiting for now)

api/core/throttling.py
gagantrivedi: Yeah, it was defined in sentry

api/core/constants.py
```diff
@@ -8,3 +8,15 @@
...
+
+SDK_ENDPOINTS = (
+    "/api/v1/flags",
+    "/api/v1/identities",
+    "/api/v1/traits",
+    "/api/v1/traits/bulk",
+    "/api/v1/traits/increment-value",
+    "/api/v1/environment-document",
+    "/api/v1/analytics/flags",
+)
```
matthewelwell: As per previous comments, should this be a set for minor perf improvements?

matthewelwell: Also, I think this will still fail for those that are still (somehow) using `/api/v1/identities/{identifier}`, right?

api/app/settings/common.py
```diff
@@ -597,6 +597,20 @@
...
+ADMIN_THROTTLE_CACHE_LOCATION = env.str(
+    "ADMIN_THROTTLE_CACHE_LOCATION", "admin-throttle"
+)
+
+IGNORE_CACHE_EXCEPTIONS = env.bool("IGNORE_CACHE_EXCEPTIONS", default=True)
+# Log exceptions generated by django-redis
+DJANGO_REDIS_LOG_IGNORED_EXCEPTIONS = env.bool(
+    "DJANGO_REDIS_LOG_IGNORED_EXCEPTIONS", True
+)
+
```
matthewelwell: These vars are slightly confusing, could we link out to some documentation here?

api/core/throttling.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class AdminRateThrottle(UserRateThrottle):
+    cache = caches[settings.ADMIN_THROTTLE_CACHE_NAME]
+    scope = "admin"
+
+    def allow_request(self, request: Request, view: object) -> bool:
+        path = request.path.rstrip("/")
+        if path in SDK_ENDPOINTS:
+            return True
```
matthewelwell: I guess the only way to resolve the issue above regarding the `/api/v1/identities/{identifier}` endpoint would be to use `startswith` or regex here which would complicate things slightly I guess. What do you think?

api/app/settings/common.py
matthewelwell: I think maybe this got lost when you refactored the settings, although I guess it technically doesn't have a place here anymore. We should add some documentation somewhere though perhaps? I know we need to add SaaS based user facing documentation but we also need to document it from a self hosted perspective - we can add an example redis URL there.

api/tests/integration/core/test_user_rate_throttle.py
```diff
@@ -0,0 +1,25 @@
...
+from pytest_lazyfixture import lazy_fixture
+from rest_framework import status
+
+
+@pytest.mark.parametrize(
+    "client",
+    [(lazy_fixture("admin_master_api_key_client")), (lazy_fixture("admin_client"))],
+)
+def test_user_throttle_can_throttle_admin_endpoints(
+    client, project, mocker, reset_cache
```
matthewelwell: Typing please

api/tests/integration/core/test_user_rate_throttle.py
```diff
@@ -0,0 +1,25 @@
...
+):
+    # Given
+    mocker.patch("core.throttling.UserRateThrottle.get_rate", return_value="1/minute")
```
matthewelwell, line 19: Shouldn't we use the `settings` fixture here? Or is that painful because of the nested dictionary issues we had previously?

api/tests/integration/core/test_user_rate_throttle.py
matthewelwell, line 1: I would definitely love to verify that all of our SDK endpoints are not throttled too.

matthewelwell (changes requested):

api/tests/integration/core/test_user_rate_throttle.py
gagantrivedi: We have those tests in their respective test modules, but I think you are right; moving them here makes more sense.

api/tests/integration/core/test_user_rate_throttle.py
gagantrivedi: Yep, correct!  Settings fixture does not work (correctly) for nested data structures

matthewelwell (approved):

khvn26 (approved):
LGTM

matthewelwell (approved):

gagantrivedi:
Will be merged after the release

PR #2864, opened 2023-10-19
feat: partial imports, off values as control value

### LINKED ISSUE #2838: Allow partial LaunchDarkly imports

Right now, LaunchDarkly imports are performed in a database transaction. This gives us two points of concern:

- Any errors during the import will lead to a complete rollback. No imported data will be saved.
- Potentially long transactions.

1. Remove the transaction, so the users are able to use the partially imported data.
1. Add per-entity status reports, so the users can access useful error info.

N/A

**Additional context**

N/A

khvn26:

## Changes

Closes #2838.

- Remove transaction from the import task
- Create control values for mulrivariate features from LD's off values
- Generic error logging

Modified relevant unit tests.

### DIFF
--- a/api/integrations/launch_darkly/services.py
+++ b/api/integrations/launch_darkly/services.py
@@ -3,7 +3,7 @@ from typing import TYPE_CHECKING, Callable, Tuple
 
 from django.core import signing
 from django.utils import timezone
-from requests.exceptions import HTTPError, RequestException
+from requests.exceptions import RequestException
 
 from environments.models import Environment
 from features.feature_types import MULTIVARIATE, STANDARD, FeatureType
@@ -39,6 +39,13 @@ def _unsign_ld_value(value: str, user_id: int) -> str:
         value,
         salt=f"ld_import_{user_id}",
     )
+
+
+def _log_error(
+    import_request: LaunchDarklyImportRequest,
+    error_message: str,
+) -> None:
+    import_request.status["error_message"] = error_message
 
 
 @contextmanager
@@ -164,24 +171,31 @@ def _create_mv_feature_states(
     environments_by_ld_environment_key: dict[str, Environment],
 ) -> None:
     variations = ld_flag["variations"]
+    variation_values_by_idx: dict[str, str] = {}
     mv_feature_options_by_variation: dict[str, MultivariateFeatureOption] = {}
 
     for idx, variation in enumerate(variations):
+        variation_idx = str(idx)
+        variation_value = variation["value"]
+        variation_values_by_idx[variation_idx] = variation_value
         (
             mv_feature_options_by_variation[str(idx)],
             _,
         ) = MultivariateFeatureOption.objects.update_or_create(
             feature=feature,
-            string_value=variation["value"],
+            string_value=variation_value,
             defaults={"default_percentage_allocation": 0, "type": STRING},
         )
 
     for ld_environment_key, environment in environments_by_ld_environment_key.items():
         ld_flag_config = ld_flag["environments"][ld_environment_key]
+
+        is_flag_on = ld_flag_config["on"]
+
         feature_state, _ = FeatureState.objects.update_or_create(
             feature=feature,
             environment=environment,
-            defaults={"enabled": ld_flag_config["on"]},
+            defaults={"enabled": is_flag_on},
         )
 
         cumulative_rollout = rollout_baseline = 0
@@ -189,6 +203,17 @@ def _create_mv_feature_states(
         if ld_flag_config_summary := ld_flag_config.get("_summary"):
             enabled_variations = ld_flag_config_summary.get("variations") or {}
             for variation_idx, variation_config in enabled_variations.items():
+                if variation_config.get("isOff"):
+                    # Set LD's off value as the control value.
+                    # We expect only one off variation.
+                    FeatureStateValue.objects.update_or_create(
+                        feature_state=feature_state,
+                        defaults={
+                            "type": STRING,
+                            "string_value": variation_values_by_idx[variation_idx],
+                        },
+                    )
+
                 mv_feature_option = mv_feature_options_by_variation[variation_idx]
                 percentage_allocation = 0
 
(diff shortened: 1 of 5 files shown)


api/integrations/launch_darkly/services.py
```diff
@@ -164,31 +171,49 @@ def _create_mv_feature_states(
...
             enabled_variations = ld_flag_config_summary.get("variations") or {}
             for variation_idx, variation_config in enabled_variations.items():
+                if variation_config.get("isOff"):
+                    # Set LD's off value as the control value.
+                    # We expect only one off variation.
+                    FeatureStateValue.objects.update_or_create(
+                        feature_state=feature_state,
+                        defaults={
+                            "type": STRING,
+                            "string_value": variation_values_by_idx[variation_idx],
```
matthewelwell, line 213: So I guess here we're expecting that the variations are in the same order in both `ld_flag["variations]` and `ld_flag["environments"][ld_environment_key]["_summary"]["variations"]`. This does seem like a fairly safe assumption but might not be guaranteed, right? Is there some other key we can use to guarantee we're using the correct variation?

api/integrations/launch_darkly/services.py
```diff
@@ -164,31 +171,49 @@ def _create_mv_feature_states(
...
+                        },
+                    )
+
```
matthewelwell, line 216: Shouldn't we `continue` after this? We don't need to create the MultivariateFeatureOption, right? Because, if I understand correctly, the control will receive the correct percentage by nature of it being the amount left over from the other variations?

api/integrations/launch_darkly/tasks.py
```diff
@@ -1,12 +1,9 @@
...
 from task_processor.decorators import register_task_handler
 
 
 @register_task_handler()
 def process_launch_darkly_import_request(import_request_id: int) -> None:
-    with transaction.atomic():
-        import_request = LaunchDarklyImportRequest.objects.get(id=import_request_id)
-        process_import_request(import_request)
+    import_request = LaunchDarklyImportRequest.objects.get(id=import_request_id)
+    process_import_request(import_request)
```
matthewelwell, line 9: Just so I'm clear here: by removing the `transaction.atomic` context manager, we're allowing imports to partially complete, right? So we might end up with half the data from LD imported? What's the thought proces here?

api/tests/unit/integrations/launch_darkly/test_services.py
```diff
@@ -60,20 +60,21 @@ def test_create_import_request__return_expected(
...
-        (Timeout(), "Timeout when requesting LaunchDarkly"),
+        (Timeout(), "Timeout when requesting /expected_path"),
     ],
 )
 def test_process_import_request__api_error__expected_status(
     ld_client_mock: MagicMock,
     ld_client_class_mock: MagicMock,
     failing_ld_client_method_name: str,
-    exception: Type[Exception],
+    exception: Type[RequestException],
```
matthewelwell: I don't think this is `Type[RequestException]` is it? Isn't it just `RequestException` ? We're instantiating a `RequestException` object.

api/tests/unit/integrations/launch_darkly/test_services.py
```diff
@@ -187,6 +188,10 @@ def test_process_import_request__success__expected_status(
     }
 
     assert percentage_mv_feature_states_by_env_name["Test"].enabled is False
+    assert (
+        percentage_mv_feature_states_by_env_name["Test"].get_feature_state_value()
+        == "variation2"
+    )
```
matthewelwell, line 195: It might be worth just adding a comment as to why we're checking the feature state value and why we expect it to be the given value?

api/integrations/launch_darkly/services.py
khvn26: I agree it's cringy, but it's the officially supported way. See the [documentation](https://apidocs.launchdarkly.com/tag/Feature-flags#section/Anatomy-of-a-feature-flag).

No idea why wouldn't they just use ids!

api/integrations/launch_darkly/services.py
khvn26: Sounds good, although we can't *not* create the `MultivariateFeatureOption`, so we'll also have this value as a 0% weight variation.

api/tests/unit/integrations/launch_darkly/test_services.py
khvn26: Thanks! Fixed.

api/tests/unit/integrations/launch_darkly/test_services.py
khvn26: Done.

api/integrations/launch_darkly/tasks.py
khvn26: We're solving a couple of problems here:

1. Created entities not available to the audit log workers ([related sentry issue](https://flagsmith.sentry.io/issues/4555706486/?alert_rule_id=14542360&alert_type=issue&notification_uuid=22537a24-c37e-433f-b1b2-bbc678327d51&project=5544478&referrer=slack))
3. Potentially long transactions

We'll also be able to surface detailed errors in import request statuses with attribution to specific entities that caused them, should we encounter such errors in the future (at this moment everything is expected to work smoothly).

The basic UX motivation/assumption for this is that having a partial import is better than having nothing imported at all. If we rethink this, we should solve 1. with delaying audit log task processor calls until the transaction ends.

api/integrations/launch_darkly/services.py
khvn26: This is done.

matthewelwell: Hmm.. based on the discussion on slack, I think actually it is better to create the MultivariateFeatureStateValue then. I think the FE may have issues if there are more Options than States. We can just create the Value with a 0 percentage as we were previously. Apologies for the confusion.

api/integrations/launch_darkly/tasks.py
matthewelwell: >The basic UX motivation/assumption for this is that having a partial import is better than having nothing imported at all.

I would probably argue the opposite here to be honest, but I don't think it's a massive issue.

api/integrations/launch_darkly/services.py
khvn26: Reverted.

api/integrations/launch_darkly/tasks.py
khvn26: Okay. Feel free to unresolve.

matthewelwell (approved):

PR #2865, opened 2023-10-19
feat: Rely on Flagsmith Engine for segment evaluation

khvn26:

## Changes

This PR removes legacy code responsible for segment rule evaluation and replaces it with Flagsmith engine implementation. 

1. Ran unit and integration tests.
2. Evaluated identities locally.


api/segments/models.py
```diff
@@ -109,9 +115,14 @@ def id_exists_in_rules_data(rules_data: typing.List[dict]) -> bool:
     def does_identity_match(
         self, identity: "Identity", traits: typing.List["Trait"] = None
     ) -> bool:
-        rules = self.rules.all()
-        return rules.count() > 0 and all(
-            rule.does_identity_match(identity, traits) for rule in rules
+        segment_model = map_segment_to_engine(self)
+        identity_model = map_identity_to_engine(identity)
+        trait_models = map_traits_to_trait_models(traits) if traits else None
```
matthewelwell: My concern here is that it's a bit hidden as to what additional database interactions are happening. I think, to plugin the engine, we need to do it much earlier in the flow and use e.g. `engine.get_identity_feature_states`.

khvn26: > additional database interactions

These are well covered by the currently failing `assertNumQueries` assertion in `test_get_segments`. IMO if we manage to keep the same amount (or less) by prefetching the data, we're good.

> do it much earlier in the flow and use e.g. engine.get_identity_feature_states.

I'll think of it but my primary focus here is to replace the segment matching logic first.

matthewelwell: >IMO if we manage to keep the same amount (or less) by prefetching the data, we're good.

Yep, fair enough.

api/environments/identities/tests/test_views.py
```diff
@@ -370,7 +370,7 @@ def test_identities_endpoint_returns_traits(self, mock_amplitude_wrapper):
         trait = Trait.objects.create(
             identity=self.identity,
             trait_key="trait_key",
-            value_type="STRING",
+            value_type=STRING,
```
matthewelwell, line 373: I tend to think this is because those tests were wrong (and we were lucky that the default type is string) but it's worth noting that this is a slight change `"STRING"` -> `"unicode"`.

matthewelwell (approved):
Approved with one minor comment. So goo d being able to delete all this code. I'm assuming as well that, since there are no test failures, we haven't had to introduce any new queries here.

api/environments/identities/tests/test_views.py
khvn26: Yep. The engine mapper relies on `Trait`'s `trait_value` property which doesn't employ the default string fallback.

PR #2867, opened 2023-10-19
ci: Run only API tests affected by changes in PRs and Upgrade GHA runners

novakzaballa:

- [X ] I have filled in the "Changes" section below?

## Changes

Use `pytest-testmon` to only execute API tests affected by changed files. Also upgrade GHA runners for this specific job.

API Tests executed as expected. Also testing manually using this PR to ensure that only tests affected by changes are exectured.

## Implementation notes
Notice that for retrieving the cache we use `github.event.pull_request.base.sha` as the final part of the cache key, and when we save the tesmon database file to the cache we use `github.sha`. That is because we restore cache only on PR events while we save the cache only on the push to main event, and `github.event.pull_request.base.sha` on the PR event context should be the same as `github.sha` of the base branch on the push event context.

Also, notice that we are using absolute paths for storing/retrieving the tesmon database file to/from the cache. We do that because if we use a relative path it will not be relative to the working directory, but to the workspace directory, which can vary depending on the context, so in this particular case I think it is better to stick to the absolute path. For more details refer to the comments below.

### DIFF
--- a/.github/workflows/api-pull-request.yml
+++ b/.github/workflows/api-pull-request.yml
@@ -69,7 +69,7 @@ jobs:
         run: make django-make-migrations
 
       - name: Restore cached testmon data
-        if: ${{ github.event_name != 'push' || github.ref != 'refs/heads/main' }}
+        if: ${{ github.event_name == 'pull_request' }}
         id: cache-testmon-restore
         uses: actions/cache/restore@v3
         with:
--- a/.github/workflows/api-pull-request.yml
+++ b/.github/workflows/api-pull-request.yml
@@ -12,6 +12,8 @@ on:
       - .github/**
     branches:
       - main
+  merge_group:
+    types: [checks_requested]
 
 defaults:
   run:
@@ -19,7 +21,7 @@ defaults:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    runs-on: General-Purpose-8c-Runner
     name: API Unit Tests
 
     services:
@@ -33,7 +35,7 @@ jobs:
         options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
 
     strategy:
-      max-parallel: 4
+      max-parallel: 2
       matrix:
         python-version: ['3.10', '3.11']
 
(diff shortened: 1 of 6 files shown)


.github/workflows/api-pull-request.yml
```diff
@@ -66,11 +68,31 @@ jobs:
...
         run: make django-make-migrations
 
+      - name: Restore cached testmon data
+        if: ${{ github.event_name != 'push' || github.ref != 'refs/heads/main' }}
+        id: cache-testmon-restore
+        uses: actions/cache/restore@v3
+        with:
+          enableCrossOsArchive: true
+          path: |
+            /home/runner/work/flagsmith/flagsmith/api/.testmondata*
```
khvn26, line 78: Any particular reason we're not using a relative path here?

.github/workflows/api-pull-request.yml
```diff
@@ -66,11 +68,31 @@ jobs:
...
+          key: testmon-data-python${{ matrix.python-version }}-${{ github.event.pull_request.base.sha }}
```
khvn26, line 79: `github.event.pull_request.base.sha` is not the same as `github.sha` used when caching the data, right? Is this intentional?

.gitignore
```diff
@@ -23,3 +23,6 @@ src/CI_COMMIT_SHA
 .DS_Store
 .idea
 *.iml
+api/.testmondata
+api/.testmondata-shm
+api/.testmondata-wal
```
khvn26: `api` has its own gitignore at `api/.gitignore`. Please use it.

khvn26 (changes requested):

.github/workflows/api-pull-request.yml
novakzaballa: Right, they are not the same if they are used in the same context, but if you look at the condition this cache restore will run in the context of a Pull Request, while the save cache action will run in the context of a push to main. I improved the condition to clearly run this restore cache action only in the context of a PR:
`if: ${{ github.event_name == 'pull_request' }}`

.github/workflows/api-pull-request.yml
novakzaballa: Good question. In my first attempts, I had issues storing and retrieving the `testmon` output files, I am using relative paths to discard the relative path as a source of issues. This change https://github.com/github/docs/pull/12853/files in the docs explains a bit the issue I had and why I preferred to use an absolute path, since the relative one would not be relative to the working directory but the workspace directory. And this one explains how the workspcae directory could be fuzzyly defined https://github.com/actions/runner/issues/2058.

.gitignore
novakzaballa: Good call. I will.

.github/workflows/api-pull-request.yml
khvn26: Thanks!

khvn26: Thanks!

.github/workflows/api-pull-request.yml
khvn26: Great, thanks for explaining. I'd love to see it explained in a comment because the reasons are not immediately obvious.

novakzaballa: I updated the PR description to mention those reasons as suggested.

khvn26 (approved):

PR #2869, opened 2023-10-22
fix(api): validate before creating projects based on current subscription

### LINKED ISSUE #2835: Projects can be created via API under the free plan
**Describe the bug**

As an admin user of a free plan org, a call like this:

```
curl 'https://api.flagsmith.com/api/v1/projects/' \
  -H 'authority: api.flagsmith.com' \
  -H 'accept: application/json' \
  -H 'accept-language: en-US,en;q=0.9' \
  -H 'authorization: Token <user token>' \
  -H 'content-type: application/json; charset=utf-8' \
  -H 'dnt: 1' \
  -H 'origin: https://app.flagsmith.com' \
  -H 'referer: https://app.flagsmith.com/' \
  -H 'sec-ch-ua: "Chromium";v="117", "Not;A=Brand";v="8"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "macOS"' \
  -H 'sec-fetch-dest: empty' \
  -H 'sec-fetch-mode: cors' \
  -H 'sec-fetch-site: same-site' \
  -H 'user-agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/117.0.0.0 Safari/537.36' \
  --data-raw '{"name":"hacked","organisation":<org id>}'
```

will result in a new project created, even though the plan forbids it. 

**Expected behavior**

A 403 error returned and no project created.

**Screenshots**

image

(Two 'hacked' projects created via API)

- [ ] Self Hosted with Docker
- [x] SaaS at flagsmith.com
- [ ] Some other way

tushar5526:
Closes #2835 

## Changes

Subscription metadata contains the number of projects allowed in a subscription, use that to validate and create project requests.

PS: All the tests are written assuming that the default subscription allows the creation of multiple projects, therefore I had to allow up to the creation of 10 projects in the default subscription for tests. Another way could to be change the default subscription for every failing test that requires the creation of multiple projects. 

Created a test ORG and tried to create multiple projects via API.


api/projects/serializers.py
```diff
@@ -55,6 +55,15 @@ def get_use_edge_identities(self, obj: Project) -> bool:
...
         )
 
+    def validate(self, data):
+        subscription_metadata = data[
+            "organisation"
+        ].subscription.get_subscription_metadata()
+        total_projects = len(Project.objects.filter(organisation=data["organisation"]))
+        if subscription_metadata.projects <= total_projects:
+            raise serializers.ValidationError("Max projects reached for subscription")
+        return data
```
khvn26: I'd very much prefer to see this logic in the `ProjectPermissions` class (api/projects/permissions.py).

khvn26: Additionally:
- Please don't skip on running tests (`cd api && make docker-up tests`) on your machine to see what's broken.
- Cover your code with tests. Once your tests pass the Codecov bot will report coverage in the PR comments.

tushar5526: Sure, I have moved it to ProjectPermissions and added some tests as well.

api/app/settings/common.py
```diff
@@ -53,6 +53,9 @@
 
 HOSTED_SEATS_LIMIT = env.int("HOSTED_SEATS_LIMIT", default=0)
 
+MAX_SEATS_IN_FREE_PLAN = 1
+MAX_PROJECTS_IN_FREE_PLAN = 1
+
```
khvn26: @matthewelwell They are not settable from env, but do we want to expose these constants as settings?

matthewelwell: I don't see a reason to move them from `organisations.subscriptions.constants` really, no.

tushar5526: @matthewelwell @khvn26 please refer the PS section in the PR description above on why I made this change. 

Apparently, there are a few tests that create multiple projects for a single organisation on the default free subscription and error out. 

I can maybe look into changing the default subscription for tests to something else.

api/projects/permissions.py
```diff
@@ -41,6 +41,16 @@ def has_permission(self, request, view):
             organisation = Organisation.objects.get(
                 id=int(request.data.get("organisation"))
             )
```
matthewelwell: Can we `select_related` here since we're using the subscription below?

api/projects/permissions.py
```diff
@@ -41,6 +41,16 @@ def has_permission(self, request, view):
...
+
+            # Allow project creation based on the active subscription
+            subscription_metadata = (
+                organisation.subscription.get_subscription_metadata()
+            )
+            total_projects_created = Project.objects.filter(
+                organisation=organisation
+            ).count()
+            if total_projects_created >= subscription_metadata.projects:
```
matthewelwell: I think `subscription_metadata.projects` can be (and is by default) `None` which indicates no limit on projects. I think we should therefore replace this with

```suggestion
            if subscription_metadata.projects and total_projects_created >= subscription_metadata.projects:
```

api/projects/tests/test_permissions.py
```diff
@@ -371,3 +372,29 @@ def test_regular_user_has_no_destroy_permission(self):
...
+    )
+    user_permission_group.users.add(user)
+    user.add_organisation(organisation, OrganisationRole.ADMIN)
+
+    project_permissions = ProjectPermissions()
+
+    mock_view.action = "create"
+    mock_view.detail = False
+    mock_request.data = {"name": "Test", "organisation": organisation.id}
+    mock_request.user = user
```
matthewelwell: I appreciate that you're following an existing pattern here, but I feel like this is going to cause us problems in the future. We're patching things onto global variables which won't be cleared before running other tests. It might result in unexpected failures depending on the order tests are run in for example. I think we should just instantiate the mocks in this individual test (and, for another PR, in the `setUp` method on the above test class perhaps). It also shouldn't add any more lines of code, because we can just pass the attributes to the MagicMock constructor.

api/projects/tests/test_permissions.py
```diff
@@ -371,3 +372,29 @@ def test_regular_user_has_no_destroy_permission(self):
...
+
+    for i in range(MAX_PROJECTS_IN_FREE_PLAN):
+        result = project_permissions.has_permission(mock_request, mock_view)
+        Project.objects.create(name=f"Test project{i}", organisation=organisation)
+        assert result
```
matthewelwell: It's slightly odd to me that we're asserting the result after creating the project. I think we should change this to something like: 

```suggestion
    for i in range(MAX_PROJECTS_IN_FREE_PLAN):
        assert project_permissions.has_permission(mock_request, mock_view)
        Project.objects.create(name=f"Test project{i}", organisation=organisation)
```

api/projects/tests/test_permissions.py
```diff
@@ -371,3 +372,29 @@ def test_regular_user_has_no_destroy_permission(self):
 
         # Then - exception thrown
         assert not result
+
+
+@pytest.mark.django_db
+def test_free_plan_has_only_fixed_projects_permission():
```
matthewelwell, line 378: Can we try to follow the `# Given / When / Then` pattern here. The test is structured quite nicely, but I'd love to see the sections split out appropriately with comments.

api/app/settings/common.py
matthewelwell: Good point! I don't love the fact that this will cause the tests to fail when not using `DJANGO_SETTINGS_MODULE=app.settings.test` but I don't necessarily see another option here. We might just see some pain while we ensure that everything in our CI process is using that module.

matthewelwell (commented):
Thanks again for this @tushar5526 ! I've added a few (mostly minor) comments but this looks good on the whole.

api/app/settings/common.py
tushar5526: https://github.com/Flagsmith/flagsmith/blob/main/api/pyproject.toml#L36C1-L38C1 - I see pytest to be defaulted to `app.settings.test`, what else can break here?

api/projects/tests/test_permissions.py
tushar5526: Yes, makes sense. I can create a separate issue for this refactoring.

api/app/settings/common.py
matthewelwell: Hopefully nothing! 😄

api/organisations/subscriptions/constants.py
```diff
@@ -1,10 +1,12 @@
 from enum import Enum
 
+from django.conf import settings
+
 from organisations.subscriptions.metadata import BaseSubscriptionMetadata
 
-MAX_SEATS_IN_FREE_PLAN = 1
+MAX_SEATS_IN_FREE_PLAN = settings.MAX_SEATS_IN_FREE_PLAN or 1
```
matthewelwell: Unless someone creates a settings file that overrides these settings to `None` (which is their own fault), I can't see any scenario where the `or 1` is needed here. I'd say we should remove.

api/projects/tests/test_permissions.py
```diff
@@ -371,3 +372,31 @@ def test_regular_user_has_no_destroy_permission(self):
...
+    # Given
+    organisation = Organisation.objects.create(name="Test organisation")
+
+    user = FFAdminUser.objects.create(email="admin@test.com")
+    user_permission_group = UserPermissionGroup.objects.create(
+        name="Users", organisation=organisation
+    )
+    user_permission_group.users.add(user)
```
matthewelwell: I don't _think_ we need to add the user to a group here?

matthewelwell (approved):
I've added a couple of minor comments but on the whole, this looks great. Thanks again @tushar5526 .

api/organisations/subscriptions/constants.py
tushar5526: Yes, makes sense to remove it. do you want me to replace each occurrence of `MAX_SEATS_IN_FREE_PLAN`  with `settings.MAX_SEATS_IN_FREE_PLAN` as well?  (same for `MAX_PROJECTS_IN_FREE_PLAN`) I find it more readable to have all constants coming from a `constants` file 🤔 but I am not sure here.

api/projects/tests/test_permissions.py
tushar5526: Yes, not needed. Thanks!

api/organisations/subscriptions/constants.py
matthewelwell: Ideally we'd just use `settings.` everywhere, yeah.

matthewelwell (approved):

PR #2877, opened 2023-10-23
feat: Allow organisation admins to mandate 2fa for their organisation

### LINKED ISSUE #411: Allow organisation admins to mandate 2fa for their organisation
This should be a scale-up only feature. 

Need to figure out how we upgrade existing accounts that don't have 2fa but are logging into an organisation that has it mandated.

novakzaballa:

## Changes

- Enable the Enforce 2FA function (This feature is not enabled for users who log in with SSO and/or SAML)

- Go to Organisation Settings
- In the "General" tab enable Enforce 2FA
- Log in with an account that does not have Two-Factor Authentication enabled


frontend/common/dispatcher/app-actions.js
```diff
@@ -21,9 +21,10 @@ const AppActions = Object.assign({}, require('./base/_app-actions'), {
       identity,
     })
   },
-  confirmTwoFactor(pin, onError) {
+  confirmTwoFactor(pin, onError, isLoginPage) {
```
kyle-ssg, line 24: Im a bit confused as to why any frontend code was changed, we already have 2fa, shouldn't this just use what was there?

frontend/web/components/App.js
```diff
@@ -275,7 +275,7 @@ const App = class extends Component {
       )
     }
     if (AccountStore.forced2Factor()) {
```
kyle-ssg, line 277: I think instead of making all of the isLoginChanges, just adjust this function to return false if 2fa has been set. It should remove the need for any other frontend change in this PR.

frontend/common/stores/account-store.js
```diff
@@ -381,8 +385,12 @@ const store = Object.assign({}, BaseStore, {
       return false
     }
 
+    if (store.samlOrOauth) {
```
kyle-ssg, line 388: Can be replaced with my other suggestion.

frontend/web/components/App.js
novakzaballa: This change was made to handle the case where you are already logged in, your organization forces 2factor and you disable your 2factor, it prevents the page from reloading, which does happen if 2factor is enabled on login.

frontend/common/stores/account-store.js
novakzaballa: This prevents the 2factor from being forced on Oauth or Saml accounts

novakzaballa:
@kyle-ssg Can you re-review, please?

kyle-ssg (approved):

PR #2878, opened 2023-10-24
feat(rate-limit): allow user to pass default throttle classes

gagantrivedi:

## Changes

allow user to pass default throttle classes(defaults to [])

Covered by unit tests


api/app/settings/common.py
```diff
@@ -228,7 +229,7 @@
     "PAGE_SIZE": 10,
     "UNICODE_JSON": False,
     "DEFAULT_PAGINATION_CLASS": "rest_framework.pagination.PageNumberPagination",
-    "DEFAULT_THROTTLE_CLASSES": ["rest_framework.throttling.UserRateThrottle"],
+    "DEFAULT_THROTTLE_CLASSES": DEFAULT_THROTTLE_CLASSES,
```
matthewelwell, line 232: I worry that this is actually a regression on what we had before we added this throttling logic. I assume we need something to pick up the `DEFAULT_THROTTLE_RATES` below, right?

matthewelwell: `[]` is the default provided by DRF. The throttle rates are used when the view has an explicit throttle class defined.

matthewelwell (approved):

Issue #2873, opened 2023-10-23
Per Project option to make Flag Description a required field


dabeeeenster:
It should be possible at the project level to define which attributes are required on a feature. Initially, this should only include the description field, but we should build this in such a way that other fields can be marked as required in the future. 

Only Project Admins should be able to configure which fields are required in a given project.

jatingodnani:
@dabeeeenster tell this problem in more detailed way.i am ready to work on this issue

gnumoreno:
Currently when you are creating a Feature Flag the Description is an optional field. See the screenshot.

![image](https://github.com/Flagsmith/flagsmith/assets/7240118/b4cb1b8a-cc59-46da-9f42-cff63eada75f)

If I understood @dabeeeenster correctly what we want is a feature flag that can make the description a required/mandatory field on the Project level so that you could have a Project where you always have to provide the description when creating a new feature flag.

jatingodnani:
Ok thanks @matthewelwell i will do this

SuperGrut:
@matthewelwell I would like to work on the issue, could you please assign it to me?

matthewelwell:
@akash19coder great, thanks! I've done that now.

matthewelwell:
Closing this as stale.

PR #2897, opened 2023-10-27
changes some code to mandatory to write description in new feature

jatingodnani:

- [ -] I have run [`pre-commit`](https://docs.flagsmith.com/platform/contributing#pre-commit) to check linting
- [- ] I have filled in the "Changes" section below?
- [ -] I have filled in the "How did you test this code" section below?
- [ -] I have used a [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) title for this Pull Request

#2873 

_Please describe._

=>changes some code to mandatory to fill the description

=>in localhost:8080

### DIFF
--- a/frontend/web/components/modals/CreateFlag.js
+++ b/frontend/web/components/modals/CreateFlag.js
@@ -541,7 +541,7 @@ const CreateFlag = class extends Component {
             }
             ds
             type='text'
-            title={identity ? 'Description' : 'Description (optional)'}
+            title={'Description*'}
             placeholder="e.g. 'This determines what size the header is' "
           />
         </FormGroup>
@@ -1534,6 +1534,7 @@ const CreateFlag = class extends Component {
                                       id='create-feature-btn'
                                       disabled={
                                         isSaving ||
+                                        !description||
                                         !name ||
                                         invalid ||
                                         !regexValid ||


matthewelwell (changes requested):
@jatingodnani this does not meet the requirements at all. You'll need to add some changes to the ORM such that the description can be made optional only for certain projects. The changes you have added make the description a required field (in the UI only) for all projects.

jatingodnani:
@matthewelwell ok,my question is which type of project require optional field.if you can give me more detail then it will easy to understand

jatingodnani:
You can tell me the min length of description

matthewelwell:
That's for the user to define. To make this generic enough so we can use it for multiple fields on different objects, controlled by different parent objects, I am imagining we ideally want something like: 

```python
class RequiredField(Model):
    parent_object_content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
    parent_object_id = GenericObjectID()
    child_object_content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
    field_name = models.CharField(max_length=200)
```

This would allow us to, for example, configure that the `description` field is required for all features in a given project but it will also allow us to, for example, say that the `description` field is required for all segments in a given project too.

This is quite complex django logic, however, so let me know if this is not in your wheelhouse. 

A slightly simpler (but acceptable) solution that will likely cover all the use cases I can think of is to link it directly to a project only. That way the model becomes more like: 

```python
class ProjectEntityRequiredField(AbstractBaseExportableModel):
    project = models.ForeignKey(Project, on_delete=models.CASCADE, related_name="required_fields_for_child_entitites")
    child_object_content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
    field_name = models.CharField(max_length=200)
```

Then, to enforce that the description field is required on the Feature model for a given project, you would need to do something like: 

```python
feature_content_type = ContentType.objects.get(app_label="features", model="feature")
project = Project.objects.get(name="my-project")
ProjectEntityRequiredField.objects.create(project=some_project, child_object_content_type=feature_content_type, field_name="description")
```

and to use that data in the validation, you can just do something like this:

```python
project = Project.objects.get(name="my-project")
required_fields = project.required_fields_for_child_entitites.filter(child_object_content_type__model="feature").all().values_list("field_name", flat=True)
```

jatingodnani:
@matthewelwell if this is form certain project,then i think 
It should be kept as optional

matthewelwell:
@jatingodnani the requirement is to make it required for certain projects. If you're not comfortable implementing this, please close the PR and we will reassign the issue in Github.

jatingodnani:
@matthewelwell thankyou for response.i dont know django and i am closing this pr

PR #2898, opened 2023-10-27
chore: upgrade django

### LINKED ISSUE #2088: Upgrade to Django 4.2
https://www.djangoproject.com/weblog/2023/apr/03/django-42-released/

Known TODOs:
 * Replace / remove third party pytest plugin with the one available in pytest-django [here](https://github.com/Flagsmith/flagsmith/blob/1aa00c97f0afea709436fac328e923bc1024740a/api/tests/unit/task_processor/test_unit_task_processor_decorators.py#L62-L63)

SD-13:
Fixes #2088 

## Changes

I started with upgrading to a new feature release in each step (e.g 2.1 -> 2.2 -> 2.3 -> 4.0) and fixing errors.

I am testing the code by running the postgresql db locally and running the tests using the make cmd. And trying to fix the failing tests.


SD-13:
I am not sure at this point what to do to fix an error and for that I opened this PR. As you can see in the commit messages I upgraded trench to 0.3.1 to avoid version conflict (ugettext was deprecated in django but the older version was still using it). After that I found that some classes we are using from trench are removed or I would say replaced in the new release (see [this commit](https://github.com/Flagsmith/flagsmith/pull/2898/commits/0e89e4a490285f705dce012b4106bcd8570a8f83). So what should we do to fix this, should we try to fit the new classes or look for an alternative?
https://github.com/Flagsmith/flagsmith/issues/2088#issuecomment-1757156585

api/pyproject.toml
```diff
@@ -70,12 +70,12 @@ djangorestframework-recursive = "~0.1.2"
 packaging = "~23.0"
 chargebee = "~2.7.7"
 python-http-client = "~3.3.7"
-django-health-check = "~3.14.3"
-django-admin-sso = "~3.1.0"
+django-health-check = "~3.17.0"
+django-admin-sso = "~5.0"
```
matthewelwell, line 74: We should make sure to manually test this

api/pyproject.toml
```diff
@@ -45,7 +45,7 @@ readme = "readme.md"
 
 [tool.poetry.dependencies]
 python = "^3.10"
-django = "~3.2.19"
+django = "~4.0"
```
matthewelwell, line 48: This should be `"~4.2"` I think?

SD-13: Yeah, I tried to jump the version to 4.2 at once and ended up failing all the tests with errors. That's why I was upgrading the version by one feature release and fixing errors in each (as recommended by [this doc](https://docs.djangoproject.com/en/4.2/howto/upgrade-version/))

api/pyproject.toml
SD-13: Noted! Once we figure out how to use the new trench classes, we can do that.

SD-13:
Hi @matthewelwell I have replied to your comments. The blocker in this PR is that I am not able to figure out how to use the new trench classes introduced in the last trench version. We are using them in `api/custom_auth/views.py`. Please let me know if you need more context on this. Thanks!

matthewelwell:
@SD-13 thank you for your efforts here but, given the importance of this change, combined with the fact that we'll need to organise notifications to certain customers due to django 4 deprecating postgres 11 support and older versions of oracle, we're going to take this work on internally at Flagsmith. 

If you'd like another task to work, please do let me know. 

Thanks again!

SD-13:
@matthewelwell Thanks for letting me know!
and yes, I would like to take some tasks.

matthewelwell:
@SD-13 please check the filtered list here: https://github.com/Flagsmith/flagsmith/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22

Issue #2893, opened 2023-10-26
DevExp: local environment


khvn26:
This is a tracking issue for current problems when trying to bring up Flagsmith locally:

cc @gnumoreno 

### Common

- [ ] No single documented way of doing this. We should also document prerequisites i.e. pyenv, Docker and npm.

### API

- [ ] `make install` failing due to Poetry not updating `$PATH` out of the box.
- [ ] `make serve` target should invoke `docker-up` and `django-migrate` targets automatically.

### Frontend

- [ ] `make build` target is incorrect, should invoke `npm run bundle` instead of `npm run build`.
- [ ] Should add `ENV=local` to `.env-local` to point frontend brought up by the `make serve` target to the local API.

PR #2902, opened 2023-10-30
docs: Add minor changes for project build

zachaysan:

## Changes
Added a few small nice to haves that would have saved me some time.

## Other possible change
In the same file it says:
    You'll also need to ensure that you have a value for `POSTGRES_PASSWORD` set as an environment variable on your development machine.

But I have not needed to add this environment variable to get the application running or to get the tests to pass. Is this no longer needed?

### DIFF
--- a/docs/docs/deployment/hosting/locally-api.md
+++ b/docs/docs/deployment/hosting/locally-api.md
@@ -17,7 +17,7 @@ make serve
 ```
 
 You can now visit `http://<your-server-domain:8000>/api/v1/users/config/init/` to create an initial Superuser and
-provide DNS settings for your installation.
+provide DNS settings for your installation or run `make test` from the `api` directory to run the test suite.
 
 Note: if you're running on on MacOS and you find some issues installing the dependencies (specifically around pyre2),
 you may need to run the following:
@@ -436,9 +436,19 @@ You can run Flagsmith as a single application/docker container using our unified
 [Docker Hub](https://hub.docker.com/repository/docker/flagsmith/flagsmith) but you can also run the front end as part of
 the Django Application. Steps to do this:
 
-1. `cd frontend; npm run bundledjango`
-2. `cd ../api; python manage.py collectstatic`
-3. `python manage.py runserver`
+```bash
+# Update packages and build django.
+cd frontend
+npm install
+npm run bundledjango
+
+# Copy additional assets with Django
+cd ../api
+python manage.py collectstatic
+
+# Boot the server
+python manage.py runserver
+```
 
 ### How it works
 


dabeeeenster (approved):

dabeeeenster:
> You'll also need to ensure that you have a value for POSTGRES_PASSWORD set as an environment variable on your development machine.

> But I have not needed to add this environment variable to get the application running or to get the tests to pass. Is this no longer needed?

Maybe you were just lucky and no PW was needed connecting to localhost?

khvn26:
> But I have not needed to add this environment variable to get the application running or to get the tests to pass. Is this no longer needed?

That's correct! We plan on updating those docs very soon (https://github.com/Flagsmith/flagsmith/issues/2893)

docs/docs/deployment/hosting/locally-api.md
```diff
@@ -436,9 +436,19 @@ You can run Flagsmith as a single application/docker container using our unified
...
+cd frontend
+npm install
+npm run bundledjango
+
+# Copy additional assets with Django
+cd ../api
+python manage.py collectstatic
+
+# Boot the server
+python manage.py runserver
```
khvn26, line 450: We have Makefile targets for that. Can we use these?

matthewelwell: @khvn26 interestingly, we _don't_ have a makefile target for `runserver`. I'm not sure if there was a reason though? 

I tend to agree that for these type of docs we _should_ use the makefile targets where possible but we should perhaps have a discussion (in the next eng meeting perhaps) about the makefiles now that we have onboarded a new engineer.

khvn26: > interestingly, we don't have a makefile target for runserver

We do: `make serve` spins up gunicorn with livereload enabled.

khvn26: If any points are to be made ahead of the discussion, I suggest adding them here: https://github.com/Flagsmith/flagsmith/issues/2893

matthewelwell: > We do: `make serve` spins up gunicorn with livereload enabled.

I don't think this is the same though, right? `runserver` does other things (e.g. you don't have to set ALLOWED_HOSTS and DEBUG environment variables).

khvn26: Good point! We probably should have those set up for the local environment regardless of which dev server we use.

matthewelwell (approved):

docs/docs/deployment/hosting/locally-api.md
khvn26: Resolving this in hopes of a later iteration — current docs are correct enough.

PR #2908, opened 2023-10-31
feat: add group owners to features

zachaysan:

## Changes

This adds two endpoints to features. The first is to allow the addition of groups as owners of a feature. Think groups like "design department" which could have many users under them. It also allows the removal of groups through a separate, similar endpoint. 

This work supports [this](https://github.com/orgs/Flagsmith/projects/2/views/11?pane=issue&itemId=43075831) ticket, though I would like to get feedback before implementing any filtering of features so that part of the ticket is not yet complete.

I wrote two new tests that go through the API. I was a little confused as to why the practice was to use a master key for authentication, but I kept with the normal way of doing things since that is how other parts of the codebase around that part did it.


api/features/serializers.py
```diff
@@ -11,7 +11,12 @@
     HideSensitiveFieldsSerializerMixin,
 )
 from projects.models import Project
-from users.serializers import UserIdsSerializer, UserListSerializer
+from users.models import UserPermissionGroup
+from users.serializers import (
+    MyUserPermissionGroupsSerializer,
```
matthewelwell: Love the reuse of this serializer, but I think we should probably find a more generic name for the serializer. Alternatively, we could do an `import MyUserPermissionGroupsSerializer as ...` but I'd rather just rename the original serializer.

api/features/serializers.py
```diff
@@ -36,8 +41,29 @@ def remove_users(self, feature: Feature):
...
+class FeatureGroupOwnerInputSerializer(serializers.Serializer):
+    group_ids = serializers.ListField(child=serializers.IntegerField())
+
+    def validate(self, data):
+        if not UserPermissionGroup.objects.filter(
+            id__in=data["group_ids"]
+        ).count() == len(data["group_ids"]):
+            raise serializers.ValidationError("Some groups not found")
+
+        return data
```
matthewelwell: We may want to consider validating that the groups belong at least to the organisation for the feature we are adding them to.

api/tests/unit/features/test_unit_features_views.py
matthewelwell, line 1: Related to the above comment, it would be great to see some negative test cases here - what if a user tries to add a group that doesn't belong to the same organisation as the feature for example? What if a user tries to remove a group that doesn't exist?

api/features/serializers.py
zachaysan: Good idea! I'll rename the serializer to be more generic

api/tests/unit/features/test_unit_features_views.py
zachaysan: Good questions.

In my opinion, trying to remove a group that doesn't exist should work without error. We could have, for example, two users trying to do this at the same time, and trying to remove something that isn't connected and having it silently pass could help in that case. 

The other question is trickier. I think putting it in the filter clause and saying "could not find group" is acceptable since the group isn't associated to the same organisation so it isn't really being found. 

Let me update the code with what I'm thinking then we can discuss further.

api/features/serializers.py
zachaysan: Ok this has been implemented.

api/features/views.py
```diff
@@ -157,6 +158,38 @@ def get_serializer_context(self):
...
+        responses={200: ProjectFeatureSerializer},
+    )
+    @action(detail=True, methods=["POST"], url_path="remove-group-owners")
+    def remove_group_owners(self, request, *args, **kwargs):
+        feature = self.get_object()
+        data = request.data
+        data["organisation_id"] = feature.project.organisation.pk
+        serializer = FeatureGroupOwnerInputSerializer(data=data)
+        serializer.is_valid(raise_exception=True)
+        feature = self.get_object()
```
matthewelwell: I don't think we need to get the `feature` again here, right?

api/features/views.py
```diff
@@ -157,6 +158,38 @@ def get_serializer_context(self):
...
 
+    @swagger_auto_schema(
+        request_body=FeatureGroupOwnerInputSerializer,
+        responses={200: ProjectFeatureSerializer},
+    )
+    @action(detail=True, methods=["POST"], url_path="add-group-owners")
+    def add_group_owners(self, request, *args, **kwargs):
+        feature = self.get_object()
+        data = request.data
+        data["organisation_id"] = feature.project.organisation.pk
```
matthewelwell: Little performance trick to save an extra query across the foreign key: 

```suggestion
        data["organisation_id"] = feature.project.organisation_id
```

api/features/serializers.py
```diff
@@ -36,8 +41,31 @@ def remove_users(self, feature: Feature):
         feature.owners.remove(*user_ids)
 
 
+class FeatureGroupOwnerInputSerializer(serializers.Serializer):
+    group_ids = serializers.ListField(child=serializers.IntegerField())
+    organisation_id = serializers.IntegerField()
```
matthewelwell: Thinking this through - my issue with adding the `organisation_id` here is that it will appear in the API docs (but will be ignored). 

I am trying to think of an alternative solution. Here are some options I came up: 

 - Add the organisation id to the serializer context in the `get_serializer_context` method on the view. The downside of this is that we'd either be hackily passing the feature object around the view class, or making 2 calls to `get_object()`. 
 - Just validate the organisation in the `add_group_owners` method (since we don't actually care about this validation for the `remove_group_owners` anyway. 

I think I prefer option (2) even though it does split up the validation logic. Happy for other suggestions though.

api/users/serializers.py
```diff
@@ -95,7 +95,7 @@ class Meta:
         read_only_fields = ("id",)
 
 
-class MyUserPermissionGroupsSerializer(serializers.ModelSerializer):
+class UserPermissionGroupNameSerializer(serializers.ModelSerializer):
```
matthewelwell: I think perhaps a better name might be `UserPermissionGroupSummarySerializer` ? But I'm ok with what you've got here too if you prefer.

api/features/views.py
zachaysan: Nice catch. Removed.

api/features/serializers.py
zachaysan: I thought the second option was the more preferable as well, and solved it that way. Thanks for the help!

api/features/views.py
```diff
@@ -157,6 +159,41 @@ def get_serializer_context(self):
...
+        data = request.data
+
+        if not UserPermissionGroup.objects.filter(
+            id__in=data["group_ids"],
+            organisation_id=feature.project.organisation_id,
+        ).count() == len(data["group_ids"]):
+            raise serializers.ValidationError("Some groups not found")
+
+        serializer = FeatureGroupOwnerInputSerializer(data=data)
+        serializer.is_valid(raise_exception=True)
```
matthewelwell: We should just switch these around since, otherwise, if someone posts an empty body (or any body missing the `group_ids` key) we would throw a 500 instead of, in the code below, a 400 explaining what the issue is. 

```suggestion
        serializer = FeatureGroupOwnerInputSerializer(data=data)
        serializer.is_valid(raise_exception=True)

        if not UserPermissionGroup.objects.filter(
            id__in=serializer.validated_data["group_ids"],
            organisation_id=feature.project.organisation_id,
        ).count() == len(serializer.validated_data["group_ids"]):
            raise serializers.ValidationError("Some groups not found")
```

matthewelwell: Apologies, the diff looks bad, but all I've done is switched the serializer validation and the ORM validation around (and used `serializer.validated_data` instead of `data["group_ids"]`)

matthewelwell (changes requested):

api/features/views.py
zachaysan: Ok cool, I see what you're going for. I've updated the PR with the changes.

api/features/views.py
```diff
@@ -157,6 +159,41 @@ def get_serializer_context(self):
...
+        feature = self.get_object()
+        data = request.data
+
+        serializer = FeatureGroupOwnerInputSerializer(data=data)
+        serializer.is_valid(raise_exception=True)
+
+        if not UserPermissionGroup.objects.filter(
+            id__in=serializer.validated_data["group_ids"],
+            organisation_id=feature.project.organisation_id,
+        ).count() == len(data["group_ids"]):
```
matthewelwell:
```suggestion
        ).count() == len(serializer.validated_data["group_ids"]):
```

matthewelwell (approved):
One minor nitpick but I've approved anyway.

api/features/views.py
zachaysan: Blast! I've committed the fix. (Not sure what black, etc would do if I use the GH interface)

PR #2918, opened 2023-11-02
feat: audit additional events

### LINKED ISSUE #2797: Support Organisation-Level Audit Log Records
Currently the platform only supports Audit Log records at the project level or below. We should modify this to support events at the Organisation-Level. 

There is a list of items that we want to be able to track, but we should focus on implementing a small number of these as a PoC for organisation-level records. These items are separated into specific sub-issues of this parent issue. 

Suggestion would be to complete [this task](https://github.com/Flagsmith/flagsmith/issues/7571) as a PoC for these events. 

The scope of this work is to extend the `AuditLog` model to add a new `organisation` field, update the existing and/or add new API endpoints for retrieving Audit Log records to retrieve them for an organisation, and to add a new Audit Log view in the dashboard at the Organisation level. The endpoint that lists the audit log records for the organisation should include all audit log records for all projects and environments inside the organisation. It should be possible to filter by project and subsequently environment (see usage page for an example of this filtering functionality). Users must be organisation admin to retrieve the organisation level audit logs.

riceyrice:

## Changes

Implements #2797

- refactors `AuditLog` to associate with organisation without needing project/environment
- audits security-related project, user, group, role, permission changes per organisation
- upgrades and patches django-simple-history to correctly handle indirect m2m signals
- audits user login success/failure and logout signals
- refactors auth mechanisms to send login success/failure and logout signals

Note: `python manage.py populate_history --auto` must be run after migrations in order to create initial history instances, otherwise the first update to any model will log a warning but create no audit log records. 

Note: There are corresponding changes in flagsmith-saml to ensure django signals are sent which must be deployed to correctly audit saml login success/failure.

Note: Does not include front-end changes to view per-organisation audit logs. `AuditLog` views look for organisation directly/indirectly and should be simplified once organisation field is backfilled for existing records.

Tested manually by making changes to memberships and permissions in the front end app or django admin or django shell. Audit log may be inspected per-project or per-environment in the front end but otherwise must be checked using REST API for per-organisation feed or django shell.
~Unit tests will be added once functionality approved. Existing unit tests may fail as I'm currently unable to run them 😞~ 
Login success/failure tested manually in front end using samltest.id and developer GitHib/Google apps.
Unit tests now complete.


api/audit/models.py
```diff
@@ -99,9 +134,12 @@ def get_history_record_model_class(
...
     @hook(BEFORE_CREATE)
-    def add_project(self):
-        if self.environment and self.project is None:
-            self.project = self.environment.project
+    def add_project_organisation(self):
+        """Ensure dependent fields are present"""
+        if self.project_id is None and self.environment:
+            self.project_id = self.environment.project_id
+        if self.organisation_id is None and self.project:
+            self.organisation_id = self.project.organisation_id
```
matthewelwell, line 151: Perhaps there is a reason for this but, from a consistency and (perhaps) functionality point of view, the following makes more sense I think? 

```suggestion
        if self.project_id is None and self.environment_id:
            self.project_id = self.environment.project_id
        if self.organisation_id is None and self.project_id:
            self.organisation_id = self.project.organisation_id
```

api/audit/signals.py
```diff
@@ -18,15 +18,10 @@
 def call_webhooks(sender, instance, **kwargs):
     data = AuditLogSerializer(instance=instance).data
 
-    if not (instance.project or instance.environment):
-        logger.warning("Audit log without project or environment. Not sending webhook.")
+    if not (organisation := instance.organisation):
```
matthewelwell, line 24: This confused me for a minute, but I think this is because, by proxy, any AuditLog record with a project or an environment will return a value for `organisation`. Let me know if I've misunderstood, otherwise you can resolve this comment off the bat.

api/audit/tasks.py
```diff
@@ -48,11 +50,11 @@ def _create_feature_state_audit_log_for_change_request(
         feature_state.change_request.title,
     )
     AuditLog.objects.create(
-        related_object_id=feature_state.id,
-        related_object_type=RelatedObjectType.FEATURE_STATE.name,
-        environment=feature_state.environment,
-        project=feature_state.environment.project,
+        environment=feature_state._get_environment(),
+        project=feature_state._get_project(),
```
matthewelwell: These methods should be 'public' if we're calling them here. Is there a reason they're 'private'?

api/audit/tasks.py
```diff
@@ -150,16 +161,114 @@ def create_segment_priorities_changed_audit_log(
...
-    feature = feature_segments[0].feature
+    # all feature segments should have the same value for environment and feature
+    feature_segment = feature_segments[0]
+    environment = feature_segment._get_environment()
+    feature = feature_segment.feature
 
     AuditLog.objects.create(
-        log=f"Segment overrides re-ordered for feature '{feature.name}'.",
         environment=environment,
-        project_id=environment.project_id,
```
matthewelwell, line 169: Is there a benefit to not storing the project_id here where we know it up front?

api/audit/tasks.py
```diff
@@ -150,16 +161,114 @@ def create_segment_priorities_changed_audit_log(
...
+    credentials: dict,
+    ip_address: str | None = None,
+    codes: list[str] | str | None = None,
+):
+    if not (username := credentials.get("username")):
+        return
+    UserModel = get_user_model()
+    try:
+        # ModelBackend looks up user this way, so we will do the same
+        user = get_user_model()._default_manager.get_by_natural_key(username)
```
matthewelwell:
```suggestion
        user = UserModel._default_manager.get_by_natural_key(username)
```

api/audit/tasks.py
```diff
@@ -150,16 +161,114 @@ def create_segment_priorities_changed_audit_log(
...
+    except UserModel.DoesNotExist:
+        return
+    if not isinstance(user, _AbstractBaseAuditableModel):
+        return
+
+    codes = [codes] if type(codes) is str else codes
```
matthewelwell: I'll probably find this later, but I'm not sure under what circumstances we get `codes` as a string instead of a list. It might be nicer to do this casting there and have this function only accept a list though.

api/core/models.py
```diff
@@ -69,89 +86,395 @@ class _AbstractBaseAuditableModel(models.Model):
...
+    def get_organisations_project_environment(
+        self, delta: ModelDelta | None
+    ) -> tuple[typing.Iterable[Organisation], Project | None, Environment | None]:
+        environment = self._get_environment(delta)
+        project = self._get_project(delta) or (
+            environment.project if environment else None
+        )
+        organisations = (
+            [project.organisation] if project else self._get_organisations(delta)
+        )
```
matthewelwell: It's a little odd that these 2 statements are somewhat opposite. The `_get_entity` method takes precedence for the project, but not for the organisation. Can we add a comment explaining this perhaps?

api/core/models.py
```diff
@@ -69,89 +86,395 @@ class _AbstractBaseAuditableModel(models.Model):
...
+        """Override the human-readable model name for the related object"""
+        if related_object_type := self.get_audit_log_related_object_type(
+            history_instance
+        ):
+            return related_object_type.value
+        else:
+            return self._meta.verbose_name or self.__class__.__name__
+
+    def _get_organisations(self, delta=None) -> typing.Iterable[Organisation] | None:
+        """Return the related organisation for this model."""
```
matthewelwell:
```suggestion
        """Return the related organisations for this model."""
```

api/core/models.py
```diff
@@ -69,89 +86,395 @@ class _AbstractBaseAuditableModel(models.Model):
...
+            )
+
+    def post_m2m_save_or_delete(self, instance, created, cls, attr, **kwargs):
+        # call post_save on parent model
+        instance_attr = cls._meta.get_field(attr).m2m_field_name()
+        self.post_save(getattr(instance, instance_attr), False, **kwargs)
+
+
+def default_get_create_log_message(
+    self: _AbstractBaseAuditableModel, history_instance
```
matthewelwell: I imagine this was moved from a method to a function at some point? 

```suggestion
    instance: _AbstractBaseAuditableModel, history_instance
```

api/core/models.py
```diff
@@ -69,89 +86,395 @@ class _AbstractBaseAuditableModel(models.Model):
...
+    if new_value is False:
+        return "set false"
+    if new_value is None:
+        return "set null"
+
+    return "updated"
+
+
+def default_get_update_log_message(
+    self: _AbstractBaseAuditableModel, history_instance, delta: ModelDelta
```
matthewelwell:
```suggestion
    instance: _AbstractBaseAuditableModel, history_instance, delta: ModelDelta
```

api/core/models.py
```diff
@@ -69,89 +86,395 @@ class _AbstractBaseAuditableModel(models.Model):
...
+
+
 def abstract_base_auditable_model_factory(
-    historical_records_excluded_fields: typing.List[str] = None,
+    related_object_type: RelatedObjectType,
+    unaudited_fields: typing.Sequence[str] | None = None,
+    audited_m2m_fields: typing.Sequence[str] | None = None,
+    audit_create=False,
+    audit_update=False,
+    audit_delete=False,
```
matthewelwell:
```suggestion
    audit_create: bool = False,
    audit_update: bool = False,
    audit_delete: bool = False,
```

api/core/models.py
```diff
@@ -69,89 +86,395 @@ class _AbstractBaseAuditableModel(models.Model):
...
         class Meta:
             abstract = True
 
+    Base.related_object_type = related_object_type
+    if audit_create:
+        Base.get_create_log_message = default_get_create_log_message
+    if audit_update:
+        Base.get_update_log_message = default_get_update_log_message
+    if audit_delete:
+        Base.get_delete_log_message = default_get_delete_log_message
```
matthewelwell: What's the benefit of doing it this way versus defining the defaults on the base class itself?

api/core/models.py
```diff
@@ -69,89 +86,395 @@ class _AbstractBaseAuditableModel(models.Model):
...
+def register_auditable_model(
+    model: type[models.Model],
+    app: str,
+    related_object_type: RelatedObjectType,
+    unaudited_fields: typing.Sequence[str] | None = None,
+    audited_m2m_fields: typing.Sequence[str] | None = None,
+    *,
+    audit_create=False,
+    audit_update=False,
+    audit_delete=False,
```
matthewelwell:
```suggestion
    audit_create: bool = False,
    audit_update: bool = False,
    audit_delete: bool = False,
```

api/projects/models.py
```diff
@@ -26,16 +29,46 @@
...
+    "hide_disabled_flags",
+    "enable_dynamo_db",
+    "prevent_flag_defaults",
+    "enable_realtime_updates",
+    "only_allow_lower_case_feature_names",
+    "feature_name_regex",
+    "max_segments_allowed",
+    "max_features_allowed",
+    "max_segment_overrides_allowed",
+)
```
matthewelwell:
```suggestion
UNAUDITED_PROJECT_FIELDS = (
    "enable_dynamo_db",
    "enable_realtime_updates",
    "max_segments_allowed",
    "max_features_allowed",
    "max_segment_overrides_allowed",
)
```

I've removed a few here that are user controlled - maybe there's another reason that they shouldn't be audited though?

api/projects/models.py
```diff
@@ -26,16 +29,46 @@
...
+
+
+class Project(
+    LifecycleModelMixin,
+    SoftDeleteExportableModel,
+    abstract_base_auditable_model_factory(
+        RelatedObjectType.PROJECT,
+        UNAUDITED_PROJECT_FIELDS,
+        audit_create=True,
+        audit_delete=True,
```
matthewelwell: What's the reasoning behind not auditing update here? (I don't think auditing the project at all was part of the scope so if that's the answer, that's cool, just checking)

api/users/models.py
```diff
@@ -91,12 +98,33 @@ def get_by_natural_key(self, email):
         return self.get(email__iexact=email)
 
 
-class FFAdminUser(LifecycleModel, AbstractUser):
+# NOTE date_joined cannot be excluded because there is also a date_joined field on
+# the organisations through model, and currently simple history would create a
+# historical through model without the field but try to store the value anyway
```
matthewelwell, line 104: Is there a path to resolve this?

matthewelwell (commented):
@riceyrice on the whole, this looks great! I've added some (mostly) minor comments, but it seems to achieve what we're looking for and the code looks good.

api/audit/models.py
riceyrice: The type checker is happier with this because it doesn't know that if `self.environment_id` is truthy then `self.environment` is not `None`.
Functionality-wise I'd check e.g. `self.environment_id` instead of `self.environment` in order to avoid hitting the database to load the `Environment` object, but the next line will access `self.environment` anyway, so I don't think this will affect functionality either way.

api/audit/signals.py
riceyrice: Yes, in time `AuditLog` is expected to have an `organisation` field. Until we add a data migration to backfill that for existing records, this is a property that checks for an organisation value and if not falls back to project (which is always set if environment was set):
```
    @property
    def organisation(self) -> Organisation | None:
        return self._organisation or (
            self.project.organisation if self.project else None
        )
```

api/audit/tasks.py
riceyrice: They are only private because they were already private - I guess they were intended for use by the public `get_environment_and_project` (now `get_organisations_project_environment`). 
Previously this code accessed member variables directly, which seemed to be against DRY and introduce the possibility that it could pick different values to those used for audit logs that go via `_get_environment`.
I guess there are some options here:

- use `get_organisations_project_environment` public interface (like `create_audit_log_from_historical_record` in the same file does)
- make `_get_organisations/project/environment` into public interface

api/audit/tasks.py
riceyrice: `AuditLog` already backfills `project_id` from `environment.project_id` so it was a DRY thing. Happy either way.

api/audit/tasks.py
riceyrice: Yeah that's a typo. Will put in next commit.

api/audit/tasks.py
riceyrice: As far as I can see it's a bug in DRF - for some exceptions `get_codes()` returns `str` and others it's `list[str]` 😞. I was 50/50 where to put the workaround, I'll push it back to the signal handler rather than the task.

api/core/models.py
riceyrice: Hmm, tbh I think that's a mistake! Possibly a hangover from some `[]` versus `None` faff I went through. Will have a look.

api/core/models.py
riceyrice: Kind of the reverse. They are intended to be added as methods. I initially wrote them using `instance` as the variable name but then the type checker complains when you assign them as methods. I can change them to `instance` if you prefer but see the base class conversation first.

api/users/models.py
riceyrice: Raise a proper ticket (among others arising from this code) with django-simple-history I guess. Currently `create_historical_record_m2ms` uses `through_model._meta.fields` to decide what values to store, and doesn't take account of the fields that were excluded when the model was created. A proper fix should probably use `tracked_fields` in the same way as `create_historical_record` should, but I couldn't get it working.

api/projects/models.py
riceyrice: The spec was to audit project creation and deletion. Unlike other entities, it did not specify auditing updates. Possibly it should? See earlier discussion about fields to include/exclude.

api/projects/models.py
riceyrice: This is somewhat redundant as we don't currently create audit logs for project updates - but equally I assumed we don't want a historical models for every minor change. My thinking was that the auditing being added in this PR is for security purposes, and the only security-related field I could think of was its name and containing organisation, hence the rest are excluded. Happy for this list to be whatever you want.

api/core/models.py
riceyrice: This dance was to avoid changing the default behaviour of `abstract_base_auditable_model_factory` as many things already call this expecting `_AbstractBaseAuditableModel` and its implementations that return `None`, logging nothing by default. 
All the things I've added want to log create/update/delete by default (except project but maybe it should).
I will see if I can do it more elegantly using mixins.

riceyrice: I've replaced this with a mixin implementation - it's a bit verbose but hopefully clearer. tbh if create/update/delete logging could be on/off rather than three toggles, that could be a fair bit simpler. Project audit being create/delete only is the only real reason for it and maybe that's not a real requirement.

api/audit/tasks.py
matthewelwell: I guess the downside of option 1 here is that we don't need the organisations (I don't think?). My preference in that case would be option 2 but if, for whatever reason, we do want the organisation here too, let's go with option 1.

api/core/models.py
matthewelwell: Yeah, I don't think that's an intentional requirement. I think having project update would also be beneficial.

api/projects/models.py
matthewelwell: Yep, let's go with what I've suggested here as per my other comment regarding project update.

api/users/models.py
matthewelwell: Ok, no worries. Perhaps let's go through the code once we're done and work out what we need to do with the dsh issues / PRs.

api/core/models.py
matthewelwell: I'm not 100% sure if this conversation is now redundant or not, but I'm happy to leave these as `self`. Maybe we should add a comment to explain why though.

api/audit/tasks.py
riceyrice: We don't need organisation so went with option 2

matthewelwell:
Finally accepting defeat and closing this PR. While we intend to resurrect the functionality to audit events at the organisation level, this PR is too old to save 😢 we will likely take a more incremental approach and start by adding a single 'org-level' audit log event (or set of events for a specific entity perhaps), and then iterating on that to add the remaining events that we attempted to add here.

PR #2919, opened 2023-11-02
feat: Add permission for manage segments overrides

zachaysan:

## Changes

Segment overrides allows release of a specific feature to broader set of users, with this change Flagsmith users are able to have a specialized permission for managing segment overrides as an environment role. See [this](https://github.com/orgs/Flagsmith/projects/2/views/11?pane=issue&itemId=14813319) for more information.

Most of the work in this PR involved trying to understand the surrounding approach and replace permission checks as needed.

I also added `django_extensions` for local development ergonomics as well.

I relied on the existing tests, following the convention of relying on `admin_master_api_key_client`, but this means that we're not really testing the new permission since it's an admin api key, not a user with the permission. 

~I would like to test this locally with the interface to verify that the permission checks work properly, but I'm opening the PR first to get feedback before setting that all up.~

I have fully tested this code through a locally running webserver and caught and debugged a few small issues. This PR must wait for merge until the frontend is ready.


api/environments/permissions/migrations/0008_add_manage_segment_overrides_permission.py
```diff
@@ -0,0 +1,55 @@
...
+    )
+    UserPermissionGroupEnvironmentPermission = apps.get_model(
+        "environment_permissions",
+        "UserPermissionGroupEnvironmentPermission",
+    )
+    manage_segment_overrides_permission, _ = PermissionModel.objects.get_or_create(
+        key=MANAGE_SEGMENT_OVERRIDES,
+        description="Permission to manage segment overrides in the given environment",
+        type=ENVIRONMENT_PERMISSION_TYPE,
+    )
```
matthewelwell, line 23: I think that, since you changed the logic in the previous migration, we don't technically need the `get_or_create` here? We can just use `create`, right?

api/features/permissions.py
```diff
@@ -164,8 +165,7 @@ def has_permission(self, request, view):
             Environment, api_key=view.kwargs["environment_api_key"]
         )
 
-        # TODO: create dedicated permission for creating segment overrides
```
matthewelwell, line 169: Nice! Deleting a very old todo!

matthewelwell (commented):
I would love to see a couple more tests here - perhaps just some unit test coverage of the permission class, or an API test with a user that has the correct permission.

api/environments/permissions/migrations/0008_add_manage_segment_overrides_permission.py
zachaysan: Yeah, but as discussed, it's easier to keep it this way for local developers.

api/features/permissions.py
zachaysan: I love deleting TODOs :)

api/features/feature_segments/tests/test_views.py
```diff
@@ -258,6 +344,35 @@ def test_get_feature_segment_by_uuid(
...
+):
+    # Given
+    url = reverse(
+        "api-v1:features:feature-segment-get-by-uuid", args=[feature_segment.uuid]
+    )
+    with_environment_permissions([UPDATE_FEATURE_STATE])
+    user_project_permission = UserProjectPermission.objects.create(
+        user=staff_user, project=project
+    )
+    user_project_permission.add_permission(VIEW_PROJECT)
```
matthewelwell: We could create a `with_project_permissions` fixture here too, right?

api/features/tests/test_views.py
```diff
@@ -787,6 +789,65 @@ def test_get_feature_evaluation_data(project, feature, environment, mocker, clie
...
 
 
+def test_create_segment_override_forbidden(feature, segment, environment, api_client):
+    # Given
+    url = reverse(
+        "api-v1:environments:create-segment-override",
+        args=[environment.api_key, feature.id],
+    )
+    user = FFAdminUser.objects.create(email="jim@example.com")
+    api_client.force_authenticate(user)
```
matthewelwell: This should just use the `staff_client` now, right?

matthewelwell (commented):
Love the test coverage here 👌 I've added a couple of minor comments but looks good for the most part.

api/features/feature_segments/tests/test_views.py
zachaysan: Yep doing that now :)

api/features/tests/test_views.py
zachaysan: Yep! Old code, nice catch!

api/conftest.py
```diff
@@ -199,6 +201,29 @@ def _with_environment_permissions(
     return _with_environment_permissions
 
 
+@pytest.fixture()
+def with_project_permissions(
+    project: Project, staff_user: FFAdminUser
+) -> typing.Callable:
+    """
+    Add project permissions to the staff_user fixture.
+    Defaults to associating to the environment fixture.
```
matthewelwell:
```suggestion
    Defaults to associating to the project fixture.
```

matthewelwell (commented):
One minor tweak to a copypasta error. It would also be good to add some typing to the test fixtures 🙏

api/conftest.py
zachaysan: Thank you!

matthewelwell (approved):

PR #2922, opened 2023-11-03
docs: network model overview

dabeeeenster:


docs/docs/clients/overview.md
```diff
@@ -120,34 +121,76 @@ running in `Local Evaluation` mode.
...
-### Local Evaluation Mode
+#### Client-side
 
-- Identities are _not_ sent to the API and so are not persisted in the datastore.
-- Because Local mode does not connect to the datastore for each Flag request, it is not able to read the Trait data of
-  Identities from the API. This means that you have to provide the full complement of Traits when requesting the Flags
-  for a particular Identity. Our SDKs all provide relevant methods to achieve this.
-- [Identity overrides](../basic-features/managing-identities#identity-overrides) do not operate at all.
-- Analytics-based Integrations do not run.
+- Flagsmith SDKs will store all the Flags for an Environment in local memory. Flag evaluations within the SDK are then a
```
matthewelwell: This isn't very clear - it doesn't _magically_ store all the flags in local memory. It does it following the first network request. I don't think it _always_ does it either - there are configuration parameters for each of the clients to configure the cache etc.

docs/docs/clients/overview.md
```diff
@@ -120,34 +121,76 @@ running in `Local Evaluation` mode.
...
+- If an entirely new Identity is provided, the SDK will make a new call to the Flagsmith API to receive new Flags.
+- Other than the above two points, the SDK will not make further network requests to the Flagsmith API. If you wish you
+  can manually trigger a network call in code and refresh Flags locally.
 
-The benefit of running in Local Evaluation mode is that you can process flag evaluations much more efficiently as they
-are all computed locally.
+#### Server-side
 
-:::tip
+- Flagsmith SDKs do not store Flags in local memory. Every Flag evaluation in your code will trigger a network request.
```
matthewelwell:
```suggestion
- Flagsmith server-side SDKs do not store Flags in local memory. Every Flag evaluation in your code will trigger a network request.
```

docs/docs/clients/overview.md
```diff
@@ -120,34 +121,76 @@ running in `Local Evaluation` mode.
...
+3. Cache these flags locally on the device.
+4. Upon subsequent application launches, immediately read the flags from the cached store from step 3 and use those as
+   your application Flag values.
+5. In the background make a request to Flagsmith to refresh their Flags.
+6. Update both your application and the local cache with these fresh flags.
+
+### Things to note
 
-## Local Evaluation Runtime Considerations
+- Flagsmith always returns **all** Environment Flags during API calls. In fact, there is no way to request the state of
```
matthewelwell, line 185: >In fact, there is no way to request the state of a single flag via our API.

I don't think this is _technically_ true - it is possible via the API we just don't advertise it (because it's poorly implemented). Happy to keep this here though, just clarifying.

docs/docs/clients/overview.md
dabeeeenster: OK tweaked. Its just a general overview but I think this info is important at this stage in the docs.

docs/docs/clients/overview.md
```diff
@@ -120,34 +121,84 @@ running in `Local Evaluation` mode.
...
+- No further network calls take place for 60 seconds.
+- After 60 seconds have elapsed, the SDK will refresh the JSON Environment document with a network call to the Flagsmith
+  API.
+
+### Client-side SDK approaches
+
+Since Client-side SDKs will generally be associated with a single Identity (the person who owns the client device!), a
+common pattern for networking implementation is:
+
+1. The user launches your application for the first time. Since your application does not know how they are, a call to
```
matthewelwell:
```suggestion
1. The user launches your application for the first time. Since your application does not know who they are, a call to
```

matthewelwell (dismissed):

matthewelwell (approved):

PR #2923, opened 2023-11-03
feat: add foundation for LDAP in core repository

matthewelwell:
## Changes

Adds the foundation for LDAP functionality in the core repository. 

Ideally, the `ldap_dn` attribute would be part of a separate model contained in the LDAP repository but, to reduce friction in moving from the EE image to the private-cloud image in the future, I've added the attribute directly to the existing model as in the EE codebase. 

This PR just adds the settings and model attribute, no logic to be tested.


api/app/settings/common.py
```diff
@@ -942,3 +942,81 @@
...
     "FLAGSMITH_ON_FLAGSMITH_SERVER_API_URL", default=FLAGSMITH_ON_FLAGSMITH_API_URL
 )
+
+# LDAP setting
+LDAP_INSTALLED = importlib.util.find_spec("ldap")
+if LDAP_INSTALLED:
+    # The URL of the LDAP server.
+    LDAP_AUTH_URL = env.str("LDAP_AUTH_URL", None)
+    if LDAP_AUTH_URL:
+        AUTHENTICATION_BACKENDS.insert(0, "django_python3_ldap.auth.LDAPBackend")
```
zachaysan, line 952: Should this throw an exception if a URL is not set?

zachaysan (approved):
I'm a little confused as to what should happen if some of the settings are left unset. Presumably the LDAP functionality would just fail somewhere, no? If it would, would it be better to ensure that the settings are present if the top LDAP configuration is set?

Since you've probably got a good answer to this question I'm approving the PR since I didn't spot any code errors. Just curious as to the approach.

api/app/settings/common.py
gagantrivedi: I think it's because we will include LDAP in every private cloud image... but it's up to the client to enable it (by setting `LDAP_AUTH_URL`). Is that correct @matthewelwell

gagantrivedi (approved):

api/app/settings/common.py
```diff
@@ -942,3 +942,81 @@
 FLAGSMITH_ON_FLAGSMITH_SERVER_API_URL = env(
     "FLAGSMITH_ON_FLAGSMITH_SERVER_API_URL", default=FLAGSMITH_ON_FLAGSMITH_API_URL
 )
+
+# LDAP setting
+LDAP_INSTALLED = importlib.util.find_spec("ldap")
```
khvn26: I think it's better to search for our own module here.

matthewelwell: I'm not sure what you mean here - `ldap` is our own module.

khvn26: python-ldap has the same module name so I got confused, apologies. Maybe we should rename our module to flagmsith-ldap.

matthewelwell: Yeah, that's probably a good shout! I'll do that.

api/app/settings/common.py
matthewelwell: Yes, that's correct.

api/app/settings/common.py
matthewelwell: This is done [here](https://github.com/Flagsmith/flagsmith/pull/2923/commits/87ab03393bc3b053134d318211e6f36c5f447058)

khvn26 (approved):

PR #2928, opened 2023-11-06
feat: Create staff fixture

zachaysan:

## Changes

Add a new fixture to transition to testing more routes by default at a lower level of access to ensure staff members are:

1. Able to do stuff they're supposed to do.
2. Not able to do stuff they're not supposed to do.

Essentially, start by default by using the staff fixture since the resulting tests are more robust since admin users are automatically able to sail through most permissions classes.

By using the same code in [this PR](https://github.com/Flagsmith/flagsmith/pull/2919/files)

### DIFF
--- a/api/conftest.py
+++ b/api/conftest.py
@@ -80,9 +80,9 @@ def test_user_client(api_client, test_user):
 def staff_user(django_user_model):
     """
     A non-admin user fixture.
-    To add to an environment with permissions use:
-    uep = UserEnvironmentPermission.objects.create ...
-    uep.permissions.add ...
+    To add to an environment with permissions use the fixture
+    with_environment_permissions.
+
     This fixture is attached to the organisation fixture.
     """
     return django_user_model.objects.create(email="staff@example.com")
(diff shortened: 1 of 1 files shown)


khvn26 (commented):
Looks good, two concerns though:

1. I'd think on the naming some more since, AFAIK, we don't use the word "staff" anywhere. `user_user` is awkward of course, maybe `user_role_user` is better.
2. I'm reluctant to merge this without any tests actually using the fixture to avoid adding dead code.

zachaysan:
> 1. I'd think on the naming some more since, AFAIK, we don't use the word "staff" anywhere. `user_user` is awkward of course, maybe `user_role_user` is better.

I bounced it off Matt and he's with me on `staff` being a sorta concise, readable word. I think it's better than the alternatives.

> 2. I'm reluctant to merge this without any tests actually using the fixture to avoid adding dead code.

I already have some tests that use this fixture. They're just in PRs that may be around for a while and I want this new fixture to use now. Plus we should all be defaulting to using this fixture anyway, since the admin client side steps many of the permissions checks.

khvn26 (approved):

matthewelwell (approved):

PR #2931, opened 2023-11-06
fix: Check that feature owners are able to view the project of a feature

zachaysan:

## Changes

Before users from anywhere could be added to a feature as an owner, even if they were outside the organisation. Now, they must be able to view the project of the feature in order to be added as feature owners. This mirrors much of the existing logic around this part of the codebase, but I could be convinced to make it differently. For example, users as members of the organisation could make sense as well. 

Updated the existing code, watched the formerly passing test fail, then updated the test and added a new one to make sure users can't be added as owners if they're not affiliated with the feature.


api/tests/unit/features/test_unit_features_views.py
```diff
@@ -529,9 +530,11 @@ def test_should_create_tags_when_feature_created(client, project, tag_one, tag_t
     "client",
     [lazy_fixture("admin_master_api_key_client"), lazy_fixture("admin_client")],
 )
-def test_add_owners_adds_owner(client, project):
+def test_add_owners_adds_owner_when_users_not_connected(client, project):
```
matthewelwell: I think this is the negative case, right? So the test name should be more like: 

```suggestion
def test_add_owners_fails_if_user_not_found(client, project):
```

api/tests/unit/features/test_unit_features_views.py
```diff
@@ -541,24 +544,47 @@ def test_add_owners_adds_owner(client, project):
     data = {"user_ids": [user_1.id, user_2.id]}
 
     # When
-    json_response = client.post(
-        url, data=json.dumps(data), content_type="application/json"
-    ).json()
+    response = client.post(url, data=json.dumps(data), content_type="application/json")
+    # Then
+    response.status_code == status.HTTP_400_BAD_REQUEST
+    response.data == ["Some users not found"]
```
matthewelwell: Let's also verify that the users were not added to the feature. Something like 

```python
assert feature.owners.filter(id__in=[user_1.id, user_2.id]).count() == 0
```

matthewelwell (changes requested):

api/tests/unit/features/test_unit_features_views.py
zachaysan: Good ideas, done!

matthewelwell (approved):

PR #2932, opened 2023-11-06
feat: Call webhooks async and add backoff to webhooks

tushar5526:
Related to #1654 

## Changes

Use task processor to exponentially backoff webhooks and make webhook calling process async. Earlier it was a sync for loop. 

Tested it locally, by running the task processor and running faulty and health webhooks. Also, updated test to support the new changes.


api/webhooks/webhooks.py
```diff
@@ -43,97 +48,199 @@ class WebhookType(enum.Enum):
...
+    :param retries: The number of times to retry the webhook in case of failure (int, default is 3).
+    """
+
     if settings.DISABLE_WEBHOOKS:
         return
-
+    try:
+        environment = Environment.objects.get(id=environment_id)
+    except Environment.DoesNotExist:
+        return
```
khvn26, line 85: Do we want to silence this? @matthewelwell

api/webhooks/webhooks.py
```diff
@@ -43,97 +48,199 @@ class WebhookType(enum.Enum):
...
         signature = sign_payload(json_data, key=webhook.secret)
         headers.update({FLAGSMITH_SIGNATURE_HEADER: signature})
-
     return requests.post(str(webhook.url), data=json_data, headers=headers, timeout=10)
 
 
-def _call_webhook_email_on_error(
-    webhook: WebhookModels, data: typing.Mapping, webhook_type: WebhookType
+@register_task_handler()
+def call_webhook_with_failure_mail_after_retires(
```
khvn26:
```suggestion
def call_webhook_with_failure_mail_after_retries(
```

api/webhooks/webhooks.py
```diff
@@ -43,97 +48,199 @@ class WebhookType(enum.Enum):
...
+                    webhook_type,
+                    send_failure_mail,
+                    max_retries,
+                    try_count + 1,
+                ),
+            )
         return
 
-    if res.status_code != 200:
+    if send_failure_mail and res.status_code != 200:
```
khvn26: I'd like to use this opportunity to remove this `if` completely in favour of using `res.raise_for_status()` inside the try/except block.

Note that without it, HTTP errors will fail on the first try, which IMO should either not happen or be well documented.

api/webhooks/webhooks.py
```diff
@@ -43,97 +48,199 @@ class WebhookType(enum.Enum):
...
+                send_failure_email(
+                    webhook,
+                    data,
+                    webhook_type,
+                    f"N/A ({exc.__class__.__name__})",
+                )
+        else:
+            call_webhook_with_failure_mail_after_retires.delay(
+                delay_until=(
+                    timezone.now() + timezone.timedelta(seconds=2**try_count)
```
khvn26: Do you think the `2` here should be configurable or at least be a constant?

khvn26 (changes requested):

api/webhooks/webhooks.py
tushar5526: Yes there was a discussion in the previous PR on this 
 -> https://github.com/Flagsmith/flagsmith/pull/2559#issuecomment-1685983292

api/webhooks/webhooks.py
tushar5526: I am of the same opinion. Webhooks should backoff in-case of a bad response.

api/webhooks/webhooks.py
tushar5526: This should be pickable from the environment, there could be a use case where some user doesn't want the backoff to be 2^x. Should I move this to `constants` file and pick the constant's value from the settings file?

tushar5526:
@khvn26 @matthewelwell 
In order to avoid any herding in large projects with a lot of webhooks configured, should I also add in jitter?

Earlier this was synchronous, but now webhooks are scheduled at almost the same time with the same constant backoffs.

api/webhooks/webhooks.py
khvn26: Yes, please do that.

tushar5526: Done. Also moved `retries` to settings. It was defaulted to 3 earlier.

matthewelwell:
> @khvn26 @matthewelwell In order to avoid any herding in large projects with a lot of webhooks configured, should I also add in jitter?
> 
> Earlier this was synchronous, but now webhooks are scheduled at almost the same time with the same constant backoffs.

I do like that idea, but I don't think it's necessarily super important. I can't see a scenario where we would have so many webhooks being retried that we would have a herding problem. That being said, if it's not too tricky, it might be worth having a go, yep.

api/webhooks/constants.py
```diff
@@ -1 +1,5 @@
+from django.conf import settings
+
 WEBHOOK_DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
+WEBHOOK_BACKOFF_BASE = settings.WEBHOOK_BACKOFF_BASE
+WEBHOOK_BACKOFF_RETRIES = settings.WEBHOOK_BACKOFF_RETRIES
```
matthewelwell: This feels unnecessary, is there a reason we shouldn't just use the `settings` variables directly?

api/features/tests/test_tasks.py
```diff
@@ -11,8 +9,7 @@
 
 
 @pytest.mark.django_db
-@mock.patch("features.tasks.Thread")
-def test_trigger_feature_state_change_webhooks(MockThread):
+def test_trigger_feature_state_change_webhooks(mocker):
```
matthewelwell:
```suggestion
def test_trigger_feature_state_change_webhooks(mocker: MockerFixture):
```

api/features/tests/test_tasks.py
```diff
@@ -30,16 +27,22 @@ def test_trigger_feature_state_change_webhooks(MockThread):
...
 
     # Then
-    call_list = MockThread.call_args_list
+    environment_webhook_call_args = mock_call_environment_webhooks.call_args
+    organisation_webhook_call_args = mock_call_organisation_webhooks.call_args
 
-    environment_webhook_call_args = call_list[0]
-    organisation_webhook_call_args = call_list[1]
+    assert environment_webhook_call_args[1]["args"][0] == environment.id
+    assert organisation_webhook_call_args[1]["args"][0] == organisation.id
```
matthewelwell: This confused me for a while until I ran the test myself. I think it's because the kwarg we pass to the `delay` method is called `args` (maybe that was a bad decision). I wonder if we can improve this with some better naming of vars (and using the attributes on the `Call` class via their names too) here though, something like: 

```python
environment_webhook_call_args = (
    mock_call_environment_webhooks.call_args.kwargs["args"]
)
organisation_webhook_call_args = (
    mock_call_organisation_webhooks.call_args.kwargs["args"]
)
```

If you agree, we should also do this in similar tests below.

api/webhooks/webhooks.py
```diff
@@ -43,97 +49,205 @@ class WebhookType(enum.Enum):
...
     _call_webhooks(
         organisation.webhooks.filter(enabled=True),
         data,
         event_type,
         WebhookType.ORGANISATION,
+        retries,
     )
 
 
 def call_integration_webhook(config, data):
```
matthewelwell: 2 points here: 

 1. While we're here, can we add typing to this function too? 
 2. Is there a reason that we're not implementing the back off logic to this webhook implementation?

api/webhooks/webhooks.py
```diff
@@ -43,97 +49,205 @@ class WebhookType(enum.Enum):
...
-    return _call_webhook(config, data)
```
matthewelwell, line 85: Based on [this](https://github.com/Flagsmith/flagsmith/pull/2932/files#diff-86f132934c5b75bd30cb37ce30cc88b07b738d2b09d772cbeea42e2186809cbaR140), this looks wrong to me. The function still seems to expect us to send an ORM model, not an id.

matthewelwell (commented):
I've added a few more comments here @tushar5526. Let me know if anythingis easier to discuss directly and we can perhaps jump on a short call.

api/webhooks/webhooks.py
tushar5526: 1. Sure, I will add it. 
2. I had second thoughts on synchronous backoffs on the main thread. If that's fine, I can add in backoff for 3 retries and max backoff time to `2s` maybe?

api/webhooks/webhooks.py
tushar5526: Yes, you are right. I missed changing this back.

api/webhooks/constants.py
tushar5526: Yes, this seems unnecessary, thanks for pointing it out. Thank you, I will revert this.

api/webhooks/webhooks.py
matthewelwell: Regarding point (2) here it's actually not run in the main thread, it's run in a daemon thread dispatched to by the main thread (and never checked afterwards). The logic here is quite convoluted and I'm not exactly happy with it, but it's served us well up to this point. 

If you follow the usages of this `call_integration_webhook`, you'll find that it's called from `integrations.webhook.WebhookWrapper` which inherits from `integrations.common.wrapper.AbstractBaseIdentityIntegrationWrapper` that has the `identify_user_async` method exposed (and is the only method actually called by the main thread). 

I'd love to refactor all of this to use task handlers instead of these daemon threads directly (the initial refactor would likely just be to use `run_in_thread`) but that's for another PR.  

Maybe _all_ of this should be handled in a separate PR to be honest and we'll leave the integrations webhook to only try once for now. I'll leave it up to you here.

api/features/tests/test_tasks.py
tushar5526: Agreed, did the changes and it is looking much more readable now. Thanks!

api/webhooks/webhooks.py
tushar5526: I have added backoff to it, using the `backoff` library. I can't use a task processor here, because the return values of `_call_webhooks` are needed in the parent calling functions.

matthewelwell (approved):
This looks good to me now! Thanks so much for all your hard work on this @tushar5526.

api/webhooks/webhooks.py
```diff
@@ -102,47 +154,116 @@ def _call_webhook(
...
         )
+        res.raise_for_status()
+    except requests.exceptions.RequestException as exc:
+        if try_count == max_retries:
+            if send_failure_mail:
+                send_failure_email(
+                    webhook,
+                    data,
+                    webhook_type,
+                    f"N/A ({exc.__class__.__name__})",
```
khvn26: Error display heuristics were lost: 

```suggestion
                    f"{f'HTTP {exc.response.status_code}' if exc.response else 'N/A'} ({exc.__class__.__name__})",
```

tushar5526:
> This looks good to me now! Thanks so much for all your hard work on this @tushar5526.

Learned lots of stuff :D, thanks for the high-quality reviews @matthewelwell @khvn26 !

khvn26 (approved):

tushar5526:
Resolved the conflicts, good to go!

tushar5526:
Just checking if there is anything else in this - that needs a second look ?

dabeeeenster:
cc @khvn26 !

matthewelwell:
@tushar5526 sorry for the delay in getting this merged, with the holidays and trying to manage the size of releases it's been waiting on the priorities a little bit. I'm merging it now. Thanks again for the hard work on this one!

Issue #1667, opened 2022-11-15
Epic: V2 API


matthewelwell:
This is expected to be a long running ticket where we can list breaking changes that we want to include in a V2 API. As part of this work, we should plan the new API interface around the key journeys that customers need to use our REST API for to simplify admin integrations. 

- Add consistent API interface across all Edge endpoints. These should all respond with a consistent schema that follows a well defined framework, e.g. https://jsonapi.org/. 
  - specifically, this will allow us to remove the 'updated at' value from the header of the `/flags/` endpoint and into the response body
- Look into fixing the typing issues. All dynamically typed fields should be removed and replaced with a combination of a string field and a provided data type. 
- Replace all references to 'unicode' with 'string'
- Create endpoints to manage state of a feature for an environment, identity or segment without needing to make multiple API requests
- Prevent updates to FeatureState objects - we should only allow new versions to be created
- Remove endpoints reliant on influx data in favour of endpoints that return data from influx or postgres for analytics purposes
- Update authentication such that it includes context of the organisation for each request
- Remove integration configuration objects from environment-document endpoint
- Remove dynatrace configuration from the engine / edge api entirely
- Remove need for FeatureSegment
- Remove `"description"` field from environment list endpoint (see discussion [here](https://github.com/Flagsmith/flagsmith/pull/2947/files#r1388331387) for reference)

gagantrivedi:
Looks like https://pypi.org/project/djoser/#history is inactive we should consider this in V2 phase

kyle-ssg:
## Displaying features

One of the main things that would make writing a frontend more feasible is to reduce the number of API calls needed to display a feature and its information for a particular environment. 

Related to https://github.com/Flagsmith/flagsmith/issues/1619:
![image](https://user-images.githubusercontent.com/8608314/204320489-7388cf1b-b435-47a4-bee7-3d0e815b6d0e.png)

Given that features are paged, if we wanted to show the number of segment overrides, the endpoints we'd be hitting to show a list of features would be as follows:

- Get every page of featurestates for the environment
```
https://api.flagsmith.com/api/v1/environments/4vfqhypYjcPoGGu8ByrBaj/featurestates/?page_size=20
```

- Get the paged features for the project 
```
https://api.flagsmith.com/api/v1/projects/12/features/?page=1&page_size=200&is_archived=false&sort_field=name&sort_direction=ASC
```

- For each feature in the list, get the "feature-segments" 
```
[...]

PR #2947, opened 2023-11-09
fix: Fix evironment metadata N+1 for environments list

### LINKED ISSUE #2900: N+1 metadata query for `/api/v1/environments/?project=<id>`
Ref: https://flagsmith.sentry.io/issues/4502384230/?project=5544478&referrer=issue-stream&statsPeriod=90d&stream_index=4

zachaysan:

## Changes

Isolated the cause of an N+1 query and implemented a fix without forcing it on every caller. We may want to consider reviewing the callers and seeing which of them load metadata as well, since the bulk of the performance enhancement is already complete.

One new test. During the testing process I also manually inspecting the types of queries that were generated to ensure that they matched those reported by Sentry.


api/environments/tests/test_views.py
```diff
@@ -536,6 +539,53 @@ def test_create_environment_without_required_metadata_returns_400(
...
+    environment_content_type: ContentType,
+) -> None:
+    # Given
+    with_environment_permissions([VIEW_ENVIRONMENT])
+
+    url = reverse("api-v1:environments:environment-list")
+    data = {"project": project.id}
+
+    # When
+    expected_query_count = 7
```
khvn26: nit: this line belongs to the `# Given` section

api/environments/tests/test_views.py
```diff
@@ -536,6 +539,53 @@ def test_create_environment_without_required_metadata_returns_400(
     assert "Missing required metadata field" in response.json()["metadata"][0]
 
 
+def test_view_environment_with_staff_track_query_count(
```
khvn26: I'd prefer a slightly more structured test name here:

```suggestion
def test_view_environment_with_staff__query_count_is_expected(
```

api/environments/tests/test_views.py
```diff
@@ -536,6 +539,53 @@ def test_create_environment_without_required_metadata_returns_400(
...
+    staff_client: FFAdminUser,
+    environment: Environment,
+    with_environment_permissions: Callable,
```
khvn26: nit: we should type this properly. Looks like it should be

```suggestion
    with_environment_permissions: Callable[[list[str], int], None],
```

I can see the fixture definition is also not typed correctly.

api/environments/tests/test_views.py
```diff
@@ -536,6 +539,53 @@ def test_create_environment_without_required_metadata_returns_400(
...
+    project: Project,
+    django_assert_num_queries: Callable,
```
khvn26:
```suggestion
    django_assert_num_queries: Callable[[int], None],
```

api/permissions/permission_service.py
```diff
@@ -130,7 +131,16 @@ def get_permitted_environments_for_user(
...
+    queryset = Environment.objects.filter(filter_)
+    if prefetch_metadata:
+        queryset = queryset.prefetch_related("metadata")
+
+    # Description is defered due to Oracle support where a
+    # query can't have a where clause if description is in
+    # the select parameters. This leads to an N+1 query for
+    # lists of environments when description is included, as
+    # each environment object re-queries the DB seperately.
+    return queryset.distinct().defer("description")
```
khvn26, line 143: Maybe we could branch out the logic here based on `connection.vendor`?

zachaysan: I was thinking that too, though it does open up a can of worms with respect to complexity.

api/environments/tests/test_views.py
```diff
@@ -536,6 +539,53 @@ def test_create_environment_without_required_metadata_returns_400(
     assert "Missing required metadata field" in response.json()["metadata"][0]
 
 
+def test_view_environment_with_staff_track_query_count(
+    staff_client: FFAdminUser,
```
gagantrivedi: This should be `APIClient`

gagantrivedi (changes requested):
Think we should try to add new tests under the `tests` directory only. The tests here should live under `/tests/unit/environment/test_views`, right @matthewelwell ?

api/permissions/permission_service.py
matthewelwell: Hmm... yeah, this is interesting. I guess for now, let's limit the scope of this PR to the N+1 caused by metadata. I think probably the better way to tackle this issue is to remove the `"description"` from wherever we list the environments. I've added this to the V2 API GH issue [here](https://github.com/Flagsmith/flagsmith/issues/1667) but we may be able to achieve this before then by creating a new endpoint perhaps.

zachaysan:
> Think we should try to add new tests under the `tests` directory only. The tests here should live under `/tests/unit/environment/test_views`, right @matthewelwell ?

Sounds good. I've moved all the tests into the target file, including the one I wrote.

matthewelwell (approved):

gagantrivedi (approved):

PR #2951, opened 2023-11-09
fix: Handle payment errors during user flow

zachaysan:

## Changes

[Ticket reference](https://github.com/orgs/Flagsmith/projects/2/views/11?pane=issue&itemId=35443590)

Before this PR we were treating all errors from the chargebee client as one thing and we responded to a user that was in the process of signing up by issuing a 500 error, we were also logging all the chargebee errors as exceptions even though a declined credit card shouldn't be. Now, if the error is payment related, a `400` error gets sent to the client with a message to contact their administrator. 

A combination of manual probing through the code with the test suite on and two new tests. 

## Remaining TODOs
- [x] Figure out what other callsites there are that will be impacted by this change, if any.
- [x] See if there are any other other areas of the codebase with invite-related code need to handle payment failures.
- [x] Manually test this somehow. I have access to flagsmith-test.chargebee.com but I'm not sure what server it is connected to. Is that staging?
- ~[ ] Team up with @novakzaballa and get the frontend hooked up so the change will actually matter.~

### DIFF
--- a/api/organisations/chargebee/chargebee.py
+++ b/api/organisations/chargebee/chargebee.py
@@ -12,6 +12,7 @@ from ..subscriptions.constants import CHARGEBEE
 from ..subscriptions.exceptions import (
     CannotCancelChargebeeSubscription,
     UpgradeSeatsError,
+    UpgradeSeatsPaymentFailure,
 )
 from .cache import ChargebeeCache
 from .constants import ADDITIONAL_SEAT_ADDON_ID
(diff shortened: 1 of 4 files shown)


api/organisations/chargebee/chargebee.py
```diff
@@ -177,6 +186,14 @@ def add_single_seat(subscription_id: str):
         )
 
     except ChargebeeAPIError as e:
+        api_error_code = e.json_obj["api_error_code"]
+        if api_error_code in CHARGEBEE_PAYMENT_ERROR_CODES:
+            logger.warning(
+                "Payment declined during additional seat upgrade to a "
+                f"CB subscription for subscription_id {subscription_id}"
+            )
```
zachaysan: I wasn't sure whether we wanted to log the specific payment failure here or not.

matthewelwell: I think probably best to include it, yep.

matthewelwell: I have added this so we can get this merged 👍

matthewelwell (approved):

api/organisations/chargebee/chargebee.py
zachaysan: Thank you!!

PR #2957, opened 2023-11-10
fix: Infinite loop 404 after leaving the organisation

### LINKED ISSUE #2955: App can get stuck in 404 loop after leaving an organisation
Steps to reproduce: 

1. At app.flagsmith.com log in as a user with access to organisations A & B and navigate to a project in Organisation B
2. In the django admin, remove access for that user to organisation B
3. At app.flagsmith.com, refresh the page

This should result in the application entering a refresh loop, displaying a 404 page each time. Seemingly the only way to resolve it is to clear cookies and log in again. 

The steps above show a way to replicate this using the Django admin console, however, it is something that I believe to be happening to other users as well so there are likely other steps to reproduce through app.flagsmith.com only.

novakzaballa:

## Changes

- Removed the item 'lastEnv' from AsyncStorage the user is no longer a member of the organization.

1. At app.flagsmith.com log in as a user with access to organisations A & B and navigate to a project in Organisation B.
2. In the Django admin, remove access for that user to organisation B.
3. At app.flagsmith.com, refresh the page.
4. The app no longer enters an infinite loop.

### DIFF
--- a/frontend/web/components/pages/NotFoundErrorPage.js
+++ b/frontend/web/components/pages/NotFoundErrorPage.js
@@ -14,8 +14,8 @@ const ComingSoon = class extends Component {
         <h3 className='pt-5'>Oops!</h3>
         <p>
           It looks like you do not have permission to view this{' '}
-          {Utils.fromParam().entity || 'page'}. Please contact a member with
-          administrator privileges.
+          {Utils.fromParam().entity || 'page'} or you are not a member of the
+          organisation . Please contact a member with administrator privileges.
         </p>
       </div>
     )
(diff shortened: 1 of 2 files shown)


frontend/web/components/pages/NotFoundErrorPage.js
```diff
@@ -14,8 +14,8 @@ const ComingSoon = class extends Component {
         <h3 className='pt-5'>Oops!</h3>
         <p>
           It looks like you do not have permission to view this{' '}
-          {Utils.fromParam().entity || 'page'}. Please contact a member with
-          administrator privileges.
+          {Utils.fromParam().entity || 'page'} or you are not a member of the
+          organisation . Please contact a member with administrator privileges.
```
matthewelwell: I don't think this needs to change, I don't want to leak any extra data.

matthewelwell (changes requested):
I've tested and confirmed this looks good. One minor comment related to the error message change.

matthewelwell (approved):
I reverted the error message change.

Issue #2197, opened 2023-05-12
When a paid SaaS account is cancelled, we should remove all but the first user from the account


dabeeeenster:
Currently cancelling a paid SaaS account doesnt remove any of the users/seats from the Organisation. Upon receiving the Account Cancelled webhook from Chargebee, we should remove all but the first Flagsmith user associated with the Organisation.

## Acceptance Criteria

 - [x] When a subscription that is linked to a Flagsmith organisation is cancelled in Chargebee, all but one users should be removed from the organisation in Flagsmith
 - [ ] (To verify if possible) Warning text is displayed with the cancel account logic to confirm that all but one users will be removed from the account in Flagsmith

matthewelwell:
Related to [#1822](https://github.com/Flagsmith/flagsmith/issues/1822)

PR #2965, opened 2023-11-13
feat: Remove all but first admin when subscription has reached cancellation date

zachaysan:

## Changes

Added code to existing cancellation workflow to limit the seats to a sole admin once the subscription cancellation date has finally been reached. 

Two new tests. Once that goes through the web request for cancellations that are immediate, and another for any that have a cancellation date in the future. I also read through a lot of code around subscriptions to see if there were other areas that could benefit from delayed processing of cancelation but I didn't find any sure contenders.


api/organisations/tasks.py
```diff
@@ -41,3 +48,16 @@ def update_organisation_subscription_information_cache():
     subscription_info_cache.update_caches(
         (SubscriptionCacheEntity.CHARGEBEE, SubscriptionCacheEntity.INFLUX)
     )
+
+
+@register_recurring_task(
+    run_every=timedelta(hours=12),
+)
```
matthewelwell, line 55: I'm curious to know if there's a benefit to implementing this as a recurring task vs just scheduling it for the cancellation date? 

e.g. if we mark this as just a regular task (`@register_task`) then we can do the following in the cancel endpoint. 

```python
finish_subscription_cancellation.delay(delay_until=...)
```

Although perhaps this approach is better because it captures any that might be cancelled, e.g. via the django admin?

api/tests/unit/organisations/test_unit_organisation_tasks.py
```diff
@@ -0,0 +1,77 @@
...
+def test_finish_subscription_cancellation():
+    # Given
+    organisation1 = Organisation.objects.create()
+    organisation2 = Organisation.objects.create()
+    organisation3 = Organisation.objects.create()
+    organisation4 = Organisation.objects.create()
+
+    # Far future cancellation will be unaffected.
+    organisation_user_count = 3
+    for __ in range(organisation_user_count):
```
matthewelwell, line 23: I've seen this in your code before but didn't say anything, is there a reason you go with `__` instead of (my understanding of the convention) `_`?

api/tests/unit/organisations/test_unit_organisation_tasks.py
```diff
@@ -0,0 +1,77 @@
...
+from users.models import FFAdminUser
+
+
+@pytest.mark.django_db
+def test_finish_subscription_cancellation():
+    # Given
+    organisation1 = Organisation.objects.create()
+    organisation2 = Organisation.objects.create()
+    organisation3 = Organisation.objects.create()
+    organisation4 = Organisation.objects.create()
```
matthewelwell, line 19: I'm kinda curious how this test is passing as it's my understanding that `name` is a required field here 🤔

api/tests/unit/organisations/test_unit_organisation_tasks.py
```diff
@@ -0,0 +1,77 @@
...
+from organisations.models import (
+    Organisation,
+    OrganisationRole,
+    UserOrganisation,
+)
+from organisations.tasks import finish_subscription_cancellation
+from users.models import FFAdminUser
+
+
+@pytest.mark.django_db
```
matthewelwell: Typically we've used the `db` fixture rather than this mark, but perhaps this is your way of getting around the typing in tests 😉 

As far as I'm aware, it does the same thing so it's not an issue.

api/organisations/models.py
```diff
@@ -157,6 +157,20 @@ def rebuild_environments(self):
...
 
+    def cancel_users(self):
+        remaining_seat_holder = (
+            UserOrganisation.objects.filter(
+                organisation=self,
+                role=OrganisationRole.ADMIN,
+            )
+            .order_by("date_joined")
+            .first()
+        )
```
matthewelwell, line 168: @dabeeeenster to confirm as per the related issue (#2197), the logic we are implementing is such that we leave the first organisation admin that joined the organisation as the only user left on cancellation. In nearly all cases I imagine this will be the user that created the organisation.

api/tests/unit/organisations/test_unit_organisation_views.py
```diff
@@ -1340,3 +1340,37 @@ def test_list_my_groups(organisation, api_client):
         "id": user_permission_group_1.id,
         "name": user_permission_group_1.name,
     }
+
+
+def test_when_subscription_is_cancelled_then_remove_all_but_the_first_user(
+    staff_client, admin_client, subscription, organisation
```
matthewelwell: typing please :)

api/tests/unit/organisations/test_unit_organisation_tasks.py
zachaysan: In Django there is the convention of doing this:

```python
from django.utils.translation import gettext_lazy as _
```

And so the modern advice for throwaway variables is now to use two underscores.

api/tests/unit/organisations/test_unit_organisation_tasks.py
zachaysan: The model doesn't doesn't have `blank=False` on it. Maybe you're thinking of a seralizer that enforces it?

api/organisations/tasks.py
zachaysan: Lots of experience with this type of thing for the erasure work at Clubhouse taught me not to rely on task scheduling far in the future, too many things can go wrong. Plus, as you say, this decouples criteria and action so it would work for Django admin as well.

api/tests/unit/organisations/test_unit_organisation_tasks.py
zachaysan: Hahaha, yeah, just a force of habit. Will fix.

api/tests/unit/organisations/test_unit_organisation_views.py
zachaysan: Nice catch! Let's get typing into the checklist when opening PRs. It's hard to remember to do it sometimes.

api/tests/unit/organisations/test_unit_organisation_tasks.py
matthewelwell: Interesting! TIL! Makes sense, thanks.

api/tests/unit/organisations/test_unit_organisation_tasks.py
matthewelwell: I was under the impression that `blank=False` was the default (as per [this](https://github.com/django/django/blob/main/django/db/models/fields/__init__.py#L192)). Clearly you're right, I'm just baffled! 😂

zachaysan: Huh. Interesting link. I don't know why it works this way for the char field, but I for one like to always specify `blank` and `null` explicitly because when I'm in a rush I don't want to have to look up what the default is for these things.

api/tests/unit/organisations/test_unit_organisation_tasks.py
zachaysan: Ok this is fixed now.

api/tests/unit/organisations/test_unit_organisation_views.py
zachaysan: Ok this is handled.

api/organisations/models.py
dabeeeenster: Yes this is fine

api/tests/unit/organisations/test_unit_organisation_tasks.py
matthewelwell: Seems like `CharField` somehow defaults to `''` which I would _still_ expect to fail! Anyway, if I'm being picky, I would rather than these were all instantiated with a name (to avoid this conversation in the future) but I'll approve the PR regardless and you can choose to add the `name` attribute or not.

api/tests/unit/organisations/test_unit_organisation_views.py
```diff
@@ -1340,3 +1340,40 @@ def test_list_my_groups(organisation, api_client):
         "id": user_permission_group_1.id,
         "name": user_permission_group_1.name,
     }
+
+
+def test_when_subscription_is_cancelled_then_remove_all_but_the_first_user(
+    staff_client: APIClient,
+    admin_client: APIClient,
```
matthewelwell: I don't think we need the `admin_client` here?

matthewelwell (approved):
Approved with one minor additional comment (and an update on the `blank=False` conversation).

api/tests/unit/organisations/test_unit_organisation_views.py
zachaysan: Yep!

PR #2973, opened 2023-11-14
fix: Logic in segment overrides readonly with the manage_segment_overrides permission

novakzaballa:

## Changes

- Invert a boolean value when the user has manage_segment_overrides permission is not `readOnly`

- Go to edit feature modal.
- Select the Segment Overrides Tab.
- If the user has manage_segment_overrides permission can create, update, and delete a segment override.
- if the user does not have permission Segment override component is `readOnly`


zachaysan (approved):
Looks good

PR #2976, opened 2023-11-15
feat: introduce dunning billing status

zachaysan:

## Changes

Introduce `billing_status` on the `Subscription` model, which will be used to show users a dunning butterbar if a subscription has a payment failed event with dunning enabled. This is done by listening for two separate webhooks from chargebee, one for payment failures and the other for payment successes and updating the billing status accordingly.

I used the webhook test feature built by chargebee to get the data format and added three new tests to cover the usecases. One test is a simple organisations list endpoint to ensure that the data is readable by a future JavaScript client.


api/organisations/views.py
```diff
@@ -272,6 +355,12 @@ def chargebee_webhook(request):
      - If subscription is cancelled or not renewing, update subscription on our end to include cancellation date and
        send alert to admin users.
     """
+    event_type = request.data.get("event_type")
+
+    if event_type == "payment_failed":
+        return chargebee_webhook_payment_failed(request)
+    if event_type == "payment_succeeded":
+        return chargebee_webhook_payment_succeeded(request)
```
zachaysan: I'm not 100% sure we want to early exit here. The way this function operated before I added the payment failed and payment succeeded early exits, rightly or wrongly, was to look at every single webhook and if it had subscription data, update the subscription information we have in the data base. This shouldn't be necessary since only the subscription specific webhooks should be necessary for keeping our subscription information up to date, but if we want to handle the payment_failed and payment_succeeded requests in the same way it is a simple change.

api/organisations/models.py
```diff
@@ -212,6 +220,7 @@ def update_mailer_lite_subscribers(self):
 
     def cancel(self, cancellation_date=timezone.now(), update_chargebee=True):
         self.cancellation_date = cancellation_date
+        self.billing_status = ""
```
matthewelwell: I kind of prefer `None` here to indicate no status but it's a minor nitpicky preference.

api/organisations/views.py
```diff
@@ -260,9 +265,87 @@ def my_permissions(self, request, pk):
         return Response(serializer.data)
 
 
+def chargebee_webhook_payment_failed(request: Request) -> Response:
```
matthewelwell: I'd argue that this (and `chargebee_webhook_payment_success`) should be refactored into functions in e.g. the `chargebee` module that get called by the view here.

api/organisations/views.py
```diff
@@ -260,9 +265,87 @@ def my_permissions(self, request, pk):
...
+
+    invoice = request.data["content"]["invoice"]
+
+    # If dunning hasn't started ignore the declined payment.
+    if invoice.get("dunning_status") != "in_progress":
+        return Response(status=status.HTTP_200_OK)
+
+    if "subscription_id" not in invoice:
+        logger.info("Payment declined for non-subscription related charge, skipping.")
+        return Response(status=status.HTTP_200_OK)
```
matthewelwell: There are a lot of returns here (I have a feeling that flake8 may even complain about the complexity). Could we take the opportunity to start using some sort of serialization here? Perhaps a DRF serializer or just a marshmallow schema?

api/organisations/views.py
matthewelwell: I think this is fine. As per the above though, I'd love to start using some level of serialization here as well.

api/organisations/views.py
matthewelwell: Actually, I'd love to see us refactor all of this logic into some kind of `ChargebeeWebhookHandler` class  (with static methods maybe), or just a `chargebee_webhook_handler` module to really separate things out.

api/tests/unit/organisations/test_unit_organisation_views.py
```diff
@@ -1340,3 +1342,91 @@ def test_list_my_groups(organisation, api_client):
         "id": user_permission_group_1.id,
         "name": user_permission_group_1.name,
     }
+
+
+def test_payment_failed_chargebee_webhook(
```
matthewelwell, line 1356: This comment may not be relevant if we rely on serialization (since we can then rely on e.g. DRF tests) but these tests are only covering the happy paths, I'd love to see more coverage.

api/organisations/models.py
zachaysan: Oh yeah, me too. Sorry for the oversight, `None` is definitely the way to go, although we'll still get `""` in the db due to Django admin sending empty strings.

api/organisations/views.py
zachaysan: Good idea, I've got something basic working here and I should have an updated PR by tomorrow.

api/organisations/views.py
zachaysan: I've switched it to a serializer locally, but it does come at the cost of detailed error messages. Even so, I've added stacktraces to the logged data so we should be ok. Will update this PR by tomorrow.

api/tests/unit/organisations/test_unit_organisation_views.py
zachaysan: I can add more tests. Seems silly sometimes to do a whole test for a single small branch, but it's often worth it.

zachaysan: Ok I added more non-happy path tests.

api/organisations/chargebee/webhook_handlers.py
```diff
@@ -0,0 +1,86 @@
...
+def payment_failed(request: Request) -> Response:
+    serializer = PaymentFailedSerializer(data=request.data)
+
+    try:
+        serializer.is_valid(raise_exception=True)
+    except ValidationError:
+        logger.warning(
+            "Serializer failure during chargebee payment failed processing",
+            exc_info=True,
+        )
```
matthewelwell, line 28: Given our current conversations about logging visibility, I'm keen for this to fail loudly. We could either change this to `logger.error` (probably the best option), or perhaps we can just remove the `try` / `except`.

api/organisations/chargebee/serializers.py
```diff
@@ -0,0 +1,32 @@
...
+    subscription_id = CharField(allow_null=False)
+
+    def validate_dunning_status(self, value):
+        if value != "in_progress":
+            raise ValidationError("To be valid dunning must be in progress")
+        return value
+
+
+class PaymentSucceededInvoiceSerializer(Serializer):
+    subscription_id = CharField(allow_null=False)
```
matthewelwell, line 16: The code is definitely much cleaner for using these serializers, but I'm not entirely sure why we need different serializers for the different event types, couldn't we use a single serializer here?

api/organisations/chargebee/webhook_handlers.py
```diff
@@ -0,0 +1,86 @@
...
+def payment_succeeded(request: Request) -> Response:
+    serializer = PaymentSucceededSerializer(data=request.data)
+
+    try:
+        serializer.is_valid(raise_exception=True)
+    except ValidationError:
+        logger.warning(
+            "Serializer failure during chargebee payment failed processing",
+            exc_info=True,
+        )
```
matthewelwell, line 63: As above, let's fail loudly here!

api/organisations/chargebee/webhook_handlers.py
zachaysan: Well, we could issue an error here, but it's not really an error in every case if we ever get payments that aren't associated with a subscription or if the dunning hasn't come into effect. We certainly don't want to remove the `try` / `except` because we want to continue to mark the webhook as processed, as a non `2xx` HTTP status code will have chargebee continue to retry the webhook over a short time.

api/organisations/chargebee/serializers.py
zachaysan: Given that only the payment failed webhook has the `dunning_status` field I think it's better to separate out the serializers. I generally prefer smaller serializers over ones which need to introspect to determine the event type in order to enforce different policies.

api/organisations/chargebee/webhook_handlers.py
zachaysan: See my other comment. I'm not sure we necessarily want to fail loudly.

api/organisations/chargebee/serializers.py
matthewelwell: Yep, ok fair enough.

api/organisations/chargebee/webhook_handlers.py
matthewelwell: Ok, makes sense.

matthewelwell (approved):

PR #2981, opened 2023-11-16
feat: add copy tooltip to feature flag names

gnumoreno:

## Changes
Adding a button to the right side of the feature name, after the tooltip with the create date.

Running on my local environment the page renders correctly, the button is there and clicking on the button it copies the feature flag name.

image

### DIFF
--- a/frontend/web/components/FeatureRow.js
+++ b/frontend/web/components/FeatureRow.js
@@ -198,6 +198,7 @@ class TheComponent extends Component {
                   rowGap: 4,
                   wordBreak: 'break-all',
                 }}
+                
               >
                 <span className='me-2'>
                   {created_date ? (
@@ -219,6 +220,24 @@ class TheComponent extends Component {
                     name
                   )}
                 </span>
+                <Tooltip
+                  html
+                  title={
+                    <div
+                      onClick={(e) => {
+                        e.stopPropagation()
+                        navigator.clipboard.writeText(
+                          projectFlag.name,
+                        )
+                        toast('Copied')
+                      }}
+                    >
+                      <Icon name='copy' width={20} fill='#9DA4AE' />
+                    </div>
+                  }
+                >
+                  Copy feature name
+                </Tooltip>
 
                 {!!projectFlag.num_segment_overrides && (
                   <div


kyle-ssg:
I'm not too sure about this to be honest. 

Features have tooltips, tags, segment overrides, identity overrides and remote config in the same row.  Adding a copy button would add to the already crowded feature row.

Not sure if @dabeeeenster has any thoughts on this?

novakzaballa:
When we have other elements in the FF title, it looks a bit messy. We would need also equalize the spaces between the elements
Screen Shot 2023-11-16 at 10 20 53

dabeeeenster:
What about moving the description to a tooltip and putting all the other icons / lozenges (?) in its place underneat the flag name?

Or what about changing the flag name treatment that shows a copy icon when its onHover?

jatingodnani:
@dabeeeenster  to make ui more clear, we can make a menu for each row which contain this properties

novakzaballa:
> What about moving the description to a tooltip and putting all the other icons / lozenges (?) in its place underneat the flag name?
> 
> Or what about changing the flag name treatment that shows a copy icon when its onHover?

I would go with that one or something similar to @jatingodnani 's suggestion, since we are showing the description because users asked for that. The description used to live in the tooltip.

jatingodnani:
@novakzaballa  If flagsmith team allow, i am intrested to do

dabeeeenster:
> @novakzaballa If flagsmith team allow, i am intersted to do

Sure please submit a PR!

jatingodnani:
@dabeeeenster What are the properties you want to add in menu

novakzaballa:
> @dabeeeenster What are the properties you want to add in menu

@jatingodnani I would like to see a mock first of each of the two options. Could you please upload a couple of mocks of how on-hover changing the view to show actionable items would look like? And also another one with the menu. 
The on-hover option should work like this:
- Initially, we show as much information as possible, that is we show all the current info like description, in addition to the info showing the tooltip, the `created at` value.
- On hover, we hide all the informational items and only show the actionable items, like the enable/disable toggle, and the button to copy the flag name, the remove feature buitton, and the history button.

novakzaballa:
For the On-hover option, we should remove the i icon in both views, and also we should also add a pen icon in the on-hover view.

novakzaballa:
This is how a feature-flag row currently  looks like without the copy button. It is a bit messy already:
image
This is how it would look like without the actionable items, gaining some space for the feature value and making it a bit cleaner:
image
The on-hover view would be something like this (Note. this is only an idea not a mock of the final UI, we are open to suggestions):
image
Or maybe even better putting the copy icon as a big button at the left of the history one. Clicking anywhere in the row but over an actionable item should open the edit feature dialog, just as the current behavior.

Notice that the current buttons' behavior should be kept, that means, considering the permissions, context, etc.

jatingodnani:
@novakzaballa if we make one button with name tags, on hover to button we can show all the tags

novakzaballa:
@jatingodnani , while hiding the tags would leave a really clean view, I would rather leave them since I think that is one of the elements that the user value to see at a glance.

jatingodnani:
@novakzaballa so you are saying we can hife the tag,and onhover we need to show the tag?

jatingodnani:
@novakzaballa i am interested to take on any frontend issue,if team decided the ui improvement for,i amready to solve this

novakzaballa:
@jatingodnani , We would love to see a PR from you for this issue.

> @novakzaballa so you are saying we can hife the tag,and onhover we need to show the tag?

We should show all the tags in the normal view and hide them all on hover. Actually, all informative elements shown be shown in the initial view, including the created at date, and all of them should be hidden on hover, except by the name and description, and the enable/disble toggle, and we should show all the actionable elements, like delete flag, history, copy, etc..

jatingodnani:
@novakzaballa if anyone hover then it show actionable element other all informative info?if yes,i will this and raise pr

novakzaballa:
> @novakzaballa if anyone hover then it show actionable element other all informative info?if yes,i will this and raise pr

Sorry, I don't get it. But we could use this mock views as a base:
Suggested initial view of the flag row:
image
Suggested On hover view of the flag view:
image
These are only suggestions, but we can elaborate using those as a base-line.

jatingodnani:
@novakzaballa i will make draft pr

novakzaballa:
Awesome! Thank you @jatingodnani.
Please notice that in the suggested initial view (the informational view) we are missing the Created at date, which is currently shown as a tooltip. We need to show that somewhere, please suggest a solution to that in your draft PR.

MatheusLasserre:
Hey! i was watching the discussion and have come with an idea. Instead of toggling the panel, what about an expansible "tool belt"?

But in this case it could introduce a reusable pattern through the app, which would be the way to collapse minor actions in one button. What do you think?

I've made some changes and recorded a video so i can show what i though. Just an idea.

See below:

https://github.com/Flagsmith/flagsmith/assets/77826309/50364017-f234-4318-88a1-aaed07815820

kyle-ssg:
After going through the comments, PR and above video with @novakzaballa we are currently thinking the following is the best way forward:

- Adding a compact mode that can be toggled (and persists in local storage), this will put the description inside the tooltip and reduce down the padding. Up until recently it was the case by default. This would mean we could have features live on pretty much a single row if you want them to.
- Moving actions (Copy, view history and remove) to a context menu similar to above only the trigger wouldn't be on the tooltip. I think we could still do this on hover however, @novakzaballa like the approach above and in @jatingodnani's PR vs a standard clickable context menu. It's great to see these suggestions, so thank you very much!

I think whatever the course is, we want to go about adding this to our design system rather than implementing the UI pattern as a one off.

jatingodnani:
@novakzaballa should i complete this  pr or not?

novakzaballa:
@jatingodnani, we think your proposal as well as @MatheusLasserre's one are both awesome. As Kyle mentioned in the original PR https://github.com/Flagsmith/flagsmith/pull/2981#issuecomment-1817553399 we will bring this back to our design team and probably perform A/B testing if needed to pick the user's preferred solution, since we want to incorporate whatever solution we pick not just to this particular view but for most of our similar views in the platform. So we can leave it as is by now since it is enough to give our design team an idea of how this option would look like, they will decide the next steps and we will communicate that to you. Again, thank you so much for helping us make this product better, we really appreciate it!

gnumoreno:
@kyle-ssg we ended up not adding the "copy" functionality. Are we still planning to?

dabeeeenster:
I think we definitely need this functionality!

PR #2986, opened 2023-11-16
chore: improve UPDATE_FEATURE_STATE permission in UI

novakzaballa:

## Changes

- The "Update segment override" button is no longer affected by UPDATE_FEATURE_STATE permission

- Give and remove UPDATE_FEATURE_STATE permission in the environment setting page
- If you have permission you can modify the feature flag.
- Having or not having permission does not affect the "segment" overrides tab

### DIFF
--- a/frontend/web/components/modals/CreateFlag.js
+++ b/frontend/web/components/modals/CreateFlag.js
@@ -1214,68 +1214,47 @@ const CreateFlag = class extends Component {
                                                   }
                                                 </strong>
                                               </p>
-                                              <Permission
-                                                level='environment'
-                                                permission={Utils.getManageFeaturePermission(
-                                                  is4Eyes,
-                                                  identity,
-                                                )}
-                                                id={this.props.environmentId}
-                                              >
-                                                {({
-                                                  permission: savePermission,
-                                                }) =>
-                                                  Utils.renderWithPermission(
-                                                    savePermission,
-                                                    Constants.environmentPermissions(
-                                                      Utils.getManageFeaturePermissionDescription(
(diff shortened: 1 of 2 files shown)


matthewelwell:
image

Looks like there's something on the API preventing it still @zachaysan

matthewelwell:
I've fixed the above in #2990.

There is one minor FE tweak that I've noticed which would be good to resolve which is that the tooltip which comes up on hover on the feature row is kind of misleading now. 

image

I have a feeling this has been misleading since we allowed users to view the feature slide out, but it's a good time to resolve it here I think @novakzaballa.

I think we can just remove the tooltip altogether here.

matthewelwell:
I've removed the tooltip [here](https://github.com/Flagsmith/flagsmith/pull/2986/commits/855e3f63604bb6673684be344d17c28ad9bde174)

matthewelwell (dismissed):

frontend/web/components/modals/CreateFlag.js
```diff
@@ -1214,68 +1214,71 @@ const CreateFlag = class extends Component {
...
+                                                        'MANAGE_SEGMENT_OVERRIDES'
+                                                      }
+                                                      id={
+                                                        this.props.environmentId
+                                                      }
+                                                    >
+                                                      {({
+                                                        permission:
+                                                          manageSegmentsOverrides,
+                                                      }) => {
```
matthewelwell, line 1241: I don't quite understand what this code is doing but I have tested it functionally and it seems to behave how I expect!

matthewelwell (approved):

Issue #2942, opened 2023-11-08
Add option to copy feature name from dashboard and edit feature modal


matthewelwell:
On the dashboard and the edit feature modal, it would be good to have an option to copy the feature name as we do for environment keys.

gnumoreno:
https://github.com/Flagsmith/flagsmith/pull/2981

PR #2999, opened 2023-11-18
feat: draft of new feature flag panel design

MatheusLasserre:

## Changes

Added a new component called "Actionables" which groups all actions represented as icons.
Modified PanelSearch component to receive "viewMode" and "toggleViewMode", so it shows a Switch on the left of the filters ta toggles the viewMode.

That's just a draft of what i understood about @kyle-ssg and @novakzaballa ideas. There's still a couple problems:

1. I'm not used to the codebase tools, so i'm probably using something that could be replaced by a already existing tool;
2. I couldn't make it initiate the view State from localStorage, and honestly, i don't really know why i couldn't grab the value calling the localStorage from componentDidMount. So it saves the state, but can't retrieve it.
3. The styling is more a POC than a real one.  

I tested through my development environment at the features page.

Video for a quick peek:

https://github.com/Flagsmith/flagsmith/assets/77826309/dc3a8a05-fee2-4277-9a4d-1f248163ed9d

Related to: #2942 and #2981

### DIFF
--- a/frontend/web/components/Actionables.tsx
+++ b/frontend/web/components/Actionables.tsx
@@ -1,0 +1,35 @@
+import React, { useState, type FC } from 'react'
+import { ellipsisVerticalCircleOutline } from 'ionicons/icons'
+import { IonIcon } from '@ionic/react'
+
+type ActionablesProps = {
+  children: React.ReactNode
+  width?: string
+}
+
+// Displays an icon that when hovered expands to the right to show a list of actionable children
+export const Actionables: FC<ActionablesProps> = ({ children, width }) => {
+  const [showList, setShowList] = useState(false)
+
+  // const onMouseLeave = (e: React.MouseEvent<HTMLDivElement, MouseEvent>) => {
+  //   e.stopPropagation()
+  //   setShowList(false)
+  // }
+
+  return (
+    <div
+      className='actionables'
+      style={{ width: width || 20 }}
+      onMouseEnter={() => setShowList(true)}
+      onTouchStart={() => setShowList(true)}
+      onMouseLeave={() => setShowList(false)}
+      onTouchEnd={() => setShowList(false)}
+      onClick={(e) => e.stopPropagation()}
+    >
+      <div className='icon--new-tooltip mr-1'>
+        <IonIcon icon={ellipsisVerticalCircleOutline} />
+      </div>
+      <div className={`list ${showList ? 'showActions' : ''}`}>{children}</div>
+    </div>
+  )
+}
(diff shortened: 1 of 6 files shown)


matthewelwell:
@MatheusLasserre please make sure that you are following the contributing guidelines. PR titles need to follow the Conventional Commit structure, I'm also not sure what 'panel actions state' really means so please make sure the title is descriptive of the change in the PR. Also, this change seems to have broken the E2E testing but I see you've mentioned that there are outstanding problems in the PR. I've marked the PR as draft while it's still in progress.

MatheusLasserre:
> @MatheusLasserre please make sure that you are following the contributing guidelines. PR titles need to follow the Conventional Commit structure, I'm also not sure what 'panel actions state' really means so please make sure the title is descriptive of the change in the PR. Also, this change seems to have broken the E2E testing but I see you've mentioned that there are outstanding problems in the PR. I've marked the PR as draft while it's still in progress. 

Sorry. I've made some changes to make sure it's following the guidelines.

The broken E2E, it's probably with the problem of the tools I've mentioned. But this PR aims more to give an idea than to be merged, unless you really like the ideia. In that case, i would change it.

dabeeeenster:
My main concern here is the mouse movement required to hit these buttons is very fiddly...I like the idea overall but maybe it should be a clickable element with a popup rathen than something that appears onHover?

MatheusLasserre:
Maybe i could make it sizes differently based on the compact/normal view?

I've made it onHover because @novakzaballa mentioned that he would like to not have the usual clickable context menu.

And since the idea of the <Actionables/> I've created was to store quick actions, like the created at, clipboard, etc, i think the onHover makes sense.

The scenario i imagine to think of this is the following:

A user wants to copy the name and check the creation date of some feature flags.

1. OnHover: 
Hover the card, clicks copy, hover created at, then next.

2. Clickable:
Clicks menu, clicks copy, hover created at, clicks outside, then next.

I feel like the onHover approach makes more sense is that scenario.

 Changing it based on the view would change the following:

1. The actual model would stay the same for the normal view, since it adapts itself for the height of the upper row, which we always know the height.

2. The actual model would change when in compact view to have the icons the same size as the current in prod "feature flag history" button, while it would maybe reduce the width of the flags, to make it easy to reach each action.

Right now the icons are with the same size as the default tooltip.

If that new idea makes sense, i can make some changes and post a new video here to show it.

matthewelwell:
Thanks for the work here @MatheusLasserre and all the conversations but, as per the conversation on [this PR](https://github.com/Flagsmith/flagsmith/pull/2981), we're going to take a step back and look at this with our design team. AS such, I'll go ahead and close this PR for now.

PR #3009, opened 2023-11-20
fix: Reading role permissions generates 500 errors

### LINKED ISSUE #3003: Reading role permissions generates 500 errors
Steps to reproduce:

- Create Role
- Create User
- Assign User to Role
- The role is persisted if you look in the "Role/User" tab but if you view the user permissions directly you get the following:

image

I can reproduce this in a fresh org on Prod

novakzaballa:

## Changes

- Skip requests when the role organization is not sent
- Use new RBAC endpoint to retrieve and delete userRoles by user

- Go to manage -> members -> Users
- Open the Network tab
- Click any User
- The error is no longer there


matthewelwell:
Based on [this Slack thread](https://flagsmith.slack.com/archives/CTF0THS2D/p1700559317931689), I'm not sure this is the right solution. All roles should have an organisation attribute. I've marked this PR as draft until we have understood more about this.

frontend/web/components/EditPermissions.tsx
```diff
@@ -733,6 +735,7 @@ const EditPermissions: FC<EditPermissionsType> = (props) => {
   } = props
   const [tab, setTab] = useState()
   const editUserPermissions = (user: User) => {
+    console.log('DEBUG: user:', user)
```
matthewelwell: Can you remove this?

frontend/web/components/EditPermissions.tsx
```diff
@@ -241,6 +241,8 @@ const _EditPermissionsModal: FC<EditPermissionModalType> = forwardRef(
           skip:
             !id ||
             envId ||
+            // TODO: In the future, perhaps use inherited components to avoid issues with merging multiple contexts into a single component.
```
matthewelwell: We discussed creating an issue for this and linking to it here, right?

matthewelwell (changes requested):

frontend/web/components/EditPermissions.tsx
novakzaballa: Done

frontend/web/components/EditPermissions.tsx
novakzaballa: Yes, created and added

api/users/models.py
```diff
@@ -246,6 +249,10 @@ def get_user_organisation(
                 % (self.id, getattr(organisation, "id", organisation))
             )
 
+        def get_user_roles(self):
+            user_roles = UserRole.objects.filter(user=self)
+            return user_roles
```
matthewelwell:
```suggestion
            return UserRole.objects.filter(user=self)
```

frontend/common/dispatcher/app-actions.js
```diff
@@ -283,10 +283,11 @@ const AppActions = Object.assign({}, require('./base/_app-actions'), {
       projectId,
     })
   },
-  getOrganisation(organisationId) {
+  getOrganisation(organisationId, force = false) {
```
matthewelwell: `force` feels a little unclear to me here, can we be more explicit with the variable name? What are we forcing?

frontend/web/components/modals/CreateRole.tsx
```diff
@@ -462,6 +462,7 @@ const CreateRole: FC<CreateRoleType> = ({
                   ))}
               </div>
             </TabItem>
+            {/* add padding:0*/}
```
matthewelwell: Not sure why this is here?

frontend/common/stores/organisation-store.js
```diff
@@ -329,7 +329,11 @@ store.dispatcherIndex = Dispatcher.register(store, (payload) => {
 
   switch (action.actionType) {
     case Actions.GET_ORGANISATION:
-      controller.getOrganisation(action.id || store.id, action.force)
+      controller.getOrganisation(
+        action.id,
+        action.force || store.id,
+        action.force,
```
matthewelwell: This is quite confusing, what are each of the arguments here?

api/users/serializers.py
```diff
@@ -56,9 +60,15 @@ class Meta:
 class UserListSerializer(serializers.ModelSerializer):
     role = serializers.SerializerMethodField(read_only=True)
     join_date = serializers.SerializerMethodField(read_only=True)
+    if settings.IS_RBAC_INSTALLED:
+        roles = UserRoleSerializer(many=True, read_only=True, source="get_user_roles")
```
matthewelwell: I'd love to see some tests for this. It might be tough to do here though because of the dependency on the rbac library. I wonder if this would be better suited as a new endpoint, which lives in the RBAC repository?

frontend/web/components/modals/CreateRole.tsx
novakzaballa: It was a comment to fix a design style, fixed and the comment deleted.

api/users/serializers.py
novakzaballa: I saw that you had a similar comment in another PR, yes I think you are right, the serializers would no longer depend on RBAC and could be tested more easily, I will do it.

dabeeeenster:
I'm now getting JS exceptions:

```
react-dom.production.min.js:187 TypeError: Cannot read properties of undefined (reading 'map')
    at EditPermissions.tsx:270:40
    at Ma (react-dom.production.min.js:188:202)
    at Ps (react-dom.production.min.js:241:307)
    at t.unstable_runWithPriority (scheduler.production.min.js:20:134)
    at mi (react-dom.production.min.js:113:325)
    at Ts (react-dom.production.min.js:241:82)
    at _s (react-dom.production.min.js:210:152)
    at bs (react-dom.production.min.js:206:372)
    at react-dom.production.min.js:114:115
    at t.unstable_runWithPriority (scheduler.production.min.js:20:134)
```

When clicking here:

image

api/users/serializers.py
novakzaballa: The endpoint was created in RBAC: https://github.com/Flagsmith/flagsmith-rbac/pull/16

frontend/common/stores/organisation-store.js
```diff
@@ -329,7 +329,11 @@ store.dispatcherIndex = Dispatcher.register(store, (payload) => {
 
   switch (action.actionType) {
     case Actions.GET_ORGANISATION:
-      controller.getOrganisation(action.id || store.id, action.force)
+      controller.getOrganisation(
+        action.id,
+        action.force || store.id,
```
kyle-ssg: I think the function here only has 2 arguments, so the last is the typo.

As for the following:

```
    action.force || store.id,
```

I've read this through along with the function itself and I don't think it makes sense. The point of the force parameter is to hit the API regardless of whether we already have data for that org. By adding ||store.id here, we essentially totally remove that functionality?

kyle-ssg (approved):
1 comment on FE code, approving from that perspective other than that

frontend/common/dispatcher/app-actions.js
novakzaballa: this change was reverted.

frontend/common/stores/organisation-store.js
novakzaballa: This change was reverted.

frontend/common/stores/organisation-store.js
novakzaballa: This change was reverted.

api/users/serializers.py
```diff
@@ -58,7 +58,10 @@ class UserListSerializer(serializers.ModelSerializer):
     join_date = serializers.SerializerMethodField(read_only=True)
 
     default_fields = ("id", "email", "first_name", "last_name", "last_login")
-    organisation_users_fields = ("role", "date_joined")
+    organisation_users_fields = (
+        "role",
```
gagantrivedi, line 62: Will this not break if rbac is not installed?

novakzaballa: This is only a format change, however, this "role" field refers to the "OrganizationRole" model, which is not part of the RBAC.

`class OrganisationRole(models.TextChoices):
    ADMIN = ("ADMIN", "Admin")
    USER = ("USER", "User")`

gagantrivedi:
image
Looks like it's trying to use environment API key for organisation ID?

frontend/web/components/base/forms/Tabs.js
```diff
@@ -55,7 +55,9 @@ const Tabs = class extends React.Component {
             return (
               <div
                 key={`content${i}`}
-                className={`tab-item ${isSelected ? ' tab-active' : ''}`}
+                className={`tab-item ${isSelected ? ' tab-active' : ''} ${
+                  this.props.isRoles && 'p-0'
```
kyle-ssg, line 59: Having a prop isRoles on tabs seems very hacky! Just add a prop tabItemClassName ??

novakzaballa:
> image Looks like it's trying to use environment API key for organisation ID?

Corrected.

gnumoreno:
@novakzaballa I'm not sure if this code touches on it but if I go to https://app.flagsmith.com/organisation-settings -> member - groups - create a new group, the new group does not show up. I have to navigate to complete refresh the page for the group to show up.

gnumoreno:
Other than the group thing I commented above, the approvers seems to be working just fine. I still think the user experience is confusing but I will take some time later this week and open a separate issue

novakzaballa:
> @novakzaballa I'm not sure if this code touches on it but if I go to https://app.flagsmith.com/organisation-settings -> member - groups - create a new group, the new group does not show up. I have to navigate to complete refresh the page for the group to show up.

@gnumoreno That part of the code was not touched, this is a good catch, we would have to create an issue.

matthewelwell (approved):

PR #3019, opened 2023-11-21
feat(ui): hide API keys from integrations list

### LINKED ISSUE #3018: Don't show API keys in Integrations list
Currently in the list of integrations we show all the information about each configured integration, including the API key. We should blank out the API key, and only show it in the edit view if the user clicks on a 'show' icon in the edit field. 

**Current View:**

image

----

**Suggested Changes:**

image

Screenshot 2023-11-21 at 18 40 14

matthewelwell:
## Changes

Fixes #3018. 

Note: requires the addition of the attribute `"hidden": true` to each field that should be hidden in each integration. I have done this in staging already. 

Ran the FE locally and verified that the relevant fields were hidden / not hidden.


novakzaballa (approved):
LGTM

PR #3025, opened 2023-11-23
chore: use self-hosted defaults environment by default

matthewelwell:
## Changes

Reconfigures the FE logic to use the self hosted default environment by default. This means that all customers downloading the flagsmith images from now on will use self-hosted defaults environment to retrieve their initial flags, rather than the staging environment. 

Note that, as part of this, I have also updated the self-hosted defaults environment to match production. 

Built and ran the `flagsmith-frontend` and `flagsmith` images locally and confirmed that they are retrieving their flags from the self hosted defaults environment.

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -11,11 +11,15 @@ COPY --chown=node:node frontend/package.json frontend/package-lock.json frontend
 COPY --chown=node:node frontend/bin/ ./bin/
 COPY --chown=node:node frontend/env/ ./env/
 
-RUN npm ci --quiet --production
+# since ENV is only used for the purposes of copying the correct
+# project_${env}.js file to common/project.js, this is a build arg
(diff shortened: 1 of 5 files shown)


frontend/Dockerfile
```diff
@@ -26,8 +36,10 @@ USER node
 WORKDIR /srv/bt
 COPY --from=build --chown=node:node /srv/bt/ .
 
-ENV ENV=prod
 ENV NODE_ENV=production
 
+# TODO: remove
+RUN cat common/project.js
```
matthewelwell: This outputs the correct `project_selfhosted.js` file contents, but then when I run the container, I still get the staging API key being used for the flagsmith on flagsmith API requests.

matthewelwell: This was user error 🤦

kyle-ssg (approved):

PR #3026, opened 2023-11-23
feat: Import / export of features across environments and orgs

zachaysan:

## Changes

This PR allows someone using Flagsmith to export the information around their feature flags from a given environment in order to import them into another environment even at a different organisation or instance of flagsmith. Changes include:

1. An endpoint to create a new feature export. This takes in an `environment_id` and a list of `tag_ids` as part of a post body and kicks off a job to generate the export.
2. An endpoint to list the feature exports that have been successfully generated. For readability, the name of the export is the environment name then a timestamp to the minute. For example, `"Production | 2023-11-23 15:36 UTC"`.
3. An endpoint to download the export as a file via FeatureExport `id`.
4. An endpoint to upload a exported and downloaded file for a given environment with the option to specify one of two strategies for conflicting feature names. The first is called `SKIP` which skips any existing features that are named the same thing. The second is called `OVERWRITE` which discards an existing feature and related objects before creating a new one to load with.
5. A task that generates exports into a fairly flat file structure. The only nesting that was necessary was around multivariate options.
6. A task that processes imports abiding by the strategy specified above.
7. A task that cleans up stale exports after they've aged two weeks. 

Many individual tests for the views and the tasks. To test import and export tasks I elected to test them together to verify the process worked from download to upload. I also manually tested the file download endpoint to make sure the header was correctly set to initiate a download. 

I was unable to test the JavaScript friendly file upload API, so although I tried my best to follow guides, the implementer of the frontend (Novak) may want to have me on a video call when he's wiring that part up.


api/features/models.py
```diff
@@ -984,3 +993,52 @@ def get_update_log_message(self, history_instance) -> typing.Optional[str]:
...
+        swappable=True,
+    )
+
+    # This is a JSON string of data used for file download
+    # once the task has completed assembly.
+    data = models.CharField(max_length=MAX_FEATURE_EXPORT_SIZE)
+    created_date = models.DateTimeField("DateCreated", auto_now_add=True)
+
+
+class FeatureImport(models.Model):
```
zachaysan: I thought about adding a field to import to mark whether or not the import was successful but I was on the fence. Thoughts?

api/features/multivariate/models.py
```diff
@@ -31,6 +31,13 @@ class MultivariateFeatureOption(
     AbstractBaseExportableModel,
     abstract_base_auditable_model_factory(["uuid"]),
 ):
+    """
+    This class holds the *value* for a given multivariate feature
+    option. This value is the same for every environment, but the
+    percent allocation is set in MultivariateFeatureStateValue
+    which varies per-environment.
+    """
```
zachaysan, line 39: I decided to add this docstring when I was learning the codebase. I found it confusing that it held the value while the value class actually held the percent.

api/features/permissions.py
```diff
@@ -185,3 +186,46 @@ def has_permission(self, request, view):
             permission=MANAGE_SEGMENT_OVERRIDES,
             environment=environment,
         )
+
+
+class FeatureImportPermissions(IsAuthenticated):
+    def has_permission(
+        self, request: Request, view: "WrappedAPIView"  # noqa: F821
```
zachaysan: Here (and in a few other places) the `WrappedAPIView` is the actual type but it isn't importable from Django Rest Framework because they generate the type in a weird scope (in a decorator, iirc).

api/features/tasks.py
```diff
@@ -1,17 +1,125 @@
...
+        feature__is_archived=False,
+    )
+
+    if tag_ids:
+        additional_filters &= Q(feature__tags__in=tag_ids)
+
+    feature_states = FeatureState.get_environment_flags_list(
+        environment_id=environment_id,
+        additional_filters=additional_filters,
+    )
```
zachaysan: This is the primary method I relied upon to keep the feature states to export consistent with the rest of the application.

api/features/views.py
```diff
@@ -747,3 +764,94 @@ def create_segment_override(
...
+    serializer = FeatureImportSerializer(feature_import)
+    import_features_for_environment.delay(
+        kwargs={"feature_import_id": feature_import.id}
+    )
+    return Response(serializer.data, status=201)
+
+
+@api_view(["GET"])
+@permission_classes([DownloadFeatureExportPermissions])
+def download_feature_export(request: Request, feature_export_id: int) -> Response:
```
zachaysan: I wasn't really sure whether to use the `swagger_auto_schema` for this endpoint or even how to do it since there isn't a serializer involved.

api/features/tasks.py
zachaysan: Since I wrote this, it has become out of date since it was moved to the `versioning_service` module and is now just called `get_environment_flags_list` and with an `environment` argument instead of an `environment_id`

api/features/migrations/0064_auto_20231121_2028.py
matthewelwell, line 1: Sorry, an (admittedly undocumented) rule is that migrations should always have user defined names (`python manage.py makemigrations app_name --name some_friendly_migration_name`).

I also think you could go ahead and delete these 2 migrations and combine them into a new, single migration.

api/features/models.py
```diff
@@ -970,3 +979,52 @@ def get_update_log_message(self, history_instance) -> typing.Optional[str]:
 
     def _get_environment(self) -> typing.Optional["Environment"]:
         return self.feature_state.environment
+
+
+class FeatureExport(models.Model):
```
matthewelwell: 2 things here: 

 - I'd perhaps argue this should be an EnvironmentExport? 
 - I'd rather these models live in either `environments/import_export` (if we decide this is an _environment_ export rather than features, `features/import_export` or the existing `import_export` app.

api/features/models.py
```diff
@@ -970,3 +979,52 @@ def get_update_log_message(self, history_instance) -> typing.Optional[str]:
...
+        "environments.Environment",
+        related_name="feature_imports",
+        on_delete=models.CASCADE,
+        swappable=True,
+    )
+
+    # This is a JSON string of data used for file download
+    # once the task has completed assembly.
+    data = models.CharField(max_length=MAX_FEATURE_EXPORT_SIZE)
+    created_date = models.DateTimeField("DateCreated", auto_now_add=True)
```
matthewelwell: Ah, I expect you got this from another `created_date` in the application, but I really dislike renaming the column to `DateCreated`. Can we use just `created_at` here as this is the new standard.

api/features/models.py
matthewelwell: I think that would be useful, yes. I wonder if we could capture any logs if it's not successful and store them here (see `task_processor.TaskRun` for example).

api/features/models.py
```diff
@@ -970,3 +979,52 @@ def get_update_log_message(self, history_instance) -> typing.Optional[str]:
...
+    strategy = models.CharField(
+        choices=FEATURE_IMPORT_STRATEGIES,
+        max_length=50,
+        blank=False,
+        null=False,
+    )
+
+    # This is a JSON string of data generated by the export.
+    data = models.CharField(max_length=MAX_FEATURE_IMPORT_SIZE)
+    created_date = models.DateTimeField("DateCreated", auto_now_add=True)
```
matthewelwell: As above, I'd rather this was just `created_at = models.DateTimeField(auto_now_add=True)`

api/features/permissions.py
```diff
@@ -185,3 +186,46 @@ def has_permission(self, request, view):
...
+        return request.user.is_environment_admin(feature_export.environment)
+
+
+class FeatureExportListPermissions(IsAuthenticated):
+    def has_permission(self, request: Request, view: ListAPIView) -> bool:
+        if not super().has_permission(request, view):
+            return False
+
+        project = Project.objects.get(id=view.kwargs["project_id"])
+        return request.user.has_project_permission(VIEW_PROJECT, project)
```
matthewelwell: I feel like perhaps we should document why this is `VIEW_PROJECT` while the others require environment admin.

api/features/tasks.py
```diff
@@ -1,17 +1,127 @@
...
+                ).first()
+            )
+
+            multivariate.append(
+                {
+                    "percentage_allocation": mv_feature_state_value.percentage_allocation,
+                    "default_percentage_allocation": mv_option.default_percentage_allocation,
+                    "value": mv_option.value,
+                }
+            )
```
matthewelwell: Couple of comments here: 

1. Couldn't we do something more like: 

```python
multivariate = [
    {
        "percentage_allocation": mv_fsv.percentage_allocation,
        "default_percentage_allocation": mv_fsv.multivariate_option.default_percentage_allocation,
        "value": mv_fsv.multivariate_option.value
    } for mv_fsv in feature_state.multivariate_feature_state_values.select_related("multivariate_option").all()
]
```

2. Without having seen the import code yet, I'm a little unsure why we'd store the value as a single field rather than the strictly typed fields that exist on the original model. Feels like it just adds extra typing complexities on the import.

api/features/tasks.py
```diff
@@ -1,17 +1,127 @@
...
+
+        payload.append(
+            {
+                "name": feature_state.feature.name,
+                "default_enabled": feature_state.feature.default_enabled,
+                "is_server_key_only": feature_state.feature.is_server_key_only,
+                "initial_value": feature_state.feature.initial_value,
+                "value": feature_state.feature_state_value.value,
```
matthewelwell: As above, would it be better to pass over the strictly typed fields?

api/features/tasks.py
```diff
@@ -1,17 +1,127 @@
...
+
+
+@register_task_handler()
+def import_features_for_environment(feature_import_id: int) -> None:
+    feature_import = FeatureImport.objects.get(id=feature_import_id)
+    environment = feature_import.environment
+    input_data = json.loads(feature_import.data)
+    project = environment.project
+
+    for rec in input_data:
```
matthewelwell: I'm not sure what `rec` means here?

api/features/tasks.py
```diff
@@ -82,3 +192,82 @@ def _get_feature_state_webhook_data(feature_state, previous=False):
...
+        feature_state_value.type = BOOLEAN
+        feature_state_value.boolean_value = value
+    else:
+        feature_state_value.type = STRING
+        feature_state_value.string_value = value
+
+    feature_state_value.save()
+
+
+def _create_multivariate_feature_option(
```
matthewelwell: I think this could be simplified by doing something like: 

```python
mvfo = MultivariateFeatureOption(feature=feature, default_percentage_allocation=default_percentage_allocation)
if isinstance(value, int):
    mvfo.integer_value = value
    mvfo.type = INTEGER
...
```

api/features/tasks.py
```diff
@@ -82,3 +192,82 @@ def _get_feature_state_webhook_data(feature_state, previous=False):
...
+    feature = Feature.objects.create(
+        name=rec["name"],
+        project=project,
+        initial_value=rec["initial_value"],
+        is_server_key_only=rec["is_server_key_only"],
+        default_enabled=rec["default_enabled"],
+    )
+    feature_state = feature.feature_states.filter(
+        environment=environment,
+    ).first()
```
matthewelwell: It's a little odd to me that we're using `.first()` here - I think we could probably just use `.get()`, right?

api/features/views.py
```diff
@@ -752,3 +769,94 @@ def create_segment_override(
...
+        }
+    )
+
+    return Response(serializer.validated_data, status=201)
+
+
+@swagger_auto_schema(
+    method="POST",
+    responses={201: FeatureImportSerializer()},
+)
```
matthewelwell: Can we use a serializer with a `serializers.FileField()` here?

api/features/views.py
```diff
@@ -752,3 +769,94 @@ def create_segment_override(
...
+
+    strategy = request.query_params.get("strategy").upper()
+
+    if not strategy:
+        return Response(
+            {"detail": "A feature import strategy query param must be supplied."},
+            status=status.HTTP_400_BAD_REQUEST,
+        )
+
+    file_content = uploaded_file.read().decode("utf-8")
```
matthewelwell: Do we want to do any validation of the file content (maybe a second PR)?

api/features/views.py
matthewelwell: Could we make use of `serializers.FileField` here again?

matthewelwell (changes requested):
For now, I've been through everything up to the tests.

api/features/models.py
zachaysan: Yes that's what I did. I strongly prefer `created_at` too. I will update it.

api/features/models.py
zachaysan: Are we thinking about exposing information to the user that imports the task? Or if we're just storing the the traceback information for our own information is there additional advantage to duplicating it out of the existing task?

zachaysan: Talking offline, go with a status field but leave the validation stuff off for now, we can always add it later.

api/features/views.py
zachaysan: Offline talked about options:
1. Look into FileField.
2. See with docstring on view to propagate the information to the api docs.

khvn26:
Should we make LD imports rely on mechanisms introduced here? We could probably make `launch_darkly/services.py` generate an export file rather than objects directly.

api/features/views.py
zachaysan: Resolved offline: First pass no validation.

api/features/migrations/0064_auto_20231121_2028.py
zachaysan: Done with the initial migration for the new app.

api/features/permissions.py
zachaysan: Sounds good, I've added this.

api/features/tasks.py
zachaysan: 1. I was a bit confused with the first point since it overlaps with a different class, but then I realized you meant `select_related("multivariate_feature_option")` in which case yes this works! Updated locally.

2. I'm adding typing to the export as a precaution since it is more robust that way.

api/features/tasks.py
zachaysan: Talked offline: rec means record, changing to another variable

api/features/tasks.py
zachaysan: Cool. I did something kinda similar where we load the type from the now saved type in the data, with an assert at the end to verify that the type is one of the three known types.

api/features/tasks.py
zachaysan: Yeah that makes sense.

api/features/views.py
zachaysan: I was able to get this working with the import, but I'm less sure of the export. Either way, a small win.

api/features/views.py
zachaysan: Ok I went with as much swagger as I could find from their docs. I wrote it up like this:
```python
@swagger_auto_schema(
    method="GET",
    responses={200: "File downloaded"},
    operation_description="This endpoint is to download a feature export file from a specific environment",
)
```

zachaysan:
> Should we make LD imports rely on mechanisms introduced here? We could probably make `launch_darkly/services.py` generate an export file rather than objects directly.

Once we're happy with the import mechanism with this working I see no harm in using it for other projects like the LD import if it makes the right sense.

api/features/import_export/migrations/0001_initial.old
matthewelwell, line 1: I'm not too sure why this file still exists?

api/features/import_export/views.py
```diff
@@ -0,0 +1,122 @@
...
+    responses={201: FeatureImportSerializer()},
+)
+@api_view(["POST"])
+@permission_classes([FeatureImportPermissions])
+def feature_import(request: Request, environment_id: int) -> Response:
+    serializer = FeatureImportUploadSerializer(data=request.data)
+    serializer.is_valid(raise_exception=True)
+    uploaded_file = serializer.validated_data["file"]
+
+    strategy = request.query_params.get("strategy").replace("-", "_").upper()
```
matthewelwell: I assume this was moved to a query parameter because of some funniness around including the file field? It certainly feels a bit strange having this as a query parameter though.

api/tests/unit/features/import_export/test_unit_features_import_export_tasks.py
```diff
@@ -0,0 +1,308 @@
...
+    kept_feature_import = FeatureImport.objects.create(
+        data="{}",
+        environment=environment,
+    )
+    lost_feature_import = FeatureImport.objects.create(
+        data="{}",
+        environment=environment,
+    )
+    lost_feature_import.created_at = four_weeks_ago
+    lost_feature_import.save()
```
matthewelwell: I'd love to see this use `freezegun` rather than hacking the `created_at` attribute (just from an aesthetics point of view).

api/tests/unit/features/import_export/test_unit_features_import_export_tasks.py
```diff
@@ -0,0 +1,308 @@
...
+    db: None,
+    environment: Environment,
+    project: Project,
+    identity: Identity,
+    feature: Feature,
+    feature_segment: FeatureSegment,
+) -> None:
+    # Given
+    # This environment should be ignored.
+    Environment.objects.create(name="Dig", project=project)
```
matthewelwell: We don't seem to use this at all, I'm not sure what it's purpose is?

api/tests/unit/features/import_export/test_unit_features_import_export_tasks.py
```diff
@@ -0,0 +1,308 @@
...
+    assert new_feature1.is_server_key_only is True
+    assert new_feature1.default_enabled is True
+
+    assert new_feature2.type == STANDARD
+    assert new_feature2.initial_value == "banana"
+    assert new_feature2.is_server_key_only is False
+    assert new_feature2.default_enabled is False
+
+
+def test_export_and_import_features_for_environment(
```
matthewelwell: Should the test name here include the strategy again?

api/features/import_export/migrations/0001_initial.old
zachaysan: Accidentally included. Will remove.

api/features/import_export/views.py
zachaysan: The original implementation didn't use the file field serializer, but now this was just an oversight. I'm switching to a serializer here.

api/tests/unit/features/import_export/test_unit_features_import_export_tasks.py
zachaysan: No problem.

api/tests/unit/features/import_export/test_unit_features_import_export_tasks.py
zachaysan: I wanted to be sure that the generated features were isolated to a single environment, but if it doesn't really seem useful I'll just remove it.

api/tests/unit/features/import_export/test_unit_features_import_export_tasks.py
zachaysan: Sure that's no problem

api/features/import_export/serializers.py
```diff
@@ -0,0 +1,51 @@
...
+            "created_at",
+        )
+
+    def get_name(self, obj: FeatureExport) -> str:
+        return (
+            f"{obj.environment.name} | {obj.created_at.strftime('%Y-%m-%d %H:%M')} UTC"
+        )
+
+
+def validate_feature_import_file_size(upload_file):
```
matthewelwell: Can we add typing to the `upload_file` argument?

api/features/import_export/views.py
```diff
@@ -0,0 +1,114 @@
...
+def create_feature_export(request: Request) -> Response:
+    serializer = CreateFeatureExportSerializer(data=request.data)
+    serializer.is_valid(raise_exception=True)
+
+    export_features_for_environment.delay(
+        kwargs={
+            "environment_id": serializer.validated_data["environment_id"],
+            "tag_ids": serializer.validated_data["tag_ids"],
+        }
+    )
```
matthewelwell: I guess there's an argument to be said that this logic should live in the `CreateFeatureExportSerializer.save` method, what do you think?

api/features/import_export/serializers.py
zachaysan: Thanks for the catch! Done!

api/features/import_export/views.py
zachaysan: Sure I can move it there. I'm a little on the fence but I could see it work either way.

api/features/import_export/views.py
```diff
@@ -0,0 +1,105 @@
...
+    feature_import = FeatureImport.objects.create(
+        environment_id=environment_id,
+        strategy=strategy,
+        data=file_content,
+    )
+    serializer = FeatureImportSerializer(feature_import)
+    import_features_for_environment.delay(
+        kwargs={"feature_import_id": feature_import.id}
+    )
+    return Response(serializer.data, status=201)
```
matthewelwell: Sorry, I should have been explicit, I think I'd like to see this done in the `serializer.save()` method too, what do you think? I do think the other view looks cleaner now.

api/features/import_export/views.py
```diff
@@ -0,0 +1,105 @@
...
+    serializer = CreateFeatureExportSerializer(data=request.data)
+    serializer.is_valid(raise_exception=True)
+    serializer.save()
+
+    return Response(serializer.validated_data, status=201)
+
+
+@swagger_auto_schema(
+    method="POST",
+    request_body=FeatureImportUploadSerializer,
```
matthewelwell:
```suggestion
    request_body=FeatureImportUploadSerializer(),
```

zachaysan: Thank you! Done!

api/features/import_export/views.py
zachaysan: Ok this has been added as well.

api/features/import_export/views.py
```diff
@@ -0,0 +1,93 @@
...
+@swagger_auto_schema(
+    method="POST",
+    request_body=FeatureImportUploadSerializer(),
+    responses={201: FeatureImportSerializer()},
+)
+@api_view(["POST"])
+@permission_classes([FeatureImportPermissions])
+def feature_import(request: Request, environment_id: int) -> Response:
+    upload_serializer = FeatureImportUploadSerializer(data=request.data)
+    upload_serializer.is_valid(raise_exception=True)
```
matthewelwell, line 51: @zachaysan I moved this check here to follow standard DRF patterns.

api/features/import_export/views.py
```diff
@@ -0,0 +1,93 @@
...
+    _feature_import = upload_serializer.save(environment_id=environment_id)
```
matthewelwell, line 52: I changed the name of this variable to stop my IDE complaining about it shadowing the name from the outer scope.

matthewelwell (approved):

api/features/import_export/views.py
matthewelwell: Thanks!

api/features/import_export/views.py
zachaysan: Ah. Okay yeah that makes more sense.

PR #3028, opened 2023-11-23
feat: Rebuild chargebee caches

zachaysan:

## Changes

Added another chargebee webhook handler function to filter events that may lead to us needing to update the chargebee cache.

Played around with the code by hand to make sure I understood the chargebee API plus some work on their test interface UI to get the webhook information then a single test based off of typical event. It's not complicated code so I'm not worried.


api/tests/unit/organisations/test_unit_organisation_views.py
```diff
@@ -1497,9 +1497,49 @@ def test_payment_failed_chargebee_webhook_no_subscription_id(
...
+    url = reverse("api-v1:chargebee-webhook")
+    task = mocker.patch("organisations.chargebee.tasks.ChargebeeCache")
+
+    # When
+    response = staff_client.post(
+        url, data=json.dumps(data), content_type="application/json"
+    )
+    # Then
+    assert response.status_code == 200
+    task.assert_called_once()
```
matthewelwell: I'd rather be more specific here - wouldn't it be simpler (and more specific) to just mock the task function in the webhook_handlers module and then call like `mocked_task.delay.assert_called_once_with()` ? 

I _think_ the current implementation also relies on `settings.TASK_RUN_METHOD == "SYNCHRONOUSLY"` whereas the above approach wouldn't?

api/organisations/views.py
```diff
@@ -274,10 +274,12 @@ def chargebee_webhook(request: Request) -> Response:
...
     event_type = request.data.get("event_type")
 
-    if event_type == "payment_failed":
+    if event_type == webhook_event_types.PAYMENT_FAILED:
         return webhook_handlers.payment_failed(request)
-    if event_type == "payment_succeeded":
+    if event_type == webhook_event_types.PAYMENT_SUCCEEDED:
         return webhook_handlers.payment_succeeded(request)
+    if event_type in webhook_event_types.CACHE_REBUILD_TYPES:
+        return webhook_handlers.cache_rebuild_event(request)
```
matthewelwell, line 282: I really wish there was a better way to do this instead of just a (probably eventually) large conditional statement. One other option would be to use a `match` / `case` but it seems there's no easy way to make it play nice with checking if an item is in a list. 

From a personal preference point of view, unless there's a reason not to, I'd prefer this to be written using `if` / `elif` just as it makes it slightly more obvious that it can only match one condition (although I guess that's the case given the fact that we're returning in each case anyway).

api/tests/unit/organisations/test_unit_organisation_views.py
```diff
@@ -1497,9 +1497,49 @@ def test_payment_failed_chargebee_webhook_no_subscription_id(
...
+                "description": "This is addon added when support is needed",
+                "type": "quantity",
+                "charge_type": "recurring",
+                "price": 1000,
+                "period": 1,
+                "period_unit": "month",
+                "status": "deleted",
+            }
+        },
+        "event_type": "addon_deleted",
```
matthewelwell: Might be overkill (but it would be super trivial to do I think) but should we parametrise this test with all of the cache rebuild event types?

matthewelwell (changes requested):
Looks good on the whole. I've left a few comments but only one that I think should definitely change.

api/tests/unit/organisations/test_unit_organisation_views.py
zachaysan: Well, I tried to place the webhook as realistic as possible without going over. To parametrize it would go against the idea of a realistic test. I think it's better to just leave a single example, in my opinion, but not a big deal if you want to enumerate the options either.

api/organisations/views.py
zachaysan: I've been raised as an early exit from an `if` means that the next exit starts as an `if` not an `elif`. I can't remember which linter enforces this, but there was one in the past. 

As for this turning into a large conditional statement, provided once we refactor this block of code I hope that every statement looks like this:

```python
if conditional1:
    return func_1()
if conditional2:
    return func_2()
if conditional3:  # And so on.
    return func_3()

return None
```

If the if checks are lean then the person reviewing the code can quickly understand what happens and in which order of preference. It's when `if` statements get subnested and variables get assigned from one sub `if` statement to another that it gets hard to reason about. At the end of the day, we receive all their webhooks through a single view so we have to filter them somehow. If we try to create a layer of indirection to, say, register functions based on the `event_type` of the post body then it may look more pleasing, but it's overhead that's doing the same thing. I think we should keep it simple.

api/tests/unit/organisations/test_unit_organisation_views.py
zachaysan: Sure I can switch it to that. I like to call tasks but if you don't think there is value here I can mock the task.

zachaysan: Ok this has been updated :)

api/organisations/views.py
matthewelwell: Ok, agreed, thanks for the comprehensive response!

api/tests/unit/organisations/test_unit_organisation_views.py
matthewelwell: Yep, I do tend to agree. Let's leave it as is for now.

api/tests/unit/organisations/test_unit_organisations_views.py
```diff
@@ -1497,9 +1497,50 @@ def test_payment_failed_chargebee_webhook_no_subscription_id(
...
+    task = mocker.patch(
+        "organisations.chargebee.webhook_handlers.update_chargebee_cache"
+    )
+    # When
+    response = staff_client.post(
+        url, data=json.dumps(data), content_type="application/json"
+    )
+    # Then
+    assert response.status_code == 200
+    task.delay.assert_called_once()
```
matthewelwell: I think I'd prefer to see this as `task.delay.assert_called_once_with()` to prove that we're calling it with no arguments, but it's not super important.

matthewelwell (approved):
Approved with one minor comment. Feel free to merge before / after implementing the suggested change :)

api/tests/unit/organisations/test_unit_organisations_views.py
zachaysan: Sure that's no problem :)

PR #3031, opened 2023-11-24
feat(auth): integrate ldap

matthewelwell:
## Changes

Adds the flagsmith-ldap module as a dependency to the main Flagsmith application. This will be installed as a dependency as part of the flagsmith-private-cloud image build workflow. 

Tests are run in flagsmith/flagsmith-ldap repository.


.github/workflows/platform-docker-publish-all-features-image.yml
```diff
@@ -120,7 +120,7 @@ jobs:
           context: .
           build-args: |
             SAML_INSTALLED=1
-            POETRY_OPTS=--with saml,auth-controller
+            POETRY_OPTS=--with saml,auth-controller,ldap
```
khvn26, line 124: Since we're resorting to a private git dependency, this step will fail without adding credentials to the git client.

See [one of the solutions](https://github.com/orgs/python-poetry/discussions/3794#discussioncomment-2469059).

matthewelwell: Hmm... I was assuming that the presence of the github token secret variable would suffice here but perhaps you're right. It's working currently where we manually checkout the packages but I guess that's because we're using a pre-built action. I'll have a play with this...

khvn26: We are providing a distinct token to the checkout action for the private repos, aren't we?

matthewelwell: Yes, you're right, we are. Well caught, thanks!

matthewelwell: See successful build [here](https://github.com/Flagsmith/flagsmith/actions/runs/7036478266/job/19149174412?pr=3031).

.github/workflows/api-pull-request.yml
```diff
@@ -43,6 +43,7 @@ jobs:
 
       - name: Install poetry
         run: pipx install poetry
+
```
matthewelwell, line 46: Sorry, irrelevant to this PR but it was bothering me.

khvn26 (approved):

PR #3040, opened 2023-11-27
fix: resolve environment N+1 caused by feature versioning v2

matthewelwell:
## Changes

Caused by the addition of [this line](https://github.com/Flagsmith/flagsmith/blob/main/api/features/models.py#L474) in the feature versioning V2, this PR resolves an N+1 query issue retrieving the environment. 

Note, this adds a test to the old location to keep things grouped for now but should be moved to the new location asap. 

Added a test to reproduce the N+1 error before resolving.

### DIFF
--- a/api/environments/identities/models.py
+++ b/api/environments/identities/models.py
@@ -97,6 +97,7 @@ class Identity(models.Model):
             full_query &= additional_filters
 
         select_related_args = [
+            "environment",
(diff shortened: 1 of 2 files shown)


khvn26 (approved):

zachaysan (approved):
Awesome fix!

PR #3042, opened 2023-11-27
feat: Add role api keys in UI

novakzaballa:

## Changes

- Add the ability in the frontend to create role API keys

- Go to manage -> Keys
- Click the "Create API Key" button
- Disable the "Is admin" switch and add roles


api/api_keys/models.py
```diff
@@ -18,3 +20,13 @@ class MasterAPIKey(AbstractAPIKey, SoftDeleteObject):
 
     objects = MasterAPIKeyManager()
     is_admin = models.BooleanField(default=True)
+
+    if settings.IS_RBAC_INSTALLED:
```
matthewelwell: This should be inside the method itself.

api/api_keys/serializers.py
```diff
@@ -11,17 +14,21 @@ class MasterAPIKeySerializer(serializers.ModelSerializer):
         "for every endpoint apart from create",
     )
     is_admin = serializers.BooleanField(default=True)
+    if settings.IS_RBAC_INSTALLED:
+        roles = MasterAPIKeyRoleSerializer(many=True, read_only=True, source="role")
```
matthewelwell: Hmm... I don't really like this behaviour of changing serializers based on the existence of external libraries. Can we think of another way to do this? Perhaps, as per the suggestion on the other PR, by creating the view in the rbac repo and adding it to URLs in the core repo. We do this for a number of views in the flagsmith-workflows repo for example.

api/api_keys/serializers.py
```diff
@@ -36,3 +43,21 @@ def validate_is_admin(self, is_admin: bool):
...
             )
         return is_admin
+
+    def to_representation(self, instance):
+        if self.context["view"].action == "retrieve":
+            is_admin = instance.is_admin if hasattr(instance, "is_admin") else False
+            if is_admin is False and not settings.IS_RBAC_INSTALLED:
+                raise serializers.ValidationError(
+                    "RBAC is not installed, cannot create non-admin key"
+                )
```
matthewelwell: This can just be done in a `validate_is_admin` method on the serializer.

matthewelwell: Also, I'm confused why we're validating this on retrieve, shouldn't this only need to be validated on create?

api/api_keys/serializers.py
novakzaballa: I saw that you had a similar comment in another PR, yes I think you are right, the serializers would no longer depend on RBAC and could be tested more easily, I will do it.

novakzaballa: The endpoint was created in RBAC: https://github.com/Flagsmith/flagsmith-rbac/pull/14

api/api_keys/models.py
novakzaballa: Moved.

kyle-ssg (approved):
FE looks fine from a code perspective 👍

api/api_keys/models.py
```diff
@@ -18,3 +20,12 @@ class MasterAPIKey(AbstractAPIKey, SoftDeleteObject):
...
     is_admin = models.BooleanField(default=True)
+
+    @hook(BEFORE_UPDATE, when="is_admin", was=False, is_now=True)
+    def delete_role_api_keys(
+        self,
+    ):
+        if settings.IS_RBAC_INSTALLED:
+            from rbac.models import MasterAPIKeyRole
+
+            MasterAPIKeyRole.objects.filter(master_api_key=self.id).delete()
```
matthewelwell, line 31: Possibly a better solution here (to keep the code related to RBAC out of the core repo as much as possible) would be to use a signal in the flagsmith-rbac repository, but for now this approach is fine.

api/api_keys/serializers.py
matthewelwell: So can we remove it from here now?

api/api_keys/serializers.py
```diff
@@ -11,17 +14,21 @@ class MasterAPIKeySerializer(serializers.ModelSerializer):
...
         fields = (
+            "id",
             "prefix",
             "created",
             "name",
             "revoked",
             "expiry_date",
             "key",
             "is_admin",
+            *([] if not settings.IS_RBAC_INSTALLED else ["roles"]),
```
matthewelwell: This should be removed too I think?

matthewelwell (changes requested):

api/api_keys/serializers.py
novakzaballa: Both were removed.

api/api_keys/serializers.py
novakzaballa: Good catch @matthewelwell, this code is unnecessary, deleted.

I also found a bug, it was already solved, I had to create a PR for the RBAC: https://github.com/Flagsmith/flagsmith-rbac/pull/19

matthewelwell:
@novakzaballa I've just tested this but it seems as though the roles aren't getting applied to the keys correctly. 

https://github.com/Flagsmith/flagsmith/assets/14089968/82ca8c8a-c75e-4aef-8235-279d7299382e

matthewelwell (commented):
Code looks ok, but as per previous comment, it doesn't appear to be working. Mostly leaving this review comment to remove this PR from my list because GH doesn't seem to have a function to do it another way.

novakzaballa:
> @novakzaballa I've just tested this but it seems as though the roles aren't getting applied to the keys correctly.
> 

@matthewelwell I found the issue, and now is solved, I had to split the backend: https://github.com/Flagsmith/flagsmith/pull/3346

matthewelwell:
Yep, seems like it works now which is great! A couple of minor things: 

1. This shouldn't say 'Terraform API Keys' any more.  
Screenshot 2024-02-15 at 11 45 16

2. Can we add an item to the table indicating whether a key is an admin or not?
3. We need to add / update the docs. Currently they're linking to the Terraform documentation which is not correct. Perhaps we should create a separate PR for the docs?

novakzaballa:
> Yep, seems like it works now which is great! A couple of minor things:
> 
> 1. This shouldn't say 'Terraform API Keys' any more.
> 
> 2. Can we add an item to the table indicating whether a key is an admin or not?
> 3. We need to add / update the docs. Currently they're linking to the Terraform documentation which is not correct. Perhaps we should create a separate PR for the docs?


Hey @matthewelwell, I changed the title and added a field in the API keys table to see if it is an admin API key or not.

I will create an issue for the API keys documentation.

matthewelwell (approved):
This looks good now, but one improvement that I'd like to see (which can maybe be done in a separate PR) is some kind of indication that a key is active / expired. I've added a `has_expired` parameter to the API [here](https://github.com/Flagsmith/flagsmith/pull/3432) I think we should just use that and display it in the table for now?

PR #3047, opened 2023-11-27
fix: Refactor existing Chargebee webhooks for subscriptions

zachaysan:

## Changes

Taking direction from earlier work, I clustered these functionalities together with the other pieces of isolated webhook modules. The basic patterns are the same, although this instance the serializers were more involved. I took a look at the Chargebee webhooks response document and tried to match things up as best as I could. 

The existing test base has good coverage of this part. I was able to add alterations to the test base where deviations had come into play based off of what is in the serializers. For example, the chargebee unixtimestamp is in integer format, not floats, so these and other things were needed to be fixed.


api/tests/unit/organisations/test_unit_organisations_views.py
```diff
@@ -662,12 +668,13 @@ def test_when_subscription_is_cancelled_then_cancellation_date_set_and_alert_sen
     ):
         # Given
         cancellation_date = datetime.now(tz=UTC) + timedelta(days=1)
+        current_term_end = int(datetime.timestamp(cancellation_date))
```
matthewelwell, line 671: I'm curious as to why this had to change as part of this PR?

matthewelwell (approved):
Approved with one minor comment question.

api/tests/unit/organisations/test_unit_organisations_views.py
zachaysan: Because the Chargebee API sends unix time in integer seconds, not milliseconds or microseconds. So to go through the API we need to fix the integer resolution to get through the serializer.

PR #3050, opened 2023-11-28
feat(sse): track usage 

gagantrivedi:

## Changes

Add a recurring task to read sse access logs from s3 and upload that data to influx

By Locally running task processor (also includes unit tests cases)


api/sse/tasks.py
```diff
@@ -27,6 +39,37 @@ def send_environment_update_message(environment_key: str, updated_at):
...
+    )
+    def update_sse_usage():
+        with influxdb_client.write_api(
+            write_options=WriteOptions(batch_size=1000, flush_interval=2000)
+        ) as write_api:
+            for log in sse_service.stream_access_logs():
+                environment = Environment.get_from_cache(log.api_key)
+
+                if not environment:
+                    logger.warning("Invalid  api_key %s", log.api_key)
```
zachaysan: Minor nit.
```suggestion
                    logger.warning("Invalid api_key %s", log.api_key)
```

api/tests/unit/sse/test_sse_service.py
```diff
@@ -96,3 +103,63 @@ def test_send_environment_update_message_for_environment_schedules_task_correctl
             realtime_enabled_project_environment_one.updated_at.isoformat(),
         )
     )
+
+
+@mock_s3
+def test_stream_access_logs(mocker: MockerFixture, settings):
```
zachaysan:
```python

@mock_s3
def test_stream_access_logs(mocker: MockerFixture, settings: SettingsWrapper) -> None:
```

api/tests/unit/sse/test_sse_service.py
```diff
@@ -96,3 +103,63 @@ def test_send_environment_update_message_for_environment_schedules_task_correctl
...
+
+    mocked_gpg.return_value.decrypt.side_effect = [
+        mocker.MagicMock(data=first_decrypted_object_data),
+        mocker.MagicMock(data=second_decrypted_object_data),
+    ]
+
+    # When
+    access_logs = list(stream_access_logs())
+
+    # The
```
zachaysan:
```suggestion
    # Then
```

api/tests/unit/sse/test_sse_service.py
```diff
@@ -96,3 +103,63 @@ def test_send_environment_update_message_for_environment_schedules_task_correctl
...
+        f"{first_log.generated_at},{first_log.api_key}\n"
+        f"{second_log.generated_at},{second_log.api_key}".encode()
+    )
+    second_encrypted_object_data = b"second_bucket_encrypted_data"
+    second_decrypted_object_data = (
+        f"{third_log.generated_at},{third_log.api_key}".encode()
+    )
+
+    # patch the s3 resource because it was created before the mock_s3 decorator was applied
+    patch_resource(s3)
```
zachaysan: I don't 100% follow what patch_resource does. Does it re-enable S3 for the test?

zachaysan (approved):
Looks overall good. Just some small questions and nits.

api/sse/tasks.py
```diff
@@ -27,6 +39,37 @@ def send_environment_update_message(environment_key: str, updated_at):
...
+
+    @register_recurring_task(
+        run_every=timedelta(seconds=60),
+    )
+    def update_sse_usage():
+        with influxdb_client.write_api(
+            write_options=WriteOptions(batch_size=1000, flush_interval=2000)
+        ) as write_api:
+            for log in sse_service.stream_access_logs():
+                environment = Environment.get_from_cache(log.api_key)
```
matthewelwell: I wonder if we'd be better off just grabbing all the (active) environments from the DB and storing in a dictionary in this task? Maybe that's a memory concern though.

api/sse/tasks.py
```diff
@@ -27,6 +39,37 @@ def send_environment_update_message(environment_key: str, updated_at):
...
+                    continue
+
+                record = _get_influx_point(environment, log.generated_at)
+                write_api.write(bucket=settings.SSE_INFLUXDB_BUCKET, record=record)
+
+
+def _get_influx_point(environment: Environment, event_time: str) -> Point:
+    return (
+        Point("sse_call")
+        .field("request_count", 1)
```
matthewelwell: I think we can aggregate here? Why send a request to influx for each data point?

infrastructure/aws/staging/ecs-task-definition-task-processor.json
```diff
@@ -114,6 +114,22 @@
...
+                    "name": "INFLUXDB_ORG",
+                    "value": "ben.rometsch@bullet-train.io"
+                },
+                {
+                    "name": "INFLUXDB_URL",
+                    "value": "https://eu-central-1-1.aws.cloud2.influxdata.com"
+                },
+                {
+                    "name": "SSE_INFLUXDB_BUCKET",
+                    "value": "sse_dev"
```
matthewelwell: This is kinda confusing, can we use sse_staging?

matthewelwell (commented):
Looks great on the whole. Added a few minor comments, the only one that will require some proper thought perhaps is the aggregation of data before sending to influx.

infrastructure/aws/staging/ecs-task-definition-task-processor.json
gagantrivedi: Done

api/sse/tasks.py
gagantrivedi: Yeah, good shout(I did not do it initially because we loose granularity, but I guess it makes sense)

api/sse/tasks.py
gagantrivedi: It kinda still does the same thing, right? Because `get_from_cache` uses local memory?

matthewelwell: Yeah, I guess that's fine (although it still might result in lots of queries if the environments aren't already in the cache). The other consideration is that the query used by `get_from_cache` pulls back a lot of data (via select_related) that I don't think we need. It also checks the bad environments cache which I don't think should be relevant here.

api/sse/tasks.py
matthewelwell: Yeah, interesting - I guess we've lost the granularity on the timestamp. I'm keen to understand though, how granular the log files are in the first place?

api/sse/tasks.py
gagantrivedi: I have reduced the query count to 1 here, but I have removed the warning for invalid keys, do we care about the warning?

api/sse/tasks.py
gagantrivedi: The file is simple access log, i.e: one line per request

matthewelwell: But how often do the log files roll over? Isn't that therefore indicative of how granular our data will be?

gagantrivedi: That can be configured in fastly... I am planning to start with 1 minutes(but if the files generated are small enough we can increase that)

api/sse/tasks.py
matthewelwell: I don't think so?

api/sse/tasks.py
matthewelwell: I'm thinking e.g. 15 minute granularity would be good. Is that possible?

matthewelwell (approved):

PR #3057, opened 2023-11-28
docs(JIRA): fix JIRA marketplace link

matthewelwell:
## Changes

Fixes the JIRA marketplace link which was previously pointing at the old dev app. 

Ran the docs locally and verified the link is correct.


dabeeeenster (approved):

dabeeeenster:
Good catch

PR #3058, opened 2023-11-28
fix: Manage members layout is broken

### LINKED ISSUE #3001: Manage members layout is broken
What I see:

image

What the designs show:

image

novakzaballa:

## Changes

- Manage members layout is fixed

Before:
Screen Shot 2023-11-28 at 14 53 43

Now:
Screen Shot 2023-11-28 at 14 45 48

- Go to manage -> Members -> Members

### DIFF
--- a/frontend/web/components/pages/OrganisationSettingsPage.js
+++ b/frontend/web/components/pages/OrganisationSettingsPage.js
@@ -26,7 +26,7 @@ import PageTitle from 'components/PageTitle'
 import { getStore } from 'common/store'
 import { getRoles } from 'common/services/useRole'
 
-const widths = [450, 150, 100, 250, 235, 255]
+const widths = [450, 255, 250, 235, 150, 100]
 const rolesWidths = [250, 600, 100]
 const OrganisationSettingsPage = class extends Component {
   static contextTypes = {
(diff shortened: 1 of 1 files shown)


frontend/web/components/pages/OrganisationSettingsPage.js
```diff
@@ -26,7 +26,7 @@ import PageTitle from 'components/PageTitle'
 import { getStore } from 'common/store'
 import { getRoles } from 'common/services/useRole'
 
-const widths = [450, 150, 100]
+const widths = [450, 150, 100, 250, 235, 255]
```
matthewelwell: I'd quite like to understand why we define these widths here rather than just directly in the style objects below. 

It also would make sense to my somewhat picky brain if these were ordered, either smallest to largest or largest to smallest, I'm not _that_ fussy!

matthewelwell:
I have tested this via the preview and it looks good to me, but I've added a comment on the implementation.

frontend/web/components/pages/OrganisationSettingsPage.js
novakzaballa: It seems that initially this array was used to give sizes to the panelSearch component, as I worked on that component I added those values ​​to the array.

Re-ordered.

matthewelwell (approved):

PR #3064, opened 2023-11-29
feat: add new endpoint to list summary objects of permission groups

matthewelwell:
## Changes

Adds a new endpoint to list a summary object for all user groups in an organisation. 

Added unit tests.


api/users/views.py
```diff
@@ -173,14 +173,16 @@ def get_queryset(self):
         organisation = Organisation.objects.get(id=organisation_pk)
 
         qs = UserPermissionGroup.objects.filter(organisation=organisation)
-        if not self.request.user.has_organisation_permission(
-            organisation, MANAGE_USER_GROUPS
+        if (
+            self.action != "summaries"
```
zachaysan, line 177: As we talked about offline, the `UserPermissionGroupPermission` needs to be added to handle the `summaries` action so that all users can see the group names.

zachaysan (approved):
We talked about it offline, but other than the one blocker the rest of this looks good.

api/users/views.py
matthewelwell: Excellent catch my man! 🙌

PR #3066, opened 2023-11-29
fix: resolve outstanding N+1 issues

matthewelwell:
## Changes

Resolves outstanding N+1 issues as observed in Sentry: 

* [list feature states](https://flagsmith.sentry.io/issues/4681801524/?environment=production&project=5544478&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=24h&stream_index=2)
* [list environment feature states](https://flagsmith.sentry.io/issues/4681165745/?environment=production&project=5544478&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=24h&stream_index=3)
* [list organisation users](https://flagsmith.sentry.io/issues/4331008588/?environment=production&project=5544478&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=24h&stream_index=4)
* [list features](https://flagsmith.sentry.io/issues/4331008588/?environment=production&project=5544478&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=24h&stream_index=4)

Added unit tests for each case and verified that, without the fix, the number of queries increased proportionately to the additional entities.


api/tests/unit/features/test_unit_features_views.py
```diff
@@ -2315,3 +2337,32 @@ def test_update_segment_override__using_simple_feature_state_viewset__denies_upd
...
+
+    base_url = reverse("api-v1:projects:project-features-list", args=[project.id])
+    url = f"{base_url}?environment={environment.id}"
+
+    # add some more versions to test for N+1 issues
+    v1_feature_state = FeatureState.objects.get(
+        feature=feature, environment=environment
+    )
+    for i in range(5):
+        v1_feature_state.clone(env=environment, version=2 + i, live_from=timezone.now())
```
zachaysan: Tiny nit: should be `range(2,7)`

api/tests/unit/users/test_unit_users_views.py
```diff
@@ -174,8 +174,16 @@ def test_admin_can_get_users_in_organisation(self):
             "api-v1:organisations:organisation-users-list", args=[self.organisation.pk]
         )
 
+        # add some more users to test for N+1 issues
+        for i in range(10):
+            additional_user = FFAdminUser.objects.create(
+                email=f"additional_user_{i}@org.com"
+            )
+            additional_user.add_organisation(self.organisation)
```
zachaysan: Do we really need to make 10 records where 2 or 3 would be fine?

api/users/models.py
```diff
@@ -235,12 +235,18 @@ def get_organisation_join_date(self, organisation):
...
+
         try:
-            return self.userorganisation_set.get(organisation=organisation)
-        except UserOrganisation.DoesNotExist:
+            return next(
+                filter(
+                    lambda uo: uo.organisation_id == organisation_id,
+                    self.userorganisation_set.all(),
+                )
+            )
```
zachaysan, line 252: Why can't a normal Django `.get` work here?

zachaysan (approved):
Some questions but approved!

api/tests/unit/features/test_unit_features_views.py
matthewelwell: Good call. I've changed to `range(2, 4)`.

api/tests/unit/users/test_unit_users_views.py
matthewelwell: Also a good call, I've changed to `range(2)`

api/users/models.py
matthewelwell: This is because of the way that `prefetch_related` works. If we do a `get()` here, it triggers another query to the database because django isn't smart enough to realise it already has the data.

zachaysan: That's good to know. Is this common enough of a pattern or should we put a small comment explaining the logic?

matthewelwell: It's (thankfully) not very common. I'll add a comment.

matthewelwell: Done 👍

PR #3079, opened 2023-11-30
fix: Reading role permissions generates 500 error backend

novakzaballa:

## Changes

- Add field roles to get organisation users

- Make a request /api/v1/organisations/{organisation_pk}/users/

### DIFF
--- a/api/users/models.py
+++ b/api/users/models.py
@@ -4,6 +4,7 @@ import typing
 from django.conf import settings
 from django.contrib.auth.base_user import BaseUserManager
 from django.contrib.auth.models import AbstractUser
+from django.core.exceptions import ImproperlyConfigured
 from django.core.mail import send_mail
 from django.db import models
 from django.db.models import Count, QuerySet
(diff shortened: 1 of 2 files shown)


api/users/models.py
```diff
@@ -243,6 +246,10 @@ def get_user_organisation(
                 % (self.id, getattr(organisation, "id", organisation))
             )
 
+    def get_user_roles(self):
+        user_roles = UserRole.objects.filter(user=self)
+        return user_roles
```
zachaysan: If `settings.IS_RBAC_INSTALLED` is not present then this method will throw an error. Maybe catch the exception and fire a custom error to stop it from looking like a random 500 error?

api/users/serializers.py
```diff
@@ -56,9 +60,15 @@ class Meta:
 class UserListSerializer(serializers.ModelSerializer):
     role = serializers.SerializerMethodField(read_only=True)
     join_date = serializers.SerializerMethodField(read_only=True)
+    if settings.IS_RBAC_INSTALLED:
+        roles = UserRoleSerializer(many=True, read_only=True, source="get_user_roles")
```
zachaysan, line 64: I don't really like the if-statement that leads to roles not being set on the serializer. Is this how we do it throughout the codebase? I feel like defaulting `roles` to `None` would look better.

api/users/models.py
novakzaballa: Done.

api/users/serializers.py
novakzaballa: I don't think this is possible since UserRoleSerializer lives in RBAC.

zachaysan: Ok makes sense.

zachaysan (approved):

PR #3081, opened 2023-11-30
fix: Move environments and features to test area

zachaysan:

## Changes

This is the second last wave of moving the test code into a centralized area of the repository. These mostly move the environment tests and merging existing tests together where appropriate.

Test it locally as I moved it over. This time there were no logical failures, though there was a set of errors that were due to the `--reuse-db` command flag.


novakzaballa (approved):
LGTM

PR #3083, opened 2023-12-01
feat: explicitly set audit log created date

matthewelwell:
## Changes

Allows AuditLog records to be set with an explicitly defined created_date. This means that, when re-running old tasks, the AuditLog records will be created with the correct timestamp. 

Updated unit tests for all modified tasks.


api/audit/migrations/0013_allow_manual_override_of_created_date.py
```diff
@@ -0,0 +1,18 @@
...
+        ('audit', '0012_auto_20230517_1006'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='auditlog',
+            name='created_date',
+            field=models.DateTimeField(verbose_name='DateCreated'),
+        ),
+    ]
```
matthewelwell, line 18: This operation is noop on the database. 

```
❯ python manage.py sqlmigrate audit 0013
System check identified some issues:

WARNINGS:
?: (axes.W002) You do not have 'axes.middleware.AxesMiddleware' in your settings.MIDDLEWARE.
BEGIN;
--
-- Alter field created_date on auditlog
--
COMMIT;
```

api/audit/tasks.py
```diff
@@ -123,6 +126,7 @@ def create_segment_priorities_changed_audit_log(
     feature_segment_ids: typing.List[int],
     user_id: int = None,
     master_api_key_id: int = None,
+    changed_at: str = None,
```
matthewelwell, line 130: I debated a larger (and arguably more elegant) change to resolve this issue which was to modify AbstractBaseTask.run() to pass in the `Task` as a key word argument to the task function and use `Task.scheduled_for`. This, however, would mean modifying all existing task functions to ensure that they can accept the additional kwarg. Since it's currently only really needed by this task (since the others have the date available from other means), I figured it was over-engineering at this stage. But I'm keen to know the thoughts of whoever reviews this PR...

novakzaballa: I agree that this looks like a more practical solution than modifying AbstractBaseTask.run().¡ in this case, it is my personal POV

novakzaballa (approved):
LGTM

PR #3090, opened 2023-12-04
fix: remove pagination from group summaries

matthewelwell:
## Changes

Since the FE needs to be able to grab the list of groups for an organisation to e.g. populate the group owners on a feature, and the summaries endpoint only returns 2 attributes for each group, we can safely remove the pagination from this endpoint. 

Note that this endpoint has not yet been released to production but this needs to be merged before the next release of the API to ensure that we're not creating breaking changes to the API. 

Updated the unit test.


zachaysan (approved):
Nice I like how clean this was.

PR #3096, opened 2023-12-05
feat: Organisation reverts to free plan

zachaysan:

## Changes

When a subscription cancels we lead to a mismatch in what the user expects, since they're no longer on the former subscription. What this change does is wipe out most of the subscription data and reverts it to what is at present for an empty free subscription. 

With a test through a task.


api/organisations/models.py
```diff
@@ -237,17 +241,51 @@ def update_mailer_lite_subscribers(self):
...
+        """
+        This method get's a subscription ready for cancelation.
+
+        If cancellation_date is in the future some aspects are
+        reserved for a task after the date has passed.
+        """
         if self.payment_method == CHARGEBEE and update_chargebee:
             cancel_chargebee_subscription(self.subscription_id)
 
+        if cancellation_date <= timezone.now():
```
gagantrivedi, line 291: Just so I have the context: did we consider using a normal task that is scheduled for `cacnellation_date` instead of using a recurring task to do that?

zachaysan: Yeah we discussed it at the time of the benefits vs drawbacks. I've had poor experiences with long running tasks in the past. Functions change their arguments and a task can fail over time. By centralized it in a single runner, if we make a change we're confident once its released.

api/organisations/models.py
```diff
@@ -237,17 +241,51 @@ def update_mailer_lite_subscribers(self):
...
+        Wipes a subscription to a normal free plan.
+
+        The only normal field that is retained is the notes field.
+        """
+        self.subscription_id = None
+        self.subscription_date = None
+        self.plan = FREE_PLAN_ID
+        self.max_seats = MAX_SEATS_IN_FREE_PLAN
+        self.max_api_calls = MAX_API_CALLS_IN_FREE_PLAN
+        self.cancellation_date = None
```
gagantrivedi, line 255: Not setting cancellation date will break this alert? Here :https://github.com/Flagsmith/flagsmith/blob/main/api/organisations/signals.py#L18 
Think we need to update that

gagantrivedi (changes requested):

api/organisations/models.py
zachaysan: This is a really good finding Gagan, thank you. I'm a little bit unsure of what to do in this scenario to restructure things without adding an additional field.

gagantrivedi: We can do it based on subscription_id/plan, i.e: if plan was not free but is now free?

gagantrivedi: I'd also just move this signal to a hook on subscription class

zachaysan: I decided that this type of method should be just a normal part of the cancellation workflow. In theory a subscriber could have their plan changed outside the cancelation workflow, but the same theory would go against non-cancelation of a Chargebee subscription, so I've deleted the signals and moved them into the cancel process.

api/organisations/models.py
```diff
@@ -237,17 +241,59 @@ def update_mailer_lite_subscribers(self):
...
+        If cancellation_date is in the future some aspects are
+        reserved for a task after the date has passed.
+        """
+        # Avoid circular import.
+        from users.models import FFAdminUser
 
         if self.payment_method == CHARGEBEE and update_chargebee:
             cancel_chargebee_subscription(self.subscription_id)
 
+        FFAdminUser.send_alert_to_admin_users(
```
gagantrivedi: nit: can we move this to a private method or maybe a task that we can delay and send emails asynchronously ... what do you think?

zachaysan: I think a small task would be a good idea. Let me implement that.

zachaysan: Ok that's been added.

api/tests/unit/organisations/test_unit_organisations_tasks.py
```diff
@@ -76,7 +78,56 @@ def test_send_org_over_limit_alert_for_organisation_with_subscription(
...
+    assert subscription.max_api_calls == MAX_API_CALLS_IN_FREE_PLAN
+    assert subscription.cancellation_date is None
+    assert subscription.customer_id is None
+    assert subscription.billing_status is None
+    assert subscription.payment_method is None
+    assert subscription.cancellation_date is None
+    assert subscription.notes == notes
+
+
+def test_finish_subscription_cancellation(db: None) -> None:
```
gagantrivedi: I think we should add an assert statement here somewhere to make sure the task was scheduled correctly?

api/organisations/tasks.py
```diff
@@ -38,6 +38,17 @@ def send_org_over_limit_alert(organisation_id):
     )
 
 
+@register_task_handler()
+def send_org_subscription_cancelled_alert(
```
gagantrivedi, line 42: I do understand this is a tiny method, but can we add a test for this as well?

gagantrivedi (changes requested):
This looks really good now! Just some minor comments

api/organisations/tasks.py
zachaysan: Haha, sure. Added one.

api/tests/unit/organisations/test_unit_organisations_tasks.py
zachaysan: Yeah that's a good call. Added.

api/tests/unit/organisations/test_unit_organisations_tasks.py
```diff
@@ -136,3 +202,23 @@ def test_finish_subscription_cancellation(db: None):
...
     assert organisation4.num_seats == organisation_user_count
+
+
+def test_send_org_subscription_cancelled_alert(db: None, mocker: MockerFixture) -> None:
+    # Given
+    send_mail_mock = mocker.patch("users.models.send_mail")
+
+    # When
+    send_org_subscription_cancelled_alert(
+        organisation_name="Nike",
```
gagantrivedi: Nike here makes me feel a bit uncomfortable for some reason… can we rename it to `test` or something

gagantrivedi (approved):

api/tests/unit/organisations/test_unit_organisations_models.py
```diff
@@ -85,7 +85,8 @@ def test_cancel_subscription_cancels_chargebee_subscription(
         )
         # refresh subscription object
         subscription.refresh_from_db()
-        assert subscription.cancellation_date
+        # Subscription has been immediately transformed to free.
+        assert subscription.cancellation_date is None
```
matthewelwell, line 90: Feels like there should be more asserts here?

api/tests/unit/organisations/test_unit_organisations_tasks.py
matthewelwell: Agree with this.

api/tests/unit/organisations/test_unit_organisations_models.py
zachaysan: This test was originally targeted at chargebee and other tests enforced the other checks, but I've added more asserts because there is no downside.

api/tests/unit/organisations/test_unit_organisations_tasks.py
zachaysan: I forget we're developing Open Source and Nike could ruffle feathers. In general, I like realistic named tests, but for this one it's the right call to change it something neutral.

PR #3107, opened 2023-12-06
docs: Add one click deploy to readme

jbgh:
Hey!

We built this one-click deploy link for your users who may not know how to stand up their own infrastructure to easily self-host this.

After deployment, users receive a link like this, which can then be CNAME-ed to any domain:

https://unleash-1665.dome.tools

Take it for a test run—we're eager to receive your feedback! We're more than willing to maintain this for the long haul, if this is valuable for your users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,13 +56,6 @@ The Flagsmith API is built using Python 3, Django 2, and DjangoRestFramework 3. 
 - [Kubernetes](https://github.com/Flagsmith/flagsmith-charts)
 - [Redhat OpenShift](https://operatorhub.io/operator/flagsmith)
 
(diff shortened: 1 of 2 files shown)


dabeeeenster:
Thanks for this - happy to add this here:

https://github.com/Flagsmith/flagsmith/blob/5033f3791450e24d221994f36e8b07e96bb0fe99/docs/docs/deployment/overview.md?plain=1#L18

but not in our main readme I'm afraid!

Also you need to rename the PR to follow our convention.

jbgh:
@dabeeeenster thanks for the feedback! Just pushed new changes to address comments.

dabeeeenster (approved):

PR #3115, opened 2023-12-08
chore: add DynamoEnvironmentV2Wrapper class

matthewelwell:
## Changes

Adds the `DynamoEnvironmentV2Wrapper` class as the ground work for multiple areas of functionality. 

This code has been ripped from another branch to speed up development.

There are currently no tests for this as the tests I created were more integration style tests. I have tested it manually by writing a basic test and adding my AWS credentials to verify that the data is returned as you expect.

### DIFF
--- a/api/environments/dynamodb/dynamodb_wrapper.py
+++ b/api/environments/dynamodb/dynamodb_wrapper.py
@@ -175,15 +175,15 @@ class DynamoEnvironmentV2Wrapper(BaseDynamoEnvironmentWrapper):
     ENVIRONMENT_API_KEY_INDEX_NAME = "environment_api_key-index"
 
     def get_environment_by_api_key(self, environment_api_key: str) -> dict:
-        filter_expression = Key(self.ENVIRONMENT_API_KEY_ATTRIBUTE).eq(
-            environment_api_key
-        ) & Key(self.DOCUMENT_KEY_ATTRIBUTE).eq("META")
-        query_kwargs = {
+        get_item_kwargs = {
             "IndexName": self.ENVIRONMENT_API_KEY_INDEX_NAME,
-            "KeyConditionExpression": filter_expression,
+            "Key": {
+                self.ENVIRONMENT_API_KEY_ATTRIBUTE: environment_api_key,
+                self.DOCUMENT_KEY_ATTRIBUTE: "META",
+            },
         }
         try:
-            return self._table.get_item(**query_kwargs)["Item"]
+            return self._table.get_item(**get_item_kwargs)["Item"]
         except IndexError:
             raise ObjectDoesNotExist()
 
--- a/api/environments/dynamodb/dynamodb_wrapper.py
+++ b/api/environments/dynamodb/dynamodb_wrapper.py
@@ -27,22 +27,25 @@ if typing.TYPE_CHECKING:
 logger = logging.getLogger()
 
 
-class DynamoWrapper:
+class BaseDynamoWrapper:
     table_name: str = None
 
     def __init__(self):
         self._table = None
-        if self.table_name:
+        if table_name := self.get_table_name():
             self._table = boto3.resource(
                 "dynamodb", config=Config(tcp_keepalive=True)
-            ).Table(self.table_name)
+            ).Table(table_name)
 
     @property
     def is_enabled(self) -> bool:
         return self._table is not None
 
+    def get_table_name(self) -> str:
+        return self.table_name
 
-class DynamoIdentityWrapper(DynamoWrapper):
+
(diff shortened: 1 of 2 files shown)


api/environments/dynamodb/dynamodb_wrapper.py
```diff
@@ -27,22 +27,25 @@
...
                 "dynamodb", config=Config(tcp_keepalive=True)
-            ).Table(self.table_name)
+            ).Table(table_name)
 
     @property
     def is_enabled(self) -> bool:
         return self._table is not None
 
+    def get_table_name(self) -> str:
+        return self.table_name
```
matthewelwell: I added this as it simplifies some of the mocking in tests that have built on top of this in my branch to add the overrides endpoint. It may not be necessary once I've refactored my tests but I don't think it hurts.

api/environments/dynamodb/dynamodb_wrapper.py
```diff
@@ -161,7 +169,45 @@ def get_item(self, api_key: str) -> dict:
...
+        filter_expression = Key("environment_api_key").eq(environment_api_key)
+        query_kwargs = {
+            "IndexName": "environment_api_key-index",
+            "Limit": 1,
+            "KeyConditionExpression": filter_expression,
+        }
+        try:
+            return self._table.query(**query_kwargs)["Items"][0]
+        except IndexError:
+            raise ObjectDoesNotExist()
```
khvn26: I've added the sort key to that index so it can be just:

```
.get_item(
    IndexName=ENVIRONMENT_TABLE_V2_SECONDARY_INDEX_NAME,
    Key={
        "environment_api_key": environment_api_key,
        "document_key": "META",
    },
)
```

api/environments/dynamodb/dynamodb_wrapper.py
khvn26: What is exactly the problem with mocking the `table_name` attribute?

api/environments/dynamodb/dynamodb_wrapper.py
matthewelwell: Ah I didn't realise you can do `get_item` with an index.

matthewelwell: Will update.

api/environments/dynamodb/dynamodb_wrapper.py
matthewelwell: Well, I was finding that the class was being loaded before the settings (using SettingsWrapper fixture) were applied. I guess I could use patch.object to set the table name directly, but I don't love using patch.object where it can be avoided.

matthewelwell: Like I said this morning though, I want to refactor my tests so I can rip this out for now if you'd rather?

api/environments/dynamodb/dynamodb_wrapper.py
matthewelwell: Done.

api/environments/dynamodb/dynamodb_wrapper.py
matthewelwell: > I can rip this out for now if you'd rather?

I've done this. Will re-add in my branch if I need it.

khvn26 (approved):

api/environments/dynamodb/dynamodb_wrapper.py
```diff
@@ -161,7 +166,47 @@ def get_item(self, api_key: str) -> dict:
...
+    ENVIRONMENT_API_KEY_INDEX_NAME = "environment_api_key-index"
+
+    def get_environment_by_api_key(self, environment_api_key: str) -> dict:
+        filter_expression = Key(self.ENVIRONMENT_API_KEY_ATTRIBUTE).eq(
+            environment_api_key
+        ) & Key(self.DOCUMENT_KEY_ATTRIBUTE).eq("META")
+        query_kwargs = {
+            "IndexName": self.ENVIRONMENT_API_KEY_INDEX_NAME,
+            "KeyConditionExpression": filter_expression,
+        }
```
khvn26: There's no need for a `KeyConditionExpression` now. Just input the primary key directly:

```suggestion
        query_kwargs = {
            "IndexName": self.ENVIRONMENT_API_KEY_INDEX_NAME,
            "Key": {
                "environment_api_key": environment_api_key,
                "document_key": "META",
            },
        }
```

matthewelwell: Yep, good point, done.

khvn26 (approved):

PR #3116, opened 2023-12-08
feat: add endpoint to list (edge) identity overrides for a feature

### LINKED ISSUE #3105: Display per-feature identity overrides from the `flagsmith_environments_v2` DynamoDB table
Migrate `/api/v1/environments/{environment_api_key}/featurestates` for select pilot environments.

Refer to schema detailed in https://github.com/Flagsmith/infrastructure/issues/18 to filter the overrides by feature id and retrieve the identity uuids.

matthewelwell:
## Changes

Fixes #3105.

Uses the new `flagsmith_environments_v2` (based on the structure defined [here](https://github.com/Flagsmith/infrastructure/issues/18)) to generate a view to show all the identity overrides for a particular feature. 

Adds an endpoint at `GET /api/v1/environments/:environment_id/edge-identity-overrides[?feature=:feature_id]` which returns data in the following structure: 

```json
{
  "results": [
    {
      "identity_uuid": "59efa2a7-6a45-46d6-b953-a7073a90eacf",
      "identifier": "identity1"
      "feature_state": {
        "feature_state_value": "foo",
        "feature": 1,
        "multivariate_feature_state_values": [],
        "enabled": true,
        "featurestate_uuid": "a7495917-ee57-41d1-a64e-e0697dbc57fb",
      }
    }
  ]
}
```

Added tests to the codebase where necessary. Also tested the dynamo integrations manually to confirm the data is returned as expected. 

I've also run my local API against staging dynamo and inserted data into the `flagsmith_environments_v2` table and verified that the endpoint picks up the data as expected.


api/edge_api/identities/serializers.py
```diff
@@ -77,7 +78,12 @@ def to_internal_value(self, data):
 
 class FeatureStateValueEdgeIdentityField(serializers.Field):
     def to_representation(self, obj):
-        identity_id = self.parent.get_identity_uuid()
+        identity: EdgeIdentity = self.parent.context["identity"]
+        environment: Environment = self.parent.context["environment"]
+        identity_id = identity.get_hash_key(
+            environment.use_identity_composite_key_for_hashing
+        )
+
```
matthewelwell, line 86: This is technically out of the scope of this PR, but this was an issue in the current implementation of the views which retrieve the feature states for a given identity.

The previous logic, although wrong, wouldn't have actually given any incorrect data (at least for users that are creating identity overrides via the dashboard, not via the API) since it only would have affected multivariate features with more than one weighted option, which we don't allow for identity overrides (you have to select a single option, which will be given 100% weighting). 

Since I had to also implement this functionality for the new endpoint, however, it made sense to fix it in this PR.

api/environments/dynamodb/dynamodb_wrapper.py
```diff
@@ -41,6 +41,9 @@ def __init__(self):
     def is_enabled(self) -> bool:
         return self._table is not None
 
+    def get_table_name(self):
+        return self.table_name
```
matthewelwell, line 45: @khvn26 I added this back in. It really simplifies the test [here](https://github.com/Flagsmith/flagsmith/pull/3116/files#diff-637cd4203bacb7503c8d4aa72d88892aa60da8fff0c726d9871ce472f351f062R62) and, in my opinion, we should migrate all other tests to use this approach since the current approach grabs the setting when the module is loaded and then we have to patch the object to change it.

api/environments/urls.py
```diff
@@ -142,4 +143,9 @@
         create_segment_override,
         name="create-segment-override",
     ),
+    path(
+        "<int:environment_pk>/edge-identity-overrides",
```
matthewelwell, line 147: I'm not sure whether to use the `pk` here or the API key. I definitely prefer using the pk, and would like to move to using the `pk` everywhere in the future so I figured this would at least not add further tech debt. It does, however, cause some consistency issues.

api/tests/unit/environments/dynamodb/test_unit_dynamodb_environment_v2_wrapper.py
```diff
@@ -0,0 +1,85 @@
...
+    os.environ["AWS_ACCESS_KEY_ID"] = "testing"
+    os.environ["AWS_SECRET_ACCESS_KEY"] = "testing"
+    os.environ["AWS_SECURITY_TOKEN"] = "testing"
+    os.environ["AWS_SESSION_TOKEN"] = "testing"
+    os.environ["AWS_DEFAULT_REGION"] = "eu-west-2"
+
+
+@pytest.fixture()
+def dynamodb(aws_credentials):
+    # TODO: move all wrapper tests to using moto
```
matthewelwell, line 28: I've left this as a todo here since I don't want to bloat this PR with extra work, moving other tests and fixtures around.

api/edge_api/identities/models.py
```diff
@@ -183,6 +183,7 @@ def save(self, user: FFAdminUser = None, master_api_key: MasterAPIKey = None):
                     "environment_api_key": self.environment_api_key,
                     "changes": changes,
                     "identity_uuid": str(self.identity_uuid),
+                    "identifier": self.identifier,
```
matthewelwell, line 186: @khvn26, I had to add the identifier here as I need it in the response to the FE. I believe we'll also need it to power identity override in local evaluation too so I think it's data worth having.

api/edge_api/identities/views.py
```diff
@@ -314,7 +330,14 @@ def put(self, request, *args, **kwargs):
         serializer = EdgeIdentityFeatureStateSerializer(
             instance=feature_state,
             data=request.data,
-            context={"view": self, "request": request},
+            context={
```
gagantrivedi, line 333: Can we use ` get_serializer_context` here to set the context?

api/environments/dynamodb/types.py
```diff
@@ -77,8 +77,13 @@ class IdentityOverrideV2(BaseModel):
     environment_id: str
     document_key: str
     environment_api_key: str
+    identifier: str
     feature_state: FeatureStateModel
 
+    @property
+    def identity_uuid(self) -> str:
+        return self.document_key.rsplit(":", maxsplit=1)[1]
+
```
khvn26: 1. I'd rather see this code in a mapper somewhere.
2. The code itself seems a bit hacky. Can't we just store the identity_uuid?

khvn26 (changes requested):

api/edge_api/identities/views.py
matthewelwell: Not really, no since this view is an `APIView`, not a viewset.

api/environments/dynamodb/types.py
matthewelwell: We could definitely just store the `identity_uuid`, yep. I think that's the better option here.

matthewelwell: This is done ^ ✅

khvn26: Please modify `map_engine_feature_state_to_identity_override`.

matthewelwell: It's already done [here](https://github.com/Flagsmith/flagsmith/pull/3116/files#diff-1a5f21351538565333f52d2a54d530cedc6e3ac8bf95e9b915a83622bc5264b3R103) 🤔

khvn26 (approved):

gagantrivedi (approved):

PR #3121, opened 2023-12-08
chore: Add multivariate warning

### LINKED ISSUE #2934: Alert users when they change a multivariate value to a flag
Changing existing multivariate values for an existing flag that is in use can change the value that an end user receives. This should show as a warning when saving a flag with new multivariate values. 

image

kyle-ssg:

## Changes

Add a warning for MV features that editing existing values can affect users.

Fixes: https://github.com/Flagsmith/flagsmith/issues/2934

image

Created / Edited an MV flag


matthewelwell:
@kyle-ssg I think this needs to be even clearer. It's not just editing the value in the current environment, right, it's editing it for all environments?

kyle-ssg:
I've adjusted the info alert to hopefully make both these points clearer 

image

matthewelwell (approved):

PR #3138, opened 2023-12-11
feat: Migrate given project's (edge) identities to environments v2

### LINKED ISSUE #3106: Populate `flagsmith_environments_v2` from `flagsmith_identities`
We need a way to migrate the data in the `identity_features` attribute on each identity in an environment into the new `flagsmith_overrides` table. 

This should be a one off migration process which we can create a Task in the task processor for.

khvn26:
Contributes to #3106.

## Changes

This PR adds a migration task that is triggered whenever a given Project model's `identity_overrides_v2_migration_status` changes to `IN_PROGRESS`.

The task iterates over project's environments and Edge identities, mapping and writing them to the new `flagsmith_environments_v2` table.

Added unit tests.


api/projects/migrations/0022_rename_identity_overrides_migration_status_project_identity_overrides_v2_migration_status.py
```diff
@@ -0,0 +1,18 @@
...
+        ('projects', '0021_add_identity_overrides_migration_status'),
+    ]
+
+    operations = [
+        migrations.RenameField(
+            model_name='project',
+            old_name='identity_overrides_migration_status',
+            new_name='identity_overrides_v2_migration_status',
+        ),
+    ]
```
matthewelwell: I'm not entirely sure this is worth doing at all but, if you think it is, I'd rather just update the old migration and fix the issues in the staging database.

api/projects/tasks.py
```diff
@@ -6,3 +8,17 @@ def write_environments_to_dynamodb(project_id: int):
...
     Environment.write_environments_to_dynamodb(project_id=project_id)
+
+
+@register_task_handler()
+def migrate_project_environments_to_v2(project_id: int):
+    from environments.dynamodb.services import migrate_environments_to_v2
+    from projects.models import IdentityOverridesV2MigrationStatus, Project
+
+    with transaction.atomic():
+        project = Project.objects.select_for_update().get(id=project_id)
```
matthewelwell, line 19: Why `select_for_update` here? The project is used heavily in the hot paths and the documentation is kind of unclear as to whether the row is locked for reads or not.

api/tests/unit/environments/dynamodb/test_unit_services.py
```diff
@@ -0,0 +1,80 @@
...
+    migrate_environments_to_v2(project_id=environment.project_id)
+
+    # Then
+    results = flagsmith_environments_v2_table.scan()["Items"]
+    assert len(results) == 2
+    assert results[0] == expected_environment_document
+    assert results[1] == expected_identity_override_document
+
+
+def test_migrate_environments_to_v2__wrappes_disabled__does_not_write(
```
matthewelwell:
```suggestion
def test_migrate_environments_to_v2__wrapper_disabled__does_not_write(
```

api/util/mappers/dynamodb.py
```diff
@@ -54,7 +54,7 @@ def map_environment_to_environment_v2_document(
 ) -> Document:
     return {
         **map_environment_to_environment_document(environment),
-        "document_key": "META",
+        "document_key": "_META",
```
matthewelwell: This should use a constant I think?

api/tests/unit/util/mappers/test_unit_mappers_dynamodb.py
```diff
@@ -143,7 +143,7 @@ def test_map_environment_to_environment_v2_document__call_expected(
 
     # Then
     assert result == {
-        "document_key": "META",
+        "document_key": "_META",
```
matthewelwell: This should use a constant I think?

matthewelwell (commented):
Added a few comments but looks pretty good on the whole. Obviously need to fix unit tests too :)

api/projects/tasks.py
khvn26: > Why `select_for_update` here? 

This is the best way I could think of to avoid accidentally spamming the task.

The row is not locked for reads in accordance with default [read committed](https://www.postgresql.org/docs/7.2/xact-read-committed.html) isolation level. This is easy to check e.g. via two Django shells. Unless some of our hot paths involve writing to a project row, we're good.

api/tests/unit/util/mappers/test_unit_mappers_dynamodb.py
khvn26: Done.

api/util/mappers/dynamodb.py
khvn26: Done.

api/projects/migrations/0022_rename_identity_overrides_migration_status_project_identity_overrides_v2_migration_status.py
khvn26: Done, but I'll probably ask for your assistance on reapplying the migration.

matthewelwell (approved):

PR #3143, opened 2023-12-12
chore: Use chunks when batch writing identity overrides

khvn26:

## Changes

This PR adds a limit of 25 items to the `BatchWriteItem` operation invoked in  `DynamoEnvironmentV2Wrapper.update_identity_overrides`. Overrides in a changeset are now updated in chunks of 25 as per the [operation documentation](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchWriteItem.html).

Covered the chunk generator with unit tests.


api/tests/unit/util/test_util.py
matthewelwell, line 1: Just for completeness, I'd suggest it's worth having a test for when the lists are the same length (and non-empty).

matthewelwell (commented):
Looks good on the whole, one minor request for an additional test but I've approved regardless. 

In general, I'd question if it's strictly _necessary_ and whether we could instead have just iterated over the to_put list to exhaustion, then the to_delete... but I do _like_ this approach 😃

matthewelwell (approved):

api/tests/unit/util/test_util.py
khvn26: Done.

matthewelwell (approved):

PR #3149, opened 2023-12-12
feat: Create flagsmith on flagsmith feature export task

zachaysan:

## Changes

This enables the easy export of Flagsmith feature flag features through the existing feature import export module. By setting two new settings constants, one for the target environment id and one for the tag id that will be selected for the feature flags that are to be exported, a daily background task will create new export ids which are saved through a new model. The most recent of which will be served through the download endpoint which is fixed and open to all without authentication.

Three new tests. The actual data itself is assert in both the task generation test and the view test.


api/app/settings/common.py
```diff
@@ -978,6 +978,15 @@
     "FLAGSMITH_ON_FLAGSMITH_SERVER_API_URL", default=FLAGSMITH_ON_FLAGSMITH_API_URL
 )
 
+FLAGSMITH_ON_FLAGSMITH_FEATURE_EXPORT_ENVIRONMENT_ID = env.int(
+    "FLAGSMITH_ON_FLAGSMITH_FEATURE_EXPORT_ENVIRONMENT_ID",
+    default=None,
+)
+FLAGSMITH_ON_FLAGSMITH_FEATURE_EXPORT_TAG_ID = env.int(
```
matthewelwell, line 985: For future proofing, should this be `env.list` ?

api/features/import_export/tasks.py
```diff
@@ -200,3 +204,33 @@ def _create_new_feature(
...
+
+@register_recurring_task(
+    run_every=timedelta(hours=24),
+)
+def create_flagsmith_on_flagsmith_feature_export():
+    if (
+        not settings.FLAGSMITH_ON_FLAGSMITH_FEATURE_EXPORT_ENVIRONMENT_ID
+        or not settings.FLAGSMITH_ON_FLAGSMITH_FEATURE_EXPORT_TAG_ID
+    ):
+        # No explicit settings on both feature settings, hence exit.
```
matthewelwell: I'd suggest that probably the settings filter should be before we register the recurring task, otherwise it might be a little odd for self hosted people that see the task registered. 

e.g. 

```python
if (settings.FLAGSMITH_ON_FLAGSMITH_FEATURE_EXPORT_ENVIRONMENT_ID and settings.FLAGSMITH_ON_FLAGSMITH_FEATURE_EXPORT_TAG_ID):
    @register_recurring_task(
        run_every=timedelta(hours=24),
    )
    def create_flagsmith_on_flagsmith_feature_export():
        ...
```

api/features/import_export/views.py
```diff
@@ -74,6 +76,37 @@ def download_feature_export(request: Request, feature_export_id: int) -> Respons
...
 
+@swagger_auto_schema(
+    method="GET",
+    responses={200: "Flagsmith on Flagsmith File downloaded"},
+    operation_description="This endpoint is to download an feature export to enable flagsmith on flagsmith",
+)
+@api_view(["GET"])
+@permission_classes([permissions.AllowAny])
+def download_flagsmith_on_flagsmith(request: Request) -> Response:
+    from django.http import Http404
```
matthewelwell: Why the local import?

api/features/import_export/models.py
```diff
@@ -76,3 +76,13 @@ class FeatureImport(models.Model):
...
+
+
+class FlagsmithOnFlagsmithFeatureExport(models.Model):
+    feature_export = models.ForeignKey(
+        FeatureExport,
+        related_name="flagsmith_on_flagsmith",
+        on_delete=models.CASCADE,
+    )
+
+    created_at = models.DateTimeField(auto_now_add=True)
```
matthewelwell: Interesting. I can certainly see why you went with this approach but I _think_ I'd prefer to just have an attribute on the FeatureExport class. Just thinking it's _slightly_ better aesthetics for the self hosted community where this functionality will be totally redundant. It'll be a redundant column vs a redundant table.

That being said, I don't have a massive preference here so if you think this is the cleanest approach, I'll take your word for it.

zachaysan: I think it's better to keep the whole functionality limited to it's own class. I can see `FeatureExport` getting expanded and used more and more as the functionality evolves and every dev investigating it will see a field that is used almost none of the time and it will clutter up the definition of the `FeatureExport` class. I think it's better to just have a single class here which self-describes itself. Plus, if there are any other features we decide to implement in the future, like say hold onto different versions of flagsmith for download, we can add it to the `FlagsmithOnFlagsmithFeatureExport` class.

api/features/import_export/views.py
zachaysan: I honestly don't recall. Probably a debugging issue. Sorry about that I will update. Thanks for the catch!

api/features/import_export/tasks.py
zachaysan: As we chatted briefly on Slack, there is the issue of importing this so in the mean time I've wrapped a private function and used it in conjunction with your suggestion to keep the task from queuing for the other clients.

api/app/settings/common.py
zachaysan: I really didn't think so given the requirements. From a UI perspective, it made more sense to me that people tag all the features that are supposed to be exported in one tag, rather than trying to combine tags together which seemed more error prone. For example, if we're using existing tags for the feature export and someone unaware of the feature export process adds a new tag to an existing one it could leak that feature by accident.

zachaysan: That being said, I can totally update this if you disagree. Ultimately I'm not the user here, it's the whole Flagsmith team.

matthewelwell: Nope, happy to go with the single tag. All good.

api/features/import_export/models.py
matthewelwell: Yep, fair enough. Happy with this approach in that case. Maybe we should add a docstring to the model however, to help explain it's purpose.

matthewelwell (approved):

api/features/import_export/models.py
zachaysan: Good idea I'll add it before merging the PR.

PR #3151, opened 2023-12-13
feat(tasks-processor): Add recurring task to clean up old recurring task runs

gagantrivedi:

- [ ] I have added information to `docs/` if required, so people know about the feature!

## Changes

Add a recurring task to delete recurring task runs older than `RECURRING_TASK_RUN_RETENTION_DAYS`

- By running task processor manually to verify, old task runs were deleted
- Adds unit tests

### DIFF
--- a/api/app/settings/common.py
+++ b/api/app/settings/common.py
@@ -910,6 +910,9 @@ TASK_DELETE_INCLUDE_FAILED_TASKS = env.bool(
 )
 TASK_DELETE_RUN_TIME = env.time("TASK_DELETE_RUN_TIME", default="01:00")
 TASK_DELETE_RUN_EVERY = env.timedelta("TASK_DELETE_RUN_EVERY", default=86400)
+RECURRING_TASK_RUN_RETENTION_DAYS = env.int(
+    "RECURRING_TASK_RUN_RETENTION_DAYS", default=30
+)
 
 # Real time(server sent events) settings
 SSE_SERVER_BASE_URL = env.str("SSE_SERVER_BASE_URL", None)
--- a/api/task_processor/tasks.py
+++ b/api/task_processor/tasks.py
@@ -9,7 +9,7 @@ from task_processor.decorators import (
     register_recurring_task,
     register_task_handler,
 )
-from task_processor.models import HealthCheckModel, Task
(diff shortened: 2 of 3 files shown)


dabeeeenster:
image

api/app/settings/common.py
```diff
@@ -910,6 +910,9 @@
 )
 TASK_DELETE_RUN_TIME = env.time("TASK_DELETE_RUN_TIME", default="01:00")
 TASK_DELETE_RUN_EVERY = env.timedelta("TASK_DELETE_RUN_EVERY", default=86400)
+RECURRING_TASK_RUN_RETENTION_DAYS = env.int(
+    "RECURRING_TASK_RUN_RETENTION_DAYS", default=30
+)
```
matthewelwell, line 915: Is there a reason not to use the same retention period for both types of task?

api/task_processor/tasks.py
```diff
@@ -50,3 +50,17 @@ def clean_up_old_tasks():
...
+    first_run_time=settings.TASK_DELETE_RUN_TIME,
+)
+def clean_up_old_recurring_task_runs():
+    if not settings.ENABLE_CLEAN_UP_OLD_TASKS:
+        return
+
+    now = timezone.now()
+    delete_before = now - timedelta(days=settings.RECURRING_TASK_RUN_RETENTION_DAYS)
+
+    RecurringTaskRun.objects.filter(finished_at__lt=delete_before).delete()
```
matthewelwell, line 66: Is there a reason that we shouldn't just use the same task as the one which removes regular tasks / task runs?

matthewelwell (commented):
Added a couple of comments but I'm happy to just approve this if you think the approach here is correct.

api/app/settings/common.py
gagantrivedi: Yeah, I think we probably want to control this independently, e.g: we can probably go with a lower value here because it's not as critical as tasks?

api/task_processor/tasks.py
gagantrivedi: Yeah, it's because of the name of the task mostly

matthewelwell (approved):

PR #3153, opened 2023-12-13
feat(task-processor): Add recurring task to clean password reset

### LINKED ISSUE #2649: Create a recurring task to clear UserPasswordResetRequest
Context: https://github.com/Flagsmith/flagsmith/pull/2619

gagantrivedi:

## Changes
Create a recurring task to clean-up `UserPasswordResetRequest` table 
closes https://github.com/Flagsmith/flagsmith/issues/2649

Adds unit test cases

### DIFF
--- a/api/task_processor/management/commands/runprocessor.py
+++ b/api/task_processor/management/commands/runprocessor.py
@@ -1,18 +1,13 @@
 import logging
-import os
 import signal
 import time
 import typing
 from argparse import ArgumentParser
 from datetime import timedelta
-from importlib import reload
 
 from django.core.management import BaseCommand
 from django.utils import timezone
 
-from custom_auth import tasks as auth_tasks
-from sse import tasks as sse_tasks
-from task_processor import tasks as processor_tasks
 from task_processor.task_registry import registered_tasks
 from task_processor.thread_monitoring import (
     clear_unhealthy_threads,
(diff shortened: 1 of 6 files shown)


api/task_processor/management/commands/runprocessor.py
```diff
@@ -21,7 +22,7 @@
 
 logger = logging.getLogger(__name__)
 
-TASKS_MODULES_TO_RELOAD = [processor_tasks, sse_tasks]
+TASKS_MODULES_TO_RELOAD = [processor_tasks, sse_tasks, auth_tasks]
```
gagantrivedi: I had to add auth_tasks here because that module is not imported by anything, i.e: python will not load it otherwise... It's not perfect, please feel free to suggest a better idea

khvn26: It's considered good practice to use `AppConfig.ready()` for such cases, e.g.

```python
# custom_auth/apps.py

class CustomAuthAppConfig(AppConfig):
    name = "custom_auth"
   
    def ready(self) -> None:
        from custom_auth import tasks
```

```python
# settings/common.py

...

INSTALLED_APPS = [
    ...
    "custom_auth",
    ...
]
```

gagantrivedi: I guess the phrased the comment wrong, we need a way to load it but not before we set `RUN_BY_PROCESSOR`. Importing tasks in ready will mean that we need to reload tasks here anyway

khvn26: I'd set `RUN_BY_PROCESSOR` outside the command then, e.g. in the `run_task_processor` function in the entrypoint script. Seems less hacky this way.

gagantrivedi: Yeah, that's a great idea! I wonder why it was not implemented like this in the first place

gagantrivedi: Done

khvn26 (approved):

PR #3161, opened 2023-12-13
feat: Import export environment flags

### LINKED ISSUE #3077: Create the feature import export frontend
We need a design solution for the feature import export. The API can be seen in [this PR](https://github.com/Flagsmith/flagsmith/pull/3026) which includes downloads and uploads of the features from and to a specific environment.

I, @zachaysan, will be available to assist with the API debugging since there are aspects of file upload that are tricky to debug.

kyle-ssg:

## Changes

- Allows a user to export an environment to a JSON file
- Allows a user to import an export to a project, specifying the merge strategy as either skip or overwrite destructive.
- Both the import and export allow you to preview changes
- Tag filtering now lets you choose between a union and intersection

image

image

image

image

- Exported multiple environments flags
- Imported using both strategies, alternating between new projects and environments


frontend/common/services/useFlagsmithProjectImport.ts
```diff
@@ -0,0 +1,59 @@
...
+export const flagsmithProjectImportService = service
+  .enhanceEndpoints({ addTagTypes: ['FlagsmithProjectImport'] })
+  .injectEndpoints({
+    endpoints: (builder) => ({
+      createFlagsmithProjectImport: builder.mutation<
+        Res['flagsmithProjectImport'],
+        Req['createFlagsmithProjectImport']
+      >({
+        invalidatesTags: [{ id: 'LIST', type: 'FlagsmithProjectImport' }],
+        queryFn: async (query, { dispatch }, _2, baseQuery) => {
```
zachaysan: Sorry if it's my ignorance here, but what does the `_2` do here?

frontend/common/types/requests.ts
```diff
@@ -130,8 +131,26 @@ export type Req = {
...
   getSubscriptionMetadata: { id: string }
-  createLaunchDarklyProjectImport: { project_id: string }
-  getLaunchDarklyProjectImport: { project_id: string }
-  getLaunchDarklyProjectsImport: { project_id: string; import_id: string }
+  createLaunchDarklyProjectImport: {
+    project_key: string
+    token: string
+  }
+  getLaunchDarklyProjectImport: { project_id: string; import_id: string }
+  getLaunchDarklyProjectsImport: { project_id: string }
```
zachaysan: These seem unrelated to the work at hand. Was it an introduction into the PR based on something new?

frontend/common/utils/toFormData.ts
```diff
@@ -0,0 +1,41 @@
...
+  formData: FormData = new FormData(),
+): FormData {
+  if (typeof data === 'string') {
+    formData.append(parentKey, data)
+  } else if (typeof data === 'number') {
+    formData.append(parentKey, data.toString())
+  } else if (typeof data === 'boolean') {
+    formData.append(parentKey, data ? 'true' : 'false')
+  } else if (data === null) {
+    // formData.append(parentKey, 'null')
```
zachaysan, line 23: This looks like a mistake. Surely the `formData` would append a `null` case?

frontend/web/components/JSONUpload.tsx
```diff
@@ -0,0 +1,98 @@
...
+        </Row>
+      ) : (
+        <div {...getRootProps()}>
+          <input {...getInputProps()} />
+          <div className='droparea text-center'>
+            <DropIcon />
+            <div className='mb-2'>
+              <strong>Select a file or drag and drop here</strong>
+            </div>
+            <div className='text-muted fs-small mb-4'>.json File</div>
```
zachaysan: Is the `.json` intentional?

frontend/web/components/import-export/FeatureExport.tsx
```diff
@@ -0,0 +1,215 @@
...
+  useEffect(() => {
+    if (
+      exports?.results.find(
+        (featureExport) => featureExport.status === 'PROCESSING',
+      )
+    ) {
+      timerRef.current = setTimeout(() => {
+        refetch()
+      }, 200)
+    }
```
zachaysan, line 72: 👍🏼

frontend/web/components/import-export/FeatureExport.tsx
```diff
@@ -0,0 +1,215 @@
...
+          title={
+            <>
+              <strong>Environment</strong>
+              <IonIcon icon={informationCircle} />
+            </>
+          }
+        >
+          {
+            'Selecting an environment will determine the feature defaults used when importing to a project'
+          }
```
zachaysan: Hmmm. Maybe this language could be tightened a bit to differentiate the environment people import to vs the other environments that take the defaults from the import.

frontend/web/components/import-export/FeatureExport.tsx
```diff
@@ -0,0 +1,215 @@
...
+          title={
+            <>
+              <strong>Filter Exported Features</strong>{' '}
+              <IonIcon icon={informationCircle} />
+            </>
+          }
+        >
+          {
+            'If any tags are selected then the exported file will only include features that are tagged.'
+          }
```
zachaysan: Maybe some clarity that any feature has any of the supplied tags will be included in the export.

frontend/web/components/import-export/FeatureImport.tsx
```diff
@@ -0,0 +1,303 @@
...
+          let newEnabled = featureState.enabled
+          if (strategy === 'OVERWRITE_DESTRUCTIVE' && importItem) {
+            if (environment === previewEnvironment) {
+              newEnabled = importItem.enabled
+              newValue = importItem.value
+            } else {
+              newEnabled = importItem.default_enabled
+              newValue = importItem.initial_value
+            }
+          }
```
zachaysan: This block needs a comment. Two or three times I thought you implemented it incorrectly, but it's done right. But it will be hard for others to follow the logic in the future.

frontend/web/components/import-export/FeatureImport.tsx
```diff
@@ -0,0 +1,303 @@
...
+              is_server_key_only: importItem.is_server_key_only,
+              multivariate_options: importItem.multivariate,
+              name: importItem.name,
+              num_identity_overrides: 0,
+              num_segment_overrides: 0,
+              owner_groups: [],
+              owners: [],
+              project: ProjectStore.model!.id,
+              tags: [],
+              type: 'STANDARD',
```
zachaysan: It's not always `STANDARD`, right? I think the multivariate option could come in here.

frontend/web/components/import-export/ImportPage.tsx
```diff
@@ -0,0 +1,244 @@
...
+      refetch()
+    }
+  }, [isSuccess, data, refetch])
+
+  const getProjectList = (LDKey: string) => {
+    setIsLoading(true)
+    _data
+      .get(`https://app.launchdarkly.com/api/v2/projects`, '', {
+        'Authorization': LDKey,
+      })
```
zachaysan, line 88: Here I'm a little confused why LD came into this feature, but it's on the same import page so maybe that is relevant?

frontend/web/components/tables/TableTagFilter.tsx
```diff
@@ -39,18 +45,39 @@ const TableTagFilter: FC<TableFilterType> = ({
...
+                }}
+                options={[
+                  {
+                    label: 'Has all',
+                    value: 'INTERSECTION',
+                  },
+                  {
+                    label: 'Has some',
+                    value: 'UNION',
+                  },
```
zachaysan, line 124: I'm happy these were added

zachaysan (approved):
I left a lot of questions and comments, but ultimately the PR looks great overall and I trust your judgement on what to add vs what to ignore from what I surfaced.

novakzaballa (approved):
It works and the code looks correct

frontend/common/utils/toFormData.ts
novakzaballa: I think the same.

frontend/common/types/requests.ts
kyle-ssg: This is just a lint fix, nothing new

frontend/common/utils/toFormData.ts
kyle-ssg: No, actually form data doesn't support null - it'd convert it to the string 'null'

frontend/web/components/import-export/ImportPage.tsx
kyle-ssg: It hasn't the file was just moved and the git diff here things it's all new code.

frontend/web/components/import-export/FeatureImport.tsx
kyle-ssg: Hmm that's a good point, I guess it's 


```
   type: importItem.multivariate?.length
                ? 'MULTIVARIATE'
                : 'STANDARD',
```

frontend/web/components/JSONUpload.tsx
kyle-ssg: It was but maybe it's better without

frontend/web/components/import-export/FeatureImport.tsx
```diff
@@ -0,0 +1,309 @@
...
+            }
+            onChange={() => setStrategy('OVERWRITE_DESTRUCTIVE')}
+            checked={strategy === 'OVERWRITE_DESTRUCTIVE'}
+          />
+        </div>
+      </div>
+      {strategy === 'OVERWRITE_DESTRUCTIVE' && (
+        <WarningMessage
+          warningMessage={
+            'This will overwrite existing features on all other environments to the defaults defined in the import, use with caution.'
```
matthewelwell:
```suggestion
            'Using overwrite destructive may affect ALL environments. Features in environments other than the one being imported to will receive the defaults defined in the import. Use with caution. For more information see the docs here.'
```

TODO: add docs link

matthewelwell (commented):
Tested this from a functionality perspective. Seems to work pretty well from what I can tell, but I've highlighted some issues I experienced here: 

image

^ this isn't strictly true when using the 'skip' import strategy. 

The UX as shown in the screencast below is kinda shonky. I'd argue that the small toast message is not enough feedback for the user here - and they can continue hitting the 'Import Project' button as many times as they like, which I imagine 80% of people will do when 'nothing' happens after they click it the first time. 

https://github.com/Flagsmith/flagsmith/assets/14089968/4a54a920-5da7-4608-9933-5e4af24ac2d5

Finally, is there a way to tell if the import is a no-op? When using the skip strategy, I think this is a valid use case. The preview can help a little here but I'd say that it's only really helpful for smaller projects. We may want to consider how we handle projects with very large numbers of features for this functionality. Perhaps a search option or filter by tag?

kyle-ssg:
> ^ this isn't strictly true when using the 'skip' import strategy.

Adjusted, I think I can just remove the first part so it's as such: 

image


> The UX as shown in the screencast below is kinda shonky. I'd argue that the small toast message is not enough feedback for the user here - and they can continue hitting the 'Import Project' button as many times as they like, which I imagine 80% of people will do when 'nothing' happens after they click it the first time.

Adjusted, it'll now clear the file and show the following: 

image

> Finally, is there a way to tell if the import is a no-op? When using the skip strategy, I think this is a valid use case. The preview can help a little here but I'd say that it's only really helpful for smaller projects. We may want to consider how we handle projects with very large numbers of features for this functionality. Perhaps a search option or filter by tag?

Have added searching / filtering as discussed.

frontend/web/components/import-export/FeatureExport.tsx
matthewelwell: Yeah, I just picked up on this. I think @kyle-ssg is maybe confused here. The environment values are used only to overwrite the values in the environment you import to. The default values that exist on the feature in the project of the _exporting_ environment are the ones that are used in other environments in the import project. 

I would suggest that we _either_ change the wording as follows, or just remove this tooltip altogether. 

```suggestion
          {
            'Select the environment to use for the export.'
          }
```

frontend/web/components/import-export/FeatureExport.tsx
```diff
@@ -0,0 +1,215 @@
...
+        <Tooltip
+          title={
+            <>
+              <strong>Filter Exported Features</strong>{' '}
+              <IonIcon icon={informationCircle} />
+            </>
+          }
+        >
+          {
+            'When tags are selected the export will contain any feature has any of the specified tags.'
```
matthewelwell:
```suggestion
            'When tags are selected the export will contain all features that have at least one of the specified tags.'
```

frontend/web/components/import-export/FeatureImport.tsx
```diff
@@ -0,0 +1,440 @@
...
+          </div>
+          <div className='my-2 col-6 col-lg-4'>
+            <EnvironmentSelect
+              projectId={projectId}
+              onChange={(e) => {
+                setPreviewEnvironment(e)
+              }}
+              value={previewEnvironment}
+            />
+          </div>
```
matthewelwell, line 347: I think it's a bit confusing that the Preview environment can be different to the environment you're importing into. Can we remove this select altogether and just use the import environment for the preview?

frontend/web/components/import-export/FeatureImport.tsx
```diff
@@ -0,0 +1,440 @@
...
+      {success ? (
+        <SuccessMessage>
+          Your import is processing, it may take several minutes for the import
+          to complete.
+        </SuccessMessage>
+      ) : (
+        <InfoMessage>
+          The selected environment will inherit the imported environment's
+          values where as all other environments will inherit the default
+          values.
```
matthewelwell:
```suggestion
          The selected environment will inherit the imported environment's
          values. All other environments will inherit the default values.
```

matthewelwell (commented):
One more thing that I've noticed, related to larger projects, is that the 'Export Features' / 'Import Features' button being at the bottom is quite painful. I would suggest this button should live above the list of features.

matthewelwell:
The feedback is still pretty weak here: 

image

I would suggest that we add a new API endpoint to get a feature import by it's ID so that the FE can poll it for the status. Currently it gets the entity back on the POST request to create the import but the `"status"` is returned as `"PROCESSING"`. This will eventually change to `"COMPLETE"` I believe, but the FE has no way of knowing this. cc @zachaysan 

This can perhaps be a future improvement, but I do think the feedback at the moment is really weak and, as a user, it would make me confused as to whether anything has happened or not.

matthewelwell:
@kyle-ssg the new endpoint is now available (in staging only until tomorrow morning) to list imports for a project at `GET /api/v1/projects/:id/feature-imports/`.

frontend/web/components/import-export/FeatureImport.tsx
kyle-ssg: Im not sure that'd be a good idea for overwrite destructive, since importing does affect other environments

matthewelwell:
@kyle-ssg I'm testing this on 2 separate projects. One has a single feature (the exporting project), the other has no features (the import project). I'm not sure if the preview was intended to handle this scenario where there are new features being created, but it's quite confusing that it's not. 

image

The import file: 

```json
[
  {
    "name": "test_feature",
    "default_enabled": false,
    "is_server_key_only": false,
    "initial_value": null,
    "value": null,
    "type": "unicode",
    "enabled": false,
    "multivariate": []
  }
]
```

Note that continuing with the import worked correctly and the feedback is definitely better.

kyle-ssg:
@matthewelwell correct it should be showing new flags, it attempts to filter them whilst filtering existing. This should now be fixed

matthewelwell (approved):

PR #3163, opened 2023-12-13
fix: Fine tune feature import export

zachaysan:

## Changes

As @kyle-ssg and I were going through the frontend integration a number of quality of life improvements came up:
1. Rather than listing feature exports by environment, it would be more useful to list them by `created_at` descending, since the newest created feature export should be at the top of the list.
2. Rather than following the feature export permissions where an environment admin is the most logical choice for an export, the *feature import* permission must be more expansive, since it can destructively overwrite all environments for the right permission, so this has been changed to organisation admin for import only.
3. To show features that are going to be import @kyle-ssg relied on the existing tagging system of the project feature list endpoint. The drawback with this approach is that it's an intersection approach, i.e., every feature must have every tag supplied, whereas the feature export uses an union approach, i.e., any one tag supplied is enough to warrant the inclusion of the feature. In order to support the proper function I added a `tag_strategy` serializer field which defaults to the old behavior and I added the new feature to the test suite plus an additional one since the old behaviour had no test coverage.

New tests and verified an existing one.


api/features/import_export/permissions.py
```diff
@@ -16,7 +16,10 @@ def has_permission(
             return False
 
         environment = Environment.objects.get(id=view.kwargs["environment_id"])
```
matthewelwell:
```suggestion
        environment = Environment.objects.select_related("project__organisation").get(id=view.kwargs["environment_id"])
```

api/features/views.py
```diff
@@ -290,12 +291,15 @@ def _filter_queryset(self, queryset: QuerySet) -> QuerySet:
         if "tags" in query_serializer.initial_data:
             if query_data.get("tags", "") == "":
                 queryset = queryset.filter(tags__isnull=True)
-            else:
+            elif query_data["tag_strategy"] == INTERSECTION:
```
matthewelwell: Feels slightly odd that we're checking if it's the default (as per the serializer) first, shouldn't we check for `UNION` first?

matthewelwell (commented):
Added a couple of minor comments but looks good on the whole.

api/features/views.py
zachaysan: I thought about it the other way around (i.e., check the default first since it's more likely), but if you're the reader and that's your conclusion then I'll take your word for it and switch them around.

api/features/import_export/permissions.py
zachaysan: Nice catch!

zachaysan: Ok this is done :)

api/features/views.py
zachaysan: Ok I switched it around for you!

matthewelwell: Ha, yeah, your logic is also sound here but let's leave it in the switched order for now.

matthewelwell (approved):

PR #3164, opened 2023-12-14
feat: Count v2 identity overrides for feature state list view

khvn26:

## Changes

- Adds identity override retrieval from `flagsmith_environments_v2` when preparing override counts.
- Base class for DynamoDB yet again changed to accommodate for all usage patterns.
- Naming improvements.

Modified and added unit tests.


api/conftest.py
```diff
@@ -623,15 +624,27 @@ def flagsmith_environments_v2_table(dynamodb: DynamoDBServiceResource) -> Table:
...
     settings: SettingsWrapper,
     flagsmith_identities_table: Table,
+    mocker: MockerFixture,
 ) -> DynamoIdentityWrapper:
     settings.IDENTITIES_TABLE_NAME_DYNAMO = flagsmith_identities_table.name
+    mocker.patch.object(
+        DynamoIdentityWrapper,
+        "get_table",
+        return_value=flagsmith_identities_table,
+    )
```
matthewelwell: I'll probably find this out later on, but I'm not sure why we need to use mocker.patch.object here, the idea behind refactoring to use the `get_table_name` was such that we didn't have to do that...

api/environments/dynamodb/dynamodb_wrapper.py
```diff
@@ -51,33 +51,44 @@
...
+    @classmethod
+    def get_table_name(cls) -> str:
+        return cls.table_name
+
+    @classmethod
+    def get_table(cls) -> typing.Optional["Table"]:
+        if table_name := cls.get_table_name():
+            return boto3.resource("dynamodb", config=Config(tcp_keepalive=True)).Table(
+                table_name
+            )
```
matthewelwell: I'm a little unclear why these have changed to classmethods

api/environments/dynamodb/dynamodb_wrapper.py
```diff
@@ -200,30 +211,31 @@ class DynamoEnvironmentWrapper(BaseDynamoEnvironmentWrapper):
...
 
 
 class DynamoEnvironmentV2Wrapper(BaseDynamoEnvironmentWrapper):
-    def get_table_name(self) -> str | None:
+    @classmethod
+    def get_table_name(cls) -> str | None:
         return settings.ENVIRONMENTS_V2_TABLE_NAME_DYNAMO
 
-    def get_identity_overrides_by_feature_id(
+    def get_identity_overrides_by_environment_id_and_feature_id(
```
matthewelwell: It seems odd to me that an optional argument is part of the name of the method. I think I'd rather change this to `get_identity_overrides_by_environment_id`.

api/features/features_service.py
```diff
@@ -1,33 +1,109 @@
...
+    project: Project,
+    environment: "Environment",
+) -> OverridesData:
+    """
+    Get correct overrides counts for a given project.
+
+    :param project: project to get overrides data for
+    :return: overrides data getter
+    """
+    match project.enable_dynamo_db, project.identity_overrides_v2_migration_status:
```
matthewelwell, line 28: Oh man, I didn't think about using tuples here, that's awesome!

api/features/features_service.py
```diff
@@ -1,33 +1,109 @@
...
+) -> OverridesData:
+    """
+    Get the number of identity / segment overrides in a given environment for each feature in the
+    project.
+    Retrieve identity override data from DynamoDB.
+
+    :param environment: the environment to get the overrides data for
+    :return OverridesData: dictionary of {feature_id: EnvironmentFeatureOverridesData}
+    """
+    with ThreadPoolExecutor() as executor:
```
matthewelwell, line 85: Nice 👌

api/features/features_service.py
```diff
@@ -1,33 +1,109 @@
...
+            feature_state.feature_id, EnvironmentFeatureOverridesData()
+        )
+        if feature_state.feature_segment_id:
+            env_feature_overrides_data.num_segment_overrides += 1
+    for identity_override in get_overrides_data_future.result():
+        env_feature_overrides_data = all_overrides_data.setdefault(
+            identity_override.feature_state.feature.id,
+            EnvironmentFeatureOverridesData(),
+        )
+        env_feature_overrides_data.add_identity_override()
```
matthewelwell: Technically we shouldn't have a case where we don't have the feature_id already in the dict, right?

api/tests/unit/features/test_unit_features_features_service.py
```diff
@@ -1,35 +1,184 @@
...
     )
+    from environments.models import Environment
+    from segments.models import Segment
+
 
-    feature_3 = Feature.objects.create(project=feature.project, name="feature_3")
-    feature_segment_for_feature_3 = FeatureSegment.objects.create(
-        feature=feature_3, segment=segment, environment=environment
+@pytest.fixture
+def distinct_segment_featurestate(
```
matthewelwell, line 32: It's not clear to me what this is 'distinct' from?

matthewelwell: Ah, I think it's a feature that is overridden by a segment only, not an identity? Is this necessary, can we use lazy_fixture below perhaps to parametrise the test?

api/features/features_service.py
matthewelwell: (I'm not saying we should necessarily change the behaviour here, but it might be worth a comment)

api/conftest.py
matthewelwell: I'm not sure I've found this out. Probably one to discuss over a short call.

khvn26: Fixed.

api/tests/unit/features/test_unit_features_features_service.py
khvn26: > It's not clear to me what this is 'distinct' from?

From the `segment_featurestate` fixture, which is for the same feature as `identity_featurestate`. I'll welcome better naming propositions. 

I don't think we need to parametrise the tests as they handle all the cases at once.

api/environments/dynamodb/dynamodb_wrapper.py
khvn26: For mocking (now reverted as unnecessary).

api/environments/dynamodb/dynamodb_wrapper.py
khvn26: Done.

api/environments/dynamodb/dynamodb_wrapper.py
```diff
@@ -51,33 +51,41 @@
...
+            )
+
+    @property
+    def is_enabled(self) -> bool:
+        return self.table is not None
+
 
 class DynamoIdentityWrapper(BaseDynamoWrapper):
-    def get_table_name(self) -> str | None:
+    def get_table_name(cls) -> str | None:
```
matthewelwell: Think this should be `get_table_name(self)` again now?

frontend/web/components/FeatureRow.js
```diff
@@ -257,7 +257,7 @@ class TheComponent extends Component {
                     </Tooltip>
                   </div>
                 )}
-                {!!projectFlag.num_identity_overrides && !Utils.getIsEdge() && (
+                {!!projectFlag.num_identity_overrides && (
```
matthewelwell: Ah, just realised that this will mean the FE gets deployed before the API, which is not great... We should pull this out to a separate PR and merge after deploying the API.

api/features/features_service.py
khvn26: Fixed.

api/features/features_service.py
```diff
@@ -1,33 +1,107 @@
...
 
 if typing.TYPE_CHECKING:
     from environments.models import Environment
 
 
+OverridesData = dict[int, EnvironmentFeatureOverridesData]
+
+
 def get_overrides_data(
+    project: Project,
```
matthewelwell: I'm not entirely sure why we need to pass in the project here rather than just using `environment.project`?

api/features/features_service.py
```diff
@@ -1,33 +1,107 @@
...
+    environment: "Environment",
+) -> OverridesData:
+    """
+    Get correct overrides counts for a given project.
```
matthewelwell: This is still getting the overrides data for an environment really.

matthewelwell (changes requested):

frontend/web/components/FeatureRow.js
khvn26: Removed, opened https://github.com/Flagsmith/flagsmith/pull/3177

api/environments/dynamodb/dynamodb_wrapper.py
khvn26: Fixed.

api/features/features_service.py
khvn26: Fixed.

api/features/features_service.py
khvn26: Fixed.

matthewelwell (approved):

PR #3165, opened 2023-12-14
chore: Simplify integration config

kyle-ssg:

## Changes
At the moment, releasing an integration requires editing 2 remote configurations, integrations and integration_data. With this PR, integrations will be determined by the keys within integration_data.

This will mean we can archive the 'integrations' feature.

Viewed the integrations page.


frontend/web/components/Aside.js
```diff
@@ -91,8 +91,8 @@ const Aside = class extends Component {
 
   render() {
     const { asideIsVisible, disabled, toggleAside } = this.props
-    let integrations = Utils.getFlagsmithValue('integrations') || '[]'
-    integrations = JSON.parse(integrations)
+    let integrations = Utils.getFlagsmithValue('integration_data') || '{}'
+    integrations = Object.keys(JSON.parse(integrations))
```
matthewelwell: I don't love the reassignment here, I think we should have 2 separate variables here, or just combine this into a single variable assignment? 

(Same comment on the `IntegrationsPage` diff too.

frontend/web/components/IntegrationList.js
```diff
@@ -266,18 +266,20 @@ class IntegrationList extends Component {
...
-                removeIntegration={this.removeIntegration}
-                projectId={this.props.projectId}
-                id={i}
-                key={i}
-                activeIntegrations={this.state.activeIntegrations[index]}
-                integration={integrationList[i]}
-              />
-            ))
+            this.props.integrations.map((i, index) =>
+              integrationList[i] ? (
```
matthewelwell: I guess this is just to protect against someone adding a key to the `integration_data` list with a null / empty value? Maybe there's a specific use case for this, but it seems a little unnecessary to me. If we're going down this line, it would make more sense to me to do some actual validation against a schema of sorts for `integrationList[i]` but I feel the chance of someone adding a falsy value here is pretty slim.

kyle-ssg: Yeah actually this PR protects against the more common scenario where theres a key in the integrations feature that didn't exist in integration_data

frontend/web/components/Aside.js
kyle-ssg: Ok I think we can do the latter since the FE waits for Flagsmith to have values

frontend/web/components/IntegrationList.js
matthewelwell: I thought that this PR would make that case completely redundant though, right, because we're not using the `integrations` feature at all?

dabeeeenster (approved):

frontend/web/components/Aside.js
```diff
@@ -91,8 +91,9 @@ const Aside = class extends Component {
 
   render() {
     const { asideIsVisible, disabled, toggleAside } = this.props
-    let integrations = Utils.getFlagsmithValue('integrations') || '[]'
-    integrations = JSON.parse(integrations)
+    const integrations = Object.keys(
+      JSON.parse(Utils.getFlagsmithValue('integration_data') || '{}'),
+    )
```
matthewelwell, line 96: I'd argue it's confusing that we now have `integrations` defined in separate places, using the same Flagsmith flag, but holding a different representation of the data. I'd recommend renaming one of the variables, but I'm not going to die on that hill so I'll approve regardless.

matthewelwell (approved):
I've also tested this via the preview link and all looks good to me.

PR #3166, opened 2023-12-14
chore: Persists tag filtering

kyle-ssg:

## Changes

When filtering features by tags the selection will be persisted, this means if the user refreshes / changes page they do not have to re-apply the filter.

Filtered by tag/archived and refreshed the page

### DIFF
--- a/frontend/common/stores/feature-list-store.js
+++ b/frontend/common/stores/feature-list-store.js
@@ -576,6 +576,10 @@ const controller = {
               : Promise.resolve(),
           ])
             .then(([features, environmentFeatures, feature]) => {
+              if (store.filter !== filter) {
+                //The filter has been changed since, ignore the api response. This will be resolved when moving to RTK.
+                return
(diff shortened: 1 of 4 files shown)


matthewelwell:
I was able to break this: 

image

Steps to reproduce (on the production preview link): 

1. Add some tag filters
2. Navigate to project settings -> import
3. Navigate back to features
4. Notice that the tags are not applied but the badge next to the 'Tags' in the table head thinks they are

Note that a refresh seems to reapply the filters correctly. Note that currently the import in project settings does not have the Flagsmith import / export functionality released but maybe I've hit some complex bug linked with that?

kyle-ssg:
@matthewelwell This has now been resolved, essentially what was happening is a race condition where the request without filters took priority. With the latest change, the FE will ignore API responses that don't match the current tag(or any) filter.

matthewelwell (approved):
LGTM

PR #3168, opened 2023-12-14
chore: Convert unit features permissions TestCase methods to normal test functions

zachaysan:
…ctions

## Changes

I'm working through re-writing the TestCase classes into normal test functions. The reasons for this are:

1. Less having to sort out what parts of a `setUp` method are actually `# Given` when it comes to replicating a failure or to make a similar method. 
2. Less adding bloat to a `setUp` method.
3. More reuse of fixtures instead of localized defined variables. For example, `with_project_permission` for the `staff_user` fixture is cleaner.

As I changed the code I added typing and I changed the assertion to check with equality with `True` or `False` instead of just a generic assert. I also got rid of the globals being used and just worked them into the tests.


api/tests/unit/features/test_unit_features_permissions.py
```diff
@@ -20,374 +21,492 @@
...
+    # Given
+    feature_permissions = FeaturePermissions()
+    mock_view = mock.MagicMock()
+    mock_request = mock.MagicMock()
+    mock_view.kwargs = {}
+    mock_request.data = {}
+    mock_view.action = "list"
+    mock_view.detail = False
+    mock_view.kwargs["project_pk"] = project.id
+    mock_request.user = admin_user
```
matthewelwell: Since we're now instantiating in each test, I think we can do this: 

```suggestion
    mock_view = mock.MagicMock(kwargs={"project_pk": project.id}, detail=False, action="list")
    mock_request = mock.MagicMock(data={}, user=admin_user)
```

I appreciate that we maybe don't want to make too many changes to the actual tests themselves but since we had to make a change here anyway, I think it makes sense. Happy for you to disagree here though. If we don't want to make the suggested change (and apply similar changes to the rest of the tests) then we should at least remove the duplication since we're assigning a value to `mock_view.kwargs` twice here (in some of the other tests `mock_request.data` is the offender). 

A final option would be to just define 2 fixtures locally in this test module that define the `mock_view` and `mock_request` objects.

matthewelwell (changes requested):

api/tests/unit/features/test_unit_features_permissions.py
zachaysan: No this is awesome! I didn't know about putting things in `MagicMock`'s constructors! Have updated!

api/tests/unit/features/test_unit_features_permissions.py
```diff
@@ -20,374 +21,453 @@
...
-        )
+    # Then
+    assert result is True
 
-        self.feature = Feature.objects.create(name="Test feature", project=self.project)
 
-        self.user = FFAdminUser.objects.create(email="user@test.com")
-        self.user.add_organisation(self.organisation, OrganisationRole.USER)
+def test_project_admin_can_list_features(
+    admin_user: FFAdminUser,
```
matthewelwell: Hmm... this isn't correct. It should be `staff_user` (and we should give them project admin permissions). Something like: 

```
UserProjectPermission.objects.create(user=staff_user, admin=True)
```

api/tests/unit/features/test_unit_features_permissions.py
```diff
@@ -20,374 +21,453 @@
...
-            mock_request, mock_view, self.feature
-        )
+    # When
+    result = feature_permissions.has_object_permission(mock_request, mock_view, feature)
+
+    # Then
+    assert result is True
+
+
+def test_project_user_cannot_edit_feature(
```
matthewelwell, line 344: This is a minor one but since the test name defines the user as a 'project user', perhaps they should have some affiliation with the project? Seems pointless, but just one to think about.

api/tests/unit/features/test_unit_features_permissions.py
```diff
@@ -20,374 +21,453 @@
...
+    staff_user: FFAdminUser,
+    feature: Feature,
+) -> None:
+    # Given
+    feature_permissions = FeaturePermissions()
+    mock_view = mock.MagicMock(detail=True, action="update")
```
matthewelwell: I wonder if (now that we have the capability), we should parametrize these `update` tests to include `action="partial_update"`.

matthewelwell (changes requested):

api/tests/unit/features/test_unit_features_permissions.py
zachaysan: That makes sense. I've added a `VIEW_PROJECT` permission to the user.

api/tests/unit/features/test_unit_features_permissions.py
zachaysan: Sure! I've added that to every action that matches `update` in the module.

matthewelwell (approved):

PR #3169, opened 2023-12-15
fix(task-processor): implement grace period for deleting old recurring task

### LINKED ISSUE #2551: Creating new Recurring Tasks only happens on second release
**Describe the bug**

When a new recurring task is added to the application, it will not be created on the first deployment that occurs. This is because, during the canary deployment, the new tasks will create the task in the database on startup, but the existing task processor units will delete it as it doesn't exist in their registry. 

See relevant code [here](https://github.com/Flagsmith/flagsmith/blob/main/api/task_processor/processor.py#L60-L63).

**To Reproduce**

1. Create a new Recurring Task.
2. Deploy the application to ECS.
3. Notice that the new task does not exist in the `task_processor_recurringtask` table.

**Expected behavior**

The recurring task is created in the database by the new deployment. 

**Screenshots**

N/a

- [ ] Self Hosted with Docker
- [x] SaaS at flagsmith.com
- [ ] Some other way

**Additional context**

N/a

gagantrivedi:

## Changes

Fixes https://github.com/Flagsmith/flagsmith/issues/2551 by implementing a grace period to not remove new tasks during deployment 

Updates/Adds unit test

### DIFF
--- a/api/tests/unit/task_processor/test_unit_task_processor_processor.py
+++ b/api/tests/unit/task_processor/test_unit_task_processor_processor.py
@@ -1,3 +1,4 @@
+import logging
 import time
 import uuid
 from datetime import timedelta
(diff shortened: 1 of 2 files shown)


zachaysan (commented):
Wouldn't the issue be solved if we could check the `created_at` on the `task_processor_recurringtask` and if the task is sufficiently old it could be deleted?

gagantrivedi:
> Wouldn't the issue be solved if we could check the `created_at` on the `task_processor_recurringtask` and if the task is sufficiently old it could be deleted?

Yeah, good shout! I have added the code to use a grace period

api/tests/unit/task_processor/test_unit_task_processor_processor.py
```diff
@@ -168,7 +176,38 @@ def _a_task():
...
+    def _a_task():
+        pass
+
+    # now - remove the task from the registry
+    registered_tasks.pop(task_identifier)
+
+    # make the task old
+    RecurringTask.objects.update(
+        created_at=timezone.now() - UNREGISTERED_RECURRING_TASK_GRACE_PERIOD
+    )
```
zachaysan: Minor nit: If we can use `freezegun` that would be better.

zachaysan (approved):
One minor nit but the rest looks good

PR #3171, opened 2023-12-15
ci: Enable merge queue checks. Run only changed tests.

novakzaballa:

## Changes
1. Added checks for merge queue on the `api-pull-request.yml` workflow
2. Added `testmon` to run only tests affected by changes during the PR workflow. The baseline will be created/updated on merges to `main`.

This needs to be tested manually during the workflow execution, so this PR will be marked as DO NOT MERGE until manual tests passed.


.github/workflows/api-pull-request.yml
```diff
@@ -12,14 +12,16 @@ on:
       - .github/**
     branches:
       - main
+  merge_group:
+    types: [checks_requested]
```
zachaysan, line 16: What does this do?

.github/workflows/api-pull-request.yml
```diff
@@ -33,7 +35,7 @@ jobs:
         options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
 
     strategy:
-      max-parallel: 4
+      max-parallel: 2
```
zachaysan, line 38: Why was this reduced to 2?

.github/workflows/api-pull-request.yml
```diff
@@ -67,11 +69,31 @@ jobs:
...
+          enableCrossOsArchive: true
+          path: |
+            /home/runner/work/flagsmith/flagsmith/api/.testmondata*
+          key: testmon-data-python${{ matrix.python-version }}-${{ github.event.pull_request.base.sha }}
+          restore-keys: testmon-data-python${{ matrix.python-version }}-
+
       - name: Run Tests
         env:
-          DOTENV_OVERRIDE_FILE: .env-ci
+          DOTENV_OVERRIDE_FILE: "${{ github.event_name == 'pull_request' && '.env-ci-testmon' || '.env-ci' }}"
```
zachaysan, line 85: Since the id is `cache-testmon-restore` it's kinda weird to fallback to `.env-ci`. When do we do that here?

api/poetry.lock
```diff
@@ -3293,6 +3293,21 @@ pytest = ">=5.0"
 [package.extras]
 dev = ["pre-commit", "pytest-asyncio", "tox"]
 
+[[package]]
+name = "pytest-testmon"
+version = "2.0.13"
+description = "selects tests affected by changed files and methods"
```
zachaysan, line 3299: I understand this only runs a subset of the tests that we have. Does this PR make that only possible for comparisons to main or does the whole CI run during the initial PR creation?

.github/workflows/api-pull-request.yml
novakzaballa: Runs the jobs of this workflow for merge-queue events, please refer to https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue and https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#merge_group for further details.

.github/workflows/api-pull-request.yml
novakzaballa: Why should we have more, if we only have 2 Python versions?

.github/workflows/api-pull-request.yml
novakzaballa: I think I don't get it. These lines are under the `Run Tests` job while the ID you are referring is under the `Restore cached testmon data` job right?

.github/workflows/api-pull-request.yml
zachaysan: Oh that's what this is for. I misunderstood. Thanks for the update.

.github/workflows/api-pull-request.yml
zachaysan: Oh I see I looked at the embedded data incorrectly. This makes sense now. Thanks!

api/poetry.lock
novakzaballa: This code uses the latest version of the base branch, main in this case, as a baseline for comparison, to run only tests related to changed code files. So every time you create a PR, or modify it, only tests affected by changes will run. Once a PR is merged into main, we will create a new baseline for comparisons.

zachaysan (approved):

PR #3175, opened 2023-12-18
docs: add beta integrations

dabeeeenster:
Also added some line breaks to space things out a bit


docs/docs/integrations/overview.md
```diff
@@ -18,6 +18,11 @@ Infrastructure as Code tooling. [Learn more](/integrations/terraform).
 
 ## Analytics Platforms
 
+<a href="/integrations/analytics/adobe" alt="Adobe"><img width="40%" src="/img/integrations/adobe/adobe-logo.svg"/></a>
+
+You can integrate Flagsmith with Adobe Analytics. You can automate AB tests by connecting the Flagsmith platform with
+Adobe Analytics. [Learn more](/integrations/analytics/amplitude).
```
matthewelwell:
```suggestion
Adobe Analytics.
```

docs/docs/integrations/analytics/adobe.md
```diff
@@ -0,0 +1,16 @@
...
+---
+
+![Adobe](/img/integrations/adobe/adobe-logo.svg)
+
+You can integrate Flagsmith with Adobe Analytics. You can automate AB tests by connecting the Flagsmith platform with
+Adobe Analytics.
+
+:::tip
+
+The Adobe Analytics integration is in Beta - please contact us to join the Beta!
```
matthewelwell: I wonder if we should be less aggressive here, could we say `alpha` ?

matthewelwell (commented):
My conservative nature is telling me this is a bad idea, but I'll defer to people with greater experience here!

docs/docs/integrations/analytics/adobe.md
dabeeeenster: OK Yep updated.

matthewelwell: 🤔 
image

matthewelwell (approved):
Approving but I'd argue that we need to revisit the styling. The `hr`s make the headings look less heading-y...

PR #3182, opened 2023-12-18
fix(subscriptions): ensure that manually added subscriptions work correctly in all deployments

matthewelwell:
## Changes

This PR ensures that manually added subscriptions work in all settings. Currently it is only possible to add a subscription manually if you define the subscription_id value as "trial". This PR will mean that manually added subscriptions will not factor in the subscription id. 

Another approach that I considered here was to have other, pre-defined constants that should be set as the subscription id in different scenarios. 

Finally, this is a stop gap improvement until we can use license files to better handle on premise deployments. 

Added / updated tests.

### DIFF
--- a/api/tests/unit/organisations/test_unit_organisations_models.py
+++ b/api/tests/unit/organisations/test_unit_organisations_models.py
@@ -427,7 +427,7 @@ def test_get_subscription_metadata_for_manually_added_enterprise_saas_licenses(
 ) -> None:
     """
     Specific test to make sure that we can manually add subscriptions to
-    the saas platform and the values stored in the django database will
+    the SaaS platform and the values stored in the Django database will
     be correctly used.
     """
     # Given
--- a/api/tests/unit/organisations/test_unit_organisations_models.py
+++ b/api/tests/unit/organisations/test_unit_organisations_models.py
@@ -2,6 +2,7 @@ from datetime import datetime
 from unittest import mock
 
 import pytest
+from django.conf import settings
 from django.test import TestCase
 from pytest_mock import MockerFixture
 from rest_framework.test import override_settings
@@ -9,7 +10,6 @@ from rest_framework.test import override_settings
 from environments.models import Environment
 from organisations.chargebee.metadata import ChargebeeObjMetadata
 from organisations.models import (
-    TRIAL_SUBSCRIPTION_ID,
     Organisation,
     OrganisationRole,
     OrganisationSubscriptionInformationCache,
@@ -19,6 +19,9 @@ from organisations.subscriptions.constants import (
     CHARGEBEE,
     FREE_PLAN_ID,
     FREE_PLAN_SUBSCRIPTION_METADATA,
+    MAX_API_CALLS_IN_FREE_PLAN,
+    MAX_SEATS_IN_FREE_PLAN,
+    TRIAL_SUBSCRIPTION_ID,
     XERO,
 )
 from organisations.subscriptions.exceptions import (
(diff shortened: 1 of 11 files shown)


api/tests/unit/organisations/test_unit_organisations_models.py
```diff
@@ -283,42 +286,48 @@ def test_organisation_is_paid_returns_false_if_cancelled_subscription_exists(
...
+        allowed_projects=projects,
     )
-
     expected_metadata = ChargebeeObjMetadata(
-        seats=seats, api_calls=api_calls, projects=10
+        seats=seats, api_calls=api_calls, projects=projects
     )
-    mock_cb_get_subscription_metadata = mocker.patch(
-        "organisations.models.Subscription.get_subscription_metadata"
+    mocker.patch("organisations.models.is_saas", return_value=True)
```
zachaysan, line 311: Much better.

api/tests/unit/organisations/test_unit_organisations_models.py
```diff
@@ -338,24 +347,125 @@ def test_subscription_get_subscription_metadata_returns_free_plan_metadata_for_n
...
+    max_seats: int,
+    max_api_calls: int,
+    expected_seats: int,
+    expected_api_calls: int,
+    expected_projects: int | None,
+    mocker: MockerFixture,
+) -> None:
+    """
+    Specific test to make sure that we can manually add subscriptions to
+    the saas platform and the values stored in the django database will
```
zachaysan: tiny nit:
```suggestion
    the SaaS platform and the values stored in the Django database will
```

api/tests/unit/organisations/test_unit_organisations_models.py
```diff
@@ -338,24 +347,125 @@ def test_subscription_get_subscription_metadata_returns_free_plan_metadata_for_n
...
+    assert subscription_metadata.api_calls == expected_api_calls
+
+
+def test_get_subscription_metadata_for_self_hosted_open_source(
+    organisation: Organisation, mocker: MockerFixture
+) -> None:
+    """
+    Open source should ignore the details provided in the
+    subscription and always return the free plan metadata.
+    """
```
zachaysan, line 458: So if I understand this correctly, we would want all our enterprise customers to add `./ENTERPRISE_VERSION` before merging this code into their copy, correct? Or did I misunderstand?

zachaysan (approved):
I think I understand this and I couldn't find any obvious problems with the code or tests so I'm approving it, but I'm just confused about how the change will rollout to our customers. Also, we'll need to add `./SAAS` to at least staging and production, right? Also, for working with things locally, we'd have to add one of the two paths to get subscription metadata outside of the free plan as well?

matthewelwell:
> I think I understand this and I couldn't find any obvious problems with the code or tests so I'm approving it, but I'm just confused about how the change will rollout to our customers. Also, we'll need to add `./SAAS` to at least staging and production, right? 

Yep, this is done [here](https://github.com/Flagsmith/flagsmith/pull/3182/files#diff-04acc058a387dc5cc3daa95f77bec112758ca8b484d17fc19eb658a290356d2cR138). 

> Also, for working with things locally, we'd have to add one of the two paths to get subscription metadata outside of the free plan as well?

Yes, that's true, but I think that's a necessary step here. If it proves to be super painful then we can revisit and see if there's a way we can work around it for local development, but obviously for testing it's super easy to mock those util functions (we may even want to create fixtures to mock them).

api/tests/unit/organisations/test_unit_organisations_models.py
matthewelwell: Nope, that is already done in the docker builds [here](https://github.com/Flagsmith/flagsmith/blob/main/.github/workflows/platform-docker-publish-all-features-image.yml#L78) and [here](https://github.com/Flagsmith/flagsmith-ee/pull/102) (this one needs merging still - I will do that now).

PR #3188, opened 2023-12-19
chore: Github integration (API)

### LINKED ISSUE #1272: Github Integration
_This probably will need splitting up into several issues and an overriding milestone_

# Overview

This feature enables you to see the state of a Flagsmith flag within an associated Github Issue/PR

image

## Version 1

Allow Flagsmith users to connect a Github Issue/Pull Request with a Flagsmith flag. Once connected, a Flagsmith Github App/Bot will create and update a comment against the issue showing the state of the Flag in each Flagsmith environment 

## Version 2

Allow users to enable/disable feature states by clicking the checkboxes. This issue currently only specs Version 1.

# Auth

Create a Github Application. Question of whether we should use a Github App or an OAuth app - https://docs.github.com/en/developers/apps/getting-started-with-apps/differences-between-github-apps-and-oauth-apps. Current thinking is use a Github app as that is most common for other apps of this nature. 

Flagsmith users will need to give the Github Application permission over their account. 

# Connect Flag with Issue/PR

Once Auth is done, allow users to paste a Github Issue/PR URL into a Flag Settings page field. Store this against the `Feature` API model object. 

# Write a comment

Once URL is pasted, create a comment against the Issue/PR containing:

- The Flagsmith Feature Name
- For all `Environments`, show a checkbox and `codeblock text field` showing the enabled state and the string value if there is one 
- A sig at the bottom of the comment saying that the comment was generated by a Flagsmith bot with links to the docs

Record the Github ID of the comment against the `Feature` in Flagsmith so that we can update it on later feature state events.

If the write fails due to lack of permission, feed that back to the user. If the write succeeds, show a green tick next to the issue url. 

# Update the comment

Whenever the `feature_state` of the connected `feature` changes state, update the connected Github comment

# Things to consider

Do we want to make this a more generic solution so that we could do similar with e.g. Gitlab/Jira etc?

### LINKED ISSUE #1769: CI/CD Integrations
We should build valuable integrations with the following tools:

GitHub Actions: https://docs.github.com/en/developers/github-marketplace/creating-apps-for-github-marketplace/requirements-for-listing-an-app

Gitlab: https://about.gitlab.com/partners/technology-partners/integrate/

CircleCI: https://circleci.com/docs/orb-author/

gz#216

novakzaballa:

## Changes

2 new models called 'external_resources', and 'github' were added, along with its views and serializers, to link GitHub information with Flagsmith data.

### Configuration:
You need the set the following env vars to enable GitHub Integration in the API:
- GITHUB_PEM
- GITHUB_APP_ID

Both of the above values can be found in the settings page of the GitHub app once you register the app.

- Unit tests added.


api/features/tasks.py
```diff
@@ -34,6 +35,11 @@ def trigger_feature_state_change_webhooks(
     )
     data = {"new_state": new_state, "changed_by": changed_by, "timestamp": timestamp}
     previous_state = _get_previous_state(history_instance, event_type)
+    print(
```
gagantrivedi: this should not be here?

api/external_resources/models.py
```diff
@@ -0,0 +1,18 @@
+from django.db import models
+
+from features.models import Feature
+
+
+class ExternalResources(models.Model):
```
gagantrivedi: What is an external resource? Can this be used for anything other than GitHub?

api/external_resources/signals.py
```diff
@@ -0,0 +1,93 @@
...
+from integrations.github.tasks import call_github_app_webhook_for_feature_state
+from webhooks.webhooks import WebhookEventType
+
+# noinspection PyUnresolvedReferences
+from .models import ExternalResources
+
+logger = logging.getLogger(__name__)
+
+
+@receiver(post_save, sender=ExternalResources)
```
gagantrivedi: can we use a hook for this instead of a signal?

api/integrations/github/views.py
```diff
@@ -0,0 +1,123 @@
+import requests
```
gagantrivedi, line 3: We will need tests for this module.
PS: we will need tests for all the new code introduced in this PR

gagantrivedi (changes requested):

api/features/tasks.py
novakzaballa: Removed.

api/external_resources/models.py
novakzaballa: The external resource model is intended to be something generic in case there are other integrations with the app.

api/external_resources/signals.py
novakzaballa: Yes, it is possible. I converted them into hooks. Anyway, I tried to base it on the original webhook that used signals, which was triggered when there was a change in the feature state.

api/external_resources/models.py
```diff
@@ -0,0 +1,102 @@
...
+    LifecycleModelMixin,
+    hook,
+)
+
+from features.models import Feature, FeatureState
+from integrations.github.tasks import call_github_app_webhook_for_feature_state
+from webhooks.webhooks import WebhookEventType
+
+
+class ExternalResources(LifecycleModelMixin, models.Model):
```
gagantrivedi: I still don't understand what external resource is? It just looks like webhook endpoint to me?

api/external_resources/models.py
```diff
@@ -0,0 +1,102 @@
...
+    url = models.URLField()
+    type = models.TextField()
```
gagantrivedi: What all type can this have? Can we make this a choice field?

api/external_resources/models.py
```diff
@@ -0,0 +1,102 @@
...
+    status = models.TextField(null=True)
```
gagantrivedi: Same as above (can status be choice field)

api/tests/unit/external_resources/test_unit_external_resources_views.py
```diff
@@ -0,0 +1,79 @@
...
+    )
+
+    # When
+    response = admin_client.get(url)
+
+    # Then
+    assert response.status_code == status.HTTP_200_OK
+
+
+def test_get_external_resource(
```
gagantrivedi: This is the duplicate of the test above

api/external_resources/models.py
```diff
@@ -0,0 +1,102 @@
...
+    feature = models.ForeignKey(
+        Feature,
+        related_name="features",
+        on_delete=models.CASCADE,
+    )
+
+    @hook(AFTER_SAVE)
+    def trigger_feature_external_resource_added_hook(self):
```
gagantrivedi: Also, do we have unit tests for this?

api/features/tasks.py
```diff
@@ -50,6 +52,49 @@ def trigger_feature_state_change_webhooks(
...
+            "feature_states": [],
+        }
+        feature_data["installation_id"] = github_configuration.installation_id
+        feature_data["organisation_id"] = github_configuration.organisation.id
+
+        feature_state = {
+            "environment_name": new_state["environment"]["name"],
+            "feature_value": new_state["enabled"],
+        }
+
```
gagantrivedi: Looks like we have the same logic in models.py as well, we should make this more DRY

api/external_resources/models.py
```diff
@@ -0,0 +1,102 @@
...
+                "id": self.feature_id,
+                "name": self.feature.name,
+                "feature_states": [],
+            }
+            feature_data["installation_id"] = github_configuration.installation_id
+            feature_data["organisation_id"] = github_configuration.organisation.id
+
+            for feature_state in feature_states:
+                feature_env_data = {}
+                if feature_state.feature_state_value.string_value is not None:
```
gagantrivedi: Can we move this type checking to `github` module?

api/external_resources/views.py
```diff
@@ -0,0 +1,21 @@
+from rest_framework import viewsets
+from rest_framework.permissions import IsAuthenticated
+
+from .models import ExternalResources
+from .serializers import ExternalResourcesSerializer
+
+
+class ExternalResourcesViewSet(viewsets.ModelViewSet):
+    serializer_class = ExternalResourcesSerializer
+    permission_classes = [IsAuthenticated]
```
gagantrivedi: Should this need feature update permission? Or are we making it open by choice?

api/integrations/github/github.py
```diff
@@ -0,0 +1,82 @@
...
+            "Accept": "application/vnd.github.v3+json",
+            "Authorization": f"Bearer {token}",
+        }
+
+        payload = {"body": body}
+        response = requests.post(url, json=payload, headers=headers)
+
+        return response.json() if response.status_code == 200 else None
+    except requests.RequestException as e:
+        print(f" {e}")
```
gagantrivedi: We should use logging here

api/integrations/github/github.py
```diff
@@ -0,0 +1,82 @@
...
+    def is_string(value):
+        return isinstance(value, str)
+
+    def is_boolean(value):
+        return isinstance(value, bool)
+
+    def parse_language(value):
+        return ""
+
+    is_update = event_type == "FLAG_UPDATED"
```
gagantrivedi: We should move them to constants

api/integrations/github/github.py
```diff
@@ -0,0 +1,82 @@
...
+    )
+    updated_text = (
+        f"### The Flagsmith Feature {name} "
+        f"was updated in the environment {feature_states[0]['environment_name']}"
+    )
+
+    result = updated_text if is_update else f"title ({name}):\n"
+
+    for v in feature_states:
+        feature_value = (
```
gagantrivedi: We won't need this parsing if we just let the user pass the value? 
related comment in `
api/external_resources/models.py
`

gagantrivedi (changes requested):
We should add more type hints

api/integrations/github/github.py
novakzaballa: Added.

api/integrations/github/github.py
novakzaballa: Corrected

api/external_resources/models.py
novakzaballa: Done.

api/tests/unit/external_resources/test_unit_external_resources_views.py
novakzaballa: This is another test.

api/external_resources/models.py
novakzaballa: Done.

api/external_resources/models.py
novakzaballa: It is any reference to an external resource using a URI. In this case it is used to link a GH PR/Issue to a flag. You can use it in the future to link any kind of external resource like a JIRA ticket, slack comment, etc. Once we have more than one external App, we could use Django Generic Relations and ContentType to better model this external-resources and its types.

api/integrations/github/github.py
novakzaballa: In a Django hook executed during updates, direct value passing may not be feasible. The current logic depends on whether it's an update (is_update) and utilizes values from feature_states. Retaining text parsing logic for value extraction seems necessary in this scenario

api/external_resources/models.py
novakzaballa: Moved

api/external_resources/models.py
novakzaballa: Yes, the unit test was added.

api/external_resources/views.py
novakzaballa: Good question, maybe this is an option, or maybe create a new permission for this?

gagantrivedi: I would not add a new permission for this(I will just tie it to feature update) if the user wants to use different permissions they can probably do that with tag based permissions?

api/external_resources/models.py
gagantrivedi: If this is generic, then methods like `create_feature_external_resource_added_comment` should not exist here, right? I think we should define a base model here and actual implementation can just inherit that and implement the hook

api/external_resources/models.py
```diff
@@ -0,0 +1,111 @@
...
+
+            call_github_app_webhook_for_feature_state.delay(
+                args=(
+                    feature_data,
+                    WebhookEventType.FEATURE_EXTERNAL_RESOURCE_REMOVED.value,
+                ),
+            )
+        else:
+            organisation_id = self.feature.project.organisation.id
+            print(
```
gagantrivedi: This should use logging instead of print

api/external_resources/models.py
```diff
@@ -0,0 +1,111 @@
...
+                f"No GitHub integration exists for organisation {organisation_id}. Not calling webhooks."
+            )
+
+    class Meta:
+        ordering = ("id",)
+        unique_together = ("url",)
```
gagantrivedi: We should use [unique constraint](https://docs.djangoproject.com/en/5.0/ref/models/constraints/#uniqueconstraint) here. Also, is there a reason why we don't want the user to attach one GitHub issue to two or more features?

api/integrations/github/constants.py
```diff
@@ -0,0 +1,3 @@
+GITHUB_API_URL = "https://api.github.com/"
+GITHUB_API_VERSION = "2022-11-28"
+GITHUB_APP_ID = 811209
```
gagantrivedi: Is this static? Or should we move this to settings?

api/integrations/github/tasks.py
```diff
@@ -0,0 +1,95 @@
...
+    generate_body_comment,
+    post_comment_to_github,
+)
+from task_processor.decorators import register_task_handler
+from webhooks.webhooks import WebhookEventType
+
+logger = logging.getLogger(__name__)
+
+
+class GithubResourceType(Enum):
```
gagantrivedi: Where is this being used?

gagantrivedi (changes requested):

api/features/tasks.py
novakzaballa: Logic moved to GitHub module

api/integrations/github/constants.py
novakzaballa: These variables won't change; they can be constants

api/external_resources/models.py
novakzaballa: Changed

api/external_resources/views.py
novakzaballa: Do you mean to UPDATE_FEATURE_STATE permission?

api/external_resources/models.py
novakzaballa: I will check the unique constraint.

api/integrations/github/tasks.py
novakzaballa: This was part of an old implementation, removed.

api/integrations/github/constants.py
gagantrivedi: Think the GITHUB_APP_ID will change if it's used in private cloud?

novakzaballa: Yes, I will move this out to external config and read it in the settings like the rest of external configurations

novakzaballa: GITHUB_APP_ID moved to settings

api/external_resources/models.py
novakzaballa: You were right, this should be a compound constraint on the fields `feature` and `url` since we want to allow any URL to be linked to several Features. Corrected.

api/conftest.py
```diff
@@ -650,3 +652,32 @@ def flagsmith_environments_v2_table(dynamodb: DynamoDBServiceResource) -> Table:
         ],
         BillingMode="PAY_PER_REQUEST",
     )
+
+
+@pytest.fixture()
+def external_resource(feature: Feature) -> FeatureExternalResource:
```
gagantrivedi: Should we rename this fixture to `feature_external..`

api/features/feature_external_resources/views.py
```diff
@@ -0,0 +1,21 @@
+from rest_framework import viewsets
+from rest_framework.permissions import IsAuthenticated
+
+from .models import FeatureExternalResource
+from .serializers import FeatureExternalResourceSerializer
+
+
+class FeatureExternalResourceViewSet(viewsets.ModelViewSet):
+    serializer_class = FeatureExternalResourceSerializer
+    permission_classes = [IsAuthenticated]
```
gagantrivedi: This should use update feature permission

api/features/feature_external_resources/views.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def get_queryset(self):
+        if "pk" in self.kwargs:
```
gagantrivedi, line 19: Can we do a check here to make sure the user has configured GitHub integration before they can create external resource?

api/integrations/github/client.py
```diff
@@ -0,0 +1,12 @@
+from django.conf import settings
+from github import Auth, Github
+
+
+def generate_token(installation_id, app_id):
```
gagantrivedi: Missing type hints

api/integrations/github/github.py
```diff
@@ -0,0 +1,146 @@
...
+from features.models import FeatureState, FeatureStateValue
+from integrations.github.client import generate_token
+from integrations.github.constants import GITHUB_API_URL
+from integrations.github.models import GithubConfiguration
+from webhooks.webhooks import WebhookEventType
+
+logger = logging.getLogger(__name__)
+
+
+def post_comment_to_github(installation_id, owner, repo, issue, body):
```
gagantrivedi: Missing type hints

api/features/tasks.py
```diff
@@ -50,6 +57,39 @@ def trigger_feature_state_change_webhooks(
         )
     )
 
+    if hasattr(instance.environment.project.organisation, "github_config"):
```
gagantrivedi: Instead of this, we should check if the feature has any external resource or not, that will help us reduce the number of tasks that don't do anything useful

api/integrations/github/client.py
```diff
@@ -0,0 +1,12 @@
...
+    auth = Auth.AppAuth(int(app_id), settings.GITHUB_PEM).get_installation_auth(
+        int(installation_id),
+        None,
+    )
+    Github(auth=auth)
+    token = auth.token
```
gagantrivedi, line 13: Do we know how long this token is valid for? We can maybe keep it in cache if it is a long-lived token?

api/integrations/github/permissions.py
```diff
@@ -0,0 +1,15 @@
...
+    Custom permission to only allow users with permission to access
+    GithubConfiguration related to their organisations.
+    """
+
+    def has_permission(self, request, view):
+        if not request.user.is_authenticated:
+            return False
+
+        organisation_id = view.kwargs.get("organisation_pk")
+        return request.user.organisations.filter(id=organisation_id).exists()
```
gagantrivedi: We should use `belong_to` method here

api/integrations/github/models.py
```diff
@@ -0,0 +1,39 @@
...
+    project = models.ForeignKey(
+        "projects.Project",
+        related_name="github_project",
+        blank=False,
+        null=False,
+        on_delete=models.CASCADE,
+    )
+
+    class Meta:
+        unique_together = (
```
gagantrivedi: Let's use `.UniqueConstraint` here
- Add a unit test to make sure user breaking this constraint results in 400

api/integrations/github/models.py
```diff
@@ -0,0 +1,39 @@
...
+        Organisation, on_delete=models.CASCADE, related_name="github_config"
+    )
+    installation_id = models.CharField(max_length=100, blank=False, null=False)
+
+
+class GithubRepository(SoftDeleteExportableModel):
+    github_configuration = models.ForeignKey(
+        GithubConfiguration, related_name="repository_config", on_delete=models.CASCADE
+    )
+    repository_owner = models.CharField(max_length=100, blank=False, null=False)
```
gagantrivedi, line 28: Why do we need `repository_owner`? Is this being used somewhere?

api/tests/unit/external_resources/test_unit_external_resources_views.py
```diff
@@ -0,0 +1,136 @@
+import simplejson as json
```
gagantrivedi: This should also be nested under features module

api/tests/unit/external_resources/test_unit_external_resources_views.py
```diff
@@ -0,0 +1,136 @@
...
+    assert len(db_record) == 1
+    assert db_record[0].metadata == json.dumps(
+        external_resource_data["metadata"], default=_django_json_encoder_default
+    )
+    assert db_record[0].feature == feature
+    assert db_record[0].type == external_resource_data["type"]
+    assert db_record[0].url == external_resource_data["url"]
+
+    # And When
+    url = reverse(
```
gagantrivedi: I think this is covered by another test?

api/tests/unit/external_resources/test_unit_external_resources_views.py
```diff
@@ -0,0 +1,136 @@
...
+    assert response.status_code == status.HTTP_200_OK
+    assert response.json()["count"] == 1
+    assert response.json()["results"][0]["type"] == external_resource_data["type"]
+    assert response.json()["results"][0]["url"] == external_resource_data["url"]
+    assert (
+        response.json()["results"][0]["metadata"] == external_resource_data["metadata"]
+    )
+
+
+def test_cannot_create_external_resources(
```
gagantrivedi: We should update the name to include why cannot create

api/tests/unit/external_resources/test_unit_external_resources_views.py
```diff
@@ -0,0 +1,136 @@
...
+from django.urls import reverse
+from rest_framework import status
+from rest_framework.test import APIClient
+
+from features.feature_external_resources.models import FeatureExternalResource
+from features.models import Feature
+
+_django_json_encoder_default = DjangoJSONEncoder().default
+
+
```
gagantrivedi: Can we add a unit test that checks that the violating following constraint does not result in 500

```
constraints = [
            models.UniqueConstraint(
                fields=["feature", "url"], name="unique_feature_url_constraint"
            )
        ]
```

api/tests/unit/external_resources/test_unit_external_resources_views.py
```diff
@@ -0,0 +1,136 @@
...
+def test_create_external_resources(
+    admin_client: APIClient,
```
gagantrivedi: Can we add `master_api_key_client` to all these tests as well?

api/integrations/github/views.py
```diff
@@ -0,0 +1,138 @@
...
+@permission_classes([IsAuthenticated, HasPermissionToGithubConfiguration])
+def fetch_pull_requests(request, organisation_pk):
+    organisation = Organisation.objects.get(id=organisation_pk)
+
+    token = generate_token(
+        organisation.github_config.installation_id,
+        settings.GITHUB_APP_ID,
+    )
+
+    repo_owner = request.query_params.get("repo_owner", "")
```
gagantrivedi: We should use a query serializer for this

api/features/tasks.py
```diff
@@ -50,6 +57,39 @@ def trigger_feature_state_change_webhooks(
...
+        feature_data["feature_states"].append(feature_state)
+
+        call_github_app_webhook_for_feature_state.delay(
+            args=(
+                feature_data,
+                event_type.value,
+            ),
+        )
+    else:
+        logger.warning(
```
gagantrivedi: I don't think there is any use for this warning?

api/integrations/github/github.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+def generate_data(
+    github_configuration: GithubConfiguration,
+    feature_id: int,
+    feature_name: str,
+    type: str,
+    feature_states: typing.Union[list[FeatureState], list[FeatureStateValue]] = None,
+    url: str = None,
+):
```
gagantrivedi: We should define a data class/or some sort of type that this function returns

api/integrations/github/github.py
```diff
@@ -0,0 +1,146 @@
...
+    feature_data["installation_id"] = github_configuration.installation_id
+    feature_data["organisation_id"] = github_configuration.organisation.id
+
+    if type == WebhookEventType.FEATURE_EXTERNAL_RESOURCE_REMOVED.value:
+        feature_data["url"] = url
+    else:
+        feature_data["feature_states"] = []
+        for feature_state in feature_states:
+            feature_env_data = {}
+            if check_not_none(feature_state.feature_state_value.string_value):
```
gagantrivedi: Why do we need `check_not_none` function? Can't we just do `feature_state.feature_state_value.string_value`?

api/integrations/github/github.py
```diff
@@ -0,0 +1,146 @@
...
+        )
+
+        url = f"{GITHUB_API_URL}repos/{owner}/{repo}/issues/{issue}/comments"
+        headers = {
+            "Accept": "application/vnd.github.v3+json",
+            "Authorization": f"Bearer {token}",
+        }
+
+        payload = {"body": body}
+        response = requests.post(url, json=payload, headers=headers)
```
gagantrivedi: We should add timeout to all the requests that we are making

api/features/feature_external_resources/models.py
```diff
@@ -0,0 +1,95 @@
...
+        ordering = ("id",)
+        constraints = [
+            models.UniqueConstraint(
+                fields=["feature", "url"], name="unique_feature_url_constraint"
+            )
+        ]
+
+    @hook(AFTER_SAVE)
+    def exectute_after_save_actions(self):
+        # Add a comment to GitHub Issue/PR when feature is linked to the GH external resource
```
gagantrivedi, line 48: We need to add tests for this module

api/features/feature_external_resources/models.py
```diff
@@ -0,0 +1,95 @@
...
+        GITHUB_PR = "GITHUB_PR", _("GitHub PR")
+
+    url = models.URLField()
+    type = models.CharField(max_length=20, choices=ResourceType.choices)
+
+    # JSON filed containing any metadata related to the external resource
+    metadata = models.TextField(null=True)
+    feature = models.ForeignKey(
+        Feature,
+        related_name="features",
```
gagantrivedi: This should be `external_resources`

api/integrations/github/tasks.py
```diff
@@ -0,0 +1,89 @@
+import logging
```
gagantrivedi, line 1: We need to add tests for this module

api/integrations/github/github.py
```diff
@@ -0,0 +1,146 @@
...
+
+    feature_data["installation_id"] = github_configuration.installation_id
+    feature_data["organisation_id"] = github_configuration.organisation.id
+
+    if type == WebhookEventType.FEATURE_EXTERNAL_RESOURCE_REMOVED.value:
+        feature_data["url"] = url
+    else:
+        feature_data["feature_states"] = []
+        for feature_state in feature_states:
+            feature_env_data = {}
```
gagantrivedi, line 119: Also, why are we not using `get_feature_state_value` here?

api/features/tasks.py
gagantrivedi: We need tests for this code

api/organisations/urls.py
```diff
@@ -133,7 +172,7 @@
     )
 
     nested_roles_router = routers.NestedSimpleRouter(
-        organisations_router, r"roles", lookup="role"
+        organisations_router, parent_prefix=r"roles", lookup="role"
```
gagantrivedi: Why are we updating roles router here?

gagantrivedi (changes requested):

api/conftest.py
novakzaballa: All occurrences have been renamed.

api/integrations/github/models.py
novakzaballa: Using UniqueConstraint now,  and a test was added

api/integrations/github/models.py
novakzaballa: We use this data in the UI

api/tests/unit/external_resources/test_unit_external_resources_views.py
novakzaballa: Done.

api/tests/unit/external_resources/test_unit_external_resources_views.py
novakzaballa: Test was added

api/integrations/github/client.py
novakzaballa: Added

api/integrations/github/github.py
novakzaballa: Added

api/integrations/github/github.py
novakzaballa: Timeouts added.

api/organisations/urls.py
novakzaballa: It seems to be the result of a merge with some branch, corrected.

api/integrations/github/views.py
novakzaballa: Serializers created

api/tests/unit/external_resources/test_unit_external_resources_views.py
novakzaballa: Changed

api/tests/unit/external_resources/test_unit_external_resources_views.py
novakzaballa: Moved.

api/features/feature_external_resources/views.py
novakzaballa: DOne

api/tests/unit/external_resources/test_unit_external_resources_views.py
novakzaballa: Yes, in another test there was also a get request

api/integrations/github/permissions.py
novakzaballa: Added

api/features/feature_external_resources/views.py
novakzaballa: Added

api/integrations/github/client.py
novakzaballa: The documentation says it's for an hour. Do you think it's a good option?

api/features/tasks.py
novakzaballa: Perhaps while the PR isn't approved, this logger can provide us with additional information.

api/features/tasks.py
novakzaballa: Changed

api/features/feature_external_resources/models.py
novakzaballa: Renamed

api/tests/unit/integrations/github/test_unit_github_views.py
```diff
@@ -0,0 +1,148 @@
+import pytest
+from django.db import IntegrityError
```
gagantrivedi: We need to add master API key client to tests in this module as well

api/users/models.py
```diff
@@ -163,6 +163,9 @@ def get_full_name(self):
             return None
         return " ".join([self.first_name, self.last_name]).strip()
 
+    def belong_to(self) -> bool:
```
gagantrivedi: Wait, why did we add this method?

api/tests/unit/integrations/github/test_unit_github_views.py
novakzaballa: OK, Added.

api/users/models.py
novakzaballa: Sorry, for some reason I didn't see that one already existed, I already replaced and removed it.

api/integrations/github/github.py
novakzaballa: Type added

api/features/feature_external_resources/views.py
```diff
@@ -0,0 +1,37 @@
...
+            return FeatureExternalResource.objects.filter(feature=features_pk)
+
+    def create(self, request, *args, **kwargs):
+        feature = get_object_or_404(
+            Feature.objects.filter(
+                id=self.kwargs["feature_pk"],
+            ),
+        )
+        if hasattr(feature.project, "github_config"):
+            raise PermissionDenied("You don't have a GitHub config")
```
gagantrivedi: I think 400 would be better here instead of 403 (returned by permission denied)

gagantrivedi: I'd just move this to the serializer and raise validation Error

api/integrations/github/github.py
novakzaballa: I put this check to ensure that the data doesn't fail when being created.

api/features/feature_external_resources/models.py
novakzaballa: I can't find a reference test for hooks, could you refer to some parts of the code?

api/integrations/github/tasks.py
novakzaballa: I think this will be included in the hooks test above.

api/integrations/github/github.py
novakzaballa: I'll see how it works

api/features/feature_external_resources/views.py
novakzaballa: I prefer to keep it here and not move it to the serializer since it is not validating directly a field received in the request.

api/integrations/github/tasks.py
novakzaballa: All these functions are tested in the `create feature external resources test`

api/features/feature_external_resources/models.py
gagantrivedi: https://github.com/Flagsmith/flagsmith/blob/main/api/tests/unit/features/test_unit_features_models.py#L236

api/tests/unit/integrations/github/test_unit_github_views.py
```diff
@@ -0,0 +1,335 @@
...
+        },
+        timeout=10,
+    )
+
+
+@pytest.mark.parametrize(
+    "client, reverse_url",
+    [
+        (
+            lazy_fixture("admin_master_api_key_client"),
```
gagantrivedi, line 345: Why are we using different URL for different client?

api/integrations/github/views.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def get_queryset(self):
+        return GithubRepository.objects.filter(
+            github_configuration=self.kwargs["github_pk"]
+        )
+
+
+@api_view(["GET"])
+@permission_classes([IsAuthenticated, HasPermissionToGithubConfiguration])
+@github_auth_required
```
gagantrivedi, line 95: Is there a reason why we are not using class based views here?

api/tests/unit/features/test_unit_feature_external_resources_views.py
```diff
@@ -0,0 +1,318 @@
...
+        "type": "GITHUB_ISSUE",
+        "url": "https://github.com/userexample/example-project-repo/issues/11",
+        "feature": feature.id,
+        "environment": environment.id,
+    }
+    url = reverse("api-v1:features:external-resources-list", args=[feature.id])
+
+    # When
+    with pytest.raises(IntegrityError) as exc_info:
+        response = client.post(url, data=feature_external_resource_data)
```
gagantrivedi: I think we should catch this exception

api/integrations/github/views.py
```diff
@@ -0,0 +1,166 @@
...
+        response = requests.get(url, headers=headers, timeout=10)
+        response.raise_for_status()
+        data = response.json()
+        return Response(data)
+    except requests.RequestException as e:
+        return JsonResponse({"error": str(e)}, status=500)
+
+
+@api_view(["GET"])
+@permission_classes([IsAuthenticated, HasPermissionToGithubConfiguration])
```
gagantrivedi, line 129: Looks like we are using admin_client(admin_user) for all the unit tests... can we write a test to make sure these permissions are working? i.e: a user without these permissions should get 403

api/features/permissions.py
```diff
@@ -124,16 +125,27 @@ def has_permission(self, request: Request, view: GenericViewSet) -> bool:
     def has_object_permission(
         self, request: Request, view: GenericViewSet, obj: FeatureState
     ) -> bool:
-        permission = (
-            MANAGE_SEGMENT_OVERRIDES if obj.feature_segment_id else UPDATE_FEATURE_STATE
-        )
+        if isinstance(obj, FeatureExternalResource):
```
gagantrivedi: I think we should extend this permission class instead of modifying it?

gagantrivedi (changes requested):

api/features/permissions.py
novakzaballa: I created a new permission for the featureExternalREsourceView inheriting from the FeaturePermission.

api/tests/unit/integrations/github/test_unit_github_views.py
novakzaballa: Good catch! we want to test both URLs with both clients. Correcting.

api/features/permissions.py
novakzaballa: Right. After talking with Matt to understand better the permissions schema, I have changed the base permission to `FeaturePermissions`, and extended it as `FeatureExternalResourcePermissions`

api/integrations/github/views.py
novakzaballa: Yes, it is because this view will have only `GET ` method always and I have seen that we use the `@api_view` decorator in several other similar views.

api/tests/unit/features/test_unit_feature_external_resources_views.py
novakzaballa: Nice. Done.

api/integrations/github/views.py
novakzaballa: Test Added.

api/integrations/github/github.py
novakzaballa: OK, now using `get_feature_state_value`

api/integrations/github/views.py
novakzaballa: Tests added

api/external_resources/models.py
novakzaballa: This has been discussed with Matt and for now, this is generic enough considering the only additional case of use for external resources which is JIRA integration. In the future, it could be necessary to extend it as discussed.

api/features/tasks.py
novakzaballa: This code is being tested here https://github.com/Flagsmith/flagsmith/pull/3188/files#diff-882717ba3a21107bfb44676303dba09977ecbddf0b5894557c25c4facf820851R40

api/integrations/github/github.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+def generate_data(
+    github_configuration: GithubConfiguration,
+    feature_id: int,
+    feature_name: str,
+    type: str,
+    feature_states: typing.Union[list[FeatureState], list[FeatureStateValue]] = None,
+    url: str = None,
+) -> dict[str, typing.Any]:
```
gagantrivedi: Can we convert the response this functions returns to a data class?

gagantrivedi (changes requested):

api/integrations/github/github.py
novakzaballa: Using a data class now

api/integrations/github/views.py
```diff
@@ -0,0 +1,166 @@
...
+        response.raise_for_status()
+        data = response.json()
+        filtered_data = [issue for issue in data if "pull_request" not in issue]
+        return Response(filtered_data)
+    except requests.RequestException as e:
+        return JsonResponse({"error": str(e)}, status=500)
+
+
+@api_view(["GET"])
+def fetch_repositories(request):
```
gagantrivedi: Should we add permissions to this method as well, or is this intentional?

api/integrations/github/github.py
```diff
@@ -0,0 +1,153 @@
...
+
+@dataclass
+class GithubData:
+    installation_id: str
+    feature_id: int
+    feature_name: str
+    feature_states: typing.List[dict[str, typing.Any]] = None
+    url: str = None
+
+    def to_dict(self) -> typing.Dict[str, typing.Any]:
```
gagantrivedi: Can we use `asdict` here instead of defining this?

api/integrations/github/github.py
```diff
@@ -0,0 +1,153 @@
...
+        }
+
+        payload = {"body": body}
+        response = response = requests.post(
+            url, json=payload, headers=headers, timeout=10
+        )
+
+        return response.json() if response.status_code == 200 else None
+    except requests.RequestException as e:
+        logger.info(f" {e}")
```
gagantrivedi: We should not log this as info

api/features/feature_external_resources/models.py
```diff
@@ -0,0 +1,95 @@
...
+                feature_states,
+            )
+
+            call_github_app_webhook_for_feature_state.delay(
+                args=(
+                    feature_data.to_dict(),
+                    WebhookEventType.FEATURE_EXTERNAL_RESOURCE_ADDED.value,
+                ),
+            )
+        else:
```
gagantrivedi: We can get rid of this else

api/features/feature_external_resources/models.py
```diff
@@ -0,0 +1,95 @@
...
+                url=self.url,
+            )
+
+            call_github_app_webhook_for_feature_state.delay(
+                args=(
+                    feature_data.to_dict(),
+                    WebhookEventType.FEATURE_EXTERNAL_RESOURCE_REMOVED.value,
+                ),
+            )
+        else:
```
gagantrivedi: We can get rid of this else

api/integrations/github/github.py
```diff
@@ -0,0 +1,153 @@
...
+) -> str:
+
+    is_update = event_type == WebhookEventType.FLAG_UPDATED.value
+    is_removed = event_type == WebhookEventType.FEATURE_EXTERNAL_RESOURCE_REMOVED.value
+    delete_text = f"### The feature flag {name} was unlinked from the issue/PR"
+
+    if is_removed:
+        return delete_text
+
+    last_updated_string = (
```
gagantrivedi: Can we move these strings to constant (sort of like we have for our audit log messages)?

api/integrations/github/permissions.py
```diff
@@ -0,0 +1,18 @@
...
+    Custom permission to only allow users with permission to access
+    GithubConfiguration related to their organisations.
+    """
+
+    def has_permission(self, request, view):
+
+        if isinstance(request.user, FFAdminUser):
+            organisation_id = view.kwargs.get("organisation_pk")
+            return request.user.belongs_to(organisation_id=organisation_id)
+        else:
```
gagantrivedi: We can get rid of this else statement here(`belongs_to` is implemented for `APIKeyUser` as well)

api/tests/unit/integrations/github/test_unit_github_views.py
```diff
@@ -0,0 +1,392 @@
...
+        args=[organisation.id, github_configuration.id],
+    )
+
+    response = client.post(url, data)
+
+    assert response.status_code == status.HTTP_201_CREATED
+    assert GithubRepository.objects.filter(repository_owner="repositoryowner").exists()
+
+
+def test_cannot_create_github_repository(
```
gagantrivedi: Can we update this test name to reflect why it cannot create GitHub repository

api/tests/unit/integrations/github/test_unit_github_views.py
```diff
@@ -0,0 +1,392 @@
...
+@pytest.mark.parametrize(
+    "client",
+    [lazy_fixture("admin_master_api_key_client"), lazy_fixture("admin_client")],
+)
+def test_create_github_repository(
+    client: APIClient,
+    organisation: Organisation,
+    github_configuration: GithubConfiguration,
+    project: Project,
+) -> None:
```
gagantrivedi, line 106: nit: can we add Give/when/then

api/tests/unit/integrations/github/test_unit_github_views.py
```diff
@@ -0,0 +1,392 @@
...
+    "client",
+    [lazy_fixture("admin_master_api_key_client"), lazy_fixture("admin_client")],
+)
+def test_github_delete_repository(
+    client: APIClient,
+    organisation: Organisation,
+    github_configuration: GithubConfiguration,
+    github_repository: GithubRepository,
+) -> None:
+    url = reverse(
```
gagantrivedi, line 198: nit: can we add Give/when/then

api/features/feature_external_resources/models.py
```diff
@@ -0,0 +1,95 @@
...
+        GITHUB_PR = "GITHUB_PR", _("GitHub PR")
+
+    url = models.URLField()
+    type = models.CharField(max_length=20, choices=ResourceType.choices)
+
+    # JSON filed containing any metadata related to the external resource
+    metadata = models.TextField(null=True)
+    feature = models.ForeignKey(
+        Feature,
+        related_name="feature_external_resources",
```
gagantrivedi: Can we change the related name to `external_resources`? (`feature` feels a bit redundant)

api/features/feature_external_resources/models.py
```diff
@@ -0,0 +1,95 @@
...
+            feature_states = FeatureState.objects.filter(feature_id=self.feature_id)
+            feature_data: GithubData = generate_data(
+                github_configuration,
+                self.feature_id,
+                self.feature.name,
+                WebhookEventType.FEATURE_EXTERNAL_RESOURCE_ADDED.value,
+                feature_states,
+            )
+
+            call_github_app_webhook_for_feature_state.delay(
```
gagantrivedi, line 61: Is there a reason why we are passing event_type explicitly here as well? It is already part of `GithubData`?

api/tests/unit/integrations/github/test_unit_github_views.py
```diff
@@ -0,0 +1,392 @@
...
+        "project": project.id,
+    }
+
+    url = reverse(
+        "api-v1:organisations:repositories-list",
+        args=[organisation.id, github_configuration.id],
+    )
+
+    # When
+    with pytest.raises(IntegrityError) as exc_info:
```
gagantrivedi: We can get rid of this and just assert the response text

api/tests/unit/integrations/github/test_unit_github_views.py
```diff
@@ -0,0 +1,392 @@
...
+    assert GithubRepository.objects.filter(repository_owner="repositoryowner").exists()
+
+
+def test_cannot_create_github_repository(
+    test_user_client: APIClient,
+    organisation: Organisation,
+    github_configuration: GithubConfiguration,
+    project: Project,
+) -> None:
+    data = {
```
gagantrivedi, line 134: nit: given/when/then

api/tests/unit/integrations/github/test_unit_github_views.py
```diff
@@ -0,0 +1,392 @@
...
+
+
+@pytest.mark.parametrize(
+    "reverse_url",
+    [
+        ("api-v1:organisations:get-github-issues"),
+        ("api-v1:organisations:get-github-issues"),
+    ],
+)
+def test_cannot_fetch_issues_or_prs_when_does_not_have_permissions(
```
gagantrivedi, line 379: Can we add this permission test for other endpoints as well

api/tests/unit/integrations/github/test_unit_github_views.py
```diff
@@ -0,0 +1,392 @@
...
+    response = client.get(url)
+
+    # Then
+    assert response.status_code == 400
+
+
+@pytest.mark.parametrize(
+    "reverse_url",
+    [
+        ("api-v1:organisations:get-github-issues"),
```
gagantrivedi, line 375: Why is the same URL defined twice

api/tests/unit/features/test_unit_feature_external_resources_views.py
```diff
@@ -0,0 +1,316 @@
...
+        headers={
+            "Accept": "application/vnd.github.v3+json",
+            "Authorization": "Bearer mocked_token",
+        },
+        timeout=10,
+    )
+    # Then
+    assert response.status_code == status.HTTP_201_CREATED
+    # assert that the payload has been save to the database
+    db_record = FeatureExternalResource.objects.filter(
```
gagantrivedi: nit: can we rename this variable

gagantrivedi (changes requested):

api/features/feature_external_resources/models.py
novakzaballa: I already tested this code in the create_feature_external_resources test and delete_feature_external_resources.

api/tests/unit/features/test_unit_feature_external_resources_views.py
novakzaballa: Renamed to `feature_external_resources`.

api/tests/unit/integrations/github/test_unit_github_views.py
novakzaballa: Corrected

api/tests/unit/integrations/github/test_unit_github_views.py
novakzaballa: I think it already exists, the problem was the name, renamed.

api/tests/unit/integrations/github/test_unit_github_views.py
novakzaballa: Added

api/tests/unit/integrations/github/test_unit_github_views.py
novakzaballa: Done.

api/features/feature_external_resources/models.py
novakzaballa: There is no explicit reason, I moved it inside GithubData now.

api/features/feature_external_resources/models.py
novakzaballa: Changed

api/tests/unit/integrations/github/test_unit_github_views.py
novakzaballa: Added.

api/tests/unit/integrations/github/test_unit_github_views.py
novakzaballa: Added

api/tests/unit/integrations/github/test_unit_github_views.py
novakzaballa: Changed.

api/integrations/github/permissions.py
novakzaballa: Implemented

api/integrations/github/github.py

api/features/feature_external_resources/models.py
novakzaballa: Removed

api/features/feature_external_resources/models.py
novakzaballa: ok, I think at this point it is correct to remove it, removed

api/integrations/github/github.py
novakzaballa: Updated to error

api/integrations/github/github.py
novakzaballa: Removed, and implemented.

api/integrations/github/views.py
novakzaballa: This was done intentionally, as in some cases, due to the installation flow, the only thing that will be obtained is the GitHub installation ID, and this is sufficient to fetch the repositories.

novakzaballa: Sorry, I misunderstood the observation. I will add permissions checking here. I think we should request organisation admin, but I will double check with Matt whether we also want to allow  a project level permission like we do with the Slack integration.

novakzaballa: I added is_admin permission

gagantrivedi (approved):
Great work, @novakzaballa, and thanks for all the patience :smile:

novakzaballa:
> Great work, @novakzaballa, and thanks for all the patience 😄

Thank you @gagantrivedi!

PR #3199, opened 2023-12-20
fix(infra/staging): Add INFLUXDB_BUCKET to task def

gagantrivedi:

## Changes

Remove runtime check to configure influx in staging correctly(because of SSE tracking)
Add `INFLUXDB_BUCKET` environment variable to staging task processor task definition 

Covered by unit tests

### DIFF
--- a/api/app/settings/common.py
+++ b/api/app/settings/common.py
@@ -295,7 +295,7 @@ INFLUXDB_ORG = env.str("INFLUXDB_ORG", default="")
 USE_POSTGRES_FOR_ANALYTICS = env.bool("USE_POSTGRES_FOR_ANALYTICS", default=False)
 
 # NOTE: Because we use Postgres for analytics data in staging and Influx for tracking SSE data,
-# we need to support both configurations at the same time.if USE_POSTGRES_FOR_ANALYTICS:
+# we need to support setting the influx configuration alongside using postgres for analytics.
 if USE_POSTGRES_FOR_ANALYTICS:
     MIDDLEWARE.append("app_analytics.middleware.APIUsageMiddleware")
 elif INFLUXDB_TOKEN:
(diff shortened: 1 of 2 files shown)


khvn26 (approved):

api/app/settings/common.py
```diff
@@ -291,14 +291,12 @@
...
-    MIDDLEWARE.append("app_analytics.middleware.InfluxDBMiddleware")
 
 if USE_POSTGRES_FOR_ANALYTICS:
-    if INFLUXDB_BUCKET:
-        raise RuntimeError("Cannot use both InfluxDB and Postgres for analytics")
-
     MIDDLEWARE.append("app_analytics.middleware.APIUsageMiddleware")
+elif INFLUXDB_TOKEN:
+    MIDDLEWARE.append("app_analytics.middleware.InfluxDBMiddleware")
+
```
matthewelwell: As per discussions in slack, I think we should add a comment to try and explain the logic here. I think it would be good to logically group all of the influx settings here too.

matthewelwell (changes requested):

api/app/settings/common.py
gagantrivedi: Yeah, Done

api/app/settings/common.py
```diff
@@ -288,17 +278,29 @@
...
+INFLUXDB_URL = env.str("INFLUXDB_URL", default="")
+INFLUXDB_ORG = env.str("INFLUXDB_ORG", default="")
 
-if USE_POSTGRES_FOR_ANALYTICS:
-    if INFLUXDB_BUCKET:
-        raise RuntimeError("Cannot use both InfluxDB and Postgres for analytics")
+USE_POSTGRES_FOR_ANALYTICS = env.bool("USE_POSTGRES_FOR_ANALYTICS", default=False)
 
+# NOTE: Because we use Postgres for analytics data in staging and Influx for tracking SSE data,
+# we need to support both configurations at the same time.if USE_POSTGRES_FOR_ANALYTICS:
```
matthewelwell:
```suggestion
# we need to support setting the influx configuration alongside using postgres for analytics.
```

matthewelwell (approved):

PR #3206, opened 2023-12-20
chore: add health checks to Docker Compose file

dreglad:

----

This PR adds a health check to `postgres` service and some other minor fixes.

## Changes

- Added a health check configuration to the `flagsmith` and `postgres` services in the dependency setup. This ensures that the Flagsmith container starts only when the database is fully operational. Previously, the depends_on directive primarily influenced the startup sequence of containers, which isn't very effective in this context.
- Removed the `link_to` fields from both services. These fields are not only redundant in this scenario but are also deprecated, further justifying their removal.
- Eliminated `flagsmith` from the `depends_on` list in the `flagsmith_processor` service. There is no direct dependency between them that necessitates this linkage.
- Omitted unnecessary quotes in YAML strings for cleaner syntax.
- Standardized the usage of single quotes, replacing any instances of mixed double/single quotes for consistency.
- Minor spelling fixes

Run `docker compose up` and verified containers are still properly started

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -40,7 +40,7 @@ services:
       # ENABLE_ADMIN_ACCESS_USER_PASS: 'true' # Uncomment to enable access to the /admin/ Django backend via your username and password
       # ALLOW_REGISTRATION_WITHOUT_INVITE: 'true'
 
-      # Enable Task Processor\
+      # Enable Task Processor
       TASK_RUN_METHOD: TASK_PROCESSOR # other options are: SYNCHRONOUSLY, SEPARATE_THREAD (default)
 
       # For more info on configuring E-Mails - https://docs.flagsmith.com/deployment/locally-api#environment-variables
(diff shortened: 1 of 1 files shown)


dabeeeenster:
Thanks for this - are you able to lint the file against our `.prettierrc` in the root folder?

docker-compose.yml
```diff
@@ -26,34 +31,43 @@ services:
...
-      # EMAIL_USE_TLS: "true" # optional
-
+      # EMAIL_USE_TLS: 'true' # optional
+    healthcheck:
+      test:
+        [
+          'CMD-SHELL',
+          'python -c "url = ''http://localhost:8000/health'';import requests;import sys;status =
+          requests.get(url).status_code;sys.exit(0 if 200 >= status < 300 else 1)"',
+        ]
```
matthewelwell, line 61: Can we extract this to a script in api/scripts perhaps? Then this could become `python ./scripts/healthcheck.py` for example and the script itself can have some more whitespace so it's easier to understand.

matthewelwell (commented):
Hi @dreglad, thanks for this - it's definitely a great addition. I've added one comment for an aesthetic improvement but I've tested it and all looks good.

PR #3209, opened 2023-12-21
fix(versioning): endpoints should return latest versions

matthewelwell:
## Changes

Fixes a FE integration issue and what appears to be a legitimate bug in the logic for getting the latest versions. 

The FE integration issue was that the API to retrieve feature segments for a given environment / feature was previously returning the feature segments across all versions, rather than just the current live version as it currently does. 

The legitimate bug was that, when evaluating the flags to retrieve the current live version, it was using the id of the feature segment as the unique key for a segment override whereas this changes for each version in the new versioning logic. As such, I have modified this logic to use the id of the segment which shouldn't change (at least until we start versioning segments too!). 

Note that as part of this, I added a `tests.types` module. I also added `tests` to the `known_first_party` attribute in isort configuration and hence there are a number of other file changes. 

Added unit tests which cover the specific functionality which exhibited the issues while integrating with the FE.

### DIFF
--- a/api/conftest.py
+++ b/api/conftest.py
@@ -60,6 +60,16 @@ from segments.models import Condition, Segment, SegmentRule
 from task_processor.task_run_method import TaskRunMethod
 from users.models import FFAdminUser, UserPermissionGroup
 
+# TODO:
+#   - move this file to the `tests/` dir and move these definitions to a `types.py` module.
+#   - migrate all existing typehints to use these
+WithProjectPermissionsCallable = typing.Callable[
+    [list[str], Project | None], UserProjectPermission
+]
+WithEnvironmentPermissionsCallable = typing.Callable[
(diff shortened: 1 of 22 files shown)


api/conftest.py
```diff
@@ -58,6 +58,10 @@
 from projects.tags.models import Tag
 from segments.models import Condition, Segment, SegmentRule
 from task_processor.task_run_method import TaskRunMethod
+from tests.types import (
+    WithEnvironmentPermissionsCallable,
+    WithProjectPermissionsCallable,
+)
```
zachaysan, line 64: I love these types, but can your PR include them in all present test function types? It should be an easy replace and then they'd be fixed everywhere.

matthewelwell: Yep, on it.

khvn26 (approved):

PR #3218, opened 2023-12-22
feat(audit): add change details to AuditLog

matthewelwell:
## Changes

This PR adds the change details to the API when retrieving a record from the AuditLog. To achieve this, I have added a new endpoint to retrieve an individual AuditLog record which includes a new `change_details` attribute in the JSON response. 

The `change_details` attribute contains a list, representing fields that have changed, e.g: 

```json
    ...
    "change_details": [
        {
            "field": "enabled",
            "new": true,
            "old": false
        }
    ]
    ...
```

Added 'integration' style tests for common use cases.


api/audit/models.py
```diff
@@ -79,9 +79,9 @@ def environment_document_updated(self) -> bool:
         return "send_environments_to_dynamodb" not in skip_signals_and_hooks
 
     @property
-    def history_record(self):
+    def history_record(self) -> typing.Optional[Model]:
         klass = self.get_history_record_model_class(self.history_record_class_path)
-        return klass.objects.get(id=self.history_record_id)
+        return klass.objects.filter(history_id=self.history_record_id).first()
```
khvn26, line 84: The only test that had to change is the one where the model is mocked. I'd surely prefer to have a more integrated test for this.

api/audit/serializers.py
```diff
@@ -23,8 +53,17 @@ class Meta:
             "related_object_id",
             "related_object_type",
             "is_system_event",
+            "change_details",
         )
 
+    @swagger_serializer_method(
+        serializer_or_field=AuditLogChangeDetailsSerializer(many=True)
+    )
+    def get_change_details(self, instance: AuditLog) -> typing.List[typing.Dict]:
```
khvn26:
```suggestion
    def get_change_details(self, instance: AuditLog) -> typing.List[typing.Dict[str, typing.Any]:
```

api/audit/models.py
khvn26: I stand corrected, this code gets called by the new tests.

khvn26 (approved):

api/audit/models.py
khvn26: As discussed offline, needs to change `first` to `get` to avoid returning `None`, though.

matthewelwell: I took a slightly different approach and handled the `None` return from this method in the calling method instead but this should be resolved now.

PR #3229, opened 2024-01-03
feat(tags/view): Add api to get tag by uuid

gagantrivedi:

## Changes

Add API to fetch tag by uuid - (will be) Used by our terraform provider to import tag

Adds unit test

### DIFF
--- a/api/projects/tags/views.py
+++ b/api/projects/tags/views.py
@@ -1,6 +1,9 @@
 from rest_framework import viewsets
+from rest_framework.decorators import action
 from rest_framework.generics import get_object_or_404
 from rest_framework.permissions import IsAuthenticated
+from rest_framework.request import Request
+from rest_framework.response import Response
 
 from projects.permissions import VIEW_PROJECT
 
@@ -32,3 +35,14 @@ class TagViewSet(viewsets.ModelViewSet):
     def perform_update(self, serializer):
         project_id = self.kwargs["project_pk"]
         serializer.save(project_id=project_id)
+
+    @action(
+        detail=False,
+        url_path=r"get-by-uuid/(?P<uuid>[0-9a-f-]+)",
+        methods=["get"],
+    )
+    def get_by_uuid(self, request: Request, project_pk: int, uuid: str):
+        qs = self.get_queryset()
+        tag = get_object_or_404(qs, uuid=uuid)
+        serializer = self.get_serializer(tag)
+        return Response(serializer.data)
--- a/api/projects/tags/views.py
+++ b/api/projects/tags/views.py
@@ -29,11 +29,11 @@ class TagViewSet(viewsets.ModelViewSet):
         return queryset
 
     def perform_create(self, serializer):
-        project_id = self.kwargs["project_pk"]
+        project_id = int(self.kwargs["project_pk"])
         serializer.save(project_id=project_id)
 
     def perform_update(self, serializer):
-        project_id = self.kwargs["project_pk"]
+        project_id = int(self.kwargs["project_pk"])
         serializer.save(project_id=project_id)
 
     @action(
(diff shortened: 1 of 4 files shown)


api/projects/tags/views.py
```diff
@@ -26,9 +29,20 @@ def get_queryset(self):
         return queryset
 
     def perform_create(self, serializer):
-        project_id = self.kwargs["project_pk"]
+        project_id = int(self.kwargs["project_pk"])
         serializer.save(project_id=project_id)
 
     def perform_update(self, serializer):
-        project_id = self.kwargs["project_pk"]
+        project_id = int(self.kwargs["project_pk"])
```
gagantrivedi, line 36: This is done to make sure the response have correct data type for project(i.e: int instead of string)

zachaysan: Why not use the serializer with `is_valid`?

api/tests/unit/projects/tags/test_views.py
```diff
@@ -0,0 +1,23 @@
...
+from rest_framework.test import APIClient
+
+from projects.models import Project
+from projects.tags.models import Tag
+
+
+@pytest.mark.parametrize(
+    "client",
+    [(lazy_fixture("admin_master_api_key_client")), (lazy_fixture("admin_client"))],
+)
```
zachaysan, line 17: We're prioritizing using lower levels of access over the `admin_client` when it comes to checking permissions. So since this endpoint follows `TagPermissions` you'll want to use `staff_client` with the `with_project_permissions([VIEW_PROJECT])` to set access level to match. 

Also, it would be good to check the `HTTP 403` case to make sure the endpoint isn't open to a user from a different project.

api/projects/tags/views.py
gagantrivedi: Sorry, I don't understand how that will help change the data type of the response?

api/tests/unit/projects/tags/test_views.py
gagantrivedi: That's a good point… Although I wonder if that should be done by the unit test that test the permission class itself instead of the test that test the view?

api/projects/tags/views.py
zachaysan: Getting the data type with `validated_data` after running the `is_valid` check should remove the requirement for the `int` cast, no?

api/tests/unit/projects/tags/test_views.py
zachaysan: Since permissions can be misconfigured when adding to a viewset (which we've found wide open ones because of a decorator misordering) I think testing through the view overall is a better strategy.

api/projects/tags/views.py
gagantrivedi: No, that does not work, I don't think any data passed to save() is part of validated data :thinking:  unless I am doing something wrong?

api/tests/unit/projects/tags/test_views.py
gagantrivedi: Yeah, that makes sense

api/projects/tags/views.py
zachaysan: Maybe we're talking past each other, and this is a minor thing so I'll approve the PR just to unblock you, but typically when I use serializers I'm able to call `is_valid` on the serializer then pass `validated_data` into whatever I need. This works for all sorts of situations including stuff like query params serializers. If the kwargs can be passed into the serializer and the serializer can validate the data then why do it manually with type casts?

zachaysan (approved):

api/projects/tags/views.py
gagantrivedi: I think the part that you are talking about is handled by drf(`update` method in this case) and I honestly don't see much value in overriding the update method

gagantrivedi: I will merge this for now, but let me know if disagree

PR #3233, opened 2024-01-03
feat: import rules from LD flags

Dogacel:

## Changes

- Support importing segments from launch darkly
- Support importing targeting rules for segments from launch darkly
- Support importing targeting rules for feature flags from launch darkly

There are mock responses from launch darkly for importing segments and flag rules.


api/integrations/launch_darkly/services.py
```diff
@@ -165,6 +175,87 @@ def _create_string_feature_states(
...
+            "after": constants.GREATER_THAN,
+            "semVerEqual": constants.EQUAL,
+            "semVerLessThan": constants.LESS_THAN,
+            "semVerGreaterThan": constants.GREATER_THAN,
+            "segmentMatch": constants.REGEX,
+        }
+    ).get(ld_operator, None)
+
+
+# TODO: Not sure what happens if it is not `IN` and there are multiple values.
```
khvn26: Don't have knowledge of how it's handled by LD, but isn't this the case for the `ANY` rule type?

api/integrations/launch_darkly/services.py
```diff
@@ -165,6 +175,87 @@ def _create_string_feature_states(
...
+            "greaterThan": constants.GREATER_THAN,
+            "greaterThanOrEqual": constants.GREATER_THAN_INCLUSIVE,
+            "before": constants.LESS_THAN,
+            "after": constants.GREATER_THAN,
+            "semVerEqual": constants.EQUAL,
+            "semVerLessThan": constants.LESS_THAN,
+            "semVerGreaterThan": constants.GREATER_THAN,
+            "segmentMatch": constants.REGEX,
+        }
+    ).get(ld_operator, None)
```
khvn26: Nit: no need for `dict()` here, the literal does the job already.

api/integrations/launch_darkly/services.py
```diff
@@ -165,6 +175,87 @@ def _create_string_feature_states(
...
+                property=_property,
+                value=value,
+                operator=operator,
+                created_with_segment=True,
+            )
+            logger.warning("Condition created: " + str(condition))
+        else:
+            logger.warning(
+                "Can't map launch darkly operator: " + clause["op"] + ", skipping..."
+            )
```
khvn26: I think we need to use `_log_error` here to mark the import request as failed.

api/integrations/launch_darkly/services.py
Dogacel: LD doesn't use CSV or other formats for `IN` operator, so it naturally makes sense to apply the `in` operator against a list of objects. But I am not sure how this works with operators like `startsWith`.

I guess they are `OR`ed, as anding doesn't make much sense. I am reading their SDK but haven't understood yet.

api/integrations/launch_darkly/services.py
Dogacel: We don't expect any other operators (based on SDK) but from User's perspective, isn't it better if we skip silently?

Both approaches just work fine, I believe the biggest blocker of skipping silently is how can we communicate this with the user 🙂

Dogacel: As far as I understand, there is no `not in` kind of operators so we won't be able to import everything. Basically all clauses in launch darkly are negatable by a flag, couldn't find the equivalent in Flagsmith so we definetly have to sacrifice some rules here.

api/integrations/launch_darkly/services.py
```diff
@@ -165,6 +175,105 @@ def _create_string_feature_states(
...
+        _property = clause["attribute"]
+
+        if operator is not None:
+            if clause["negate"] is True:
+                negated_child = SegmentRule.objects.create(
+                    rule=parent_rule, type=SegmentRule.NONE_RULE
+                )
+                rule = negated_child
+            else:
+                rule = child_rule
```
Dogacel: @khvn26 this block needs some attention 🙂

Dogacel: Also LMK if there are any other testing strategies than manual testing. Probably we can download the JSON responses from a test project and try to mock it here.

api/integrations/launch_darkly/client.py
```diff
@@ -35,7 +35,7 @@ def _iter_paginated_items(
         self,
         collection_endpoint: str,
         additional_params: Optional[dict[str, str]] = None,
-    ) -> Iterator[dict[str, Any]]:
+    ) -> Iterator[dict[str, Any] | str]:
```
Dogacel: Fixed this for `getTags` endpoint.

Line 48:

```python
yield from response_json.get("items") or []
```

Here we assume `items` is another struct but for get tags, items is a list of strings.

api/integrations/launch_darkly/services.py
```diff
@@ -109,15 +119,156 @@ def _create_tags_from_ld(
...
+                rule=rule,
+                property=_property,
+                value=value,
+                operator=operator,
+                created_with_segment=True,
+            )
+            logger.warning("Condition created: " + str(condition))
+        elif clause["op"] == "segmentMatch":
+            # TODO: Assign the segment to the feature
+            pass
```
Dogacel: I believe this can be done by creating a `FeatureSegment` object without creating a new segment but rather pointing to an existing one.

api/integrations/launch_darkly/services.py
```diff
@@ -156,21 +310,29 @@ def _create_string_feature_states(
 
         feature_state, _ = FeatureState.objects.update_or_create(
             feature=feature,
+            feature_segment=None,
```
Dogacel, line 693: Previously, there was only a single `FeatureState`, which allowed running import multiple times. After my changes, multiple `FeatureState` objects belong to this feature because they target a `feature_segment`.

So, we should match for states that do not belong to any segment for this piece of code. This should set the state that we see when we click the feature from UI (the main state)

api/integrations/launch_darkly/services.py
```diff
@@ -316,6 +482,38 @@ def _create_features_from_ld(
...
+    for ld_segment, env in ld_segments:
+        if ld_segment["deleted"]:
+            continue
+
+        Segment.objects.create(
+            name="imported-" + env + "-" + ld_segment["name"],
+            project_id=project_id,
+        )
+
+        # TODO: Tagging segments is not supported yet.
```
Dogacel: cc: @khvn26 Correct me if wrong.

Dogacel: Maybe we can include it in the description

api/integrations/launch_darkly/services.py
```diff
@@ -316,6 +482,38 @@ def _create_features_from_ld(
...
+    """
+
+    :param ld_segments: A list of mapping from (env, segment).
+    """
+    for ld_segment, env in ld_segments:
+        if ld_segment["deleted"]:
+            continue
+
+        Segment.objects.create(
+            name="imported-" + env + "-" + ld_segment["name"],
```
Dogacel: @khvn26 Also seems like a limitation to me. I think there are two options:
- Create a segment for each environment, let rules target the segment "for" their environment
- Create a single segment, add rules for each environment (not sure if overlaps will cause issues)

api/ld-openapi-filter.yaml
```diff
@@ -3,6 +3,8 @@ inverseOperationIds:
   - getProject
   - getEnvironmentsByProject
   - getFeatureFlags
+  - getTags
```
Dogacel, line 6: Was already being used.

api/integrations/launch_darkly/services.py
khvn26: We cover all of our code with unit tests. Refer to [existing LaunchDarkly importer unit tests](https://github.com/Flagsmith/flagsmith/blob/93df958b8fc67d17c6cfc298184bfef0f83847b4/api/tests/unit/integrations/launch_darkly) to see how mocking is organised.

khvn26: Regarding the code block in question, yeah, I was about to suggest using the `NONE_RULE` rule type for the negated LD clauses. Looks good but I expect it to be thoroughly covered with tests!

api/integrations/launch_darkly/client.py
khvn26: The correct typing fix for this would be something like:

```python
T = TypeVar("T")

...

    def _iter_paginated_items(
        self,
        collection_endpoint: str,
        item_type: Type[T],
        additional_params: Optional[dict[str, str]] = None,
    ) -> Iterator[T]:

...

    # calling code
    list_of_environments = self._iter_paginated_items(collection_endpoint=endpoint, item_type=ld_types.Environment)
    list_of_str = self._iter_paginated_items(collection_endpoint=endpoint, item_type=str)
```

api/integrations/launch_darkly/services.py
khvn26: Option 1 is the way to go. We should create distinct `Segment` model instances, and `FeatureSegment` model instances that point to the corresponding environment.

I think we should opt for more human-readable naming, e.g. `name=f'{ld_segment["name"]} (Override for {env})'`

api/integrations/launch_darkly/services.py
khvn26: Given we're now solving negated clauses with `NONE_RULE`, are there any operators left we don't support? If not, I still believe we should mark the import request as failed in this case.

In any case, IMO, we should think on informing the users on what will and will not be imported on import request creation.

This can be done either by gathering more thorough import summary data [here](https://github.com/Flagsmith/flagsmith/blob/79851ce74d26c5f712105402c131023ce3b7cc0e/api/integrations/launch_darkly/services.py#L319-L342), or, if the difference we expect is minimal, just adding the info [here](https://github.com/Flagsmith/flagsmith/blob/93df958b8fc67d17c6cfc298184bfef0f83847b4/docs/docs/integrations/importers/launchdarkly.md).

api/integrations/launch_darkly/services.py
khvn26: Add issue link https://github.com/Flagsmith/flagsmith/issues/3241 to the TODO and ignore for now.

api/integrations/launch_darkly/services.py
khvn26: So, just for completeness' sake, I'd wrap those in the `ANY_RULE`, but ultimately it's up to you. Just make sure you're using good test data for the unit tests.

api/integrations/launch_darkly/services.py
khvn26: I believe you're right. Maybe @matthewelwell has more to say here.

matthewelwell: From following this through quickly, it seems as though `"segmentMatch"` is some kind of operator in LaunchDarkly for supporting nested segments? If so, I'm not too sure how we'd support that in Flagsmith. I guess you could grab all the (supported) operators from that segment and create a new `"ALL"` rule in Flagsmith to nest each of the rules from the LD segment in the new 'monster' Flagsmith segment. Sounds like some recursion might be needed here which would get complicated real fast. 

Let me know if I've misunderstood something here.

api/integrations/launch_darkly/services.py
Dogacel: That makes sense 👍

api/integrations/launch_darkly/services.py
Dogacel: Exactly, this operator kinda allow for nested segments and allows segment belonging to be used as a condition.

> I guess you could grab all the (supported) operators from that segment 

We need to traverse segments in the right dependency order, which can be tricky to figure out. We probably gotta implement DFS and calculate nodes that are not yet calculated. A little too complex to implement 😛

Dogacel: I think generating the summary of which flags, and segments are not fully imported can be helpful for the user in this scenario. There are definitely some limitations here and clearly communicating on which flags did not work makes a lot of sense as users can go and check them 1 by 1 rather than checking everything 1 by 1 😆

matthewelwell: Yep, agreed on all of this. @khvn26 we should discuss adding this summary.

api/integrations/launch_darkly/services.py
```diff
@@ -109,15 +119,199 @@ def _create_tags_from_ld(
...
+    ld_flag_config: ld_types.FeatureFlagConfig,
+    feature: Feature,
+    environment: Environment,
+) -> None:
+    if "targets" in ld_flag_config:
+        logger.warning("Targets: " + str(ld_flag_config["targets"]))
+
+    if "contextTargets" in ld_flag_config:
+        logger.warning("Context targets: " + str(ld_flag_config["contextTargets"]))
+
```
Dogacel: @khvn26 I might need some guidence here. Which objects are storing the variations for individual targets?

api/integrations/launch_darkly/services.py
```diff
@@ -109,15 +119,199 @@ def _create_tags_from_ld(
...
+        "contains": constants.CONTAINS,
+        "lessThan": constants.LESS_THAN,
+        "lessThanOrEqual": constants.LESS_THAN_INCLUSIVE,
+        "greaterThan": constants.GREATER_THAN,
+        "greaterThanOrEqual": constants.GREATER_THAN_INCLUSIVE,
+        "before": constants.LESS_THAN,
+        "after": constants.GREATER_THAN,
+        "semVerEqual": constants.EQUAL,
+        "semVerLessThan": constants.LESS_THAN,
+        "semVerGreaterThan": constants.GREATER_THAN,
```
Dogacel, line 146: Does flagsmith automatically understand those are versions and compares them according to the semver conventions?

I saw `SemVer` stuff on UI but it seems like they are just regular gt/lt in the backend.

dabeeeenster: no - flagsmith does not try and interpret string values with gt/lt - we have specific semver-aware operators that are in addiotn to gt/lt

api/integrations/launch_darkly/services.py
Dogacel: I have double-checked, multiple values are `OR`ed. I believe this is the same behavior when a rule has multiple conditions (they are shown as `X or Y` in the UI).

Correct me if I am wrong, also implemented this as described above.

api/integrations/launch_darkly/services.py
Dogacel: I am going to leave this as a warning because we will need a better way to communicate which flags / rules are not imported successfully.

Can fix it all together with other stuff.

api/integrations/launch_darkly/services.py
Dogacel: Kinda verified this by checking the frontend after import, they are shown as `AND NOT`.

One UI issue I identified is first rule that is `NONE` is not properly shown in the UI. Probably we should show like a header that says `---- Not ----`?

wdy think?

api/integrations/launch_darkly/services.py
Dogacel: [Remind myself]: Discuss this in Wednesday's meeting. I kinda understand how this will work, but need to double check with the team as users can't be added to segments individually but rather they can only be added with property & segment rule matches.

api/integrations/launch_darkly/services.py
Dogacel: Seems like there is a conflicting view, we can go over this together Wednesday's meeting.

api/integrations/launch_darkly/services.py
```diff
@@ -109,15 +119,367 @@ def _create_tags_from_ld(
...
+    """
+    Import the individual targeting rules for the given Launch Darkly's feature flag.
+
+    :param ld_flag_config: the feature flag config from Launch Darkly.
+    :param feature: the feature to target for identities.
+    :return:
+    """
+    # TODO: Update flag states for individual targets.
+    if "targets" in ld_flag_config:
+        logger.warning("Targets: " + str(ld_flag_config["targets"]))
```
khvn26: 1. For now, just create an identity with a trait that houses the identifier, and a segment against that trait.
1. Eventually this will be migrated to using a special property name added in https://github.com/Flagsmith/flagsmith-engine/pull/198.

api/integrations/launch_darkly/services.py
```diff
@@ -303,19 +705,73 @@ def _create_features_from_ld(
...
+            _create_segment_rule_for_segment(
+                segment,
+                [],
+            )
+
+        # TODO: Import users
+        logger.warning("Included segment users: " + str(ld_segment["included"]))
+        logger.warning("Excluded segment users: " + str(ld_segment["excluded"]))
+        logger.warning("Included context users: " + str(ld_segment["includedContexts"]))
+        logger.warning("Excluded context users: " + str(ld_segment["excludedContexts"]))
```
khvn26: For user matching, add an identifier-valued trait with the same key as https://github.com/Flagsmith/flagsmith/pull/3233/files#r1447487956

For contexts, we'll have to maintain a comma-delimited trait value containing context names.

To segments, add corresponding `IN`/`NONE(IN)` rules against identity and context name containing traits.

api/integrations/launch_darkly/services.py
khvn26: You need to add a `:semver` suffix to the segment value. This will prompt the engine to use the correct matching function as seen here:

[flag_engine/utils/semver.py
](https://github.com/Flagsmith/flagsmith-engine/blob/2f09b52be8fc8a872a43cd4732b41f769e487fcb/flag_engine/utils/semver.py#L15-L16)

api/integrations/launch_darkly/client.py
```diff
@@ -57,6 +62,14 @@ def _iter_paginated_items(
...
                     endpoint=next_endpoint,
                 )
+            elif use_offset_pagination and len(items) == params["limit"]:
+                # Offset based pagination
+                offset += params["limit"]
+                params["offset"] = offset
+                response_json = self._get_json_response(
+                    endpoint=collection_endpoint,
+                    params=params,
+                )
```
khvn26: My guess is the reason that we have to calculate offsets client-side is that the `/api/v2/segments/{project_key}/{environment_key}` endpoint does not return proper `next` link, correct? In that case I suggest we migrate the other collection endpoint-consuming methods to local offset calculation and remove the `next` link consumption entirely.

I'm extremely reluctant to support both of those code paths. As they effectively serve the same purpose, we should leave only one of them.

api/integrations/launch_darkly/client.py
```diff
@@ -82,6 +95,7 @@ def get_flags(self, project_key: str) -> list[ld_types.FeatureFlag]:
         return list(
             self._iter_paginated_items(
                 collection_endpoint=endpoint,
+                additional_params={"summary": "0"},
```
khvn26, line 109: Any reason to use an integer and not boolean/boolean string here?

Also, would love a comment explaining why we're doing it (E.g., `# This makes sure rules and targets are returned along with flag configurations`).

api/integrations/launch_darkly/models.py
```diff
@@ -15,7 +15,7 @@ class LaunchDarklyImportStatus(TypedDict):
     requested_environment_count: int
     requested_flag_count: int
     result: NotRequired[Literal["success", "failure"]]
-    error_message: NotRequired[str]
+    error_messages: list[str]
```
khvn26, line 18:
```suggestion
    error_messages: NotRequired[list[str]]
```

api/integrations/launch_darkly/models.py
```diff
@@ -44,8 +44,9 @@ def get_update_log_message(self, _) -> Optional[str]:
             return None
         if self.status.get("result") == "success":
             return "LaunchDarkly import completed successfully"
-        if error_message := self.status.get("error_message"):
-            return f"LaunchDarkly import failed with error: {error_message}"
+        if error_messages := self.status.get("error_messages"):
+            if len(error_messages) > 0:
+                return f"LaunchDarkly import failed with errors: {str(error_messages)}"
```
khvn26: Let's make this a bit more human-readable:

```suggestion
                return f"LaunchDarkly import failed with errors:\n{'\n'.join('- ' + error_message for error_message in error_messages)}"
```

api/integrations/launch_darkly/services.py
```diff
@@ -368,17 +1063,35 @@ def process_import_request(
...
             project_id=import_request.project_id,
         )
-        tags_by_ld_tag = _create_tags_from_ld(
-            ld_tags=ld_tags,
+
+        # Create segments
+        segment_tags_by_ld_tag = _create_tags_from_ld(
+            ld_tags=ld_segment_tags,
+            project_id=import_request.project_id,
+        )
```
khvn26: I'd argue we don't need to create these before https://github.com/Flagsmith/flagsmith/issues/3241 is closed.

How about we leave the following code instead?

```suggestion
        segment_tags_by_ld_tag = {}  # TODO populate with LD tags when https://github.com/Flagsmith/flagsmith/issues/3241 is done 
```

api/integrations/launch_darkly/services.py
```diff
@@ -109,15 +120,513 @@ def _create_tags_from_ld(
...
+    match ld_operator:
+        case "in":
+            # TODO: How to escape the comma itself?
+            return list([",".join(values)])
+        case "endsWith":
+            return [".*" + re.escape(value) for value in values]
+        case "startsWith":
+            return [re.escape(value) + ".*" for value in values]
+        case _:
+            return [value for value in values]
```
khvn26: Looks like the semver value conversion mentioned [here](https://github.com/Flagsmith/flagsmith/pull/3233/files#r1447505113) has not been done yet?

api/integrations/launch_darkly/services.py
```diff
@@ -109,15 +120,513 @@ def _create_tags_from_ld(
...
+    :param segments_by_ld_key: a mapping from Launch Darkly segment key to Segment. Used to find right segment
+    from the "segmentMatch" operator
+    :return: a list of "FeatureSegment" operators created for each "segmentMatch" operator.
+    """
+
+    if any(clause["op"] != "segmentMatch" for clause in clauses):
+        _log_error(
+            import_request=import_request,
+            error_message=f"Nested segment match is not supported, skipping"
+            f" for {feature.name} in {environment.name}",
```
khvn26: I would like a slightly more rigid format for the error messages, e.g.

`Could not import {entity type} {entity name}: {reason}`

In this case, it would be

`Could not import segment clause {clause.attribute} {clause.operator}: nested segment match is not supported`

api/integrations/launch_darkly/services.py
```diff
@@ -109,15 +120,513 @@ def _create_tags_from_ld(
...
+    :param feature: the feature to target for identities.
+    :param environment: the environment to target for identities.
+    :param mv_feature_options_by_variation: a mapping from variation index to MultivariateFeatureOption if the
+    flag is multivariate.
+    """
+    if "targets" in ld_flag_config:
+        # Identifiers are grouped by their variation index. So each target has the same variation index.
+        for target in ld_flag_config["targets"]:
+            # Create a segment override for those identities. This is a work-around to support individual
+            # targeting in local evaluation mode.
```
khvn26, line 421: Please add a TODO to remove this when local evaluation is supported for identity overrides.

api/integrations/launch_darkly/services.py
```diff
@@ -126,11 +635,30 @@ def _create_boolean_feature_states(
...
+            mv_feature_options_by_variation={},
+        )
+        _import_rules(
+            import_request=import_request,
+            ld_flag_config=ld_flag_config,
+            feature=feature,
+            environment=environment,
+            segments_by_ld_key=segments_by_ld_key,
+            mv_feature_options_by_variation={},
+        )
```
khvn26, line 659: I don't like that we're introducing side effects here. We should consider one of the following:

- Factor out `_import_targets` and `_import_rules` calls so they're invoked directly from `process_import_request` rather than as a side effect. Make the factory functions return created entities, including the MV options, so they can be used by later code, or think of some common context object that can be used throughout the import process.
- Rename factory functions to `_create_boolean_feature_states_with_segments_identities` and so on.

khvn26 (changes requested):

api/integrations/launch_darkly/models.py
Dogacel: I found it easier to default it to an empty array rather than not required. As there is no logic that benefits from `None` over `[]`.

That's how we simplify the optionality of repeated fields in our scala codebase.

api/integrations/launch_darkly/client.py
Dogacel: Yeah, there is no `next` link.

I am not sure what is the intention there, I think this endpoint is just not refactored to have that `next` link. IMO removing the next link support is degradation,


 .. |  ..  | .. 
-- | -- | --
continuationToken | string <string>Limits results to contexts with sort values after the value specified. You can use this for pagination, however, we recommend using the next link we provide instead.
-- | --

This is from context search. So I think it is OK considering this is to support their legacy operation.

api/integrations/launch_darkly/services.py
Dogacel: Done, also updated the tests.

api/integrations/launch_darkly/services.py
Dogacel: Added a link that tracks SDK support for identity overrides.

api/integrations/launch_darkly/services.py
Dogacel: Hmm why you think this is a side-effect? Isn't targeting rules a part of the feature flag itself? 

I am not clear on how we are going to achieve this 🤔 As there is already lots of automated tests, can you push a commit that would do this refactoring? I feel like this  should not change how we create stuff.

api/integrations/launch_darkly/services.py
khvn26: The display bug is now fixed.

api/integrations/launch_darkly/services.py
khvn26: For later UI work, the error messages are exposed via API.

Additionally, the errors are reported via audit log.

api/integrations/launch_darkly/services.py
khvn26: > Isn't targeting rules a part of the feature flag itself?

In Flagsmith world they aren't, rather, it's the other way around — feature states are referenced by segment/identity overrides and are otherwise oblivious to them.

> As there is already lots of automated tests, can you push a commit that would do this refactoring?

I'm off for 4 work days so it won't happen soon. As the change is mostly for better code structure/readability, I'm fine with having a TODO + an issue for this. Leaving this for @matthewelwell to sort out.

api/integrations/launch_darkly/models.py
khvn26: Nice, agreed 👍

api/integrations/launch_darkly/client.py
khvn26: Sounds reasonable enough. I suggest we rename `use_offset_pagination` to `use_legacy_offset_pagination`, add a comment with an explanation, and resolve this.

api/integrations/launch_darkly/services.py
Dogacel: This should be resolved.

api/integrations/launch_darkly/client.py
Dogacel: That's just how Launch Darkly's API defines it. Added a comment that links to the API.

api/integrations/launch_darkly/services.py
Dogacel: Created an issue here: https://github.com/Flagsmith/flagsmith/issues/3383

Also added TODOs 👍

Dogacel:
I am hoping the failing E2E will be handled by #3399

khvn26 (approved):

matthewelwell:
Thanks again for this @Dogacel !

PR #3235, opened 2024-01-03
feat: Create split testing for multivariate

zachaysan:

## Changes

**Edit: These changes are out of date after multiple rounds of review and now with ongoing work to pull the split testing code into it's own repository.**

Overall, the changes introduced here support adding split testing to Flagsmith for the customers of ours that use the `settings.USE_POSTGRES_FOR_ANALYTICS` option, so this does not support our current SaaS method. The basic idea is that each of our multivariate options are linked into a `SplitTest` model and serve as the source of identification as to which split test is represented. In detail changes include:

1. A modification to `FeatureEvaluationRaw` to accept a new identifier so we can tell when a user has encountered a feature. This has been implemented as a query param since the body of the request just has feature names. Also, `FeatureEvaluationRaw` needed faster lookups so an index on `feature_name` was added concurrently as well.
2. An introduction of the `chi2_contingency` summary statistic that follows the G-Test version of split testing which was verified as current on Wikipedia as well as elsewhere. This comes with all of `scipy` which may be a drawback on some systems, but if there is significant issues with it we can always extract the one function we need out into its own helper.
3. The creation of a new tracking metric called `ConversionEvent` which can be called for events like new user signup, just as a normal analytics app would. This is used to track the actual split test results.
4. After a recurring task has processed the latest `SplitTest` results they are available via a `SplitTestViewSet` which serializes the related attributes of `feature` and `multivariate_feature_option` which reused serializers from existing areas of the codebase.

Around seven or eight new tests, mostly on the new functionality. I wasn't quite happy with the split test task test since it is not deterministic and there didn't seem to be a clean way to force some identities to some multivariate options and other ones to others. Also, it would have been nice to test it with more objects, but with the size of the objects that would have been necessary, the statistics part was tested with raw numbers instead.


api/app_analytics/models.py
```diff
@@ -73,11 +73,14 @@ def check_overlapping_buckets(self):
...
 
 class FeatureEvaluationRaw(models.Model):
-    feature_name = models.CharField(max_length=2000)
+    feature_name = models.CharField(db_index=True, max_length=2000)
     environment_id = models.PositiveIntegerField()
     evaluation_count = models.IntegerField(default=0)
     created_at = models.DateTimeField(auto_now_add=True)
 
+    # Identity identifier stored for tracking multivariate split testing.
+    identifier = models.CharField(max_length=2000, null=True, default=None)
```
matthewelwell: I think I'd prefer `identity_identifier` here.

api/app_analytics/split_testing/apps.py
```diff
@@ -0,0 +1,6 @@
+from django.apps import AppConfig
+
+
+class AppAnalyticsConfig(AppConfig):
+    name = "app_analytics.split_testing"
+    label = "app_analytics_split_testing"
```
matthewelwell: I think, unless there's a reason to change this, the `label` can be left as default (which I think will be `split_testing`).

api/app_analytics/split_testing/helpers.py
```diff
@@ -0,0 +1,36 @@
+import numpy as np
+from scipy.stats import chi2_contingency
```
matthewelwell: Since numpy and scipy (if I remember correctly) are fairly large dependencies we should make sure they're added as optional dependencies in poetry. How we control when they're installed we can of course discuss but just calling this out now.

api/app_analytics/split_testing/helpers.py
```diff
@@ -0,0 +1,36 @@
...
+def gather_split_test_metrics(
+    evaluation_counts: dict[int, int], conversion_counts: dict[int, int]
+) -> tuple[float, float]:
+    _evaluation_counts = []
+    _conversion_counts = []
+    for mv_feature_option_id, evaluation_count in evaluation_counts.items():
+        _evaluation_counts.append(evaluation_count)
+        _conversion_counts.append(conversion_counts[mv_feature_option_id])
+    input_data = np.array([_conversion_counts, _evaluation_counts])
+    return analyse_split_test(input_data)
```
matthewelwell: I'd love to see some docstrings on these functions. The code looks fairly sound, but the actual logic is going straight over my head.

api/app_analytics/split_testing/models.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class SplitTest(models.Model):
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["environment", "feature", "multivariate_feature_option"],
+                name="unique_environment_feature_mvfo",
+            )
+        ]
```
matthewelwell: Is there a benefit to doing this over `unique_together = ("environment", "feature", "multivariate_feature_option")` ?

api/app_analytics/split_testing/models.py
```diff
@@ -0,0 +1,55 @@
...
+    # number of unique identifiers for a single feature /
+    # environment combination. Multiple occurences ignored.
+    evaluation_count = models.PositiveIntegerField()
+    # from the ConversionEvent model for matching identifiers.
+    conversion_count = models.PositiveIntegerField()
+
+    # Split test metrics, where the pvalue is the most useful.
+    # See the analyse_split_test helpers function for more details.
+    pvalue = models.FloatField(null=False)
+    statistic = models.FloatField(null=False)
```
matthewelwell: This is perhaps obvious to most, but `statistic` seems like a fairly generic term here.

api/app_analytics/split_testing/serializers.py
matthewelwell, line 1: Based on the permissions class above, I would expect to see a query parameter serializer here too, to validate that the environment id is provided.

api/app_analytics/split_testing/serializers.py
```diff
@@ -0,0 +1,40 @@
...
+    def save(self, *args, **kwargs) -> ConversionEvent:
+        environment = self.context["request"].environment
+        identity = Identity.objects.get(
+            environment=environment,
+            identifier=self.validated_data["identity_identifier"],
+        )
+        return ConversionEvent.objects.create(
+            environment=environment,
+            identity=identity,
+        )
```
matthewelwell: Should we defer this behaviour to the task processor?

api/app_analytics/split_testing/tasks.py
```diff
@@ -0,0 +1,118 @@
...
+            environment_id__in=environment_ids,
+            identifier__isnull=False,
+        ).values_list("environment_id", "identifier")
+
+        # Eliminate duplicate identifiers
+        qs_values_list = qs_values_list.distinct()
+
+        environment_identifiers = defaultdict(list)
+        for env_identifier in qs_values_list:
+            environment_id, identifier = env_identifier
```
matthewelwell:
```suggestion
        for environment_id, identifier in qs_values_list:
```

api/app_analytics/split_testing/tasks.py
```diff
@@ -0,0 +1,118 @@
...
+            environment_identifiers[environment_id].append(identifier)
+
+        for environment_id, identifiers in environment_identifiers.items():
+            _save_environment_split_test(
```
matthewelwell: Might confuse things but I wonder if we should consider fanning out to additional (non-recurring) tasks here? I have a slight concern that the health check process for the task processor relies on each worker thread publishing it's ability to process tasks in a given time window. If we let a single worker do all of the work here, we may exceed the default grace period given to each worker. Perhaps this is the lesser of 2 evils but one to consider I think.

api/app_analytics/split_testing/tasks.py
```diff
@@ -0,0 +1,118 @@
...
+        conversion_counts[mv_option.id] = 0
+
+    conversion_events = ConversionEvent.objects.filter(
+        environment=environment,
+        identity__in=identities,
+    )
+
+    conversion_identities = set(
+        Identity.objects.filter(conversion_events__in=conversion_events)
+    )
```
matthewelwell: Maybe this is better performance-wise, but I think we can get this information without going back tot he database, right? Something like: 

```python
conversion_identities = {ce.identity for ce in conversion_events}
```

api/app_analytics/split_testing/tasks.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def _save_environment_split_test(
+    feature: Feature, environment_id: int, identifiers: list[str]
+) -> None:
+    environment = Environment.objects.get(id=environment_id)
+    identities = Identity.objects.filter(
+        environment_id=environment_id,
+        identifier__in=identifiers,
+    )
```
matthewelwell: I think we should select_related the environment here for the calls to `get_hash_key` later on.

api/app_analytics/split_testing/tasks.py
```diff
@@ -0,0 +1,118 @@
...
+        for environment_id, identifiers in environment_identifiers.items():
+            _save_environment_split_test(
+                feature=feature,
+                environment_id=environment_id,
+                identifiers=identifiers,
+            )
+
+
+def _save_environment_split_test(
+    feature: Feature, environment_id: int, identifiers: list[str]
```
matthewelwell: based on the below logic, I think we could be more explicit with the `identifiers` argument here. Perhaps `evaluated_identifiers` or even `identifiers_with_at_least_one_evaluation`. Something to differentiate between converted identities and those which just evaluate the feature.

api/app_analytics/split_testing/tasks.py
```diff
@@ -0,0 +1,118 @@
...
+    pvalue, statistic = gather_split_test_metrics(
+        evaluation_counts,
+        conversion_counts,
+    )
+
+    # Delete all the former split tests to avoid stale MV options.
+    SplitTest.objects.filter(
+        feature=feature,
+        environment=environment,
+    ).delete()
```
matthewelwell: Could we just do an `update_or_create` instead?

matthewelwell: I guess maybe that would be less optimal in terms of database interactions since we're using bulk_create below. I am always a little worried though when I see a delete like this followed by a create. Could / should we do some explicit transaction management here?

api/app_analytics/split_testing/tasks.py
```diff
@@ -0,0 +1,118 @@
...
+        conversion_count = conversion_counts[mv_feature_option_id]
+        split_tests.append(
+            SplitTest(
+                environment=environment,
+                feature=feature,
+                multivariate_feature_option_id=mv_feature_option_id,
+                evaluation_count=evaluation_count,
+                conversion_count=conversion_count,
+                pvalue=pvalue,
+                statistic=statistic,
```
matthewelwell: So, just to confirm here, the pvalue and statistic are shared between all of the `SplitTest` entities for the same feature / environment.  I think that makes sense and, as mentioned previously, removes the need for the parent model class. 

It does make me think, however, that the model name could probably be improved.

api/app_analytics/split_testing/views.py
```diff
@@ -0,0 +1,39 @@
...
+from .models import SplitTest
+from .permissions import SplitTestPermissions
+from .serializers import ConversionEventSerializer, SplitTestSerializer
+
+
+class ConversionEventViewSet(viewsets.ModelViewSet):
+    permission_classes = (EnvironmentKeyPermissions,)
+    authentication_classes = (EnvironmentKeyAuthentication,)
+    serializer_class = ConversionEventSerializer
+    http_method_names = ["post"]
```
matthewelwell: Unless there's a specific reason for doing it this way, I'd rather use the `GenericViewset` / mixins approach. So the class definition for this viewset for example becomes: 

```python
from rest_framework import mixins

...

class ConversionEventViewSet(viewsets.GenericViewSet, mixins.CreateModelMixin):
```

api/app_analytics/split_testing/views.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class SplitTestViewSet(viewsets.ModelViewSet):
+    permission_classes = (SplitTestPermissions,)
```
matthewelwell: This should probably still also include `IsAuthenticated`

api/app_analytics/split_testing/views.py
```diff
@@ -0,0 +1,39 @@
...
+    serializer_class = SplitTestSerializer
+    http_method_names = ["get"]
+
+    def get_queryset(self) -> QuerySet[SplitTest]:
+        environment_id = self.request.query_params.get("environment_id")
+        if not environment_id:
+            raise Http404(
+                "Unable to find split test without environment_id query param set"
+            )
```
matthewelwell: As previously mentioned I think we should use a serializer to validate the query parameters here, we can then also use it in a `swagger_auto_schema` decorator so the documentation is correct.

api/app_analytics/tasks.py
```diff
@@ -62,14 +62,19 @@ def clean_up_old_analytics_data():
 
 
 @register_task_handler()
-def track_feature_evaluation(environment_id, feature_evaluations):
+def track_feature_evaluation(
+    environment_id: int,
+    feature_evaluations: dict[str, int],
+    identifier: Optional[str] = None,
```
matthewelwell: I _think_ the more recent approach for defining optional types is `identifier: str | None = None` ?

api/app_analytics/views.py
```diff
@@ -73,8 +73,20 @@ def post(self, request, *args, **kwargs):
...
             )
+
+        # Since the post body data has bare names of features with counts,
+        # if we put the identifier in the post body it could collide with
+        # a feature name. It's more straightforward to set it as a query parm.
+        identifier = request.query_params.get("identifier")
+        if identifier:
+            assert isinstance(identifier, str)
+        else:
+            identifier = None
```
matthewelwell: Let's add a TODO to fix this in v2 as this is ugly. We can also an item to the issue here: https://github.com/Flagsmith/flagsmith/issues/1667

api/pyproject.toml
```diff
@@ -103,6 +103,7 @@ pydantic = "~1.10.9"
 pyngo = "~1.6.0"
 flagsmith = "^3.4.0"
 python-gnupg = "^0.5.1"
+scipy = "^1.11.4"
```
matthewelwell: as above, let's add this as an optional dependency.

api/tests/unit/app_analytics/split_testing/test_unit_split_testing_helpers.py
```diff
@@ -0,0 +1,31 @@
+import numpy as np
+from app_analytics.split_testing.helpers import analyse_split_test
+
+
+def test_analyse_split_tests_with_clear_results() -> None:
+    # Given
+    # Three options, with the last as clear winner as 400 / 500.
+    conversion_counts = [100, 200, 400]
+    evaluation_counts = [500, 500, 500]
+    input_data = np.array([conversion_counts, evaluation_counts])
```
matthewelwell: I might be missing something but it feels to me as though we could abstract any interactions with numpy from the user of these helper functions and just pass in the conversion counts and evaluation counts directly.

api/tests/unit/app_analytics/split_testing/test_unit_split_testing_tasks.py
```diff
@@ -0,0 +1,139 @@
...
+        default_percentage_allocation=50,
+        type=STRING,
+        string_value="mv_feature_option3",
+    )
+    MultivariateFeatureOption.objects.create(
+        feature=feature2,
+        default_percentage_allocation=50,
+        type=STRING,
+        string_value="mv_feature_option4",
+    )
```
matthewelwell: It's odd to me that we're not using the control value at all here. Both features have multivariate feature options defined that add up to 100%. This doesn't really seem like a realistic use case that we're testing.

api/tests/unit/app_analytics/split_testing/test_unit_split_testing_tasks.py
```diff
@@ -0,0 +1,139 @@
...
+    _update_split_tests()
+
+    # Then
+    split_tests = SplitTest.objects.all()
+    assert len(split_tests) == 4
+    assert sum([st.evaluation_count for st in split_tests]) == 5
+    assert sum([st.conversion_count for st in split_tests]) == 2
+    for st in split_tests:
+        assert st.pvalue == 1.0
+        assert st.statistic == 0.0
```
matthewelwell: These seem like very specific numbers, is there a reason we're getting these? Probably because the test has no statistical significance but I would have expected them to be (infinitesimally) close to 1 and 0, but probably not exactly those values.

api/tests/unit/app_analytics/split_testing/test_unit_split_testing_tasks.py
matthewelwell, line 1: One thing that we might be missing here, based on my nervousness around the bulk delete that happens in the task code, is a test that checks multiple tests across multiple environments. That might be a lot of effort for what is quite simple logic though so I'm happy to be rebuffed here!

api/tests/unit/app_analytics/split_testing/test_unit_split_testing_views.py
```diff
@@ -0,0 +1,171 @@
...
+
+def test_split_test_list(
+    staff_client: APIClient,
+    with_environment_permissions: Callable[[list[str], int], None],
+    environment: Environment,
+    project: Project,
+) -> None:
+    # Given
+    feature = Feature.objects.create(
+        name="single feature for test",
```
matthewelwell: nit: although it's not enforced in the ORM, we should create a more realistic feature here by not including spaces (since the FE enforces this).

api/tests/unit/app_analytics/split_testing/test_unit_split_testing_views.py
```diff
@@ -0,0 +1,171 @@
...
+        pvalue=0.2,
+        statistic=15,
+    )
+
+    with_environment_permissions([VIEW_ENVIRONMENT])
+    url = reverse("api-v1:split-tests-list")
+    url = f"{url}?environment_id={environment.id}"
+
+    # When
+    response = staff_client.get(url)
```
matthewelwell: I'd love to see a `django_assert_num_queries` here to ensure we're not introducing N+1 issues.

matthewelwell (commented):
@zachaysan on the whole, this is really impressive. I have gone through with quite a fine-toothed comb so there are a fair amount of comments but most of them are my lack of knowledge in the world of statistics, or a little bit nit-picky :)

api/app_analytics/split_testing/models.py
gagantrivedi: Yeah `unique-together` is practically deprecated

```


Use UniqueConstraint with the constraints option instead.

UniqueConstraint provides more functionality than unique_together. unique_together may be deprecated in the future.

```
source: https://docs.djangoproject.com/en/dev/ref/models/options/#unique-together

gagantrivedi: nit: can we move fields definition up and meta down?

api/app_analytics/split_testing/permissions.py
```diff
@@ -0,0 +1,23 @@
...
+    def has_permission(self, request: Request, view: ModelViewSet) -> bool:
+        if not super().has_permission(request, view):
+            return False
+
+        environment_id = request.query_params.get("environment_id")
+
+        if not environment_id:
+            return False
+
+        environment = Environment.objects.get(id=environment_id)
```
gagantrivedi: This will crash if environment ID is not valid, right?

api/app_analytics/split_testing/apps.py
zachaysan: Ah, of course. I was going off of the feature export apps where I needed a `label` there because of the collision with the other import export module.

api/app_analytics/models.py
zachaysan: Oh yeah, 100% good catch. I don't know why I didn't think of it.

zachaysan: FYI: I'm updating a lot of the code to match this behaviour since there are a lot of places that refer to this only as `identifier` when `identity_identifier` would be more sound. It's taking some time but it's the right call.

api/app_analytics/split_testing/models.py
zachaysan: Yeah I can, although I preferred seeing the Meta class near the top. It seems strange to me buried further down in the code, but if that's what we do here I'm fine to move it.

api/app_analytics/split_testing/tasks.py
zachaysan: 👍🏼

api/app_analytics/split_testing/tasks.py
zachaysan: Yep this is a good call.

api/app_analytics/split_testing/tasks.py
zachaysan: I think `evaluated_identifiers` matches the existing language better. Will update.

api/app_analytics/split_testing/views.py
zachaysan: As discussed briefly during standup, `IsAuthenticated` is the parent class of the `SplitTestPermissions`

api/app_analytics/split_testing/views.py
zachaysan: Ok cool, I'll default to that pattern for this project when using class based ViewSets.

api/app_analytics/split_testing/serializers.py
zachaysan: I think it belongs in the serializer, no? This is how we did things in the feature import export stuff.

api/app_analytics/split_testing/serializers.py
zachaysan: Ok this has been added.

api/app_analytics/split_testing/views.py
zachaysan: Sounds good

api/app_analytics/split_testing/tasks.py
zachaysan: I decided to shift this into a `bulk_update` if there are records and a `bulk_create` if there are not in order to get rid of the delete.

api/app_analytics/split_testing/tasks.py
zachaysan: I decided that we don't really need the `statistic` measure after all and removed it from the codebase. It's useful for debugging, but it will end up with too much cognitive overhead for the frontend. 

As for the model name, I'm open to feedback, but `SplitTest` seemed the most natural one to land on.

api/app_analytics/tasks.py
zachaysan: Yeah you're right. This will take me so long to internalize. Feel free to keep correcting me on it.

api/app_analytics/views.py
zachaysan: Ok sounds good. I've added the TODO locally and I added a comment to #1667 about it as well.

api/tests/unit/app_analytics/split_testing/test_unit_split_testing_helpers.py
zachaysan: Sure that's easy.

api/tests/unit/app_analytics/split_testing/test_unit_split_testing_tasks.py
zachaysan: I'm not sure if I 100% follow what the connection is, but I have re-written the bulk delete code to use updates now. As for testing across multiple environments, I don't think we really need that much of a test build up for any edge cases there. They're already pretty large tests as it is.

api/tests/unit/app_analytics/split_testing/test_unit_split_testing_tasks.py
zachaysan: In a lot of statistics stuff you have cut offs for the lower numbers. It's not a smooth curve until a threshold has been reached.

api/tests/unit/app_analytics/split_testing/test_unit_split_testing_views.py
zachaysan: Right! Thank you!

api/tests/unit/app_analytics/split_testing/test_unit_split_testing_views.py
zachaysan: Sure that's a good idea.

api/tests/unit/app_analytics/split_testing/test_unit_split_testing_tasks.py
zachaysan: From the call with @matthewelwell there is a fallback that I missed in the last line of `get_multivariate_feature_state_value` where the default value is selected if there isn't a matching multivariate feature state value.

The fix for this will necessitate adding in a null value to the multivariate feature option and falling back to the feature value for those records, which should be one per collection of feature, environment, and mv feature option.

api/app_analytics/split_testing/helpers.py
zachaysan: Updating this comment from our discussion: 

We're actually going to move the split testing code to its own repository like we presently do with LDAP which will isolate these dependencies.

api/app_analytics/split_testing/helpers.py
zachaysan: Ok I've written some docstrings for these.

api/app_analytics/split_testing/models.py
zachaysan: Ok I've moved this to the bottom.

api/app_analytics/split_testing/models.py
zachaysan: I decided that while `statistic` can be useful in verifying results while developing, it doesn't have much use in reporting successfully passing split test, so I've removed it.

api/app_analytics/split_testing/permissions.py
zachaysan: Nice catch, I've altered it to return false if there is no matching environment id.

api/app_analytics/split_testing/tasks.py
zachaysan: Good suggestion. The worker has been altered to fanning out each feature (with all environments) to its own task.

api/app_analytics/split_testing/helpers.py
```diff
@@ -0,0 +1,53 @@
...
+    log-likelihood argument set in order to accomplish what is
+    more commonly known as the G-Test with the p-value returned.
+    The p-value is a measure indicating the strength of evidence
+    against a null hypothesis with smaller values indicating more
+    statistical strength in the relationship being non-random.
+    """
+    # Reform input data to scipy's format.
+    observed_matrix = np.array([conversion_counts, evaluation_counts])
+
+    # Replace zero values in order for the chi-squared results can
```
matthewelwell:
```suggestion
    # Replace zero values in order for the chi-squared results to be
```

api/app_analytics/split_testing/helpers.py
```diff
@@ -0,0 +1,53 @@
...
+    # as large as 5% is acceptable for some tests.
+    return results.pvalue
+
+
+def gather_split_test_metrics(
+    evaluation_counts: dict[int, int], conversion_counts: dict[int, int]
+) -> float:
+    """
+    Take in evalaution counts (aka, views of the individual features) and
+    matches them up against the conversion counts in order to run the
```
matthewelwell:
```suggestion
    match them up against the conversion counts in order to run the
```

api/app_analytics/split_testing/serializers.py
matthewelwell: Hmm... well, in theory yes - that makes sense. My thought process here is that we want to optimise this endpoint for performance and I don't _think_ (correct me if I'm wrong here) the client cares about the response to this request. So, a single trip to the database to create a task would likely be better than multiple round trips as per the current implementation. It's not a major issue, but something to think about perhaps.

api/app_analytics/split_testing/views.py
```diff
@@ -0,0 +1,96 @@
...
+from .serializers import (
+    ConversionEventSerializer,
+    ConversionEventTypeQuerySerializer,
+    ConversionEventTypeSerializer,
+    SplitTestQuerySerializer,
+    SplitTestSerializer,
+)
+
+
+class ConversionEventViewSet(viewsets.GenericViewSet, mixins.CreateModelMixin):
```
matthewelwell: Sorry, I'd rather this was a `CreateAPIView` - I'm not too sure why really, it's more of a feeling. I tend to think the `ViewSet` classes should be reserved for dashboard crud management, whereas this is used by the SDK.

api/app_analytics/split_testing/serializers.py
zachaysan: @matthewelwell and I talked about this offline and decided that we'll respond with a 202 to keep the options open in the future to move this to a task if we see fit.

api/app_analytics/split_testing/views.py
zachaysan: No problem.

api/app_analytics/split_testing/views.py
```diff
@@ -0,0 +1,107 @@
...
+class CreateConversionEventView(CreateAPIView):
+    permission_classes = (EnvironmentKeyPermissions,)
+    authentication_classes = (EnvironmentKeyAuthentication,)
+    serializer_class = ConversionEventSerializer
+
+    def post(self, *args, **kwargs) -> Response:
+        response = super().post(*args, **kwargs)
+
+        # Return an empty client to keep options open to process
+        # this in a task one day.
```
matthewelwell:
```suggestion
        # Return an empty response to the client to keep options open to process
        # this in a task one day.
```

api/app_analytics/split_testing/views.py
```diff
@@ -0,0 +1,107 @@
...
+        if response.status_code == 201:
+            return Response(status=status.HTTP_202_ACCEPTED)
```
matthewelwell: Hmm... this is going to mean the documentation isn't correct. Probably the easiest way to fix this would be to add the `swagger_auto_schema` decorator to the `post` method here.

zachaysan: Nice catch! I've added the swagger response but the format looks weird for 202s

```python
@swagger_auto_schema(request_body=no_body, responses={202: ""})
```

matthewelwell (approved):
Approving this PR since I've reviewed all the code. We will not merge this until the bulk of the split_testing code is moved into its own repository however.

api/pyproject.toml
zachaysan: Ok `scipy` has been removed

api/poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.7.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.5.1 and should not be changed by hand.
```
matthewelwell: Hmm... I think the change to this file can probably be reverted since there are no changes to pyproject.toml?

api/tests/unit/app_analytics/test_unit_app_analytics_views.py
```diff
@@ -131,3 +145,73 @@ def test_get_total_usage_count_for_non_admin_user_returns_403(
...
+def test_set_sdk_analytics_flags_with_enabled_when_evaluated(
+    api_client: APIClient,
+    environment: Environment,
+    feature: Feature,
+    settings: SettingsWrapper,
+) -> None:
+    # Given
+    settings.USE_POSTGRES_FOR_ANALYTICS = True
+    url = reverse("api-v1:analytics-flags")
+    url += "?enabled_when_evaluated=true"
```
matthewelwell: So I guess this means that _all_ features sent in the request will be marked as 'enabled_when_evaluated` ? This might make the integration pretty tricky in all SDKs since they all keep a local cache (with counts) and then flush that periodically. It's possibly ok for now, but it's going to require some heavy changes in any of the SDKs that we want to implement this in. 

It's making me think that we could just create a brand new endpoint...

zachaysan: Notes from call with Matt:

1. Make a new endpoint under a new V2 path.
2. Fix the body of the post to not just use feature names directly (see below).

image

api/poetry.lock
zachaysan: I copied the main branch in, committed, then merged with main since there were too many main merges between poetry updates to do a clean revert. Either way, should be fixed now.

api/tests/unit/app_analytics/test_unit_app_analytics_views.py
zachaysan: Ok this is now implemented.

api/app_analytics/serializers.py
```diff
@@ -16,3 +16,14 @@ class UsageDataQuerySerializer(serializers.Serializer):
 
 class UsageTotalCountSerializer(serializers.Serializer):
     count = serializers.IntegerField()
+
+
+class SDKAnalyticsFlagsSerializerDetail(serializers.Serializer):
+    feature_name = serializers.CharField()
+    identity_identifier = serializers.CharField()
```
matthewelwell: Maybe this is the default, but this should be `required=False` I think? Flags can be evaluated without the context of an identifier.

api/app_analytics/views.py
```diff
@@ -32,6 +40,59 @@
...
+    serializer_class = SDKAnalyticsFlagsSerializer
+
+    def create(self, request: Request, *args, **kwargs) -> Response:
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        self.evaluations = serializer.validated_data["evaluations"]
+        if not self._is_data_valid():
+            # Mirror v1 implementation and return 200 to avoid
+            # breaking client integrations.
```
matthewelwell: Since we're creating a new endpoint here, we can probably change this if we want to?

api/app_analytics/views.py
```diff
@@ -32,6 +40,59 @@
...
+            track_feature_evaluation_v2.delay(
+                args=(
+                    request.environment.id,
+                    self.evaluations,
+                )
+            )
+        elif settings.INFLUXDB_TOKEN:
+            track_feature_evaluation_influxdb_v2(
+                request.environment.id, self.evaluations
+            )
```
matthewelwell: This should be a task as well.

api/tests/unit/app_analytics/test_unit_app_analytics_views.py
```diff
@@ -131,3 +146,88 @@ def test_get_total_usage_count_for_non_admin_user_returns_403(
...
+    # Given
+    settings.USE_POSTGRES_FOR_ANALYTICS = True
+    url = reverse("api-v2:analytics-flags")
+    api_client.credentials(HTTP_X_ENVIRONMENT_KEY=environment.api_key)
+    feature_request_count = 2
+    data = {
+        "evaluations": [
+            {
+                "feature_name": feature.name,
+                "identity_identifier": identity.identifier,
```
matthewelwell, line 173: As per the above, I'd like to see a test where the `identity_identifier` is omitted.

api/app_analytics/serializers.py
zachaysan: Nice catch! I've set it to `required=False, default=None` so the client can fail to include the key but it will still save as `None`.

api/app_analytics/views.py
zachaysan: Ok I've changed the name and sent a HTTP `400` status code.

api/tests/unit/app_analytics/test_unit_app_analytics_views.py
zachaysan: This is done.

api/app_analytics/views.py
zachaysan: I mirrored the existing function but since this one should be a task then both of them should be, right? So I've updated both versions to operate as tasks. Let me know if this is a problem.

zachaysan: I also wrote an integration test that was missing for the v1 version of this just to make sure the task is configured correctly there as well.

matthewelwell (approved):

PR #3239, opened 2024-01-04
chore: Bump default stuck limit for task processor tasks

zachaysan:

## Changes

Our SaaS application has bumped the task worker limit to 20 seconds, but our on prem customers still use the old default. It's probably a better idea to extend the same worker limit to them as well, since 3 seconds of processing time is a bit tight for some tasks.

I did not test this code because it's just a default value change.

### DIFF
--- a/api/task_processor/management/commands/runprocessor.py
+++ b/api/task_processor/management/commands/runprocessor.py
@@ -45,7 +45,7 @@ class Command(BaseCommand):
             "--graceperiodms",
             type=int,
             help="Number of millis before running task is considered 'stuck'.",
-            default=3000,
+            default=20000,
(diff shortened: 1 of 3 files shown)


matthewelwell (approved):
This is good, but we can probably also remove the override in the `infrastructure/aws/production/...task-processor....json` file.

zachaysan:
> This is good, but we can probably also remove the override in the `infrastructure/aws/production/...task-processor....json` file.

Nice catch. I've added it and while I was at it I grepped the codebase for the flag to see if there were any other changes and I found the documentation needed to be updated as well.

matthewelwell (approved):

PR #3245, opened 2024-01-05
feat: make segment condition value dynamic

matthewelwell:
## Changes

For self hosted customers, the limit on the length of each segment condition is very restrictive for little benefit. The restriction in our SaaS environment is primarily driven by the size limit that we have imposed on us by dynamodb (although hopefully this will be worked around in the not too distant future). 

This PR allows self hosted customers to define the length of that column via an environment variable. I have added documentation warning against it's use for already running instances of Flagsmith. 

Ran the following:

```
export DATABASE_URL=<url of an empty database>
export SEGMENT_CONDITION_VALUE_LIMIT=1999999
python manage.py migrate
python manage.py runserver
```

and validated the following: 

 1. The database column was created with a max size limit of 1999999B
 2. It is possible to create a segment condition with a size exceeding the default limit of 1000B
 3. It is not possible to create a segment condition with a size exceeding the configured limit of 1999999B
 4. Using a segment with a value that has a size close to the limit still works in a reasonable manner

Note that I also ran this to confirm that django doesn't think there are migrations to run (assuming the environment variable is not modified): 

```
❯ python manage.py makemigrations segments
System check identified some issues:

WARNINGS:
?: (axes.W002) You do not have 'axes.middleware.AxesMiddleware' in your settings.MIDDLEWARE.
No changes detected in app 'segments'
```


docs/docs/deployment/hosting/locally-api.md
```diff
@@ -246,6 +246,10 @@ the below variables will be ignored.
   ignored.
 - `DISABLE_FLAGSMITH_UI`: Disable the Flagsmith UI which can be rendered by the API containers in a single container
   environment. Use `True` to disable, defaults to `False`.
+- `SEGMENT_CONDITION_VALUE_LIMIT`: Configure the size of the segment condition value in bytes. Default is 1000.
```
zachaysan, line 249: Perhaps we should add a warning that going above the default will make the data inoperable with the SaaS offering since we won't be able to import their data? This is kinda locking them in to hosting themselves, no?

zachaysan (approved):
One small question / suggestion but overall the approach makes sense.

docs/docs/deployment/hosting/locally-api.md
matthewelwell: Great catch - yes I'll do that!

zachaysan (approved):

PR #3250, opened 2024-01-08
docs: Fix missing js client functions

MatheusLasserre:

## Changes

Added to [Available Functions](https://docs.flagsmith.com/clients/javascript#available-functions) the following functions that were missing:
`setState(state: IState)=>void`;
`getAllFlags()=>IFlags`

I've opened the MarkDown Preview inside VSCode.

## Other

I also found that some available properties are also not documented, but since I couldn't find a place for them, I've chosen to mention them here and ask for suggestions about it. Should I make a PR for them? If so, in the same place as the Available functions, or somewhere else?

The properties are:

     // The stored identity of the user
    identity?: string;
   
     // Whether the flagsmith SDK is initialised
    initialised?: boolean;

     // Returns the current loading state
    loadingState?: LoadingState;


MatheusLasserre:
Marked as draft because i made a mistake in one of my pushes that marks the whole table as changed instead of just the 2 lines i added.

docs/docs/clients/client-side/javascript.md
```diff
@@ -256,12 +256,14 @@ All function and property types can be seen
...
 | <code>getState()=>IState</code>                                                                                                                     |                                                                               Retrieves the current state of flagsmith, useful in NextJS / isomorphic applications. `flagsmith.getState()`                                                                                |
+| <code>setState(state: IState)=>void</code>                                                                                                          |                                                                                                                    Set the current state of flagsmith.                                                                                                                    |
```
matthewelwell: Maybe this will be obvious to users of the JS client, but I think some kind of example here would be useful?

MatheusLasserre: Added. Ive added `flagsmith.setState({environmentID: 'iefBdPk5jLjp3WfFjiB4MS'})`, since the envID change is unique to the setState function after the init as far as i know. What do you think?

MatheusLasserre: This environmentID is the same used at the javascript-vanilla-example-v2 in the examples PR.

kyle-ssg: I think perhaps mention that this is useful for hydration via SSR and link to https://docs.flagsmith.com/clients/next-ssr#comparing-ssr-and-client-side-flagsmith-usage, this is probably a better example. We'd rarely setState with just the environment id, more with flags/identity state.

MatheusLasserre: Maybe this way?
	
"Set the current state of flagsmith, [useful in NextJS / isomorphic applications.](https://docs.flagsmith.com/clients/next-ssr#comparing-ssr-and-client-side-flagsmith-usage) e.g. `flagsmith.setState({identity: 'mary@mycompany.com'})`."

Or should i give a different or bigger setState example? I thought of putting just one so the one-line code would not be ugly. Instead of keeping it in my thoughts, let's check:

`flagsmith.setState({identity: 'mary@mycompany.com', flags: {font_size: 12}})`.

What do you think?

MatheusLasserre: @kyle-ssg Should i do one of these two examples? I would like to close this PR before i get another one.

matthewelwell: @MatheusLasserre let's just go with the first option. As long as the link to the relevant docs is there, I think that's good.

MatheusLasserre: Done!

matthewelwell (approved):

PR #3255, opened 2024-01-08
feat: Add endpoints for feature imports

zachaysan:

## Changes

Introduce two new endpoints to support ongoing feature imports. One is a list the other is a retrieve. These were initially going to be as part of the same viewset, but the router differences got in the way and so it was more straightforward to split the functionality apart, since the list view relies on the `project_pk` nested part of the router.

Four new tests, all of which are simple.


api/features/import_export/permissions.py
```diff
@@ -56,3 +57,28 @@ def has_permission(self, request: Request, view: ListAPIView) -> bool:
...
+
+class FeatureImportRetrievePermissions(IsAuthenticated):
+    def has_permission(self, request: Request, view: GenericViewSet) -> bool:
+        if not super().has_permission(request, view):
+            return False
+
+        feature_import = view.get_queryset().filter(pk=view.kwargs["pk"]).first()
+
+        if not feature_import:
+            return False
```
matthewelwell: I'd say it's perhaps better to use `get_object_or_404` here?

api/features/import_export/views.py
```diff
@@ -123,3 +129,28 @@ def get_queryset(self) -> QuerySet[FeatureExport]:
...
+
+        return FeatureImport.objects.filter(environment__in=environment_ids).order_by(
+            "-created_at"
+        )
+
+
+class FeatureImportRetrieve(viewsets.GenericViewSet, mixins.RetrieveModelMixin):
+    serializer_class = FeatureImportSerializer
+    permission_classes = [FeatureImportRetrievePermissions]
+    queryset = FeatureImport.objects.all()
```
matthewelwell: Since we're only including one action on each of these viewsets, perhaps we should just use standard `APIView` classes? `ListAPIView` / `RetrieveAPIView` I mean.

matthewelwell: Thinking about it, I'm not sure why we don't just combine them? I get that in one we have the object to do the permission check, but I don't really see how it's better than just filtering out in the queryset (which would essentially have the same effect as the `get_object_or_404` that I suggested above).

zachaysan: From our discussion on slack: Switch to just using the List and used the `ListAPIView` as mentioned.

api/features/import_export/permissions.py
zachaysan: Would have been better but no longer necessary as that permission is now deleted.

matthewelwell (approved):

PR #3257, opened 2024-01-09
feat: redesign organisation layout

azaiko-akveo:

## Changes

Apply new design to organization management functionality

Test application locally


dabeeeenster:
Can we fix the layout shifting as per the screen recording


https://github.com/Flagsmith/flagsmith/assets/173290/48cbcf38-01f8-4f8f-859a-273ee8c131d4

azaiko-akveo:
> Can we fix the layout shifting as per the screen recording

The shift occurs due to hiding/displaying the global application scrollbar (the effect also exists on the Account page).
The effect is not reproduced for most recent browser versions (verified for Chrome, Firefox, and Safari).

To resolve this issue, we need to update the global application layout styles. It will take significant time and considers testing all existing pages to handle unexpected issues.

kyle-ssg:
Upon logging in I see the following with no ability to change organisation. I suspect that we haven't considered permissions / viewing organisations as non admins ?

This needs to be tested with the following permissions

- Org admin
- Non org admin
- Non org admin but with Manage user groups / Manage users permissions

image

azaiko-akveo:
> Upon logging in I see the following with no ability to change organisation. I suspect that we haven't considered permissions / viewing organisations as non admins ?
> 
> This needs to be tested with the following permissions
> 
> * Org admin
> * Non org admin
> * Non org admin but with Manage user groups / Manage users permissions
> 
> image

Fixed. Now it is possible to choose an organization if user doesn't have permissions

image

kyle-ssg:
@azaiko-akveo I can switch org now but I can't do what I previously could?

After: 

image


Previously: 

Note: I can select a project to view.

image


Please check that you can perform the same actions that you can on staging.flagsmith.com.

Also as mentioned, please check this with Manage user groups / Manage users permissions

dabeeeenster:
@azaiko-akveo can we get an update?

azaiko-akveo:
> an we get an update?

still in the process of applying the logic of permissions

dabeeeenster:
> > an we get an update?
> 
> still in the process of applying the logic of permissions

Thanks for the update

azaiko-akveo:
@dabeeeenster , @kyle-ssg.
Permissions logic was updated. The PR is ready for review

dabeeeenster:
@azaiko-akveo If I'm on the "Keys" tab, and change organisation using the dropdown, the Keys area doesnt update. Similar issues with Webhooks.

azaiko-akveo:
> @azaiko-akveo If I'm on the "Keys" tab, and change organisation using the dropdown, the Keys area doesnt update. Similar issues with Webhooks.

fixed

frontend/web/components/ProjectManageWidget.tsx
```diff
@@ -0,0 +1,250 @@
...
+                    items={projects}
+                    renderRow={(
+                      { environments, id, name }: Project,
+                      i: number,
+                    ) => {
+                      return (
+                        <>
+                          {i === 0 && (
+                            <div className='col-md-6 col-xl-3'>
+                              <Permission
```
kyle-ssg: Since this a functional component now, can we replace the <Permission components with the equivelent useHasPermission

frontend/web/components/ProjectManageWidget.tsx
```diff
@@ -0,0 +1,250 @@
...
+    const { state } = router.route.location as { state: { create?: boolean } }
+    if (state && state.create) {
+      handleCreateProjectClick()
+    }
+  }, [handleCreateProjectClick, router.route.location])
+
+  return (
+    <div data-test='project-manage-widget' id='project-manage-widget'>
+      <OrganisationProvider>
+        {({ isLoading, projects }: OrganisationContext) => (
```
kyle-ssg: I think we can replace this with the useProjects hook

kyle-ssg (changes requested):
Just a couple comments

azaiko-akveo:
> Just a couple comments

@kyle-ssg , code was updated

kyle-ssg:
> @azaiko-akveo When I click on the flagsmith logo it goes to /projects which looks just like the organisation link except without tabs. Is there a need for /projects? It doesn't seem useful.
> 
> Unless I'm missing something I think we can:
> 
> - Remove the /projects route
> - Make the home button behave the same was as clicking organisation 
> 
> 

See following comment instead

dabeeeenster:
It should redirect to the new projects tab in the org screen?

kyle-ssg:
After discussion, let's just remove the logo from the nav so that we have the following @azaiko-akveo. If I'm not mistaken, we can remove the projects page unless it's being used?
image

azaiko-akveo:
@kyle-ssg , @dabeeeenster 
* there is a case when user has access only for reading projects. To avoid displaying single tab we display a separate page. The same behaviour currently exists for `/organisation-groups`
* navigation to the `/projects` (projects page) is used in many other components withing the app including login functionality

azaiko-akveo:
> It should redirect to the new projects tab in the org screen?

Yes, if we need to keep existing application behaviour

kyle-ssg (approved):

PR #3263, opened 2024-01-10
feat: add fields necessary for stale flags

matthewelwell:
## Changes

Adds logic necessary to support the stale flags functionality. The relevant changes are: 

 1. Adds fields to the list features endpoint which the FE can use to display / sort data about when a flag was last modified
 2. Updates the Tag model to support 3 new fields: 
    a. `is_system_tag` - field denoting that the tag is system created, rather than user created (necessary for the tag we plan to add to any flags determined to be stale)
    b. `is_permanent` - a replacement for the current FE functionality which uses regex to determine if a tag indicates that a feature is 'permanent' and should not be deleted. This is necessary as we will ignore any features tagged with a tag that has this attribute when evaluating if a feature is stale or not. 
    c. `type` - a field to add a new enumerated 'type' field to a tag. This allows us to deterministically find the tag which indicates a feature is stale.

Note that there is a related PR in the workflows repository [here](https://github.com/Flagsmith/flagsmith-workflows/pull/23) which adds the relevant tasks, etc. to supplement these changes.  

Added unit tests.


api/projects/tags/migrations/0005_add_tag_fields_for_stale_flags_logic.py
```diff
@@ -0,0 +1,75 @@
...
+    tag_class = apps.get_model("tags", "tag")
+
+    to_update = []
+
+    for tag in filter(
+        lambda t: get_sanitised_label(t.label) in PROTECTED_LABELS,
+        tag_class.objects.all(),
+    ):
+        tag.is_permanent = True
+        to_update.append(tag)
```
khvn26, line 35: Do we have an approximation of how long will it run for SaaS?

api/projects/tags/models.py
```diff
@@ -12,6 +16,22 @@ class Tag(AbstractBaseExportableModel):
...
+        default=False,
+        help_text="Indicates that a tag was created by the system, not the user.",
+    )
+    is_permanent = models.BooleanField(
+        default=False,
+        help_text="When applied to a feature, it means this feature should be excluded from stale flags logic.",
+    )
+    type = models.CharField(
+        default=None,
+        null=True,
```
khvn26: I'd prefer to have a default tag type.

api/projects/tags/models.py
```diff
@@ -12,6 +16,22 @@ class Tag(AbstractBaseExportableModel):
     description = models.CharField(max_length=512, blank=True, null=True)
     project = models.ForeignKey(Project, on_delete=models.CASCADE, related_name="tags")
 
+    is_system_tag = models.BooleanField(
+        default=False,
+        help_text="Indicates that a tag was created by the system, not the user.",
+    )
```
khvn26, line 23: Are we going to allow the users to mark flags as stale?

api/tests/unit/features/test_unit_features_views.py
```diff
@@ -2554,3 +2559,66 @@ def test_simple_feature_state_returns_only_latest_versions(
...
+        EnvironmentFeatureVersion.objects.create(
+            environment=environment_v2_versioning, feature=feature
+        )
+
+    # let's add a few more features to ensure we aren't adding N+1 issues
+    for i in range(2):
+        Feature.objects.create(name=f"feature_{i}", project=project)
+
+    # When
+    with django_assert_num_queries(14):  # TODO: reduce this number of queries!
```
khvn26, line 2608: Yeah, that crossed my mind as soon as I saw the annotate + prefetch related combo.

khvn26:
The code looks fine, but I felt I needed some questions cleared up for me before approving it.

api/projects/tags/migrations/0005_add_tag_fields_for_stale_flags_logic.py
matthewelwell: There are only ~3000 tags in our production database in total. We'll have to iterate over all of those and run the regex. Of those that match the regex, there are only 62. 

Query to get this data: 

```sql
select count(*)
from tags_tag t
where replace(replace(lower("label"), ' ', ''), '_', '') in ('protected', 'donotdelete', 'permanent');
```

Here's a simple script that attempts to replicate it: 

```python
import random
import re
import time
from dataclasses import dataclass

PROTECTED_LABELS = {"protected", "donotdelete", "permanent"}
LABEL_REGEX = re.compile(r"[ _]")


def get_sanitised_label(label: str) -> str:
    return LABEL_REGEX.sub("", label.lower())


@dataclass
class Dummy:
    """Dummy dataclass to at least attempt to simulate model serialization"""

    label: str


matching_labels = [
    random.choice(
        [
            "do not delete",
            "protected",
            "Protected",
            "do_not_delete",
            "permanent",
            "Permanent",
        ]
    )
    for _ in range(65)
]
non_matching_labels = ["not matching" for _ in range(3000)]
labels = [*matching_labels, *non_matching_labels]


if __name__ == "__main__":
    start = time.time()
    print(f"Starting at: {start}")

    to_update = []

    for label in filter(
        lambda l: get_sanitised_label(l) in PROTECTED_LABELS,
        labels,
    ):
        to_update.append(label)

    end = time.time()
    print(f"Finished at: {end}")

    print(f"Process took {end - start}")
```

And the output: 

```
Starting at: 1707500721.4080262
Finished at: 1707500721.408973
Process took 0.0009467601776123047
```

tl;dr

>Do we have an approximation of how long will it run for SaaS?

Probably 10s of ms.

api/projects/tags/models.py
matthewelwell: You mean like `TagType.NONE`?

api/projects/tags/models.py
matthewelwell: Yes, that's the plan for sure. Not sure if my code here caters for that though. I will check...

matthewelwell: Ah, well, yes - it will. That's why we're creating the tag on project create. Users will then be able to use that tag on their features.

khvn26: > Yes, that's the plan for sure.

No further questions in that case.

> Not sure if my code here caters for that 

Having seen the workflows PR I believe it does, but feel free to double-check.

api/projects/tags/models.py
khvn26: Yes, or something among the lines of `TagType.NORMAL`.

api/tests/unit/features/test_unit_features_views.py
matthewelwell: Just for a sanity check, I commented out all the annotations and re-ran the test which confirmed that I haven't actually added any queries here (since the `prefetch_related` call hasn't actually changed).

api/projects/tags/models.py
matthewelwell: I went with TagType.NONE as I feel that is the most relevant here. It also means that I've had to complicate the migration a little bit to create the nullable column, populate the data, then add the default and remove the null option.

khvn26 (approved):

PR #3267, opened 2024-01-10
fix: Adjust segment not rule

### LINKED ISSUE #382: Add "Not" predicate operator in the Segment rules
We should have an "Add AND NOT" as per below:

image

Would also mean updating https://github.com/Flagsmith/flagsmith-engine

kyle-ssg:

## Changes

Note: this feature is controlled via the **segment_operator** flag. 

- Adjusts the not condition to send the right payload to the API
- Adjusts UI to distinguish between AND and AND NOT conditions 

image

- Created the above segment in production, viewed a user and added a trait that matches one of the rules. The user as a result was no longer part of the segment.


novakzaballa (approved):
Screen Shot 2024-01-10 at 16 26 21


LGTM, just one thing, by removing the first condition and adding another one, the text is incorrect.

kyle-ssg:
Fixed

PR #3295, opened 2024-01-14
fix(webhooks): prevent raise on give up

matthewelwell:
## Changes

Prevent the code from raising the exception on give up off the backoff logic. Should resolve [this sentry issue](https://flagsmith.sentry.io/issues/4848505909). I've also changed the log level on backoff give up to warning which should resolve [this sentry issue](https://flagsmith.sentry.io/issues/4848505814).

This change required an update to backoff which subsequently involved upgrading both segment and rudderstack packages since the existing versions had both pinned older versions of backoff. 

I've added a unit test to test the change to the backoff functionality and manually tested (by running the API locally) the integrations with segment and rudderstack.


khvn26 (approved):

khvn26 (changes requested):
```
TypeError: expo() got an unexpected keyword argument 'raise_on_giveup'
```

matthewelwell:
> ```
> TypeError: expo() got an unexpected keyword argument 'raise_on_giveup'
> ```

Whoops, looks like it's only available since 2.0 which something in our deps graph is preventing. I'll fix this up. Sorry!

zachaysan (approved):
It took me a double take to understand how the test works, but it looks good.

api/integrations/rudderstack/rudderstack.py
```diff
@@ -16,7 +16,7 @@
 class RudderstackWrapper(AbstractBaseIdentityIntegrationWrapper):
     def __init__(self, config: RudderstackConfiguration):
         rudder_analytics.write_key = config.api_key
-        rudder_analytics.data_plane_url = config.base_url
+        rudder_analytics.dataPlaneUrl = config.base_url
```
matthewelwell, line 19: This was a change in v2 of the rudderstack library. It blows my mind that they've moved from snake case to camel in a python library 🤯 but it is what it is. I've tested it locally to verify this is correct.

matthewelwell:
> It took me a double take to understand how the test works, but it looks good.

This seems fair on my own review so I've tried to add a bit more verbiage in the test to help clarify.

khvn26 (approved):

PR #3297, opened 2024-01-15
chore: Convert users models TestCase to normal test function

zachaysan:

## Changes

I'm working through re-writing the `TestCase` classes into normal test functions. Rewrote the existing `TestCase` and replaced with fixtures instead of recreating everything. There are other PRs that are similar to this one.

I manually changed the tests one at a time to refactor them as I went then ran the full test suite afterwards. Also, I changed assertions as I went to the new fixtures which resulted in shorter tests.

### DIFF
--- a/api/tests/unit/users/test_unit_users_models.py
+++ b/api/tests/unit/users/test_unit_users_models.py
@@ -1,168 +1,154 @@
-from unittest import TestCase, mock
+from unittest import mock
 
 import pytest
 from django.db.utils import IntegrityError
 
-from environments.models import Environment
 from organisations.models import Organisation, OrganisationRole
 from organisations.permissions.models import UserOrganisationPermission
 from organisations.permissions.permissions import ORGANISATION_PERMISSIONS
-from projects.models import (
-    Project,
-    ProjectPermissionModel,
-    UserProjectPermission,
-)
+from projects.models import Project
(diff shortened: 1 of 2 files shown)


api/tests/unit/users/test_unit_users_models.py
```diff
@@ -1,168 +1,154 @@
...
+    assert projects.count() == 1
+    assert projects.first() == project
 
-        # Then
-        assert projects.count() == 1
 
-    def test_get_admin_organisations(self):
-        # Given
-        self.user.add_organisation(self.organisation, OrganisationRole.ADMIN)
+def test_get_admin_organisations(
```
gagantrivedi, line 60: Can we create another organization here and assign the user as a non-administrator to it, just to make sure it's not returning all the organizations

gagantrivedi (changes requested):

api/tests/unit/users/test_unit_users_models.py
zachaysan: Yes this is a good idea. I should have caught that with the existing test.

zachaysan: Ok this is fixed. I also found a typing issue in another file that was missed by a recent cleanse so I just added it into this PR.

gagantrivedi (approved):

PR #3298, opened 2024-01-15
feat: Add GitHub Integration

novakzaballa:

## Changes

- Add a new GitHub Setup Page
- Add a new GitHub integration in Integrations
- The new components are wrapped by the `github_integration` flag.


frontend/common/types/responses.ts
```diff
@@ -137,6 +146,211 @@ export type LaunchDarklyProjectImport = {
...
+  id: number
+  node_id: string
+  number: number
+  title: string
+  user: User
+  labels: any[]
+  state: string
+  locked: boolean
+  assignee: null | any
+  assignees: any[]
```
kyle-ssg: What's the actual type ?

frontend/common/types/responses.ts
```diff
@@ -137,6 +146,211 @@ export type LaunchDarklyProjectImport = {
...
+  milestone: null | any
+  comments: number
+  created_at: string
+  updated_at: string
+  closed_at: null | string
+  author_association: string
+  active_lock_reason: null | string
+  body: string
+  reactions: Reaction
```
kyle-ssg: Is this a type ?

docs/docs/integrations/project-management/github.md
```diff
@@ -0,0 +1,34 @@
...
+
+:::tip
+
+- The GitHub integration is currently only supported with our hosted Flagsmith SaaS service.
+
+:::
+
+## Integration Setup
+
+1. In GitHub, add the app from the [GitHub Marketplace](https://github.com/apps/flagsmith-github-integration).
```
kyle-ssg, line 30: This should be navigable from the admin ui, if I'm not mistaken connection to GH can be done without navigating the user away from the app too.

docs/docs/integrations/project-management/github.md
```diff
@@ -0,0 +1,34 @@
...
+2. Select your organisation.
+3. Select your repositories where you want install the app.
+4. You was redirected to Flagsmith app -> GitHub Setup.
```
kyle-ssg: "You will be redirected back to the Flagsmith app to finish the integration setup". 


How do you get to this step if the integration is already added? I think as I mentioned later on this should just be a permanent page

frontend/web/components/OrganisationSelect.js
```diff
@@ -8,6 +8,12 @@ const OrganisationSelect = class extends Component {
     this.state = {}
   }
 
+  componentDidMount() {
+    if (this.props.firstOrganisation) {
+      this.props.onChange(AccountStore.getOrganisation().id)
```
kyle-ssg: This will probably crash the app if the org hasn't loaded yet

frontend/web/components/ProjectFilter.tsx
```diff
@@ -30,7 +32,7 @@ const ProjectFilter: FC<ProjectFilterType> = ({
         (data || [])?.map((v) => ({ label: v.name, value: `${v.id}` })),
       )}
       onChange={(value: { value: string; label: string }) =>
-        onChange(value?.value || '')
+        onChange(projectComplete ? value : value?.value ? value.value : '')
```
kyle-ssg: Unclear what this is doing, what is the new prop and why is this adjusting the project filter

frontend/web/components/modals/CreateFlag.js
```diff
@@ -187,6 +198,50 @@ const CreateFlag = class extends Component {
     ) {
       this.getFeatureUsage()
     }
+
+    if (
```
kyle-ssg: Could we put all of this into a separate components (e.g. the github integration, github issue select(allow a prop to include prs)? CreateFlag is already very bloated. Also that way the hooks can be used to cut down the amount of code this is calling

frontend/web/components/pages/GitHubSetupPage.tsx
```diff
@@ -0,0 +1,245 @@
...
+  label: string
+}
+
+type repoType = {
+  value: string
+  label: string
+  name?: string
+}
+
+const GitHubSetupPage: FC<GitHubSetupPageType> = (props) => {
```
kyle-ssg: I don't think this should be a separate page, it should either be part of the integrations page or project settings. If you check out Vercel for example as a good flow for connecting GH.

Testing this out it has to work more like that anyway, since if you've already connected to GitHub there's no callback url, you just see the ability to configure.

frontend/web/components/modals/CreateFlag.js
```diff
@@ -187,6 +198,50 @@ const CreateFlag = class extends Component {
...
+              .catch((error) => {
+                console.log(error)
+              })
+              .then((pulls) => {
+                this.setState({ prExternalResources: pulls?.data })
+              })
+          }
+        })
+        .catch((error) => {
+          console.log(error)
```
kyle-ssg: all the above logs / catches arent necessary I dont' think

frontend/web/components/modals/CreateFlag.js
```diff
@@ -187,6 +198,50 @@ const CreateFlag = class extends Component {
...
     }
+
+    if (
+      this.props.projectFlag &&
+      Utils.getFlagsmithHasFeature('github_integration')
+    ) {
+      getExternalResource(getStore(), {
+        feature_id: this.props.projectFlag.id,
+      }).then((res) => {
+        this.setState({ externalResources: res.data.results })
```
kyle-ssg: We'll be able to remove state in tsx components with hooks

frontend/web/components/pages/GitHubSetupPage.tsx
```diff
@@ -0,0 +1,245 @@
...
+          const promises = []
+          for (const project of projects) {
+            const promise = createGithubRepository({
+              body: {
+                project: project.value,
+                repository_name: repositoryName,
+                repository_owner: repositoryOwner,
+              },
+              github_pk: data.id,
+              organisation_pk: organisation,
```
kyle-ssg: organisation_pk is just the org id right? Is it possible to just replace these instances with organisation_id, I think requests.ts normally use this

frontend/web/components/pages/GitHubSetupPage.tsx
kyle-ssg: I think even if we did have a separate page, this looks a bit unfinished? 

image

kyle-ssg: Managed to get here with an installation id but see a 500 pointing to the api branch url, I think that should work unless environment vars need adding ? 
image

frontend/web/components/pages/GitHubSetupPage.tsx
novakzaballa: Changed.

frontend/web/components/modals/CreateFlag.js
novakzaballa: OK, removed.

frontend/web/components/modals/CreateFlag.js
novakzaballa: Ok, I created some new components.

docs/docs/integrations/project-management/github.md
novakzaballa: Corrected.

frontend/web/components/modals/CreateFlag.js
```diff
@@ -601,9 +643,156 @@ const CreateFlag = class extends Component {
...
+                        _createExternalResourse()
+                      }}
+                    >
+                      Link
+                    </Button>
+                  )}
+                </Row>
+              </FormGroup>
+              <ExternalResourcesTable featureId={projectFlag.id} />
+              {/* <PanelSearch
```
kyle-ssg: This can be removed it looks like

frontend/web/components/pages/GitHubSetupPage.tsx
```diff
@@ -0,0 +1,286 @@
...
+
+  useEffect(() => {
+    const createRepositories = async () => {
+      try {
+        if (
+          isSuccessCreateGithubIntegration &&
+          localStorage?.githubIntegrationSetupFromFlagsmith
+        ) {
+          const dataToSend = { 'installationId': installationId }
+          window.opener.postMessage(dataToSend, '*')
```
kyle-ssg, line 99: Does this mean you got a modal way to do this ?

frontend/web/components/pages/GitHubSetupPage.tsx
```diff
@@ -0,0 +1,286 @@
...
+        }
+        if (
+          data &&
+          isSuccessCreateGithubIntegration &&
+          !localStorage?.githubIntegrationSetupFromFlagsmith
+        ) {
+          const promises = []
+          for (const project of projects) {
```
kyle-ssg: I think this would be neater 
```
await Promise.all(projects.map((project)=>{

})

```

frontend/web/components/pages/GitHubSetupPage.tsx
```diff
@@ -0,0 +1,286 @@
...
+        }
+      } catch (error) {
+        console.error('Error creating:', error)
+      }
+    }
+
+    createRepositories()
+  }, [
+    data,
+    isSuccessCreateGithubIntegration,
```
kyle-ssg: This pattern leads to confusing, hard to read code. Instead, can we do 

```
createGithubIntegration({}).then((res)=>{
  if(!res.error){
   do stuff
  }
})
```

frontend/web/components/pages/GitHubSetupPage.tsx
```diff
@@ -0,0 +1,286 @@
...
+    if (reposLoaded && repos.repositories) {
+      setRepositoryOwner(repos?.repositories[0].owner.login)
+      setRepositories(repos)
+    }
+  }, [repos, reposLoaded])
+
+  useEffect(() => {
+    if (
+      isSuccessCreatedGithubRepository &&
+      !localStorage?.githubIntegrationSetupFromFlagsmith
```
kyle-ssg: Can we make this a variable rather than having this everywhere, it will also help if it was typed as it's not clear what this is

frontend/web/components/pages/GitHubSetupPage.tsx
```diff
@@ -0,0 +1,286 @@
...
+                },
+                organisation_id: organisation,
+              })
+            }}
+          >
+            Save Configuration
+          </Button>
+        </>
+      ) : (
+        <h1>Installation Completed</h1>
```
kyle-ssg, line 257: Is this necessary ? When does this show, immediately as a window is closing ?

docs/docs/integrations/overview.md
```diff
@@ -120,6 +120,14 @@ Manage your Flagsmith Change Requests inside ServiceNow. [Learn more](/integrati
 You can integrate Flagsmith with Slack. Send flag change events from Flagsmith into your Slack channels.
 [Learn more](/integrations/project-management/slack).
 
+---
+
+<a href="/integrations/project-management/jira" alt="GitHub"><img width="30%" src="/img/integrations/github/github-logo.svg"/></a>
```
kyle-ssg: URL looks incorrect

docs/docs/integrations/project-management/github.md
```diff
@@ -0,0 +1,34 @@
+---
```
kyle-ssg, line 1: Should this be under project management? Not sure if that's the right category

frontend/web/components/App.js
```diff
@@ -183,7 +192,15 @@ const App = class extends Component {
             )
             return
           }
-
+          if (
```
kyle-ssg: I could be being silly but is this duplicated from 170 ?

frontend/web/components/GitHubRepositoriesSelect.tsx
```diff
@@ -0,0 +1,76 @@
...
+    }
+  }, [isSuccessCreatedGithubRepository])
+
+  return (
+    <div style={{ width: '100%' }}>
+      <Select
+        size='select-md'
+        placeholder={'Select Your Repository'}
+        disabled={disabled}
+        onChange={(r: any) => {
```
kyle-ssg: {value:string}

frontend/web/components/GithubRepositoriesTable.tsx
```diff
@@ -0,0 +1,77 @@
...
+            <Flex className='table-column px-3'>
+              <div className='font-weight-medium mb-1'>{`${repo.repository_owner} - ${repo.repository_name}`}</div>
+            </Flex>
+            <div
+              className='table-column  text-center'
+              style={{ width: '80px' }}
+            >
+              <Button
+                onClick={() => {
+                  deleteGithubRepository({
```
kyle-ssg: Should openConfirm this and destructive things

frontend/web/components/IntegrationList.js
```diff
@@ -253,6 +363,8 @@ class IntegrationList extends Component {
               }
             : null
         }
+        githubId={githubId}
```
kyle-ssg: Can we make these github props just a generic meta field, otherwise as we add others that need this kind of custom stuff these props will get out of control. 

```
meta={{githubId, installationId... etc}}
```
In the CreateEditIntegration component you can do (meta as GitHubMeta).githubId

frontend/web/components/GitHubRepositoriesSelect.tsx
novakzaballa: Sorry, corrected

docs/docs/integrations/overview.md
novakzaballa: Thanks, corrected it.

frontend/web/components/modals/CreateFlag.js
novakzaballa: Good catch, removed

frontend/web/components/pages/GitHubSetupPage.tsx
novakzaballa: Ok, updated.

frontend/web/components/pages/GitHubSetupPage.tsx
novakzaballa: Sure, Originally it was used once, the variable was created and typed.

frontend/web/components/pages/GitHubSetupPage.tsx
novakzaballa: Yes, it shows before the child window closes.

docs/docs/integrations/project-management/github.md
novakzaballa: Yes, I think so, as it helps to see the status of your features in your GitHub project(Repo).

frontend/web/components/App.js
novakzaballa: It was part of an old logic, I've updated it and removed it.

frontend/web/components/pages/GitHubSetupPage.tsx
novakzaballa: Updated

frontend/web/components/GithubRepositoriesTable.tsx
novakzaballa: Added.

frontend/web/components/pages/GitHubSetupPage.tsx
novakzaballa: >I don't think this should be a separate page, it should either be part of the integrations page or project settings. If you check out Vercel for example as a good flow for connecting GH.
Testing this out it has to work more like that anyway, since if you've already connected to GitHub there's no callback url, you just see the ability to configure.

The ability to install the app without leaving the app has been added. However, the old flow will take you to the setup page if you want to install it from GitHub.

> Managed to get here with an installation id but see a 500 pointing to the api branch url, I think that should work unless environment vars need adding ?

Yes, a var env named GITHUB_PEM is necessary, which is described in the API PR.

frontend/web/components/modals/CreateFlag.js
novakzaballa: Was removed, and a new component was added with hooks.

docs/docs/integrations/project-management/github.md
novakzaballa: This is already implemented.

frontend/common/types/responses.ts
novakzaballa: This type was deleted.

frontend/common/types/responses.ts
novakzaballa: This type was corrected.

frontend/web/components/OrganisationSelect.js
novakzaballa: Corrected.

frontend/web/components/ProjectFilter.tsx
novakzaballa: This accessory is evaluated solely when utilizing the project filter on the GitHub configuration page to return the correct value. Therefore, renaming it will enhance clarity.

kyle-ssg: Adjusting a reusable component just for the sake of a GitHub integration doesn't seem to make sense. Can't you just pass value appropriately ?

frontend/web/components/IntegrationList.js
novakzaballa: Good, added.

frontend/web/components/ProjectFilter.tsx
```diff
@@ -30,7 +32,13 @@ const ProjectFilter: FC<ProjectFilterType> = ({
         (data || [])?.map((v) => ({ label: v.name, value: `${v.id}` })),
       )}
       onChange={(value: { value: string; label: string }) =>
-        onChange(value?.value || '')
+        onChange(
+          gitHubProjectFilterValidator
```
kyle-ssg: Re-adding this comment, there's no reason to adjust this reusable component just for a GitHub integration. We can just provide {value} in GitHub's usage.

novakzaballa: I need this change to get not only the project id, because I need also the project name.

kyle-ssg: Ok, then I think it'd be better to change onChange to 

```
onChange: (id: string, name:string) => void
```

This would definitely be better than having generic components inherit random features

novakzaballa: I made some changes, in the `onChange`, and removed the `gitHubProjectFilterValidator` prop.

kyle-ssg (approved):

Issue #2874, opened 2023-10-23
Better control of database read replicas


dabeeeenster:
- Add the ability to specify read database regions for failover or performance.
- I want to be able to set the order that Flagsmith uses for reads. Right [now the code randomly](https://github.com/Flagsmith/flagsmith/blob/c231749272d52d7b1a987aa2fdd82855e05f6d07/api/app/routers.py#L4) picks a replica which isn’t super helpful.

zachaysan:
I did a review online of common approaches and here are some of my findings:
1. Cross regional database reads are rare. Most of the documentation focusses on local replicas.
2. There are two main approaches for failover from one replica to another. The more prevalent way of approaching it is using a heartbeat connection to the replica in question (either before handing off to Django or by storing it locally as a cache that's refreshed once a second) and if the replica is still online, hand off the replica to the reader otherwise hand it off to the next available replica (or even the default database if there is none). The secondary approach uses middleware to intercept database queries, this approach looked inferior to the first solution. 
3. Falling back from the primary database to a promoted read replica is not widely covered by online sources. I think it should be possible but we would be venturing into the unknown and would come with unknowns like WAL log non-replication between the primary and the secondary databases.

Given what I've read online so far, I think we can pretty easily handle this ticket the following way:
1. Implement a heartbeat connection to our replicas using a Django `cache` to avoid querying every cycle. 
2. Create a secondary set of `CROSS_REGION_REPLICA_DATABASE_URLS` in `settings.py`. These would not be used unless the local regional replication databases have fallen over.
3. Keep the current `REPLICA_DATABASE_URLS` in `settings.py` which are the first line of querying for reads.
4. In order to support setting orders instead of randomly distributed the reads I suggest a new `settings.py` variable called `REPLICA_READ_STRATEGY` set to either `DISTRIBUTED` which is the current approach of spreading reads across replicas, or set to `SEQUENTIAL` which would try, sequentially, the `REPLICA_DATABASE_URLS` in order and then fallback to `CROSS_REGION_REPLICA_DATABASE_URLS` and follow them in order once they're all exhausted. 

[...]

PR #3300, opened 2024-01-16
feat: Add support for replicas and cross region replicas

zachaysan:

## Changes

Support fallbacking from replicas that go down and support multiple replica options. See [this ticket's comment](https://github.com/Flagsmith/flagsmith/issues/2874#issuecomment-1856492754) for full details.

I hard coded a single default database. Extensive testing by hand to make sure the caches and connections worked, then later I introduced one test that verifies the connection works against the default database and in addition to that I added in several tests against the router logic with mocked logic to mimic database connection issues.


api/app/routers.py
```diff
@@ -1,15 +1,70 @@
...
+        usable = False
+        logger.error(
+            "Encountered exception during connection",
+            exc_info=True,
+        )
+
+    if usable:
+        cache.set(
+            f"db_connection_active.{database}", "online", CONNECTION_CHECK_CACHE_TTL
+        )
```
matthewelwell, line 40: Interesting, I'm not sure I've seen the `cache` used directly like this, I've always used `my_cache = caches[<cache name>]` then called the methods on `my_cache`. How does the core cache work?

api/app/routers.py
```diff
@@ -30,6 +85,27 @@ def allow_relation(self, obj1, obj2, **hints):
...
+    def _get_replica(self, replicas: list[str]) -> None | str:
+        while replicas:
+            if settings.REPLICA_READ_STRATEGY == "DISTRIBUTED":
+                database = random.choice(replicas)
+            elif settings.REPLICA_READ_STRATEGY == "SEQUENTIAL":
+                database = replicas[0]
+            else:
+                assert (
+                    False
+                ), f"Unknown REPLICA_READ_STRATEGY {settings.REPLICA_READ_STRATEGY}"
```
matthewelwell: This should probably raise `ImproperlyConfiguredError` ?

api/app/settings/common.py
```diff
@@ -177,11 +178,33 @@
...
+    CROSS_REGION_REPLICA_DATABASE_URLS = env.list(
+        "CROSS_REGION_REPLICA_DATABASE_URLS",
+        default=[],
+        delimiter=CROSS_REGION_REPLICA_DATABASE_URLS_DELIMITER,
+    )
+    NUM_CROSS_REGION_DB_REPLICAS = len(CROSS_REGION_REPLICA_DATABASE_URLS)
+
+    # DISTRIBUTED spreads the load out across replicas while
+    # SEQUENTIAL only falls back once the first replica connection is faulty
+    REPLICA_READ_STRATEGY = env("REPLICA_READ_STRATEGY", "DISTRIBUTED")
```
matthewelwell: Might be overkill but we could define an enum for this purpose.

api/tests/unit/app/test_unit_app_routers.py
```diff
@@ -0,0 +1,188 @@
...
+    settings.NUM_DB_REPLICAS = 0
+    settings.NUM_CROSS_REGION_DB_REPLICAS = 0
+
+    conn_patch = mocker.MagicMock()
+
+    # All replicas should be ignored.
+    conn_patch.is_usable.return_value = True
+    create_connection_patch = mocker.patch(
+        "app.routers.connections.create_connection", return_value=conn_patch
+    )
```
matthewelwell: Do we actually need this then?

zachaysan: For this one I wanted to make sure that the test would fail if there were replicas present and that all the call counts to the `create_connection_patch` and to the `conn_patch.is_usable` mocks were zero. May be a bit overkill, but I think it's better than introducing a test which would silently fail by falling back to the default database after already trying the replicas and failing there.

api/app/settings/common.py
zachaysan: Easy enough, done.

api/app/routers.py
zachaysan: Ok I'll add this. You're right that it shouldn't be an assert since the person may have fat fingered the input for the setting.

api/app/routers.py
zachaysan: Really? I've used the `cache` directly most of the time unless I need something specific. 

It's just the default cache with the normal key-value-ttl arguments.

https://docs.djangoproject.com/en/5.0/topics/cache/#django.core.cache.cache

api/tests/unit/app/test_unit_app_routers.py
matthewelwell: I'd argue that at least this code is redundant then though, right? 

```
conn_patch.is_usable.return_value = True
```

matthewelwell (commented):
I've left one (minor) comment thread outstanding but on the whole this looks good now. Can we add some docs to this PR?

api/tests/unit/app/test_unit_app_routers.py
zachaysan: Yep that's fair. I've removed that line.

zachaysan:
CI failure is related to this: https://github.com/Flagsmith/flagsmith/issues/3361

matthewelwell:
> CI failure is related to this: #3361

Looks like you've actually got some legit failures.

api/app/routers.py
```diff
@@ -1,15 +1,80 @@
+import logging
 import random
+from enum import Enum
 
 from django.conf import settings
+from django.core.cache import cache
+from django.db import connections
+
+logger = logging.getLogger(__name__)
+CONNECTION_CHECK_CACHE_TTL = 2
```
matthewelwell: nit: imo these should be separated by a (blank) new line whitespace.

api/app/routers.py
```diff
@@ -1,15 +1,80 @@
...
+
+
+class ReplicaReadStrategy(Enum):
+    DISTRIBUTED = "DISTRIBUTED"
+    SEQUENTIAL = "SEQUENTIAL"
+
+
+class ImproperlyConfiguredError(RuntimeError):
+    pass
```
matthewelwell: I don't love that this exception isn't defined in an `exceptions` module, but perhaps it's overkill if `app.exceptions` doesn't already exist?

api/app/routers.py
```diff
@@ -1,15 +1,80 @@
...
+            # This return is the most likely as replicas should be
+            # online and properly functioning.
+            return replica
+
+        # Since no replicas are available, fall back to the cross
+        # region replicas which have worse availability.
+        cross_region_replicas = [
+            f"cross_region_replica_{i}"
+            for i in range(1, settings.NUM_CROSS_REGION_DB_REPLICAS + 1)
+        ]
```
khvn26, line 66: Nit: the `cross_region_replica_` and `replica_` prefixes could be constants used both by the router and the logic in settings module.

khvn26 (approved):
Approved with one extremely minor comment.

matthewelwell (approved):
Also approved with 2 minor comments.

api/app/routers.py
zachaysan: I think it looks fine my way, but I don't mind doing it your way either so I've updated it.

api/app/routers.py
zachaysan: I'm happy to move it there.

api/app/routers.py
zachaysan: I see what you're saying, but I don't think it's worth the legibility loss / added complexity.

PR #3212, opened 2023-12-21
feat: Add metadata fields to core entities (FE)

novakzaballa:
Thanks for submitting a PR! Please check the boxes below:

- [X] I have run [`pre-commit`](https://docs.flagsmith.com/platform/contributing#pre-commit) to check linting
- [x] I have added information to `docs/` if required so people know about the feature!
- [x] I have filled in the "Changes" section below?
- [x] I have filled in the "How did you test this code" section below?
- [x] I have used a [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) title for this Pull Request

## Changes
PR Re-created (ORIGINAL: https://github.com/Flagsmith/flagsmith/pull/2861)
FE changes:
- Add and create metadata to Environments, segments, and Flags
- The feature was wrapped inside the 'enable_metadata' flag

BE changes:
- https://github.com/Flagsmith/flagsmith/pull/3315

## How did you test this code?

- Run `feat/add-metadata-fields-to-core-entities-api` 
- Go to Project settings
- Click the Metadata tab.
- Create Metadata.
- Click on the metadata created, And select the entities you want to have metadata.

PR #3315, opened 2024-01-23
feat: Add metadata fields to core entities (API)

novakzaballa:

## Changes
FE changes: 
- https://github.com/Flagsmith/flagsmith/pull/3212

BE changes:
- Support metadata for segments, and Flags
- Support required metadata model fields for segments, and Flags

Tests added


api/conftest.py
```diff
@@ -504,6 +504,66 @@ def required_a_environment_metadata_field(
...
+
+
+@pytest.fixture()
+def optional_b_segment_metadata_field(organisation, b_metadata_field, segment):
+    segment_type = ContentType.objects.get_for_model(segment)
+
+    return MetadataModelField.objects.create(
+        field=b_metadata_field,
+        content_type=segment_type,
+    )
```
matthewelwell: These should all be typed.

api/conftest.py
```diff
@@ -541,6 +601,16 @@ def environment_content_type():
...
 
 
+@pytest.fixture()
+def feature_content_type():
+    return ContentType.objects.get_for_model(Feature)
+
+
+@pytest.fixture()
+def segment_content_type():
+    return ContentType.objects.get_for_model(Segment)
```
matthewelwell: typing please

api/features/serializers.py
```diff
@@ -245,11 +246,33 @@ def get_num_identity_overrides(self, instance) -> typing.Optional[int]:
...
+                return Project.objects.get(pk=project_pk)
+            except Project.DoesNotExist:
+                raise serializers.ValidationError("Project not found.")
+
+        raise serializers.ValidationError(
+            "Unable to retrieve project for metadata validation."
+        )
+
+
+class UpdateFeatureSerializer(FeatureSerializerWithMetadata):
```
matthewelwell: This should be `UpdateFeatureSerializerWithMetadata` I think so that people don't misuse it?

api/metadata/models.py
```diff
@@ -21,7 +21,19 @@
...
+        "project": lambda project_id: Project.objects.get(
+            id=project_id
+        ).organisation_id,
+    },
+    "segment": {
+        "organisation": lambda org_id: org_id,
+        "project": lambda project_id: Project.objects.get(
+            id=project_id
+        ).organisation_id,
+    },
```
matthewelwell: @gagantrivedi @novakzaballa I'm a little confused as to why this is needed. It seems like we're defining the same thing for each model class. Do we need this mapping dict at all? 

Also, one for @novakzaballa, if we _do_ need to define this, should it have `"project"` as a key too?

api/segments/serializers.py
```diff
@@ -120,6 +126,22 @@ def _update_or_create_segment_rules(
                 child_rules, rule=child_rule, is_create=is_create
             )
 
+    def _update_or_create_metadata(self, metadata_data, segment=None):
```
matthewelwell: typing

api/segments/serializers.py
```diff
@@ -120,6 +126,22 @@ def _update_or_create_segment_rules(
...
+        for metadata_item in metadata_data:
+            metadata_id = metadata_item.pop("id", None)
+            if metadata_item.get("delete"):
+                Metadata.objects.filter(id=metadata_id).delete()
+                continue
```
matthewelwell: Is there any other option than using this behaviour? We only had to introduce that because the segments API was poorly created initially, can we avoid it here? How are metadata items being removed from e.g. features?

api/tests/unit/features/test_unit_features_views.py
matthewelwell, line 1: Commenting here but all the tests across all the files are missing typing.

api/metadata/models.py
```diff
@@ -12,7 +12,7 @@
 
 FIELD_VALUE_MAX_LENGTH = 2000
 
-METADATA_SUPPORTED_MODELS = ["environment"]
+METADATA_SUPPORTED_MODELS = ["environment", "feature", "project", "segment"]
```
matthewelwell: I'm confused as I don't actually see any logic defined for supporting metadata on the project model?

matthewelwell (changes requested):

api/segments/serializers.py
novakzaballa: I think that @gagantrivedi could better answer this, since he wrote the be side of metadata, and I made some adjustments.

api/features/serializers.py
novakzaballa: Corrected.

api/tests/unit/features/test_unit_features_views.py
novakzaballa: I believe that all the tests are now typed.

novakzaballa: I believe that all the tests are now typed.

api/metadata/models.py
novakzaballa: I added this only to return the information about the content type of the project.

api/segments/serializers.py
novakzaballa: Added.

api/conftest.py
novakzaballa: Added

api/conftest.py
novakzaballa: I believe that all the fixture are now typed.

api/metadata/models.py
matthewelwell: I don't understand. My understanding is that this constant is defining the models which support the addition of metadata?

api/segments/serializers.py
matthewelwell: The other implementations of managing metadata for an object use a writeable nested serializer though. Changing this might be significant effort here because of the way in which the segment endpoints work, but I'd certainly prefer to look into it. 

Perhaps for now, to get these PRs over the line, we should just ignore metadata for segments?

api/tests/unit/features/test_unit_features_views.py
```diff
@@ -2229,6 +2230,110 @@ def test_cannot_update_feature_of_a_feature_state(
...
 
+@pytest.mark.parametrize(
+    "client",
+    [lazy_fixture("admin_master_api_key_client"), lazy_fixture("admin_client")],
+)
+def test_create_feature_with_required_metadata_returns_400(
+    project: Project,
+    client: APIClient,
+    required_a_segment_metadata_field,
+):
```
```suggestion
) -> None:
```

(also applies to tests below)

api/tests/unit/metadata/test_views.py
```diff
@@ -301,6 +307,72 @@ def test_create_model_metadata_field(
...
 
 
+def test_create_model_metadata_field_for_features(
+    admin_client: APIClient,
+    a_metadata_field: MetadataField,
+    organisation: Organisation,
+    project_content_type: ContentType,
+    feature_content_type: ContentType,
+    project: Project,
+):
```
matthewelwell:
```suggestion
) -> None:
```

api/tests/unit/features/test_unit_features_views.py
matthewelwell: Yep, mostly! A few still just need the return type adding.

api/metadata/models.py
novakzaballa: According to this code, the only place where this constant is used, and it returns the data of the content types that are in it.
https://github.com/Flagsmith/flagsmith/blob/a1c0ea17c9a27151514de19edfe28ce9b8e32bd4/api/metadata/views.py#L75-L88

api/tests/unit/metadata/test_views.py
novakzaballa: Added.

api/tests/unit/features/test_unit_features_views.py
novakzaballa: Added.

api/tests/unit/features/test_unit_features_views.py
novakzaballa: added.

api/segments/serializers.py
novakzaballa: I think that the segments metadata should not be most used so we could release the feature without metadata for now

api/tests/unit/features/test_unit_features_views.py
```diff
@@ -2229,6 +2230,110 @@ def test_cannot_update_feature_of_a_feature_state(
...
+
+@pytest.mark.parametrize(
+    "client",
+    [lazy_fixture("admin_master_api_key_client"), lazy_fixture("admin_client")],
+)
+def test_create_feature_with_optional_metadata_returns_201(
+    project: Project,
+    client: APIClient,
+    optional_b_feature_metadata_field: MetadataModelField,
+):
```
matthewelwell: Missing return type

api/segments/serializers.py
gagantrivedi: Yeah, no, this code is not correct e.g: if you update this test case: https://github.com/Flagsmith/flagsmith/pull/3315/files#diff-9a4d991aa7d7c0a8180bd898794518ff854111cbe01de6c6a0143c52f344319aR734 and remove metadata from the data it will return 200, i.e: the following test will fail

```
@pytest.mark.parametrize(
    "client",
    [lazy_fixture("admin_master_api_key_client"), lazy_fixture("admin_client")],
)
def test_create_environment_without_required_metadata_returns_400(
    project,
    client,
    required_a_environment_metadata_field,
    optional_b_environment_metadata_field,
):
    # Given
    url = reverse("api-v1:environments:environment-list")
    description = "This is the description"
    data = {
        "name": "Test environment",
        "project": project.id,
        "description": description,
    }

    # When
    response = client.post(url, data=data)

    # Then
    assert response.status_code == status.HTTP_400_BAD_REQUEST
    assert "Missing required metadata field" in response.json()["metadata"][0]


```

api/tests/unit/features/test_unit_features_views.py
```diff
@@ -2229,6 +2230,110 @@ def test_cannot_update_feature_of_a_feature_state(
...
 
 
+@pytest.mark.parametrize(
+    "client",
+    [lazy_fixture("admin_master_api_key_client"), lazy_fixture("admin_client")],
+)
+def test_create_feature_with_required_metadata_returns_400(
+    project: Project,
+    client: APIClient,
+    required_a_segment_metadata_field: MetadataModelField,
```
gagantrivedi: This should be `required_a_feature_metadata_field`

api/tests/unit/features/test_unit_features_views.py
```diff
@@ -2229,6 +2230,110 @@ def test_cannot_update_feature_of_a_feature_state(
...
+) -> None:
+    # Given
+    url = reverse("api-v1:projects:project-features-list", args=[project.id])
+    description = "This is the description"
+    field_value = 10
+    content_type = 9999
+    data = {
+        "name": "Test feature",
+        "description": description,
+        "metadata": [
```
gagantrivedi, line 2339: metadata field should not exists here

api/tests/unit/segments/test_unit_segments_views.py
```diff
@@ -579,6 +580,78 @@ def test_update_segment_delete_existing_rule(project, client, segment, segment_r
...
+    url = reverse("api-v1:projects:project-segments-list", args=[project.id])
+    description = "This is the description"
+    field_value = 10
+    content_type = 9999
+    data = {
+        "name": "Test Segment",
+        "description": description,
+        "project": project.id,
+        "rules": [{"type": "ALL", "rules": [], "conditions": []}],
+        "metadata": [
```
gagantrivedi, line 668: Metadata field should not exist here if we are testing for without Metadata

gagantrivedi (changes requested):

api/metadata/models.py
gagantrivedi: Hmm yeah, we should be able to create generic functions for them (maybe in the serializer)

api/tests/unit/features/test_unit_features_views.py
novakzaballa: Corrected

api/tests/unit/features/test_unit_features_views.py
novakzaballa: Corrected

api/metadata/models.py
novakzaballa: How do you propose to do this?

gagantrivedi: Okay, so this dict does two things:
 1. It provides a way to get organisation_id of the requirement object (those lambda functions) that can be solved by adding a property like:
 ```
     @property
    def get_org_id(self):
        return self.organisation_id


 ```
 to all the requirement objects(i.e: project and organisation)
 and update the serializer like this:
```


class MetaDataModelFieldSerializer(DeleteBeforeUpdateWritableNestedModelSerializer):
    is_required_for = MetadataModelFieldRequirementSerializer(many=True, required=False)

    class Meta:
        model = MetadataModelField
        fields = ("id", "field", "content_type", "is_required_for")

    def validate(self, data):
        data = super().validate(data)
        for requirement in data.get("is_required_for", []):
            org_id = (
                requirement["content_type"]
                .model_class()
                .objects.get(id=requirement["object_id"])
                .get_org_id
            )
            if org_id != data["field"].organisation_id:
                raise serializers.ValidationError(
                    "The requirement organisation does not match the field organisation"
                )
        return data


```
 2: It provides a way to check that the relationship is valid, i.e: you should not be able to nest project under environment (not really possible right now but just an example)
 To solve that, we can turn this dict(`SUPPORTED_REQUIREMENTS_MAPPING`) into something like
 
 ```
 SUPPORTED_REQUIREMENTS_MAPPING = {
    "environment": ["organisation", "project"],
    "feature": ["organisation", "project"],
    ....
 ```
 and check that the relationship is valid in the validate method

novakzaballa: Ok, I applied the changes you proposed.

api/segments/serializers.py
novakzaballa: I appreciate the heads-up and the example. I'll review the test at the provided URL for similar adjustments. Thanks for your input.

api/projects/models.py
```diff
@@ -194,6 +194,10 @@ def show_edge_identity_overrides_for_feature(self) -> bool:
             == IdentityOverridesV2MigrationStatus.COMPLETE
         )
 
+    @property
+    def get_org_id(self):
```
gagantrivedi: typing

api/tests/unit/segments/test_unit_segments_views.py
```diff
@@ -579,6 +580,78 @@ def test_update_segment_delete_existing_rule(project, client, segment, segment_r
     assert segment_rule.conditions.count() == 0
 
 
+@pytest.mark.parametrize(
+    "client",
+    [lazy_fixture("admin_master_api_key_client"), lazy_fixture("admin_client")],
+)
+def test_create_segment_with_required_metadata_returns_201(
```
gagantrivedi, line 616: We should also add test where metadata filed for segments is required across organisation and not just project (we support that, right? )
ps: we should do the same for features as well

api/tests/unit/segments/test_unit_segments_views.py
zachaysan: The test name is confusing if we're testing _without_ metadata.

gagantrivedi (changes requested):

api/tests/unit/features/test_unit_features_views.py
```diff
@@ -2229,6 +2230,102 @@ def test_cannot_update_feature_of_a_feature_state(
     )
 
 
+@pytest.mark.parametrize(
+    "client",
+    [lazy_fixture("admin_master_api_key_client"), lazy_fixture("admin_client")],
+)
+def test_create_feature_with_required_metadata_returns_400(
```
gagantrivedi: This test is named incorrectly *without..*

api/tests/unit/segments/test_unit_segments_views.py
```diff
@@ -579,6 +580,78 @@ def test_update_segment_delete_existing_rule(project, client, segment, segment_r
...
+        == required_a_segment_metadata_field.id
+    )
+    assert response.json()["metadata"][0]["field_value"] == str(field_value)
+
+
+@pytest.mark.parametrize(
+    "client",
+    [lazy_fixture("admin_master_api_key_client"), lazy_fixture("admin_client")],
+)
+def test_create_segment_with_required_metadata_returns_400(
```
gagantrivedi: Same as feature one: this test should also be renamed

novakzaballa: Renamed.

api/tests/unit/features/test_unit_features_views.py
novakzaballa: Renamed.

api/tests/unit/segments/test_unit_segments_views.py
novakzaballa: Removed, and renamed.

api/segments/serializers.py
novakzaballa: Ok, I reviewed it, you were right, there was an issue with the SegmentSerializer, now fixed.

api/tests/unit/segments/test_unit_segments_views.py
novakzaballa: Done.

api/conftest.py
```diff
@@ -532,7 +647,10 @@ def environment_metadata_a(environment, required_a_environment_metadata_field):
 
 
 @pytest.fixture()
-def environment_metadata_b(environment, optional_b_environment_metadata_field):
+def environment_metadata_b(
+    environment: Environment,
+    optional_b_environment_metadata_field: MetadataModelFieldRequirement,
```
gagantrivedi: Type hint in wrong here, right? It should be `MetadataModelField`

api/conftest.py
```diff
@@ -521,7 +633,10 @@ def optional_b_environment_metadata_field(organisation, b_metadata_field, enviro
 
 
 @pytest.fixture()
-def environment_metadata_a(environment, required_a_environment_metadata_field):
+def environment_metadata_a(
+    environment: Environment,
+    required_a_environment_metadata_field: MetadataModelFieldRequirement,
```
gagantrivedi: Type hint in wrong here, right? It should be MetadataModelField

api/metadata/models.py
gagantrivedi: This feels wrong… can we use `SUPPORTED_REQUIREMENTS_MAPPING ` and generate this list on the fly?

api/features/serializers.py
```diff
@@ -270,11 +271,33 @@ def get_last_modified_in_current_environment(
...
+):
+    metadata = MetadataSerializer(required=False, many=True)
 
     class Meta(ListCreateFeatureSerializer.Meta):
-        read_only_fields = ListCreateFeatureSerializer.Meta.read_only_fields + (
+        fields = ListCreateFeatureSerializer.Meta.fields + ("metadata",)
+
+    def get_project(self, validated_data: dict = None) -> Project:
+        view = self.context.get("view")
+
```
gagantrivedi: Can't we get project from `self.context["project"]`?

api/segments/serializers.py
```diff
@@ -156,6 +167,25 @@ def _update_or_create_segment_rules(
                 child_rules, rule=child_rule, is_create=is_create
             )
 
+    def _update_or_create_metadata(
```
gagantrivedi, line 172: Do we have to do this because we are not using NestedModelSerializer Here?

gagantrivedi (changes requested):
This looks pretty good now - have left some minor comments.
Great work @novakzaballa

api/conftest.py
novakzaballa: Corrected.

api/conftest.py
novakzaballa: Corrected.

api/features/serializers.py
novakzaballa: Sure, Changed.

novakzaballa:
@gagantrivedi I'm not sure what you're referring to here: https://github.com/Flagsmith/flagsmith/pull/3315#discussion_r1534062976

api/segments/serializers.py
novakzaballa: Yes.

api/metadata/models.py
gagantrivedi:
```
from itertools import chain
need_content_type_of = list(chain.from_iterable((key, *value) for key, value in SUPPORTED_REQUIREMENTS_MAPPING.items()))
qs = ContentType.objects.filter(model__in=need_content_type_of) 
```

api/segments/serializers.py
gagantrivedi: I understand that it is not related to your code, but do we know why are we not using NestedModelSerializer  here?

novakzaballa: I think the decision to not employ NestedModelSerializer here is based on the simplicity and structure of the models and data relationships we are dealing with.

api/metadata/models.py
novakzaballa: Implemented.

gagantrivedi: We should be able to remove this constant now?

api/metadata/models.py
```diff
@@ -6,22 +6,20 @@
...
-    "environment": {
-        "organisation": lambda org_id: org_id,
-        "project": lambda project_id: Project.objects.get(
-            id=project_id
-        ).organisation_id,
-    }
+    "environment": ["organisation", "project"],
+    "feature": ["organisation", "project"],
+    "segment": ["organisation", "project"],
+    "project": [],
```
gagantrivedi: We don't need `project` and `organisation` to generate that content type, right? Did we add them for something else?

gagantrivedi (changes requested):

api/metadata/models.py
novakzaballa: Yes, it was removed.

api/metadata/models.py
novakzaballa: We need the content type of both to create a MetadataModelFieldRequirement which requires the content type ID of the organization or the project, their respective IDs, and the model_field ID of the segment, feature, or environment.

gagantrivedi: But that's what this code [here](https://github.com/Flagsmith/flagsmith/pull/3315/files#diff-679ebeca94a92941172d9fd3c4891c62d17e65226a8b3ed19dcc8e62f688e907R86) is doing, right?

novakzaballa: We needed this to return those content types in the endpoint: https://github.com/Flagsmith/flagsmith/blob/f98f8c3974e9a1eeec3582fad548e4bb0c545068/api/metadata/views.py#L84,

gagantrivedi: I think there is some confusion here...

```
need_content_type_of = list(
            chain.from_iterable(
                (key, *value) for key, value in SUPPORTED_REQUIREMENTS_MAPPING.items()
            )
        )
```
This code returns a list that includes both key and values, i.e: we don't need to add `project` and `organisation` in order to return their content type

novakzaballa: OK, I think it's clear now, I made the correction, thank you for the help @gagantrivedi.

api/features/serializers.py
```diff
@@ -270,11 +271,32 @@ def get_last_modified_in_current_environment(
...
+        fields = ListCreateFeatureSerializer.Meta.fields + ("metadata",)
+
+    def get_project(self, validated_data: dict = None) -> Project:
+        try:
+            project = self.context.get("project")
+            if project:
+                return project
+            else:
+                raise serializers.ValidationError("Project not found.")
+        except Exception as exc:
```
gagantrivedi: We can get rid of this except now, right?

gagantrivedi (changes requested):

api/features/serializers.py
novakzaballa: Removed.

gagantrivedi (approved):

api/projects/models.py
```diff
@@ -194,6 +194,10 @@ def show_edge_identity_overrides_for_feature(self) -> bool:
             == IdentityOverridesV2MigrationStatus.COMPLETE
         )
 
+    @property
+    def get_org_id(self) -> int:
+        return self.organisation_id
```
matthewelwell: I don't understand why this exists and, if it does need to exist, why it's not named appropriately. `property`s should not have the `get_` prefix by convention. 

It should either be: 
 1. Removed entirely
 2. Changed to `org_id` if it needs to be a property
 3. Change to a method, not a property (and probably updated to `get_organisation_id` - I don't see the need to shorten `organisation` here)

Note: I _really_ don't like option 2 here because we're seemingly just reimplementing the existing `organisation_id` property provided by django. Let's try to go with option 1 or 3.

matthewelwell (commented):
@novakzaballa added a couple of (I think fairly minor) comments here. 

Also, can you and @gagantrivedi make sure that all the existing conversation are resolved. I think the onus here is on @gagantrivedi to resolve the conversations where he requested changes.

api/projects/models.py
novakzaballa: I opted for the first option, this property has been removed, and the organisation ID is obtained directly.

api/tests/unit/features/test_unit_features_views.py
```diff
@@ -2878,7 +3011,7 @@ def test_feature_list_last_modified_values(
         Feature.objects.create(name=f"feature_{i}", project=project)
 
     # When
-    with django_assert_num_queries(16):  # TODO: reduce this number of queries!
+    with django_assert_num_queries(19):  # TODO: reduce this number of queries!
```
matthewelwell, line 3105: What are the extra queries we're adding here, and have we confirmed that we don't have any N+1 issues?

novakzaballa: You're right, I found an N+1 problem. I've fixed it by adding metadata to the prefetch_related used in the queryset.

matthewelwell (approved):

PR #3325, opened 2024-01-24
chore: Convert organisations models TestCase to normal test function

zachaysan:

## Changes

I'm working through re-writing the TestCase classes into normal test functions. Rewrote the existing TestCase and replaced with fixtures instead of recreating everything. There are other PRs that are similar to this one.

I deleted one of the tests that was providing no value and I was on the fence about the value of a few of the other ones. Feel free to ask me to just remove them if they're too simple to be worthwhile to keep.

I manually changed the tests one at a time to refactor them as I went then ran the full test suite afterwards. Also, I changed assertions as I went to the new fixtures which resulted in shorter tests.


api/tests/unit/organisations/test_unit_organisations_models.py
```diff
@@ -29,72 +27,63 @@
...
-        subscription.refresh_from_db()
-        # Subscription has been immediately transformed to free.
-        assert subscription.cancellation_date is None
-        assert subscription.subscription_id is None
-        assert subscription.billing_status is None
-        assert subscription.payment_method is None
-        assert subscription.plan == FREE_PLAN_ID
+
+
+def test_has_paid_subscription_true(db: None) -> None:
```
gagantrivedi: Can we update the test name to add the name of the class as well... something like `test_organisation...` it looks a bit confusing otherwise

gagantrivedi (changes requested):

api/tests/unit/organisations/test_unit_organisations_models.py
zachaysan: Sure I updated that and other tests in the module.

api/tests/unit/organisations/test_unit_organisations_models.py
```diff
@@ -193,19 +182,14 @@ def test_organisation_is_auto_seat_upgrade_available(organisation, settings):
...
-    def setUp(self) -> None:
-        self.organisation = Organisation.objects.create(name="Test org")
-
-    def tearDown(self) -> None:
-        Subscription.objects.all().delete()
-
-    def test_max_seats_set_as_one_if_subscription_has_no_subscription_id(self):
-        # Given
-        subscription = Subscription.objects.get(organisation=self.organisation)
+def test_organisation_max_seats_set_as_one_if_subscription_has_no_subscription_id(
```
gagantrivedi: nit: the test name is slightly confusing, can we change it to something like: `test_organisation_default_subscription_have_one_max_seat`

gagantrivedi (approved):
Looks good apart from one minor comment

api/tests/unit/organisations/test_unit_organisations_models.py
zachaysan: Sure that's easy

PR #3337, opened 2024-01-29
feat(sse/tracking): Add project and org name to the influx event

gagantrivedi:

## Changes

Add project and org names to the Influx event to make analyzing usage data easier.

Updated the unit test

### DIFF
--- a/api/tests/unit/sse/test_tasks.py
+++ b/api/tests/unit/sse/test_tasks.py
@@ -121,10 +121,30 @@ def test_track_sse_usage(
         [
             call("sse_call"),
             call().field("request_count", 2),
-            call().field().tag("organisation_id", environment.project.organisation_id),
-            call().field().tag().tag("project_id", environment.project_id),
-            call().field().tag().tag().tag("environment_id", environment.id),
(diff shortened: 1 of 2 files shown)


dabeeeenster (approved):

api/tests/unit/sse/test_tasks.py
```diff
@@ -121,10 +121,30 @@ def test_track_sse_usage(
...
+            .tag()
+            .tag("organisation", environment.project.organisation.name),
+            call()
+            .field()
+            .tag()
+            .tag()
+            .tag()
+            .tag()
+            .tag()
+            .time(second_access_log.generated_at),
```
matthewelwell, line 147: This is ugly as heck but I'm not sure if there are any other options. I tried doing _something_ like this but to no avail: 

```python
mocked_influx_point.assert_called_once_with(
    call("see_call").field("request_count", 2).tag("environment_id", environment.id)...
)
```

matthewelwell (approved):

api/tests/unit/sse/test_tasks.py
gagantrivedi: Haha yeah, it is a weird one

PR #3340, opened 2024-01-29
feat: Create api usage function

zachaysan:

## Changes

This PR enables our application to respond to organisations which are using the API close to or above our limits. A few considerations were necessary in order to get all the moving pieces together. First, additional information from Chargebee was required to see what part of the billing cycle the customer is at at the time of comparing their API use. This is then combined with the realtime data we store in Influxdb. And then code that is capable of treating yearly plans as smoothly as monthly ones is in place with relative date counting. Finally an additional model was created in order to keep track of sending notifications only once for a given threshold in a given month.

Influxdb related code was manually tested with ENV vars set to point to our development influxdb instance. In addition to that one large test is available to validate the rest of the work.


api/app_analytics/influxdb_wrapper.py
```diff
@@ -314,6 +313,46 @@ def get_top_organisations(date_range: str, limit: str = ""):
     return dataset
 
 
+def get_current_api_usage(organisation_id: int, date_range: str) -> int:
+    """
+    Query influx db top used organisations
```
matthewelwell: Think this is stale?

api/app_analytics/influxdb_wrapper.py
```diff
@@ -314,6 +313,46 @@ def get_top_organisations(date_range: str, limit: str = ""):
...
+
+    :param organisation_id: filtered organisation
+    :param date_range: data range for current api usage window
```
matthewelwell, line 321: I assume this is some kind of proprietary influx date range value, can we either give more information here or (perhaps ideally) accept actual python datetimes and convert to whatever influx needs here?

api/organisations/tasks.py
```diff
@@ -73,3 +86,85 @@ def finish_subscription_cancellation():
...
+        "organisation": organisation,
+        "matched_threshold": matched_threshold,
+    }
+    send_mail(
+        subject=f"Flagsmith API use has reached {matched_threshold}%",
+        message=render_to_string("organisations/api_usage_notification.txt", context),
+        from_email=settings.DEFAULT_FROM_EMAIL,
+        recipient_list=recipient_list,
+        fail_silently=True,
+    )
```
matthewelwell: We may need to discuss this, but we should use an HTML template here.

api/organisations/templates/organisations/api_usage_notification.txt
```diff
@@ -0,0 +1,7 @@
+Hi there,
+
+API usage for {{ organisation.name }} has reached {{ matched_threshold }} within the current subscription period. Please consider upgrading your organisations account limits.
```
matthewelwell: We might need to be more aggressive for the higher thresholds (100 / 120) but for now this is probably fine.

api/organisations/tasks.py
```diff
@@ -73,3 +86,85 @@ def finish_subscription_cancellation():
...
+@register_recurring_task(
+    run_every=timedelta(hours=12),
+)
+def handle_api_usage_notifications():
+    for organisation in Organisation.objects.filter(
+        subscription_information_cache__current_billing_term_starts_at__isnull=False,
+        subscription_information_cache__current_billing_term_ends_at__isnull=False,
+    ).select_related(
+        "subscription_information_cache",
+    ):
```
matthewelwell, line 175: I'm a little bit unsure how / if we plan on backfilling data for existing organisations that don't have these billing_term attributes.

matthewelwell (changes requested):
I've added a few comments to specific pieces here but I'm also concerned that we might need to consider how this affects the dashboard functionality to show organisations their usage.

My understanding is that we haven't changed that at all here, but we perhaps should (although perhaps a separate PR makes sense if we can wrap this logic in e.g. an env var or feature flag).

I'm talking about this graph in the 'Usage' section of the organisation settings. 

image

api/app_analytics/influxdb_wrapper.py
zachaysan: Yeah, I've updated it.

api/app_analytics/influxdb_wrapper.py
zachaysan: Hmm. Well all the existing functions in this module uses the `date_range: str` pattern and I sorta see why. The module is about being a wrapper around influx db and I don't find that interacting with it is onerous.

matthewelwell:
@zachaysan as discussed, I think this PR should include a small informational message on the Organisation Usage page which states the following: 

> These numbers show the API usage for the last 30 days, not your current billing period which may differ.

api/organisations/tasks.py
zachaysan: Ok this is implemented now, though as we discussed more follow on work will come with better visuals for our emails.

api/organisations/templates/organisations/api_usage_notification.txt
zachaysan: I decided that we should implement these higher threshold emails now, so I've created a second set of emails and wrote them out in both html and txt formats.

api/organisations/tasks.py
zachaysan: Well there are two different cases to consider:

1. Existing ChargeBee subscribers.
2. Non-ChargeBee subscribers.

For the first case we rely on subscriptions coming in regularly enough that the billing terms should be added in eventually and should be kept in sync often enough as well.

For the second case, which I think is mostly non-SaaS users that have set up outside arrangements the current approach will need some sort of addition in order for it to work. But even there, I'm not sure what the source of truth is for those arrangements. Where do we record their billing dates and could we work that in?

zachaysan:
> @zachaysan as discussed, I think this PR should include a small informational message on the Organisation Usage page which states the following:
> 
> > These numbers show the API usage for the last 30 days, not your current billing period which may differ.

As discussed, I included it on the existing note in the page.

api/app_analytics/influxdb_wrapper.py
matthewelwell: Given our time again, I'd like to see the influx wrapper require more human understandable inputs, rather than relying on the knowledge of Influx's query language. But... you're right that the rest of the module behaves this way so it's not a hill that I'll die on.

api/organisations/templates/organisations/api_usage_notification.txt
matthewelwell: Nice ok 👍

api/organisations/tasks.py
matthewelwell: >For the first case we rely on subscriptions coming in regularly enough that the billing terms should be added in eventually and should be kept in sync often enough as well.

I'm not sure I fully understand this. What will trigger the data being updated for an existing chargebee subscription? 

You're right that we don't really need to care about non-CB subs though, yes.

matthewelwell: As discussed, this data will be updated on any webhook from chargebee that includes the subscription. Worst case scenario is that an annual subscription doesn't get updated for 1 year, but we can manually trigger a webhook from chargebee to update this if necessary. Under normal circumstances, this should only ever be 1 month before we have all paying subscribers up to date.

api/app_analytics/influxdb_wrapper.py
zachaysan: This has been created as an issue so we can handle the whole file in one pass, as discussed.

https://github.com/Flagsmith/flagsmith/issues/3417

api/organisations/tasks.py
```diff
@@ -73,3 +86,93 @@ def finish_subscription_cancellation():
...
+) -> None:
+    """
+    Send notification to admins that the API has breached a threshold.
+    """
+    recipient_list = list(
+        FFAdminUser.objects.filter(
+            userorganisation__organisation=organisation,
+            userorganisation__role=OrganisationRole.ADMIN,
+        ).values_list("email", flat=True)
+    )
```
matthewelwell: One thing that came up in a conversation with Greg about this yesterday is whether we can change the distribution list based on the severity of the threshold. Essentially, for the 100 / 120 thresholds, we want to email everyone in the organisation.

api/organisations/templates/organisations/api_usage_notification_limit.html
```diff
@@ -0,0 +1,32 @@
...
+               <td>Hi there,</td>
+
+        </tr>
+
+        <tr>
+
+               <td>
+                 The API usage for {{ organisation.name }} has breached
+                 {{ matched_threshold }}% within the current subscription period.
+                 Please upgrade your organisations account to ensure continued service.
```
matthewelwell, line 14: Nicely worded 👌

api/tests/unit/organisations/test_unit_organisations_tasks.py
```diff
@@ -222,3 +224,84 @@ def test_send_org_subscription_cancelled_alert(db: None, mocker: MockerFixture)
...
+            "               </td>\n\n\n        </tr>\n\n        "
+            "<tr>\n\n               <td>Thank you!</td>\n\n      "
+            "  </tr>\n\n        <tr>\n\n               "
+            "<td>The Flagsmith Team</td>\n\n        "
+            "</tr>\n\n</table>\n"
+        ),
+        from_email="noreply@flagsmith.com",
+        recipient_list=["admin@example.com"],
+        fail_silently=True,
+    )
```
matthewelwell: I guess there's an argument to be said that we should parametrise this test to handle the different threshold 'groups'. The logic is pretty basic so I'll leave it for the purposes of approving this PR, but I think strictly we should test both paths.

api/organisations/tasks.py
zachaysan: Ok that's no problem.

api/tests/unit/organisations/test_unit_organisations_tasks.py
zachaysan: Sounds good. I'm going to write a separate test though. The output of the blobs of parameterization look ugly when the test suite is run with them, see below:

image

api/organisations/tasks.py
zachaysan: This has been added.

api/organisations/tasks.py
```diff
@@ -73,3 +86,103 @@ def finish_subscription_cancellation():
...
+    else:
+        message = "organisations/api_usage_notification_limit.txt"
+        html_message = "organisations/api_usage_notification_limit.html"
+
+        # Since threshold >= 100 include everyone in the organisation.
+        recipient_list = list(
+            FFAdminUser.objects.filter(
+                userorganisation__organisation=organisation,
+            ).values_list("email", flat=True)
+        )
```
matthewelwell: I'd probably have done: 

```suggestion
    recipient_list = FFAdminUser.objects.filter(
        userorganisation__organisation=organisation,
    )

    if matched_threshold < 100:
        message = "organisations/api_usage_notification.txt"
        html_message = "organisations/api_usage_notification.html"

        # Since threshold < 100 only include admins.
        recipient_list = recipient_list.filter(
            userorganisation__role=OrganisationRole.ADMIN,
        )
    else:
        message = "organisations/api_usage_notification_limit.txt"
        html_message = "organisations/api_usage_notification_limit.html"
```

... and then in the send_mail call you can convert to a flat list of emails: 

```python
send_mail(
    ...
    recipient_list=list(recipient_list.values_list("email", flat=True))
```

api/tests/unit/organisations/test_unit_organisations_tasks.py
```diff
@@ -222,3 +224,167 @@ def test_send_org_subscription_cancelled_alert(db: None, mocker: MockerFixture)
...
+    # Now re-run the usage to make sure the notification isn't resent.
+    handle_api_usage_notifications()
+
+    assert (
+        OranisationAPIUsageNotification.objects.filter(
+            organisation=organisation,
+        ).count()
+        == 1
+    )
+    assert OranisationAPIUsageNotification.objects.first() == api_usage_notification
```
matthewelwell, line 313: A neat addition here might be to use the `mailoutbox` fixture available in pytest-django. 

https://pytest-django.readthedocs.io/en/latest/helpers.html#mailoutbox

matthewelwell (approved):
I've approved this as the comments I've left are only minor improvements that aren't strictly necessary.

api/organisations/tasks.py
zachaysan: Ok that's no problem to update.

api/tests/unit/organisations/test_unit_organisations_tasks.py
zachaysan: What a cool fixture! I've updated the code to use that instead of the email send mocks.

PR #3341, opened 2024-01-29
fix: tweak sdk copy

dabeeeenster:
Old: 
image

New:
image

### DIFF
--- a/frontend/web/components/AdminAPIKeys.js
+++ b/frontend/web/components/AdminAPIKeys.js
@@ -224,11 +224,10 @@ export default class AdminAPIKeys extends PureComponent {
         {!!apiKeys && !!apiKeys.length && (
           <PanelSearch
             className='no-pad'
-            title='API Keys'
             items={apiKeys}
             header={
               <Row className='table-header'>
-                <Flex className='table-column px-3'>API Keys</Flex>
+                <Flex className='table-column px-3'>Terraform API Keys</Flex>
                 <Flex className='table-column'>Created</Flex>
                 <div
                   className='table-column text-center'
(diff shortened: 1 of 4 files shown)


khvn26 (approved):

frontend/web/components/AdminAPIKeys.js
```diff
@@ -224,11 +224,10 @@ export default class AdminAPIKeys extends PureComponent {
         {!!apiKeys && !!apiKeys.length && (
           <PanelSearch
             className='no-pad'
-            title='API Keys'
             items={apiKeys}
             header={
               <Row className='table-header'>
-                <Flex className='table-column px-3'>API Keys</Flex>
+                <Flex className='table-column px-3'>Terraform API Keys</Flex>
```
matthewelwell, line 230: Why did this change back to Terraform API keys? This was changed on purpose since we're soon going to support regular API keys (using roles etc.)

dabeeeenster: It's clearer right now. We would probably want to split out legacy keys in a separate table anyway?

matthewelwell: What do you mean by 'legacy keys' ??

dabeeeenster: Ah I mean I'm not sure what we are doing with "Terraform keys" are we going to deprecate them or call them something different?

matthewelwell: They're just going to exist in the same 'API keys' section

matthewelwell: It's the exact same entity under the hood.

PR #3342, opened 2024-01-29
feat(datadog): add source type name to datadog

matthewelwell:

## Changes

Adds the `source_type_name` attribute to requests to the datadog request to display the source in the datadog dashboard as per documentation [here](https://docs.datadoghq.com/integrations/faq/list-of-api-source-attribute-value/).

Modified / added tests. Also made a manual call to the datadog API with the source_type_name parameter set to "flagsmith" to verify it was accepted (as you can see in screenshot below)

image

I also ran the FE locally to verify that the changes made in the `CreateEditIntegration` component work as expected.


docs/docs/integrations/apm/datadog.md
```diff
@@ -42,9 +42,10 @@ The second type of integration allows you to send Flag change events in Flagsmit
 4. Add the Datadog URL into Flagsmith as the Base URL - (This is either `https://api.datadoghq.com/` or
    `https://api.datadoghq.eu/`)
 
-![Datadog](/img/integrations/datadog/datadog-1.png)
+### Custom Source
 
-![Datadog](/img/integrations/datadog/datadog-2.png)
```
matthewelwell, line 47: These were old, stale, images that I think are not necessary. The documentation is clear enough and these images will always become stale over time.

api/integrations/datadog/datadog.py
```diff
@@ -9,14 +9,22 @@
...
 
 class DataDogWrapper(AbstractBaseEventIntegrationWrapper):
-    def __init__(self, base_url: str, api_key: str, session: requests.Session = None):
+    def __init__(
+        self,
+        base_url: str,
+        api_key: str,
+        session: requests.Session = None,
+        use_custom_source: bool = False,
+    ):
```
zachaysan: Nit: Missing ` -> None:`

api/integrations/datadog/datadog.py
```diff
@@ -34,6 +42,9 @@ def generate_event_data(audit_log_record: AuditLog) -> dict:
         }
 
     def _track_event(self, event: dict) -> None:
+        if self.use_custom_source:
+            event["source_type_name"] = SOURCE_TYPE_NAME
```
zachaysan: This reads a bit weird. `SOURCE_TYPE_NAME` should be renamed to something like `FLAGSMITH_SOURCE_TYPE_NAME`.

api/tests/unit/integrations/datadog/test_unit_datadog.py
```diff
@@ -19,28 +24,46 @@ def test_datadog_initialized_correctly(base_url, expected_events_url):
     api_key = "123key"
 
     # When initialized
-    data_dog = DataDogWrapper(base_url=base_url, api_key=api_key)
+    data_dog = DataDogWrapper(
+        base_url=base_url, api_key=api_key, use_custom_source=True
+    )
```
zachaysan, line 29: This makes me think that `use_custom_source` defaulting to `False` is kinda backwards, isn't it? Shouldn't we always default to using `flagsmith` as the source for our very own wrapper around data dog?

api/tests/unit/integrations/datadog/test_unit_datadog_views.py
```diff
@@ -26,6 +30,11 @@ def test_should_create_datadog_config_when_post(
     assert response.status_code == status.HTTP_201_CREATED
     assert DataDogConfiguration.objects.filter(project=project).count() == 1
 
+    created_config = DataDogConfiguration.objects.first()
```
zachaysan: Nit: filter by project is either necessary or unnecessary for the test.

frontend/web/components/base/forms/Checkbox.tsx
```diff
@@ -0,0 +1,34 @@
...
+    onChange(!checked)
+  }
+
+  return (
+    <>
+      <input type='checkbox' />
+      <label
+        onClick={handleChange}
+        className='mb-0'
+        htmlFor='mailinglist'
```
zachaysan: I think I understand most of this file, but I don't understand why the htmlFor would point to mailinglist.

zachaysan: After reading the rest of the code I understand that its being used on the home page in an email list. It would be better if something as reusable as a checkbox was less hardcoded to its use elsewhere.

zachaysan (approved):
Lots of minor small nits but the overall picture looks good.

api/tests/unit/integrations/datadog/test_unit_datadog.py
matthewelwell: Well, the issue here is backwards compatibility. If existing customers that are using the datadog integration have filtering rules set up based on the previous (default) source type, then releasing this for all users would break those rules. I've added the default as False, then changed it at the model level to True so that all new instances get the custom source type name.

frontend/web/components/base/forms/Checkbox.tsx
matthewelwell: Yes, good question. I'll defer to @kyle-ssg who wrote that code for an answer here.

kyle-ssg: Oops, this is fixed.

matthewelwell:
@kyle-ssg I've just retested this in the Uffizzi environment and it's not working. I'm sure that it previously was sending the correct data, but it seems like it's no longer doing so - I can't quite see why though. 

This should include the `use_custom_source` attribute as well, but it's excluding it from the request. 

image

frontend/web/components/modals/CreateEditIntegrationModal.js
```diff
@@ -61,6 +61,7 @@ const CreateEditIntegration = class extends Component {
     const handleOauthSignature = (res, isProject) => {
       const signature = res && res.signature
       if (signature) {
+        q
```
matthewelwell: qué?

frontend/web/components/modals/CreateEditIntegrationModal.js
```diff
@@ -217,7 +218,11 @@ const CreateEditIntegration = class extends Component {
                   <Input
                     id={field.label.replace(/ /g, '')}
                     ref={(e) => (this.input = e)}
-                    value={this.state.data[field.key] || field.default}
+                    value={
+                      typeof this.state.data[field.key] === 'undefined'
+                        ? this.state.data[field.key]
+                        : field.default
+                    }
```
matthewelwell: Shouldn't this be inverted? 

```js
                    value={
                      typeof this.state.data[field.key] !== 'undefined'
                        ? this.state.data[field.key]
                        : field.default
                    }
```

PR #3346, opened 2024-01-29
feat: Add role API key (BE)

novakzaballa:

## Changes

- BE code from https://github.com/Flagsmith/flagsmith/pull/3042

### DIFF
--- a/api/api_keys/models.py
+++ b/api/api_keys/models.py
@@ -1,4 +1,6 @@
+from django.conf import settings
 from django.db import models
+from django_lifecycle import BEFORE_UPDATE, LifecycleModelMixin, hook
 from rest_framework_api_key.models import AbstractAPIKey, APIKeyManager
 from softdelete.models import SoftDeleteManager, SoftDeleteObject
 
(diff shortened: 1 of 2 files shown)


api/api_keys/models.py
```diff
@@ -18,3 +20,12 @@ class MasterAPIKey(AbstractAPIKey, SoftDeleteObject):
 
     objects = MasterAPIKeyManager()
     is_admin = models.BooleanField(default=True)
+
+    @hook(BEFORE_UPDATE, when="is_admin", was=False, is_now=True)
+    def delete_role_api_keys(
```
matthewelwell: We should add a test for this method.

matthewelwell (changes requested):

api/api_keys/models.py
novakzaballa: The test is here: https://github.com/Flagsmith/flagsmith-rbac/pull/19

matthewelwell: In that case, I think we should move all of this code to a signal in the `RBAC` repository. 

See here for an example: 

https://github.com/Flagsmith/flagsmith-ldap/blob/main/flagsmith_ldap/signals.py
https://github.com/Flagsmith/flagsmith-ldap/blob/main/flagsmith_ldap/apps.py

novakzaballa: Moved to RBAC

matthewelwell (approved):

PR #3354, opened 2024-01-31
fix(webhook/logging): log response code only if response is not none

gagantrivedi:

## Changes

Handle `call_integration_webhook` returning `None` in case of failure
details: https://flagsmith.sentry.io/issues/4874822372/?project=5544478&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=24h&stream_index=0

I have tested it manually but writing a test for this will involve unit testing a private method. Keen to get some input on this

### DIFF
--- a/api/integrations/webhook/webhook.py
+++ b/api/integrations/webhook/webhook.py
@@ -22,9 +22,10 @@ class WebhookWrapper(AbstractBaseIdentityIntegrationWrapper):
 
     def _identify_user(self, data: typing.Mapping) -> None:
         response = call_integration_webhook(self.config, data)
-        logger.debug(
-            "Sent event to Webhook. Response code was: %s" % response.status_code
-        )
+        if response:
+            logger.debug(
+                "Sent event to Webhook. Response code was: %s" % response.status_code
+            )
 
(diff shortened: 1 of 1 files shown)


api/integrations/webhook/webhook.py
```diff
@@ -22,9 +22,10 @@ def __init__(self, config: WebhookConfiguration):
 
     def _identify_user(self, data: typing.Mapping) -> None:
         response = call_integration_webhook(self.config, data)
```
matthewelwell, line 24: Under what circumstances does `call_integration_webhook` return `None` ?

gagantrivedi: On error: https://github.com/Flagsmith/flagsmith/blob/main/api/tests/unit/webhooks/test_unit_webhooks.py#L324

matthewelwell: Ah ok, interesting. This solution seems fine for now I guess, but perhaps my original solution was wrong, and we should just try/except the `call_integration_webhook` function 🤔

matthewelwell (approved):

PR #3359, opened 2024-01-31
fix(audit): add details for override creation

matthewelwell:

## Changes

Adds logic to ensure that change details are included when an override (segment or identity) is created. Previously, the logic is such that only updates are tracked. 

Since the only real use case where this is important is for creating overrides, I have restricted this functionality (using a parameter on the history model base class) to only the FeatureState model. 

Added some integration tests.


gagantrivedi:
Tests that are failing for this looks unrelated :confused: 
Edit: Ah, it's related to  https://github.com/Flagsmith/flagsmith/issues/3361

gagantrivedi (approved):

PR #3365, opened 2024-02-01
fix(webhooks): prevent unnecessary organisation webhook tasks

matthewelwell:

## Changes

Prevents tasks being created to call organisation webhooks unnecessarily. Previously this check was done inside the task (and still is for belt and braces) but it seems unnecessary to create the task if we know it won't execute. 

Added unit tests.


api/audit/signals.py
```diff
@@ -27,9 +26,15 @@ def call_webhooks(sender, instance, **kwargs):
...
         else instance.environment.project.organisation
     )
-    call_organisation_webhooks.delay(
-        args=(organisation.id, data, WebhookEventType.AUDIT_LOG_CREATED.value)
-    )
+
+    if (
+        not settings.DISABLE_WEBHOOKS
+        and organisation.webhooks.filter(enabled=True).exists()
+    ):
```
zachaysan: Very minor nit: This would look better if it were inverted and used as an early return from the function. Seems weird to wait to the bottom of the function to discover that `DISABLE_WEBHOOKS` was set.

api/tests/unit/audit/test_unit_audit_signals.py
```diff
@@ -0,0 +1,51 @@
...
+    )
+
+    # When
+    call_webhooks(sender=AuditLog, instance=audit_log)
+
+    # Then
+    mocked_call_organisation_webhooks.assert_not_called()
+
+
+def test_call_webhooks_does_not_create_task_if_organisation_has_no_webhooks(
```
zachaysan, line 32: The two tests you've added are good, but you're missing a test that the task is created if the organisation has webhooks and `DISABLE_WEBHOOKS` is `False`.

zachaysan (approved):
Overall looks good, but we really should add that one test.

zachaysan (approved):
Looks great!

PR #3371, opened 2024-02-02
docs: refactor sdk compatibility

dabeeeenster:
- Also bump Docusaurus
- Also fix a bunch of links

### DIFF
--- a/docs/docs/basic-features/managing-segments.md
+++ b/docs/docs/basic-features/managing-segments.md
@@ -193,6 +193,19 @@ after being divided by 2.
 
 `userId % 2 == 0`
 
+These minimum SDK versions support segments with the `Modulo` operator in
+[local evaluation mode](/clients/overview#local-evaluation):
+
+- Python SDK: `2.3.0+`
+- Java SDK: `5.1.0+`
+- .NET SDK: `4.2.0+`
+- NodeJS SDK: `2.4.0+`
+- Ruby SDK: `3.1.0+`
+- PHP SDK: `3.1.0+`
+- Go SDK: `2.2.0+`
+- Rust SDK: `0.2.0+`
+- Elixir SDK: `1.1.0+`
+
 </TabItem>
 </Tabs>
 
--- a/docs/docs/basic-features/managing-segments.md
+++ b/docs/docs/basic-features/managing-segments.md
@@ -2,6 +2,8 @@
 description: Group your users based on a set of rules, then control Feature Flags and Remote Config for those groups.
 ---
 
+import Tabs from '@theme/Tabs'; import TabItem from '@theme/TabItem';
+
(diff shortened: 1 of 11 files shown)


docs/docs/basic-features/managing-segments.md
```diff
@@ -240,3 +240,30 @@ More simply, the order of precedence is:
 1. Identity
 2. Segment
 3. Flag
+
+## SDK Compatibility
+
+### `In` Segment operator
```
matthewelwell: Hmm... I actually meant more that we should create a section for the `IN` operator with the other operators, e.g. the Modulo operator has it's own section [here](https://github.com/Flagsmith/flagsmith/pull/3371/files#diff-5c8f900088e9b7550b4df2e8d2db0ad8fc3f9c1c3c99e295688043f7cad924ebR190). I think we should add a new section for the IN operator and define the compatibility there.

dabeeeenster: OK Yep - have shifted some other stuff around its more logically laid out now.

docs/docs/basic-features/managing-segments.md
```diff
@@ -77,35 +79,36 @@ individual value weightings as part of a Segment override.
 :::important
 
 Some of the operators in local evaluation mode are only supported in later SDK versions, see
-[SDK Compatibility](../clients/overview.md#sdk-compatibility).
+[SDK Compatibility](#sdk-compatibility).
```
matthewelwell: I think this link is no longer relevant, right?

matthewelwell (changes requested):

docs/docs/basic-features/managing-segments.md
dabeeeenster: yep pushed

docs/docs/basic-features/managing-segments.md
```diff
@@ -118,7 +114,31 @@ contains the tenancy ID of that user.
...
-### SemVer-aware operators
+:::important
+
+Earlier SDK versions will not work in local evaluation mode if your environment has segments with the `In` operator
+defined.
+
+To keep local evaluation from breaking, please ensure you have your SDK versions updated before you add such segments to
+your environment.
+
+:::
```
matthewelwell, line 125: It's kinda odd that we have this alert here but not in the Modulo tab. We should probably standardise this in some way.

matthewelwell (approved):
Approved because I feel like this review cycle is becoming tedious but I've added one comment that I think we should consider addressing.

docs/docs/basic-features/managing-segments.md
dabeeeenster: OK added to the modulo and merged

PR #3376, opened 2024-02-04
feat: Issue 166 json formatted logs

masmontanas:

## Changes

Extending logging.Formatter with a custom class for json formatted messages.  Unfortunately specifying json format as a string in the LOGGING dictionary presented a corner-case with the axes module (which returns a dictionary [with double-quoted values] with request information).  Traditional string formatting tokens also had to be replaced with '{}' since the tokens were not being interpolated properly after the logging formatter was added.

1. Build `flagsmith:local` container locally.
2. Update `docker-compose.yaml` to use `flagsmith:local` container.
3. Update environment variables in docker-compose file to include `LOG_FORMAT: json`
4. Basic local testing through the UI to ensure logs were properly formatted.


api/app/settings/logger.py
matthewelwell, line 1: I think I'd prefer to see this in e.g. `util.logging`

api/app/settings/logger.py
```diff
@@ -0,0 +1,22 @@
+import json
+import logging
+
+
+class JsonFormatter(logging.Formatter):
+    """Custom formatter for json logs."""
+
+    def format(self, record):
+        try:
+            log_message = record.msg.replace('"', "'").replace("%s", "{}")
```
matthewelwell: I'd love to see a comment here explaining why we're having to do these replaces. I think you explained it in the description of this PR so perhaps we can just add that detail here too?

api/app/settings/logger.py
```diff
@@ -0,0 +1,22 @@
...
+            log_record = {
+                "levelname": record.levelname,
+                "message": formatted_message,
+                "timestamp": self.formatTime(record, self.datefmt),
+                "logger_name": record.name,
+                "process_id": record.process,
+                "thread_name": record.threadName,
+            }
+            return json.dumps(log_record)
+        except Exception as e:
```
matthewelwell: Is it normal in a logging formatter to just catch all exceptions or can we be more explicit here?

matthewelwell (commented):
Hi @masmontanas, thanks for this PR, it looks really good on the whole! I've added a few minor comments but I think we can get this merged very soon for sure.

api/app/settings/logger.py
masmontanas: Good point, was just trying to avoid an unhandled exception/app crash due to logging formatting -- catching the json encoding error is likely good enough.  I'll take a look at the possible exceptions cases and narrow it down.

api/app/settings/logger.py
masmontanas: json.dumps seems to be handling the quote character replacement fine after some testing, so it was removed.  Comments were added for clarity around the issue with the logging module not passing arguments  correctly with legacy string-formatting tokens.

masmontanas:
Hi @matthewelwell , required updates are ready for review.  Thank you.

masmontanas:
Hi @matthewelwell , I saw that e2e tests failed but doesn't seem to be related -- anything needed from my end?

api/util/logging.py
```diff
@@ -0,0 +1,27 @@
...
+            log_record = {
+                "levelname": record.levelname,
+                "message": formatted_message,
+                "timestamp": self.formatTime(record, self.datefmt),
+                "logger_name": record.name,
+                "process_id": record.process,
+                "thread_name": record.threadName,
+            }
+            return json.dumps(log_record)
+        except json.JSONDecodeError as e:
```
khvn26: I don't think this is a correct exception class to catch. `JSONDecodeError` is relevant to `json.load[s]` interface, not `json.dumps`.

`json.dumps`, when using the default encoder, emits `TypeError` and `ValueError` (see [Python docs](https://docs.python.org/3/library/json.html#json.dump)).

masmontanas: Thank you for catching that, fixed and added some tests.

khvn26 (approved):
I'm approving this, but we need to be conscious that, from observability perspective, it isn't a _great_ solution.

We need to evaluate integrating something like https://django-structlog.readthedocs.io/en/latest/ so Flagsmith provides truly insightful structured logs in the future.

matthewelwell (approved):

matthewelwell:
@masmontanas I've approved, but it looks like you have some legit test failures to resolve.

masmontanas:
Fixed unit test issue.

matthewelwell:
thanks again @masmontanas !

PR #3380, opened 2024-02-05
fix: Avoid errors when missing subscription information cache id

zachaysan:

## Changes

A [Sentry issue](https://flagsmith.sentry.io/issues/4945988284/?alert_rule_id=14542360&alert_type=issue&environment=production&notification_uuid=807455be-47d6-417f-ade1-b075a549842d&project=5544478&referrer=slack) alerted the ability of an unset subscription information cache id during an organisational delete. This code simply checks the presence of the model in question and does not delete it if the id is not present. 

Though only three occurrences at the time of this writing have happened, it's not clear as to how the application got into the state where an model is set without an id. 

I am going to rely on CI since the change is very simple.

### DIFF
--- a/api/organisations/models.py
+++ b/api/organisations/models.py
@@ -267,6 +267,9 @@ class Subscription(LifecycleModelMixin, SoftDeleteExportableModel):
         if not getattr(self.organisation, "subscription_information_cache", None):
             return
 
+        if not self.organisation.subscription_information_cache.id:
+            return
+
         self.organisation.subscription_information_cache.delete()
 
     def prepare_for_cancel(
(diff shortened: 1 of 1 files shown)


api/organisations/models.py
```diff
@@ -267,6 +267,9 @@ def save_as_free_subscription(self):
         if not getattr(self.organisation, "subscription_information_cache", None):
             return
 
+        if not self.organisation.subscription_information_cache.id:
+            return
```
matthewelwell, line 273: Should we add a comment here linking to the sentry issue? Maybe the git history is sufficient. I'll leave it up to your decision.

matthewelwell (approved):

api/organisations/models.py
zachaysan: Yeah a comment here certainly makes sense. Good call.

matthewelwell (approved):

PR #3392, opened 2024-02-07
fix(redis_cache): extend DefaultClient class to add support  for RedisClusterException

gagantrivedi:

## Changes
- Create a client class that converts `RedisClusterException` to `ConnectionInterrupted`(handled/defined by django_redis)
- set `connection_factory` as part of __init__ to avoid setting it explicitly 
-  Add explicit retry

- Adds unit tests
- Tested manually against dead redis cluster to make sure the exceptions are only logged and not raised

### DIFF
--- a/api/core/redis_cluster.py
+++ b/api/core/redis_cluster.py
@@ -11,8 +11,6 @@ Include the following configuration in Django project's settings.py file:
 ```python
 # settings.py
 
-DJANGO_REDIS_CONNECTION_FACTORY = "core.redis_cluster.ClusterConnectionFactory"
-
 "cache_name: {
         "BACKEND": ...,
         "LOCATION": ...,
@@ -69,6 +67,9 @@ class SafeRedisClusterClient(DefaultClient):
             setattr(
                 self, method_name, self._safe_operation(getattr(super(), method_name))
             )
+
+        # Let's use our own connection factory here
+        self.connection_factory = ClusterConnectionFactory(options=self._options)
 
 
 class ClusterConnectionFactory(ConnectionFactory):
--- a/api/core/redis_cluster.py
+++ b/api/core/redis_cluster.py
@@ -12,14 +12,63 @@ Include the following configuration in Django project's settings.py file:
 # settings.py
 
 DJANGO_REDIS_CONNECTION_FACTORY = "core.redis_cluster.ClusterConnectionFactory"
+
+"cache_name: {
+        "BACKEND": ...,
+        "LOCATION": ...,
+        "OPTIONS": {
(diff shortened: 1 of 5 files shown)


api/core/redis_cluster.py
```diff
@@ -57,17 +106,23 @@ def get_connection(self, connection_params: dict) -> RedisCluster:
...
+                    )
+                client_cls_kwargs[key] = value
+
+            # Add explicit retry
+            client_cls_kwargs["retry"] = Retry(DecorrelatedJitterBackoff(), 3)
+            # ... and then build and return the client
+            return RedisCluster(**client_cls_kwargs)
+        except Exception as e:
+            # Let django redis handle the exception
+            raise ConnectionInterrupted(connection=None) from e
```
khvn26, line 127: We're wrapping our own `ImproperlyConfigured` here, is that the plan? If not, I argue we should move the `try:` keyword to directly above line 122.

api/core/redis_cluster.py
```diff
@@ -12,14 +12,63 @@
...
+    ]
+
+    def _safe_operation(self, func):
+        def wrapper(*args, **kwargs):
+            try:
+                return func(*args, **kwargs)
+            except RedisClusterException as e:
+                raise ConnectionInterrupted(connection=None) from e
+
+        return wrapper
```
khvn26: Nit: probably should be a staticmethod.

api/core/redis_cluster.py
```diff
@@ -12,14 +12,63 @@
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Dynamically generate safe versions of methods
+        for method_name in self.SAFE_METHODS:
+            setattr(
+                self, method_name, self._safe_operation(getattr(super(), method_name))
+            )
```
khvn26, line 70: Ugh.

Kind of a dumb question, but have you considered monkeypatching `django_redis.client.default._main_exceptions` instead?

(See [here](https://github.com/jazzband/django-redis/blob/1abeaa538a125ecbbcc63661e427b604bedf5c50/django_redis/client/default.py#L20))

gagantrivedi: Yeah, I did not do it for two reasons
* It's a private variable
* The library supports a less hacky way to do it(i.e: implement your own client)

api/core/redis_cluster.py
gagantrivedi: No, that's part of the plan. I don't think cache miss configuration should take down the entire application

api/core/redis_cluster.py
gagantrivedi: Yeah, done

khvn26 (approved):

PR #3397, opened 2024-02-07
fix: Limit segment rules and conditions

zachaysan:

## Changes

Introduce a limit via settings constant to limit the number of conditions a segment can undergo before rejection.

Two long unit tests one for the update the other for the create. Existing tests can handle the cases where we want to see segments created in response to API input.


api/segments/serializers.py
```diff
@@ -74,13 +85,34 @@ def validate_project_segment_limit(self, project: Project) -> None:
...
-        return super().update(instance, validated_data)
+    def validate_segment_rules_conditions_limit(
+        self, rules_data: dict[str, object]
+    ) -> None:
+        count = self._calculate_condition_count(rules_data)
+        if count > settings.SEGMENT_RULES_CONDITIONS_LIMIT:
+            raise ValidationError(
+                {
+                    "segment": f"The segment has {count} of the maximum condition count "
+                    f"limit of {settings.SEGMENT_RULES_CONDITIONS_LIMIT}."
```
matthewelwell: I think this could be worded better. Perhaps something like the following:

```suggestion
                    "segment": f"The segment has {count} conditions, which exceeds the maximum "
                    f"condition count of {settings.SEGMENT_RULES_CONDITIONS_LIMIT}."
```

This also triggers a couple of questions: 

 1. Will the FE handle this error correctly? 
 2. Are there segments in the database right now that people will not be able to update due to this limit? (I will check this)

matthewelwell: Unfortunately, the answer to (2) is yes. We have 57 such segments in the database already. This indicates we may also have issues for self hosted customers too. 

The biggest problem we have is that we can't just tell people 'you'll have to reduce the size of the segment', because that relies on a PUT request... which will fail. 

So, I think we have a few options (whether they are all feasible / worthwhile I'm not sure): 

 1. Only implement the validation on new segments 
 2. Change the logic so that the limit is enforced in the database at a project level so that we can add exceptions for those customers
 3. Add some logic that only validates newly added conditions? 

My preference would be (2) or (3) for sure, but I'm not sure which is most feasible. 

_Query used to obtain data:_ 
```sql
SELECT s.id, count(*)
FROM segments_segment s
LEFT outer JOIN segments_segmentrule sr1 on s.id = sr1.segment_id
left outer join segments_segmentrule sr2 on sr1.id = sr2.rule_id 
left outer join segments_condition sc on sr2.id = sc.rule_id
group by s.id
having count(*) > 100
order by count(*) desc;
```

zachaysan: Over a call with @matthewelwell: 

Decided to run a one off migration and include a new model to allow whitelisted segments to evade the limits.

Also, need to look into FE if it's really handles this error structure.

api/segments/serializers.py
```diff
@@ -74,13 +85,38 @@ def validate_project_segment_limit(self, project: Project) -> None:
...
-        # remove rules from validated data to prevent error trying to create segment with nested rules
-        del validated_data["rules"]
-        return super().update(instance, validated_data)
+    def validate_segment_rules_conditions_limit(
+        self, rules_data: dict[str, object]
+    ) -> None:
+        count = self._calculate_condition_count(rules_data)
+
+        if self.instance and getattr(self.instance, "whitelisted_segment", None):
+            return
```
matthewelwell: We could probably swap these 2 statements around, right? If it is a whitelisted segment, then it must exceed the limit.

api/segments/migrations/0021_create_whitelisted_segments.py
```diff
@@ -0,0 +1,55 @@
...
+
+    whitelisted_segments = []
+    with connection.cursor() as cursor:
+        cursor.execute(sql)
+        results = cursor.fetchall()
+        for result in results:
+            segment_id = result[0]
+            whitelisted_segments.append(model_class(segment_id=segment_id))
+
+    model_class.objects.bulk_create(whitelisted_segments)
```
matthewelwell, line 32: They're painful to write, but I'd love to see a migration test written for this.

matthewelwell (commented):
Added a couple of comments to the code. I think we should also document this in the system limits section of the docs. 

Have we confirmed yet how the FE behaves in this case too?

api/segments/serializers.py
zachaysan: Oh yeah that was dumb of me. Will move the early return.

zachaysan:
> Have we confirmed yet how the FE behaves in this case too?

The mishandled errors have been fixed in this PR: https://github.com/Flagsmith/flagsmith/pull/3413

zachaysan:
> I think we should also document this in the system limits section of the docs.

Ok no problem. Added this locally.

api/segments/migrations/0021_create_whitelisted_segments.py
zachaysan: Ok this is now accomplished.

matthewelwell (approved):

PR #3398, opened 2024-02-07
chore: (WIP) Test ci to debug testmon issue

zachaysan:
## Changes

Rerunning CI to track down the problem with the testmon CI issue.

Painfully re-running CI to figure out what's going wrong.


zachaysan:
@matthewelwell and I have concluded that removing testmon is the right call given the issues we're having debugging the setup.

PR #3419, opened 2024-02-14
fix(migrations): Fix tags migrations

matthewelwell:

## Changes

**Motivation behind this PR** 

Migrations are currently failing on any existing database (note that it seems that they work on new databases, e.g. for running tests). 

**Implementation**

Removes steps from a migrations file in the tags project which created the new column as nullable, manually populated the defaults and then set the column to not null. 

The reason I added this originally is that, technically, adding a new non-nullable column to a table requires acquiring a temporary lock on the table I believe. Since the tags table is small enough in production however (~3000 records), I think we can not worry about this. 

Reproduced the migration error locally, changed the migration file and re ran to confirm the issue is fixed.


khvn26 (approved):

zachaysan (approved):
PR makes sense and yeah with only that few number of records it should be fast enough.

PR #3426, opened 2024-02-15
fix(db): Fix read replica strategy

matthewelwell:

## Changes

Fixes the logic to look for the actual enum instead of the value of the enum. 

Updated the tests but also ran the API locally with `REPLICA_DATABASE_URLS` configured to point at a replicated database and confirmed that it worked as expected.

### DIFF
--- a/api/app/routers.py
+++ b/api/app/routers.py
@@ -100,9 +100,9 @@ class PrimaryReplicaRouter:
 
     def _get_replica(self, replicas: list[str]) -> None | str:
         while replicas:
-            if settings.REPLICA_READ_STRATEGY == ReplicaReadStrategy.DISTRIBUTED.value:
+            if settings.REPLICA_READ_STRATEGY == ReplicaReadStrategy.DISTRIBUTED:
                 database = random.choice(replicas)
-            elif settings.REPLICA_READ_STRATEGY == ReplicaReadStrategy.SEQUENTIAL.value:
+            elif settings.REPLICA_READ_STRATEGY == ReplicaReadStrategy.SEQUENTIAL:
                 database = replicas[0]
             else:
                 raise ImproperlyConfiguredError(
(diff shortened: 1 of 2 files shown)


khvn26:
Do we need to add a default setting value for the folks that have read replicas enabled?

matthewelwell:
> Do we need to add a default setting value for the folks that have read replicas enabled?

The default is `ReadReplicaStrategy.DISTRIBUTED`, which maintains the previous behaviour.

zachaysan (approved):
I'm a little confused as to why the `.value` was added in the first place. I vaguely remember thinking it was necessary based on something.

khvn26:
> > Do we need to add a default setting value for the folks that have read replicas enabled?
> 
> 
> 
> The default is `ReadReplicaStrategy.DISTRIBUTED`, which maintains the previous behaviour. 

 ̶I̶t̶ ̶s̶h̶o̶u̶l̶d̶ ̶b̶e̶ ̶`̶R̶e̶a̶d̶R̶e̶p̶l̶i̶c̶a̶S̶t̶r̶a̶t̶e̶g̶y̶.̶D̶I̶S̶T̶R̶I̶B̶U̶T̶E̶D̶.̶v̶a̶l̶u̶e̶`̶ ̶s̶h̶o̶u̶l̶d̶n̶'̶t̶ ̶i̶t̶?̶ UPD It should not due to `env.enum` behaviour.

matthewelwell:
> I'm a little confused as to why the `.value` was added in the first place. I vaguely remember thinking it was necessary based on something and I'm surprised that the test suite didn't caught it.

The test suite was essentially monkey patching the settings module to whatever it wanted, so the logic was sound when the setting was monkeypatched to be `.value`. The issue is that the tests weren't exercising the `env.enum` parsing logic which actually returns an instance of the `enum` class.

khvn26 (approved):

PR #3429, opened 2024-02-16
ci: upgrade github actions to node v20

dabeeeenster:


khvn26 (approved):
LGTM

Note: the Uffizzi artifact uploading steps have not been updated due to them apparently uploading the same artifact more than once.

PR #3453, opened 2024-02-22
fix: org switcher

kyle-ssg:

## Changes

Switching from an organisation that's enterprise to one that's not crashes the application, this is because hooks should not return null

Switch from Flagsmith to another org


novakzaballa (approved):
LGTM

PR #3473, opened 2024-02-27
feat: Track leads to Hubspot

zachaysan:

## Changes

Upon a user being successfully added to a new organisation we trigger a task where the user's basic info (first and last name as well as email) are associated with the organisation in Hubspot via their API. The lead tracking implementation was partially mirrored with the existing Pipedrive implementation, including matching leads based off of domain names or regexes as well as using the same abstract base class `LeadTracker` although modelling the responses in the same way didn't seem necessary to get the functionality in place.

Tested the actual API responses with a locally set access token and then moved the relevant responses into mocked requests for the three unit tests that cover the various edge cases.


api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -0,0 +1,78 @@
...
+
+logger = logging.getLogger(__name__)
+
+try:
+    import re2 as re
+
+    logger.info("Using re2 library for regex.")
+except ImportError:
+    logger.warning("Unable to import re2. Falling back to re.")
+    import re
```
zachaysan, line 19: This import logic was to mirror the functionality of the existing pipedrive code which had this way of importing `re`.

dabeeeenster:
Should we remove the pipedrive code here too?

api/integrations/lead_tracking/hubspot/client.py
```diff
@@ -0,0 +1,77 @@
...
+            archived=False,
+        )
+
+        results = response.to_dict()["results"]
+        if not results:
+            return None
+
+        if len(results) > 1:
+            logger.warning(
+                "Hubspot contact endpoint is non-unique but which should not be possible"
```
matthewelwell:
```suggestion
                "Hubspot contact endpoint is non-unique which should not be possible"
```

matthewelwell (approved):
Approved with one minor typo fix.

zachaysan:
> Should we remove the pipedrive code here too?

According to @matthewelwell we're holding onto the pipedrive code for now since there is still a need for a bit.

PR #3499, opened 2024-02-28
fix: self regulate task processor threads (WIP)

matthewelwell:

## Changes

This PR updates the task processor management to self regulate the number of worker threads. If certain threads are determined to be unhealthy, it will kill them and spawn new threads in their place. The main changes are: 

 1. When checking for unhealthy threads, the management process will replace any unhealthy threads with new worker threads
 2. The management command to check the health of the task processor now simply checks that the correct number of threads are running, since the main manager process should maintain this number
 3. I've added a bunch more debug logging which helps in testing

I have tested this manually and it seems to work well, but more testing is probably necessary.

### DIFF
--- a/api/task_processor/management/commands/checktaskprocessorthreadhealth.py
+++ b/api/task_processor/management/commands/checktaskprocessorthreadhealth.py
@@ -3,15 +3,19 @@ import sys
 
 from django.core.management import BaseCommand
 
-from task_processor.thread_monitoring import get_unhealthy_thread_names
+from task_processor.thread_monitoring import get_thread_counts
 
 logger = logging.getLogger(__name__)
 
 
 class Command(BaseCommand):
     def handle(self, *args, **options):
-        if get_unhealthy_thread_names():
-            sys.exit("Task processor has unhealthy threads.")
+        thread_counts = get_thread_counts()
 
-        logger.info("Task processor has no unhealthy threads.")
+        if thread_counts.running != thread_counts.expected:
+            sys.exit(
+                "Processor is running %d threads, but expected %d"
+                % (thread_counts.running, thread_counts.expected)
+            )
+
         sys.exit(0)
--- a/api/task_processor/management/commands/runprocessor.py
+++ b/api/task_processor/management/commands/runprocessor.py
@@ -9,10 +9,7 @@ from django.core.management import BaseCommand
 from django.utils import timezone
 
 from task_processor.task_registry import registered_tasks
-from task_processor.thread_monitoring import (
-    clear_unhealthy_threads,
-    write_unhealthy_threads,
-)
+from task_processor.thread_monitoring import ThreadCounts, write_thread_counts
 from task_processor.threads import TaskRunner
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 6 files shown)


api/task_processor/management/commands/checktaskprocessorthreadhealth.py
```diff
@@ -3,15 +3,19 @@
...
 from django.core.management import BaseCommand
 
-from task_processor.thread_monitoring import get_unhealthy_thread_names
+from task_processor.thread_monitoring import get_thread_counts
 
 logger = logging.getLogger(__name__)
 
 
 class Command(BaseCommand):
     def handle(self, *args, **options):
```
zachaysan, line 12: Small nit: While you're here you may as well add typing for the method.

api/task_processor/management/commands/runprocessor.py
```diff
@@ -111,5 +110,28 @@ def _get_unhealthy_threads(
...
+                TaskRunner(
+                    sleep_interval_millis=sleep_interval_ms,
+                    queue_pop_size=queue_pop_size,
+                )
+                for _ in range(n)
+            ]
+        )
+
+        for thread in self._threads:
+            thread.start()
```
zachaysan, line 137: In the case where unhealthy threads have been removed, while healthy ones were left to run, when the unhealthy threads are replaced with new threads via `_spawn_threads` then this code here would be calling the previously healthy ones via a `.start()` method. 

Is it a costless method call or should the addition of new threads be called `.start()` instead of the whole pool?

api/task_processor/processor.py
```diff
@@ -28,6 +28,8 @@ def run_tasks(num_tasks: int = 1) -> typing.List[TaskRun]:
         executed_tasks = []
         task_runs = []
 
+        logger.debug("Processing %d tasks", len(list(tasks)))
```
zachaysan, line 31: Wouldn't this work better?

```suggestion
        logger.debug("Processing %d tasks", len(tasks))
```

api/task_processor/thread_monitoring.py
```diff
@@ -1,34 +1,23 @@
...
-UNHEALTHY_THREADS_FILE_PATH = "/tmp/task-processor-unhealthy-threads.json"
+from pydantic import BaseModel
 
-logger = logging.getLogger(__name__)
-
-
-def clear_unhealthy_threads():
-    if _unhealthy_threads_file_exists():
-        os.remove(UNHEALTHY_THREADS_FILE_PATH)
+THREAD_COUNTS_FILE_PATH = "/tmp/task-processor-thread-counts.json"
```
zachaysan, line 5: It's not really clear why this file exists. A docstring or comment somewhere would probably be helpful.

zachaysan (approved):
The overall approach looks good and makes sense to me. I have some questions though.

api/task_processor/management/commands/checktaskprocessorthreadhealth.py
matthewelwell: Hmm... I'll see what I can do but this is a django management command which needs the generic args / kwargs so I think all I'll be doing is adding the return type as `None`. I'll see if django 4 has any better info here though.

api/task_processor/management/commands/runprocessor.py
matthewelwell: Good point, I'm not sure - I'll check on this. I've only been testing on a single worker thread so this is areally good catch.

api/task_processor/processor.py
matthewelwell: I'm not sure I can call `len()` on a `RawQuerySet` class - you certainly can't on a `QuerySet` class. I tried to use `.count()` but that doesn't exist on the `RawQuerySet` class.

api/task_processor/thread_monitoring.py
matthewelwell: Yep, good point, I'll add.

api/task_processor/management/commands/runprocessor.py
gagantrivedi: Yeah, calling start second time should raise RuntimeError

matthewelwell:
Closing this as it's probably not a good idea.

Issue #3461, opened 2024-02-26
API usage alerting issues


matthewelwell:
In our deployment this morning a few issues have come up: 

 1. It is currently erroring in staging with the exception provided below, it seems as though it expects usage to be retrieved from influx, but staging is using postgres for analytics. 
 2. We need to be able to test this on a per organisation basis perhaps. We don't want to deploy this and start sending alerts out to all organisations immediately (and perhaps incorrectly?)

```
Traceback (most recent call last):
  File "/app/task_processor/processor.py", line 96, in _run_task
    task.run()
  File "/app/task_processor/models.py", line 66, in run
    return self.callable(*self.args, **self.kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/organisations/tasks.py", line 174, in handle_api_usage_notifications
    _handle_api_usage_notifications(organisation)
  File "/app/organisations/tasks.py", line 145, in _handle_api_usage_notifications
    api_usage = get_current_api_usage(organisation.id, f"{days}d")
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/app_analytics/influxdb_wrapper.py", line 350, in get_current_api_usage
    assert len(result.records) == 1
           ^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError

```

PR #3507, opened 2024-02-28
fix: API usage alerting in production

zachaysan:

## Changes

In order to address https://github.com/Flagsmith/flagsmith/issues/3461 two main issues were handled. First, an incorrectly implemented InfluxDB query was fixed in order to get the full API usage counts in a single scalar. Second, a local evaluation mode Flagsmith client was introduced in order to allowlist organisations into alerting so that any given mistake in the API alerting code won't disrupt too many users. 

The API usage code was tested against the staging InfluxDB setup with a date range window set to about a year to gather the most amount of data. Thousands of organisations were tested without the same failure once the fix was implemented. 

The local evaluation Flagsmith client code was tested locally with settings variables set to appropriate values. During testing an issue with local evaluation cropped up where flags were not present if the method call happened before the local evaluation cache had been set, so this was temporarily worked around with a small sleep on the client's instantiation.


api/integrations/flagsmith/client.py
```diff
@@ -21,14 +19,24 @@
...
 
-    if not _flagsmith_client:
-        _flagsmith_client = Flagsmith(**_get_client_kwargs())
+    try:
+        _flagsmith_client = _flagsmith_clients[name]
+    except (KeyError, TypeError):
+        kwargs = _get_client_kwargs()
+        kwargs["enable_local_evaluation"] = local_eval
+        _flagsmith_client = Flagsmith(**kwargs)
+        # TODO: Remove this once client has been fixed.
```
matthewelwell: Let's link the relevant PR.

api/tests/unit/organisations/test_unit_organisations_tasks.py
```diff
@@ -247,6 +247,8 @@ def test_handle_api_usage_notifications_below_100(
         "organisations.tasks.get_current_api_usage",
     )
     mock_api_usage.return_value = 91
+    mocker.patch("organisations.tasks.get_client")
```
matthewelwell: Should we test what happens when the flag is enabled / disabled?

zachaysan: Sure I've added a new test for when the flag is disabled and reworked some of the mock logic to return the actual expected feature flag value.

api/integrations/flagsmith/client.py
zachaysan: Done, although I'm sorta hoping we get that PR merged and deployed before this code is ready to merge.

matthewelwell: I've just released v3.6.0 with this in it.

zachaysan: Ok great I've removed the temporary shim.

matthewelwell: Well, I'm expecting to also see an updated poetry.lock to update the flagsmith client lib in this PR, right?

api/edge_api/identities/edge_identity_service.py
```diff
@@ -15,4 +15,7 @@ def get_edge_identity_overrides(
             environment_id=environment_id, feature_id=feature_id
         )
     )
-    return [IdentityOverrideV2.parse_obj(item) for item in override_items]
+    for item in override_items:
+        item["environment_id"] = str(item["environment_id"])
+
+    return [IdentityOverrideV2.model_validate(item) for item in override_items]
```
khvn26: I'd avoid double iteration here:

```suggestion
    return [
        IdentityOverrideV2.model_validate(
            {**item, "environment_id": str(item["environment_id"])}
        )
        for item in override_items
    ]
```

api/integrations/flagsmith/client.py
zachaysan: Great catch! I've added the update which came with a bunch of dependency upgrades, which pulled in an update to the `flagsmith-flag-engine` library which required a series of updates to a bunch of tests and some changes to the code that was using the `flagsmith-flag-engine` library.

api/util/mappers/dynamodb.py
```diff
@@ -77,10 +77,12 @@ def map_environment_api_key_to_environment_api_key_document(
 def map_engine_identity_to_identity_document(
     engine_identity: "IdentityModel",
 ) -> Document:
-    return {
+    response = {
         field_name: _map_value_to_document_value(value)
         for field_name, value in engine_identity
     }
+    response["composite_key"] = engine_identity.composite_key
+    return response
```
matthewelwell, line 85: I'm a little unsure why this had to change @khvn26 @zachaysan ?

zachaysan: I believe it was an update to pydantic making computed properties no longer part of the generated document. @khvn26 had even opened an issue on their repo about it and the conclusion was that it was not going to get fixed at least until the first major upgrade.

matthewelwell (approved):

PR #3512, opened 2024-02-29
fix: N+1 on segment overrides for environment-document endpoint

matthewelwell:

## Changes

Fixes an N+1 issue on environment-document endpoint caused by retrieving the environment for each segment override. 

Updated the test for the environment-document endpoint to include segment overrides. Confirmed that the number of queries increased linearly with the number of segment overrides, then reduced once the new select related argument was added.

### DIFF
--- a/api/tests/unit/environments/test_unit_environments_views_sdk_environment.py
+++ b/api/tests/unit/environments/test_unit_environments_views_sdk_environment.py
@@ -5,7 +5,9 @@ from rest_framework import status
 from rest_framework.test import APIClient
 
 from environments.models import Environment, EnvironmentAPIKey
-from features.models import Feature
+from features.feature_types import MULTIVARIATE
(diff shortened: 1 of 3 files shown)


api/tests/unit/environments/test_unit_environments_views_sdk_environment.py
```diff
@@ -44,11 +46,37 @@ def test_get_environment_document(
...
+            default_percentage_allocation=10,
+            string_value="option-2",
+        )
+
     # and the relevant URL to get an environment document
     url = reverse("api-v1:environment-document")
 
     # When
-    with django_assert_num_queries(11):
+    with django_assert_num_queries(13):
```
matthewelwell, line 79: This number increased to 23 before I added the select related. I think the additional 2 queries are legitimate for retrieving the MV data and the segment overrides.

khvn26 (approved):

PR #3537, opened 2024-03-04
chore(self-hosting): default `ENABLE_ADMIN_ACCESS_USER_PASS` to true

matthewelwell:

## Changes

Default `ENABLE_ADMIN_ACCESS_USER_PASS` to True. I don't see a reason not to do this, especially as we create by default an admin user. 

Ran the docker-compose.yml locally and confirmed that the admin was accessible by user / password.


zachaysan (approved):
I'm left just wondering why we ever had this disabled in the first place.

PR #3541, opened 2024-03-04
feat: Add state feature filter

### LINKED ISSUE #3243: Filter flags in dashboard based on flag state

image

zachaysan:

## Changes

Fixes https://github.com/Flagsmith/flagsmith/issues/3243

Introduce the ability to filter based off of the feature's value or enabled status, but only when connected to a given environment. Note that there is a limitation at present when filtering off of features with an integer value since I was not able to find a workable `icontains` that is handled by that workflow in Django.

Manually tested a lot of combinations, kept two of them for the actual tests since they are lengthy tests.


api/features/serializers.py
```diff
@@ -98,6 +98,10 @@ class FeatureQuerySerializer(serializers.Serializer):
         required=False,
         help_text="Integer ID of the environment to view features in the context of.",
     )
+    state_enabled = serializers.BooleanField(
+        allow_null=True, required=False, default=None
+    )
+    state_search = serializers.CharField(required=False, default=None)
```
matthewelwell: Let's add some `help_text` to these so the FE knows what they are for (it will show up in the api docs)

api/features/views.py
```diff
@@ -187,6 +189,48 @@ def get_serializer_context(self):
...
+    def apply_state_to_queryset(
+        self, query_data: dict[str, typing.Any], queryset: QuerySet[Feature]
+    ) -> QuerySet[Feature]:
+        if not query_data.get("environment"):
+            raise serializers.ValidationError(
+                "Environment is required in order to filter by state search or by state enabled"
+            )
+        state_enabled = query_data["state_enabled"]
+        state_search = query_data["state_search"]
+        environment = query_data["environment"]
```
matthewelwell: Let's call this variable `environment_id`?

api/features/views.py
```diff
@@ -187,6 +189,48 @@ def get_serializer_context(self):
...
+        filter_enabled_q = Q()
+        if state_enabled is not None:
+            filter_enabled_q = filter_enabled_q | Q(
+                feature_states__enabled=state_enabled
+            )
+        queryset = queryset.filter(
+            filter_search_q,
+            filter_enabled_q,
+            feature_states__environment=environment,
+        )
```
matthewelwell: Hmnm... I don't think this is actually what we want. I think this is going to return incorrect results because of 2 things: 

1. Versioning
2. Overrides

I think that (like the other PR which returns the feature state in the JSON response body), we'll likely need to use `FeatureState.objects.get_live_feature_states` somehow. I'm not entirely sure how we'll do that. We might need to filter that queryset based on the provided parameters and then just filter the main queryset by the feature ids.

matthewelwell (changes requested):

api/features/views.py
zachaysan: Ok that's easy. However, the query_data already has it named `environment` instead of `environment_id` so that part will be kept the same for backwards compatibility.

matthewelwell: agreed.

api/features/serializers.py
zachaysan: Good idea. Have added locally.

api/features/views.py
zachaysan: As we discussed during standup, I've proceeded with what you suggested. Also as discussed, when we bring the PRs together there will be more requests due to some duplication, but only for this method of searching the features page, so the extra traffic will likely to be small.

api/features/views.py
```diff
@@ -187,6 +189,53 @@ def get_serializer_context(self):
...
+        state_search = query_data["state_search"]
+        environment_id = query_data["environment"]
+
+        filter_search_q = Q()
+        if state_search is not None:
+            filter_search_q = filter_search_q | Q(
+                feature_state_value__string_value__icontains=state_search,
+            )
+
+            if state_search.lower() in {"true", "false"}:
```
matthewelwell:
```suggestion
            if (boolean_search := state_search.lower()) in {"true", "false"}:
```

Let's get the walrus involved here because it's the best thing to ever happen in python!

api/features/views.py
```diff
@@ -187,6 +189,53 @@ def get_serializer_context(self):
...
+        filter_enabled_q = Q()
+        if state_enabled is not None:
+            filter_enabled_q = filter_enabled_q | Q(enabled=state_enabled)
+
+        base_q = Q(
+            identity__isnull=True,
+            feature_segment__isnull=True,
+        )
+        if not getattr(self, "environment", None):
+            self.environment = Environment.objects.get(id=environment_id)
```
matthewelwell, line 233: I'm not sure what scenarios the environment is / isn't set on the instance already?

api/tests/unit/features/test_unit_features_views.py
```diff
@@ -2510,6 +2510,157 @@ def test_list_features_with_intersection_tag(
     assert response.data["results"][0]["tags"] == [tag1.id, tag2.id]
 
 
+def test_list_features_with_sort_by_state_string_and_int(
```
```suggestion
def test_list_features_with_filter_by_state_string_and_int(
```

api/features/serializers.py
```diff
@@ -98,6 +98,17 @@ class FeatureQuerySerializer(serializers.Serializer):
         required=False,
         help_text="Integer ID of the environment to view features in the context of.",
     )
+    state_enabled = serializers.BooleanField(
+        allow_null=True,
+        required=False,
+        default=None,
+        help_text="Boolean value to filter features as enabled or disabled.",
+    )
+    state_search = serializers.CharField(
```
matthewelwell: I'm being picky here but I think I'd rather this was `value_search`

api/features/views.py
```diff
@@ -187,6 +189,53 @@ def get_serializer_context(self):
...
+        if state_search is not None:
+            filter_search_q = filter_search_q | Q(
+                feature_state_value__string_value__icontains=state_search,
+            )
+
+            if state_search.lower() in {"true", "false"}:
+                boolean_search = state_search.lower() == "true"
+                filter_search_q = filter_search_q | Q(
+                    feature_state_value__boolean_value=boolean_search
+                )
```
matthewelwell: This should be:

```suggestion
                filter_search_q = filter_search_q | Q(
                    feature_state_value__boolean_value=boolean_search,
                    feature_state_value__type=BOOLEAN
                )
```

(same goes for the string and integer filters below)

api/tests/unit/features/test_unit_features_views.py
```diff
@@ -2510,6 +2510,157 @@ def test_list_features_with_intersection_tag(
...
+    feature_state_value1.integer_value = 1945
+    feature_state_value1.save()
+
+    feature_state2 = feature2.feature_states.filter(environment=environment).first()
+    feature_state2.enabled = True
+    feature_state2.save()
+
+    feature_state_value2 = feature_state2.feature_state_value
+    feature_state_value2.string_value = None
+    feature_state_value2.boolean_value = True
```
matthewelwell, line 2554: As per my comment above, we might need to also be careful of setting the `type` attribute here.

matthewelwell (changes requested):
On the whole this looks 👌 but I've added a few minor comments where I think some improvements can be made. 

I think there's only one actual necessary change which is related to setting the `type` in the query filter as well.

api/tests/unit/features/test_unit_features_views.py
zachaysan: Yeah great catch. I'll fix this and the other test and of course the type check in the view as well.

api/features/views.py
zachaysan: I think it's more legible without the parenthesis and walrus operator, but I don't want to crush your infectious enthusiasm so I'll do it your way just this one time.

zachaysan: Running the code I've found that it doesn't actually match what we need, which is a value of `True` or `False`, instead it simply holds the value of `state_search.lower()` which is not what we need.

But just to keep you happy here's a walrus all for yourself:

image

matthewelwell: I think we could fix it by moving the parenthesis... but my walrus needs have been appeased (for now), so do as you see fit.

api/features/serializers.py
zachaysan: Ok done and as discussed, we changed `state_enabled` to `is_enabled` at the same time.

api/features/views.py
zachaysan: If I recall correctly, this was done this way in order to play nicely with another PR that also needed to set the environment variable and it wasn't clear which of the methods would run first.

matthewelwell (approved):

PR #3542, opened 2024-03-05
feat: Enabled state filter (Frontend)

### LINKED ISSUE #3243: Filter flags in dashboard based on flag state

kyle-ssg:

## Changes

Frontend functionality for filtering feature enabled state and feature value, based on https://github.com/Flagsmith/flagsmith/pull/3541. This is controlled by the feature flag **feature_enabled_state_filter**.

image

Note: I believe the API is searching based on the project default_value/default_enabled, it should be based on the provided environment's feature state. We'd never want to filter based on defaults.

- Search a feature by value / enabled state


zachaysan (commented):
The code looks good thus far, but just wanted to confirm that we're designing in a search field as well to match feature values based on boolean, string, and integer values.

kyle-ssg:
It’s an easy addition but I couldn’t imagine that common a use case , thinking about itI’ll do it as part of this pr

zachaysan (approved):
Code looks good, but please hold off on merging this until the server issue you highlighted gets fixed.

Edit: Or wrap this all up under a feature flag so we can deploy and fix whenever we want.

kyle-ssg:
Yep its under a feature flag.

PR #3550, opened 2024-03-05
fix: Enable faster feature loading

zachaysan:

## Changes

In order to support faster loading on the feature page an alteration was done to the feature response, if the environment id is present, then the `FeatureState` that matches the correct environment and has the proper `is_live` attributes is set as the feature state for the feature.

I tried to follow the `featurestates` associated code, although the inclusion of the `identity` serializer didn't seem necessary for this endpoint so I used a slightly slimmer serializer.

Since using a queryset to slim the relationship from many-to-one to one-to-one wasn't possible, I took a page from the viewset and associated the individual models manually with an `IN` clause then kept the collection for association in the serializer. 

If the environment is not present then all that is different is that the value for the `feature_state` part of the response is set to `None`. 

None of the existing test methods really suited this test so I made a new one and checked the multivariate part of the response as well as basic functioning of the endpoint as well.


api/features/serializers.py
```diff
@@ -247,6 +251,11 @@ def validate(self, attrs):
 
         return attrs
 
+    def get_feature_state(self, instance: Feature) -> FeatureState | None:
+        for feature_state in self.context.get("feature_states", []):
```
matthewelwell: Couldn't we build a dictionary when we load it into the context, then each lookup here is O(1) rather than O(n)?

api/features/views.py
```diff
@@ -154,8 +154,26 @@ def get_queryset(self):
         )
         queryset = queryset.order_by(sort)
 
+        if environment_id:
+            page = self.paginate_queryset(queryset)
+
+            environment = Environment.objects.get(id=environment_id)
+            feature_states = FeatureState.objects.get_live_feature_states(
+                environment,
+                additional_filters=Q(feature_id__in=[feature.id for feature in page]),
```
matthewelwell: I think we also need to filter out segment overrides and identity overrides here, right?

api/features/serializers.py
```diff
@@ -112,6 +112,9 @@ class ListCreateFeatureSerializer(DeleteBeforeUpdateWritableNestedModelSerialize
     )
     owners = UserListSerializer(many=True, read_only=True)
     group_owners = UserPermissionGroupSummarySerializer(many=True, read_only=True)
+
+    feature_state = serializers.SerializerMethodField()
```
matthewelwell: Let's be explicit with this attribute. It should be `environment_default_feature_state` or maybe just `environment_feature_state` is sufficient.

api/features/serializers.py
zachaysan: Since we're dealing with small collections of lists I don't think it really impacts performance, but I've re-written it to using dictionaries because it was easy to do so.

api/features/views.py
zachaysan: Oh! I was under the wrong impression that the `get_live_feature_states` method handled more than it does. I'll add in the filters for the segment overrides and identity overrides here and in the other PR.

api/features/serializers.py
matthewelwell: Well, some projects will have upwards of 200 features, I think at that scale it could make a difference, no? If we're doing O(200), 200 times.

api/features/serializers.py
zachaysan: Sounds good.

api/features/serializers.py
zachaysan: I was under the impression that the page length was 20 records. If we're dealing with O(200) 200 times it does start to add up.

api/features/serializers.py
```diff
@@ -247,6 +251,14 @@ def validate(self, attrs):
 
         return attrs
 
+    def get_environment_feature_state(self, instance: Feature) -> FeatureState | None:
+        feature_states = self.context.get("feature_states", {})
+        feature_state = feature_states.get(instance.id)
+        if feature_state is None:
+            return
+
+        return FeatureStateSerializerSmall(instance=feature_state).data
```
matthewelwell: I think we can simplify this with our walrus-y friend again: 

```suggestion
        if (feature_states := self.context.get("feature_states")) and feature_state = feature_states.get(instance.id):
            return FeatureStateSerializerSmall(instance=feature_state).data
```

```suggestion
    def get_environment_feature_state(self, instance: Feature) -> FeatureState | None:
        feature_states = self.context.get("feature_states", {})
        feature_state = feature_states.get(instance.id)
        if feature_state is None:
            return

        return FeatureStateSerializerSmall(instance=feature_state).data
```

api/features/serializers.py
```diff
@@ -247,6 +251,14 @@ def validate(self, attrs):
 
         return attrs
 
+    def get_environment_feature_state(self, instance: Feature) -> FeatureState | None:
```
matthewelwell: We can type this by using the following: 

```python
    @swagger_serializer_method(
        serializer_or_field=FeatureStateSerializerSmall()
    )
```

There might be a way of indicating that it's nullable too, but I'll leave that to you to determine :)

api/features/views.py
```diff
@@ -154,8 +154,32 @@ def get_queryset(self):
         )
         queryset = queryset.order_by(sort)
 
+        if environment_id:
+            page = self.paginate_queryset(queryset)
+
+            if not getattr(self, "environment", None):
+                self.environment = Environment.objects.get(id=environment_id)
```
matthewelwell: I'm not sure under which scenarios here we do / don't have the environment already on the instance?

api/tests/unit/features/test_unit_features_views.py
```diff
@@ -2510,6 +2511,94 @@ def test_list_features_with_intersection_tag(
     assert response.data["results"][0]["tags"] == [tag1.id, tag2.id]
 
 
+def test_list_features_with_feature_state(
```
matthewelwell, line 2514: I think there's a few extra cases to test here (either in this test, or in a separate test): 

 1. Segment / Identity overrides are ignored
 2. The latest version is the one that is returned

matthewelwell (changes requested):

api/features/serializers.py
zachaysan: I'm not 100% sure I like the change but I've implemented it.

api/features/views.py
zachaysan: I believe I did this to avoid loading the environment twice with a separate PR which similarly needed the environment. But on second look I think the environment will always be loaded in this codepath first, so I'll update this to just set the environment without checking.

api/features/serializers.py
zachaysan: I don't think there is a nullable option. I checked [the source](https://drf-yasg.readthedocs.io/en/stable/_modules/drf_yasg/utils.html#swagger_serializer_method) and it's of no help.

zachaysan: Ok I've added the decorator 👍🏼

api/tests/unit/features/test_unit_features_views.py
zachaysan: I added them into the existing test

api/features/serializers.py
matthewelwell: > I don't think there is a nullable option. I checked [the source](https://drf-yasg.readthedocs.io/en/stable/_modules/drf_yasg/utils.html#swagger_serializer_method) and it's of no help.

Yep, you just need to add `allow_null=True` to the constructor. This at least then shows an indication in the docs that the field is nullable. I think strictly it's more targeted at whether it's nullable from an inbound perspective, but it's probably the best we can do here. 

Also, the type hint isn't quite correct here. We're returning `dict[str, typing.Any] | None`.

zachaysan: Ok neat! That's been added and so has the typing fix—nice catch there!

matthewelwell (approved):

PR #3590, opened 2024-03-08
chore: Convert identities views TestCase to normal test function part 2

zachaysan:

## Changes

I'm working through re-writing the TestCase classes into normal test functions. Rewrote the existing TestCase and replaced with fixtures instead of recreating everything. There are other PRs that are similar to this one.

This file was exceptionally large, so I decided to split it up into two different PRs. This PR is the second part of this one: https://github.com/Flagsmith/flagsmith/pull/3519


api/tests/unit/environments/identities/test_unit_identities_views.py
```diff
@@ -525,396 +525,439 @@ def test_identities_endpoint_returns_value_for_segment_if_identity_in_segment_an
...
+    environment.save()
+
+    # When
+    api_client.credentials(HTTP_X_ENVIRONMENT_KEY=environment.api_key)
+    response = api_client.post(
+        url, data=json.dumps(data), content_type="application/json"
+    )
+
+    # Then
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
```
gagantrivedi, line 857: We should probably add a check to make sure the new traits don't exists

api/tests/unit/environments/identities/test_unit_identities_views.py
```diff
@@ -525,396 +525,439 @@ def test_identities_endpoint_returns_value_for_segment_if_identity_in_segment_an
...
+    environment.allow_client_traits = False
+    environment.save()
+    # When
+    api_client.credentials(HTTP_X_ENVIRONMENT_KEY=environment_api_key.key)
+    response = api_client.post(
+        url, data=json.dumps(data), content_type="application/json"
+    )
+
+    # Then
+    assert response.status_code == status.HTTP_200_OK
```
gagantrivedi, line 885: Add a check to make sure traits exists

gagantrivedi (approved):
One minor comment

api/tests/unit/environments/identities/test_unit_identities_views.py
zachaysan: Good idea! Added.

api/tests/unit/environments/identities/test_unit_identities_views.py
zachaysan: Added this also.

PR #3593, opened 2024-03-11
fix(change-requests): prevent incorrect scheduled changes warning

matthewelwell:

## Changes

The FE shows the following warning based on the logic changed as part of this PR. I can't see any reason why it should use the start of the current hour instead of the current time so I have removed the `startOf('hour')` clause. 

image

Tested using the vercel previews.

### DIFF
--- a/frontend/web/components/ExistingChangeRequestAlert.tsx
+++ b/frontend/web/components/ExistingChangeRequestAlert.tsx
@@ -1,4 +1,4 @@
-import { FC } from 'react'
+import { FC, useRef } from 'react';
 import { useGetChangeRequestsQuery } from 'common/services/useChangeRequest'
 import WarningMessage from './WarningMessage'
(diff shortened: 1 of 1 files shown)


frontend/web/components/ExistingChangeRequestAlert.tsx
```diff
@@ -19,10 +19,11 @@ const ExistingChangeRequestAlert: FC<ExistingChangeRequestAlertType> = ({
     environmentId,
     feature_id: featureId,
   })
+  const date = useRef(moment().toISOString())
```
matthewelwell, line 22: Note that it was necessary (thanks to an assist from @kyle-ssg) to move this out to `useRef` to avoid an infinite recursion caused by the date being calculated on each re-render.

kyle-ssg (approved):

PR #3594, opened 2024-03-11
refactor: reduce DB lock time by changing trait update logic

### LINKED ISSUE #3595: Encountered high database lock time during trait updates

- [ ] Self Hosted with Docker
- [X] Self Hosted with Kubernetes
- [ ] SaaS at flagsmith.com

### Describe the bug

Setup: 

Flagsmith deployed in Kubernetes with the following configuration:
- 55 flagsmith-api replicas each with a limit of 1 CPU and 1GB RAM
- 3 pgbouncer replicas each with a limit of 0.5 CPU and 0.5GB RAM
- A PostgreSQL instance with 12 CPUs and 16GB RAM

The load pattern is around 500-1000 RPS for updating traits (`api/v1/traits/`) from the Android client (https://github.com/Flagsmith/flagsmith-kotlin-android-client). There are over 1.5 million devices in the system.

I observed a performance issue in this setup. Requests made to update traits from mobile clients resulted in extensive row locks in the database.

The problematic SQL statement is:

SELECT
  "environments_trait"."id",
  "environments_trait"."identity_id",
  "environments_trait"."trait_key",
  "environments_trait"."value_type",
  "environments_trait"."boolean_value",
  "environments_trait"."integer_value",
  "environments_trait"."string_value",
  "environments_trait"."float_value",
  "environments_trait"."created_date"
FROM
  "environments_trait"
WHERE
  ("environments_trait"."identity_id" = $1
    AND "environments_trait"."trait_key" = $2)
LIMIT
  $3 FOR
UPDATE

The statistics for this statement under the given load are:
- AVG execution time: 125.41 ms
- 99th percentile Query latency: Approaching 10 seconds
- Lock wait time: Extending up to 180 CPU-seconds

Given that an overwhelming majority (99.99%) of the execution time comprises waiting for rows to unlock, and with incoming updates from apps, the total query execution time increases to 30-45-60 seconds. As expected, the deployment experiences crashes. Essentially, the current architecture isn't prepared for this level of load.

The relevant code at: https://github.com/Flagsmith/flagsmith/blob/90e10cf656dba13c2a137e2c5630b325b8290776/api/environments/sdk/serializers.py#L49 uses the `update_or_create` method, which according to Django documentation, tries to fetch an object from the database based on provided kwargs and updates the fields in the defaults dictionary if a match is found.

In my view, executing a lock-update operation every time, regardless of whether the data has changed, isn't the most optimal solution. It leads to additional database work and potential locking issues(that I see), particularly under heavy load.

I've tried to fix problem - instead of lock-update every time, it reads the data, compares the existing and incoming data, and only issues an update when the data actually changes. This strategy drastically reduces unnecessary UPDATE operations and consequently, the chances of lock contention.

[Please see my PR for the detailed changes.
](https://github.com/Flagsmith/flagsmith/pull/3594)
With these changes, I observed:
- Reduced DB CPU utilization (max 8% instead of 30%)
- Lower RAM consumption (50% free instead of 25%)
- Reduction in write/update queries (0.07k/s as opposed to a peak of 1.5k/s)
- Most importantly, a significant decrease in lock time, I don't see it at all now.

Note that our use-case is quite rare, we quite often updating traits concurrently.

I'm interested in hearing your thoughts on this. Is it acceptable to make these types of changes?

### Steps To Reproduce

.

### Expected behavior

.

### Screenshots

Screenshot at Mar 01 20-52-56

cat5e-inv:

## Changes

Instead of lock-update every time, it reads the data, compares the existing and incoming data, and only issues an update when the data actually changes. This strategy drastically reduces unnecessary UPDATE operations and consequently, the chances of lock contention.

_Please describe._

Unit tests, manually

_Please describe._

### DIFF
--- a/api/environments/sdk/serializers.py
+++ b/api/environments/sdk/serializers.py
@@ -46,9 +46,22 @@ class SDKCreateUpdateTraitSerializer(serializers.ModelSerializer):
             else STRING,
         }
 
-        return Trait.objects.update_or_create(
-            identity=identity, trait_key=trait_key, defaults=defaults
-        )[0]
+        try:
+            trait_obj = Trait.objects.get(identity=identity, trait_key=trait_key)
+            updated = False
+            for field, value in defaults.items():
+                existing_value = getattr(trait_obj, field)
+                if value is not None and value != existing_value:
+                    setattr(trait_obj, field, value)
+                    updated = True
+            if updated:
+                trait_obj.save()
+        except Trait.DoesNotExist:
+            trait_obj = Trait.objects.create(
+                identity=identity, trait_key=trait_key, **defaults
+            )
+
+        return trait_obj
 
     def validate(self, attrs):
         request = self.context["request"]


api/environments/sdk/serializers.py
```diff
@@ -46,9 +46,22 @@ def create(self, validated_data):
...
+                existing_value = getattr(trait_obj, field)
+                if value is not None and value != existing_value:
+                    setattr(trait_obj, field, value)
+                    updated = True
+            if updated:
+                trait_obj.save()
+        except Trait.DoesNotExist:
+            trait_obj = Trait.objects.create(
+                identity=identity, trait_key=trait_key, **defaults
+            )
```
matthewelwell, line 62: I wonder if this would be better defined on the `TraitManager` class itself? Maybe `update_if_changed_or_create` or something? That way the logic in this serializer stays cleaner, and we can much more easily unit test this logic.

matthewelwell (changes requested):
Thanks for this @cat5e-inv, I think this is a really nice improvement. I've added a comment which I think would be really nice to implement, but really it's just a small(ish) refactor, rather than a change.

matthewelwell:
@cat5e-inv I also updated the title of the PR to fit our naming conventions.

api/environments/sdk/serializers.py
cat5e-inv: Thank you for feedback and suggestions. 

However, currently, my Python skill level might not be sufficient to implement the proposed changes accurately and efficiently. I believe that these modifications would be better suited for someone with deeper knowledge and expertise in it.

I'd be utterly grateful if someone from the team could take ownership of making these suggested improvements, ensuring the best result for the codebase and the project.

matthewelwell: @cat5e-inv I've actually been looking into this further and I'm not wholly convinced that this change is a good idea. As you can see from the django code [here](https://github.com/django/django/blob/3.2.25/django/db/models/query.py#L596-L614), the lock is implemented to ensure that we don't have conflicting writes in the database. If we remove that lock, we could end up with inconsistent behaviour. For example: 

Request 1 comes in with an update to trait `trait_key=foo` to set `trait_value=bar`, retrieves the trait from the database, and starts trying to determine whether it needs to update or not
Request 2 comes in with an update to the same trait, to set `trait_value=baz`, retrieves the trait from the database and (quicker than request 1) updates the `trait_value` successfully. 
Request 1 finishes determining whether it needs to update the trait and sets the `trait_value` to `bar`. 

Since we can't guarantee the order of execution in this scenario (because threading), we would end up with the final result being that the trait takes the value of request 1, when it should have taken the value of request 2. 

I can certainly understand where you're coming from here, and the actual scenario where an issue like this might occur is pretty slim, but I am certainly reluctant to implement this change based on this. We could perhaps implement an environment variable with a warning in the docs that it might result in inconsistent writes, but will likely reduce DB lock time.

matthewelwell: Do you have any quantifiable data around these locks causing a problem for you?

arronax: @matthewelwell @cat5e-inv the screenshot in https://github.com/Flagsmith/flagsmith/issues/3595 shows the Cloud SQL PG instance metrics, and how locks and update volume drops.

The locking issue in particular is probably not going to be a big deal for most users, because one needs to concurrently update the same traits of the same identity. Even in our case, I must admit, that's a deficiency of our approach, not Flagsmith's.

If we concentrate just on the locking, however, and not on the volume of writes, what do you think about the use of PG's conflict handling? If each trait update writes anyway, an `INSERT...ON CONFLICT UPDATE` at least removes the need to run a `SELECT...FOR UPDATE`. The only problem is that it doesn't work with standard `update_or_create()`: https://django-postgres-extra.readthedocs.io/en/master/conflict_handling.html. This approach should also be slightly more performant due to at least one less roundtrip. This code would not be database-agnostic, however, and I'm not sure how much of a problem that is.

The originally proposed change, however, reduces the sheer volume of updates for the situation where trait values don't change that often. That seems to be a good thing. However, you're right that it breaks the ordering of updates, and allows the oldest transaction to "win". I tried to think of a way to avoid the current pessimistic locking approach but failed to come to something concrete that is also not overly complicated. I'll write out my thoughts anyway, perhaps someone will get inspired.

One way out could be to use optimistic locking, but that would necessitate using some way of tracking not only the record version, but also transaction itself against its peers. That should be possible with PG's xmin, but at the very least will be quite complicated. Another option is to run this particular transaction under SERIALIZABLE isolation in PG, which will reduce locking substantially, but would also cause concurrent transactions to fail on commit due to serialization failures. First committer wins with PG SSI (https://wiki.postgresql.org/wiki/SSI), and I think that would still require the same tracking to decide whether to re-try or not. If t1 commits when t2 and t3 have just finished their reads, both t2 and t3 will fail, and the earlier transaction wins. If t2 and t3 retry simultaneously, it's not possible to say whether the end result will be that of t2 or t3 unless they can order themselves.

matthewelwell: > This code would not be database-agnostic, however, and I'm not sure how much of a problem that is.

Yes, this is unfortunately an issue for us as we support multiple database engines for our Enterprise offering. We could add some conditional logic to separate out based on database engine, but it gets quite complicated and I'd rather avoid it if possible. 

Could you compare the traits on your side before sending the update using the `getTrait` or `getTraits` method in the Kotlin client?

matthewelwell: @arronax @cat5e-inv do you have anything further to add here?

cat5e-inv: @matthewelwell 
Apologies for the delayed response.

Due to the current structure of our application, implementing this feature on the client side is rather complex, especially considering it needs to be integrated across multiple types of clients.

By the way, the Android-Kotlin library's API implementation is quite minimal. It only allows updating traits individually, which is quite server-intensive, as evidenced by this issue. [We've developed a bulk update](https://github.com/Flagsmith/flagsmith-kotlin-android-client/pull/45) feature according to the API within the Kotlin library and have started to incorporate it. This should reduce the amount of requests to update traits and eliminate the need for the "select for update" logic(as I see, bulk update has another implemenation). However, every request will still update rows in the database, even if the data remains the same. We will see.

matthewelwell: @cat5e-inv thanks for the contribution to the kotlin client and apologies for the delay in getting back to you. 

As mentioned on the PR, I have merged it (too hastily), but I cannot release a new version as the implementation is not compatible with our SaaS platform. I have a PR [here](https://github.com/Flagsmith/flagsmith-kotlin-android-client/pull/47) to revert the change, but if you're able to submit a small update to use the endpoint I mentioned on the comment on the PR then that would likely be the best next step.

matthewelwell:
@cat5e-inv , based on the discussion thread [here](https://github.com/Flagsmith/flagsmith/pull/3594#discussion_r1519903464) and the (now released in v1.6.0) work in the kotlin client to expose the bulk traits functionality (See PRs [here](https://github.com/Flagsmith/flagsmith-kotlin-android-client/pull/45) and [here](https://github.com/Flagsmith/flagsmith-kotlin-android-client/pull/48)) I am going to close this PR as there are significant concerns about the behaviour that would result from this change.

If you strongly disagree with this, please let me know and we can discuss whether this change could be included but hidden behind some sort of switch perhaps.

PR #3605, opened 2024-03-13
feat(tags): prevent system tag modifications

matthewelwell:

## Changes

Prevent system generated tags from being deleted or modified via the API. 

Added new tests against the API.


api/tests/unit/projects/tags/test_unit_projects_tags_views.py
```diff
@@ -68,3 +70,51 @@ def test_get_tag_by__uuid_returns_200_for_user_with_view_project_permission(
...
+def test_cannot_delete_a_system_tag(
+    staff_client: APIClient,
+    with_project_permissions: WithProjectPermissionsCallable,
+    project: Project,
+    system_tag: Tag,
+) -> None:
+    # Given
+    url = reverse("api-v1:projects:tags-detail", args=(project.id, system_tag.id))
+
+    with_project_permissions(admin=True)
```
zachaysan, line 84: I take it you are setting the admin level here to go at the lowest level of permissions over using the `admin_client` instead. Do I have that right?

zachaysan (approved):
Looks good. Nice and simple code 👍🏼

api/tests/unit/projects/tags/test_unit_projects_tags_views.py
matthewelwell: Exactly, yeah - the `admin_client` is an organisation admin so I wanted to make it possible for a client to only be a project admin.

PR #3606, opened 2024-03-13
feat: stale flags (FE)

### LINKED ISSUE #184: Show "stale" flags
This will help people identity flags that have not changed for a while and are good candidates for removal.

Add a button (somewhere! Not sure where - maybe by the current tags or by the search filter top right) to filter flags that haven't had a state change in N days (suggest N=30 initially but can be changed by the user).

The following events should be considered a state change:

- Creation of flag
- Modifying core flag state (boolean, config, MV)
- Adding/removing/changing segment override
- Adding/removing Identity override

## UI

Add a "Last Modified" option here: 

image

### LINKED ISSUE #576: Protected flag tags feature is not findable
There is no way to find out about the tags that protect a flag (Protected, Permanent, Do Not Delete, Read Only). The only way to use this feature right now would be to have someone show it to you. 

There should be some explanation that would lead someone to add one of those tags. Or they could be automatically created (but this could add clutter)

My suggestion is that it is a toggle like Archived is in the settings. Something like that, doesn't really matter as long as it informs users of the feature.

kyle-ssg:

## Changes

- Allows defining whether a tag indicates a permanent feature
- Prevents deletion / adds icon for system generated tags
- Adds tooltip for STALE tags
- Standardises rendering tag content
- Resets create/edit tag modal on close (before it would stay on the create tag state)
- Fixes a 1px tooltip arrow gap and incorrect background

 

https://github.com/Flagsmith/flagsmith/assets/8608314/7a476e54-1413-4da2-9f1d-83bb818643c0

_Please describe._


frontend/web/components/tags/TagContent.tsx
```diff
@@ -0,0 +1,87 @@
...
+const getTooltip = (tag: TTag | undefined) => {
+  if (!tag) {
+    return null
+  }
+  const truncated = Format.truncateText(tag.label, 12)
+  const isTruncated = truncated !== tag.label ? tag.label : null
+  let tooltip = null
+  switch (tag.type) {
+    case 'STALE': {
+      tooltip = `A feature is marked as stale if no changes have been made to it in any environment within ${Project.stale_flags_limit_days} days. This is automatically applied and will be re-evaluated every 3 days if you remove this tag unless you apply a permanent tag to the feature.`
```
matthewelwell:
```suggestion
      tooltip = `A feature is marked as stale if no changes have been made to it in any environment within ${Project.stale_flags_limit_days} days. This is automatically applied and will be re-evaluated if you remove this tag unless you apply a permanent tag to the feature.`
```

This was technically wrong as it's currently every 3 hours, but I think in general I'd like the freedom to tweak this in the API without having to update here as well. I don't think it's important to include the time period here.

frontend/api/index.js
```diff
@@ -108,6 +108,7 @@ app.get('/config/project-overrides', (req, res) => {
     { name: 'amplitude', value: process.env.AMPLITUDE_API_KEY },
     { name: 'delighted', value: process.env.DELIGHTED_API_KEY },
     { name: 'capterraKey', value: process.env.CAPTERRA_API_KEY },
+    { name: 'stale_flags_limit_days', value: process.env.STALE_FLAG_LIMIT_DAYS || "30" },
```
matthewelwell: This is an attribute on the project, although I see that I haven't exposed it in the serializer. I will do that now, but you should update this PR to add a configurable attribute in the project settings.

matthewelwell (commented):
We should update this tooltip as well based on the addition of the is_permanent attribute: 

image

frontend/api/index.js
kyle-ssg: Gotcha, I missinterpreted what Project. stale_flags_limit_days, will remove

kyle-ssg:
> We should update this tooltip as well based on the addition of the is_permanent attribute:
> 
> image

I think you already did here https://github.com/Flagsmith/flagsmith/pull/3606/files#diff-87ffbe66e88677f5a26bc18b4f651b84a9a9264f03a350a11caa74f6df5077e2L23

matthewelwell:
> > We should update this tooltip as well based on the addition of the is_permanent attribute:
> > image
> 
> I think you already did here https://github.com/Flagsmith/flagsmith/pull/3606/files#diff-87ffbe66e88677f5a26bc18b4f651b84a9a9264f03a350a11caa74f6df5077e2L23

I updated the logic, but I'm talking about the actual content of the Tooltip itself ([this line of code](https://github.com/Flagsmith/flagsmith/blob/main/frontend/web/components/FeatureAction.tsx#L172)). It's still referencing the old behaviour.

kyle-ssg:
Added 
image

novakzaballa (approved):
LGTM

kyle-ssg:
@matthewelwell Just checked and project settings is also persisting the newly added stale days project config

matthewelwell:
@kyle-ssg one thing I've noticed here is that the FE seems to let you edit the project setting regardless of the plan that the organisation has. The API restricts this to enterprise subscriptions only - can we update this?

kyle-ssg:
@matthewelwell ah thanks, that has been updated now.

matthewelwell:
@kyle-ssg we might want to consider some extra documentation around this somewhere (not exactly sure where yet), because this feature will only work with feature versioning enabled. Perhaps an extra line on the tooltips for the setting in projects settings and the one on the tag itself?

kyle-ssg:
Hmmm I guess this will be a little tricky since versioning is enabled per environment.

I'll note it in the tooltip and also start managing this UI with the versioning feature flag

matthewelwell:
@kyle-ssg looks like this has broken: 

image

kyle-ssg:
@matthewelwell sorted

matthewelwell (approved):

PR #3610, opened 2024-03-13
chore: Versioned change requests

kyle-ssg:

## Changes

- If versioning is enabled, change requests will now be created with associated versions


matthewelwell:
@kyle-ssg perhaps I didn't communicate this very well, but I was under the impression that this work would also bring segment overrides into the change request workflow when versioning was enabled?

matthewelwell:
> @kyle-ssg perhaps I didn't communicate this very well, but I was under the impression that this work would also bring segment overrides into the change request workflow when versioning was enabled?

As discussed, this will be done in a separate PR.

matthewelwell (dismissed):

frontend/common/stores/feature-list-store.js
```diff
@@ -414,7 +414,25 @@ const controller = {
   ) => {
     store.saving()
     API.trackEvent(Constants.events.EDIT_FEATURE)
-
+    const env = ProjectStore.getEnvironment(environmentId)
+    let environment_feature_versions = null
```
matthewelwell: The API seemingly doesn't like explicitly setting `null` here so this is breaking the existing change request workflow. I can update the API, but the quickest fix would probably be the following (or not sending the key at all). 

```suggestion
    let environment_feature_versions = []
```

matthewelwell (changes requested):

frontend/common/stores/feature-list-store.js
kyle-ssg: Yep ok makes sense

matthewelwell (approved):

PR #3618, opened 2024-03-14
chore: Import export docs

zachaysan:

## Changes

Documentation for feature import export came with a reorganization of the existing import export code.

Also, a README addition comes with a quality of life improvement for running this locally. 

Locally checked documentation looked accurate.


docs/docs/system-administration/importing-and-exporting/features.md
```diff
@@ -0,0 +1,51 @@
+---
+title: Features
+sidebar_position: 110
+---
+
+The import and export of feature data associated with a given environment is possible on Flagsmith. The feature data
+that's export includes multivariate features, but does not include other data that's associated with tags, owners, group
```
matthewelwell:
```suggestion
that's exported includes multivariate features, but does not include other data that's associated with tags, owners, group
```

docs/docs/system-administration/importing-and-exporting/features.md
```diff
@@ -0,0 +1,51 @@
...
+owners, etc.
+
+## Exporting
+
+On the Export tab of the project settings page it is possible to export the project's features using the default values
```
matthewelwell: We should be careful about terminology here. These are the 'environment' values rather than the 'default' values.

docs/docs/system-administration/importing-and-exporting/bulk-data.md
```diff
@@ -1,14 +1,8 @@
 ---
-title: Importing and Exporting
+title: Bulk Data
```
matthewelwell: Maybe we could use 'Data Migration' here?

dabeeeenster:
I think we need to add context to the bulk export and flag export as to when you might use either of these features. I'm not clear what the point of the feature export is as much as bulk? Is it for folk with network-isolated environments?

dabeeeenster:
can we use `info` or `tip` in place of `warn` - the warn is scary...

docs/docs/system-administration/importing-and-exporting/bulk-data.md
zachaysan: Good idea, I've updated it

zachaysan:
> can we use `info` or `tip` in place of `warn` - the warn is scary...

I've changed one of the cautionary flags to `tip` but the other one I think should stay as a caution because it is scary!

zachaysan:
> I think we need to add context to the bulk export and flag export as to when you might use either of these features. I'm not clear what the point of the feature export is as much as bulk? Is it for folk with network-isolated environments?

Ok I've added a bit more clarity in the introductory paragraphs of the two documents.

dabeeeenster (approved):

PR #3624, opened 2024-03-15
feat: Enterprise licensing

### LINKED ISSUE #3597: Enterprise Licence Management
The UX around installing Flagsmith on-prem can be improved with some additional license management. Suggested approach:

- Flagsmith generates an encrypted file with its _private_ key that includes meta data such as customer name, department name, number of seats, expiry date and possibly other data
- This file is then sent to the customer
- When starting the EE version of Flagsmith, this file can be passed in with different methods (ENV var, attached volume etc) so that the file can be read
- The file is decrypted with the Flagsmith _public_ key and the meta data used to help configure the service

We should also display the org name and department name in the header of the dashboard.

matthewelwell:
## Changes

This PR sets up licensing work to enable us to digitally sign licensing agreements and enable them to be evaluated towards the enterprise's limits.

A few management commands have been added to allow us to create signatures and to digitally sign licensing agreements.

To create a private key (which has been done for production already, stored in Bitwarden) run:

`poetry run python manage.py create_licensing_private_key`

To create a public key (which has been done for production already, added to the settings file) run:

`poetry run python manage.py create_licensing_public_key`

To create and sign a licence agreement run:

```
poetry run python manage.py create_signed_licence \
--organisation-name "Some Org" \
--plan-id "SomePlanID" \
--num-seats 20 \ 
--num-projects 10 \
--num-api-calls 2000000
```

Which will output the JSON file for the license and an cryptographic blob which is the licence signature.

To add a licence to a Flagsmith organisation upload via `PUT` the two files, one the license doc and the other the licence signature to the following URL path:

`/api/v1/organisations/:organisation_id/licence`

New tests were added as well as manually verifying that the management commands could work.


api/organisations/management/commands/create_licensing_private_key.py
```diff
@@ -0,0 +1,10 @@
+from typing import Any
+
+from django.core.management import BaseCommand
+
+from organisations.subscriptions.licensing.helpers import create_private_key
+
+
+class Command(BaseCommand):
+    def handle(self, *args: Any, **options: Any) -> None:
+        print(create_private_key())
```
matthewelwell:
```suggestion
        self.stdout.write(create_private_key())
```

We should use `self.stdout` here instead of print as per advice in django docs (see [here](https://docs.djangoproject.com/en/5.1/howto/custom-management-commands/))

matthewelwell: (this comment applies to all usages of print below)

api/organisations/subscriptions/licensing/__init__.py
```diff
@@ -0,0 +1 @@
+# TODO: split this into a private package?
```
matthewelwell: Ah yes, good point, we probably should do this... 🤔

api/organisations/subscriptions/licensing/helpers.py
```diff
@@ -0,0 +1,91 @@
...
+logger: logging.Logger = logging.getLogger(name=__name__)
+
+
+def sign_licence(licence: str) -> str:
+    message = licence.encode("utf-8")
+
+    # Load the private key from PEM
+    private_key = serialization.load_pem_private_key(
+        settings.SUBSCRIPTION_LICENCE_PRIVATE_KEY.encode("utf-8"), password=None
+    )
```
matthewelwell: Just for ergonomics, perhaps we should throw an exception (and catch it in the management command) if `SUBSCRIPTION_LICENCE_PRIVATE_KEY` is not set.

api/organisations/subscriptions/licensing/migrations/0002_alter_organisationlicence_organisation.py
matthewelwell, line 1: I think we can remove this migration if we just update `0001` to have the correct field definition.

api/organisations/subscriptions/licensing/models.py
```diff
@@ -0,0 +1,17 @@
...
+    updated_at = models.DateTimeField(auto_now=True)
+
+    organisation = models.OneToOneField(
+        "organisations.Organisation", related_name="licence", on_delete=models.CASCADE
+    )
+
+    content = models.TextField(blank=True)
+
+    def get_licence_information(self) -> LicenceInformation:
+        return LicenceInformation.parse_raw(self.content)
```
matthewelwell: I guess this implies that the licence information is still stored in the database in plain text. That's probably fine (and better than the alternative of having to decrypt every time we need access). 

I guess an (overkill perhaps) alternative would be adding some kind of time-based cache to hold the decrypted licence information so that we're only decrypting it once every 24h for example?

api/tests/unit/features/versioning/test_unit_versioning_views.py
```diff
@@ -1664,6 +1665,20 @@ def test_list_versions_returns_all_versions_for_enterprise_plan(
...
+            "plan_id": "Enterprise",
+            "num_seats": 20,
+            "num_projects": 3,
+            "num_api_calls": 3_000_000,
+        }
+
+        OrganisationLicence.objects.create(
+            organisation=environment_v2_versioning.project.organisation,
+            content=json.dumps(licence_content),
+        )
```
matthewelwell: This makes me think that we should split this case out from this parametrised test.

zachaysan: Sounds good.

api/organisations/subscriptions/licensing/models.py
zachaysan: We discussed this over standup and we've determined to leave it unencrypted for now.

api/organisations/subscriptions/licensing/migrations/0002_alter_organisationlicence_organisation.py
zachaysan: Ok this is done.

api/organisations/subscriptions/licensing/helpers.py
zachaysan: Sounds good.

api/organisations/management/commands/create_licensing_private_key.py
zachaysan: Interesting. Thanks!

api/organisations/subscriptions/licensing/__init__.py
zachaysan: That was actually one of your TODOs, not mine. I'm not convinced it is worth the trouble to splitting this into a private package.

matthewelwell: Well, I think it also makes sense for another reason since this code also isn't relevant for SaaS so, if it was in it's own package, we could bundle it into the on premise version only. 

I appreciate that it's a pain, but I do think that we should do it, but maybe we can use it as the first test of the new private hosted pypi server work that we're doing?

api/organisations/models.py
```diff
@@ -415,16 +415,26 @@ def _get_subscription_metadata_for_chargebee(self) -> ChargebeeObjMetadata:
...
+                feature_history_visibility_days=None,
+            )
+        elif is_enterprise():  # pragma: no cover
+            return BaseSubscriptionMetadata(
+                seats=self.max_seats,
+                api_calls=self.max_api_calls,
+                projects=None,
+                audit_log_visibility_days=None,
+                feature_history_visibility_days=None,
+            )
```
matthewelwell, line 438: As discussed previously, I think we'll want to remove this eventually at least. Let's add a comment explaining the reasoning.

api/tests/unit/features/versioning/test_unit_versioning_views.py
```diff
@@ -1639,8 +1639,7 @@ def test_list_versions_always_returns_current_version_even_if_outside_limit(
 
 
 @pytest.mark.freeze_time(now - timedelta(days=DEFAULT_VERSION_LIMIT_DAYS + 1))
-@pytest.mark.parametrize("is_saas", (True, False))
-def test_list_versions_returns_all_versions_for_enterprise_plan(
```
matthewelwell, line 1643: Aren't we losing a test here? I understood that we were going to split the test into 2 individual tests? 🤔

zachaysan: It was moved into the licensing repo over [here](https://github.com/Flagsmith/licensing/blob/main/tests/integration/test_integration_versioning_views.py#L29) since it needed the `OrganisationLicence` model.

api/organisations/models.py
zachaysan: Ok no problem I've added a comment.

matthewelwell (commented):
@zachaysan approving via comment since it's my PR. Can you approve (but let's hold off on merging until next week)?

zachaysan (approved):

PR #3648, opened 2024-03-21
feat: Add domain to Hubspot company

### LINKED ISSUE #3626: Hubspot: send domain name with Company to Hubspot
When we create a company in Hubspot, we should also send the domain name. 

image

Looks like the field is a built in Hubspot field so we should be able to provide it in the JSON we send across when we create the company. The domain name should be taken from the email address of the user creating / joining the organisation. 

If the user has a generic email address (gmail.com, hotmail.com, etc.) then we should try to avoid setting it where possible.

zachaysan:

## Changes

Fixes https://github.com/Flagsmith/flagsmith/issues/3626

Simple change to introduce adding domain to the Hubspot company model.

Since this change was purely against their API, I elected to just test it manually which was successful.


api/app/settings/common.py
```diff
@@ -1033,7 +1033,33 @@
 ENABLE_HUBSPOT_LEAD_TRACKING = env.bool("ENABLE_HUBSPOT_LEAD_TRACKING", False)
 HUBSPOT_IGNORE_DOMAINS = env.list("HUBSPOT_IGNORE_DOMAINS", [])
 HUBSPOT_IGNORE_DOMAINS_REGEX = env("HUBSPOT_IGNORE_DOMAINS_REGEX", "")
-
+HUBSPOT_IGNORE_ORGANISATION_DOMAINS = env.list(
+    "HUBSPOT_IGNORE_ORGANISATION_DOMAINS",
+    [
+        "gmail.com",
+        "hotmail.com",
```
kyle-ssg, line 1049: (being nosey) hotmail is duplicated, this is a list I use to filter out big-signups if it helps https://github.com/Kikobeats/free-email-domains/blob/master/domains.json#L1

zachaysan: Is this huge list being used on something at Flagsmith or for some other project? I'd just be wary of adding so many domains which are largely dormant. 

Edit: And thanks for spotting the duplication, I've removed it and sorted the list for easier visual verification and lookup.

api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -54,18 +54,24 @@ def create_lead(self, user: FFAdminUser, organisation: Organisation) -> None:
...
         """
         if getattr(organisation, "hubspot_organisation", None):
             return organisation.hubspot_organisation.hubspot_id
 
-        response = self.client.create_company(name=organisation.name)
+        if user.email_domain in settings.HUBSPOT_IGNORE_ORGANISATION_DOMAINS:
+            domain = None
+        else:
+            domain = user.email_domain
+        response = self.client.create_company(name=organisation.name, domain=domain)
```
matthewelwell: We should update the unit tests for this logic.

matthewelwell (changes requested):

api/integrations/lead_tracking/hubspot/lead_tracker.py
zachaysan: Ok altered one existing test to capture when it is set and created a new test for when the domain name is in the filter list.

api/app/settings/common.py
kyle-ssg: Anther project

kyle-ssg: Another project

matthewelwell (approved):

PR #3661, opened 2024-03-25
fix: No pagination when querying `environments_v2`

khvn26:

## Changes

This adds pagination to `environments_v2` querying code to allow returning more than 1MB of identity overrides.

Added a unit test for the case when DDB query gets more than one page of results.

### DIFF
--- a/api/tests/unit/environments/dynamodb/wrappers/test_unit_dynamodb_environment_v2_wrapper.py
+++ b/api/tests/unit/environments/dynamodb/wrappers/test_unit_dynamodb_environment_v2_wrapper.py
@@ -2,11 +2,15 @@ import uuid
 
 from mypy_boto3_dynamodb.service_resource import Table
 from pytest_django.fixtures import SettingsWrapper
+from pytest_mock import MockerFixture
 
 from environments.dynamodb import DynamoEnvironmentV2Wrapper
 from environments.dynamodb.types import (
     IdentityOverridesV2Changeset,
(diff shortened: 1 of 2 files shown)


api/tests/unit/environments/dynamodb/wrappers/test_unit_dynamodb_environment_v2_wrapper.py
```diff
@@ -53,6 +59,45 @@ def test_environment_v2_wrapper__get_identity_overrides_by_environment_id__retur
...
+        [
+            mocker.call(
+                KeyConditionExpression=mocker.ANY,
+            ),
+            mocker.call(
+                KeyConditionExpression=mocker.ANY,
+                ExclusiveStartKey=last_evaluated_key,
+            ),
+        ]
+    )
```
zachaysan, line 98: So just to make sure I'm understanding this correctly: All the pages are returned in a single call. The response to the client could potentially span multiple documents that could be quite long?

zachaysan (approved):
I think I understand the code and it looks good. One question below.

api/tests/unit/environments/dynamodb/wrappers/test_unit_dynamodb_environment_v2_wrapper.py
khvn26: Yep. We have a customer with thousands of identity overrides for a single feature which prompted us to come up with this fix

PR #3663, opened 2024-03-25
ci: fix uffizzi

dabeeeenster:
I _think_ the `uffizzi-preview.yml` file runs from `main` meaning we need to merge this to test it 🤮

The zip file content is now nested due to the changes in the previous step, so this change just flattens the zip contents which might/should fix things

### DIFF
--- a/.github/workflows/uffizzi-preview.yml
+++ b/.github/workflows/uffizzi-preview.yml
@@ -20,7 +20,7 @@ jobs:
     steps:
       - name: 'Download artifacts'
         # Fetch output (zip archive) from the workflow run that triggered this workflow.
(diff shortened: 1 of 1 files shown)


zachaysan (approved):
Approved, but be aware that `unzip -j` will cause user prompts under some circumstances. Not sure how it's going to run as it is.

dabeeeenster:
Ah well! I tested on MacOS and it didnt ask so...maybe just merge and see?

zachaysan:
> Ah well! I tested on MacOS and it didnt ask so...maybe just merge and see?

Sure, haha, what's the downside? :)

PR #3665, opened 2024-03-25
style: Roles

kyle-ssg:

### Roles Table

**Before:**

image

**After:**

image

### Tabs

**Before**
image

**After**
image

## Members

**Before** 
image

**After**
image

## Permissions

**Before** 
image
image

**After**
image
image

https://github.com/Flagsmith/flagsmith/assets/8608314/58af3cb9-8863-414a-b20f-cda117620169


novakzaballa (approved):

matthewelwell:
@kyle-ssg this seems to be a regression from the current production code. 

https://github.com/Flagsmith/flagsmith/assets/14089968/4a1ddf5f-dd7f-4147-9cf5-e37fce270c46

I wonder if there's also a way that we can keep the list of assigned permissions visible in the list view, I really liked that feature.

matthewelwell:
image

It seems like the FE is not correctly rendering the response from the API as it's not showing that a permission exists already for this environment / role combination. I think this is what was causing the 500 error because the FE is trying to send a new role / permission assignment rather than update the existing one.

matthewelwell (changes requested):
As per comments with screenshot / screen recording.

kyle-ssg:
> @kyle-ssg this seems to be a regression from the current production code.
> 
>  Screen.Recording.2024-04-03.at.11.19.12.mov 
> I wonder if there's also a way that we can keep the list of assigned permissions visible in the list view, I really liked that feature.



> @kyle-ssg this seems to be a regression from the current production code.
> 
>  Screen.Recording.2024-04-03.at.11.19.12.mov 
> I wonder if there's also a way that we can keep the list of assigned permissions visible in the list view, I really liked that feature.

This is fixed, I'll take a look at why the permissions aren't showing on the list items.

kyle-ssg:
Resolved the above

matthewelwell (approved):
This looks good for the most part. There are still 2 minor niggles but I think perhaps they can be sorted in a separate PR because, on the whole, this is a lot better than what exists currently. 

1. When you make a change to a project / environment permission set, then close the panel without saving it should either remember the changes, or remove the 'unsaved' badge. At the moment, it's confusing that it forgets the changes, but keeps the unsaved badge. See video. 

https://github.com/Flagsmith/flagsmith/assets/14089968/edce86bb-0b1a-416a-ba25-44d79fbfc8ac

2. When a permission includes only the admin role, the badge shows '+1' rather than the label. See screenshot. 

image

PR #3666, opened 2024-03-25
fix: Add Flagsmith signature header when testing webhook.

### LINKED ISSUE #2786: Include `x-flagsmith-signature` header when webhook is triggered by test webhook widget
Include `x-flagsmith-signature` header when webhook is triggered by test webhook widget

**To Reproduce**

1. Click on Create webhook and url and secret 
2. Click on test webhook
3. Observe: no  `x-flagsmith-signature` header is sent 

It should include ` x-flagsmith-signature` header

shubham-padia:
Fixes #2786.

No need to add info to the docs

## Changes

We are trying to create the same signature as the webhook in the python code. This commit assumes that the python code will use the same approach to create signature for long term.

We are relying on notes here to make sure that both frontend and backend are implementing the same signature function, which is not the perfect approach. An alternative would be a backend endpoint to test the webhook so that the implementation always remains the same, but not sure if its overkill at this point or not. Note to maintainers: Please let me know what you think about both the approaches and what makes sense in this case since you might know best how the users might use this feature and how often the sign function might change. 

I'm also modifying the string on the frontend to be the exact same as the json.dumps output on the python side, so that the sign value is the same on both side. 

- Created a webhook on the project settings, set the secret to `test123` and a dead example URL and noted down the header value: `X-Flagsmith-Signature: 1907ef6a4c7a3e0010504757728d8dbbf3980247e3e2c5835757b80e9fd1f085`
- 
Screenshot 2024-03-26 at 12 01 27 AM

- You can check that the replit [here](https://replit.com/@ShubhamPadia/test-hmac?v=1#main.py) will give the same signature. I also generated it in `test_unit_webhooks.py` by modifying the code to use the same example as frontend just to make sure there is no different when using the replit, but didn't commit the code since the current test is sufficient enough. 
Screenshot 2024-03-26 at 12 07 32 AM

I have not tested this on a real webhook, just a dead link, let me know if I need to test it on a real webhook, would be nice if there's an existing test endpoint for that. 
I have not added any tests on the frontend side, since I could only find e2e tests, please let me know if tests needed to be added on frontend for this change. 

NOTE: I found a lot of unrelated errors to my changes when running `npm run lint:fix`, I commited the changes related to my PR and ignored the rest, let me know what was the ideal approach here, same with pre-commit

### DIFF
--- a/api/core/signing.py
+++ b/api/core/signing.py
@@ -3,6 +3,8 @@ import hmac
 
 
 def sign_payload(payload: str, key: str):
+    # signPayload of frontend/web/components/TestWebHook on the frontend replicates this
+    # exact function, change the function there if this changes.
     return hmac.new(
         key=key.encode(), msg=payload.encode(), digestmod=hashlib.sha256
     ).hexdigest()
--- a/api/edge_api/identities/edge_request_forwarder.py
+++ b/api/edge_api/identities/edge_request_forwarder.py
@@ -81,9 +81,6 @@ def _get_headers(request_method: str, headers: dict, payload: str = "") -> dict:
     # ref: https://groups.google.com/g/django-developers/c/xjYVJN-RguA/m/G9krDqawchQJ
     if request_method == "GET":
         headers.pop("Content-Length", None)
-
-    # signPayload of frontend/web/components/TestWebHook on the frontend replicates this
-    # exact function, change the function there if this changes.
     signature = sign_payload(payload, settings.EDGE_REQUEST_SIGNING_KEY)
     headers[FLAGSMITH_SIGNATURE_HEADER] = signature
     return headers
--- a/api/edge_api/identities/edge_request_forwarder.py
+++ b/api/edge_api/identities/edge_request_forwarder.py
@@ -81,6 +81,9 @@ def _get_headers(request_method: str, headers: dict, payload: str = "") -> dict:
     # ref: https://groups.google.com/g/django-developers/c/xjYVJN-RguA/m/G9krDqawchQJ
     if request_method == "GET":
         headers.pop("Content-Length", None)
+
+    # signPayload of frontend/web/components/TestWebHook on the frontend replicates this
+    # exact function, change the function there if this changes.
     signature = sign_payload(payload, settings.EDGE_REQUEST_SIGNING_KEY)
     headers[FLAGSMITH_SIGNATURE_HEADER] = signature
     return headers
--- a/api/webhooks/webhooks.py
+++ b/api/webhooks/webhooks.py
@@ -153,8 +153,6 @@ def _call_webhook(
     headers = {"content-type": "application/json"}
     json_data = json.dumps(data, sort_keys=True, cls=DjangoJSONEncoder)
     if webhook.secret:
-        # signPayload of frontend/web/components/TestWebHook on the frontend replicates this
-        # exact function, change the function there if this changes.
         signature = sign_payload(json_data, key=webhook.secret)
         headers.update({FLAGSMITH_SIGNATURE_HEADER: signature})
 
--- a/api/webhooks/webhooks.py
+++ b/api/webhooks/webhooks.py
@@ -153,6 +153,8 @@ def _call_webhook(
     headers = {"content-type": "application/json"}
     json_data = json.dumps(data, sort_keys=True, cls=DjangoJSONEncoder)
     if webhook.secret:
+        # signPayload of frontend/web/components/TestWebHook on the frontend replicates this
+        # exact function, change the function there if this changes.
         signature = sign_payload(json_data, key=webhook.secret)
         headers.update({FLAGSMITH_SIGNATURE_HEADER: signature})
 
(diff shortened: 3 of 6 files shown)


frontend/web/components/TestWebhook.tsx
```diff
@@ -8,18 +8,72 @@ import Button from './base/forms/Button'
...
+        })
+    })
+}
+
+const TestWebhook: FC<TestWebhookType> = ({ json, secret, webhook }) => {
   const [error, setError] = useState<string | null>(null)
   const [loading, setLoading] = useState(false)
   const [success, setSuccess] = useState(false)
+  const [sign, setSign] = useState('')
+  getSignature(json, secret, setSign)
```
kyle-ssg: I think rather than re-evaluating this each render we may as well just do this on submit

frontend/web/components/TestWebhook.tsx
```diff
@@ -8,18 +8,72 @@ import Button from './base/forms/Button'
...
+    'X-Flagsmith-Signature': sign,
+  }
+
   const submit = () => {
     setError(null)
     setLoading(true)
     setSuccess(false)
     data
-      .post(webhook, JSON.parse(json), null)
+      .post(webhook, JSON.parse(json), headers)
```
kyle-ssg: RE above comment, this could be something more like 
```
getSignature().then((sign)=>{
 const headers = {
    'X-Flagsmith-Signature': sign,
  }
 return data.post(...)
}).then
```

frontend/web/components/TestWebhook.tsx
shubham-padia: Agreed! Fixed at https://github.com/Flagsmith/flagsmith/compare/446e37228ee813f6f6503e5d2b950850f3b593a6..39cc696750aa1916fe6376fc6babbae6676cac0b along with a bit of refactoring

frontend/web/components/TestWebhook.tsx
shubham-padia: Thanks for the review! Fixed at https://github.com/Flagsmith/flagsmith/compare/446e37228ee813f6f6503e5d2b950850f3b593a6..39cc696750aa1916fe6376fc6babbae6676cac0b along with a bit of refactoring

kyle-ssg (approved):
Great from a FE perspective, I'll handover to another member of the team from a backend perspective. Thank you very much for this!

shubham-padia:
Bump for review

api/core/signing.py
```diff
@@ -3,6 +3,8 @@
 
 
 def sign_payload(payload: str, key: str):
+    # TestWebHook on the frontend replicates this exact function, change the
+    # function there if this changes.
```
khvn26: 1. For consistency's sake, let's call the frontend function `signPayload`.
2. I'd prefer to explicitly state the frontend module name in this here comment.

shubham-padia: @khvn26 I've addressed the comments and the pushed the fix

api/webhooks/webhooks.py
```diff
@@ -153,6 +153,8 @@ def _call_webhook(
     headers = {"content-type": "application/json"}
     json_data = json.dumps(data, sort_keys=True, cls=DjangoJSONEncoder)
     if webhook.secret:
+        # TestWebHook on the frontend replicates this exact function, change the
```
khvn26:
```suggestion
        # signPayload of frontend/web/components/TestWebHook on the frontend replicates this exact function, change the
```

api/webhooks/webhooks.py
```diff
@@ -199,6 +201,8 @@ def call_webhook_with_failure_mail_after_retries(
...
```
khvn26:
```suggestion
        # signPayload of frontend/web/components/TestWebHook on the frontend replicates this exact function, change the
```

khvn26 (approved):

api/core/signing.py
khvn26: Great, thanks! I've replicated your change across the rest of the new code comments.

shubham-padia: Missed that, thanks for replicating the changes!

matthewelwell:
@shubham-padia @khvn26 looks like we've got CI failures here now, can we fix?

shubham-padia:
> @shubham-padia @khvn26 looks like we've got CI failures here now, can we fix?

@kyle-ssg fixed!

shubham-padia:
Just checking if there are any blockers to getting this merged :) !

api/edge_api/identities/edge_request_forwarder.py
```diff
@@ -81,6 +81,9 @@ def _get_headers(request_method: str, headers: dict, payload: str = "") -> dict:
     # ref: https://groups.google.com/g/django-developers/c/xjYVJN-RguA/m/G9krDqawchQJ
     if request_method == "GET":
         headers.pop("Content-Length", None)
+
```
gagantrivedi: This comment should not be here; edge-forwarding is different from webhooks

gagantrivedi (changes requested):

frontend/web/components/TestWebhook.tsx
```diff
@@ -8,34 +8,82 @@ import Button from './base/forms/Button'
...
+  let result = JSON.stringify(obj, null, 1) // stringify, with line-breaks and indents
+  result = result.replace(/^ +/gm, ' ') // remove all but the first space for each line
+  result = result.replace(/\n/g, '') // remove line-breaks
+  result = result.replace(/{ /g, '{').replace(/ }/g, '}') // remove spaces between object-braces and first/last props
+  result = result.replace(/\[ /g, '[').replace(/ \]/g, ']') // remove spaces between array-brackets and first/last items
+  return result
+}
+
+const signPayload = async (body: string, secret: string): Promise<string> => {
+  if(!secret) {
```
kyle-ssg, line 26: Just a note, added this as it was rejecting attempting to sign with an empty secret. Instead it just resolves with '' in this case.

gagantrivedi (approved):

PR #3674, opened 2024-03-26
fix: prevent tasks dying from temporary loss of db connection

matthewelwell:
## Changes

Fixes issue where threads die from a temporary loss of connection to the database by catching DatabaseError and logging.

Added a unit test to verify that the exception is handled and the correct logs are written.


api/tests/unit/environments/dynamodb/wrappers/test_unit_dynamodb_environment_v2_wrapper.py
```diff
@@ -65,9 +65,9 @@ def test_environment_v2_wrapper__get_identity_overrides_by_environment_id__last_
 ) -> None:
     # Given
     wrapper = DynamoEnvironmentV2Wrapper()
-    mocker.patch.object(
-        wrapper, "get_table"
-    ).return_value = table_mock = mocker.MagicMock(spec=flagsmith_environments_v2_table)
+    mocker.patch.object(wrapper, "get_table").return_value = table_mock = (
+        mocker.MagicMock(spec=flagsmith_environments_v2_table)
+    )
```
matthewelwell: This seems to be related to the black upgrade but I'm not sure why it wasn't caught in a previous PR. It seems fairly harmless though so I'll leave it here for now.

khvn26: Please rebase on latest main, should be fixed.

api/tests/unit/task_processor/conftest.py
```diff
@@ -1,6 +1,35 @@
...
+
+@pytest.fixture
+def get_task_processor_caplog(
+    caplog: pytest.LogCaptureFixture,
+) -> GetTaskProcessorCaplog:
+    # caplog doesn't allow you to capture logging outputs from loggers that don't
+    # propagate to root. Quick hack here to get the task_processor logger to
+    # propagate.
+    # TODO: look into using loguru.
+
```
khvn26, line 26: @matthewelwell Why doesn't it propagate in the first place?

matthewelwell: TBH, I'm not sure but I don't really want to open that can of worms here. [Context](https://github.com/Flagsmith/flagsmith/blob/main/api/app/settings/common.py#L564-L588).

khvn26 (approved):
LGTM.

I'd love to see a test with less mocking (e.g. with db fixture taken out), but it's probably not worth the hassle. In case of errors different than `DatabaseError`, we'll take it as it comes I guess.

PR #3676, opened 2024-03-26
feat: Add subscription to Hubspot tracker

zachaysan:

## Changes

Introduce tracking subscriptions in Hubspot. During a first pass at creating the Hubspot company we now add the subscription data and in addition there's a new hook on subscriptions getting saved where a plan change will update the information in Hubspot if there is already a Hubspot related organisation.

Manual testing for the actual API calls then created two tests to ensure the client code is called.


api/integrations/lead_tracking/hubspot/client.py
```diff
@@ -64,8 +67,11 @@ def create_contact(self, user: FFAdminUser, hubspot_company_id: str) -> dict:
         )
         return response.to_dict()
 
-    def create_company(self, name: str) -> dict:
-        properties = {"name": name}
+    def create_company(self, name: str, active_subscription: str) -> dict:
+        properties = {
+            "name": name,
+            "active_subscription": active_subscription,
```
matthewelwell, line 75: same comment as on the other PR, I've love to pull this out as a constant.

api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -65,14 +69,36 @@ def get_or_create_organisation_hubspot_id(self, organisation: Organisation) -> s
...
         HubspotOrganisation.objects.create(
             organisation=organisation,
             hubspot_id=response["id"],
         )
+
         return response["id"]
 
+    def update_company_active_subscription(self, subscription: Subscription) -> dict:
+        if not subscription.plan:
+            return
```
matthewelwell, line 91: This return contradicts the type definition. Not sure if the solution would be to raise an exception here instead or just to change the return type to `dict | None`

api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -65,14 +69,36 @@ def get_or_create_organisation_hubspot_id(self, organisation: Organisation) -> s
...
+
+        organisation = subscription.organisation
+
+        # Check if we're missing the associated hubspot id.
+        if not getattr(organisation, "hubspot_organisation", None):
+            return
```
matthewelwell, line 97: As above.

api/organisations/models.py
```diff
@@ -251,6 +254,13 @@ def can_auto_upgrade_seats(self) -> bool:
     def is_free_plan(self) -> bool:
         return self.plan == FREE_PLAN_ID
 
+    @hook(AFTER_SAVE, when="plan", has_changed=True)
```
matthewelwell, line 257: Does this also fire when the subscription is created, or do we not care because it's sent when we first create the organisation anyway?

api/tests/unit/integrations/lead_tracking/hubspot/test_unit_hubspot_lead_tracking.py
```diff
@@ -212,3 +214,61 @@ def test_hubspot_with_existing_contact_and_new_organisation(
...
+
+    # When
+    organisation.subscription.plan = "scale-up-v2"
+    organisation.subscription.save()
+
+    # Then
+    mock_update_company.assert_called_once_with(
+        active_subscription=organisation.subscription.plan,
+        hubspot_company_id=hubspot_id,
+    )
```
matthewelwell, line 251: I'm _slightly_ concerned that I think this test relies on the `TASK_RUN_METHOD` environment variable being set to `SYNCHRONOUSLY`, right? 

I guess our options are: 

 1. force `settings.TASK_RUN_METHOD=TaskRunMethod.SYNCHRONOUSLY`
 2. Mock the task and assert task.delay is called, not the actual underlying hubspot client method (and then unit test the task function itself)
 3. something else?

api/tests/unit/integrations/lead_tracking/hubspot/test_unit_hubspot_lead_tracking.py
```diff
@@ -212,3 +214,61 @@ def test_hubspot_with_existing_contact_and_new_organisation(
...
+
+    assert organisation.subscription.plan == "free"
+
+    # When
+    organisation.subscription.plan = "scale-up-v2"
+    organisation.subscription.save()
+
+    # Then
+    mock_update_company.assert_not_called()
```
matthewelwell, line 281: As above.

matthewelwell (changes requested):

api/tests/unit/integrations/lead_tracking/hubspot/test_unit_hubspot_lead_tracking.py
zachaysan: Well, I can force `settings.TASK_RUN_METHOD=TaskRunMethod.SYNCHRONOUSLY` if you think it's necessary, but many other tests I've written relied upon the implicit synchronicity but since you're concerned about it I can update the task to set it explicitly.

api/tests/unit/integrations/lead_tracking/hubspot/test_unit_hubspot_lead_tracking.py
zachaysan: Ok updated the test here as well

api/organisations/models.py
zachaysan: We don't care because when a user is added to a organisation we include the plan with the initial creation anyway.

api/integrations/lead_tracking/hubspot/lead_tracker.py
zachaysan: Nice catch. And yeah I think updating the return type is the right call. An exception would just be an annoyance for the caller.

api/integrations/lead_tracking/hubspot/client.py
zachaysan: As discussed on the other PR we decided against it.

api/tests/unit/integrations/lead_tracking/hubspot/test_unit_hubspot_lead_tracking.py
zachaysan: Ok that's been updated. I also merged in the first of many merge conflicts from all this Hubspot work.

matthewelwell (approved):

PR #3680, opened 2024-03-26
feat: Add org id to hubspot company

zachaysan:

## Changes

On the creation of a Hubspot company resource include the `organsation_id` in a new field which is titled `OrgID` in Hubspot, but is only accessible via `orgid` through the API due to restrictions on Hubspot's end.  

Manually checked the API response once I was able to set the param in all lowercase then updated an existing test that already checked for the right mocked call to a client resource.

### DIFF
--- a/api/integrations/lead_tracking/hubspot/client.py
+++ b/api/integrations/lead_tracking/hubspot/client.py
@@ -64,8 +64,8 @@ class HubspotClient:
         )
         return response.to_dict()
 
-    def create_company(self, name: str) -> dict:
-        properties = {"name": name}
+    def create_company(self, name: str, organisation_id: int) -> dict:
+        properties = {"name": name, "orgid": str(organisation_id)}
         simple_public_object_input_for_create = SimplePublicObjectInputForCreate(
             properties=properties,
         )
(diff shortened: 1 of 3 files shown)


api/integrations/lead_tracking/hubspot/client.py
```diff
@@ -64,8 +64,8 @@ def create_contact(self, user: FFAdminUser, hubspot_company_id: str) -> dict:
         )
         return response.to_dict()
 
-    def create_company(self, name: str) -> dict:
-        properties = {"name": name}
+    def create_company(self, name: str, organisation_id: int) -> dict:
+        properties = {"name": name, "orgid": str(organisation_id)}
```
matthewelwell, line 68: Minor: I'd love to see the hubspot parameters pulled out as constants in e.g. `hubspot.constants`.

matthewelwell (approved):

api/integrations/lead_tracking/hubspot/client.py
zachaysan: We discussed offline and decided against it for now.

PR #3687, opened 2024-03-27
docs: refactor feature state value

dabeeeenster:


docs/docs/clients/client-side/ios.md
```diff
@@ -74,7 +74,6 @@ Note that you can use:
 
 - `flag.value?.stringValue`
 - `flag.value?.intValue`
-- `flag.value?.floatValue`
```
zachaysan, line 77: Shouldn't the `boolValue` equivalent be placed here?

zachaysan (approved):
One small question otherwise looks good

PR #3701, opened 2024-04-01
fix: database Compose warnings and set a project name

rolodato:

## Changes

Fixes two warnings when launching the database from Docker Compose:

```
% docker --version
Docker version 26.0.0, build 2ae903e86c
% docker compose -f db.yml up
WARN[0000] /Users/rolodato/source/flagsmith/flagsmith/docker/db.yml: `version` is obsolete
[+] Running 1/1
 ✔ db Pulled                                                                                                   1.6s
WARN[0001] mount of type `volume` should not define `bind` option
[+] Running 0/1
 ⠋ Container docker-db-1  Created                                                                              0.0s
Attaching to db-1
[...]
```

Also, this PR adds a top-level `name` field to the Compose file. Otherwise, the default project name is `docker` (the directory name where the Compose file lives), which is not very informative.

Before:

image

After:

image

More details on the top-level `version` and `name` Compose field names: https://docs.docker.com/compose/compose-file/04-version-and-name/

Manually tested by starting the database using Compose as usual:

```
docker compose -f docker/db.yml up
```

### DIFF
--- a/docker/db.yml
+++ b/docker/db.yml
@@ -1,5 +1,3 @@
-version: '3'
-
 volumes:
   pg_11_data:
 
--- a/docker/db.yml
+++ b/docker/db.yml
@@ -1,3 +1,5 @@
+name: flagsmith
+
 volumes:
   pg_11_data:
 
--- a/docker/db.yml
+++ b/docker/db.yml
@@ -7,7 +7,7 @@ services:
     pull_policy: always
     restart: unless-stopped
     volumes:
-      - pg_11_data:/var/lib/postgresql/data:Z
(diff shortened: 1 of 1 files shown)


docker/db.yml
```diff
@@ -1,4 +1,4 @@
-version: '3'
```
matthewelwell, line 1: What's the reason for removing the `version` attribute here? Can we not keep it, and just add the `name` attribute below / above?

matthewelwell: Sorry, just read the PR description in more depth 🤦

matthewelwell (approved):

PR #3703, opened 2024-04-01
chore: upgrade and unpin docusaurus to ^3.2.0

rolodato:

## Changes

Sets the version of all docusaurus packages to `^3.2.0` (compatible-with 3.2.0), up from `3.1.0` (exactly 3.1.0). This ensures we get docusaurus fixes and improvements sooner.

I am not sure why it was originally pinned to `3.1.0` - before the upgrade to `3.1.0`, [we were not pinning an exact version](https://github.com/Flagsmith/flagsmith/blob/2c56526706ea48b48714f03103bbb4de1db7ba3a/docs/package.json#L19-L22), and currently there is a [mix of exact and non-exact versions](https://github.com/Flagsmith/flagsmith/blob/65fd5907e4d928c933e1ab6c5fb4c1527b8b745d/docs/package.json#L17-L19).

Manually by running the docs locally: `npm install && npm run start`


zachaysan (approved):
Looks good to me!

PR #3713, opened 2024-04-03
fix: allow deletion of scheduled change requests

### LINKED ISSUE #3710: Unable to delete scheduled flag update

- [ ] Self Hosted with Docker
- [X] SaaS at flagsmith.com

### Describe the bug

It is not possible to delete a scheduled change to a flag once it has been created. The delete button appears to do nothing and the network console shows that the delete request failed with an http 400 error and the following message: 

```json
{"detail":"Cannot delete a Change Request that has been committed."}
```

### Steps To Reproduce

1. Schedule an update to a flag by choosing 'Schedule Update' from the feature slide out
2. Navigate to 'Scheduling' in the environment menu and to the update that was scheduled in step 1
3. Click on 'Delete' 

### Expected behavior

The pending scheduled update is deleted

### Screenshots

_No response_

matthewelwell:

## Changes

Allow change requests that are scheduled for the future to be deleted. 

Resolves #3710 

Added 2 new unit tests.

### DIFF
--- a/api/features/versioning/models.py
+++ b/api/features/versioning/models.py
@@ -118,7 +118,8 @@ class EnvironmentFeatureVersion(
         persist: bool = True,
     ) -> None:
         now = timezone.now()
-        self.live_from = live_from or now
+
+        self.live_from = live_from or (self.live_from or now)
         self.published_at = now
         self.published_by = published_by
         if persist:
--- a/api/features/workflows/core/models.py
+++ b/api/features/workflows/core/models.py
@@ -1,6 +1,7 @@
 import importlib
 import logging
 import typing
+from datetime import datetime
 
 from core.helpers import get_current_site_url
 from core.models import (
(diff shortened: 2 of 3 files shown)


khvn26 (approved):
I'm approving this as I don't see any problems in code, but it slightly worries me from the UX perspective.

I'd much rather continue restricting deletion of scheduled CRs in the backend, adding a UI flow which offers users to remove the schedule (i.e., clear the related `live_from` attributes) prior to deleting the CR.

This would make this process more transparent and resilient to human error, IMO.

matthewelwell:
> I'm approving this as I don't see any problems in code, but it slightly worries me from the UX perspective.
> 
> I'd much rather continue restricting deletion of scheduled CRs in the backend, adding a UI flow which offers users to remove the schedule (i.e., clear the related `live_from` attributes) prior to deleting the CR.
> 
> This would make this process more transparent and resilient to human error, IMO.

This would be quite challenging I think, because we would need to 'unpublish' the change request at the same time (which isn't currently possible). Otherwise, you would essentially be marking the change request as 'complete'. 

This does make me think that there might be an issue I haven't considered here though that there is a task added to the queue when a scheduled change request is published which we may need to clear from the queue. This could be challenging, however due to issues interacting with the task queue db table. I'll take a look.

PR #3714, opened 2024-04-03
feat: Block access after seven days notice of API overage

zachaysan:

## Changes

Building upon my [previous PR](https://github.com/Flagsmith/flagsmith/pull/3698) which itself builds upon the use of API usage notification models this PR provides the following improvements:

- Setting two variables, `stop_serving_flags` and `block_access_to_admin`, to `True` for organisations that have received an API usage warning at least 7 days old. This only applies to free accounts as paid accounts are going to be automatically charged in a subsequent PR.
- Automatically undoing the above states in cases where an organisation's subscription plan is changed to something new.
- Automatically undoing the above states when an API usage notification about the 100% threshold has not been filed within the past month.

Two large tests and a decent amount of time with a debugger.


api/organisations/tasks.py
```diff
@@ -189,7 +197,81 @@ def handle_api_usage_notifications():
             )
 
 
+def restrict_use_due_to_api_limit_grace_period_over() -> None:
```
matthewelwell, line 200: I think we should perhaps have a flag to control this behaviour similar to the one that we added for the usage notifications task? 

I wonder if the flag could also include a value that defines which fields to update on the organisation in case we want to start by just blocking access to admin for example, rather than immediately stopping serving flags.

api/organisations/tasks.py
```diff
@@ -189,7 +197,81 @@ def handle_api_usage_notifications():
...
+    grace_period = timezone.now() - timedelta(days=API_USAGE_GRACE_PERIOD)
+    month_start = timezone.now() - timedelta(30)
```
matthewelwell, line 209: I think perhaps we should add a comment here explaining the logic around the fact that free plans don't have a period start / end, so we just look at the rolling 30 day period.

api/organisations/tasks.py
```diff
@@ -189,7 +197,81 @@ def handle_api_usage_notifications():
...
+    ).update(stop_serving_flags=True, block_access_to_admin=True)
+
+
+def unrestrict_after_api_limit_grace_period_is_stale() -> None:
+    month_start = timezone.now() - timedelta(30)
+    queryset = (
+        OranisationAPIUsageNotification.objects.filter(
+            notified_at__gt=month_start,
+            percent_usage__gte=100,
+        )
```
matthewelwell, line 276: I'm struggling to understand why the queryset here is different to the one in the task function above (i.e. grace period is missing).

api/organisations/tasks.py
```diff
@@ -189,7 +197,81 @@ def handle_api_usage_notifications():
...
+    # queryset after it has been iterated through fails to save.
+    # So to work around that, we reload the queryset from the ids.
+    Organisation.objects.filter(
+        id__in=organisation_ids,
+    ).update(stop_serving_flags=True, block_access_to_admin=True)
+
+
+def unrestrict_after_api_limit_grace_period_is_stale() -> None:
+    month_start = timezone.now() - timedelta(30)
+    queryset = (
```
matthewelwell: I think we should be more explicit with our variable naming here rather than `queryset`

api/organisations/tasks.py
```diff
@@ -189,7 +197,81 @@ def handle_api_usage_notifications():
...
+
+    # Due to a bug in Django, calling update on the organisations
+    # queryset after it has been iterated through fails to save.
+    # So to work around that, we reload the queryset from the ids.
+    Organisation.objects.filter(
+        id__in=organisation_ids,
+    ).update(stop_serving_flags=True, block_access_to_admin=True)
+
+
+def unrestrict_after_api_limit_grace_period_is_stale() -> None:
```
matthewelwell, line 261: I actually don't quite understand the logic for this task. My understanding is that, once an organisation has been blocked their only way to get unblocked is to upgrade their plan. What do we mean here by 'grace period is stale' ? 

The flag mentioned above should probably control this task function too...

api/tests/unit/organisations/test_unit_organisations_tasks.py
```diff
@@ -452,3 +458,182 @@ def test_handle_api_usage_notifications_above_100(
...
+
+    # Organisation which is not on the free plan is ok.
+    assert organisation5.stop_serving_flags is False
+    assert organisation5.block_access_to_admin is False
+
+    # Organisations that breached 100 are blocked.
+    assert organisation.stop_serving_flags is True
+    assert organisation.block_access_to_admin is True
+    assert organisation2.stop_serving_flags is True
+    assert organisation2.block_access_to_admin is True
```
matthewelwell: We should check that the `APILimitAccessBlock` objects are created too.

api/organisations/tasks.py
zachaysan: I think I understand what you're asking for here, but just to be clear:

Once this task has been run, but before the individual filtered organisations have been modified we would have to run something like this:

```python
for organisation in organisations:
    stop_serving_flags_enabled = flagsmith_client.get_identity_flags(
        f"org.{organisation.id}.{organisation.name}",
        traits={"organisation_id": organisation.id},
    ).is_feature_enabled("api_limiting_stop_serving_flags")
    
    blog_access_to_admin_enabled = flagsmith_client.get_identity_flags(
        f"org.{organisation.id}.{organisation.name}",
        traits={"organisation_id": organisation.id},
    ).is_feature_enabled("api_limiting_block_access_to_admin")
```

And then if either of the features are enabled, then the bulk update would include those files. Am I understanding this correctly?

api/organisations/tasks.py
zachaysan: Sounds good. I decided to go with a docstring for the function over a comment.

api/organisations/tasks.py
zachaysan: Ok I've updated it, though I tend to like the variable `queryset` for intermediary django logic.

api/organisations/tasks.py
zachaysan: Because the selected records are for users that have already been marked to be blocked. The grace period window only covers accounts that are going to be marked as blocked passed the window. But once the window has passed we don't need to check the grace window to unmark blocked accounts, we just need to check if that they're continuing breaching the API limits and if they are then the blocked window stays present.

api/tests/unit/organisations/test_unit_organisations_tasks.py
zachaysan: Good idea. I already had done so in the unrestrict test but should have done it here too.

api/organisations/tasks.py
zachaysan: (We discussed during standup but for the benefit of others reading I'll reexplain it here.)

Once an account has entered a new month *and* they no longer have recent API usage notifications they're marked to be re-enabled and added back in. They can again be blocked if they're account usage puts them over the threshold again, but in the mean time they can serve flags and access the admin.

As for the flag being used on this task as well, I don't really see the usecase since this task is unsetting a punishment and I don't think we want to incorporate flags to stop the relief action from happening.

api/organisations/tasks.py
matthewelwell: I was more thinking something along the lines of: 

```python
for organisation in organisations:
    flags = flagsmith_client.get_identity_flags(...)
    if flags.is_feature_enabled("api_limiting"):
        fields_to_update = json.looads(flags.get_feature_value("api_limiting"))["fields_to_update"]
        for field, value in fields_to_update.items():
            setattr(organisation, field, value)
        organisation.save()
```

... but maybe that's a bit over complicated / error prone. Perhaps your solution is better (although we should at least make sure that we separate out the call to get the flags and call `is_feature_enabled()` on the resultant Flags object rather than calling `get_identity_flags` twice).

zachaysan: I think the biggest deciding factor should be how onerous it will be to update the flags once we're ready in production. My gut is slightly in favour of using individual flags over the feature values because it seems simpler to set them to enabled for the identity flags, but that may just be an incorrect understanding of how much time it will take to individually set the fields to update in your method. 

If you were in change of the Flagsmith flags for this change which approach would be more efficient?

matthewelwell: I think you're right tbh, individual flags are easier to manage.

matthewelwell: @zachaysan I think there's still work to be done here?

zachaysan: Ok I've added the flagsmith flags and updated how the update logic works.

matthewelwell (approved):

PR #3715, opened 2024-04-03
feat: Support multiple OR'd search terms in sales-dashboard

rolodato:

## Changes

* Support searching by multiple terms in the Sales dashboard
* Add text to the Search box placeholder to clarify what exactly is being searched:

image

This naive-ish implementation splits search terms by commas and ORs them all together. This has implications on any existing search queries:

* `ACME, LLC` will now search for `ACME` OR `LLC`, whereas today it's an exact match. This will return the same results as before, in addition to potentially other unrelated results
* It's now impossible to use `,` as a search criterion - I don't think this matters

Given the internal scope of the sales dashboard and the value of allowing multiple search queries, I think the tradeoff for this change is worth it. This seems like a good 80/20 solution.

Example (searching for multiple queries):

image

Manually on a local instance, validated with manual testing.

### DIFF
--- a/api/sales_dashboard/templates/sales_dashboard/base.html
+++ b/api/sales_dashboard/templates/sales_dashboard/base.html
@@ -16,10 +16,10 @@
   </head>
   <body>
     <nav class="navbar navbar-dark sticky-top bg-dark flex-md-nowrap p-0 shadow">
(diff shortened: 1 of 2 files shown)


khvn26 (approved):
LGTM 👍

PR #3716, opened 2024-04-03
fix: Remove CSRF parameter from sales-dashboard search form

rolodato:

## Changes

The search form in the Sales dashboard includes a CSRF token. This token is not validated (it can be omitted with no consequence) and it is not necessary (this is a read-only GET endpoint, so it can be safely removed).

The main benefit of removing the CSRF token is it allows us to share nicer URLs to pre-made queries, such as `/sales-dashboard/?search=foo`. Otherwise, the search URLs look like `/sales-dashboard/?csrfmiddlewaretoken=obnoxiously_long_random_string_that_appears_before_the_actual_search_term&search=foo`.

Manually in development only. I've confirmed that this endpoint in production also does not validate the presence of the `csrfmiddleware` parameter.


khvn26 (approved):
Looks good 👍

matthewelwell (approved):

PR #3717, opened 2024-04-04
fix: async feature versioning test

### LINKED ISSUE #3712: Locally and sometimes in the e2e tests, the value of use_v2_feature_versioning does not update from false to true when you enable this feature.

- [ ] Self Hosted with Docker
- [ ] SaaS at flagsmith.com
- [X] Some other way (add details in description below)

### Describe the bug

The versioning e2e test may fail because the hideHistory value, which depends on whether `use_v2_feature_versioning` is enabled, is `false`.

Locally, if you create a new project and enable the Feature versioning function without any flags, the value does not change.

### Steps To Reproduce

In the e2e test:
- Run the e2e versioning test
Local:
- Create a new Organisation and a new project
- Go to the setting from an environment
- Enable Feature Versioning -> Confirm the modal
- Refresh the page
- The Feature Versioning switch is disabled, and the API response indicates that the value of `use_v2_feature_versioning` is false.

### Expected behavior

If you enabled the Feature Versioning switch and the modal is confirmed, the `use_v2_feature_versioning` would be changed to true.

### Screenshots

https://github.com/Flagsmith/flagsmith/assets/41410593/7af09141-721b-4953-8a6f-fea6c1652ea5

In the Versioning test:

Screen Shot 2024-04-02 at 18 40 38

kyle-ssg:

## Changes

We had intermittent failures in the versioning test, there are 2 reasons why this would happen

- The action to enable an environment versioning is async, it could take a while to process
- The frontend was incorrectly saving the environment when the feature was enabled


novakzaballa (approved):
LGTM, but some test are now failing.

frontend/web/components/pages/EnvironmentSettingsPage.js
```diff
@@ -115,32 +115,10 @@ const EnvironmentSettingsPage = class extends Component {
...
-        hide_sensitive_data: !!this.state.hide_sensitive_data,
-        minimum_change_request_approvals: has4EyesPermission
-          ? this.state.minimum_change_request_approvals
-          : null,
-        name: name || env.name,
-        use_identity_composite_key_for_hashing:
-          !!this.state.use_identity_composite_key_for_hashing,
-        use_mv_v2_evaluation: !!this.state.use_mv_v2_evaluation,
-      }),
-    )
```
novakzaballa, line 143: It looks like removing these lines breakes the ability to update the environment name.
image

novakzaballa (changes requested):

novakzaballa (approved):
LGTM

PR #3720, opened 2024-04-04
fix: convert CharFields to TextFields for FeatureImport / FeatureExport models

matthewelwell:

## Changes

To support Oracle and MySQL (in our enterprise edition), we need to convert the CharFields here to TextFields, since the underlying column types in these database don't support the size we need. 

Ran the migration and confirmed that no data was lost.


zachaysan (approved):
Looks good, although the constants could have been used for the migration.

PR #3729, opened 2024-04-08
feat: Add API usage billing

zachaysan:

## Changes

Pretty complicated PR. This introduces charges for API usage overages for monthly billed customers. Once a customer is billed for overages, a record is created and stops them from being billed twice in the same month. Only monthly billed customers are permitted to be billed, as the yearly paying customers need to be handled by sales. In order to ensure that customers are notified before they're billed, there is a pre-step that checks to see if the API usage notification has been sent out before including the customer in the billing step.

Manually tested against ChargeBee's backend and verified the results and also added four or five tests to ensure the workflow is proceeding according to plan.


api/organisations/chargebee/chargebee.py
```diff
@@ -203,6 +209,54 @@ def add_single_seat(subscription_id: str):
...
+) -> None:
+    if not count:
+        return
+    try:
+        chargebee.Subscription.update(
+            subscription_id,
+            {
+                "addons": [{"id": addon_id, "quantity": count}],
+                "prorate": True,
+                "invoice_immediately": invoice_immediately,
```
matthewelwell: This doesn't seem to fit what we discussed, right? particularly, I'm not sure why we'd prorate the amount here. My understanding was that we were going to process this at the end of each month and invoice them for the amount they went over for the previous month.

api/organisations/migrations/0053_create_api_billing.py
```diff
@@ -0,0 +1,30 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('organisations', '0052_create_hubspot_organisation'),
+    ]
+
+    operations = [
+        migrations.RenameModel(
+            old_name='OranisationAPIUsageNotification',
+            new_name='OrganisationAPIUsageNotification',
```
matthewelwell: I'm a bit concerned that this will create a schema mismatch when we perform the deployment 🤔 I guess it's possibly not a concern because nothing (at least nothing in any current user journeys) is querying that table.

matthewelwell: An alternative if we are concerned about it though would be to just keep the table name as the typo'd name and update the name in the code.

api/organisations/models.py
```diff
@@ -478,3 +478,15 @@ class HubspotOrganisation(models.Model):
...
+class OrganisationAPIBilling(models.Model):
+    organisation = models.ForeignKey(
+        Organisation, on_delete=models.CASCADE, related_name="api_billing"
+    )
+    api_overage = models.IntegerField(null=False)
+    immediate_invoice = models.BooleanField(null=False, default=False)
+    billed_at = models.DateTimeField(null=False)
+
+    created_at = models.DateTimeField(null=True, auto_now_add=True)
+    updated_at = models.DateTimeField(null=True, auto_now=True)
```
matthewelwell: I think this class could definitely use some documentation.

api/organisations/tasks.py
```diff
@@ -189,7 +201,76 @@ def handle_api_usage_notifications():
             )
 
 
+def charge_for_api_call_count_overages():
+    now = timezone.now()
+    api_usage_notified_at = now - timedelta(days=30)
+    month_window_start = timedelta(days=25)
+    month_window_end = timedelta(days=35)
```
matthewelwell, line 223: I'm not sure why these numbers are 25 / 35, they seem fairly arbitrary - perhaps a comment here would be helpful?

api/organisations/tasks.py
```diff
@@ -189,7 +201,76 @@ def handle_api_usage_notifications():
             )
 
 
+def charge_for_api_call_count_overages():
+    now = timezone.now()
+    api_usage_notified_at = now - timedelta(days=30)
```
matthewelwell, line 216: I'm not sure what this is?

api/organisations/tasks.py
```diff
@@ -189,7 +201,76 @@ def handle_api_usage_notifications():
...
+    month_window_start = timedelta(days=25)
+    month_window_end = timedelta(days=35)
+    closing_billing_term = now + timedelta(hours=12)
```
matthewelwell: I'm not sure what this is?

api/organisations/tasks.py
```diff
@@ -189,7 +201,76 @@ def handle_api_usage_notifications():
...
+        subscription_information_cache__current_billing_term_starts_at__lte=F(
+            "subscription_information_cache__current_billing_term_ends_at"
+        )
+        - month_window_start,
+        subscription_information_cache__current_billing_term_starts_at__gte=F(
+            "subscription_information_cache__current_billing_term_ends_at"
+        )
+        - month_window_end,
+    ).select_related(
+        "subscription_information_cache",
```
matthewelwell, line 253: We might as well select related the `subscription` here too, right?

api/organisations/migrations/0053_create_api_billing.py
zachaysan: I think that since nothing is using the table right now we should just rename the table. Nothing is going to break and who wants to deal with a legacy typo for decades?

api/organisations/chargebee/chargebee.py
zachaysan: Yeah you're definitely right about the prorate option, I should have caught that. I've updated it. 

Otherwise the rest of what you said should be true. We add this charge to the clients subscription near the end of the month and it should be automatically added to their account and be charged with that month's billing.

api/organisations/models.py
zachaysan: Good idea, added it locally.

api/organisations/tasks.py
zachaysan: Nice catch. Added.

api/organisations/tasks.py
zachaysan: I've added a comment to explain it 👍🏼

api/organisations/tasks.py
zachaysan: Added a comment. 👍🏼

api/organisations/tasks.py
zachaysan: I've added a comment!

api/organisations/models.py
```diff
@@ -478,3 +478,29 @@ class HubspotOrganisation(models.Model):
...
+
+class OrganisationAPIBilling(models.Model):
+    """
+    Tracks API billing for when accounts go over their API usage
+    limits. This model is what allows subsequent billing runs
+    to not double bill an organisation for the same use.
+
+    Even though api_overage is charge per thousand API calls, this
+    class tracks the actual rounded count of API calls that are
+    billed for (i.e., 52000 for an account with 52233 api calls).
```
matthewelwell, line 503: Shouldn't this be 53000?

api/organisations/tasks.py
```diff
@@ -189,7 +201,88 @@ def handle_api_usage_notifications():
...
+        .exclude(
+            organisation__api_billing__billed_at__gt=api_usage_notified_at,
+        )
+        .values_list("organisation_id", flat=True)
+    )
+
+    for organisation in Organisation.objects.filter(
+        id__in=organisation_ids,
+        subscription_information_cache__current_billing_term_ends_at__lte=closing_billing_term,
+        subscription_information_cache__current_billing_term_ends_at__gte=now,
```
matthewelwell, line 243: Ok, so we're essentially finding organisations that are up to 12 hours away from the end of their billing period. Does this mean that any usage in the last 12 hours of their billing period will not be billed? Should we actually do this within 12 hours after the end of their billing period? I don't know when Chargebee actually generates the invoice, is there a window that we can squeeze into?

api/organisations/tasks.py
```diff
@@ -189,7 +201,88 @@ def handle_api_usage_notifications():
...
+        subscription_information_cache__current_billing_term_starts_at__lte=F(
+            "subscription_information_cache__current_billing_term_ends_at"
+        )
+        - month_window_start,
+        subscription_information_cache__current_billing_term_starts_at__gte=F(
+            "subscription_information_cache__current_billing_term_ends_at"
+        )
+        - month_window_end,
```
matthewelwell, line 251: Is this essentially just to exclude any plans on a non-monthly billing cycle then? (if so, can we add that to the comments somewhere - probably just on the `month_window_x` variable definitions above).

api/organisations/models.py
zachaysan: I was thinking we should under charge the usage to the nearest thousand. I would be kinda ticked off if I found out one of my providers were nickel and diming me by rounding up!

api/organisations/tasks.py
zachaysan: Well we can do it after the billing period is over, but it just feels better to bill them within the same period. It's how my cell phone bill works, after all. Plus we get to collect the overage a little sooner.

One thing we can consider is to run the task more frequently and bill them, say, one hour before the subscription is billed. 

Or we can do it your way and push it into the next billing period. I'm not totally opposed to that.

api/organisations/tasks.py
zachaysan: Ok I've tweaked the existing language at the variable definitions to highlight the non-inclusion aspect of it.

api/organisations/models.py
matthewelwell: I'd have said it makes most sense to round up since all our plans are 'up to X amount' but, since 1000 api requests will be about ~$0.07 I really don't care. Maybe just add an extra line here to say that it was a conscious decision to round down so that it's not confusing in the future.

api/organisations/tasks.py
matthewelwell: > Well we can do it after the billing period is over, but it just feels better to bill them within the same period. It's how my cell phone bill works, after all. Plus we get to collect the overage a little sooner.

Yeah, this would be the ideal, but I think for e.g. your cell phone bill there's a window of time between when your billing period ends, and when your bill is calculated. If there's a way we can do that, then great, otherwise maybe this is the best option: 

> One thing we can consider is to run the task more frequently and bill them, say, one hour before the subscription is billed.

api/organisations/models.py
zachaysan: Ok I've added an extra line to the docstring.

api/organisations/tasks.py
zachaysan: I've read through the documentation and I don't think there is a way of separating the end of the subscription term and the application of addons, so I'm going to run the billing more frequently. I think it's ok to have unbilled API usage at the last ~30 minutes of a month.

zachaysan: Ok this has been added.

matthewelwell (approved):

zachaysan:
If the merge conflicts wasn't enough now I'm hitting the codecov :(

Will update the tests for this when I can.

zachaysan:
Ok the code coverage is now at 100%. @matthewelwell did you want to take a look at the introduced tests before merging or are we good to go?

api/tests/unit/organisations/test_unit_organisations_tasks.py
```diff
@@ -649,3 +968,20 @@ def test_unrestrict_after_api_limit_grace_period_is_stale(
     assert organisation4.stop_serving_flags is True
     assert organisation4.block_access_to_admin is True
     assert getattr(organisation4, "api_limit_access_block", None) is None
+
+
+def test_recurring_tasks(mocker: MockerFixture, settings: SettingsWrapper) -> None:
```
matthewelwell: minor: 
```suggestion
def test_register_recurring_tasks(mocker: MockerFixture, settings: SettingsWrapper) -> None:
```

api/tests/unit/organisations/test_unit_organisations_tasks.py
```diff
@@ -649,3 +968,20 @@ def test_unrestrict_after_api_limit_grace_period_is_stale(
...
+    # When
+    register_recurring_tasks()
+
+    # Then
+    register_task_mock.call_args_list == [
+        call(run_every=timedelta(seconds=43200)),
+        call(run_every=timedelta(seconds=1800)),
+        call(run_every=timedelta(seconds=43200)),
+        call(run_every=timedelta(seconds=43200)),
+    ]
```
matthewelwell, line 990: I have a feeling you're going to hate me for this, because I know you only really implemented this test for the coverage... but this test feels fairly pointless without actually asserting the functions that we're wrapping. How tricky is it to extend this test to check the task functions? 

I think we'd just need to do something like: 

```python
assert register_task_mock.return_value.call_args_list = [
    call(task_function_1),
    ...,
    call(task_function_n),
]
```

matthewelwell (approved):
I've added a couple of minor comments, but I've approved it nonetheless as I don't think they're critical.

api/tests/unit/organisations/test_unit_organisations_tasks.py
zachaysan: Yeah good call. I considered doing this before but I (incorrectly) thought the mocking would be more complicated.

Added.

PR #3734, opened 2024-04-09
fix: segment override assignment

kyle-ssg:

## Changes

We were accidentally reassigning segment override feature_segments due to how we were mapping multivariate feature state values. This happened to work for everyone so far as the order of feature segments always matched the order of feature states.


matthewelwell (approved):
I've manually tested (using the preview URL) that this resolves the issue found in production.

PR #3742, opened 2024-04-09
chore: Convert project permissions TestCase to normal test function

zachaysan:

## Changes


zachaysan:
@matthewelwell I tagged you for this PR since we talked about the removal of the dead code which is part of this PR.

api/tests/types.py
```diff
@@ -1,11 +1,15 @@
...
 WithProjectPermissionsCallable = Callable[
-    [list[str], Project | None], UserProjectPermission
+    [list[str], int | None], UserProjectPermission
+]
+WithOrganisationPermissionsCallable = Callable[
+    [list[str], int | None], UserOrganisationPermission
 ]
 WithEnvironmentPermissionsCallable = Callable[
-    [list[str], Project | None], UserEnvironmentPermission
+    [list[str], int | None, bool], UserEnvironmentPermission
```
matthewelwell: I think this fix was added to the wrong type here? I think it's just the `WithProjectPermissionsCallable` that currently accepts the `admin` boolean?

matthewelwell (approved):
Approved with one minor comment regarding the typing fix.

api/tests/types.py
zachaysan: Wow that's a great catch during a PR review! Thank you!!

zachaysan: Also, if we were using mypy these types of things would be caught more automatically.

PR #3755, opened 2024-04-11
fix: add migration file to generate stale flags tag for all existing projects

matthewelwell:

## Changes

Currently the stale flags task is failing because it is trying to create stale flags in projects that don't have the tag created in. 

This PR adds a migration to add the default tag to all existing projects, note that all new projects already have the tag automatically created. 

I also considered the option of having the task create the stale flags tag lazily, but we decided that we want to be able to show our customers that they can get the benefit of stale flags functionality if they upgrade their plan. They are also free to use the tag manually themselves if they want. 

Ran the migration manually locally to confirm that the tag was created for projects that didn't already have the tag and was not created for those projects that do have it.


novakzaballa (approved):

matthewelwell:
I'd like to discuss this with @kyle-ssg because this will add the Stale tag to all existing projects in the Flagsmith platform, but I'm not sure that the experience is quite right since it seems to suggest that the Stale flag behaviour will work for all subscription plans, but that's not the case - it will only work for Enterprise.

matthewelwell:
Converted to draft until the above is resolved.

matthewelwell:
Closing this until we can focus on this (after versioning is released)

PR #3756, opened 2024-04-11
fix: Filter versioned features

### LINKED ISSUE #3243: Filter flags in dashboard based on flag state

zachaysan:

## Changes

Fixes https://github.com/Flagsmith/flagsmith/issues/3243

Replace the existing mechanism of filtering feature states with a new one that preselects fewer models and runs a slimmed down filter step to select features based on their most recent version.

Added a versioned section to an existing test and ensured that the most recent version is selected.


api/features/versioning/versioning_service.py
```diff
@@ -3,15 +3,35 @@
...
 from django.utils import timezone
 
+from environments.models import Environment
 from features.models import FeatureState
 from features.versioning.models import EnvironmentFeatureVersion
 
-if typing.TYPE_CHECKING:
-    from environments.models import Environment
+
+def get_environment_flags_latest(
```
matthewelwell: I don't quite understand how this differs from `get_environment_flags_list` ? 

Can we get any re-use here? If not, we should at least add docstrings to these functions explaining why they're different.

zachaysan: I've added the docstring. There are enough differences between the two functions where I think we need it. The `select_related` in `get_environment_flags_list` the least of them.

kyle-ssg:
@zachaysan feel free to turn on the feature_enabled_state_filter flag when this is in 🚀

api/features/versioning/versioning_service.py
matthewelwell: Ok, yep - I think I see now. My argument then though is that this method could be confusing since (at least as of now) there isn't a use case for mixing up identity & segment overrides with environment default flags - so why offer that functionality?   

Then, if we follow that thread, we should probably code in the `identity__isnull` and `feature_segment__isnull` filters... but then isn't this equivalent to: 

```python
additional_filters = additional_filters or Q() & Q(feature_segment__isnull=True, identity__isnull=True)
return get_environment_flags_list(environment)
```

The code in this module is already complicated enough, I'm really keen not to add more complications if we can avoid it.

zachaysan: Ok I still think the extra `select_related` in that function slows it down more than it helps, but I'm fine just doing it this way if it's what you want. I've updated the PR with the changes.

matthewelwell (approved):

PR #3757, opened 2024-04-11
docs: Add JSON environment document tip

zachaysan:

- [x] I have added information to `docs/` if required so people know about the feature! ❤️ 

## Changes

Added text to explain the JSON Environment Document near the local evaluation mode section of the documentation.

Booted the document server and verified it rendered correctly.

### DIFF
--- a/docs/docs/clients/overview.md
+++ b/docs/docs/clients/overview.md
@@ -166,6 +166,16 @@ Local Evaluation mode is only available with Server-side SDKs.
 - After 60 seconds have elapsed, the SDK will refresh the JSON Environment document with a network call to the Flagsmith
   API.
 
+:::tip
+
+A JSON Environment Document in the context of a feature flag company is a structured file containing configuration
+settings for feature flags within different environments, such as development, staging, and production. It typically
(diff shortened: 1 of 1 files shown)


docs/docs/clients/overview.md
```diff
@@ -166,6 +166,16 @@ Local Evaluation mode is only available with Server-side SDKs.
 - After 60 seconds have elapsed, the SDK will refresh the JSON Environment document with a network call to the Flagsmith
   API.
 
+:::tip
+
+A JSON Environment Document in the context of a feature flag company is a structured file containing configuration
+settings for feature flags within different environments, such as development, staging, and production. It typically
```
matthewelwell: This makes it sound like a single document contains information about multiple environments which isn't true.

I think we should also be more specific here, I'm not sure why we're using 'feature flag company' rather than just talking in the context of Flagsmith?

zachaysan: Good feedback. I've updated the wording.

dabeeeenster (approved):

PR #3760, opened 2024-04-12
fix: edge API not updated when versioned change request committed

matthewelwell:

## Changes

Adds logic to trigger the `rebuild_environment_task` when a change request containing environment feature version objects is committed.  

This PR also includes some minor tidying up of imports to remove ugly relative imports. 

Updated existing unit test.

### DIFF
--- a/api/features/workflows/core/models.py
+++ b/api/features/workflows/core/models.py
@@ -35,11 +35,11 @@ from audit.tasks import (
     create_feature_state_updated_by_change_request_audit_log,
     create_feature_state_went_live_audit_log,
 )
+from environments.tasks import rebuild_environment_document
 from features.models import FeatureState
+from features.versioning.models import EnvironmentFeatureVersion
 from features.versioning.tasks import trigger_update_version_webhooks
-
-from ...versioning.models import EnvironmentFeatureVersion
-from .exceptions import (
+from features.workflows.core.exceptions import (
     CannotApproveOwnChangeRequest,
     ChangeRequestDeletionError,
     ChangeRequestNotApprovedError,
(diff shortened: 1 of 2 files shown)


api/features/workflows/core/models.py
```diff
@@ -35,11 +35,11 @@
...
     create_feature_state_went_live_audit_log,
 )
+from environments.tasks import rebuild_environment_document
 from features.models import FeatureState
+from features.versioning.models import EnvironmentFeatureVersion
 from features.versioning.tasks import trigger_update_version_webhooks
-
-from ...versioning.models import EnvironmentFeatureVersion
-from .exceptions import (
+from features.workflows.core.exceptions import (
```
zachaysan, line 42: In general I think absolute imports are always fine, but one level of relative imports like `from .exceptions import ...` are acceptable too. So I agree with your redefinition of `from ...versioning.models import EnvironmentFeatureVersion` since it's too many levels deep but I don't really see the importance of redefining the exceptions module. 

That said, I'm not really hung up on it, I'm more just interested in knowing why you're favouring the absolute import for exceptions here.

api/features/workflows/core/models.py
```diff
@@ -165,6 +165,10 @@ def _publish_environment_feature_versions(
                     },
                     delay_until=environment_feature_version.live_from,
                 )
+                rebuild_environment_document.delay(
+                    kwargs={"environment_id": self.environment_id},
+                    delay_until=environment_feature_version.live_from,
+                )
```
zachaysan, line 171: Just to confirm: The `live_from` of the environment feature version could be well into the future, correct? But that's ok since our task system is designed to handle it? Just confirming we'd survive, say, a task scheduled further in the future to survive a backend deployment or similar.

zachaysan (approved):
Everything looks good, just a couple of small questions.

api/features/workflows/core/models.py
matthewelwell: Yeah, that's a fair point. I think I just like standardising to a single approach.

api/features/workflows/core/models.py
matthewelwell: Yep, it's persisted to the postgres database.

PR #3761, opened 2024-04-12
docs: seo optimisations

matthewelwell:

## Changes

Adds 2 changes to the docs site as requested by SEO agency. 

 1. Add 'feature flag tool' to the intro statement
 2. Adds all our how-to guides as links in the footer

I don't love the aesthetic of the how-to guides in the footer, but if it's helpful for SEO then we should probably just swallow it for now. I can't see any obvious options to split the columns, so we do end up with one large column of all of the guides. 

image

Ran the docs site locally and verified the changes.

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -127,48 +127,48 @@ const config = {
                                 href: 'https://www.flagsmith.com/blog/angular-feature-flag',
                             },
                             {
-                                label: 'Python Feature Flags',
-                                href: 'https://www.flagsmith.com/blog/python-feature-flag',
-                            },
-                            {
-                                label: 'Java Feature Flags',
-                                href: 'https://www.flagsmith.com/blog/java-feature-toggle',
-                            },
-                            {
-                                label: 'Ruby Feature Flags',
-                                href: 'https://www.flagsmith.com/blog/ruby-feature-flags',
-                            },
-                            {
-                                label: 'JavaScript Feature Flags',
-                                href: 'https://www.flagsmith.com/blog/javascript-feature-flags',
-                            },
-                            {
                                 label: 'Flutter Feature Flags',
                                 href: 'https://www.flagsmith.com/blog/flutter-feature-flags',
-                            },
-                            {
-                                label: 'Node.js Feature Flags',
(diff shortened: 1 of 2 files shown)


docs/docusaurus.config.js
```diff
@@ -119,6 +119,63 @@ const config = {
                             },
                         ],
                     },
+                    {
+                        title: 'How-to guides',
+                        items: [
+                            {
+                                label: 'Angular Feature Flags',
```
rolodato, line 126: Is it necessary to have "Feature Flags" on each label? It looks visually noisy IMO but if there's truly a SEO benefit I'm guessing it's okay.

docs/docusaurus.config.js
```diff
@@ -119,6 +119,63 @@ const config = {
...
+                            {
+                                label: 'Java Feature Flags',
+                                href: 'https://www.flagsmith.com/blog/java-feature-toggle',
+                            },
+                            {
+                                label: 'Ruby Feature Flags',
+                                href: 'https://www.flagsmith.com/blog/ruby-feature-flags',
+                            },
+                            {
+                                label: 'Javascript Feature Flags',
```
rolodato:
```suggestion
                                label: 'JavaScript Feature Flags',
```

JavaScript is the correct capitalisation :)

docs/docusaurus.config.js
```diff
@@ -119,6 +119,63 @@ const config = {
                             },
                         ],
                     },
+                    {
+                        title: 'How-to guides',
```
rolodato, line 123: Is the order for these links intentional? I'm curious why Angular is first but more popular platforms like iOS, React Native etc are lower.

I don't have a strong preference either way - maybe alphabetical order makes it easier to skim through the list and makes it easier to decide what goes first.

rolodato (approved):
Left some thoughts, nothing worth blocking the PR 👍

docs/docusaurus.config.js
dabeeeenster: Yes I think alphabetical here?

docs/docusaurus.config.js
matthewelwell: I think there probably is an SEO benefit here, but I will check on this as it's a good point.

Issue #1762, opened 2022-12-12
Enable per-identity overrides in Local Evaluation mode


dabeeeenster:
# Overview

There is currently a functional deficiency in running the SDKs in Local Evaluation mode in that they ignore Identity overrides. This issue seeks to fix that deficiency. To do this, we plan to implement an (optional) limit to the number of Identity overrides per Environment. 

# Caveats that would be solved by Identity limiting

## Allow for Identity overrides in Local Evaluation mode

When running the server side SDKs in Local Evaluation mode, we don't store any of the Identity overrides in the `GET Environment Document` endpoint as it could contain unlimited data. If we limited the number of Identity overrides, we could include that either in the main Environment document or as a supplementary endpoint. With these two pieces of data, local evaluation mode would include Identity Overrides.

## Show identity overrides for an environment

When moving to Edge, we removed the option of seeing a list of Indentities that are overriding a particular Environment, as we have no index that can form this data. Limiting the number of overrides would allow us to store this data as a single large object of denormalised data. We could then reintroduce this view. 

# Technical Approach

If we limited the number of Identity overrides we could store the override data as a single piece, either in the main Environment document or as a supplementary endpoint. With these two pieces of data, local evaluation mode would include Identity Overrides.

# Questions and Complications

- Would the limit be based on the number of identity overrides or the total amount of data they consume?
- What would the limit be?
- DynamoDB has a max document size of 400kb.
- What would we do with existing Edge overrides? 
- If we do this, the only easily accessible source of identity override data in Edge is the Dynamo Environment Document. Are we ok reading then writing to this?

kyle-ssg:
From a frontend perspective, I think that the only requirement is to show an appropriate error when adding an override in the dashboard fails ? Potentially if we return this usage in /environments/:id I can display a warning too.

kyle-ssg:
Please assign back to me when we have a method to retrieving / detecting max identity overrides.

matthewelwell:
The only thing left here, is to support pagination, which only affects very large environments. I have created a separate issue to track the progress of this [here](https://github.com/Flagsmith/flagsmith/issues/4664).

PR #3766, opened 2024-04-15
feat: Identity overrides in environment document

khvn26:

## Changes

This PR adds identity overrides to the environment document response for local evaluation.

Added a unit test for the new mapper.

Contributes to #1762.

### DIFF
--- a/api/environments/identities/managers.py
+++ b/api/environments/identities/managers.py
@@ -1,6 +1,6 @@
 from typing import TYPE_CHECKING, Iterable
 
-from django.db.models import Manager
+from django.db.models import Count, Manager, QuerySet
 
 if TYPE_CHECKING:
     from environments.identities.models import Identity
@@ -29,4 +29,12 @@ class IdentityManager(Manager["Identity"]):
             )
             .prefetch_related("identity_traits")
             .get_or_create(identifier=identifier, environment=environment)
+        )
+
+    def with_overrides(self) -> QuerySet["Identity"]:
+        """
+        Only include identities with overridden features.
+        """
+        return self.annotate(overridden_count=Count("identity_features")).filter(
+            overridden_count__gt=0
         )
--- a/api/environments/identities/managers.py
+++ b/api/environments/identities/managers.py
@@ -1,6 +1,6 @@
 from typing import TYPE_CHECKING, Iterable
 
-from django.db.models import Manager, Prefetch, QuerySet
+from django.db.models import Manager
 
 if TYPE_CHECKING:
     from environments.identities.models import Identity
@@ -30,11 +30,3 @@ class IdentityManager(Manager["Identity"]):
             .prefetch_related("identity_traits")
             .get_or_create(identifier=identifier, environment=environment)
         )
-
-    def only_overrides(self) -> QuerySet["Identity"]:
-        """
-        Only include identities with overridden features.
-        """
-        return self.prefetch_related(
-            Prefetch("identity_features"),
-        ).filter(identity_features__isnull=False)
--- a/api/environments/identities/managers.py
+++ b/api/environments/identities/managers.py
@@ -1,6 +1,6 @@
 from typing import TYPE_CHECKING, Iterable
 
-from django.db.models import Count, Manager, QuerySet
+from django.db.models import Manager, Prefetch, QuerySet
 
 if TYPE_CHECKING:
     from environments.identities.models import Identity
@@ -31,10 +31,10 @@ class IdentityManager(Manager["Identity"]):
             .get_or_create(identifier=identifier, environment=environment)
         )
 
-    def with_overrides(self) -> QuerySet["Identity"]:
+    def only_overrides(self) -> QuerySet["Identity"]:
         """
         Only include identities with overridden features.
         """
-        return self.annotate(overridden_count=Count("identity_features")).filter(
-            overridden_count__gt=0
-        )
+        return self.prefetch_related(
+            Prefetch("identity_features"),
+        ).filter(identity_features__isnull=False)
--- a/api/environments/models.py
+++ b/api/environments/models.py
@@ -1,17 +1,15 @@
-# -*- coding: utf-8 -*-
-from __future__ import unicode_literals
-
 import logging
 import typing
 from copy import deepcopy
 
 from core.models import abstract_base_auditable_model_factory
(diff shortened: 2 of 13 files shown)


api/environments/identities/managers.py
```diff
@@ -30,3 +30,11 @@ def get_or_create_for_sdk(
...
             .get_or_create(identifier=identifier, environment=environment)
         )
+
+    def with_overrides(self) -> QuerySet["Identity"]:
+        """
+        Only include identities with overridden features.
+        """
+        return self.annotate(overridden_count=Count("identity_features")).filter(
+            overridden_count__gt=0
+        )
```
khvn26: I'm not sure it's the best way to do this, given that `filter_for_environment_document` prefetches all the feature states.

Perhaps this could be delegated to pure Python in the mapper by iterating `environment.feature_states.all()`.

matthewelwell: This might also get painful in the EE merge so I'm certainly for doing this in pure python.

api/util/mappers/sdk.py
```diff
@@ -0,0 +1,51 @@
...
+
+def map_environment_to_sdk_document(
+    environment: "Environment",
+) -> SDKDocument:
+    """
+    Map an `environments.models.Environment` instance to an SDK document
+    used by SDKs with local evaluation mode.
+
+    It's virtually the same data that gets indexed in DynamoDB,
+    except it presents identity overrides and omits integrations configurations.
```
matthewelwell, line 28: Intrigued to understand which part of the below actually omits the integrations configurations?

khvn26: See the `exclude` kwarg on line 50: https://github.com/Flagsmith/flagsmith/blob/ade08b0f6f84b0d27d96f1602d3ec0518b7ad578/api/util/mappers/sdk.py#L50

matthewelwell: Blimey, yep, not sure how I missed that 🤦

api/environments/identities/managers.py
khvn26: Couldn't avoid it but the SQL is much more straightforward now.

api/tests/unit/util/mappers/test_unit_mappers_sdk.py
```diff
@@ -0,0 +1,115 @@
...
+
+    identities_with_overrides = (
+        Identity.objects.only_overrides().filter(environment=environment).all()
+    )
+
+    # When
+    result = map_environment_to_sdk_document(
+        environment,
+        identities_with_overrides=identities_with_overrides,
+    )
```
matthewelwell: I'd love to see an assertion on the number of queries here (along with maybe multiple identities / identity overrides).

api/util/mappers/engine.py
```diff
@@ -333,7 +321,6 @@ def map_environment_to_engine(
         #
         # Integrations:
         amplitude_config=amplitude_config_model,
-        dynatrace_config=dynatrace_config_model,
```
matthewelwell, line 336: I'm unsure why this has been removed here?

khvn26: Dynatrace integration isn't currently (or wasn't ever) an identity one. We don't need it indexed for Edge API.

api/tests/unit/util/mappers/test_unit_mappers_sdk.py
khvn26: This is a valid request but the assertion won't belong to the mapper test as it expects already fetched data with the `identities_with_overrides` kwarg. I'll add multiple overrides to the environment-document view test.

khvn26: Done, and fixed an N+1 issue in the process 😬

api/environments/models.py
```diff
@@ -369,8 +381,41 @@ def _get_environment_document_from_db(
...
+            extra_prefetch_related=[
+                Prefetch(
+                    "feature_states",
+                    queryset=FeatureState.objects.select_related(
+                        "feature",
+                        "feature_state_value",
+                        "identity",
+                        "identity__environment",
+                    ).prefetch_related(
+                        "identity__identity_traits",
```
matthewelwell: I'm unsure why we would need traits here?

api/tests/unit/environments/test_unit_environments_views_sdk_environment.py
```diff
@@ -76,7 +108,7 @@ def test_get_environment_document(
     url = reverse("api-v1:environment-document")
 
     # When
-    with django_assert_num_queries(13):
+    with django_assert_num_queries(16):
```
matthewelwell: I guess this increase is due to all the extra prefetch queries added to handle the identity overrides?

docs/docs/clients/overview.md
```diff
@@ -350,5 +350,3 @@ are all computed locally.
 - [Analytics-based Integrations](/integrations/overview#analytics-platforms) do not run.
   [Flag Analytics](/advanced-use/flag-analytics) do still work, if enabled within the
   [SDK setup](/clients/server-side#configuring-the-sdk).
-- In circumstances where you need to target a specific identity, you can do this by creating a segment to target that
-  specific user and subsequently adding a segment override for that segment.
```
matthewelwell: If we're removing this, shouldn't we also remove this line: https://github.com/Flagsmith/flagsmith/pull/3766/files#diff-a4321d583bb04a398f14c6f6164fe85200d8687ac5c617f8e52b631376463809L349 

... but should we also add some data about the Edge API opt-in process here?

api/tests/unit/environments/test_unit_environments_views_sdk_environment.py
khvn26: Yes.

api/environments/models.py
khvn26: I thought I had fair reasoning for that, but, upon reviewing, I tend to agree with your question/comment. Removed both trait prefetching and trait mapping to the environment document.

docs/docs/clients/overview.md
khvn26: I've removed Ben's commit as we should really update the docs as part of FE work after #3881.

matthewelwell (approved):

PR #3771, opened 2024-04-15
fix: odd behaviour seen when using REPLICA_DATABASE_URLS

### LINKED ISSUE #3681: After upgrade from v2.66.2 to 2.104.0 some POST calls fail when REPLICA_DATABASE_URLS is set

- [ ] Self Hosted with Docker
- [X] Self Hosted with Kubernetes
- [ ] SaaS at flagsmith.com

### Describe the bug

We use Aurora RDS/Postgres with one writer and one reader. We've been using REPLICA_DATABASE_URLS to split reads/writes between these 2 instances. After upgrading from v2.66.2 to v2.104.0, we've seen failing POSTS (both directly through the api and also the UI) that throw a 500 error.  Things like creating feature, create segment override. When we don't set REPLICA_DATABASE_URLS, these all work just fine. Seems like some of the POSTS are not routing to the write instance.

### Steps To Reproduce

Set REPLICA_DATABASE_URLS
Create environment
Try to create feature fails with 500 error

### Expected behavior

Set REPLICA_DATABASE_URLS
Create environment
Try to create feature succeeds

### Screenshots

_No response_

matthewelwell:

## Changes

Fixes https://github.com/Flagsmith/flagsmith/issues/3681. 

Ran the API locally with `REPLICA_DATABASE_URLS` set up to point to the same database as `DATABASE_URL` and observed that, without this change, I received an exception when trying to create a Feature. 

Once I made this change, I no longer received the exception.

### DIFF
--- a/api/core/signals.py
+++ b/api/core/signals.py
@@ -30,7 +30,7 @@ def create_audit_log_from_historical_record(
         else None
     )
 
-    environment, project = history_instance.instance.get_environment_and_project()
+    environment, project = instance.get_environment_and_project()
     if project != history_instance.instance and (
         (project and project.deleted_at)
         or (environment and environment.project.deleted_at)


matthewelwell:
@zachaysan @khvn26 I am requesting a review from both of you on this PR. Zach, because you have the context from our investigation of this issue. Kim, because you might have some context as to why `history_instance.instance` was used in the first place, rather than just `instance`.

zachaysan (approved):
I'm happy to see this fixed, but we really should make a ticket for creating a test suite to run with a replica read database in addition to running the whole suite as just the default database. I know 99% of the time it would just duplicate the work, but the small part of the time to catch these regressions would really help out to find issues for our clients that use the replica databases.

khvn26 (approved):
Approved but echoing Zach's proposal here. I'm sure we're going to bump into problems with replicated setups in the future, see https://andrewbrookins.com/python/scaling-django-with-postgres-read-replicas/ ("Solving Consistency Problems Caused By Replication Lag").

matthewelwell:
@zachaysan @khvn26 yeah, you're both right for sure. I hoped it would be as simple as e.g.: 

```shell
DATABASE_URL=postgres://postgres:password@127.0.0.1:5432/flagsmith \
 REPLICA_DATABASE_URLS=postgres://postgres:password@127.0.0.1:5432/flagsmith \
 pytest
```

but this causes all the tests to fail with a spurious error: 

```
ERROR tests/integration/ core/test_commands.py::test_bootstrap__empty_instance__creates_expected - django.db.utils.ProgrammingError: column users_ffadminuser.organisation_id does not exist
```

I agree we should implement some testing against this, but it looks like it needs some thought.

PR #3773, opened 2024-04-16
feat: Clone identity flag states

### LINKED ISSUE #2871: Compare / Clone Identities
Add the ability in the dashboard to compare flags between two Identities and then copy/clone the flags from Identity A to B

novakzaballa:

## Changes

Add Core and Edge edpoints for clonning identity's flag states.

Integrations tests added

### DIFF
--- a/api/edge_api/identities/models.py
+++ b/api/edge_api/identities/models.py
@@ -252,7 +252,7 @@ class EdgeIdentity:
 
         # Clone identity_source's feature states to identity_target
         for feature_in_source in source_identity.feature_overrides:
-            feature_state_target: FeatureStateModel = FeatureStateModel(
+            feature_state_target = FeatureStateModel(
                 feature=feature_in_source.feature,
                 feature_state_value=feature_in_source.feature_state_value,
                 enabled=feature_in_source.enabled,
--- a/api/edge_api/identities/models.py
+++ b/api/edge_api/identities/models.py
@@ -62,16 +62,6 @@ class EdgeIdentity:
     def add_feature_override(self, feature_state: FeatureStateModel) -> None:
         self._engine_identity_model.identity_features.append(feature_state)
 
-    # Method that receives no parameters and returns only the overridden
-    # feature_states of an identity as a dictionary with the feature ID as the key
-    # and the overridden feature_state as the value.
-    def get_overridden_feature_states_dict(self) -> dict[str, FeatureStateModel]:
-        overridden_feature_states: dict[str, FeatureStateModel] = {}
-
-        for feature_state in self.feature_overrides:
-            overridden_feature_states[str(feature_state.feature.id)] = feature_state
-        return overridden_feature_states
-
     def get_all_feature_states(
         self,
     ) -> typing.Tuple[
@@ -256,26 +246,15 @@ class EdgeIdentity:
         """
         Clone the feature states from the source identity to the target identity.
         """
-        target_feature_states: dict[str, FeatureStateModel] = (
-            self.get_overridden_feature_states_dict()
-        )
-        source_feature_states: dict[str, FeatureStateModel] = (
-            source_identity.get_overridden_feature_states_dict()
-        )
-
         # Delete identity_target's feature states
-        for feature_state_id in target_feature_states:
-            self.remove_feature_override(
-                feature_state=target_feature_states[feature_state_id]
-            )
+        for feature_state in list(self.feature_overrides):
+            self.remove_feature_override(feature_state=feature_state)
 
         # Clone identity_source's feature states to identity_target
-        for feature_id_source in source_feature_states:
+        for feature_in_source in source_identity.feature_overrides:
             feature_state_target: FeatureStateModel = FeatureStateModel(
-                feature=source_feature_states[feature_id_source].feature,
-                feature_state_value=source_feature_states[
-                    feature_id_source
-                ].feature_state_value,
-                enabled=source_feature_states[feature_id_source].enabled,
+                feature=feature_in_source.feature,
+                feature_state_value=feature_in_source.feature_state_value,
+                enabled=feature_in_source.enabled,
             )
             self.add_feature_override(feature_state_target)
--- a/api/edge_api/identities/models.py
+++ b/api/edge_api/identities/models.py
@@ -252,7 +252,7 @@ class EdgeIdentity:
     def _reset_initial_state(self):
         self._initial_state = copy.deepcopy(self)
 
-    def clone_flag_states_from(self, source_identity: "EdgeIdentity"):
+    def clone_flag_states_from(self, source_identity: "EdgeIdentity") -> None:
         """
         Clone the feature states from the source identity to the target identity.
         """
@@ -279,5 +279,3 @@ class EdgeIdentity:
                 enabled=source_feature_states[feature_id_source].enabled,
             )
             self.add_feature_override(feature_state_target)
-
-        return list(target_feature_states.values())
--- a/api/edge_api/identities/models.py
+++ b/api/edge_api/identities/models.py
@@ -62,6 +62,16 @@ class EdgeIdentity:
     def add_feature_override(self, feature_state: FeatureStateModel) -> None:
         self._engine_identity_model.identity_features.append(feature_state)
 
+    # Method that receives no parameters and returns only the overridden
+    # feature_states of an identity as a dictionary with the feature ID as the key
+    # and the overridden feature_state as the value.
+    def get_overridden_feature_states_dict(self) -> dict[str, FeatureStateModel]:
+        overridden_feature_states: dict[str, FeatureStateModel] = {}
+
+        for feature_state in self.feature_overrides:
+            overridden_feature_states[str(feature_state.feature.id)] = feature_state
+        return overridden_feature_states
+
     def get_all_feature_states(
         self,
     ) -> typing.Tuple[
@@ -241,3 +251,33 @@ class EdgeIdentity:
 
     def _reset_initial_state(self):
         self._initial_state = copy.deepcopy(self)
+
+    def clone_flag_states_from(self, source_identity: "EdgeIdentity"):
+        """
+        Clone the feature states from the source identity to the target identity.
+        """
+        target_feature_states: dict[str, FeatureStateModel] = (
+            self.get_overridden_feature_states_dict()
+        )
+        source_feature_states: dict[str, FeatureStateModel] = (
+            source_identity.get_overridden_feature_states_dict()
+        )
+
+        # Delete identity_target's feature states
+        for feature_state_id in target_feature_states:
+            self.remove_feature_override(
+                feature_state=target_feature_states[feature_state_id]
+            )
+
+        # Clone identity_source's feature states to identity_target
+        for feature_id_source in source_feature_states:
+            feature_state_target: FeatureStateModel = FeatureStateModel(
+                feature=source_feature_states[feature_id_source].feature,
+                feature_state_value=source_feature_states[
+                    feature_id_source
+                ].feature_state_value,
+                enabled=source_feature_states[feature_id_source].enabled,
+            )
+            self.add_feature_override(feature_state_target)
+
+        return list(target_feature_states.values())
--- a/api/edge_api/identities/views.py
+++ b/api/edge_api/identities/views.py
@@ -307,6 +307,9 @@ class EdgeIdentityFeatureStateViewSet(viewsets.ModelViewSet):
 
     @action(detail=False, methods=["POST"], url_path="clone-from-given-identity")
     def clone_from_given_identity(self, request, *args, **kwargs) -> Response:
+        """
+        Clone feature states from a given source identity.
+        """
         source_identity: EdgeIdentity = self.get_identity(
             edge_identity_identity_uuid=request.data.get("source_identity_uuid")
         )
--- a/api/edge_api/identities/views.py
+++ b/api/edge_api/identities/views.py
@@ -305,8 +305,8 @@ class EdgeIdentityFeatureStateViewSet(viewsets.ModelViewSet):
 
         return Response(serializer.data)
 
-    @action(detail=False, methods=["POST"], url_path="clone-flag-states-from")
-    def clone_flag_states_from(self, request, *args, **kwargs) -> Response:
+    @action(detail=False, methods=["POST"], url_path="clone-from-given-identity")
+    def clone_from_given_identity(self, request, *args, **kwargs) -> Response:
         source_identity: EdgeIdentity = self.get_identity(
             edge_identity_identity_uuid=request.data.get("source_identity_uuid")
         )
--- a/api/edge_api/identities/views.py
+++ b/api/edge_api/identities/views.py
@@ -312,6 +312,12 @@ class EdgeIdentityFeatureStateViewSet(viewsets.ModelViewSet):
         """
         Clone feature states from a given source identity.
         """
+        # Get and validate source identity
+        serializer = EdgeIdentitySourceIdentityRequestSerializer(
+            data=request.data, context={"request": request}
+        )
+        serializer.is_valid(raise_exception=True)
+
         source_identity: EdgeIdentity = self.get_identity(
             edge_identity_identity_uuid=request.data.get("source_identity_uuid")
         )
--- a/api/edge_api/identities/views.py
+++ b/api/edge_api/identities/views.py
@@ -216,7 +216,7 @@ class EdgeIdentityFeatureStateViewSet(viewsets.ModelViewSet):
 
     def get_identity(self, edge_identity_identity_uuid) -> EdgeIdentity:
         identity_document = EdgeIdentity.dynamo_wrapper.get_item_from_uuid_or_404(
-            uuid=edge_identity_identity_uuid
+            edge_identity_identity_uuid
         )
 
         if (
--- a/api/edge_api/identities/views.py
+++ b/api/edge_api/identities/views.py
@@ -215,7 +215,7 @@ class EdgeIdentityFeatureStateViewSet(viewsets.ModelViewSet):
     ]
     pagination_class = None
 
-    def get_identity(self, edge_identity_identity_uuid) -> EdgeIdentity:
+    def get_identity(self, edge_identity_identity_uuid: str) -> EdgeIdentity:
         identity_document = EdgeIdentity.dynamo_wrapper.get_item_from_uuid_or_404(
             edge_identity_identity_uuid
         )
@@ -319,7 +319,9 @@ class EdgeIdentityFeatureStateViewSet(viewsets.ModelViewSet):
         serializer.is_valid(raise_exception=True)
 
         source_identity: EdgeIdentity = self.get_identity(
-            edge_identity_identity_uuid=request.data.get("source_identity_uuid")
+            edge_identity_identity_uuid=serializer.validated_data[
+                "source_identity_uuid"
+            ]
         )
 
         self.identity.clone_flag_states_from(source_identity)
--- a/api/edge_api/identities/views.py
+++ b/api/edge_api/identities/views.py
@@ -7,7 +7,7 @@ from boto3.dynamodb.conditions import Key
 from django.shortcuts import get_object_or_404
 from django.utils.decorators import method_decorator
 from drf_yasg.utils import swagger_auto_schema
-from flag_engine.identities.models import IdentityModel
+from flag_engine.identities.models import IdentityFeaturesList, IdentityModel
 from flag_engine.identities.traits.models import TraitModel
 from pyngo import drf_error_details
 from rest_framework import status, viewsets
@@ -202,7 +202,6 @@ class EdgeIdentityFeatureStateViewSet(viewsets.ModelViewSet):
     lookup_field = "featurestate_uuid"
 
     serializer_class = EdgeIdentityFeatureStateSerializer
-
     # Patch is not supported
     http_method_names = [
         "get",
@@ -215,10 +214,9 @@ class EdgeIdentityFeatureStateViewSet(viewsets.ModelViewSet):
     ]
     pagination_class = None
 
-    def initial(self, request, *args, **kwargs):
-        super().initial(request, *args, **kwargs)
+    def get_identity(self, edge_identity_identity_uuid) -> EdgeIdentity:
         identity_document = EdgeIdentity.dynamo_wrapper.get_item_from_uuid_or_404(
-            self.kwargs["edge_identity_identity_uuid"]
+            uuid=edge_identity_identity_uuid
         )
 
         if (
@@ -233,8 +231,15 @@ class EdgeIdentityFeatureStateViewSet(viewsets.ModelViewSet):
                 environment__api_key=identity.environment_api_key
             ).values_list("feature__name", flat=True)
         )
-        identity.synchronise_features(valid_feature_names)
-        self.identity = identity
+        identity.synchronise_features(valid_feature_names=valid_feature_names)
+
+        return identity
+
+    def initial(self, request, *args, **kwargs):
+        super().initial(request, *args, **kwargs)
+        self.identity: EdgeIdentity = self.get_identity(
+            edge_identity_identity_uuid=self.kwargs["edge_identity_identity_uuid"]
+        )
 
     def get_object(self):
         feature_state = self.identity.get_feature_state_by_featurestate_uuid(
@@ -261,7 +266,7 @@ class EdgeIdentityFeatureStateViewSet(viewsets.ModelViewSet):
         )
         q_params_serializer.is_valid(raise_exception=True)
 
-        identity_features = self.identity.feature_overrides
+        identity_features: IdentityFeaturesList = self.identity.feature_overrides
 
         feature = q_params_serializer.data.get("feature")
         if feature:
@@ -299,6 +304,19 @@ class EdgeIdentityFeatureStateViewSet(viewsets.ModelViewSet):
         )
 
         return Response(serializer.data)
+
+    @action(detail=False, methods=["POST"], url_path="clone-flag-states-from")
+    def clone_flag_states_from(self, request, *args, **kwargs) -> Response:
+        source_identity: EdgeIdentity = self.get_identity(
+            edge_identity_identity_uuid=request.data.get("source_identity_uuid")
+        )
+
+        self.identity.clone_flag_states_from(source_identity)
+        self.identity.save(
+            user=request.user.id,
+            master_api_key=getattr(request, "master_api_key", None),
+        )
+        return Response(status=status.HTTP_200_OK)
 
 
 class EdgeIdentityWithIdentifierFeatureStateView(APIView):
--- a/api/edge_api/identities/views.py
+++ b/api/edge_api/identities/views.py
@@ -306,7 +306,10 @@ class EdgeIdentityFeatureStateViewSet(viewsets.ModelViewSet):
 
         return Response(serializer.data)
 
-    @swagger_auto_schema(request_body=EdgeIdentitySourceIdentityRequestSerializer())
+    @swagger_auto_schema(
+        request_body=EdgeIdentitySourceIdentityRequestSerializer(),
+        responses={200: IdentityAllFeatureStatesSerializer(many=True)},
+    )
     @action(detail=False, methods=["POST"], url_path="clone-from-given-identity")
     def clone_from_given_identity(self, request, *args, **kwargs) -> Response:
         """
@@ -330,24 +333,7 @@ class EdgeIdentityFeatureStateViewSet(viewsets.ModelViewSet):
             master_api_key=getattr(request, "master_api_key", None),
         )
 
-        # Prepare response
-        (
-            feature_states,
-            identity_feature_names,
-        ) = self.identity.get_all_feature_states()
-
-        serializer = IdentityAllFeatureStatesSerializer(
-            instance=feature_states,
-            many=True,
-            context={
-                "request": request,
-                "identity": self.identity,
-                "environment_api_key": self.identity.environment_api_key,
-                "identity_feature_names": identity_feature_names,
-            },
-        )
-
-        return Response(serializer.data)
+        return self.all(request, *args, **kwargs)
 
 
 class EdgeIdentityWithIdentifierFeatureStateView(APIView):
--- a/api/edge_api/identities/views.py
+++ b/api/edge_api/identities/views.py
@@ -38,6 +38,7 @@ from edge_api.identities.serializers import (
     EdgeIdentityFsQueryparamSerializer,
     EdgeIdentityIdentifierSerializer,
     EdgeIdentitySerializer,
+    EdgeIdentitySourceIdentityRequestSerializer,
     EdgeIdentityTraitsSerializer,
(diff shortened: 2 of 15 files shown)


novakzaballa:
I am tempted to create one or a couple of fixtures that return all the assets used in the integration tests since they both, API Core and the EDdge API tests, are very similar, even the asserts are almost the same.

api/edge_api/identities/views.py
```diff
@@ -215,10 +215,9 @@ class EdgeIdentityFeatureStateViewSet(viewsets.ModelViewSet):
     ]
     pagination_class = None
 
-    def initial(self, request, *args, **kwargs):
-        super().initial(request, *args, **kwargs)
+    def get_identity(self, edge_identity_identity_uuid) -> EdgeIdentity:
```
matthewelwell:
```suggestion
    def get_identity(self, edge_identity_identity_uuid: str) -> EdgeIdentity:
```

api/edge_api/identities/views.py
```diff
@@ -300,6 +306,29 @@ def all(self, request, *args, **kwargs):
...
+        source_identity: EdgeIdentity = self.get_identity(
+            edge_identity_identity_uuid=request.data.get("source_identity_uuid")
+        )
+
+        self.identity.clone_flag_states_from(source_identity)
+        self.identity.save(
+            user=request.user.id,
+            master_api_key=getattr(request, "master_api_key", None),
+        )
+        return Response(status=status.HTTP_200_OK)
```
matthewelwell: Couldn't this return the identity to save another API request from the FE?

api/environments/identities/models.py
```diff
@@ -141,6 +141,32 @@ def get_all_feature_states(
 
         return list(identity_flags.values())
 
+    def get_overridden_feature_states(self) -> dict[str, FeatureState]:
```
matthewelwell: I don't _think_ this is returning a dict with string keys - it looks like integers to me?

api/environments/identities/models.py
```diff
@@ -141,6 +141,32 @@ def get_all_feature_states(
...
+            "identity",
+        ]
+
+        identity_flags: models.BaseManager[FeatureState] = (
+            FeatureState.objects.select_related(*select_related_args).filter(query)
+        )
+
+        for feature_state in identity_flags:
+            overridden_feature_states[feature_state.feature.id] = feature_state
+        return overridden_feature_states
```
matthewelwell: Small aesthetic improvement: 

```suggestion
        return {fs.feature.id: fs for fs in identity_flags}
```

api/environments/identities/models.py
```diff
@@ -141,6 +141,32 @@ def get_all_feature_states(
...
+    def get_overridden_feature_states(self) -> dict[str, FeatureState]:
+        """
+        Get all overridden feature states for an identity.
+
+        :return: dict[str, FeatureState] - Key: feature ID. Value: Overridden feature_state.
+        """
+        overridden_feature_states: dict[str, FeatureState] = {}
+
+        query = Q(environment=self.environment) & Q(identity=self)
+        query &= Q(live_from__lte=timezone.now(), version__isnull=False)
```
matthewelwell: We don't currently version identity overrides in any way, so I _think_ we can simplify this to just `self.identity_features.all()`, right?

api/features/models.py
```diff
@@ -947,6 +947,48 @@ def _is_more_recent_version(self, other: "FeatureState") -> bool:
...
+
+        # Delete own feature states not in source_identity
+        feature_states_to_delete = list(
+            target_feature_states.keys() - source_feature_states.keys()
+        )
+        for feature_state_id in feature_states_to_delete:
+            target_feature_states[feature_state_id].delete()
+
+        # Clone source_identity's feature states to target_identity
+        for feature_id_source in source_feature_states:
```
matthewelwell: I think you could simplify this code by using the following code to iterate over the keys and values in the dictionary 

```suggestion
        for feature_id, feature_state in source_feature_states.items():
```

api/features/views.py
```diff
@@ -636,6 +637,31 @@ def all(self, request, *args, **kwargs):
...
+        source_identity = get_object_or_404(
+            queryset=Identity, pk=request.data.get("source_identity_id")
+        )
+
+        # Clone feature states
+        FeatureState.copy_identity_feature_states(
+            target_identity=target_identity, source_identity=source_identity
+        )
+
+        return Response(status=status.HTTP_200_OK)
```
matthewelwell: Again, we should return some data here - the target identity and their overrides perhaps? I guess we want to return whatever data the FE needs to re-render the identity detail page.

api/tests/integration/conftest.py
```diff
@@ -100,7 +100,10 @@ def environment(
 
 @pytest.fixture()
 def dynamo_enabled_environment(
-    admin_client, dynamo_enabled_project, environment_api_key
+    admin_client,
+    dynamo_enabled_project,
+    environment_api_key,
+    app_settings_for_dynamodb,
```
matthewelwell: typing please

api/tests/conftest.py
```diff
@@ -90,7 +94,49 @@ def dynamo_enabled_project_environment_one_document(
...
+) -> None:
+    settings.ENVIRONMENTS_TABLE_NAME_DYNAMO = flagsmith_environment_table.name
+    settings.ENVIRONMENTS_V2_TABLE_NAME_DYNAMO = flagsmith_environments_v2_table.name
+    settings.IDENTITIES_TABLE_NAME_DYNAMO = flagsmith_identities_table.name
+    return
+
+
+@pytest.fixture()
+def features_for_identity_clone_flag_states_from() -> (
+    typing.Callable[..., tuple[Feature, Feature, Feature]]
```
matthewelwell:
```suggestion
    typing.Callable[[Project], tuple[Feature, Feature, Feature]]
```

api/tests/conftest.py
```diff
@@ -90,7 +94,49 @@ def dynamo_enabled_project_environment_one_document(
...
+    flagsmith_identities_table: Table,
+) -> None:
+    settings.ENVIRONMENTS_TABLE_NAME_DYNAMO = flagsmith_environment_table.name
+    settings.ENVIRONMENTS_V2_TABLE_NAME_DYNAMO = flagsmith_environments_v2_table.name
+    settings.IDENTITIES_TABLE_NAME_DYNAMO = flagsmith_identities_table.name
+    return
+
+
+@pytest.fixture()
+def features_for_identity_clone_flag_states_from() -> (
```
matthewelwell: I'm not sure this fixture is really necessary and even if it is, it shouldn't live in `api/tests/conftest` it should live in the most local conftest.py file possible. 

The code, however, can be simplified to just the following, so I'd argue that we don't really need the fixture perhaps? 

```python
features = []
for i in range(1, 4): 
    features.append(Feature.objects.create(name=f"feature_{i}", project=project, default_enabled=True))
```

api/tests/integration/edge_api/identities/conftest.py
```diff
@@ -1,8 +1,29 @@
...
+                args=(environment_api_key, "$identity_uuid"),
+            )
+        ),
+        "all": Template(
+            reverse(
+                "api-v1:environments:edge-identity-featurestates-all",
+                args=(environment_api_key, "$identity_uuid"),
+            )
+        ),
+    }
```
matthewelwell: This fixture feels like overkill to me

api/tests/integration/edge_api/identities/test_edge_identity_featurestates_viewset.py
```diff
@@ -1033,3 +1044,135 @@ def _create_segment_override(
...
+    # Add feature states for features 1 and 2 to source identity
+    source_identity.add_feature_override(feature_state=source_feature_state_1)
+    source_identity.add_feature_override(feature_state=source_feature_state_2)
+
+    # Add feature states for features 2 and 3 to target identity.
+    # This feature state must be deleted as part of the cloning process.
+    target_identity.add_feature_override(feature_state=target_feature_state_2)
+    target_identity.add_feature_override(feature_state=target_feature_state_3)
+
+    # Save identities to identities document
```
matthewelwell:
```suggestion
    # Save identities to identities table
```

matthewelwell (commented):
@novakzaballa in general, this PR is great! I'm really impressed with the quality of the code, and the tests. 

I've added quite a few comments, but they're all fairly minor and, realistically, none of them are real blockers - they're mostly just me being picky!

api/edge_api/identities/views.py
novakzaballa: Yeah, I was originally sending the new Identity feature states but in the FE that would require a lot of additional logic, currently, the API client we use (RTK Query) invalidates the state to build the view, and then calls the endpoint originally used to get the data view, the get all feature states, and while we could tweak it,  that's the RTK's philosophy.

api/edge_api/identities/views.py
novakzaballa: Suggestion applied, thankis!

api/environments/identities/models.py
novakzaballa: Right, thanks! My IDE suggests usually the right typing but it can't deal with django predefined fields like `id` yet.

api/edge_api/identities/models.py
```diff
@@ -241,3 +241,20 @@ def _get_changes(self) -> IdentityChangeset:
...
+        """
+        Clone the feature states from the source identity to the target identity.
+        """
+        # Delete identity_target's feature states
+        for feature_state in list(self.feature_overrides):
+            self.remove_feature_override(feature_state=feature_state)
+
+        # Clone identity_source's feature states to identity_target
+        for feature_in_source in source_identity.feature_overrides:
+            feature_state_target: FeatureStateModel = FeatureStateModel(
```
zachaysan: The typing here is completely redundant and it's easier to follow without it.
```suggestion
            feature_state_target = FeatureStateModel(
```

api/environments/identities/models.py
novakzaballa: Yeah, this is the right/nice pythonic way :) thanks!

api/environments/identities/models.py
novakzaballa: Yeah, Thanks! I was not absolutly sure so I based this query on the `get_all_feature_states` method's one. But yeah tests are runing well with this change. Nice one line method now I am wondering if it woth's it to keep it :)

api/features/models.py
novakzaballa: Right, thanks!

api/features/views.py
novakzaballa: This is answered in the same observation for the Core endpoint.

api/edge_api/identities/views.py
```diff
@@ -300,6 +306,29 @@ def all(self, request, *args, **kwargs):
...
+        """
+        # Get and validate source identity
+        serializer = EdgeIdentitySourceIdentityRequestSerializer(
+            data=request.data, context={"request": request}
+        )
+        serializer.is_valid(raise_exception=True)
+
+        source_identity: EdgeIdentity = self.get_identity(
+            edge_identity_identity_uuid=request.data.get("source_identity_uuid")
+        )
```
zachaysan: Nit: Use the serializer to get the `source_identity_uuid` since it's been validated.

api/edge_api/identities/views.py
zachaysan: It seems strange to me that RTK can't use data included in the response.

api/environments/identities/models.py
```diff
@@ -141,6 +141,32 @@ def get_all_feature_states(
 
         return list(identity_flags.values())
 
+    def get_overridden_feature_states(self) -> dict[str, FeatureState]:
+        """
+        Get all overridden feature states for an identity.
+
+        :return: dict[str, FeatureState] - Key: feature ID. Value: Overridden feature_state.
+        """
+        overridden_feature_states: dict[str, FeatureState] = {}
```
zachaysan: This part needs updating too.
```suggestion
        overridden_feature_states: dict[int, FeatureState] = {}
```

api/tests/conftest.py
novakzaballa: The reason it is a fixture and lives there is because we are using it in both tests, the API Core one and the Edge one. But I can write it in both. A lot of logic and assets are duplicated any way. BTM I am just optimizing the code as you suggested.

api/environments/identities/models.py
zachaysan: I think @matthewelwell means `identity.identity_features.all()` and I think it can simplify to that as well.

api/tests/integration/edge_api/identities/conftest.py
novakzaballa: It could simplify the tests if we use it everywhere, but for only this test ... it could be too much. Should I move it to the test?

api/features/models.py
```diff
@@ -947,6 +947,48 @@ def _is_more_recent_version(self, other: "FeatureState") -> bool:
...
+    @staticmethod
+    def copy_identity_feature_states(
+        target_identity: "Identity", source_identity: "Identity"
+    ) -> None:
+        target_feature_states: dict[str, FeatureState] = (
+            target_identity.get_overridden_feature_states()
+        )
+        source_feature_states: dict[str, FeatureState] = (
+            source_identity.get_overridden_feature_states()
+        )
```
zachaysan: I think these dicts have `int`s instead of `str`.

```suggestion
        target_feature_states: dict[int, FeatureState] = (
            target_identity.get_overridden_feature_states()
        )
        source_feature_states: dict[int, FeatureState] = (
            source_identity.get_overridden_feature_states()
        )
```

api/features/views.py
```diff
@@ -636,6 +637,31 @@ def all(self, request, *args, **kwargs):
 
         return Response(serializer.data)
 
+    @swagger_auto_schema(request_body=IdentitySourceIdentityRequestSerializer())
+    @action(methods=["POST"], detail=False, url_path="clone-from-given-identity")
+    def clone_from_given_identity(self, request, *args, **kwargs) -> Response:
```
zachaysan, line 645: nit: missing typing in the args

api/features/views.py
```diff
@@ -636,6 +637,31 @@ def all(self, request, *args, **kwargs):
...
+        serializer = IdentitySourceIdentityRequestSerializer(
+            data=request.data, context={"request": request}
+        )
+        serializer.is_valid(raise_exception=True)
+        # Get and validate source and target identities
+        target_identity = get_object_or_404(
+            queryset=Identity, pk=self.kwargs["identity_pk"]
+        )
+        source_identity = get_object_or_404(
+            queryset=Identity, pk=request.data.get("source_identity_id")
```
zachaysan, line 658: nit: use the serializer which has already been validated

api/tests/integration/edge_api/identities/test_edge_identity_featurestates_viewset.py
```diff
@@ -1033,3 +1044,135 @@ def _create_segment_override(
...
+    def create_identity(identifier: str) -> EdgeIdentity:
+        identity_model = IdentityModel(
+            identifier=identifier,
+            environment_api_key=environment_api_key,
+            identity_features=IdentityFeaturesList(),
+            identity_uuid=uuid.uuid4(),
+        )
+        return EdgeIdentity(engine_identity_model=identity_model)
+
+    # GIVEN
```
zachaysan:
```suggestion
    # Given
```

api/tests/integration/edge_api/identities/test_edge_identity_featurestates_viewset.py
```diff
@@ -1033,3 +1044,135 @@ def _create_segment_override(
...
+
+    project: Project = Project.objects.get(id=dynamo_enabled_project)
+
+    # Create 3 features
```
zachaysan: nit: This comment is unnecessary.

api/tests/integration/edge_api/identities/test_edge_identity_featurestates_viewset.py
```diff
@@ -1033,3 +1044,135 @@ def _create_segment_override(
...
+    target_identity.add_feature_override(feature_state=target_feature_state_3)
+
+    # Save identities to identities document
+    target_identity_document = target_identity.to_document()
+    source_identity_document = source_identity.to_document()
+
+    flagsmith_identities_table.put_item(Item=target_identity_document)
+    flagsmith_identities_table.put_item(Item=source_identity_document)
+
+    # WHEN
```
zachaysan:
```suggestion
    # When
```

api/tests/integration/edge_api/identities/test_edge_identity_featurestates_viewset.py
```diff
@@ -1033,3 +1044,135 @@ def _create_segment_override(
...
+    )
+
+    # Get current feature states for fixture target identity
+    get_feature_states_response = admin_client.get(
+        path=featurestates_urls["all"].substitute(
+            {"identity_uuid": target_identity.identity_uuid}
+        ),
+    )
+
+    # THEN
```
zachaysan:
```suggestion
    # Then
```

api/tests/integration/edge_api/identities/test_edge_identity_featurestates_viewset.py
```diff
@@ -1033,3 +1044,135 @@ def _create_segment_override(
...
+    assert response[0]["feature_state_value"] == source_feature_state_1_value
+    assert response[0]["overridden_by"] == "IDENTITY"
+
+    assert response[1]["feature"]["id"] == feature_2.id
+    assert response[1]["enabled"] == source_feature_state_2.enabled
+    assert response[1]["feature_state_value"] == source_feature_state_2_value
+    assert response[1]["overridden_by"] == "IDENTITY"
+
+    assert response[2]["feature"]["id"] == feature_3.id
+    assert response[2]["overridden_by"] is None
```
zachaysan: I'd add asserts for `enabled` and `feature_state_value` even if they're not part of this response even testing their non-inclusion would be useful.

api/tests/unit/environments/identities/test_unit_identities_feature_states_views.py
```diff
@@ -106,3 +112,121 @@ def test_user_with_view_environment_permission_can_retrieve_all_feature_states_f
...
     assert response.status_code == status.HTTP_200_OK
+
+
+def test_identity_clone_flag_states_from(
+    project: Project,
+    environment: Environment,
+    admin_client: Client,
+    features_for_identity_clone_flag_states_from: typing.Callable,
+) -> None:
+    # GIVEN
```
zachaysan:
```suggestion
    # Given
```

api/tests/unit/environments/identities/test_unit_identities_feature_states_views.py
```diff
@@ -106,3 +112,121 @@ def test_user_with_view_environment_permission_can_retrieve_all_feature_states_f
...
+
+    # Create 3 features
```
zachaysan: Again, I know you like to comment things as you go, but for really obvious stuff it's not worth the mental overhead.

api/tests/unit/environments/identities/test_unit_identities_feature_states_views.py
```diff
@@ -106,3 +112,121 @@ def test_user_with_view_environment_permission_can_retrieve_all_feature_states_f
...
+        "api-v1:environments:identity-featurestates-clone-from-given-identity",
+        args=[environment.api_key, target_identity.id],
+    )
+
+    get_feature_states_url = reverse(
+        "api-v1:environments:identity-featurestates-all",
+        args=[environment.api_key, target_identity.id],
+    )
+
+    # WHEN
```
zachaysan:
```suggestion
    # When
```

api/tests/unit/environments/identities/test_unit_identities_feature_states_views.py
```diff
@@ -106,3 +112,121 @@ def test_user_with_view_environment_permission_can_retrieve_all_feature_states_f
...
+    clone_identity_feature_states_response = admin_client.post(
+        clone_identity_feature_states_url,
+        data=json.dumps({"source_identity_id": str(source_identity.id)}),
+        content_type="application/json",
+    )
+
+    # And getting feature states for target identity
+    get_feature_states_response = admin_client.get(get_feature_states_url)
+
+    # THEN
```
zachaysan:
```suggestion
    # Then
```

api/tests/unit/environments/identities/test_unit_identities_feature_states_views.py
```diff
@@ -106,3 +112,121 @@ def test_user_with_view_environment_permission_can_retrieve_all_feature_states_f
...
+    assert response[0]["feature_state_value"] == source_feature_state_1_value
+    assert response[0]["overridden_by"] == "IDENTITY"
+
+    assert response[1]["feature"]["id"] == feature_2.id
+    assert response[1]["enabled"] == source_feature_state_2.enabled
+    assert response[1]["feature_state_value"] == source_feature_state_2_value
+    assert response[1]["overridden_by"] == "IDENTITY"
+
+    assert response[2]["feature"]["id"] == feature_3.id
+    assert response[2]["overridden_by"] is None
```
zachaysan: Again I'd add tests to this response record for `enabled` and `feature_state_value`

zachaysan (commented):
Overall the code looks well written. I've filed a bunch of nits, but overall this looks good. 

For the nits of changing `# GIVEN` to `# Given` (and similar) it's because I want the codebase to be consistent and we should keep to the same standard as we're currently using.

api/edge_api/identities/views.py
novakzaballa: It can, I just would need extra logic to merge the data coming from the original endpoint which includes much more data and this one, while in the RTKway you just invalidate the data being changed and it will make the request, update the view, or pick it from cache, or do whatever is needed automatically.

api/environments/identities/models.py
novakzaballa: I understood Matt's suggestion. thanks!

api/environments/identities/models.py
novakzaballa: Correct, my changes will include also that, thanks!

novakzaballa: Usually the suggestions breke code since are amde on the fly without an IDE so I apply them in my IDE checking what else is needed, and usually additional changes are needed :)

api/features/models.py
novakzaballa: Right. I will do this as part of applying Matt's suggestion

api/features/views.py
novakzaballa: I haven't seen our selves using typing in rest_framework handlers, in general I haven't seen that Python HTTP event handlers, even my IDE doesn't complaint about it, or including the unsed params args and kwargs.

api/features/views.py
novakzaballa: My IDE compliants about using the serialized.validated_data since it doesn't understand that, as you mentioned, they were alreay validated, but precily because of that we can use any of both ways right?

api/tests/integration/edge_api/identities/test_edge_identity_featurestates_viewset.py
novakzaballa: I prefer to use the capitalized version since I have a lot comments and that helps to identiy the sections more easily

api/tests/unit/environments/identities/test_unit_identities_feature_states_views.py
novakzaballa: That is not part of this changes and below we verify that the feature state has been removed.

api/tests/unit/environments/identities/test_unit_identities_feature_states_views.py
novakzaballa: I don't think 3 words can cause a mental overhead :)

api/tests/integration/edge_api/identities/test_edge_identity_featurestates_viewset.py
novakzaballa: Definitely not part of the scope of the functionality being tested as far as it is not an identity override is OK, right?

api/tests/integration/edge_api/identities/conftest.py
matthewelwell: Yes, let's move it to the test for now.

api/tests/integration/edge_api/identities/test_edge_identity_featurestates_viewset.py
novakzaballa: Added anyway since in this case is very cheap in all senses.

api/tests/integration/edge_api/identities/test_edge_identity_featurestates_viewset.py
novakzaballa: I need it :)

api/edge_api/identities/views.py
matthewelwell: Ok, I'm not sure that necessarily means that the view shouldn't return the data though - other direct API users might want to see that data.

api/environments/identities/models.py
matthewelwell: Sorry, just realised that this should also be optimised slightly, as follows:  

```python
return {fs.feature_id: fs for fs in identity_flags}
```

api/features/views.py
matthewelwell: As per other conversation, I still think we should return the data.

api/tests/unit/environments/identities/test_unit_identities_feature_states_views.py
matthewelwell: It's not the mental overhead so much as the risk of it becoming stale. IMO we should be stringent with comments and use them only when really necessary because it's additional code that needs to be kept up to date when the actual test code changes.

api/tests/integration/edge_api/identities/test_edge_identity_featurestates_viewset.py
matthewelwell: I think we can resolve all of these. I tend to agree that I like the consistency of using sentence case, but I don't think it's worth the back and forth here. @zachaysan unless you feel very strongly about this, can you resolve each of these comments?

api/tests/integration/edge_api/identities/test_edge_identity_featurestates_viewset.py
matthewelwell: Unless you can provide a good reason why you need it @novakzaballa , I agree with Zach that it should be removed.

api/tests/conftest.py
matthewelwell: Yes, I'd like to remove this fixture - it's too specific to the needs of the given tests. Please localise it or write it into the tests.

matthewelwell:
@novakzaballa @zachaysan can you make sure all outstanding conversations on this PR are responded to or resolved please?

api/tests/unit/environments/identities/test_unit_identities_feature_states_views.py
novakzaballa: I removed most of the inline comments so now I am also using Title case for Given, When, Then

api/edge_api/identities/views.py
novakzaballa: Right. Returning the identity feature flags again.

api/features/views.py
novakzaballa: Right. Data being returned again.

api/tests/conftest.py
novakzaballa: Done

api/tests/integration/edge_api/identities/test_edge_identity_featurestates_viewset.py
novakzaballa: Originally it was marking a section with several lines for readibility, but now that is one sentence in a few lines, it is removed.

api/tests/integration/edge_api/identities/test_edge_identity_featurestates_viewset.py
novakzaballa: I removed some of the inline comments and applied Zack suggestions.

api/environments/identities/models.py
novakzaballa: Yeah, I noticed that while addressing your first suggestion, now it is a single-line function.

api/edge_api/identities/views.py
```diff
@@ -300,6 +306,49 @@ def all(self, request, *args, **kwargs):
...
+            ]
+        )
+
+        self.identity.clone_flag_states_from(source_identity)
+        self.identity.save(
+            user=request.user.id,
+            master_api_key=getattr(request, "master_api_key", None),
+        )
+
+        # Prepare response
```
matthewelwell: Couldn't we just do `return self.all(request, *args, **kwargs)` ? Also, bear in mind that the documentation for this endpoint needs updating as well!

api/features/views.py
```diff
@@ -636,6 +637,43 @@ def all(self, request, *args, **kwargs):
...
+        )
+
+        # Clone feature states
+        FeatureState.copy_identity_feature_states(
+            target_identity=target_identity, source_identity=source_identity
+        )
+
+        # Prepare response
```
matthewelwell: Again, can we use `return self.all(request, *args, **kwargs)` here?

api/environments/identities/models.py
matthewelwell: No, my point is that we should use `feature_id` instead of `feature.id` this will provide a small performance optimisation. I'll add a suggestion to the actual line in another comment to make it clear.

api/environments/identities/models.py
```diff
@@ -141,6 +141,15 @@ def get_all_feature_states(
...
         return list(identity_flags.values())
 
+    def get_overridden_feature_states(self) -> dict[int, FeatureState]:
+        """
+        Get all overridden feature states for an identity.
+
+        :return: dict[int, FeatureState] - Key: feature ID. Value: Overridden feature_state.
+        """
+
+        return {fs.feature.id: fs for fs in self.identity_features.all()}
```
matthewelwell:
```suggestion
        return {fs.feature_id: fs for fs in self.identity_features.all()}
```

Related thread: https://github.com/Flagsmith/flagsmith/pull/3773#discussion_r1568640624

api/features/views.py
novakzaballa: Right. Done. Also updated the autogenerate swagger command

api/features/views.py
zachaysan: There's a ton of code that sets `request: Request` and our orders from @matthewelwell is to always set the typing even in places that seem silly, like typing in test fixtures.

api/features/views.py
zachaysan: It's the right way of doing it. When people refactor code they may easily miss accessing data directly but with the moved code it could be done now *before* validation. To keep things safe, it's best to use the validated data directly.

api/environments/identities/models.py
novakzaballa: Right, we don't need to load the whole related feature but just the FK which is already in the feature_sate data.
Done

zachaysan (approved):
I responded to your comments but they're just a suggestion at this point. Overall this looks good.

matthewelwell (approved):

PR #3790, opened 2024-04-17
feat: versioned segment override change request

### LINKED ISSUE #1204: Change requests for changes to segment overrides
Change requests currently cannot be used for changes to segment overrides. 

This means that any user with permissions to `update feature state` can change segment overrides without creating a change request. As a result, we must heavily restrict who has `update feature state` permissions in production, and therefore we are heavily restricting who can create change requests.

It would be awesome if changes to segment overrides could also be done through change requests.

kyle-ssg:

## Changes

- Environments with versioning + Change Requests enabled will now have the ability to create a change request for segment overrides
image

- Change requests now diff all of the feature states in comparison to live, irrespective of if versioning is enabled. 

Much like versioning this will tell you changed, created and unchanged values 

image

image

image

- The create feature modal toast will now differentiate when a change request was created vs editing a feature. Creating a change request will close the create feature modal

- Removed a confusing unused field "hide_from_client" from feature states

- Note: versioned change requests cannot be edited due to the limited use case and complexity of updating a versions feature states in a change request. 

- Created, edited and removed segment overrides in a change request on an environment with versioning


frontend/common/services/useFeatureState.ts
```diff
@@ -0,0 +1,79 @@
...
+          const {
+            data,
+          }: {
+            data: PagedResponse<
+              Omit<FeatureState, 'feature_segment'> & {
+                feature_segment: number | null
+              }
+            >
+          } = await baseQuery({
+            url: `features/featurestates/?${Utils.toParam(query)}`,
```
matthewelwell, line 40: I don't _think_ GET requests will be an issue here, but it's probably safest to add the trailing slash here for when this endpoint is used on the combined image. 

```suggestion
            url: `features/featurestates/?${Utils.toParam(query)}/`,
```

matthewelwell (approved):
There's an outstanding bug in the API which is resolved [here](https://github.com/Flagsmith/flagsmith/pull/3850) but this PR seems sound. 

I've also added one minor suggestion which we should probably add before merging.

matthewelwell:
@kyle-ssg I've just noticed that the change request view doesn't actually show which segment a feature is being overridden by? 

image

matthewelwell:
Also, is there anything we can do to improve the view when you remove a segment override? I'm just seeing this

image

kyle-ssg:
> Also, is there anything we can do to improve the view when you remove a segment override? I'm just seeing this
> 
> image

I've resolved the first point however cannot replicate the second, I see the following : 

image

kyle-ssg:
@matthewelwell both issues should be resolved now

matthewelwell:
@kyle-ssg when you create a CR with more than one segment override, the priority values don't seem right. 

image

matthewelwell:
@kyle-ssg I can't create any change requests now with segment overrides following the latest change: 

image

matthewelwell (approved):

kyle-ssg:
Related #1204

rolodato:
@kyle-ssg Did we decide to explicitly exclude scheduling changes on segment overrides on environments without change requests enabled?

Editing a feature's state without change requests enabled let you schedule changes:

image

In comparison, editing segment overrides without change requests only let you make an immediate update:

image

PR #3792, opened 2024-04-17
chore: Improve enable versioning and tests

kyle-ssg:

## Changes

- Versioning tests would sometimes fail due to enabling versioning not yet taking effect, the tests will now recursively wait and refresh
- The environment settings page will poll for changes in this setting and update the environment if it detects it

https://github.com/Flagsmith/flagsmith/assets/8608314/c5be16ef-f6cc-44a2-af55-26ec0e0cbb9e

### DIFF
--- a/frontend/e2e/tests/versioning-tests.ts
+++ b/frontend/e2e/tests/versioning-tests.ts
@@ -8,10 +8,10 @@ import {
     createRemoteConfig,
     editRemoteConfig,
     log,
-    login,
+    login, refreshUntilElementVisible,
     waitAndRefresh,
-    waitForElementVisible,
-} from '../helpers.cafe';
+    waitForElementVisible
+} from "../helpers.cafe";
 import { E2E_USER, PASSWORD } from '../config';
 import fetch from 'node-fetch';
 import flagsmith from 'flagsmith/isomorphic';
@@ -32,7 +32,8 @@ export default async () => {
     await click('#env-settings-link')
     await click(byId('enable-versioning'))
     await click('#confirm-toggle-feature-btn')
-    await waitAndRefresh(10000)
+    await waitAndRefresh(1000)
+    await refreshUntilElementVisible(byId('feature-versioning-enabled'))
 
     log('Create feature 1')
     await createRemoteConfig(0, 'a', 'small')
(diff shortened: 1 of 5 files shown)


frontend/e2e/tests/versioning-tests.ts
```diff
@@ -31,8 +31,9 @@ export default async () => {
     await waitForElementVisible(byId('features-page'))
     await click('#env-settings-link')
     await click(byId('enable-versioning'))
-    await click('#confirm-toggle-feature-btn')
-    await waitAndRefresh(10000)
+    await click('#confirm-btn-yes')
+    await waitAndRefresh(1000)
+    await refreshUntilElementVisible(byId('feature-versioning-enabled'))
```
matthewelwell: Do we need both of these still?

frontend/web/components/EnvironmentVersioningListener.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+const EnvironmentVersioningListener: FC<EnvironmentVersioningListenerType> = ({
+  id,
+  onChange,
+  versioningEnabled,
+}) => {
+  const { data, isFetching, isSuccess } = useGetEnvironmentQuery(
+    { id: `${id}` },
+    {
+      pollingInterval: versioningEnabled ? undefined : 2000,
```
matthewelwell, line 18: This ternary feels like it's the wrong way around to me 🤔 ? Aren't we trying to enable polling if versioning is enabled? If so, shouldn't we define the polling interval in this case?

kyle-ssg: No, we're polling if it's currently disabled from an API standpoint, this will then cover a user staying on the page vs revisiting / refreshing. It polls until it's enabled then stopping since it's irreversible.

frontend/e2e/tests/versioning-tests.ts
kyle-ssg: Actually yeah since we have the FE change we can wait without refreshing at all. I'll keep the refreshUntilElementVisible util since I think we can use it in future.

novakzaballa (approved):

PR #3795, opened 2024-04-18
feat: Capability for Pydantic-based OpenAPI response schemas

khvn26:

## Changes

This PR solves incorrect rendering of `Optional` fields of the environment document Swagger spec, and adds a generalised way to use Pydantic models to describe API responses with `swagger_auto_schema`.

 - Add `exclude_model_fields` utility function
 - Add `PydanticResponseCapableSwaggerAutoSchema` drf-yasg view inspector
 - Fix invalid nullable field schema generation

- Observed correct Swagger schema locally.
- Added a unit test.

### DIFF
--- a/api/api/openapi.py
+++ b/api/api/openapi.py
@@ -33,6 +33,27 @@ class _GenerateJsonSchema(GenerateJsonSchema):
 
 
 class PydanticResponseCapableSwaggerAutoSchema(SwaggerAutoSchema):
+    """
+    A `SwaggerAutoSchema` subclass that allows to generate view response Swagger docs
+    from a Pydantic model.
+
+    Example usage:
+
+    ```
+    @drf_yasg.utils.swagger_auto_schema(
+        responses={200: YourPydanticSchema},
+        auto_schema=PydanticResponseCapableSwaggerAutoSchema,
+    )
+    def your_view(): ...
+    ```
+
+    To adapt Pydantic-generated schemas, the following is taken care of:
+
+    1. Pydantic-generated definitions are unwrapped and added to drf-yasg's global definitions.
+    2. Rather than using `anyOf`, nullable fields are annotated with `x-nullable`.
+    3. As there's no way to annotate a reference, all nested models are assumed to be `x-nullable`.
+    """
+
     def get_response_schemas(
         self,
         response_serializers: dict[str | int, Any],
--- a/api/api/openapi.py
+++ b/api/api/openapi.py
@@ -36,7 +36,7 @@ class PydanticResponseCapableSwaggerAutoSchema(SwaggerAutoSchema):
     ) -> dict[str, Response]:
         result = {}
 
-        for status_code in response_serializers:
+        for status_code in list(response_serializers):
             if isinstance(response_serializers[status_code], type) and issubclass(
                 model_cls := response_serializers[status_code], BaseModel
             ):
--- a/api/api/openapi.py
+++ b/api/api/openapi.py
@@ -11,6 +11,9 @@ class _GenerateJsonSchema(GenerateJsonSchema):
     def nullable_schema(self, schema: core_schema.NullableSchema) -> JsonSchemaValue:
         """Generates an OpenAPI 2.0-compatible JSON schema that matches a schema that allows null values.
 
+        (The catch is OpenAPI 2.0 does not allow them, but some clients are capable
+        to consume the `x-nullable` annotation.)
+
         Args:
             schema: The core schema.
 
@@ -36,6 +39,8 @@ class PydanticResponseCapableSwaggerAutoSchema(SwaggerAutoSchema):
     ) -> dict[str, Response]:
         result = {}
 
+        definitions = self.components.with_scope(SCHEMA_DEFINITIONS)
+
         for status_code in list(response_serializers):
             if isinstance(response_serializers[status_code], type) and issubclass(
                 model_cls := response_serializers[status_code], BaseModel
@@ -46,7 +51,6 @@ class PydanticResponseCapableSwaggerAutoSchema(SwaggerAutoSchema):
                     ref_template=f"#/{SCHEMA_DEFINITIONS}/{{model}}",
                 )
 
-                definitions = self.components.with_scope(SCHEMA_DEFINITIONS)
                 for ref_name, schema_kwargs in model_json_schema.pop("$defs").items():
                     definitions.setdefault(
                         ref_name,
--- a/api/api/openapi.py
+++ b/api/api/openapi.py
@@ -1,38 +1,68 @@
-from copy import deepcopy
-from functools import lru_cache
+from typing import Any
 
-import jsonref
-from drf_yasg.openapi import Schema
-from flag_engine.environments.models import EnvironmentModel
-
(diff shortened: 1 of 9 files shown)


api/util/pydantic.py
```diff
@@ -0,0 +1,21 @@
+from pydantic import BaseModel, create_model
+
+
+def exclude_model_fields(
+    model_cls: type[BaseModel],
+    *exclude_fields: str,
```
matthewelwell, line 6: Feels like it might be more extensible in the future if `exclude_fields` was a single argument instead? e.g. `exclude_fields: typing.List[str]`

api/api/openapi.py
```diff
@@ -1,38 +1,68 @@
...
+from drf_yasg.inspectors import SwaggerAutoSchema
+from drf_yasg.openapi import SCHEMA_DEFINITIONS, Response, Schema
+from pydantic import BaseModel
+from pydantic.json_schema import GenerateJsonSchema, JsonSchemaValue
+from pydantic_core import core_schema
+
+
+class _GenerateJsonSchema(GenerateJsonSchema):
+    def nullable_schema(self, schema: core_schema.NullableSchema) -> JsonSchemaValue:
+        """Generates an OpenAPI 2.0-compatible JSON schema that matches a schema that allows null values.
```
matthewelwell, line 13: It would be good to explain here (or in an inline comment), what change is needed to make it OpenAPI 2.0 compatible.

api/api/openapi.py
```diff
@@ -1,38 +1,68 @@
...
+
+
+class PydanticResponseCapableSwaggerAutoSchema(SwaggerAutoSchema):
+    def get_response_schemas(
+        self,
+        response_serializers: dict[str | int, Any],
+    ) -> dict[str, Response]:
+        result = {}
+
+        for status_code in list(response_serializers):
```
matthewelwell, line 66: I'm not sure why we elected not to iterate over `response.serializers.items()` here?

api/api/openapi.py
```diff
@@ -1,38 +1,68 @@
...
+            if isinstance(response_serializers[status_code], type) and issubclass(
+                model_cls := response_serializers[status_code], BaseModel
+            ):
+                model_json_schema = model_cls.model_json_schema(
+                    mode="serialization",
+                    schema_generator=_GenerateJsonSchema,
+                    ref_template=f"#/{SCHEMA_DEFINITIONS}/{{model}}",
+                )
+
+                definitions = self.components.with_scope(SCHEMA_DEFINITIONS)
```
matthewelwell: This looks like it doesn't need to be called inside the loop?

api/api/openapi.py
```diff
@@ -1,38 +1,68 @@
...
+                for ref_name, schema_kwargs in model_json_schema.pop("$defs").items():
+                    definitions.setdefault(
+                        ref_name,
+                        maker=lambda: Schema(
+                            **schema_kwargs,
+                            # We can not annotate references with `x-nullable`,
+                            # So just assume all nested models as nullable for now.
```
matthewelwell, line 82: This is all quite abstract so I'm struggling to understand exactly where this limitation exists? I think this means that whenever we define schema in a swagger response schema definition as a pydantic model that has a nested pydantic model, it will always be shown as nullable regardless of the type definition?

matthewelwell (commented):
Added some general comments and also tested the swagger auto generation by running the code locally and viewing the generated docs.

api/api/openapi.py
khvn26: We're mutating the `response_serializers` during iteration.

api/api/openapi.py
khvn26: > exactly where this limitation exists?

Any properties sibling to `$ref` are ignored according to [OpenAPI spec](https://swagger.io/docs/specification/using-ref/) (see "$ref Syntax").

> it will always be shown as nullable regardless of the type definition

Correct. This, IMO, is fine for response schemas.

api/util/pydantic.py
khvn26: Extending this crutch? Not on my watch.

In a better state of things we'll simply use separately defined schemas instead.

api/api/openapi.py
khvn26: Added more context.

api/api/openapi.py
khvn26: Good call, moved this.

api/api/openapi.py
```diff
@@ -1,38 +1,72 @@
...
+            for any_of in anyof_schema_value["anyOf"]
+            if any_of.get("type") != "null"
+        )
+        if type := elem.get("type"):
+            return {"type": type, "x-nullable": True}
+        # Assuming a reference here (which we can not annotate)
+        return elem
+
+
+class PydanticResponseCapableSwaggerAutoSchema(SwaggerAutoSchema):
```
matthewelwell, line 36: Since this is the entry point to this logic for anyone coming across the class reference in the settings file, I think it would be good to add a docstring to this class with a short summary explaining why we're overriding the default behaviour.

api/util/pydantic.py
matthewelwell: Ok, fair enough.

api/api/openapi.py
khvn26: Done.

api/api/openapi.py
```diff
@@ -1,38 +1,93 @@
...
+    def get_response_schemas(
+        self,
+        response_serializers: dict[str | int, Any],
+    ) -> dict[str, Response]:
+        result = {}
+
+        definitions = self.components.with_scope(SCHEMA_DEFINITIONS)
+
+        for status_code in list(response_serializers):
+            if isinstance(response_serializers[status_code], type) and issubclass(
```
zachaysan: I believe this if check will always fail since `type` is being passed in as a built-in function.

zachaysan (approved):
The rest of it looks 👌🏼 but I'm concerned about `type` not being set as intentioned. Assuming that is fixed feel free to merge this PR.

api/api/openapi.py
khvn26: `type` is, in fact, a class:

```python
>>> type
<class 'type'>
>>> type.__mro__
(<class 'type'>, <class 'object'>)
```

So the instance check works as intended:


```python
>>> obj = object()
>>> isinstance(obj, type)
False
>>> class Class: pass
... 
>>> isinstance(Class, type)
True
```

Additionally, you can see the [coverage report](https://app.codecov.io/gh/Flagsmith/flagsmith/pull/3795?src=pr&el=tree&filepath=api%2Fapi%2Fopenapi.py&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Flagsmith#07cb16953ed0a71679bd302276251914-R69) to make sure the code path is being run.

zachaysan: Well, it's good to see that `type` is set as intentioned, but it's a surprising check that at least warrants a comment so that confused developers are set right.

khvn26: Changed to `inspect.isclass` for more clarity.

PR #3808, opened 2024-04-18
feat: Show organisation name in header

rolodato:

## Changes

Always show the organisation name in the header:

Screenshot 2024-04-18 at 15-38-26 Flagsmith - Feature Flags Feature Toggles and Remote Config

This is what it looks like with very long organisation names:

Screenshot 2024-04-18 at 17 35 40

This ensures users can quickly check which organisation they're working in. This is especially important for self-hosting users and Flagsmith support working within customer organisations.

It also hopefully makes this navigation element more evident as a click target - it's not immediately obvious that "Organisation" is clickable, I've seen a few users miss this.

There's probably a smarter way to implement this, but this was the least offense/effort/risk solution I came up with.

Manually, see screenshots.

### DIFF
--- a/frontend/web/components/App.js
+++ b/frontend/web/components/App.js
@@ -361,7 +361,7 @@ const App = class extends Component {
                                         fill='#9DA4AE'
                                       />
                                     </span>
-                                    {'Organisation'}
+                                    Organisation <strong>{AccountStore.getOrganisation()?.name}</strong>
                                   </NavLink>
                                 </Row>
                                 <Row>
--- a/frontend/web/styles/project/_project-nav.scss
+++ b/frontend/web/styles/project/_project-nav.scss
@@ -128,6 +128,10 @@
   }
 }
 .nav-link {
+  text-wrap: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  max-width: 30em;
   img {
     max-height: 21px;
   }


kyle-ssg:
I agree it's a bit confusing, to be honest this is the one area I find not quite right about the navigation redesign but I get the rationale was to reduce links in the header.

Having the org name next to the word organisation feels like we're masking something that was just unintuitive to begin with. Do you think that maybe we should consider a breadcrumb here? 

As a rough thought, something like this

image

kyle-ssg:
I remember you mentioned this being confusing in a demo @agregoryfs, might be good to get your opinion.

rolodato:
I agree there's much better ways to implement this - I submitted this as a low effort stop-gap solution to hopefully prevent someone messing up the wrong organisation until we get a proper design in place. It's a step in the right direction with low risk that we can easily revisit later on. I'll let you guys decide if we want to ship this as-is or wait until we have a better design in place.

kyle-ssg:
Yeah understood, ok let's do it.

@dabeeeenster I wonder if we should get a design prototyped that does something like the above. I do think there are still improvements we should make. The more I compare our navigation to GH (which has very similar hierarchy) the more I think we should bring it closer to that.

kyle-ssg (approved):

PR #3815, opened 2024-04-20
feat: Add Pytest CI mode to optimise migrations

khvn26:

## Changes

This PR introduces a "CI mode" to Core API test runner, invoked by providing the `--ci` option to pytest.

In this mode, the following optimisations are applied to how pytest manages Django databases when in xdist mode:
 - Databases for workers are created using Django's native cloning method. This means that the migrations are run only once, and all the subsequent databases are copied using `CREATE DATABASE ... WITH TEMPLATE` query.
 - Database setup is only run on the controller node outside of the test session.
 - Database teardown is completely skipped.

Locally, this yields a substantial decrease in overall running time:

 ```sh
make test  708.02s user 34.22s system 216% cpu 5:43.19 total
make test opts='--ci'  558.61s user 29.81s system 255% cpu 3:50.14 total
```

I suspect it won't be as prominent in CI as it probably improves linearly with the number of workers (which is going to be less in CI).

The only proper way was to run the whole test suite.

### CODE REVIEW

File: api/conftest.py [hunk 1]
```diff
@@ -63,6 +67,51 @@
...
+        default=False,
+        help="Enable CI mode",
+    )
+
+
+@pytest.hookimpl(trylast=True)
+def pytest_configure(config: pytest.Config) -> None:
+    if (
+        config.option.ci
+        and config.option.dist != "no"
```
zachaysan, line 86: I'm surprised that the option's value comes out as `"no"` instead of `False`.

api/conftest.py
```diff
@@ -63,6 +67,51 @@
...
+        and not hasattr(config, "workerinput")
```
zachaysan, line 87: It seems weird that if `ci` is present that if `workerinput` is set that this will silently fail. Don't we prefer an exception if `ci` is set and so is `workerinput`?

api/conftest.py
```diff
@@ -63,6 +67,51 @@
 )
 from users.models import FFAdminUser, UserPermissionGroup
 
+
+def pytest_addoption(parser: pytest.Parser) -> None:
+    parser.addoption(
```
zachaysan, line 74: Just confirming that this function will be auto-invoked without a decorator.

api/conftest.py
```diff
@@ -63,6 +67,51 @@
...
+                interactive=False,
+                parallel=config.option.numprocesses,
+            )
+
+
+@pytest.fixture(scope="session")
+def django_db_setup(request: pytest.FixtureRequest) -> None:
+    if (
+        request.config.option.ci
+        and (xdist_worker_id_suffix := get_xdist_worker_id(request)[2:]).isnumeric()
```
zachaysan, line 102: It's not obvious why 

```python
get_xdist_worker_id(request)[2:]
```

That part of the logic ends with a `[2:]`. Can you add a comment?

zachaysan (approved):
Approved with some suggestions and questions, but overall looks useful 👍🏼

api/conftest.py
khvn26: See coverage report :)

Dive into [pytest hooks](https://docs.pytest.org/en/7.1.x/how-to/writing_hook_functions.html) if you're curious about how this works.

api/conftest.py
khvn26: `"no"` is, in fact, one of xdist's distribution modes. See [here](https://pytest-xdist.readthedocs.io/en/latest/distribution.html#running-tests-across-multiple-cpus).

api/conftest.py
khvn26: Having `ci` and having `workerinput` is a legitimate case when we need this hook to be a noop.

api/conftest.py
khvn26: Added.

PR #3822, opened 2024-04-22
chore: Convert features views APITestCase and TestCase to a normal test function

zachaysan:

## Changes


api/tests/unit/features/test_unit_features_views.py
```diff
@@ -62,572 +55,569 @@
 one_hour_ago = now - timedelta(hours=1)
 
 
-@pytest.mark.django_db
-class ProjectFeatureTestCase(TestCase):
-    project_features_url = "/api/v1/projects/%s/features/"
-    project_feature_detail_url = "/api/v1/projects/%s/features/%d/"
-    post_template = '{ "name": "%s", "project": %d, "initial_value": "%s" }'
+def test_project_owners_is_read_only_for_feature_create(
```
gagantrivedi, line 59: Would it be possible to take this opportunity to add `admin_master_api_key_client` to every test? We have some examples here: https://github.com/Flagsmith/flagsmith/blob/main/api/tests/unit/segments/test_unit_segments_views.py

gagantrivedi (changes requested):
Can we take this optimity to add

api/tests/unit/features/test_unit_features_views.py
zachaysan: Since we use that pattern so often (and since the default really should be to include `admin_master_api_key_client` into tests easily) I've created a new fixture that automatically cycles between the two clients for us without having to create them with pytest. 

The new default for using the `admin_client` fixture is now `admin_client_new`. Once I've had some time to change the tests `admin_client` will eventually take over from `admin_client_new`. 

I've talked with @matthewelwell about this and he's on board with the change.

gagantrivedi: Good idea :bulb:

gagantrivedi (approved):

PR #3824, opened 2024-04-23
docs(footer): add .NET guide link

matthewelwell:
Thanks for submitting a PR! Please check the boxes below:

- [x] I have run [`pre-commit`](https://docs.flagsmith.com/platform/contributing#pre-commit) to check linting
- [x] I have added information to `docs/` if required so people know about the feature!
- [x] I have filled in the "Changes" section below?
- [x] I have filled in the "How did you test this code" section below?
- [x] I have used a [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) title for this Pull Request

## Changes

Add link to the .NET how-to guide to the docs footer. 

## How did you test this code?

Ran docs locally and confirmed link was added correctly.

PR #3828, opened 2024-04-23
chore: fix anchor link

dabeeeenster:


rolodato (approved):

matthewelwell:
I think the workflow failure here is probably the same one that I fixed in my branch [here](https://github.com/Flagsmith/flagsmith/pull/3824). We should merge that one first and rebase here.

PR #3830, opened 2024-04-23
chore(saas/hubspot): create contacts with default domain

matthewelwell:

## Changes

Add logic to create a lead in Hubspot when a user signs up but doesn't create or join an organisation. 

Wrote some unit tests, as well as testing the hubspot API directly where needed.,

### DIFF
--- a/api/integrations/lead_tracking/hubspot/client.py
+++ b/api/integrations/lead_tracking/hubspot/client.py
@@ -15,9 +15,9 @@ logger = logging.getLogger(__name__)
 
 
 class HubspotClient:
-    def __init__(self) -> None:
+    def __init__(self, client: hubspot.Client = None) -> None:
         access_token = settings.HUBSPOT_ACCESS_TOKEN
-        self.client = hubspot.Client.create(access_token=access_token)
+        self.client = client or hubspot.Client.create(access_token=access_token)
 
     def get_contact(self, user: FFAdminUser) -> None | dict:
         public_object_id = BatchReadInputSimplePublicObjectId(
@@ -68,7 +68,11 @@ class HubspotClient:
         )
         return response.to_dict()
 
-    def get_company(self, domain: str) -> dict | None:
+    def get_company_by_domain(self, domain: str) -> dict | None:
+        """
+        Domain should be unique in Hubspot by design, so we should only ever have
+        0 or 1 results.
+        """
         public_object_search_request = PublicObjectSearchRequest(
             filter_groups=[
                 {
@@ -79,7 +83,7 @@ class HubspotClient:
             ]
         )
 
-        response = self.client.crm.contacts.batch_api.read(
+        response = self.client.crm.companies.search_api.do_search(
             public_object_search_request=public_object_search_request,
         )
 
@@ -109,8 +113,10 @@ class HubspotClient:
             properties["domain"] = domain
         if active_subscription:
             properties["active_subscription"] = active_subscription
-        if organisation_id:
-            properties["orgid"] = organisation_id
+
+        # hubspot doesn't allow null values for numeric fields, so we
+        # set this to -1 for auto generated organisations.
+        properties["orgid"] = organisation_id or -1
 
         simple_public_object_input_for_create = SimplePublicObjectInputForCreate(
             properties=properties,
--- a/api/integrations/lead_tracking/hubspot/client.py
+++ b/api/integrations/lead_tracking/hubspot/client.py
@@ -70,17 +70,22 @@ class HubspotClient:
     def create_company(
         self,
         name: str,
-        active_subscription: str,
-        organisation_id: int,
-        domain: str | None,
+        active_subscription: str = None,
+        organisation_id: int = None,
+        domain: str | None = None,
     ) -> dict:
         properties = {
             "name": name,
             "active_subscription": active_subscription,
             "orgid": str(organisation_id),
         }
+
         if domain:
             properties["domain"] = domain
+        if active_subscription:
+            properties["active_subscription"] = active_subscription
+        if organisation_id:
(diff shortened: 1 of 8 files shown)


api/users/models.py
```diff
@@ -123,6 +127,14 @@ def __str__(self):
     def subscribe_to_mailing_list(self):
         mailer_lite.subscribe(self)
 
+    @hook(AFTER_CREATE)
+    def schedule_hubspot_tracking(self) -> None:
+        if settings.ENABLE_HUBSPOT_LEAD_TRACKING:
+            track_hubspot_lead_without_organisation.delay(
+                kwargs={"user_id": self.id},
+                delay_until=timezone.now() + timedelta(minutes=30),
```
matthewelwell: This 30 minute number was chosen fairly arbitrarily - I wonder if I should add it to a constants file somewhere with a description of what it is so that we can easily change it in the future?

api/integrations/lead_tracking/hubspot/client.py
```diff
@@ -67,20 +68,55 @@ def create_contact(self, user: FFAdminUser, hubspot_company_id: str) -> dict:
...
         }
+
         if domain:
             properties["domain"] = domain
+        if active_subscription:
+            properties["active_subscription"] = active_subscription
+
+        # hubspot doesn't allow null values for numeric fields, so we
+        # set this to -1 for auto generated organisations.
+        properties["orgid"] = organisation_id or -1
```
zachaysan, line 119: 😢

api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -112,5 +116,15 @@ def update_company_active_subscription(
 
         return response
 
+    def _get_or_create_company_by_domain(self, domain: str) -> dict:
+        # try to find the organisation via it's domain
+        company = self.client.get_company_by_domain(domain)
+        if not company:
+            # otherwise create a new organisation with the domain
+            # as the name.
+            company = self.client.create_company(name=domain)
```
zachaysan: I'm not really sure what benefit the comments here bring, since it's all self-documented by the method names.

api/tests/unit/integrations/lead_tracking/hubspot/_hubspot_responses.py
```diff
@@ -0,0 +1,106 @@
+from datetime import datetime
+
+from dateutil.tz import tzlocal
+
+
+class DummyHubspotResponse:
+    """
+    Dummy class to replicate the to_dict() method of the Hubspot response classes.
+    """
```
zachaysan: The phrase "Dummy" is frowned upon as it originally and derogatorily means "mute person". See [here](https://github.com/pandas-dev/pandas/issues/35724) for a random discussion about it.

For classes like this I prefer the term `Fake` so `FakeHubspotResponse` in this case.

api/tests/unit/integrations/lead_tracking/hubspot/_hubspot_responses.py
```diff
@@ -0,0 +1,106 @@
...
+            "results": [
+                {
+                    "archived": False,
+                    "archived_at": None,
+                    "created_at": datetime(
+                        2024, 1, 25, 21, 48, 28, 655000, tzinfo=tzlocal()
+                    ),
+                    "id": "9765318341",
+                    "properties": {
+                        "createdate": "2024-01-25T21:48:28.655Z",
```
zachaysan, line 40: I know its not your fault, but it does seem weird that `created_at` is a `datetime` while `createdate` is a `str`.

api/tests/unit/integrations/lead_tracking/hubspot/_hubspot_responses.py
```diff
@@ -0,0 +1,106 @@
...
+        data={
+            "paging": None,
+            "results": [],
+            "total": 0,
+        }
+    )
+
+
+def generate_create_company_response(
+    name: str, domain: str = None, organisation_id: int = None
```
zachaysan: You're missing the typing for the `None` values here, but in addition to that I'm not sure if organisation should be set null since the default case as noted about is actually the value `-1`, right?

api/users/models.py
zachaysan: Yeah either in settings or as a feature flag with a default value set to 30 minutes.

zachaysan (approved):
Test coverage looks great and the rest of it looks good too. Just some small comments but no showstoppers.

api/integrations/lead_tracking/hubspot/lead_tracker.py
matthewelwell: Lol, yep, very good point. I've become everything I hate in a developer 😂

api/tests/unit/integrations/lead_tracking/hubspot/_hubspot_responses.py
matthewelwell: Hmm... yeah, this is what I got back from the hubspot API. I think the fact that I added docstrings to the functions explaining them should cover us here.

api/tests/unit/integrations/lead_tracking/hubspot/_hubspot_responses.py
matthewelwell: Yep, good spot. Fixed.

api/tests/unit/integrations/lead_tracking/hubspot/_hubspot_responses.py
matthewelwell: Good call, updated.

api/users/models.py
matthewelwell: Done.

PR #3834, opened 2024-04-24
chore: old features warning

kyle-ssg:

## Changes

- Shows an alert for old-non permanent features.
- The yellow tag was far too bright for light mode, toned it down.

image

### DIFF
--- a/frontend/web/components/StaleFlagWarning.tsx
+++ b/frontend/web/components/StaleFlagWarning.tsx
@@ -2,11 +2,12 @@ import { FC } from 'react'
 import Tag from './tags/Tag'
 import Constants from 'common/constants'
 import moment from 'moment'
-import { ProjectFlag } from 'common/types/responses'
+import { Project, ProjectFlag } from 'common/types/responses'
 import { getProtectedTags } from 'common/utils/getProtectedTags'
 import { IonIcon } from '@ionic/react'
 import { close, warning } from 'ionicons/icons'
 import Tooltip from './Tooltip'
+import ProjectStore from 'common/stores/project-store'
 
 type StaleFlagWarningType = {
   projectFlag: ProjectFlag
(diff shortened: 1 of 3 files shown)


frontend/web/components/StaleFlagWarning.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const protectedTags = getProtectedTags(projectFlag, `${projectFlag.project}`)
+  if (protectedTags?.length) {
+    return null
+  }
+  const created_date = projectFlag?.created_date
+  const daysAgo = created_date && moment().diff(moment(created_date), 'days')
+  console.log(ProjectStore.model)
+  const suggestStale =
+    daysAgo >=
+    ((ProjectStore.model as Project | null)?.stale_flags_limit_days || 30)
```
matthewelwell, line 26: Hmm... I did consider whether we should be using this field, but I'm not 100% certain we should be. The reason being that (a) this field is only available to enterprise customers and (b) it is used by the stale flags tagging logic as a comparison on when a feature was last changed in a given environment. 

It's probably fine for now, since it will be 30 days for everyone, and maybe we should create another field to hold the data for this.

matthewelwell (approved):
Approved with a minor comment.

frontend/web/components/StaleFlagWarning.tsx
kyle-ssg: Hmm yeah I also deliberated this quite a bit, I wonder if the duration should be longer than stale flags too. 

I'll leave for now as suggested

matthewelwell: Yeah, I think so

matthewelwell: But agree we can leave for now!

PR #3844, opened 2024-04-25
fix(hubspot): create hubspot company with domain

matthewelwell:

## Changes

Ensure that companies created when a user doesn't join an organisation are created with their domain. 

Updated a unit test.

### DIFF
--- a/api/integrations/lead_tracking/hubspot/lead_tracker.py
+++ b/api/integrations/lead_tracking/hubspot/lead_tracker.py
@@ -119,7 +119,7 @@ class HubspotLeadTracker(LeadTracker):
     def _get_or_create_company_by_domain(self, domain: str) -> dict:
         company = self.client.get_company_by_domain(domain)
         if not company:
-            company = self.client.create_company(name=domain)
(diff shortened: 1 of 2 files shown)


api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -119,7 +119,7 @@ def update_company_active_subscription(
     def _get_or_create_company_by_domain(self, domain: str) -> dict:
         company = self.client.get_company_by_domain(domain)
         if not company:
-            company = self.client.create_company(name=domain)
+            company = self.client.create_company(name=domain, domain=domain)
```
zachaysan: Since the domain param is passed twice (which looks weird) this could benefit from a short comment I think.

zachaysan (approved):
Looks good. One tiny small suggestion.

api/integrations/lead_tracking/hubspot/lead_tracker.py
matthewelwell: Done ✅

PR #3845, opened 2024-04-25
chore: fix github app link

dabeeeenster:


novakzaballa (dismissed):
I will create a new PR getting the GitHub App URL from an env var so self hosted customers can create their own GH App

matthewelwell (approved):

PR #3850, opened 2024-04-26
fix(versioning): prevent deleted segment overrides returning

matthewelwell:

## Changes

This PR sets out to resolve a bug in the v2 versioning which meant that deleted segment overrides from older versions were being returned alongside other feature states in more recent versions. 

This is achieved by grabbing the latest version for each feature in a given environment upfront and using this to filter the feature states. 

Some notes on the specifics of this PR:

1. Overall, I'm not particularly happy with the branching logic we now have in the `versioning_service` module and this is something that I'd like to work on refactoring, but if possible I'd like to do that after merging this PR so we can continue with testing. 
2. I resorted to using a raw SQL query to retrieve the latest versions from the database. I am sure it is possible using `annotate` in the Django ORM but I couldn't get it to work, if the reviewer has any suggestions, I'm all ears! 

Added tests.


api/tests/unit/features/versioning/test_unit_versioning_versioning_service.py
```diff
@@ -150,6 +151,47 @@ def test_get_environment_flags_v2_versioning_returns_latest_live_versions_of_fea
...
 
 
+def test_get_environment_flags_v2_versioning_does_not_return_removed_segment_override(
+    project: Project,
+    feature: Feature,
+    admin_user: FFAdminUser,
+    segment: Segment,
+    segment_featurestate: FeatureState,
+    environment_v2_versioning: Environment,
+):
```
khvn26:
```suggestion
) -> None:
```

khvn26 (approved):

PR #3864, opened 2024-05-01
style: Navigation improvements

kyle-ssg:

## Changes

Improves navigation such that

- Segments, integrations and compare no longer have environment ids in their urls
- Header nav sits above sidebar and show sub navigation based on if we are viewing the organisation or project
- The side menu will only show for environment specific pages
- Switching environments is done via a select vs showing them all as a collapsable list
- The top nav has a breadcrumb and menu allowing users to navigate to any organisation or project
- Clicking home will go to an organisation select page  
- Users and permissions and usage is separated from settings at the organisation level
- Settings / usage links are restricted to organisation admin
- Reduces font sizes in some places 
- Moves SDK keys to top level navigation
- Dev experience: Migrates some components to RTK and TypeScript 
- Dev experience: removes a lot of custom css 
- Minor: Remove unused images

_Please describe._


frontend/common/stores/account-store.js
```diff
@@ -299,12 +302,18 @@ const controller = {
       if (user && user.organisations) {
         store.organisation = user.organisations[0]
         const cookiedID = API.getCookie('organisation')
-        if (cookiedID) {
+        const pathID = matchPath(document.location.pathname, {
```
kyle-ssg, line 305: Prioritises url based organisation over last visited

frontend/common/stores/organisation-store.js
```diff
@@ -125,7 +126,7 @@ const controller = {
       })
   },
   getOrganisation: (id, force) => {
-    if (id !== store.id || force) {
+    if (`${id}` !== `${store.id}` || force) {
```
kyle-ssg, line 129: Protects against refetching organisation where the only difference was id vs "id"

frontend/common/utils/utils.tsx
```diff
@@ -384,6 +384,14 @@ const Utils = Object.assign({}, require('./base/_utils'), {
     return Constants.projectColors[index % (Constants.projectColors.length - 1)]
   },
 
+  getOrganisationHomePage(id?: string) {
```
kyle-ssg: Reusable function to go to the most relevant homepage

frontend/web/components/ButterBar.tsx
```diff
@@ -18,9 +19,12 @@ const ButterBar: React.FC<ButterBarProps> = ({ billingStatus, projectId }) => {
...
   const environment = matches && matches[1]
   const timerRef = useRef<NodeJS.Timer>()
-  const { data: featureImports, refetch } = useGetFeatureImportsQuery({
-    projectId,
-  })
+  const { data: featureImports, refetch } = useGetFeatureImportsQuery(
+    {
+      projectId,
+    },
+    { skip: !projectId },
```
kyle-ssg, line 26: Fixes an observed issue where we were sending requests to a url with project as undefined

frontend/web/components/GithubStar.tsx
```diff
@@ -36,11 +34,11 @@ const GithubStar: FC<GithubStarType> = ({}) => {
         style={{ width: 90 }}
         target='_blank'
         href='https://github.com/flagsmith/flagsmith'
-        className='btn btn-sm btn-with-icon text-body me-2'
+        className='btn btn-sm btn-with-icon text-body'
         rel='noreferrer'
       >
         <div className='d-flex flex-row justify-content-center align-items-center'>
-          <Icon name='github' width={20} fill='#9DA4AE' />
+          <IonIcon style={{ fontSize: 16 }} icon={logoGithub} />
```
kyle-ssg, line 41: Fixes an issue where the colour of dark mode wasn't matching the app

rolodato (commented):
This looks great, really excited to get it shipped! Here's my feedback, in no particular order:

1) I see an error in the projects page, haven't tested this locally so not sure if it's a result of running in a preview environment. I don't see any failed requests or JS errors, and this error doesn't show in production:

image

2) I like the clear hierarchy presented by the top navigation, but I find it confusing that the two up/down chevron icons do the same thing, and that there are 4 click targets for a hierarchy that is at most 2 elements deep:

image

I would consider turning each breadcrumb and its arrows into a single click target. Then, within each dropdown, only show organisations or projects depending on what was clicked.

For example, this is how Auth0 handles it:

https://github.com/Flagsmith/flagsmith/assets/829698/325b7e29-05c9-4fc2-ae2a-6f4b216b8c96

3) Have you considered naming the root projects tab "Environments"? That would make it consistent with the Organisation settings, and would also make it extra clear that everything inside this section is environment-specific:

image

image

To implement this and avoid having redundant "Environments" headings, we could show the current environment name in place of the "Environments" header, and make that show a dropdown selector when clicked.

This would also solve something I don't like about the current behaviour in prod - switching between environments looks like opening a collapsed menu, but in reality it's changing the current context. This can be confusing especially with high latency, which is something I regularly experience being far from the Core API servers. For example, notice how long it takes for the flag to change its current state when I change environments, even on a wired connection:

https://github.com/Flagsmith/flagsmith/assets/829698/a561cdcd-7434-4bd3-a011-f0e1f75d87f3

Also, because the click targets shuffle around as you change between environments, accidentally double-clicking an environment name can take you to a different menu. Example:

https://github.com/Flagsmith/flagsmith/assets/829698/9bad7e64-ff2f-4015-985a-397cb7a55a12

4) The Integrations tab feels like it has too much importance, for something that is almost always going to be set-and-forget. Trying to think of alternatives that don't sacrifice discoverability too much.

5) There might be something wrong with code snippets for segments - the environment key shows as `undefined` (doesn't happen in prod):

image

6) This is more of a thought than actual feedback, but have we considered shipping this new design right away but put it behind a feature flag? Curious how much work that would be :) I'd be happy to test this in production over time with real data.

dabeeeenster:
image

The separator/left hand area seems too wide? Also is it possible to make it user adjustable?

kyle-ssg:
**Done**
1 - Fixed
5 - Fixed

**Done but not sure** 
3 - Yeah this does make sense, it might help make it more obvious which environment we are on whilst reducing nav items. I think there are pros and cons to this, I've opened up internally for conversation.
image

**Not sure**
2 - This followed inspiration from Vercel after discussing with Ben, it felt like navigation between organisations and projects was the most intuitive example we found. It's a bit subjective, it might be worth leaving this one to wider feedback.
4 - I think this could be said about a lot of links in products (e.g , view a project in Vercel, view a repository in GH). We could put it in settings but I feel like there's more than enough real estate to keep it here. Also, we want to encourage and highlight new integrations, we're going to push more and more into them.
6 - Yes but there are far to many small changes (See changes) for it to be viable.

kyle-ssg:
> The separator/left hand area seems too wide? Also is it possible to make it user adjustable?

Ok yeah will do that

rolodato:
>  2 - This followed inspiration from Vercel after discussing with Ben, it felt like navigation between organisations and projects was the most intuitive example we found. It's a bit subjective, it might be worth leaving this one to wider feedback.

I hadn't noticed this from Vercel, that makes sense. I like that Vercel uses a higher contrast background colour here, I struggle to see the difference in ours:

image

vs

image

Also the changing background colour on mouseover doesn't seem to work as well (tested on Firefox):

https://github.com/Flagsmith/flagsmith/assets/829698/7a0fb4b5-64bb-43c1-aaea-0e1e1ae2bd22


https://github.com/Flagsmith/flagsmith/assets/829698/7cb94de7-ba9a-4c0e-befa-acb269b0df28

Lastly, I think we're missing the option to create a new organisation? We could add it to the same dropdown. Same for creating new projects, but at least that's still doable from the Projects page.


>  4 - I think this could be said about a lot of links in products (e.g , view a project in Vercel, view a repository in GH). We could put it in settings but I feel like there's more than enough real estate to keep it here. Also, we want to encourage and highlight new integrations, we're going to push more and more into them.

That's fair - in that case, could we put Integrations next to Settings? So the order from left to right would be:

1. Environments
2. Segments
3. Compare
4. Integrations
5. Settings

That way Integrations and Settings have a similar importance.

kyle-ssg:
> Also the changing background colour on mouseover doesn't seem to work as well (tested on Firefox):

Ok yeah I'll look at improving this.

> Lastly, I think we're missing the option to create a new organisation? We could add it to the same dropdown.

I'll do that just like with the env dropdown, however you can click the brand icon to go to organisations.

> That's fair - in that case, could we put Integrations next to Settings? So the order from left to right would be:

Yep that makes sense

frontend/web/components/App.js
```diff
@@ -443,64 +553,162 @@ const App = class extends Component {
...
+                              lastEnvironmentId || environmentId
+                            }/features`}
+                          >
+                            Environments
+                          </NavSubLink>
+                          <NavSubLink
+                            icon={<SegmentsIcon />}
+                            id={`segments-link`}
+                            to={`/project/${projectId}/segments`}
+                          >
```
novakzaballa, line 576: I think this is a great opportunity to clearly show in the UI/UX that a feature is a first-class citizen in our app, and belongs to Project level, we could easily do that by showing another tab called Features beside Environments showing exactly what we show today but without the ability to override the feature for an Environment, or even just asking for the environment if the user wants to do it. To distinguish the `features` option under the `Environments` tab we could name the option `Environment Features`. I think redundancy in this proposal worths it to gain clarity.

rolodato:
One last thing, there's still an error shown after creating a new organisation, but before the first project is created:

image

I'm sure there's little things we can find here and there to improve but I'd love to get this merged ASAP. Ship it! :shipit:

frontend/web/components/App.js
kyle-ssg: Yeah this will clear the way for deciding how we have a separate screen for features, though it won't be part of this PR

kyle-ssg:
Done ^^

frontend/web/components/App.js
novakzaballa: Resolving since as discussed internally this is out of the scope of this PR and will be addressed in the future.

zachaysan (approved):
All the stuff I've tested looks good. Really like this refactor.

matthewelwell:
Seems like there's still some nasty debugging going on here. I managed to get this to appear by just navigating to project settings. 

image

kyle-ssg:
Figured this out, was due to tags and is an existing issue in production. This PR will resolve it.

matthewelwell (approved):
Confirmed above issue as fixed.

Issue #3417, opened 2024-02-13
Handle influxdb wrapper function parameters better


zachaysan:
In `app_analytics/influxdb_wrapper.py` there are a number of parameters that take a string which represents a date range, like `30d` for thirty days. The task is to replace these string date ranges with `timedelta` and `datetime` without changing query responses.

matthewelwell:
Resolved in #4450

PR #3870, opened 2024-05-01
feat: Add api usage metrics for different periods

zachaysan:

## Changes

This provides the ability to get API usage data for different billing periods. The supported new periods are the current billing period (i.e., the first day of the billing period till now), the previous billing period, and a 90-day rolling period to the present.

Part of these changes involved with a refactoring of how code is called to influxdb. The former code incorrectly attributed `date_range` as if it were `date_start` with an inverse modifier. With the former code any callers that had set a `date_stop` to anything other than `now()` the resultant range would be smaller than the interface contract implied. These issues have been addressed and subsequent calling methods have been updated.

There was also an issue that cropped up with Pydantic and Freeze Gun, which has been temporarily solved with the addition of an auto-use function in `conftest.py`. 

Tested manually against production influxdb to verify results then also added 3 tests to ensure calling signatures to related influxdb functions.


api/app_analytics/analytics_db_service.py
```diff
@@ -14,31 +14,103 @@
...
 ANALYTICS_READ_BUCKET_SIZE = 15
 
 
 def get_usage_data(
-    organisation, environment_id=None, project_id=None
-) -> List[UsageData]:
+    organisation: Organisation,
+    environment_id: int | None = None,
+    project_id: int | None = None,
+    period: str | None = None,
```
khvn26: How about we define a type like `Literal["current_billing_period", "previous_billing_period"]` for this annotation?

api/app_analytics/analytics_db_service.py
```diff
@@ -14,31 +14,103 @@
...
+) -> list[UsageData]:
+    now = timezone.now()
+
+    date_stop = date_start = None
+    period_starts_at = period_ends_at = None
```
matthewelwell, line 39: TIL! Never seen this multi-assignment before.

api/app_analytics/analytics_db_service.py
```diff
@@ -14,31 +14,103 @@
...
+
+    if period == "current_billing_period":
```
matthewelwell: Should this be a constant defined somewhere?

api/app_analytics/analytics_db_service.py
```diff
@@ -14,31 +14,103 @@
...
+        if not getattr(organisation, "subscription_information_cache", None):
+            return []
+        sub_cache = organisation.subscription_information_cache
+        starts_at = sub_cache.current_billing_term_starts_at
+        month_delta = relativedelta(now, starts_at).months
+        period_starts_at = relativedelta(months=month_delta) + starts_at
+        period_ends_at = now
+        date_start = f"-{(now - period_starts_at).days}d"
```
matthewelwell: I don't think it particularly matters here, but it's odd to me that we define `period_ends_at` above, but then use `now` here still.

api/app_analytics/analytics_db_service.py
```diff
@@ -14,31 +14,103 @@
...
+        sub_cache = organisation.subscription_information_cache
+        starts_at = sub_cache.current_billing_term_starts_at
+        month_delta = relativedelta(now, starts_at).months - 1
+        month_delta += relativedelta(now, starts_at).years * 12
+        period_starts_at = relativedelta(months=month_delta) + starts_at
+        period_ends_at = relativedelta(months=month_delta + 1) + starts_at
+        date_start = f"-{(now - period_starts_at).days}d"
+        date_stop = f"-{(now - period_ends_at).days}d"
+
+    elif period == "90_day_period":
```
matthewelwell: again, should we use a constant or even an enum for this?

api/app_analytics/analytics_db_service.py
matthewelwell: Separately (and feel free to ignore), but we could be fancy (well, `3.10` fancy) and use pattern matching here instead, right?

api/app_analytics/analytics_db_service.py
```diff
@@ -14,31 +14,103 @@
...
+            )
         return get_usage_data_from_local_db(
-            organisation, environment_id=environment_id, project_id=project_id
+            organisation=organisation,
+            environment_id=environment_id,
+            project_id=project_id,
+        )
+
+    if date_start:
+        assert date_stop
```
matthewelwell, line 92: Minor: couldn't we do these assertions in the `get_usage_data_from_x` functions and simplify the code here (by just defining the kwargs and keeping a single return)

api/app_analytics/influxdb_wrapper.py
```diff
@@ -137,7 +140,9 @@ def get_events_for_organisation(organisation_id: id, date_range: str = "30d") ->
     return total
 
 
-def get_event_list_for_organisation(organisation_id: int, date_range: str = "30d"):
+def get_event_list_for_organisation(
+    organisation_id: int, date_start: str = "-30d", date_stop: str = "now()"
+) -> [dict[list[int]], list[str]]:
```
khvn26:
```suggestion
) -> tuple[dict[str, list[int]], list[str]]:
```

api/app_analytics/influxdb_wrapper.py
```diff
@@ -238,13 +261,16 @@ def get_multiple_event_list_for_feature(
     :param environment_id: an id of the environment to get usage for
     :param feature_name: the name of the feature to get usage for
     :param period: the influx time period to filter on, e.g. 30d, 7d, etc.
+      Note that "period" is tied to existing serializer logic. Prefer
+      "date_start" for new interfaces.
```
khvn26: Is it too much of a lift to change the argument to `date_start` here?

api/conftest.py
```diff
@@ -844,7 +845,11 @@ def github_repository(
         "admin_master_api_key_client",
     ]
 )
-def admin_client_new(request, admin_client_original, admin_master_api_key_client):
+def admin_client_new(
+    request: SubRequest,
```
khvn26:
```suggestion
    request: pytest.FixtureRequest,
```

api/app_analytics/serializers.py
```diff
@@ -12,6 +12,7 @@ class UsageDataSerializer(serializers.Serializer):
 class UsageDataQuerySerializer(serializers.Serializer):
     project_id = serializers.IntegerField(required=False)
     environment_id = serializers.IntegerField(required=False)
+    period = serializers.CharField(required=False)
```
khvn26: Do you think it's a good idea to use [RegexField](https://www.django-rest-framework.org/api-guide/fields/#regexfield) here?

api/app_analytics/influxdb_wrapper.py
```diff
@@ -149,14 +154,20 @@ def get_event_list_for_organisation(organisation_id: int, date_range: str = "30d
...
+        date_stop=date_stop,
     )
     dataset = defaultdict(list)
     labels = []
     for result in results:
         for record in result.records:
             dataset[record["resource"]].append(record["_value"])
-            required_records = int(date_range[:-1]) + 1
+            if date_stop == "now()":
+                required_records = abs(int(date_start[:-1])) + 1
```
matthewelwell, line 166: Pragmatic comment: I wonder if we should perhaps factor out the code to convert the influx date range string to an integer here so we can do some error handling. I'm a _little_ concerned that there's technically nothing stopping the consumer from passing e.g. `now()` as the value for date start for example (other than the obvious point that it wouldn't return any data)

matthewelwell: Longer term comment: I think this is a good argument from factoring out the influx specific syntax and just using python types as per this issue: https://github.com/Flagsmith/flagsmith/issues/3417

api/app_analytics/influxdb_wrapper.py
```diff
@@ -149,14 +154,20 @@ def get_event_list_for_organisation(organisation_id: int, date_range: str = "30d
...
+            else:
+                required_records = (
+                    abs(int(date_start[:-1])) - abs(int(date_stop[:-1])) + 1
+                )
             if len(labels) != required_records:
```
matthewelwell, line 171: I'm not sure what cases this might happen?

api/app_analytics/influxdb_wrapper.py
```diff
@@ -201,14 +215,23 @@ def get_multiple_event_list_for_organisation(
...
             dataset[i]["name"] = record.values["_time"].strftime("%Y-%m-%d")
+
     return dataset
 
 
 def get_usage_data(
-    organisation_id: int, project_id: int = None, environment_id=None
-) -> typing.List[UsageData]:
+    organisation_id: int,
+    project_id: int = None,
```
matthewelwell: nit: 

```suggestion
project_id: int | None = None
```

api/app_analytics/influxdb_wrapper.py
```diff
@@ -238,13 +261,16 @@ def get_multiple_event_list_for_feature(
...
+      "date_start" for new interfaces.
     :param aggregate_every: the influx time period to aggregate the data by, e.g. 24h
 
     :return: a list of dicts with feature and request count in a specific environment
     """
 
+    date_start = f"-{period}"
     results = InfluxDBWrapper.influx_query_manager(
-        date_range=period,
+        date_start=date_start,
```
matthewelwell, line 270: Probably my own misunderstanding, but I'm not sure why this doesn't include `date_end` - perhaps it's just not needed.

api/tests/unit/app_analytics/test_analytics_db_service.py
```diff
@@ -308,3 +313,106 @@ def test_get_feature_evaluation_data_calls_get_feature_evaluation_data_from_loca
...
+        organisation=organisation,
+        environment_id=None,
+        project_id=None,
+        date_start=datetime(2022, 12, 30, 9, 9, 47, 325132, tzinfo=timezone.utc),
+        date_stop=datetime(2023, 1, 19, 9, 9, 47, 325132, tzinfo=timezone.utc),
+    )
+
+
+@pytest.mark.freeze_time("2023-01-19T09:09:47.325132+00:00")
+def test_get_usage_data_returns_get_usage_data_from_local_db_with_set_period_starts_at_with_previous_billing_period(
```
matthewelwell:
```suggestion
def test_get_usage_data_calls_get_usage_data_from_local_db_with_set_period_starts_at_with_previous_billing_period(
```

api/tests/unit/app_analytics/test_analytics_db_service.py
```diff
@@ -308,3 +313,106 @@ def test_get_feature_evaluation_data_calls_get_feature_evaluation_data_from_loca
...
+    # When
+    usage_data = get_usage_data(organisation, period=period)
+
+    # Then
+    assert usage_data == []
+    mocked_get_usage_data_from_local_db.assert_not_called()
+
+
+@pytest.mark.freeze_time("2023-01-19T09:09:47.325132+00:00")
+def test_get_usage_data_returns_get_usage_data_from_local_db_with_set_period_starts_at_with_current_billing_period(
```
matthewelwell:
```suggestion
def test_get_usage_data_calls_get_usage_data_from_local_db_with_set_period_starts_at_with_current_billing_period(
```

api/tests/unit/app_analytics/test_analytics_db_service.py
```diff
@@ -308,3 +313,106 @@ def test_get_feature_evaluation_data_calls_get_feature_evaluation_data_from_loca
...
+
+
+@pytest.mark.freeze_time("2023-01-19T09:09:47.325132+00:00")
+def test_get_usage_data_returns_get_usage_data_from_local_db_with_set_period_starts_at_with_previous_billing_period(
+    mocker: MockerFixture, settings: SettingsWrapper, organisation: Organisation
+) -> None:
+    # Given
+    period: str = "previous_billing_period"
+
+    settings.USE_POSTGRES_FOR_ANALYTICS = True
```
khvn26: nit: Could this be just one test with parametrised `period`, `date_start` and `date_stop` arguments?

khvn26 (changes requested):
LGTM in general. Didn't manage to find any flaws in logic/architecture. Commented a couple minor issues/nitpicks. Requesting changes because some of the comments point out incorrect typing.

api/tests/unit/app_analytics/test_unit_app_analytics_views.py
```diff
@@ -99,6 +105,189 @@ def test_get_usage_data(mocker, admin_client, organisation):
...
+    # Given
+    url = reverse("api-v1:organisations:usage-data", args=[organisation.id])
+    url += "?period=previous_billing_period"
+
+    mocked_get_usage_data = mocker.patch(
+        "app_analytics.analytics_db_service.get_usage_data_from_influxdb",
+        autospec=True,
+        return_value=[
+            UsageData(flags=10, day=date.today()),
+            UsageData(flags=10, day=date.today() - timedelta(days=1)),
```
matthewelwell: I get that this is just getting passed through and asserted in the json below, but we should probably keep this realistic and use dates that would fall in the period we're filtering on.

matthewelwell (changes requested):

api/app_analytics/analytics_db_service.py
zachaysan: Yeah it's nice. Although, doesn't work with typing, unfortunately.

api/app_analytics/analytics_db_service.py
zachaysan: Good idea. Done.

api/app_analytics/analytics_db_service.py
zachaysan: Defined the constant and switched to neat-o pattern matching.

(As an aside, pattern matching feels a bit weird because *defining a variable* means you have to access it through a module or a class for it to *read the value for matching*).

api/app_analytics/analytics_db_service.py
zachaysan: Because `date_start` is always relative to the present, we should always use `now` even if we have the same value in `period_ends_at`.

That said, I agree that it looks odd.

api/app_analytics/analytics_db_service.py
zachaysan: Good idea. I've used the new constants in the definition so it's even better.

api/app_analytics/analytics_db_service.py
zachaysan: The assertions have to stay in the module because they are not common to the code that lives in where they are calling. It's asserting that we're understanding the world correctly, not assertions that we're calling the function correctly.

But I did simplify the code by using kwargs and keeping the number of function calls to a minimum 👍🏼

api/app_analytics/influxdb_wrapper.py
zachaysan: Well the one view that calls this code now transforms it to a negative signed string, so `now()` isn't technically possible. I think we should keep the refactoring work to #3417 since partially fixing it here will make things inconsistent. I'd rather have the whole codebase be upgraded at once.

api/app_analytics/influxdb_wrapper.py
zachaysan: It's existing code and I'm not sure when `labels` wouldn't align with the `required_records`. My best guess is that it's trying to add time strings for records that are previously unseen.

api/app_analytics/influxdb_wrapper.py
zachaysan: The lift costed about 30 minutes of time refactoring the code around it. While I was at it I noticed another function that had the same param issue and way more callers so I've decided not to fix that one right now. I'll have to get to it once the timedelta refactor comes around.

api/app_analytics/influxdb_wrapper.py
zachaysan: Yeah the calling code always wants it to the present. Keep in mind that this is sort of outside the work that this PR is supposed to be focussed on. I tried to fix the other interfaces while I was at it, but I don't really want to refactor all the influxdb code we have, haha 😄

api/app_analytics/serializers.py
zachaysan: Done. Used the constants to be extra 👌🏼

api/conftest.py
zachaysan: When I check the type using pdb I get `SubRequest` as the type. Why should we switch it to `FixtureRequest`?

api/tests/unit/app_analytics/test_analytics_db_service.py
zachaysan: Fixed

api/tests/unit/app_analytics/test_analytics_db_service.py
zachaysan: Fixed

api/conftest.py
khvn26: 1. `SubRequest` is a subtype of `FixtureRequest`.
2. `SubRequest` is not meant to be imported directly. `FixtureRequest` is a public export of pytest and is sufficient to be used as an annotation here.

api/tests/unit/app_analytics/test_analytics_db_service.py
zachaysan: I don't really like over parametrizing tests with the mocked assertion having to be added into the parametrized decorator. The downside of duplicated code between tests is small and I'd rather more easily read them and debug them. Not to say I never parametrize tests, I do, I just don't like the really convoluted stuff.

api/conftest.py
khvn26: In general, `pytest` exports all types usually needed when annotating tests at package level, e.g.

- `pytest.FixtureRequest`
- `pytest.MonkeyPatch`
- `pytest.CaptureFixture`

`_` in the beginning of the `_pytest` package hints that it's private and is not meant to be used directly, at least in the test code.

api/tests/unit/app_analytics/test_analytics_db_service.py
khvn26: How about factoring out the `OrganisationSubscriptionInformationCache` object to a common fixture then?

api/app_analytics/serializers.py
khvn26: > Used the constants

So the period is not meant to be arbitrary? If we're using a fixed set of pre-defined periods, I'd suggest using ChoiceField instead.

api/tests/unit/app_analytics/test_unit_app_analytics_views.py
zachaysan: Yeah good call. I normally do things like that but I'm not sure why I didn't line up the dates this time. Updated.

api/tests/unit/app_analytics/test_analytics_db_service.py
zachaysan: That sounds like a good idea. Implementing now.

api/conftest.py
zachaysan: Ok you've convinced me. Will update.

api/app_analytics/constants.py
```diff
@@ -0,0 +1,6 @@
+ANALYTICS_READ_BUCKET_SIZE = 15
+
+# get_usage_data() related period constants
+CURRENT_BILLING_PERIOD = "current_billing_period"
+PREVIOUS_BILLING_PERIOD = "previous_billing_period"
+_90_DAY_PERIOD = "90_day_period"
```
khvn26: Why is this prefixed with underscore (ergo, private)?

api/app_analytics/analytics_db_service.py
```diff
@@ -14,34 +14,109 @@
...
 
 
 def get_usage_data(
-    organisation, environment_id=None, project_id=None
-) -> List[UsageData]:
+    organisation: Organisation,
+    environment_id: int | None = None,
+    project_id: int | None = None,
+    period: (
+        Literal[
```
khvn26: It's nicer to avoid defining the type inline here, i.e.

```python
UsageDataPeriod = Literal[
    constants.CURRENT_BILLING_PERIOD,
    constants.PREVIOUS_BILLING_PERIOD,
    constants._90_DAY_PERIOD,
]

...

def get_usage_data(
    ...
    period: UsageDataPeriod | None = None,
```

api/app_analytics/serializers.py
```diff
@@ -12,6 +14,17 @@ class UsageDataSerializer(serializers.Serializer):
...
+        regex=r"^(%s)$"
+        % "|".join(
+            [
+                constants.CURRENT_BILLING_PERIOD,
+                constants.PREVIOUS_BILLING_PERIOD,
+                constants._90_DAY_PERIOD,
+            ]
+        ),
+        required=False,
+    )
```
khvn26: When I offered to use `RegexField` initially, I was thinking more of a `^\d+[d|h|m]$` regex, as I didn't know that the periods were meant to be predefined.

Now it seems more appropriate to use `ChoiceField` instead (and if you define the literal type outside the method definition as suggested [here](https://github.com/Flagsmith/flagsmith/pull/3870/files#r1604137042), you'll be able to use `typing.get_args` with it to provide the choices in a DRY manner).

api/app_analytics/constants.py
zachaysan: Because python (very annoyingly) doesn't allow variables or constants to start with a number.

api/app_analytics/analytics_db_service.py
zachaysan: Yeah I agree I'll do that when I update this PR.

api/app_analytics/serializers.py
zachaysan: Yeah I should have gone with a `ChoiceField`

api/app_analytics/constants.py
matthewelwell: I also had this question, so I think it warrants a better resolution tbh. Why not `NINETY_DAY_PERIOD` ? If we're adamant on keeping the leading underscore for whatever reason, lets at least add a comment here.

khvn26: Great solution Matt, I agree

zachaysan: Ok this is done.

api/app_analytics/analytics_db_service.py
zachaysan: Moved this into its own module called `typing` with the other changes as noted above.

api/app_analytics/serializers.py
zachaysan: Ok created the `ChoiceField` with the `typing.get_args` trick. Nice.

api/app_analytics/analytics_db_service.py
khvn26: Sorry if I'm annoying, but we have modules like this in the codebase and they're called `types`. Let's rename yours for consistency.

api/app_analytics/analytics_db_service.py
```diff
@@ -14,34 +14,103 @@
...
+    period: PERIOD_TYPE | None = None,
+) -> list[UsageData]:
+    now = timezone.now()
+
+    date_stop = date_start = None
+    period_starts_at = period_ends_at = None
+
+    match period:
+        case constants.CURRENT_BILLING_PERIOD:
+            if not getattr(organisation, "subscription_information_cache", None):
```
khvn26, line 43: nit:

```suggestion
            if not organisation.subscription_information_cache.exists():
```

api/app_analytics/analytics_db_service.py
```diff
@@ -14,34 +14,103 @@
...
+            sub_cache = organisation.subscription_information_cache
+            starts_at = sub_cache.current_billing_term_starts_at
+            month_delta = relativedelta(now, starts_at).months
+            period_starts_at = relativedelta(months=month_delta) + starts_at
+            period_ends_at = now
+            date_start = f"-{(now - period_starts_at).days}d"
+            date_stop = "now()"
+
+        case constants.PREVIOUS_BILLING_PERIOD:
+            if not getattr(organisation, "subscription_information_cache", None):
```
khvn26, line 54: nit:

```suggestion
            if not organisation.subscription_information_cache.exists():
```

khvn26 (approved):
Two nitpicks, one naming request I trust @zachaysan to address, otherwise LGTM.

api/app_analytics/serializers.py
```diff
@@ -12,6 +16,12 @@ class UsageDataSerializer(serializers.Serializer):
 class UsageDataQuerySerializer(serializers.Serializer):
     project_id = serializers.IntegerField(required=False)
     environment_id = serializers.IntegerField(required=False)
+    period: PERIOD_TYPE | None = serializers.ChoiceField(
```
khvn26: nit: I'm not sure we're able to type-annotate serializer fields this way. Doesn't matter much until we enforce mypy, of course, but I think this annotation is incorrect.

api/app_analytics/analytics_db_service.py
khvn26: The above suggestion is incorrect.

api/app_analytics/analytics_db_service.py
zachaysan: As mentioned here: https://github.com/Flagsmith/flagsmith/pull/3870#discussion_r1611667665

The suggestion is incorrect

api/app_analytics/serializers.py
zachaysan: Ok I'll remove typing for it. Not a huge deal either way.

api/app_analytics/analytics_db_service.py
zachaysan: Ah yes of course. Will update 👍🏼

matthewelwell (approved):

PR #3872, opened 2024-05-02
docs: Rewrite segments doc

rolodato:

## Changes

See https://github.com/Flagsmith/flagsmith/pull/3872/commits/c5fa3cc3539571574fe55179c6612435fa3a897a for the main changes and the commit description. 

The title of the document was renamed from "Managing Segments" to just "Segments". Eventually we should rename the other documents to follow suit:

* Managing Features
* Managing Identities

N/A


docs/docs/basic-features/segments.md
```diff
@@ -0,0 +1,239 @@
...
+
+<Tabs groupId="operators" queryString>
+<TabItem value="in" label="In">
+
+The `In` operator lets you match a trait value against a comma-separated list of values. For example, the segment rule
+value might read `21,682,8345`. This would match against a trait value of `682` but not against a trait value of `683`
+or `834`.
+
+The `In` operator can be useful to build segments that represent a specific set of tenants in your application. For
+example, you could create a segment with the following rule: `tenant_id In tenant_1,tenant_2,tenant_3`
```
khvn26, line 175: We have a lot of incoming questions about the segment condition value size limit specifically for the `In` operator.

We should clearly state the condition value size limit somewhere in this doc, and reference it here.

rolodato: Good call, I think I just forgot to add it as part of the rewrite. I'll add that back in

frontend/web/components/pages/SegmentsPage.tsx
```diff
@@ -344,7 +344,7 @@ const SegmentsPage: FC<SegmentsPageType> = (props) => {
                       added to the segments based on the rules you create.{' '}
                       <Button
                         theme='text'
-                        href='https://docs.flagsmith.com/basic-features/managing-segments'
+                        href='https://docs.flagsmith.com/basic-features/segments'
```
kyle-ssg, line 347: It might be worth adding this redirect in vercel.json, we'll probably be ranking for this url in search

docs/docs/basic-features/segments.md
rolodato: Actually that was a false memory 😅 the limits are only mentioned here https://docs.flagsmith.com/system-administration/system-limits

Added a separate section just for the limits in 3de8a58d3c4762eb45329989f56f95a49cf452b5, that should cover all operators and not just `In`.

khvn26 (approved):

PR #3879, opened 2024-05-03
chore: Linkedin partner tracking

### LINKED ISSUE #3837: Set up tracking linkedin tracking for saas
We are looking to set up tracking for the app.flagsmith subdomain so that we can understand user behaviour of campaign traffic. 

LinkedIn Insight Tag 
Paste the Insight Tag code below in your website’s global footer, right above the closing HTML <body> tag. Adding the tag to the footer will let you track conversions or retarget on any page across your whole site.

```
<script type="text/javascript">
_linkedin_partner_id = "5747572";
window._linkedin_data_partner_ids = window._linkedin_data_partner_ids || [];
window._linkedin_data_partner_ids.push(_linkedin_partner_id);
</script><script type="text/javascript">
(function(l) {
if (!l){window.lintrk = function(a,b){window.lintrk.q.push([a,b])};
window.lintrk.q=[]}
var s = document.getElementsByTagName("script")[0];
var b = document.createElement("script");
b.type = "text/javascript";b.async = true;
b.src = "https://snap.licdn.com/li.lms-analytics/insight.min.js";
s.parentNode.insertBefore(b, s);})(window.lintrk);
</script>
<noscript>
<img height="1" width="1" style="display:none;" alt="" src="https://px.ads.linkedin.com/collect/?pid=5747572&fmt=gif" />
</noscript>
```

Once the insight tag is installed, the following insight tag event conversions can be created:

Created Organisation
This code needs to load when users have completed the “Created Organisation” action:
`window.lintrk('track', { conversion_id: 16798338 });
`
Created Project
This code needs to load when users have completed the “Created Project” action:
`window.lintrk('track', { conversion_id: 16798346 });
`
Created Feature
This code needs to load when users have completed the “Created Feature” action:

`window.lintrk('track', { conversion_id: 16798354 });
`
Please contact Anna Redbond in case more background or details are needed.

kyle-ssg:

## Changes

Adds linkedin partner tracking when LINKEDIN_PARTNER_TRACKING=true 

- Ran the frontend with LINKEDIN_PARTNER_TRACKING=true, validated conversion events were sent.
- Validated no requests are sent to Linkedin when the variable is disabled


api/app/views.py
```diff
@@ -47,7 +47,7 @@ def project_overrides(request):
         "fpr": "FIRST_PROMOTER_ID",
         "headway": "HEADWAY_API_KEY",
         "hideInviteLinks": "DISABLE_INVITE_LINKS",
-        "linkedin_api_key": "LINKEDIN_API_KEY",
+        "linkedinPartnerTracking": "LINKEDIN_PARTNER_TRACKING",
```
matthewelwell, line 50: I see here that we've replaced `linkedin_api_key` but not removed it from the settings [here](https://github.com/Flagsmith/flagsmith/pull/3880) - just checking this was intentional?

matthewelwell (commented):
Added one minor specific comment based on the implementation but I'm also keen to make sure that we've verified this doesn't cause any unnecessary outbound requests when the option is disabled, e.g. for self hosted installations. I can't see any reference to it in the test section of the PR description.

api/app/views.py
kyle-ssg: Yeah it isnt used anywhere by the looks of it

kyle-ssg: linkedin_api_key doesnt seem to be used anywhere else unless I'm missing something, have removed from common.py

kyle-ssg:
> Added one minor specific comment based on the implementation but I'm also keen to make sure that we've verified this doesn't cause any unnecessary outbound requests when the option is disabled, e.g. for self hosted installations. I can't see any reference to it in the test section of the PR description.

Yeah, it wouldn't. The script only gets added to the page if the env var is set https://github.com/Flagsmith/flagsmith/pull/3879/files#diff-d2af3ec690c6d6d6c4691ed1d668ec2cf39c84b774cef27c7346e5efd555d321R106

api/app/views.py
novakzaballa: I understand that this API configuration is only needed for self-hosted deployments, right? Are we interested in having this tracking enabled for them?

novakzaballa (approved):
The code LGTM but I left a comment regarding self-hosted case.

api/app/views.py
kyle-ssg: I'd say it's as good as any place to put it regardless of that, maybe we should have a wider conversation around our env vars, permanent flags and project_prod.js at some point.

PR #3881, opened 2024-05-03
feat: Edge V2 migration opt-in, capacity budget for migration

khvn26:

### Naming

`IdentityOverridesV2MigrationStatus` enumeration is renamed to `EdgeV2MigrationStatus`.

### Settings

Added:
- `EDGE_V2_MIGRATION_READ_CAPACITY_BUDGET`: controls how much read capacity is allowed during migration to `environments_v2`. Default value is `0`. The idea is to increase it to a sensible amount whenever Edge V2 is live. 

### Business logic 

Renamed:
- `Project.identity_overrides_v2_migration_status` -> `Project.edge_v2_migration_status`.

Modified:
- `DynamoIdentityWrapper.iter_all_items_paginated` now accepts `capacity_budget` argument. If the budget is reached or exceeded during iteration, it raises a `CapacityBudgetExceeded` exception. If `capacity_budget` is `0`, the error is raised before any request to DynamoDB is made. 

Added: 
- `Project.edge_v2_read_capacity_budget` overrides the `EDGE_V2_MIGRATION_READ_CAPACITY_BUDGET` for the given project. Field value is `null` by default.
- `EdgeV2MigrationStatus.INCOMPLETE` marks projects that have exceeded the read capacity budget during migration and have only environment documents migrated. Identity override migration is completely skipped for such projects.
 
### API 
- Added `edge_v2_migration_status` to `project` endpoints so the migration can be triggered via REST API.
- `edge_v2_read_capacity_budget` is only modifiable via Django admin page. 

All code covered with unit tests.


api/environments/dynamodb/services.py
```diff
@@ -26,39 +45,55 @@ def migrate_environments_to_v2(project_id: int) -> IdentityOverridesV2Changeset
...
-        "Retrieved %d identity overrides to migrate for project %d",
-        len(changeset.to_put),
-        project_id,
-    )
+    identity_overrides_changeset = IdentityOverridesV2Changeset(to_put=[], to_delete=[])
+    result_status = EdgeV2MigrationStatus.COMPLETE
+
+    try:
+        to_put = list(
+            _iter_paginated_overrides(
```
gagantrivedi, line 57: We should use project_expression to   only fetch the data we need for migration

api/environments/dynamodb/services.py
```diff
@@ -26,39 +45,55 @@ def migrate_environments_to_v2(project_id: int) -> IdentityOverridesV2Changeset
...
-    logger.info(
-        "Retrieved %d identity overrides to migrate for project %d",
-        len(changeset.to_put),
-        project_id,
-    )
+    identity_overrides_changeset = IdentityOverridesV2Changeset(to_put=[], to_delete=[])
+    result_status = EdgeV2MigrationStatus.COMPLETE
+
+    try:
+        to_put = list(
```
gagantrivedi, line 56: Do we have a rough idea about how much data will this be for an environment with several million identities? I just want to make sure we don't hurt task processor

gagantrivedi (changes requested):

api/environments/dynamodb/services.py
khvn26: Added.

api/environments/dynamodb/services.py
khvn26: `len(to_put)` is equal to the count of only the identities that have overridden feature states. The rest of identities is iterated over but discarded.

api/environments/dynamodb/wrappers/identity_wrapper.py
```diff
@@ -89,37 +91,49 @@ def get_all_items(
...
         }
-        if projection_expression:
-            query_kwargs["ProjectionExpression"] = projection_expression
-
         if start_key:
             query_kwargs["ExclusiveStartKey"] = start_key
+        if projection_expression:
+            query_kwargs["ProjectionExpression"] = projection_expression
+        if return_consumed_capacity:
+            query_kwargs["ReturnConsumedCapacity"] = "TOTAL"
```
matthewelwell: This seems a bit cryptic, I'd love a comment here.

api/projects/migrations/0023_rename_identity_overrides_v2_migration_status_project_edge_v2_migration_status.py
```diff
@@ -0,0 +1,18 @@
...
+        ("projects", "0022_add_stale_flags_threshold_to_project"),
+    ]
+
+    operations = [
+        migrations.RenameField(
+            model_name="project",
+            old_name="identity_overrides_v2_migration_status",
+            new_name="edge_v2_migration_status",
+        ),
+    ]
```
matthewelwell: Erk, we'll get schema mismatch issues here during deployment! We'll probably have to just rename the python attribute and set the `column_name` on the field definition to keep it the same as it is currently.

matthewelwell (changes requested):
I've gone through the code for now (still need to review the tests) but wanted to submit this now (and go get some dinner...)

api/environments/dynamodb/services.py
gagantrivedi: But are we still fetching all the identities? I thought we found a way to only fetch identities that have overrides?

khvn26: No, we haven't. Even if we use a filter expression we'll spend the same amount of read capacity.

gagantrivedi: I think the following should work?

```
 query_kwargs = {
        "IndexName": "environment_api_key-identifier-index",
        "KeyConditionExpression": filter_expression,
        "FilterExpression": Attr("identity_features").ne([]),
    }
``

File: api/tests/unit/environments/dynamodb/test_unit_services.py [hunk 5]
```diff
@@ -73,8 +81,56 @@ def test_migrate_environments_to_v2__wrapper_disabled__does_not_write(
...
+    def iter_all_items_paginated_gen_mock(**_):
+        yield map_identity_to_identity_document(identity)
+        raise CapacityBudgetExceeded(expected_capacity_budget, Decimal(13))
+
+    mocked_dynamodb_identity_wrapper.iter_all_items_paginated.side_effect = (
+        iter_all_items_paginated_gen_mock
+    )
+    mocker.patch(
+        "environments.dynamodb.services.DynamoIdentityWrapper",
+        autospec=True,
```
matthewelwell, line 113: minor: it's odd that we use `autospec` here as well as defining the `spec` in the MagicMock constructor above.

api/environments/dynamodb/services.py
khvn26: Let me compare performance between the two variants. If there's no major difference, I'm in favour of this change.

api/projects/migrations/0023_rename_identity_overrides_v2_migration_status_project_edge_v2_migration_status.py
khvn26: Done.

api/environments/dynamodb/wrappers/identity_wrapper.py
khvn26: Added.

api/environments/dynamodb/services.py
khvn26: Done.

api/tests/unit/environments/dynamodb/test_unit_services.py
khvn26: I agree it's a bit overprotective, but it doesn't hurt.

matthewelwell (approved):

gagantrivedi (approved):

PR #3883, opened 2024-05-03
chore: Convert features models TestCase to a normal test function

zachaysan:

## Changes


api/tests/unit/features/test_unit_features_models.py
```diff
@@ -220,322 +218,295 @@ def test_when_create_feature_with_tags_then_success(project: Project) -> None:
...
+@mock.patch("features.signals.trigger_feature_state_change_webhooks")
+def test_cannot_create_duplicate_feature_state_in_an_environment_for_segment(
+    mock_trigger_webhooks: mock.MagicMock,
+    project: Project,
+    feature: Feature,
+    environment: Environment,
+) -> None:
+    """
+    Note that although the mock isn't used in this test, it throws an exception on
+    it's thread so we mock it here anyway.
```
khvn26: I don't understand this comment.

khvn26 (approved):

api/tests/unit/features/test_unit_features_models.py
zachaysan: I removed the mock and it looks like it passes without it, so I'll just delete this docstring and update the code. Thanks for checking it!

zachaysan: (Note I'm also removing the other similar test mocks in the module.)

Issue #444, opened 2021-10-23
Search Identities based on Trait key/value


dabeeeenster:
Need to design a way to index traits with the global data store. Some options:

- https://aws.amazon.com/about-aws/whats-new/2015/08/amazon-dynamodb-elasticsearch-integration/
- https://typesense.org/docs/guide/dynamodb-full-text-search.html

In SaaS, this is problematic due to the size of the data we are storing and would need to index. 

For self hosted/open source we should investigate using postgres to drive this.

landonjs:
We could really use something like this as well +1

Rednegniw:
Hi! I'd really love this feature as well.

matthewelwell:
Superseded by #4016

Issue #290, opened 2021-09-06
Add additional meta data to Identities


dabeeeenster:
Use case:

> Wondering if there's a way we can add something like notes or tags to identities in the Flagsmith UI that are not accessible via the non-secret environment key. The use case is that we want to be able to add actual customer names to identities so it's easier to toggle flags per customer, but we of course want it to be impossible to get that info via the env key. 

So these would need to be indexed and searchable as well.

kyle-ssg:
Needs fleshing out, assign to me when we've got a bit more of a spec.

matthewelwell:
Closed accidentally after being linked to an incorrect PR.

matthewelwell:
Superseded by #4016

PR #3898, opened 2024-05-08
chore: Identify by id

kyle-ssg:

## Changes

We're currently calling flagsmith.identify with email as the identity. This changes the behaviour to identify by user id.

## How did I test

Validated we are identifying by uid, checked identity existed in Staging.

### DIFF
--- a/frontend/web/project/api.js
+++ b/frontend/web/project/api.js
@@ -91,7 +91,7 @@ global.API = {
     }
 
     flagsmith
-      .identify(user.email, {
+      .identify(user.id, {
         email: user.email,
         organisations: user.organisations
           ? user.organisations.map((o) => `"${o.id}"`).join(',')


matthewelwell (commented):
This seems fine, but it really makes me want to implement [this](https://github.com/Flagsmith/flagsmith/issues/290) or [this](https://github.com/Flagsmith/flagsmith/issues/444) (but that's much harder!).

frontend/web/project/api.js
```diff
@@ -91,7 +91,7 @@ global.API = {
     }
 
     flagsmith
-      .identify(user.email, {
+      .identify(user.id, {
```
matthewelwell, line 94: I'd argue maybe this should be some concatenated string to avoid interspersion with the previous email based identities. Something like the following perhaps? 

```suggestion
      .identify(`user-${user.id}`, {
```

kyle-ssg: I was thinking we should delete them tbh @matthewelwell, it won't affect anything

matthewelwell: Yep, fair enough. I do still think that using a string concatenation like this is still a good idea, even if just for aesthetics.

kyle-ssg: I think your suggestion around having a user guid would be the best, if we want to do that I'll hold off on this PR until we do that @matthewelwell ?

PR #3899, opened 2024-05-08
fix: Validate if the GitHub installation is valid

novakzaballa:

## Changes

- Add an infoMessage if the GitHub installation is invalid.
- Add the option to update the installation id or delete the integration, if it is invalid.
- Correct issue when try link a GitHub PR.
- Add extra logic in the GitHubSetupPage

- Create an integration with GH in the Flagsmith App.
- Uninstall the app in GitHub
- Go to integration option in the side menu -> GitHub Integration -> Manage integration

### DIFF
--- a/frontend/web/components/App.js
+++ b/frontend/web/components/App.js
@@ -18,6 +18,7 @@ import { getStore } from 'common/store'
 import { resolveAuthFlow } from '@datadog/ui-extensions-sdk'
 import ConfigProvider from 'common/providers/ConfigProvider'
 import { getOrganisationUsage } from 'common/services/useOrganisationUsage'
+import { getGithubIntegration } from 'common/services/useGithubIntegration'
 import Button from './base/forms/Button'
 import Icon from './Icon'
 import AccountStore from 'common/stores/account-store'
@@ -38,10 +39,12 @@ const App = class extends Component {
   state = {
     activeOrganisation: 0,
     asideIsVisible: !isMobile,
+    githubMessage: '',
     lastEnvironmentId: '',
     lastProjectId: '',
     pin: '',
     showAnnouncement: true,
+    showGithubMessage: false,
   }
 
   constructor(props, context) {
@@ -53,6 +56,7 @@ const App = class extends Component {
     getBuildVersion()
     this.listenTo(ProjectStore, 'change', () => this.forceUpdate())
     this.listenTo(AccountStore, 'change', this.getOrganisationUsage)
+    this.getGithubIntegration()
     this.getOrganisationUsage()
     window.addEventListener('scroll', this.handleScroll)
     AsyncStorage.getItem('lastEnv').then((res) => {
@@ -74,9 +78,27 @@ const App = class extends Component {
       getOrganisationUsage(getStore(), {
         organisationId: AccountStore.getOrganisation()?.id,
       }).then((res) => {
+        if (AccountStore.getOrganisation().id) {
+          this.getGithubIntegration()
+        }
         this.setState({
           activeOrganisation: AccountStore.getOrganisation().id,
         })
+      })
+    }
+  }
+
+  getGithubIntegration = () => {
+    if (this.state.activeOrganisation) {
+      getGithubIntegration(getStore(), {
+        organisation_id: this.state.activeOrganisation,
+      }).then((res) => {
+        if (res.isError && res.error.data.status === 410) {
+          this.setState({
+            githubMessage: res.error.data.detail,
+            showGithubMessage: true,
+          })
+        }
       })
     }
   }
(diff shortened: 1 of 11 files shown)


frontend/web/components/App.js
```diff
@@ -53,6 +56,7 @@ const App = class extends Component {
     getBuildVersion()
     this.listenTo(ProjectStore, 'change', () => this.forceUpdate())
     this.listenTo(AccountStore, 'change', this.getOrganisationUsage)
+    this.getGithubIntegration()
```
kyle-ssg, line 59: Can we move GH stuff in app to a component. App needs to have less noise

frontend/web/components/GitHubRepositoriesSelect.tsx
```diff
@@ -73,6 +77,30 @@ const GitHubRepositoriesSelect: FC<GitHubRepositoriesSelectType> = ({
...
+          <Button
+            className='mr-3'
+            id='open-github-win-installations-btn'
+            data-test='open-github-win-installations-btn'
+            onClick={manageIntegration}
+            size='small'
+          >
+            Manage available GitHub Repositories
+          </Button>
+          <DeleteGithubIntegracion
```
kyle-ssg, line 95: DeleteGithubIntegration

frontend/web/components/GitHubRepositoriesSelect.tsx
```diff
@@ -73,6 +77,30 @@ const GitHubRepositoriesSelect: FC<GitHubRepositoriesSelectType> = ({
...
+            githubId={githubId}
+            organisationId={organisationId}
+            onConfirm={() => {
```
kyle-ssg, line 98: onComplete I assume

frontend/web/components/modals/InvalidGithubIntegration.tsx
```diff
@@ -0,0 +1,82 @@
...
+      <div className='modal-body'>
+        <ErrorMessage error={errorMessage} />
+        <InputGroup
+          title='New Installation ID'
+          inputProps={{
+            className: 'full-width',
+            name: 'installationId',
+          }}
+          value={installationId}
+          onChange={(event: React.ChangeEvent<HTMLInputElement>) => {
```
kyle-ssg, line 48: I really don't understand what this is doing, why would we be typing an installation id ?

frontend/web/components/modals/CreateEditIntegrationModal.js
```diff
@@ -233,29 +231,8 @@ const CreateEditIntegration = class extends Component {
...
-                  >
-                    Manage available GitHub Repositories
-                  </Button>
-                  <DeleteGithubIntegracion
-                    githubId={this.props.githubMeta.githubId}
-                    organisationId={AccountStore.getOrganisation().id}
-                    onConfirm={() => {
-                      closeModal()
-                    }}
+                    manageIntegration={this.openGitHubWinInstallations}
```
kyle-ssg, line 234: openGitHubWithInstallations

frontend/web/components/InfoMessage.js
```diff
@@ -8,7 +8,11 @@ export default class InfoMessage extends PureComponent {
   static displayName = 'InfoMessage'
 
   handleOpenNewWindow = () => {
-    window.open(this.props.url, '_blank')
+    if (this.props.goToIntegrations) {
```
kyle-ssg, line 11: Let's come up with a better solution than adding goToIntegrations as a prop to a generic component like InfoMessage, it doesn't make sense.

frontend/web/components/App.js
```diff
@@ -474,6 +496,28 @@ const App = class extends Component {
                               id={AccountStore.getOrganisation()?.id}
                             />
                           )}
+                          {user && this.state.showGithubMessage && (
+                            <Row className={'px-3'}>
```
kyle-ssg, line 500: RE above comment, we can just put all the changes and markup into ``<GitHubMessage/>``

kyle-ssg:
Reading through the purpose of this PR, I just want to make sure this is the best solution. There must be a way to handle this without UI?

E.g.

- Do GitHub Apps have WebHooks that detect uninstall ? If so we could invalidate the installation ID
- Is there a way to get a GH installation ID ? If so we could attempt to fetch it and invalidate if it doesnt exist

How does e.g. Vercel handle this ? 

Am I right in saying we could just subscribe to this https://docs.github.com/en/webhooks/webhook-events-and-payloads#github_app_authorization

frontend/web/components/modals/InvalidGithubIntegration.tsx
novakzaballa: That's for the case they want to keep the existing links between PR/Issue and feature flags, while they updated for some reason the GH App Installation ID, i.e. they removed the integration from GH accidentally. Maybe too much for an edge case?

novakzaballa:
> Reading through the purpose of this PR, I just want to make sure this is the best solution. There must be a way to handle this without UI?
> 
> E.g.
> 
> * Do GitHub Apps have WebHooks that detect uninstall ? If so we could invalidate the installation ID
> * Is there a way to get a GH installation ID ? If so we could attempt to fetch it and invalidate if it doesnt exist
> 
> How does e.g. Vercel handle this ?
> 
> Am I right in saying we could just subscribe to this https://docs.github.com/en/webhooks/webhook-events-and-payloads#github_app_authorization

Yes, that is the plan but it was not part of the scope of this first version. However, maybe it is worth it to implement the webhook instead of this. Please let's discuss.

novakzaballa:
> * Do GitHub Apps have WebHooks that detect uninstall ? If so we could invalidate the installation ID

Yep, 

> * Is there a way to get a GH installation ID ? If so we could attempt to fetch it and invalidate if it doesnt exist
> 
> How does e.g. Vercel handle this ?
> 
> Am I right in saying we could just subscribe to this https://docs.github.com/en/webhooks/webhook-events-and-payloads#github_app_authorization

>

PR #3902, opened 2024-05-08
fix: send all users on paid subscriptions to hubspot

matthewelwell:

## Changes

Previously we filtered out users that were part of an organisation that had a paid subscription before sending them to hubspot. This PR removes that filtering. 

I haven't, but it's just removing a few lines of trivial code which gets called in a single location.


zachaysan (approved):
Looks good, but we should have had test coverage for this.

PR #3905, opened 2024-05-08
fix: Move call to GitHub integration tasks out from trigger_feature_state_change_webhooks

novakzaballa:

## Changes
- Move call to GitHub integration tasks out from trigger_feature_state_change_webhooks_signal for compatibility with versioning v2

- Create a GitHub integration, add any repository, link a feature with a GH issue, and update the feature values, show the comment in your GH issue

### DIFF
--- a/api/features/models.py
+++ b/api/features/models.py
@@ -74,7 +74,7 @@ from features.value_types import (
     STRING,
 )
 from features.versioning.models import EnvironmentFeatureVersion
-from organisations.models import Organisation
+from integrations.github.models import GithubConfiguration
 from projects.models import Project
 from projects.tags.models import Tag
 
@@ -1007,10 +1007,8 @@ class FeatureState(
             and self.environment.project.github_project.exists()
             and self.environment.project.organisation.github_config.exists()
         ):
-            github_configuration = (
-                Organisation.objects.prefetch_related("github_config")
-                .get(id=self.environment.project.organisation_id)
-                .github_config.first()
+            github_configuration = GithubConfiguration.objects.get(
+                organisation_id=self.environment.project.organisation_id
             )
             feature_states = []
             feature_states.append(self)
--- a/api/features/models.py
+++ b/api/features/models.py
@@ -1013,13 +1013,23 @@ class FeatureState(
             feature_states = []
             feature_states.append(self)
 
-            feature_data: GithubData = generate_data(
-                github_configuration=github_configuration,
-                feature_id=self.feature.id,
-                feature_name=self.feature.name,
-                type=WebhookEventType.FLAG_UPDATED.value,
-                feature_states=feature_states,
-            )
+            if self.deleted_at is None:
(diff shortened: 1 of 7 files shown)


api/features/signals.py
```diff
@@ -12,6 +18,34 @@
 
 @receiver(post_save, sender=FeatureState)
 def trigger_feature_state_change_webhooks_signal(instance, **kwargs):
```
khvn26: I'd prefer this code to reside in a separate receiver function rather than adding to an existing one.

Additionally, consider using an `AFTER_SAVE` hook instead.

api/features/signals.py
```diff
@@ -12,6 +18,34 @@
...
+    # Enqueue Github integration tasks if feature has GitHub external resources
+    if (
+        not instance.identity_id
+        and not instance.feature_segment
+        and instance.feature.external_resources.exists()
+        and instance.environment.project.github_project.exists()
+        and hasattr(instance.environment.project.organisation, "github_config")
```
khvn26: I'm not sure `hasattr` is used correctly here.

Also I'd love to see a comment explaining the lengthy condition.

khvn26 (changes requested):
Left two comments. Also, the code needs test coverage.

api/features/signals.py
khvn26:
```suggestion
        and instance.environment.project.organisation.github_config
```

or, to be more explicit

```suggestion
        and instance.environment.project.organisation.github_config is not None
```

novakzaballa: This doesn't work in this case, using exists() instead.

api/features/models.py
```diff
@@ -989,6 +992,41 @@ def copy_identity_feature_states(
...
+            and not self.feature_segment
+            and self.feature.external_resources.exists()
+            and self.environment.project.github_project.exists()
+            and self.environment.project.organisation.github_config.exists()
+        ):
+            github_configuration = (
+                Organisation.objects.prefetch_related("github_config")
+                .get(id=self.environment.project.organisation_id)
+                .github_config.first()
+            )
```
matthewelwell: Err... what? Why not just: 

```suggestion
            github_configuration = (
                GithubConfig.objects.get(organisation_id=self.environment.project.organisation_id)
            )
```

matthewelwell (changes requested):

api/features/signals.py
novakzaballa: Yeah. Moving the code to an AFTER_SAVE hook

api/features/models.py
novakzaballa: Long story, but basically I had issues with the `soft-delete` model and the `one-to-one` relationship and other things, but that complex query is not needed. Removed

khvn26 (approved):

matthewelwell (approved):

PR #3906, opened 2024-05-09
feat: Implement GitHub Webhook

novakzaballa:

## Changes
Implement GitHub Webhook to receive events from GitHub App.
Initially we only handle the Integration Delete action. We delete the configuration and any related object in the DB when an installation is removed from GitHub.

Also, the env var `GITHUB_WEBHOOK_SECRET` was added to ECS task definition web secrets. This env var is not used by the tasks executed by the task processor. It is only used by the GitHub webhook.

Once we generate the `GITHUB_WEBHOOK_SECRET` we need to update our GitHub App to use it as shown here
image

https://github.com/organizations/Flagsmith/settings/apps/flagsmith

Unit tests added.

### DIFF
--- a/api/integrations/github/views.py
+++ b/api/integrations/github/views.py
@@ -1,3 +1,6 @@
+import hashlib
+import hmac
+import json
 import re
 from functools import wraps
 
@@ -8,7 +11,7 @@ from django.http import JsonResponse
 from rest_framework import status, viewsets
 from rest_framework.decorators import api_view, permission_classes
 from rest_framework.exceptions import ValidationError
-from rest_framework.permissions import IsAuthenticated
+from rest_framework.permissions import AllowAny, IsAuthenticated
 from rest_framework.response import Response
 
 from integrations.github.client import generate_token
@@ -23,6 +26,22 @@ from integrations.github.serializers import (
 )
 from organisations.models import Organisation
 from organisations.permissions.permissions import GithubIsAdminOrganisation
+
+
+def github_webhook_payload_is_valid(payload_body, secret_token, signature_header):
+    """Verify that the payload was sent from GitHub by validating SHA256.
+    Raise and return 403 if not authorized.
+    """
+    if not signature_header:
+        return False
+    hash_object = hmac.new(
+        secret_token.encode("utf-8"), msg=payload_body, digestmod=hashlib.sha1
+    )
+    expected_signature = "sha1=" + hash_object.hexdigest()
+    if not hmac.compare_digest(expected_signature, signature_header):
+        return False
+
+    return True
 
 
 def github_auth_required(func):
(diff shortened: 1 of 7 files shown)


api/integrations/github/views.py
```diff
@@ -25,6 +28,22 @@
 from organisations.permissions.permissions import GithubIsAdminOrganisation
 
 
+def github_webhook_payload_is_valid(payload_body, secret_token, signature_header):
```
gagantrivedi: Type hints

api/integrations/github/views.py
```diff
@@ -204,3 +223,27 @@ def fetch_repositories(request, organisation_pk: int):
         return Response(data)
     except requests.RequestException as e:
         return JsonResponse({"error": str(e)}, status=500)
+
+
+@api_view(["POST"])
+@permission_classes([AllowAny])
+def github_webhook(request) -> JsonResponse:
+    # Validate payload signature
```
gagantrivedi: I don't thik his comment is needed

api/integrations/github/views.py
gagantrivedi: Also, this function should not be part of views

api/integrations/github/views.py
```diff
@@ -204,3 +223,27 @@ def fetch_repositories(request, organisation_pk: int):
...
+    if github_webhook_payload_is_valid(
+        payload_body=payload, secret_token=secret, signature_header=signature
+    ):
+        data = json.loads(payload.decode("utf-8"))
+        # handle GitHub Webhook "installation" event with action type "deleted"
+        if github_event == "installation" and data["action"] == "deleted":
+            GithubConfiguration.objects.filter(
+                installation_id=data["installation"]["id"]
+            ).delete()
+            return JsonResponse({"detail": "Event processed"}, status=200)
```
gagantrivedi: Is there a reason why we are not using `Response` here?

gagantrivedi (changes requested):

api/integrations/github/views.py
novakzaballa: I put it there because It's a helper used to validate a request header so it looked very view-specific for me.  Where would you suggest to move it?

api/integrations/github/views.py
novakzaballa: Not Really I will change this and other views to use Response

novakzaballa: Not Really I will change this and other views to use Response

api/integrations/github/views.py
gagantrivedi: signature.py

infrastructure/aws/production/ecs-task-definition-web.json
```diff
@@ -245,6 +245,10 @@
                     "name": "SSE_AUTHENTICATION_TOKEN",
                     "valueFrom": "arn:aws:secretsmanager:eu-west-2:084060095745:secret:ECS-API-LxUiIQ:SSE_AUTHENTICATION_TOKEN::"
                 },
+                {
```
gagantrivedi, line 248: Have we added these secrete?

novakzaballa: Not yet, I don't have the credentials or authorization to do it. Could you?

gagantrivedi: Have we added it to staging?

novakzaballa: I added instructions in the description on how to update the GitHub App once we have a secret defined and set.

gagantrivedi: But we can't merge it without it, right?

api/integrations/github/views.py
novakzaballa: Done. Moved to `helpers.py`

api/integrations/github/views.py
novakzaballa: Done

api/integrations/github/views.py
novakzaballa: Removed

infrastructure/aws/production/ecs-task-definition-web.json
novakzaballa: For another env var added @matthewelwell asked me to do first the PR declaring the env var in the infrastructure folder. And he set the value after that, however, I would say that he has the last word.

gagantrivedi: Yeah, that makes sense, but we still can't merge it before the secrets are added because the deployment will fail

novakzaballa: @matthewelwell have set the env var secret.

gagantrivedi (approved):

PR #3913, opened 2024-05-10
fix(versioning): handle mapping of environment to engine post v2 versioning migration

matthewelwell:

## Changes

Fixes issue mapping environment to engine environment post v2 versioning merge. 

Note that another option here would be to change the mapping function to use e.g. the `versioning_service.get_environment_feature_states` function, but this would result in database query issues. 

Added a specific unit test to replicate the issue seen in production.

### DIFF
--- a/api/tests/unit/util/mappers/test_unit_mappers_engine.py
+++ b/api/tests/unit/util/mappers/test_unit_mappers_engine.py
@@ -3,6 +3,7 @@ from typing import TYPE_CHECKING
 
 import pytest
 import pytz
+from django.utils import timezone
 from flag_engine.environments.integrations.models import IntegrationModel
 from flag_engine.environments.models import (
     EnvironmentAPIKeyModel,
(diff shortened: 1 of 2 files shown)


api/tests/unit/util/mappers/test_unit_mappers_engine.py
```diff
@@ -533,3 +535,44 @@ def test_map_environment_to_engine__returns_correct_feature_state_for_different_
...
+    assert result
+
+    assert len(result.feature_states) == 1
+    assert result.feature_states[0].django_id == v2_environment_feature_state.id
+
+    assert len(result.project.segments) == 1
+    assert len(result.project.segments[0].feature_states) == 1
+    assert (
+        result.project.segments[0].feature_states[0].django_id == v2_segment_override.id
+    )
```
novakzaballa, line 592: My only question is why we are not checking the values but only the IDs? I guess the answer is they are being checked in another unit test?

matthewelwell: Yeah, exactly. I just want to make sure here that the correct versioned feature state is being returned. The value will be fixed with that version so, as long as the id is correct, the value will be correct. I'll happily add some checks in here about the state / value though.

matthewelwell: Done ☑️

novakzaballa (approved):
LGTM

PR #3916, opened 2024-05-10
fix(versioning): fix exception getting feature states for edge identity post v2 versioning migration

matthewelwell:

## Changes

Fixes an issue where retrieving the flags for an Edge Identity following a v2 versioning migration in an environment which had previously used the change requests functionality caused an exception. 

Reproduced the issue with a unit test that I was then able to get to pass following the changes. Existing unit tests should cover the refactoring that was included as part of this PR.

### DIFF
--- a/api/tests/unit/edge_api/identities/test_edge_identity_models.py
+++ b/api/tests/unit/edge_api/identities/test_edge_identity_models.py
@@ -9,7 +9,9 @@ from freezegun import freeze_time
 from pytest_mock import MockerFixture
 
 from edge_api.identities.models import EdgeIdentity
-from features.models import FeatureSegment, FeatureState, FeatureStateValue
+from environments.models import Environment
+from features.models import Feature, FeatureSegment, FeatureState
+from features.versioning.tasks import enable_v2_versioning
 from features.workflows.core.models import ChangeRequest
 from segments.models import Segment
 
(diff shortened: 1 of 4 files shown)


api/tests/unit/edge_api/identities/test_edge_identity_models.py
```diff
@@ -38,17 +40,10 @@ def test_get_all_feature_states_for_edge_identity_uses_segment_priorities(
...
+    FeatureState.objects.create(
         feature=feature, environment=environment, feature_segment=feature_segment_p2
     )
 
-    FeatureStateValue.objects.filter(feature_state=segment_override_p1).update(
-        string_value="p1"
-    )
-    FeatureStateValue.objects.filter(feature_state=segment_override_p2).update(
-        string_value="p2"
-    )
```
matthewelwell, line 50: These seemed superfluous since we're using the feature state's ID for the assertion below.

zachaysan (approved):
The code looks good, but I wonder if existing code has any `django_assert_num_queries` against it. If not, would be good to get some coverage.

matthewelwell:
> The code looks good, but I wonder if existing code has any `django_assert_num_queries` against it. If not, would be good to get some coverage.

There are certainly some sprinkled throughout the test suite on the important code paths that will use these functions, however, I have added some additional uses of the fixture in on some specific unit tests to also verify things.

PR #3948, opened 2024-05-14
fix: Create GitHub comment as table

novakzaballa:

## Changes

- Improve GitHub comment to look like a table

image

- Some unit tests updated.
- Link an external resource manually.

### DIFF
--- a/api/integrations/github/constants.py
+++ b/api/integrations/github/constants.py
@@ -3,12 +3,11 @@ GITHUB_API_VERSION = "2022-11-28"
 
 LINK_FEATURE_TITLE = """**Flagsmith feature linked:** `%s`
 Default Values:\n"""
-FEATURE_TABLE_HEADER = """| Environment | Enabled | Value | Type | Updated (UTC) |
-| :--- | :----- | :------ | :------- | :------ |\n"""
-FEATURE_TABLE_ROW = "| [%s](%s) | %s | `%s` | %s | %s |\n"
+FEATURE_TABLE_HEADER = """| Environment | Enabled | Value | Last Updated (UTC) |
+| :--- | :----- | :------ | :------ |\n"""
+FEATURE_TABLE_ROW = "| [%s](%s) | %s | %s | %s |\n"
 LINK_SEGMENT_TITLE = "Segment `%s` values:\n"
 UNLINKED_FEATURE_TEXT = "### The feature flag `%s` was unlinked from the issue/PR"
 UPDATED_FEATURE_TEXT = "Flagsmith Feature `%s` has been updated:\n"
-LAST_UPDATED_FEATURE_TEXT = "Last Updated %s"
 DELETED_FEATURE_TEXT = "### The Feature Flag `%s` was deleted"
 FEATURE_ENVIRONMENT_URL = "%s/project/%s/environment/%s/features?feature=%s&tab=%s"
--- a/api/integrations/github/constants.py
+++ b/api/integrations/github/constants.py
@@ -1,8 +1,18 @@
+from core.helpers import get_current_site_url
+
 GITHUB_API_URL = "https://api.github.com/"
 GITHUB_API_VERSION = "2022-11-28"
 
-LINK_FEATURE_TEXT = "### This pull request is linked to a Flagsmith Feature (`%s`):\n"
+LINK_FEATURE_TITLE = """**Flagsmith feature linked:** `%s`
+Default Values:\n"""
+FEATURE_TABLE_HEADER = """| Environment | Enabled | Value | Type | Updated (UTC) |
+| :--- | :----- | :------ | :------- | :------ |\n"""
+FEATURE_TABLE_ROW = "| [%s](%s) | %s | `%s` | %s | %s |\n"
+LINK_SEGMENT_TITLE = "Segment `%s` values:\n"
 UNLINKED_FEATURE_TEXT = "### The feature flag `%s` was unlinked from the issue/PR"
-UPDATED_FEATURE_TEXT = "### The Flagsmith Feature `%s` was updated in the environment "
+UPDATED_FEATURE_TEXT = "Flagsmith Feature `%s` has been updated:\n"
 LAST_UPDATED_FEATURE_TEXT = "Last Updated %s"
 DELETED_FEATURE_TEXT = "### The Feature Flag `%s` was deleted"
+FEATURE_ENVIRONMENT_URL = (
+    get_current_site_url() + "/project/%s/environment/%s/features?feature=%s&tab=%s"
+)
--- a/api/integrations/github/constants.py
+++ b/api/integrations/github/constants.py
@@ -1,5 +1,3 @@
-from core.helpers import get_current_site_url
-
 GITHUB_API_URL = "https://api.github.com/"
 GITHUB_API_VERSION = "2022-11-28"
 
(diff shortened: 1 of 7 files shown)


api/integrations/github/constants.py
```diff
@@ -1,8 +1,14 @@
 GITHUB_API_URL = "https://api.github.com/"
 GITHUB_API_VERSION = "2022-11-28"
 
-LINK_FEATURE_TEXT = "### This pull request is linked to a Flagsmith Feature (`%s`):\n"
+LINK_FEATURE_TITLE = """**Flagsmith feature linked:** `%s`
+Default Values:\n"""
+FEATURE_TABLE_HEADER = """| Environment | Enabled | Value | Type | Updated (UTC) |
```
khvn26:
```suggestion
FEATURE_TABLE_HEADER = """| Environment | Enabled | Value | Type | Last Updated (UTC) |
```

This way, we won't have to prepend "Last Updated" to every timestamp in this column.

api/integrations/github/github.py
```diff
@@ -81,22 +89,37 @@ def generate_body_comment(
     last_updated_string = LAST_UPDATED_FEATURE_TEXT % (
         datetime.datetime.now().strftime("%dth %b %Y %I:%M%p")
     )
```
khvn26: I'm not a fan of the custom datetime format.

Let's consider either sticking to ISO8601 or using Django locale, i.e.:

```python
from django.utils.formats import get_format

datetime_format = get_format('DATETIME_INPUT_FORMATS')[0]
```

api/integrations/github/github.py
```diff
@@ -81,22 +89,37 @@ def generate_body_comment(
...
+        if v.get("segment_name") is not None and v["segment_name"] != last_segment_name:
+            result += "\n" + LINK_SEGMENT_TITLE % (v["segment_name"])
+            last_segment_name = v["segment_name"]
+            result += FEATURE_TABLE_HEADER
+        table_row = FEATURE_TABLE_ROW % (
+            v["environment_name"],
+            environment_link_url,
+            "✅ Enabled" if v["feature_value"] else "❌ Disabled",
+            feature_value if feature_value else "",
+            feature_value_type,
```
khvn26: I don't think we should surface feature value type here:

1. We plan to move away from feature value types to only support strings in the future.
2. String value type awkwardly gets shown as "Unicode" which, ideally, should be replaced by "String" but it's messy and I'd rather skip the whole ordeal.

api/integrations/github/github.py
```diff
@@ -81,22 +89,37 @@ def generate_body_comment(
...
+        )
+        if v.get("segment_name") is not None and v["segment_name"] != last_segment_name:
+            result += "\n" + LINK_SEGMENT_TITLE % (v["segment_name"])
+            last_segment_name = v["segment_name"]
+            result += FEATURE_TABLE_HEADER
+        table_row = FEATURE_TABLE_ROW % (
+            v["environment_name"],
+            environment_link_url,
+            "✅ Enabled" if v["feature_value"] else "❌ Disabled",
+            feature_value if feature_value else "",
```
khvn26: Empty feature value is currently displayed as two backticks (``). I'd like to avoid this.

khvn26 (changes requested):

api/integrations/github/github.py
```diff
@@ -81,22 +89,37 @@ def generate_body_comment(
...
+            tab,
+        )
+        if v.get("segment_name") is not None and v["segment_name"] != last_segment_name:
+            result += "\n" + LINK_SEGMENT_TITLE % (v["segment_name"])
+            last_segment_name = v["segment_name"]
+            result += FEATURE_TABLE_HEADER
+        table_row = FEATURE_TABLE_ROW % (
+            v["environment_name"],
+            environment_link_url,
+            "✅ Enabled" if v["feature_value"] else "❌ Disabled",
```
khvn26: I don't think the logic here is correct. Disabled feature states can have values.

I'm not entirely sure where the feature state data comes from, but, assuming it has the "enabled" key, the code should be:

```suggestion
            "✅ Enabled" if v["enabled"] else "❌ Disabled",
```

khvn26:
Please cover the code reported as uncovered with tests, or add a `# pragma: no cover` annotation, preferably with an explanation why.

api/integrations/github/constants.py
novakzaballa: Done.

api/integrations/github/github.py
novakzaballa: Done.

api/integrations/github/github.py
novakzaballa: Done.

api/integrations/github/github.py
novakzaballa: Done.

api/integrations/github/github.py
novakzaballa: The logic is correct, but the variable name is confusing. Variable renamed

novakzaballa:
> Please cover the code reported as uncovered with tests, or add a `# pragma: no cover` annotation, preferably with an explanation why.

I ended up adding the coverage. Let's keep this coverage while it is possible and is worth it.

api/tests/unit/features/test_unit_feature_external_resources_views.py
```diff
@@ -41,20 +40,31 @@ def json(self):
...
+
+    expected_comment_body = (
+        "**Flagsmith feature linked:** `feature_with_value`\n"
+        + "Default Values:\n"
+        + "| Environment | Enabled | Value | Last Updated (UTC) |\n"
+        + "| :--- | :----- | :------ | :------ |\n"
+        + f"| [Test Environment](https://example.com/project/{project.id}/"
+        + f"environment/{environment.api_key}/features?feature={feature_with_value.id}&tab=value) "
+        + "| ❌ Disabled | `value` | 2024-01-01 00:00:00 |\n"
+    )
```
khvn26: This is nice! nit: Python supports multiline strings so you could do this:

```suggestion
    expected_comment_body = (
        "**Flagsmith feature linked:** `feature_with_value`\n"
        "Default Values:\n"
        "| Environment | Enabled | Value | Last Updated (UTC) |\n"
        "| :--- | :----- | :------ | :------ |\n"
        f"| [Test Environment](https://example.com/project/{project.id}/"
        f"environment/{environment.api_key}/features?feature={feature_with_value.id}&tab=value) "
        "| ❌ Disabled | `value` | 2024-01-01 00:00:00 |\n"
    )
```

Looks a bit nicer this way, IMO.

api/integrations/github/github.py
```diff
@@ -78,25 +86,38 @@ def generate_body_comment(
...
+            get_current_site_url(),
+            project_id,
+            v.get("environment_api_key"),
+            feature_id,
+            tab,
+        )
+        if v.get("segment_name") is not None and v["segment_name"] != last_segment_name:
+            result += "\n" + LINK_SEGMENT_TITLE % (v["segment_name"])
+            last_segment_name = v["segment_name"]
+            result += FEATURE_TABLE_HEADER
```
khvn26, line 105: We should add a segment to the test or slap a `# pragma: no cover` on L108.

api/tests/unit/features/test_unit_feature_external_resources_views.py
novakzaballa: Done.

api/integrations/github/github.py
novakzaballa: Done.

khvn26 (approved):

PR #3959, opened 2024-05-15
fix(versioning): handle Master API Keys when publishing a version

matthewelwell:

## Changes

Allow EnvironmentFeatureVersion objects to be published using a MasterAPIKey. 

Added a new unit test. Note that I tried using the new `admin_client_new` fixture, but I decided against it since it would have meant adding some conditional logic to the assert regarding the `published_by` attribute (which is set to `None`, and hence interpreted as 'System' when using a MasterAPIKey).

### DIFF
--- a/api/tests/unit/features/versioning/test_unit_versioning_views.py
+++ b/api/tests/unit/features/versioning/test_unit_versioning_views.py
@@ -26,10 +26,10 @@ tomorrow = now + timedelta(days=1)
 
 
 def test_get_versions_for_a_feature_and_environment(
-    admin_client: "APIClient",
-    admin_user: "FFAdminUser",
+    admin_client: APIClient,
+    admin_user: FFAdminUser,
     environment_v2_versioning: Environment,
-    feature: "Feature",
+    feature: Feature,
 ) -> None:
     # Given
     url = reverse(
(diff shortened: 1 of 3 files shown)


api/tests/unit/features/versioning/test_unit_versioning_views.py
```diff
@@ -194,6 +194,44 @@ def test_publish_feature_version(
     )
 
 
+@pytest.mark.parametrize("live_from", (None, tomorrow))
+def test_publish_feature_version_using_master_api_key(
+    admin_master_api_key_client: "APIClient",
+    environment_v2_versioning: Environment,
+    feature: "Feature",
+    live_from: typing.Optional[datetime],
```
zachaysan: This is minor, but given that this is a test module shouldn't we be able to do the following?
```suggestion
    admin_master_api_key_client: APIClient,
    environment_v2_versioning: Environment,
    feature: Feature,
    live_from: datetime | None,
```

zachaysan (approved):
One minor question, but the rest looks good

api/tests/unit/features/versioning/test_unit_versioning_views.py
matthewelwell: Yes, fair point, I can do that.

matthewelwell: Done (although I didn't update the rest of the quoted uses in the test module as I didn't want to dilute this PR). I think it's worth doing that in a separate PR.

matthewelwell: Scrap that, I've updated the rest of the file here too.

PR #3962, opened 2024-05-15
feat: Improvements in the GitHub integration BE

novakzaballa:

## Changes

- Add delete request to GitHub endpoint: `/app/installations/{installation_id}`
- Fix flaky unit test
- Move GitHub API requests to `integration/github/client.py` file.
- Obtaining the name and state of the GitHub resource. 

-  Unit test added and updated


api/integrations/github/client.py
```diff
@@ -12,3 +12,12 @@ def generate_token(installation_id: str, app_id: int) -> str:
     Github(auth=auth)
     token = auth.token
     return token
+
+
+def generate_jwt_token(app_id: int) -> str:  # pragma: no cover
```
khvn26, line 60: I'd prefer to mock `github_auth.create_jwt` in the test instead of pushing uncovered code.

api/integrations/github/client.py
```diff
@@ -12,3 +12,12 @@ def generate_token(installation_id: str, app_id: int) -> str:
...
+    github_auth: Auth.AppAuth = Auth.AppAuth(
+        app_id=int(app_id),
```
khvn26: According to function typing annotation, `app_id` is already `int`.

api/integrations/github/client.py
khvn26, line 60: nit: Naming of this function could be better. `generate` verb assumes Core API generates something on its end.

```suggestion
def get_jwt_token(app_id: int) -> str:  # pragma: no cover
```

api/tests/unit/integrations/github/test_unit_github_views.py
```diff
@@ -19,6 +19,18 @@
...
+def mocked_requests_delete(*args, **kwargs):
+    class MockResponse:
+        def __init__(self, json_data, status_code):
+            self.json_data = json_data
+            self.status_code = status_code
+
+        def raise_for_status(self) -> None:
+            pass
+
+    return MockResponse(json_data={"data": "data"}, status_code=204)
```
khvn26: Is there a reason to keep this at module level instead of having this code inside the test?

api/tests/unit/integrations/github/test_unit_github_views.py
```diff
@@ -114,10 +127,18 @@ def test_delete_github_configuration(
             github_configuration.id,
         ],
     )
+
+    mock_generate_token = mocker.patch(
+        "integrations.github.views.generate_jwt_token",
+    )
+    mock_generate_token.return_value = "mocked_token"
+    mocker.patch("requests.delete", side_effect=mocked_requests_delete)
```
khvn26: Please use the `responses` fixture to generate a response.

See example usage [here](https://github.com/Flagsmith/flagsmith/blob/e8d133726c5b724057175bfc7d945e12a733f9d1/api/tests/unit/integrations/flagsmith/test_unit_flagsmith_service.py#L98-L103).

api/tests/unit/integrations/github/test_unit_github_views.py
khvn26: I don't think this is needed at all, see [this comment](https://github.com/Flagsmith/flagsmith/pull/3962/files#r1603266828).

api/integrations/github/views.py
```diff
@@ -71,6 +71,23 @@ def create(self, request, *args, **kwargs):
...
+
+        url = f"{GITHUB_API_URL}app/installations/{installation_id}"
+
+        headers = {
+            "X-GitHub-Api-Version": GITHUB_API_VERSION,
+            "Accept": "application/vnd.github.v3+json",
+            "Authorization": f"Bearer {token}",
+        }
+
+        response = requests.delete(url, headers=headers, timeout=10)
```
khvn26: nit: This is okay, but I'd very much prefer to have a dedicated interface in the `github.github` module for this, rather than having all this code right here in the viewset.

api/integrations/github/views.py
```diff
@@ -71,6 +71,23 @@ def create(self, request, *args, **kwargs):
...
+        response.raise_for_status()
```
khvn26: We're assuming we're returning 500 in case of any errors from Github here?

Do we want to return the error message to the client or at least log it in some way?

khvn26 (changes requested):
Requested a couple of changes, see comments.

api/integrations/github/client.py
novakzaballa: We also would need to mock `Auth.AppAuth` in which case we would not be covering that function since those are its only two lines of code. For transparency, I'd prefer to mark that portion of code as uncovered.

api/integrations/github/client.py
novakzaballa: Right, since we are not yet using MyPy or another runtime type checker I would argue that, but in this case, it is guaranteed it will be an int so I am removing the explicit casting.

api/integrations/github/client.py
novakzaballa: In this context is valid since we are inside the integrations.github App and in its client module.

api/tests/unit/integrations/github/test_unit_github_views.py
novakzaballa: Right. Removed at all in favor of the simpler solution you recommended.

api/tests/unit/integrations/github/test_unit_github_views.py
novakzaballa: Done. Thanks!

api/integrations/github/views.py
novakzaballa: Good catch. It is not OK, it should go into the business logic layer. I am moving all this kind of business logic out from the views and into the github.client or github.github modules.

api/integrations/github/views.py
novakzaballa: Right, good catch. Done.

api/integrations/github/client.py
```diff
@@ -1,8 +1,41 @@
...
+    PULL_REQUESTS = "pulls"
+
+
+def build_request_headers(installation_id: str, useJWT: bool = False) -> dict[str, str]:
+    token = (
+        generate_token(
+            installation_id,
+            settings.GITHUB_APP_ID,
+        )
+        if not useJWT
```
khvn26: Python names should be in snake_case:

```suggestion
def build_request_headers(installation_id: str, use_jwt: bool = False) -> dict[str, str]:
    token = (
        generate_token(
            installation_id,
            settings.GITHUB_APP_ID,
        )
        if not use_jwt
```

api/integrations/github/client.py
khvn26: Well, we'll still be verifying that the `AppAuth` interface is used correctly — it's won't be a meaningless test if you use `autospec` with the mocks.

I'm conscious of the time constraints and agree that there's not too much to cover here, so I'll resolve this, but I'll appreciate a `TODO` near that `pragma: no cover`.

api/integrations/github/client.py
```diff
@@ -12,3 +45,66 @@ def generate_token(installation_id: str, app_id: int) -> str:
...
+    return response
+
+
+def fetch_github_resource(
+    resource_type: ResourceType, organisation_id: int, repo_owner: str, repo_name: str
+) -> requests.Response:
+    organisation = Organisation.objects.get(id=organisation_id)
+    github_configuration = GithubConfiguration.objects.get(
+        organisation=organisation, deleted_at__isnull=True
+    )
```
khvn26: What's the point for an extra select here?

```suggestion
    github_configuration = GithubConfiguration.objects.get(
        organisation_id=organisation_id, deleted_at__isnull=True
    )
```

api/integrations/github/client.py
```diff
@@ -12,3 +45,66 @@ def generate_token(installation_id: str, app_id: int) -> str:
...
+    url = f"{GITHUB_API_URL}repos/{repo_owner}/{repo_name}/{resource_type.value}"
+    headers: dict[str, str] = build_request_headers(
+        github_configuration.installation_id
+    )
+    response = requests.get(url, headers=headers, timeout=10)
```
khvn26: I'd like to see a constant, preferably configurable, for that `timeout` argument value, or not use a custom timeout at all.

api/integrations/github/client.py
```diff
@@ -12,3 +45,66 @@ def generate_token(installation_id: str, app_id: int) -> str:
...
+    response.raise_for_status()
+    return response
+
+
+def fetch_github_repositories(installation_id: str) -> requests.Response:
+    url = f"{GITHUB_API_URL}installation/repositories"
+
+    headers: dict[str, str] = build_request_headers(installation_id)
+
+    response = requests.get(url, headers=headers, timeout=10)
```
khvn26: Same as [here](https://github.com/Flagsmith/flagsmith/pull/3962/files#r1605248710).

api/integrations/github/client.py
novakzaballa: Ooops, right

api/integrations/github/client.py
```diff
@@ -12,3 +45,66 @@ def generate_token(installation_id: str, app_id: int) -> str:
...
+    response.raise_for_status()
+
+    return response.json()
+
+
+def delete_github_installation(installation_id: str) -> requests.Response:
+    url = f"{GITHUB_API_URL}app/installations/{installation_id}"
+    headers = build_request_headers(installation_id, useJWT=True)
+    response = requests.delete(url, headers=headers, timeout=10)
+    return response
```
khvn26, line 89: Why no `raise_for_status()` here?

khvn26 (changes requested):

api/integrations/github/views.py
```diff
@@ -71,6 +99,22 @@ def create(self, request, *args, **kwargs):
             if re.search(r"Key \(organisation_id\)=\(\d+\) already exists", str(e)):
                 raise DuplicateGitHubIntegration
 
+    def destroy(self, request, *args, **kwargs):
+        github_response = delete_github_installation(self.get_object().installation_id)
+        if github_response.status_code != 204:
+            return Response(
```
khvn26: Why no decorator here? Related to [this comment](https://github.com/Flagsmith/flagsmith/pull/3962/files#r1605261746).

api/integrations/github/client.py
novakzaballa: Good point. I based most of my work on the existing Webhook code here https://github.com/Flagsmith/flagsmith/blob/feat/uninstall-github-app-installation-from-flagsmith/api/webhooks/webhooks.py#L162-L165
I think we should also consider your observation there.

khvn26: We absolutely should, good call.

api/integrations/github/client.py
novakzaballa: Covered in the view

khvn26: Is there a reason to handle the status code in the view specifically?

novakzaballa: Being a class View it is just easier and clearer to me.

api/integrations/github/views.py
novakzaballa: Being a class View it is no decorator is needed here.

khvn26: You should be able to decorate class methods without any problems.

novakzaballa: Yes, but I don't see that we need a decorator, here. Maybe I misunderstood your comment, Which decorator do you think we are missing here?

khvn26: The one you wrote to wrap GitHub errors, `github_api_call_error_handler`.

novakzaballa: RIght. It just because I would need to adapt it  or create a new one since the signature of the method is different from the signature of the functional views.

novakzaballa: So it was just easier this way

khvn26: You are using a generic `*args, **kwargs` signature when wrapping, so it shouldn't be a problem, unless I'm missing something here.

novakzaballa: Oh you're right, I was thinking all the time in this other decorator I wrote which has a specific signature https://github.com/Flagsmith/flagsmith/blob/feat/uninstall-github-app-installation-from-flagsmith/api/integrations/github/views.py#L33-L36
I'll see if can apply it to this class View.

novakzaballa:
@khvn26 just a heads up: While adding `djangorestframework-dataclasses` package I noticed that `requests ` package was also upgraded to `v2.32.0`. It seems like it was recently added by dependabot. I saw a warning message from Poetry about this package stating that it would be "Yanked due to conflicts CVE-2024-35195 mitigation".

api/integrations/github/views.py
```diff
@@ -86,9 +120,13 @@ def perform_create(self, serializer):
         serializer.save(github_configuration_id=github_configuration_id)
 
     def get_queryset(self):
-        return GithubRepository.objects.filter(
-            github_configuration=self.kwargs["github_pk"]
-        )
+        github_pk = self.kwargs.get("github_pk")
+        if github_pk is not None:
```
khvn26: nit: This method will cause an internal server error if `github_pk` is `None`. This renders the elaborate validation below incomplete.

api/features/feature_external_resources/views.py
```diff
@@ -25,6 +26,25 @@ def get_queryset(self):
...
+        # get organisation id from feature and get feature from validated data
+        organisation_id = get_object_or_404(
+            Feature.objects.filter(id=self.kwargs["feature_pk"]),
+        ).project.organisation_id
+
+        for resource in data if isinstance(data, list) else []:
+            if resource["url"]:
+                resource["metadata"] = get_github_issue_pr_title_and_state(
+                    organisation_id=organisation_id, resource_url=resource["url"]
+                )
```
khvn26:
```suggestion
            if resource_url := resource.get("url"):
                resource["metadata"] = get_github_issue_pr_title_and_state(
                    organisation_id=organisation_id, resource_url=resource_url
                )
```

api/features/models.py
```diff
@@ -406,6 +409,35 @@ def get_delete_log_message(self, history_instance) -> typing.Optional[str]:
...
+            feature_data: GithubData = generate_data(
+                github_configuration=github_configuration,
+                feature=self.feature,
+                type=WebhookEventType.SEGMENT_OVERRIDE_DELETED.value,
+                segment_name=self.segment.name,
+            )
+
+            call_github_app_webhook_for_feature_state.delay(
+                args=(asdict(feature_data),),
+            )
```
khvn26: I'd rather this code belonged in `integrations.github` somewhere (the amount of imports alone suggests it).

khvn26: Looking at both this code and the one in `api/features/versioning/serializers.py`, I'm pretty sure it can be factored out to a single interface.

api/features/versioning/serializers.py
```diff
@@ -14,6 +21,34 @@ class Meta(CreateSegmentOverrideFeatureStateSerializer.Meta):
...
+                github_configuration=github_configuration,
+                feature=feature_state.feature,
+                type=WebhookEventType.FLAG_UPDATED.value,
+                feature_states=feature_states,
+            )
+
+            call_github_app_webhook_for_feature_state.delay(
+                args=(asdict(feature_data),),
+            )
+
```
khvn26: Same as [this comment](https://github.com/Flagsmith/flagsmith/pull/3962/files#r1609667423). Factor out this code to a function in `integrations.github` namespace.

api/integrations/github/client.py
```diff
@@ -12,3 +54,154 @@ def generate_token(installation_id: str, app_id: int) -> str:
     Github(auth=auth)
     token = auth.token
     return token
+
+
+# TODO: Add coverage tests for this function
```
khvn26:
```suggestion
# TODO: Add test coverage for this function
```

api/integrations/github/client.py
```diff
@@ -12,3 +54,154 @@ def generate_token(installation_id: str, app_id: int) -> str:
...
+
+    if response.links.get("next"):
+        data["next"] = response.links.get("next")
+
+    # Return a Response object
+    return Response(
+        data=data,
+        content_type="application/json",
+        status=status.HTTP_200_OK,
+    )
```
khvn26: I appreciate we're trying to do the smart thing here, but you still validate this data in a view before returning it, so I don't see a valid reason to use the `Response` object here.

I think it's more fitting to use a `dict[str, Any]` (or, better yet, a `TypedDict`) as a return type for this function.

You may regard this comment as a nitpick, but, IMO, separation of concerns is important.

api/integrations/github/client.py
```diff
@@ -12,3 +54,154 @@ def generate_token(installation_id: str, app_id: int) -> str:
...
+    ]
+    data = {
+        "repositories": results,
+        "total_count": json_response["total_count"],
+    }
+    return Response(
+        data=data,
+        content_type="application/json",
+        status=status.HTTP_200_OK,
+    )
```
khvn26, line 176: Same as [this comment](https://github.com/Flagsmith/flagsmith/pull/3962/files#r1609687693).

api/features/versioning/serializers.py
```diff
@@ -14,6 +21,34 @@ class Meta(CreateSegmentOverrideFeatureStateSerializer.Meta):
...
+            not feature_state.identity_id
+            and feature_state.feature.external_resources.exists()
+            and feature_state.environment.project.github_project.exists()
+            and feature_state.environment.project.organisation.github_config.exists()
+        ):
+            github_configuration = GithubConfiguration.objects.get(
+                organisation_id=feature_state.environment.project.organisation_id
+            )
+            feature_states = []
+            feature_states.append(feature_state)
```
khvn26:
```suggestion
            feature_states = [feature_state]
```

api/integrations/github/client.py
```diff
@@ -1,8 +1,50 @@
...
+    installation_id: str, use_jwt: bool = False
+) -> dict[str, str]:
+    token = (
+        generate_token(
+            installation_id,
+            settings.GITHUB_APP_ID,
+        )
+        if not use_jwt
+        else generate_jwt_token(settings.GITHUB_APP_ID)
+    )
```
khvn26:
```suggestion
    token = (
        generate_jwt_token(settings.GITHUB_APP_ID)
        if use_jwt
        else generate_token(
            installation_id,
            settings.GITHUB_APP_ID,
        )
    )
```

api/integrations/github/client.py
```diff
@@ -1,8 +1,50 @@
...
+
+    return {
+        "X-GitHub-Api-Version": GITHUB_API_VERSION,
+        "Accept": "application/vnd.github.v3+json",
+        "Authorization": f"Bearer {token}",
+    }
+
+
+# TODO: Add coverage tests for this function
```
khvn26:
```suggestion
# TODO: Add test coverage for this function 
```

api/integrations/github/client.py
```diff
@@ -12,3 +54,154 @@ def generate_token(installation_id: str, app_id: int) -> str:
...
+    token = generate_token(
+        installation_id,
+        settings.GITHUB_APP_ID,
+    )
+
+    url = f"{GITHUB_API_URL}repos/{owner}/{repo}/issues/{issue}/comments"
+    headers = {
+        "Accept": "application/vnd.github.v3+json",
+        "Authorization": f"Bearer {token}",
+    }
```
khvn26: Can we use `build_request_headers` here?

api/integrations/github/client.py
```diff
@@ -12,3 +54,154 @@ def generate_token(installation_id: str, app_id: int) -> str:
...
+    number = url_parts[-1]
+    installation_id = GithubConfiguration.objects.get(
+        organisation_id=organisation_id, deleted_at__isnull=True
+    ).installation_id
+
+    url = f"{GITHUB_API_URL}repos/{owner}/{repo}/issues/{number}"
+    headers = build_request_headers(installation_id)
+    response = requests.get(url, headers=headers, timeout=GITHUB_API_CALLS_TIMEOUT)
+    response.raise_for_status()
+    return {"title": response.json()["title"], "state": response.json()["state"]}
```
khvn26:
```suggestion
    response_json = response.json()
    return {"title": response_json["title"], "state": response_json["state"]}
```

api/integrations/github/dataclasses.py
```diff
@@ -0,0 +1,44 @@
...
+from typing import Optional
+
+
+@dataclass
+class GithubData:
+    installation_id: str
+    feature_id: int
+    feature_name: str
+    type: str
+    feature_states: typing.List[dict[str, typing.Any]] | None = None
```
khvn26:
```suggestion
    feature_states: list[dict[str, typing.Any]] | None = None

```

api/integrations/github/github.py
```diff
@@ -73,17 +30,23 @@ def generate_body_comment(
     project_id: int,
     feature_id: int,
     feature_states: typing.List[typing.Dict[str, typing.Any]],
```
khvn26:
```suggestion
    feature_states: list[dict[str, typing.Any]],
```

api/integrations/github/client.py
```diff
@@ -12,3 +54,154 @@ def generate_token(installation_id: str, app_id: int) -> str:
...
+        app_id=app_id,
+        private_key=settings.GITHUB_PEM,
+    )
+    token = github_auth.create_jwt()
+    return token
+
+
+def post_comment_to_github(
+    installation_id: str, owner: str, repo: str, issue: str, body: str
+) -> typing.Dict[str, typing.Any]:
```
khvn26:
```suggestion
) -> dict[str, typing.Any]:
```

api/integrations/github/github.py
```diff
@@ -1,70 +1,27 @@
...
-            settings.GITHUB_APP_ID,
-        )
-
-        url = f"{GITHUB_API_URL}repos/{owner}/{repo}/issues/{issue}/comments"
-        headers = {
-            "Accept": "application/vnd.github.v3+json",
-            "Authorization": f"Bearer {token}",
-        }
+from .dataclasses import GithubData
+from .models import GithubConfiguration
```
khvn26: Let's avoid relative imports:

```suggestion
from integrations.github.dataclasses import GithubData
from integrations.github.models import GithubConfiguration
```

api/integrations/github/dataclasses.py
```diff
@@ -0,0 +1,44 @@
...
+    page: Optional[int] = 1
+    page_size: Optional[int] = 100
+    state: Optional[str] = "open"
+    author: Optional[str] = None
+    assignee: Optional[str] = None
+    search_in_body: Optional[bool] = True
+    search_in_comments: Optional[bool] = False
+
+    @classmethod
+    def from_dict(cls, data_dict: dict) -> "RepoQueryParams":
```
khvn26:
```suggestion
    def from_dict(cls, data_dict: dict[str, typing.Any]) -> "RepoQueryParams":

```

api/integrations/github/dataclasses.py
```diff
@@ -0,0 +1,44 @@
...
+    feature_id: int
+    feature_name: str
+    type: str
+    feature_states: typing.List[dict[str, typing.Any]] | None = None
+    url: str | None = None
+    project_id: int | None = None
+    segment_name: str | None = None
+
+    @classmethod
+    def from_dict(cls, data_dict: dict) -> "GithubData":
```
khvn26:
```suggestion
    def from_dict(cls, data_dict: dict[str, typing.Any]) -> "GithubData":

```

api/integrations/github/serializers.py
```diff
@@ -14,6 +17,7 @@ class Meta:
 class GithubRepositorySerializer(ModelSerializer):
     class Meta:
         model = GithubRepository
+        optional_fields = "search_text" "page"
```
khvn26: Looks like a typo? Should this be 
```suggestion
        optional_fields = ("search_text", "page")
```

api/integrations/github/tasks.py
```diff
@@ -85,10 +76,15 @@ def generate_feature_external_resources(
...
+        or github_event_data.type == WebhookEventType.SEGMENT_OVERRIDE_DELETED.value
+    ):
         feature_external_resources = generate_feature_external_resources(
-            FeatureExternalResource.objects.filter(
-                feature_id=github_event_data.feature_id
+            list(
+                FeatureExternalResource.objects.filter(
+                    feature_id=github_event_data.feature_id
+                ).all()
             )
```
khvn26: I think a cast to `list` is enough:
```suggestion
            list(
                FeatureExternalResource.objects.filter(
                    feature_id=github_event_data.feature_id
                )
            )
```

api/integrations/github/views.py
```diff
@@ -46,6 +50,31 @@ def wrapper(request, organisation_pk):
...
+def github_api_call_error_handler(
+    error: str | None = None,
+) -> Callable[..., Callable[..., Any]]:
+    def decorator(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs) -> Response:
+            try:
+                return func(*args, **kwargs)
+            except requests.RequestException as e:
+                return Response(
```
khvn26, line 67: I'd throw in a `logger.error`/`logger.warning` here to log the GH error for our own convenience.

khvn26 (changes requested):
Left a bunch of comments.

api/integrations/github/views.py
novakzaballa: Done

api/integrations/github/tasks.py
novakzaballa: Done

api/integrations/github/serializers.py
novakzaballa: Done

api/integrations/github/dataclasses.py
novakzaballa: Done

api/integrations/github/dataclasses.py
novakzaballa: Done

api/integrations/github/github.py

api/integrations/github/client.py
novakzaballa: Done

api/integrations/github/github.py

api/integrations/github/dataclasses.py
novakzaballa: Done

api/integrations/github/client.py
novakzaballa: Done

api/integrations/github/client.py
novakzaballa: Done

api/integrations/github/client.py
novakzaballa: Done

api/integrations/github/client.py
novakzaballa: Done

api/features/versioning/serializers.py
novakzaballa: Done

api/integrations/github/client.py
novakzaballa: ok, Done

api/integrations/github/client.py
novakzaballa: Ok, done

api/integrations/github/client.py
novakzaballa: Done

api/features/feature_external_resources/views.py
novakzaballa: Done

api/integrations/github/views.py
novakzaballa: Done

api/features/versioning/serializers.py
novakzaballa: I created a common function to handle this code.

api/features/models.py
novakzaballa: I created a common function to handle this code.

api/integrations/github/views.py
```diff
@@ -46,6 +53,28 @@ def wrapper(request, organisation_pk):
...
+    error: str | None = None,
+) -> Callable[..., Callable[..., Any]]:
+    def decorator(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs) -> Response:
+            default_error = "Failed to retrieve requested information from GitHub API."
+            try:
+                return func(*args, **kwargs)
+            except requests.RequestException as e:
+                logger.error(f"{error or default_error} Error: {str(e)}")
```
khvn26:
```suggestion
                logger.error(f"{error or default_error} Error: {str(e)}", exc_info=e)
```

khvn26 (approved):

PR #3974, opened 2024-05-17
fix(versioning): multiple versioned segment overrides added to environment document

matthewelwell:

## Changes

This PR fixes an issue found in testing where we were previously adding all versions of a given segment override to the environment document. It achieves this by (unfortunately) adding an extra query to the mapping function to grab the latest versions for an environment and then filtering out any segment overrides that don't have a matching version. 

Added an explicit unit test on the mapper function to test it.

### DIFF
--- a/api/util/mappers/engine.py
+++ b/api/util/mappers/engine.py
@@ -1,6 +1,7 @@
 from collections.abc import Iterable
 from itertools import chain
 from typing import TYPE_CHECKING, Dict, List, Optional
+from uuid import UUID
 
 from flag_engine.environments.integrations.models import IntegrationModel
 from flag_engine.environments.models import (
@@ -430,7 +431,7 @@ def _get_prioritised_feature_states(
 def _get_segment_feature_states(
     segments: Iterable["Segment"],
     environment_id: int,
-    latest_environment_feature_version_uuids: Iterable["EnvironmentFeatureVersion"],
+    latest_environment_feature_version_uuids: Iterable[UUID],
 ) -> Dict[int, List["FeatureState"]]:
     feature_states_by_segment_id = {}
 
--- a/api/util/mappers/engine.py
+++ b/api/util/mappers/engine.py
@@ -25,6 +25,7 @@ from flag_engine.segments.models import (
 )
 
 from environments.constants import IDENTITY_INTEGRATIONS_RELATION_NAMES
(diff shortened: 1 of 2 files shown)


api/util/mappers/engine.py
```diff
@@ -419,14 +425,29 @@ def _get_prioritised_feature_states(
 def _get_segment_feature_states(
     segments: Iterable["Segment"],
     environment_id: int,
+    latest_environment_feature_versions: (
+        Iterable["EnvironmentFeatureVersion"] | None
+    ) = None,
```
khvn26: nit: optional argument is not needed here

api/util/mappers/engine.py
```diff
@@ -419,14 +425,29 @@ def _get_prioritised_feature_states(
...
 ) -> Dict[int, List["FeatureState"]]:
     feature_states_by_segment_id = {}
+    latest_environment_feature_versions = latest_environment_feature_versions or []
```
khvn26: remove along with https://github.com/Flagsmith/flagsmith/pull/3974/files#r1604807951

khvn26: Convert to set for faster membership checks

khvn26 (changes requested):
LGTM overall, requested a minor cleanup

api/util/mappers/engine.py
matthewelwell: Done.

api/util/mappers/engine.py
matthewelwell: Since we had to do an iteration to convert to a set, I ended up using a set of uuids instead of the object itself. This should give us an extra (very minor) boost in performance too I think.

api/util/mappers/engine.py
```diff
@@ -419,14 +430,26 @@ def _get_prioritised_feature_states(
 def _get_segment_feature_states(
     segments: Iterable["Segment"],
     environment_id: int,
+    latest_environment_feature_version_uuids: Iterable["EnvironmentFeatureVersion"],
```
khvn26: The typing is incorrect here

matthewelwell: Gah, good spot! Fixing...

matthewelwell: Fixed.

khvn26 (approved):

PR #3976, opened 2024-05-17
chore: remove deprecated functions url and ugettext_lazy

gagantrivedi:

## Changes

- Replace `django.conf.urls url` with `django.urls.re_path`
- Replace `ugettext_lazy` with `gettext_lazy`

Note: ugettext_lazy and url were deprecated in Django 3.x.

covered by existing unit tests( :crossed_fingers: )

### DIFF
--- a/api/app/urls.py
+++ b/api/app/urls.py
@@ -1,9 +1,8 @@
 import importlib
 
 from django.conf import settings
-from django.conf.urls import include, url
 from django.contrib import admin
-from django.urls import path
+from django.urls import include, path, re_path
 from django.views.generic.base import TemplateView
 
 from users.views import password_reset_redirect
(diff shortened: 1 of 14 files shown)


api/app/urls.py
```diff
@@ -70,4 +69,4 @@
 
 if settings.SERVE_FE_ASSETS:
```
matthewelwell:
```suggestion
if settings.SERVE_FE_ASSETS:  # pragma: no cover
```

api/environments/models.py
```diff
@@ -10,7 +10,7 @@
 from django.db import models
 from django.db.models import Prefetch, Q
 from django.utils import timezone
-from django.utils.translation import ugettext_lazy as _
+from django.utils.translation import gettext_lazy as _
```
matthewelwell: I had a look at where we're actually using this and I think we can just remove it, right?

api/features/models.py
```diff
@@ -21,7 +21,7 @@
 from django.db import models
 from django.db.models import Max, Q, QuerySet
 from django.utils import timezone
-from django.utils.translation import ugettext_lazy as _
+from django.utils.translation import gettext_lazy as _
```
matthewelwell: Probably the same comment as above.

api/users/urls.py
```diff
@@ -24,4 +23,4 @@
 ]
 
 if settings.ALLOW_ADMIN_INITIATION_VIA_URL:
```
matthewelwell:
```suggestion
if settings.ALLOW_ADMIN_INITIATION_VIA_URL:  # pragme: no cover
```

api/environments/models.py
gagantrivedi: Yeah, we don't really support any other language; I will remove them

gagantrivedi: Done

api/features/models.py
gagantrivedi: Done

matthewelwell (approved):

PR #3988, opened 2024-05-21
fix(deps): Migrate MFA code to our codebase and bump djangorestframework

gagantrivedi:

## Changes

- Because Trench is no longer maintained, this commit adds all the code that we depend on.
- Bump djangorestframework
- Fixes update feature external resource endpoint

The following scenarios were tested manually:

  -  Generate MFA code from the old library and log in using the code from this PR and vice versa.
   - Log in using the backup codes generated by the old library with the code from this PR and vice versa.


api/custom_auth/mfa/trench/__init__.py
```diff
@@ -0,0 +1 @@
+__version__ = "0.3.1"
```
matthewelwell: Do we need this? Feels like another version number that will inevitably become stale.

api/custom_auth/urls.py
```diff
@@ -25,12 +26,13 @@
...
+    path("logout/", TokenDestroyView.as_view(), name="authtoken-logout"),
     path("", include(ffadmin_user_router.urls)),
     path("token/", delete_token, name="delete-token"),
     # NOTE: endpoints provided by `djoser.urls`
     # are deprecated and will be removed in the next Major release
     path("", include("djoser.urls")),
-    path("", include("trench.urls")),  # MFA
-    path("", include("trench.urls.djoser")),  # override necessary urls for MFA auth
+    path("", include("custom_auth.mfa.trench.urls")),  # MFA
+    # path("", include("trench.urls")),  # MFA
```
matthewelwell: Is there a reason that this commented out? If not, let's remove it. If there is a reason for this to still be here, let's add a comment explaining why it's commented out.

matthewelwell (approved):
Approving with 2 minor comments.

api/custom_auth/mfa/trench/__init__.py
gagantrivedi: Yeah, we don't need this

api/custom_auth/urls.py
gagantrivedi: Good catch! I forgot to remove this after our discussion around removing endpoints that are not used by UI

matthewelwell (approved):

matthewelwell:
@gagantrivedi this looks good to me now, I think we can resolve the merge conflicts and merge it in.

PR #3991, opened 2024-05-21
feat(versioning): add logic to create version in single endpoint

matthewelwell:

## Changes

This PR updates the endpoint to create a new version to add new attributes that allow you to create, update and delete feature states when creating the new version and optionally publish it immediately. This will avoid the FE having to create the version, retrieve the feature states, update any necessary feature states and then publish it. 

I'm keen to get feedback on the naming of the new fields as I'm not 100% pleased with them. The endpoint looks like this: 

image

Added tests to cover all the use cases.


api/tests/unit/features/versioning/test_unit_versioning_views.py
```diff
@@ -638,3 +638,75 @@ def test_disable_v2_versioning_returns_bad_request_if_not_using_v2_versioning(
...
+            {
+                "feature_segment": None,
+                "enabled": True,
+                "feature_state_value": {"type": "unicode", "string_value": "updated!"},
+            }
+        ],
+        "feature_states_to_create": [
+            {
+                "feature_segment": {
+                    # TODO: priorities?
```
matthewelwell: This is going to be tricky! But one that we might need to try and tackle.

api/features/feature_segments/serializers.py
```diff
@@ -37,10 +37,24 @@ def validate(self, data):
...
+    # (to adhere to the django-ordered-model functionality), we redefine the priority
+    # field here, and use it manually in the save method.
+    priority = serializers.IntegerField(min_value=1, required=False)
+
     class Meta:
         model = FeatureSegment
         fields = ("id", "segment", "priority", "uuid")
-        read_only_fields = ("priority",)
+
+    def save(self, **kwargs):
```
zachaysan: nit: typing

api/features/versioning/serializers.py
```diff
@@ -64,6 +64,163 @@ class Meta:
...
+            "publish_immediately",
+        )
+        non_model_fields = (
+            "feature_states_to_create",
+            "feature_states_to_update",
+            "segment_ids_to_delete_overrides",
+            "publish_immediately",
+        )
+
+    def create(self, validated_data):
```
zachaysan: nit: typing

api/features/versioning/serializers.py
```diff
@@ -64,6 +64,163 @@ class Meta:
...
+        for field_name in self.Meta.non_model_fields:
+            validated_data.pop(field_name, None)
```
zachaysan, line 148: I'm a little confused how these popped fields are actually used, but I'll keep reading the PR for more info.

zachaysan: Ok I understand it now. Maybe the comment below that explains it could go into more detail why we want raw data.

api/features/versioning/serializers.py
```diff
@@ -64,6 +64,163 @@ class Meta:
...
+                "environment": version.environment,
+                "environment_feature_version": version,
+            },
+        )
+        fs_serializer.is_valid(raise_exception=True)
+        fs_serializer.save(
+            environment_feature_version=version,
+            environment=version.environment,
+            feature=version.feature,
+        )
```
zachaysan: Is there a DRYer way of doing this? The context was already set in the constructor, it seems weird to have to set them again.

api/features/versioning/serializers.py
```diff
@@ -64,6 +64,163 @@ class Meta:
...
+
+    def _update_feature_state(
+        self, feature_state: dict, version: EnvironmentFeatureVersion
+    ) -> None:
```
zachaysan: It would be better if the typing on the `dict` had more information.

api/tests/unit/features/versioning/test_unit_versioning_views.py
```diff
@@ -638,3 +638,255 @@ def test_disable_v2_versioning_returns_bad_request_if_not_using_v2_versioning(
...
+
+    assert new_version.published is True
+    assert new_version.is_live is True
+
+
+def test_update_and_create_segment_override_in_single_request(
+    feature: Feature,
+    segment: Segment,
+    segment_featurestate: FeatureState,
+    admin_client: APIClient,
```
zachaysan: Is this intentional use of `admin_client` instead of `admin_client_new`? Also, many of the tests use the `admin_client_new` which is great, but are we missing out on testing the ACL with `staff_client` too?

zachaysan (approved):
A mess of small suggestions or nits but overall the code looks great. I understand you want better naming but I couldn't think of better ideas than the ones you've already put in place.

api/features/feature_segments/serializers.py
matthewelwell: Added.

api/features/versioning/serializers.py
matthewelwell: Added.

api/features/versioning/serializers.py
matthewelwell: Hmm... I certainly see your point, but I'm struggling to go into too much more detail without writing an essay. The main reason is just that we're utilising the serializers to create the feature states and the serializers expect raw data, not ORM models. I'm obviously keen to make the comment as useful as possible, so let me know if you have any suggestions on how to improve.

api/features/versioning/serializers.py
matthewelwell: I'm not really too sure why we need both the context and the save kwargs. I tried to follow it down the rabbit hole (of my own making), but couldn't see anything obvious. I've tidied it up a little bit to use a single dict definition in both cases, and I'll try to tackle a better refactor in a separate piece of work.

api/features/versioning/serializers.py
matthewelwell: I've added as much information as I think I can without it being beneficial to just create e.g. a pydantic model.

api/tests/unit/features/versioning/test_unit_versioning_views.py
matthewelwell: Good point. I've updated an older test to use staff client and standardised to using `admin_client_new` on these new tests.

matthewelwell:
I've now tested this PR against the FE branch [here](https://github.com/Flagsmith/flagsmith/pull/4050) and confirmed that everything seems to work well. 

I used the Uffizzi deployment to do this too which was a good test latency-wise because the Uffizzi deployment runs somewhere in the US. It was noticeable that the latency was greatly improved when changing the state of the flag compared even to making a change in production SaaS from London.

api/features/versioning/serializers.py
zachaysan: The comment below seems to cover it mostly. Maybe add a one liner that says `# self.initial_data is used instead of popping validated data.`

zachaysan (approved):
Changes look good!

api/features/versioning/serializers.py
matthewelwell: I've moved (and slightly improved the comment) in the latest commit.

PR #3996, opened 2024-05-21
fix: Set api usage billing to 100k

zachaysan:

## Changes

1. Set up billing to 100k from 1000 from the Chargebee add-on code.
2. Fix a horrible bug that would have allowed customers to be billed multiple times for the same API calls.

Updated tests and added for codecov.

### DIFF
--- a/api/organisations/models.py
+++ b/api/organisations/models.py
@@ -500,8 +500,10 @@ class OrganisationAPIBilling(models.Model):
 
     Even though api_overage is charge per thousand API calls, this
     class tracks the actual rounded count of API calls that are
-    billed for (i.e., 52000 for an account with 52233 api calls).
-    We're intentionally rounding down to the closest thousands.
+    billed for (i.e., 200000 for an account with 234323 api calls
+    and a allowed_30d_api_calls set to 100000, the overage is
+    beyond the allowed api calls).
+    We're intentionally rounding up to the closest hundred thousand.
 
     The option to set immediate_invoice means whether or not the
     API billing was processed immediately versus pushed onto the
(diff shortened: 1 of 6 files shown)


api/organisations/models.py
```diff
@@ -500,8 +500,10 @@ class OrganisationAPIBilling(models.Model):
 
     Even though api_overage is charge per thousand API calls, this
```
matthewelwell:
```suggestion
    Even though api_overage is charge per 100k API calls, this
```

api/organisations/tasks.py
```diff
@@ -255,21 +252,35 @@ def charge_for_api_call_count_overages():
...
 
-        if api_usage_ratio < 1.0:
-            logger.warning("API Usage does not match API Notification")
-            continue
+        api_billing = (
+            OrganisationAPIBilling.objects.filter(
+                billed_at__gte=subscription_cache.current_billing_term_starts_at
+            )
+            .order_by("-billed_at")
+            .first()
```
matthewelwell: There should only ever be one of these records, right? Do we care about the `order_by`? Should we assert that there is only one?

matthewelwell (commented):
Added a couple of minor comments.

api/organisations/tasks.py
zachaysan: No these `OrganisationAPIBilling` objects are created each time someone has been billed. So if in a given month a customer has multiple overages each time a new billing. That said, I made a mistake with the logic around the billed amount since the overages are only the latest billings when it should be a sum of the billings within the month. I'll update this pr with a fix when I can.

zachaysan: Ok I've written a fix and verified it works.

matthewelwell (approved):
I'm approving here, but I'm not sure what's going on with the codecov workflow that seems stuck..

PR #3999, opened 2024-05-21
fix(versioning): fix cloning environments using v2 versioning

matthewelwell:

## Changes

This fixes an issue found in testing when cloning an environment that has enabled v2 versioning. 

Added a unit test for cloning an environment using v2 versioning.

### DIFF
--- a/api/environments/models.py
+++ b/api/environments/models.py
@@ -188,7 +188,7 @@ class Environment(
             # Create a dictionary holding the environment feature versions (unique per feature)
             # to use in the cloned environment.
             clone_environment_feature_versions = {
-                efv.feature_id: efv.clone(environment=clone)
+                efv.feature_id: efv.clone_to_environment(environment=clone)
                 for efv in latest_environment_feature_versions
             }
 
--- a/api/environments/models.py
+++ b/api/environments/models.py
@@ -41,6 +41,7 @@ from environments.exceptions import EnvironmentHeaderNotPresentError
 from environments.managers import EnvironmentManager
 from features.models import Feature, FeatureSegment, FeatureState
 from features.multivariate.models import MultivariateFeatureStateValue
+from features.versioning.models import EnvironmentFeatureVersion
 from metadata.models import Metadata
 from projects.models import Project
 from segments.models import Segment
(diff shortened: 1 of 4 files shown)


api/environments/models.py
```diff
@@ -174,8 +175,36 @@ def clone(self, name: str, api_key: str = None) -> "Environment":
...
+            # to clone these later on in the logic, we need the ORM objects, not the
+            # RawSQL queryset, so we have to nest the queries here.
+            latest_environment_feature_versions = EnvironmentFeatureVersion.objects.filter(
+                uuid__in=[
+                    efv.uuid
+                    for efv in EnvironmentFeatureVersion.objects.get_latest_versions(
+                        self
+                    )
+                ]
+            )
```
matthewelwell: I'm not particularly happy with this, but the alternative was to update the sql to return all the fields and then manually create the EnvironmentFeatureVersion, rather than using a `clone` method on the model class. To keep things consistent with other models, this seemed like the lesser of 2 evils. 

One option might be to conceal this in another method on the EnvironmentFeatureVersion model manager.

api/features/versioning/managers.py
```diff
@@ -15,9 +16,25 @@
...
+
+        Note that it is often required to return the proper QuerySet to carry out
+        operations on the ORM object.
+        """
+        return self.filter(
+            uuid__in=[
+                efv.uuid
+                for efv in EnvironmentFeatureVersion.objects.get_latest_versions(self)
+            ]
+        )
```
matthewelwell: I'm not particularly happy with this, but the alternative was to update the sql to return all the fields and then (in the `Environment.clone` method) manually create the EnvironmentFeatureVersion, rather than using a clone method on the model class. To keep things consistent with other models, this seemed like the lesser of 2 evils.

api/features/versioning/models.py
```diff
@@ -128,3 +129,12 @@ def publish(
         if persist:
             self.save()
             environment_feature_version_published.send(self.__class__, instance=self)
+
+    def clone(self, environment: "Environment") -> "EnvironmentFeatureVersion":
```
khvn26: I'm aware that the naming's already pretty deeply integrated but I really think `clone_to_environment` or `migrate_to_environment` would be better.

matthewelwell: Sure, fair point. On it.

api/features/versioning/managers.py
```diff
@@ -1,11 +1,12 @@
 import typing
 from pathlib import Path
 
-from django.db.models.query import RawQuerySet
+from django.db.models.query import QuerySet, RawQuerySet
 from softdelete.models import SoftDeleteManager
 
 if typing.TYPE_CHECKING:
     from environments.models import Environment
+    from features.versioning.models import EnvironmentFeatureVersion
```
khvn26, line 9: This should be a normal import.

matthewelwell: No, it can't be. The `models` module imports directly from the managers (and has to).

api/features/versioning/models.py
matthewelwell: Done.

api/features/versioning/managers.py
khvn26: Why not write a second sql with all the fields?

> manually create the EnvironmentFeatureVersion, rather than using a clone method on the model class

I don't think so as long as the model instances were correctly mapped?

api/features/versioning/managers.py
khvn26: You need an import inside the new interfaces, then.

khvn26: (Have I told you I hate Django btw?)

api/features/versioning/managers.py
novakzaballa: I think that "Cloning environments" is not used frequently enough to make this a real issue.

novakzaballa (approved):
LGTM

api/features/versioning/managers.py
matthewelwell: Sorry, my bad, that was just a copypasta error. I've replaced it with `self.get_latest_versions`.

api/features/versioning/managers.py
khvn26: Fair point Novak

khvn26 (approved):

api/features/versioning/managers.py
matthewelwell: >I think that "Cloning environments" is not used frequently enough to make this a real issue.

Well, I wish this were true. But the FE made the decision to remove the functionality to 'create' an environment, and now all environments are cloned from another. 

> Why not write a second sql with all the fields?

I'd still rather avoid this to avoid code duplication or hacky ways of concatenating SQL strings.

PR #4005, opened 2024-05-22
feat: async the logic for cloning feature states into a cloned environment

matthewelwell:

## Changes

Adds a new field to the clone environment endpoint which allows the creation of feature states to be offloaded to the task processor.  This stemmed from work I did to handle the cloning of an environment that is using v2 versioning, however, even the cloning of an environment on the old versioning system is pretty intensive. 

This PR also adds a new field `is_creating` to the environment model which can be used by the frontend to determine when an environment has been successfully cloned. 

Related issue: #4489

Added a new unit test and updated existing ones where needed, but I have not changed the default behaviour so existing test coverage should be sufficient in most cases.


matthewelwell:
Note that this PR is only marked as draft as we need to discuss with the FE team how to implement the 'pending creation' logic. I don't want to review / merge this PR before then as it includes a migration that might need to change.

khvn26 (approved):

PR #4013, opened 2024-05-22
feat: Add endpoint to fetch GitHub repository contributors

novakzaballa:

## Changes
Fetch GitHub repository contributors endpoint added.
Some Data Classes were optimized

Unit tests added to cover the new code

### DIFF
--- a/api/integrations/github/client.py
+++ b/api/integrations/github/client.py
@@ -191,7 +191,7 @@ def fetch_github_repositories(
             "id": i["id"],
             "name": i["name"],
         }
-        for i in json_response["results"]
+        for i in json_response["repositories"]
     ]
 
     return build_paginated_response(results, response, json_response["total_count"])
--- a/api/integrations/github/client.py
+++ b/api/integrations/github/client.py
@@ -5,15 +5,17 @@ from typing import Any
 import requests
 from django.conf import settings
 from github import Auth, Github
-from rest_framework import status
-from rest_framework.response import Response
 
 from integrations.github.constants import (
     GITHUB_API_CALLS_TIMEOUT,
     GITHUB_API_URL,
     GITHUB_API_VERSION,
 )
-from integrations.github.dataclasses import IssueQueryParams
+from integrations.github.dataclasses import (
+    IssueQueryParams,
+    PaginatedQueryParams,
+    RepoQueryParams,
+)
 from integrations.github.models import GithubConfiguration
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


api/integrations/github/dataclasses.py
```diff
@@ -1,44 +1,54 @@
...
-    feature_external_resources: list[dict[str, typing.Any]]
+    feature_external_resources: list[dict[str, Any]]
 
 
+# Dataclasses for external calls to GitHub API
 @dataclass
-class RepoQueryParams:
+class PaginatedQueryParams(ABC):
+    page: int = field(default=1, init=False)
+    page_size: int = field(default=100, init=False)
```
khvn26: 1. Why derive from `ABC`?
2. Why `init=False`? What if we want to customise pagination?

khvn26 (commented):
LGTM in general with one question outstanding.

api/integrations/github/dataclasses.py
novakzaballa: 1. Because I forgot we need to use it for a request and thought we would only need it as a base class for others. Now I changed it to a regular class and it's in use by the get repos endpoint.
2. Because derivating without either `init=false` or `kw_only` gives an error. I now changed to use `kw_only` to have a better result.
Thanks for cathing this out!

novakzaballa:
@khvn26 I solved your question, but that implied making some adjustments and improvements. Can you please re-review?

api/tests/unit/integrations/github/test_unit_github_views.py
```diff
@@ -776,3 +780,209 @@ def test_cannot_fetch_repositories_when_there_is_no_installation_id(
...
+    url = reverse(
+        "api-v1:organisations:get-github-installation-repos", args=[organisation.id]
+    )
+    # When
+    response = admin_client.get(url, data)
+
+    # Then
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
+    response_json = response.json()
+    if is_invalid_page:
```
khvn26: An `if` inside a test is a code smell. I think it's better to have a parametrised `detail` or `response_json` argument instead.

api/tests/unit/integrations/github/test_unit_github_views.py
```diff
@@ -776,3 +780,209 @@ def test_cannot_fetch_repositories_when_there_is_no_installation_id(
...
```
khvn26: Same here; consider parametrising `detail` or `response_json` argument instead.

api/integrations/github/client.py
```diff
@@ -66,9 +68,34 @@ def generate_jwt_token(app_id: int) -> str:  # pragma: no cover
     return token
 
 
+def build_paginated_response(
+    results: list[dict[str, Any]],
+    response: requests.Response,
+    total_count: int | None = None,
+    incomplete_results: bool | None = None,
+) -> dict[str, Any]:
```
khvn26, line 76: nit: It's not critical, but, since we fully control the creation of this dict, we could add a `TypedDict` return type to this function that would help introspection.

khvn26 (approved):
Three minor comments, otherwise LGTM.

api/tests/unit/integrations/github/test_unit_github_views.py
novakzaballa: Right, thanks! Done.

api/tests/unit/integrations/github/test_unit_github_views.py
novakzaballa: Done

api/integrations/github/client.py
novakzaballa: Maybe Dataclass or would it be too much? Let's please discuss this. Merging as is FTM.

PR #4037, opened 2024-05-28
fix: GitHub repos unique constraint and delete

novakzaballa:

## Changes

- Fix soft-deleted issues with GitHub repos/project uniqueness.
- Fix when removing a GitHub repo from integration delete only the resources of that repo.
- Improve unit test.

- Run the GitHub view, and feature external resources unit tests


api/conftest.py
```diff
@@ -85,6 +87,25 @@ def pytest_sessionstart(session: pytest.Session) -> None:
     fix_issue_3869()
 
 
+@pytest.fixture()
+def post_request_mock(mocker: MockerFixture) -> MagicMock:
+    def mocked_request(*args, **kwargs):
+        class MockResponse:
+            def __init__(self, json_data, status_code):
```
zachaysan: Missing typing

api/conftest.py
```diff
@@ -85,6 +87,25 @@ def pytest_sessionstart(session: pytest.Session) -> None:
     fix_issue_3869()
 
 
+@pytest.fixture()
+def post_request_mock(mocker: MockerFixture) -> MagicMock:
+    def mocked_request(*args, **kwargs):
```
zachaysan: Missing typing

api/conftest.py
```diff
@@ -85,6 +87,25 @@ def pytest_sessionstart(session: pytest.Session) -> None:
...
+        class MockResponse:
+            def __init__(self, json_data, status_code):
+                self.json_data = json_data
+                self.status_code = status_code
+
+            def raise_for_status(self) -> None:
+                pass
+
+            def json(self):
```
zachaysan: Missing typing

api/integrations/github/github.py
```diff
@@ -25,6 +26,21 @@
 logger = logging.getLogger(__name__)
 
 
+def handle_installation_deleted(payload: dict[str, typing.Any]) -> None:
```
zachaysan: nit: `Any` has already been imported from `typing` so this should be like the below
```suggestion
def handle_installation_deleted(payload: dict[str, Any]) -> None:
```

api/integrations/github/github.py
```diff
@@ -25,6 +26,21 @@
...
+    installation_id = payload.get("installation", {}).get("id")
+    try:
+        GithubConfiguration.objects.get(installation_id=installation_id).delete()
+    except GithubConfiguration.DoesNotExist:
```
zachaysan: `installation_id` could be `None` which may match the queryset if there are `installation_id` matching records with `None` in them. I would add an explicit check for `installation_id is None` and handle that case separately.

api/integrations/github/github.py
```diff
@@ -25,6 +26,21 @@
...
+        logger.error(
+            f"Github Configuration with installation_id {installation_id} does not exist"
```
zachaysan: nit:
```suggestion
            f"GitHub Configuration with installation_id {installation_id} does not exist"
```

api/integrations/github/models.py
```diff
@@ -57,21 +58,26 @@ class Meta:
...
         FeatureExternalResource.objects.filter(
             feature_id__in=self.project.features.values_list("id", flat=True),
             type__in=[
-                FeatureExternalResource.ResourceType.GITHUB_ISSUE,
-                FeatureExternalResource.ResourceType.GITHUB_PR,
+                ResourceType.GITHUB_ISSUE,
+                ResourceType.GITHUB_PR,
             ],
+            # Filter by url containing the repository owner and name
+            url__contains=f"{self.repository_owner}/{self.repository_name}",
```
zachaysan: The match criteria isn't good enough. If a repository owner is named `foo` it would also match to `examplefoo`. you need to bound the region of the `url__contains` so that you're always matching the right object.

api/tests/unit/integrations/github/test_unit_github_views.py
```diff
@@ -639,6 +650,32 @@ def test_github_webhook_delete_installation(
     assert not GithubConfiguration.objects.filter(installation_id=1234567).exists()
 
 
+def test_github_webhook_with_non_existing_installation(
+    github_configuration: GithubConfiguration,
```
zachaysan, line 661: nit:
```suggestion
    github_configuration: GitHubConfiguration,
```

api/tests/unit/integrations/github/test_unit_github_views.py
```diff
@@ -639,6 +650,32 @@ def test_github_webhook_delete_installation(
...
+    mocker: MockerFixture,
+) -> None:
+    # Given
+    settings.GITHUB_WEBHOOK_SECRET = WEBHOOK_SECRET
+    url = reverse("api-v1:github-webhook")
+    mocker_logger = mocker.patch("integrations.github.github.logger")
+
+    # When
+    client = APIClient()
```
zachaysan: nit: use `api_client` fixture

api/tests/unit/integrations/github/test_unit_github_views.py
```diff
@@ -639,6 +650,32 @@ def test_github_webhook_delete_installation(
...
+        path=url,
+        data=WEBHOOK_PAYLOAD_WITHOUT_INSTALLATION_ID,
+        content_type="application/json",
+        HTTP_X_HUB_SIGNATURE=WEBHOOK_SIGNATURE_WITHOUT_INSTALLATION_ID,
+        HTTP_X_GITHUB_EVENT="installation",
+    )
+
+    # Then
+    mocker_logger.error.assert_called_once_with(
+        "Github Configuration with installation_id 765432 does not exist"
```
zachaysan: nit:
```suggestion
        "GitHub Configuration with installation_id 765432 does not exist"
```

zachaysan (approved):
Overall looks good, but left a number of nits. There is one change I think you should address which is the `url__contains` bit since it would bleed data.

api/integrations/github/github.py

api/tests/unit/integrations/github/test_unit_github_views.py
novakzaballa: Done

api/tests/unit/integrations/github/test_unit_github_views.py
novakzaballa: Done

api/integrations/github/github.py

api/integrations/github/github.py

api/conftest.py
novakzaballa: Done

api/conftest.py
novakzaballa: Done

api/conftest.py
novakzaballa: Done

api/tests/unit/integrations/github/test_unit_github_views.py
novakzaballa: I think this shouldn't be changed as it's the model name.

api/integrations/github/models.py
novakzaballa: I'm using regex instead of contains

PR #4054, opened 2024-05-29
fix: Set billing starts at to reasonable default for API usage notifications

zachaysan:

- [x ] I have filled in the "Changes" section below?

## Changes

Set billing starts at to reasonable default for API usage notifications. This shouldn't be common in practice since most subscriptions set a billing period. Solves [this Sentry issue](https://flagsmith.sentry.io/issues/5424660386/?environment=staging&project=5544478&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=24h&stream_index=2).

N/A

### DIFF
--- a/api/organisations/views.py
+++ b/api/organisations/views.py
@@ -2,6 +2,7 @@
 from __future__ import unicode_literals
 
 import logging
+from datetime import timedelta
 
 from app_analytics.influxdb_wrapper import (
     get_events_for_organisation,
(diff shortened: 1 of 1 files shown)


matthewelwell:
>This shouldn't be common in practice since most subscriptions set a billing period.

I don't think this is actually true since most subscriptions are on the 'free' plan.

matthewelwell (approved):
I'm going to approve and merge this for now because I think it's not relevant to organisations on the free plan, but we should discuss when you're around @zachaysan .

PR #4055, opened 2024-05-29
feat: Add UI for configuring SAML in Flagsmith

### LINKED ISSUE #4006: Add UI for configuring SAML via Flagsmith dashboard
Currently, configuration of SAML is restricted to the django admin. We should add the option to configure it in the organisation settings in the Flagsmith dashboard. 

This should only be possible for organisation admins, and is only available to enterprise customers. 

The flow should be: 

 1. FE sends a request to `/api/v1/auth/saml/configuration/` with the following information: 

```json
{
  "organisation": 1,  // id of the organisation to add SAML to
  "name": "org-short-name",  // a short name for the organisation, used as the input when clicking 'Single Sign-on' at login (note this is unique across all tenants and will form part of the URL so should only be alphanumeric + '-,_')
  "frontend_url": "https://app.flagsmith.com",  // this should be the base URL of the Flagsmith dashboard
  "allow_idp_initiated": true,  // this field determines whether logins can be initiated from the IdP
  "idp_metadata_xml": "...",  // the metadata from the IdP (this should not be required on creation as it might depend on our service provider metadata)
}
```

Once this is created, it can be updated / deleted using the relevant HTTP verbs and the URL `/api/v1/auth/saml/configuration/:name/`. 

2. The FE offers the option to download our service provider metadata XML which can be obtained using the endpoint at `/api/v1/auth/saml/{name}/metadata/`. 

3. Finally the FE should offer the option to upload the IdP metadata in response, as per the description above. 

There are additional entities that can be used for mapping fields between the IdP and Flagsmith, but these do not yet have endpoints. Once these are ready, we will update this issue or create a new one.

novakzaballa:

## Changes

- Add UI for configuring SAML in Flagsmith wrapped with the `saml_configuration` flag.

- Use an organisation with an enterprise plan
- Go to Organisation Settings -> SAML tab
- Create a SAML configuration

### DIFF
--- a/docs/docs/system-administration/authentication/01-SAML/index.md
+++ b/docs/docs/system-administration/authentication/01-SAML/index.md
@@ -15,17 +15,17 @@ SAML tab, you'll be able to configure it.
 
 In the UI, you will be able to configure the following fields.
 
-**Name:** (**Requierd**) A short name for the organisation, used as the input when clicking "Single Sign-on" at login
+**Name:** (**Required**) A short name for the organisation, used as the input when clicking "Single Sign-on" at login
 (note this is unique across all tenants and will form part of the URL so should only be alphanumeric + '-,\_').
 
-**Frontend URL**: (**Requierd**) This should be the base URL of the Flagsmith dashboard.
+**Frontend URL**: (**Required**) This should be the base URL of the Flagsmith dashboard.
 
 **Allow IdP initiated**: This field determines whether logins can be initiated from the IdP.
 
 **IdP metadata xml**: The metadata from the IdP.
 
-Once Flagsmith has configured your identity provider, you can download the service provider metadata XML document whit
-the button "Download Service Provider Metadata".
+Once Flagsmith you have configured your identity provider, you can download the service provider metadata XML document
+with the button "Download Service Provider Metadata".
 
 ### Assertion Consumer Service URL
 
--- a/docs/docs/system-administration/authentication/01-SAML/index.md
+++ b/docs/docs/system-administration/authentication/01-SAML/index.md
@@ -24,8 +24,8 @@ In the UI, you will be able to configure the following fields.
 
 **IdP metadata xml**: The metadata from the IdP.
 
-Once you have configured your identity provider, you can download the service provider metadata XML document
-with the button "Download Service Provider Metadata".
+Once you have configured your identity provider, you can download the service provider metadata XML document with the
+button "Download Service Provider Metadata".
 
 ### Assertion Consumer Service URL
 
--- a/docs/docs/system-administration/authentication/01-SAML/index.md
+++ b/docs/docs/system-administration/authentication/01-SAML/index.md
@@ -8,45 +8,24 @@ SAML authentication requires an [Enterprise subscription](https://flagsmith.com/
 
 :::
 
-## Setup (SaaS)
+## Setup
 
-To enable SAML authentication for your Flagsmith organisation, you must send your identity provider metadata XML
-document to [support@flagsmith.com](mailto:support@flagsmith.com).
+To enable SAML authentication for your Flagsmith organisation, you have to go to your organisations settings, and in the
+SAML tab, you'll be able to configure it.
 
-Once Flagsmith has configured your identity provider, we will send you a service provider metadata XML document or an
-Assertion Consumer Service (ACS) URL to use with your identity provider.
+In the UI, you will be able to configure the following fields.
 
-## Setup (self-hosted)
+**Name:** (**Requierd**) A short name for the organisation, used as the input when clicking "Single Sign-on" at login
+(note this is unique across all tenants and will form part of the URL so should only be alphanumeric + '-,\_').
 
-To enable SAML for your Flagsmith organisation in a self-hosted environment, you will need access the
-[Django admin interface](/deployment/configuration/django-admin).
+**Frontend URL**: (**Requierd**) This should be the base URL of the Flagsmith dashboard.
 
-In the Django admin interface, click on the "SAML Configurations" option in the menu on the left. To create a new SAML
-configuration, click on "Add SAML Configuration" in the top right corner.
+**Allow IdP initiated**: This field determines whether logins can be initiated from the IdP.
 
-You should see a screen similar to the following:
+**IdP metadata xml**: The metadata from the IdP.
 
-![SAML Auth Setup](/img/saml-auth-setup.png)
-
-From the drop down next to **Organisation**, select the organisation that you want to configure for SAML authentication.
-
-Next to **Organisation name**, add a URI-safe name that uniquely identifies the organisation. Users will need to provide
-this name when selecting the "Single Sign-On" option at the Flagsmith login screen.
-
-Next to **Frontend URL**, add the URL where your Flagsmith frontend is running. Users will be redirected to this URL
-when they authenticate using SAML.
-
-Copy your identity provider's XML metadata document into the **IdP metadata XML** field, or leave it blank and come back
-to this step later if you do not have it.
-
-If you want to enable IdP-initiated SSO, check the box next to **Allow IdP-initiated (unsolicited) login**. If you are
-unsure, leave this box unchecked.
-
-Hit the **Save** button to create the SAML configuration.
-
-Once your SAML configuration is created, you can download your Flagsmith service provider metadata by going back to the
-list of SAML configurations in the Django admin interface and clicking "Download" on the SAML configuration you just
-created.
+Once Flagsmith has configured your identity provider, you can download the service provider metadata XML document whit
+the button "Download Service Provider Metadata".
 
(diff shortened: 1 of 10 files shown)


matthewelwell:
@novakzaballa I've fixed a couple of minor issues here, but there are some fundamental issues that need resolving. We will, for example, need to add a new endpoint in the SAML repository (which I will work on now).

matthewelwell:
I've added a new list endpoint in the PR [here](https://github.com/Flagsmith/flagsmith-saml/pull/41) which we will need to update this code to use. It's also worth noting that we allow multiple `SamlConfiguration` objects per organisation, so this PR will need to be updated to handle that case - it seems like it only supports a single configuration at the moment. 

Some other changes that will need to be made: 

 1. We should validate the `name` attribute as it needs to be a valid URL path parameter. This means we should restrict the input to only alphanumeric + `-` and `_`. 
 2. We need to make sure that we handle errors from the API since the name field is unique across all organisations in the platform. 
 3. We should add a tooltip explaining what the name field _is_: "An alphanumeric string to uniquely identify the SAML configuration. You will use this as the 'Organisation name' when logging in via the Flagsmith login page. Must only include alphanumeric characters or hyphens / underscores." 
 4. The SamlConfiguration endpoints use the `name` field as the lookup field whereas this PR seems to be trying to use the numeric `id` field after creation. 
 5. The IdP metadata can only be XML, and will be multiline. Can we update the input accordingly? Can the syntax validation be locked to XML perhaps? 

We should also note that (in the near future), we will want to also be able to manage a list of attribute mappings (which consist of a string input, and a select box (values TBC). So we should consider that when designing the UI to allow for multiple SamlConfiguration objects.

matthewelwell:
@novakzaballa thanks for the changes, it's looking a lot better now but there are still a few issues: 

 1. The IdP metadata is quite unmanageable in an auto-expanding text input. Ideally we'd have a file upload option here, but I think we'd need a new endpoint for that? 
 2. The bottom of the modal is missing padding (see screenshots below)
 3. It's not obvious that the 'Name' input label has a tooltip, let's add an icon there to make it clear. We should probably add tooltips to the remaining fields (probably with the exception of the IdP metadata which is fairly self explanatory. 
 4. The alignment on the table is quite odd. 
 5. The 'Edit Configuration' button should say 'Update Configuration'
 6. The 'Download Service Provider Metadata' option doesn't work correctly - it downloads a JSON file which isn't what we need. I think you need to set the accept header to `accept: application/xml` ? 

image
image
image

novakzaballa:
> 1. The IdP metadata is quite unmanageable in an auto-expanding text input. Ideally we'd have a file upload option here, but I think we'd need a new endpoint for that?

@matthewelwell Could the file be loaded locally and automatically parsed as XML, then converted into text? Would that be acceptable?


> 2. The bottom of the modal is missing padding (see screenshots below)

Done
Screenshot 2024-06-06 at 11 39 57 AM


> 3. It's not obvious that the 'Name' input label has a tooltip, let's add an icon there to make it clear. We should probably add tooltips to the remaining fields (probably with the exception of the IdP metadata which is fairly self explanatory.

Done:

Screenshot 2024-06-06 at 11 39 50 AM


> 4. The alignment on the table is quite odd.

Corrected

Screenshot 2024-06-06 at 11 40 09 AM


> 5. The 'Edit Configuration' button should say 'Update Configuration'

Changed

> 6. The 'Download Service Provider Metadata' option doesn't work correctly - it downloads a JSON file which isn't what we need. I think you need to set the accept header to `accept: application/xml` ?

Now we download .XML files instead of .JSON.

matthewelwell:
>@matthewelwell Could the file be loaded locally and automatically parsed as XML, then converted into text? Would that be acceptable?

I'm not sure I fully understand this, but it sounds reasonable - yep. Could we also then add a 'download' option to re-download the IdP metadata instead of showing it directly in the text input?

novakzaballa:
> I'm not sure I fully understand this, but it sounds reasonable - yep. Could we also then add a 'download' option to re-download the IdP metadata instead of showing it directly in the text input?
Screenshot 2024-06-07 at 10 13 23 AM

matthewelwell:
I'm not seeing the 'Show More' UI option here: 

image

It seems like the first time you click on one of the SAML configurations after browsing to the SAML settings page, it doesn't collapse the field. If you close it and reopen, it does collapse the field correctly. 

I also added a minor commit [here](https://github.com/Flagsmith/flagsmith/pull/4055/commits/04f3bbc62d6b76b596bbaceabd576b3961cea707) to ensure the 'Download IDP Metadata' option shows as clickable. Previously, it was just showing the regular cursor.

docs/docs/system-administration/authentication/01-SAML/index.md
```diff
@@ -8,45 +8,24 @@ SAML authentication requires an [Enterprise subscription](https://flagsmith.com/
...
-document to [support@flagsmith.com](mailto:support@flagsmith.com).
+To enable SAML authentication for your Flagsmith organisation, you have to go to your organisations settings, and in the
+SAML tab, you'll be able to configure it.
 
-Once Flagsmith has configured your identity provider, we will send you a service provider metadata XML document or an
-Assertion Consumer Service (ACS) URL to use with your identity provider.
+In the UI, you will be able to configure the following fields.
 
-## Setup (self-hosted)
+**Name:** (**Requierd**) A short name for the organisation, used as the input when clicking "Single Sign-on" at login
```
matthewelwell:
```suggestion
**Name:** (**Required**) A short name for the organisation, used as the input when clicking "Single Sign-on" at login
```

docs/docs/system-administration/authentication/01-SAML/index.md
```diff
@@ -8,45 +8,24 @@ SAML authentication requires an [Enterprise subscription](https://flagsmith.com/
...
+(note this is unique across all tenants and will form part of the URL so should only be alphanumeric + '-,\_').
 
-To enable SAML for your Flagsmith organisation in a self-hosted environment, you will need access the
-[Django admin interface](/deployment/configuration/django-admin).
+**Frontend URL**: (**Requierd**) This should be the base URL of the Flagsmith dashboard.
```
matthewelwell:
```suggestion
**Frontend URL**: (**Required**) This should be the base URL of the Flagsmith dashboard, e.g. https://app.flagsmith.com for SaaS. 
```

docs/docs/system-administration/authentication/01-SAML/index.md
```diff
@@ -8,45 +8,24 @@ SAML authentication requires an [Enterprise subscription](https://flagsmith.com/
...
-
-If you want to enable IdP-initiated SSO, check the box next to **Allow IdP-initiated (unsolicited) login**. If you are
-unsure, leave this box unchecked.
-
-Hit the **Save** button to create the SAML configuration.
-
-Once your SAML configuration is created, you can download your Flagsmith service provider metadata by going back to the
-list of SAML configurations in the Django admin interface and clicking "Download" on the SAML configuration you just
-created.
+Once Flagsmith has configured your identity provider, you can download the service provider metadata XML document whit
```
matthewelwell:
```suggestion
Once Flagsmith you have configured your identity provider, you can download the service provider metadata XML document with
```

matthewelwell (changes requested):
I've tested the UI aspects and all seems good now. 

Added a few comments for some typos / improvements to the documentation.

docs/docs/system-administration/authentication/01-SAML/index.md
novakzaballa: Done

docs/docs/system-administration/authentication/01-SAML/index.md
novakzaballa: Done

docs/docs/system-administration/authentication/01-SAML/index.md
novakzaballa: Done

docs/docs/system-administration/authentication/01-SAML/index.md
```diff
@@ -8,45 +8,24 @@ SAML authentication requires an [Enterprise subscription](https://flagsmith.com/
...
-
-If you want to enable IdP-initiated SSO, check the box next to **Allow IdP-initiated (unsolicited) login**. If you are
-unsure, leave this box unchecked.
-
-Hit the **Save** button to create the SAML configuration.
-
-Once your SAML configuration is created, you can download your Flagsmith service provider metadata by going back to the
-list of SAML configurations in the Django admin interface and clicking "Download" on the SAML configuration you just
-created.
+Once Flagsmith you have configured your identity provider, you can download the service provider metadata XML document
```
matthewelwell:
```suggestion
Once you have configured your identity provider, you can download the service provider metadata XML document
```

matthewelwell (approved):
Approved with one more minor typo (which I'll fix now)

PR #4069, opened 2024-05-31
fix: Metadata UI issues

### LINKED ISSUE #3986: Metadata issues
1. This Tooltip doesn't make any sense to me. I don't think it's necessary as we have the description and link to the docs below. 

image

2. I'm not sure what these `null` values are. Maybe this is old data. 

image

The JSON responses are as follows: 

```json
// GET /metadata/fields/

        {
            "id": 1,
            "name": "JIRA Ticket Number",
            "type": "str",
            "description": "",
            "organisation": 18
        }
```

```json
// GET organisations/:id/metadata-model-fields/
        {
            "id": 1,
            "field": 1,
            "content_type": 39,
            "is_required_for": []
        },
        {
            "id": 2,
            "field": 1,
            "content_type": 55,
            "is_required_for": []
        }
```

3. Requeried -> Required

Screenshot 2024-05-20 at 17 27 58

4. Be more descriptive in locations where metadata should be shown, guiding users to where they can configure the metadata. 

image (3)

novakzaballa:

## Changes

Related with: https://github.com/Flagsmith/flagsmith/issues/3986

> 1. This Tooltip doesn't make any sense to me. I don't think it's necessary as we have the description and link to the docs below. 

Now

Screenshot 2024-05-31 at 12 49 22 PM

> 2. I'm not sure what these `null` values are. Maybe this is old data.

I couldn't reproduce this issue; perhaps it was because the data at that time was correct.

> 3. Requeried -> Required

Screenshot 2024-05-31 at 12 51 01 PM

> 4. Be more descriptive in locations where metadata should be shown, guiding users to where they can configure the metadata.

Screenshot 2024-05-31 at 12 50 33 PM

- Create metadata for some entity


matthewelwell:
This looks a lot better but I still think the 'Metadata' section when creating / editing an entity can be improved. Specifically for Segments and Features. I think we should: 

 1. Add an option to show / hide the metadata
 2. When there are no metadata fields for an entity type, we should perhaps display something beneath the header (perhaps instead of the tooltip).
 
With the above 2 suggestions, I think we could end with something like the following: 

image

Generally, I am also concerned regarding segments that, if the metadata gets too large, then we are overtaking the 'Rules' tab with the metadata which feels wrong. Should we add a new tab to the Segment modal for the metadata instead? 

Also, we need to make sure that we're paying attention to these things on creation of entities. For example, I just noticed that it's not currently possible to create an environment which has required metadata. We should check if the clone behaviour works correctly (and decide whether it _should_ be included in the clone behaviour).

matthewelwell:
@novakzaballa this is better, but there are still some issues. I've fixed a couple of typos and added a link to the project settings in a commit [here](https://github.com/Flagsmith/flagsmith/pull/4069/commits/d15d6c877c4ba10b458e1bd72234c3054fe3b3f5). 

I'm also still not happy with the approach for showing metadata for segments. I think it should be it's own tab. Can we add that please?

novakzaballa:
> @novakzaballa this is better, but there are still some issues. I've fixed a couple of typos and added a link to the project settings in a commit [here](https://github.com/Flagsmith/flagsmith/pull/4069/commits/d15d6c877c4ba10b458e1bd72234c3054fe3b3f5).
> 
> I'm also still not happy with the approach for showing metadata for segments. I think it should be it's own tab. Can we add that please?

@matthewelwell I added the Tab

matthewelwell:
Thanks for adding the tab, I think that's much better. I've added a few more pieces of feedback below: 

This shouldn't be collapsed by default, or even collapsible at all really now that it's in its own tab

image

When creating a new segment, this icon indicating that something has changed looks like it's something in the metadata tab that needs attention. Can this icon be aware of the tab in which something has changed? 

image

I like that this icon shows when a specific metadata field has been edited, but it can be removed when just switching between the 2 tabs. If this is difficult to resolve, we can just remove it entirely. 

image

novakzaballa:
> Thanks for adding the tab, I think that's much better. I've added a few more pieces of feedback below:
> 
> This shouldn't be collapsed by default, or even collapsible at all really now that it's in its own tab
> 
> image
> When creating a new segment, this icon indicating that something has changed looks like it's something in the metadata tab that needs attention. Can this icon be aware of the tab in which something has changed?
> 
> image
> I like that this icon shows when a specific metadata field has been edited, but it can be removed when just switching between the 2 tabs. If this is difficult to resolve, we can just remove it entirely.
> 
> image

I have already made the changes.

matthewelwell:
Looks like there's still an issue regarding the icon displaying whether something has changed. When I edit the metadata, the icon shows that the rules have changed. 

https://www.loom.com/share/07667735cdf041d4b1cc4c6ac08c1e1e?sid=f76cb099-0c77-4e15-8f47-1d30525f97cd

Also, it's odd to me that, when creating a segment, the metadata tab has an icon indicating that it's been changed, but the Basic Configuration tab doesn't. My preference would be that, on create, we don't show the icon at all. On edit, we show the icon only on the tabs that has/have been edited. 

image

novakzaballa:
> Also, it's odd to me that, when creating a segment, the metadata tab has an icon indicating that it's been changed, but the Basic Configuration tab doesn't. My preference would be that, on create, we don't show the icon at all. On edit, we show the icon only on the tabs that has/have been edited.


Ok, I made the last changes.

matthewelwell (approved):

Issue #4051, opened 2024-05-29
Boolean feature state changes in versioning v2 are not saved to audit logs


rolodato:
### How are you running Flagsmith

- [ ] Self Hosted with Docker
- [ ] Self Hosted with Kubernetes
- [X] SaaS at flagsmith.com
- [ ] Some other way (add details in description below)

### Describe the bug

See reproduction steps.

### Steps To Reproduce

1. Go to any environment with Versioning v2 enabled
2. Change the boolean state of any feature (not the remote config value) and save changes

This change is not reflected in the audit log.

### Expected behavior

All changes should be saved to the audit log.

### Screenshots

https://github.com/Flagsmith/flagsmith/assets/829698/1dd75192-1159-423e-9744-508fc5c33a67

matthewelwell:
This behaviour is being resolved as follows: 

 1. I have added logic to create an audit log record when a new version is published for a given feature
 2. This audit log record will link directly to the history to show the change details in the new version
 
 The first PR for this is here: #4064, the next stage is to implement the FE. 
 
@kyle-ssg I'm using this issue here to provide the information about the changes we discussed earlier today. 

Some details you will need to know: 

 1. The AuditLog record will have `related_object_type=EF_VERSION` and `related_object_uuid=<uuid>` - this will be how you determine the uuid to link to the new version page we discussed. 
 2. The endpoint to retrieve a feature version has been added (you can see it [here](https://api-staging.flagsmith.com/api/v1/docs/#/api/api_v1_environments_features_versions_read)) and includes the `previous_version_uuid` attribute which you can use to get the previous version for comparison purposes.

PR #4077, opened 2024-06-03
fix: facilitate FE display of environment version from audit log

matthewelwell:

## Changes

This PR addresses 2 issues found by the FE in the integration as per [this issue](https://github.com/Flagsmith/flagsmith/issues/4051): 

 1. Adds `related_object_uuid` to the response when retrieving audit log records from the API
 2. Adds a new endpoint to retrieve an `EnvironmentFeatureVersion` without the need for the environment and feature args in the URL path. This is due to the need from the FE to retrieve the version directly from the audit log which does not have the id of the feature directly available. 

Note that, based on this, I have removed the `RetrieveModelMixin` from the `EnvironmentFeatureVersionViewSet` since it is not needed. Technically this is a breaking API change but, since the versioning feature has not been released anywhere yet, I don't see that this is an issue. 

Added a bunch of new tests.


api/features/versioning/views.py
```diff
@@ -46,7 +47,6 @@
 )
 class EnvironmentFeatureVersionViewSet(
     GenericViewSet,
-    RetrieveModelMixin,
```
matthewelwell, line 49: As mentioned in the description, this is technically a breaking API change but, since the versioning feature has not been released to anyone yet, I don't see this as an issue.

gagantrivedi (approved):

PR #4084, opened 2024-06-03
ci: standardise conventional commits

dabeeeenster:
- Rationalise Conventional Commits to this list added to the docs in this PR
- Auto label PRs based on commit title
- Add docs with approved project labels

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -15,3 +15,7 @@ api:
 github:
   - changed-files:
       - any-glob-to-any-file: .github/**
+
+infrastructure:
+  - changed-files:
+      - any-glob-to-any-file: infrastructure/**
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -12,10 +12,6 @@ api:
   - changed-files:
       - any-glob-to-any-file: api/**
 
-ci-cd:
-  - changed-files:
-      - any-glob-to-any-file: .github/**
-
 infrastructure:
   - changed-files:
       - any-glob-to-any-file: infrastructure/**
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -12,7 +12,7 @@ api:
   - changed-files:
       - any-glob-to-any-file: api/**
 
-github:
+ci-cd:
   - changed-files:
       - any-glob-to-any-file: .github/**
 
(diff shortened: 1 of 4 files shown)


matthewelwell:
@dabeeeenster heads up that I cancelled the vercel deployments to try and jump the ones I need up the queue. This PR doesn't need vercel preview deployments.

.github/workflows/platform-pull-request.yml
```diff
@@ -12,22 +12,40 @@ jobs:
...
         with:
           types: | # mirrors changelog-sections in the /release-please-config.json
-            fix
             feat
+            fix
             infra
+            ci
             docs
-            chore
-            build
```
zachaysan, line 25: Do we not need the `build` designator?

zachaysan (approved):
Conditionally approved but I'm not sure if we want to discard the `build` category

.github/workflows/platform-pull-request.yml
dabeeeenster: I wanted to reduce the number of labels overall where possible and don't feel its too different to `ci` unless I'm missing something?

.github/labeler.yml
```diff
@@ -15,3 +15,7 @@ api:
 github:
```
matthewelwell: I'd say the `github` label is unclear tbh. We should just use the `ci` label here imo or, if this is a different _category_ of labelling to what we're doing with conv commits [here](https://github.com/Flagsmith/flagsmith/pull/4084/files#diff-88b40629df399d3611837cb436b6d43f743791f1f5944ea002bb76bf3713046aR32), maybe we should use `workflows` instead?

docs/docs/platform/contributing.md
```diff
@@ -16,6 +16,22 @@ suggesting the next steps.
 - If your PR involves a lot of commits, squash them using `git rebase -i` as this makes it easier for us to review.
 - Keep lines under 80 characters.
 
+### Conventional Commits
+
+Please title your PR's with starting with one of the following
```
matthewelwell:
```suggestion
Please prefix the title of your PRs with one of the following
```

matthewelwell (changes requested):

.github/labeler.yml
dabeeeenster: Agreed - had this thought in the shower this morning 🤣

.github/workflows/platform-pull-request.yml
zachaysan: Ok that makes sense.

matthewelwell (approved):

PR #4088, opened 2024-06-03
fix(versioning): prevent FeatureSegment from writing audit log on delete when v2 versioning enabled

matthewelwell:

## Changes

Prevents the application from writing an AuditLog record when a feature segment is deleted if v2 versioning is enabled. 

Added a new unit test.

### DIFF
--- a/api/features/models.py
+++ b/api/features/models.py
@@ -389,6 +389,11 @@ class FeatureSegment(
     def get_audit_log_related_object_id(self, history_instance) -> int:
         return self.feature_id
 
+    def get_skip_create_audit_log(self) -> bool:
+        # Don't create audit logs when deleting feature segments using versioning
+        # v2 as we rely on the version history instead.
+        return self.environment_feature_version_id is not None
+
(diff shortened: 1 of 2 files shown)


api/features/models.py
```diff
@@ -389,6 +389,11 @@ def to_id_priority_tuple_pairs(
     def get_audit_log_related_object_id(self, history_instance) -> int:
         return self.feature_id
 
+    def get_skip_create_audit_log(self) -> bool:
+        # Don't create audit logs when deleting feature segments using versioning
+        # v2 as we rely on the version history instead.
+        return self.environment_feature_version_id is not None
```
zachaysan, line 395: nit: I would do the following
```suggestion
    @property
    def is_skip_create_audit_log(self) -> bool:
        # Don't create audit logs when deleting feature segments using versioning
        # v2 as we rely on the version history instead.
        return self.environment_feature_version_id is not None
```

zachaysan (approved):
Approved with one minor quality of life improvement suggestion

PR #4091, opened 2024-06-03
fix(versioning): ensure that audit log record is created when committing versions via CR

matthewelwell:

## Changes

This PR ensures that the correct `environment_feature_version_published` signal is sent when we publish a ChangeRequest that includes EnvironmentFeatureVersion objects. 

Added new test

### DIFF
--- a/api/features/workflows/core/models.py
+++ b/api/features/workflows/core/models.py
@@ -38,6 +38,7 @@ from audit.tasks import (
 from environments.tasks import rebuild_environment_document
 from features.models import FeatureState
 from features.versioning.models import EnvironmentFeatureVersion
+from features.versioning.signals import environment_feature_version_published
 from features.versioning.tasks import trigger_update_version_webhooks
 from features.workflows.core.exceptions import (
     CannotApproveOwnChangeRequest,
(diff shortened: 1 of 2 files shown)


api/features/workflows/core/models.py
```diff
@@ -169,6 +170,9 @@ def _publish_environment_feature_versions(
                     kwargs={"environment_id": self.environment_id},
                     delay_until=environment_feature_version.live_from,
                 )
+                environment_feature_version_published.send(
+                    EnvironmentFeatureVersion, instance=environment_feature_version
+                )
```
zachaysan, line 175: I can't see from this review where `environment_feature_version_published` is actually coming from.

zachaysan (commented):
The testing code looks good but I'm not sure about the variable noted below

api/features/workflows/core/models.py
matthewelwell: It's a signal defined [here](https://github.com/Flagsmith/flagsmith/blob/main/api/features/versioning/signals.py) and consumed [here](https://github.com/Flagsmith/flagsmith/blob/main/api/features/versioning/receivers.py#L42-L64).

I kind of hate using signals because it can be a bit hard to track the execution stack, but I seem to remember having no choice in this case.

zachaysan (approved):

PR #4095, opened 2024-06-04
feat: Add default rule for segments

### LINKED ISSUE #3998: Segment condition editor "Or" and "Delete" buttons are buggy

- [ ] Self Hosted with Docker
- [X] SaaS at flagsmith.com

### Describe the bug

I'm not sure how to describe the behaviour here, but the "Or" and "delete" buttons for segment editors seem very broken. Depending on what order things are clicked in, either button will not work, the "Delete" button will add a condition, or sometimes the "Or" button will add more than 1 condition in a different place.

### Steps To Reproduce

Click around randomly as seen in this video. It doesn't always happen, I'm not exactly sure what the pattern is to trigger this behaviour.

https://github.com/Flagsmith/flagsmith/assets/829698/c6e2643b-a836-4b61-91a6-5cb65546f0bd

### Expected behavior

Segment condition editor should work as expected.

### Screenshots

_No response_

### LINKED ISSUE #4000: Improve UX for segment condition descriptions

This is the UI to create a new segment:

image

The button "Show condition descriptions" does nothing until you figure out you need to create a condition first for this to make sense. This button is also confusingly named, since the option "Show condition descriptions" looks to be disabled, but this changes to "Hide condition descriptions" once enabled:

image

I understand why this button exists, since adding a "description" field below every condition can get big fast:

image

Segment descriptions can be very valuable and we should encourage their use. We should get rid of the "Show condition descriptions" button to remove a useless choice for unfamiliar users and provide a better way to show and edit descriptions.

One alternative would be to collapse all the available actions for a condition (edit, delete, add "Or") into a collapsible menu. Example from Vercel's environment variable editor:

Screenshot 2024-05-21 at 16 51 05

This would have the added benefit of making it less likely to fat-finger the Delete button, which currently requires no confirmation and has no undo.

A quick fix might be to simply reduce the placeholder height for descriptions, and expand them as needed up to a certain limit. For example, this screenshot shows the default textbox height compared to the smallest my browser allows me to resize them:

Screenshot 2024-05-21 at 16 54 26

I would be okay with something similar to this as well - the goal here is to reduce friction for creating segments.

### Additional context

Related issues:

https://github.com/Flagsmith/flagsmith/issues/3920
https://github.com/Flagsmith/flagsmith/issues/3918

### LINKED ISSUE #4001: Add a placeholder AND condition when creating segments

This is the UI to create a new segment:

image

To proceed, users must choose between "Add AND Condition" or "Add AND NOT Condition". I don't have data to back this up, but I would guess the vast majority of users will start by creating an AND condition. For unfamiliar users and demos, this masks what segments are for - if we showed the "trait" "condition" "value" boxes straight away, this becomes more self-evident by just looking at the UI. The word "trait" is mentioned twice in the text here, but it's not clear what it's referring to if you're not already familiar with segments.

The only situation that absolutely requires an AND NOT condition is if using an operator that does not have an opposite equivalent. As of today, these are only `Modulo` and `In`. `Matches regex` also does not have an opposite, but any regex is trivial to negate. All of these are fairly advanced use cases anyway.

Show the segment creation UI with one empty AND condition placeholder, something like this:

image

This makes it more obvious what segments are for and lets users define a simple segment without needing to choose between AND or AND NOT.

This also has the benefit of removing the second "NOTE" banner which just looks ugly.

We might want to make a larger redesign of the segment condition editor as described here https://github.com/Flagsmith/flagsmith/issues/4000

However this issue is a quick win that can be implemented today with the current UI.

### Additional context

_No response_

kyle-ssg:

## Changes

Adds placeholder rule for segments

Click create segment

### DIFF
--- a/frontend/web/components/modals/CreateSegment.tsx
+++ b/frontend/web/components/modals/CreateSegment.tsx
@@ -319,49 +319,51 @@ const CreateSegment: FC<CreateSegmentType> = ({
   }, [operators, rules])
   //Find any non-deleted rules
   const hasNoRules = !rules[0]?.rules?.find((v) => !v.delete)
-
+  const rulesToShow = rules[0].rules.filter((v) => !v.delete)
   const rulesEl = (
     <div className='overflow-visible'>
       <div>
         <div className='mb-4'>
-          {rules[0].rules
-            ?.filter((v) => !v?.delete)
-            .map((rule, i) => {
-              return (
-                <div key={i}>
-                  <Row
-                    className={classNames('and-divider my-1', {
-                      'text-danger': rule.type !== 'ANY',
-                    })}
-                  >
-                    <Flex className='and-divider__line' />
-                    {Format.camelCase(
-                      `${i > 0 ? 'And ' : ''}${
-                        rule.type === 'ANY'
-                          ? 'Any of the following'
-                          : 'None of the following'
-                      }`,
-                    )}
-                    <Flex className='and-divider__line' />
-                  </Row>
-                  <Rule
-                    showDescription={showDescriptions}
-                    readOnly={readOnly}
-                    data-test={`rule-${i}`}
-                    rule={rule}
-                    operators={operators}
-                    onRemove={() => {
-                      setValueChanged(true)
-                      removeRule(0, i)
-                    }}
-                    onChange={(v: SegmentRule) => {
-                      setValueChanged(true)
-                      updateRule(0, i, v)
-                    }}
-                  />
-                </div>
-              )
-            })}
+          {rules[0].rules.map((rule, i) => {
+            if (rule.delete) {
+              return null
+            }
+            const displayIndex = rulesToShow.indexOf(rule)
+            return (
+              <div key={i}>
+                <Row
+                  className={classNames('and-divider my-1', {
+                    'text-danger': rule.type !== 'ANY',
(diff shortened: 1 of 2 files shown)


matthewelwell:
@kyle-ssg I think E2E is legitimately broken here.

kyle-ssg:
Ah, yep I bet I adjusted tests when I originally removed this

rolodato (approved):

kyle-ssg:
@rolodato could you re-test this since it now also fixes the linked buggy segment rules

frontend/web/components/modals/CreateSegment.tsx
```diff
@@ -319,49 +319,51 @@ const CreateSegment: FC<CreateSegmentType> = ({
   }, [operators, rules])
   //Find any non-deleted rules
   const hasNoRules = !rules[0]?.rules?.find((v) => !v.delete)
-
+  const rulesToShow = rules[0].rules.filter((v) => !v.delete)
```
rolodato, line 322: Feel free to ignore this if I'm wrong - it just caught my attention that we're guarding against `rules[0]` being undefined or null on the line above, but not this one.

rolodato (approved):
Works great thanks! Left a minor comment, don't even take the time to reply if it doesn't apply :D

PR #4104, opened 2024-06-04
fix: environment click sizes

### LINKED ISSUE #4089: Audit log click target for environment too large

- [ ] Self Hosted with Docker
- [X] SaaS at flagsmith.com

### Describe the bug

In the list of audit logs, within a single entry, there are 2 click targets: one to visit the actual audit log entry, and another one to visit the environment that the entry is about. Here is the click target to view the current environment - clicking anywhere else will open the audit log entry:

image

I've had it happen way too often that I accidentally click on the blank area next to the environment name, and it opens the environment instead of the audit log entry. The vast majority of users opening the audit log will want to look at audit log entries and not a specific environment.

### Steps To Reproduce

1. Go to Audit Log
2. Click on the blank space between the environment name and the entry description

The environment opens and not the audit log entry, which is unexpected.

### Expected behavior

There are 2 alternatives I've considered here:

1. Change the click target for the environment name to only include the environment name itself, and not any blank space
2. Remove the click target for the environment altogether, and add a link to the relevant environment from within the audit log entry as shown here ("Development" here is not a link to anything):

image

I would prefer option 2 as there is no visual indication that there are different click targets, besides looking at the browser's target link when hovering. Once a user has identified the entry they need (without needing to worry about misclicking), they can drill down into the environment from the audit log entry if needed.

### Screenshots

_No response_

kyle-ssg:

## Changes
- Adjusts click area for environment column of audit log

### DIFF
--- a/frontend/web/components/AuditLog.tsx
+++ b/frontend/web/components/AuditLog.tsx
@@ -65,7 +65,7 @@ const AuditLog: FC<AuditLogType> = (props) => {
       search,
     },
     {
-      refetchOnMountOrArgChange: true
+      refetchOnMountOrArgChange: true,
     },
   )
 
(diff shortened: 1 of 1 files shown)


rolodato (approved):
I've tested this and it works well 👍

Any reason why you preferred this option instead of removing the link to the environment from the audit log list, and adding a link to the environment from the individual audit log entry instead?

Other things I don't like about having a link to the environment in the list:

* Environments that were deleted will link to a 404
* We support filtering audit log by environment, and it's not unreasonable to expect that clicking on the environment name on an audit log row would perform the same filtering
* It's still possible to accidentally click on the environment name and not realise this is a different click target. It's reasonable to expect customers to never use audit logs except during a crisis/downtime situation, so I strongly prefer any alternatives with the least surprises/cognitive load here

kyle-ssg:
Hmm yeah good point I think you're right, I think let's keep this for now (I think it also improves UI) and plan to migrate actions to the far right(like in the feature list) since with versioning we'll also have a link to the version

PR #4113, opened 2024-06-05
feat: add UserSignUpMeta model

matthewelwell:

- [ ] I have run [`pre-commit`](https://docs.flagsmith.com/platform/contributing#pre-commit) to check linting

## Changes

As part of a request from the marketing team we need to capture, and send to Hubspot, information about how the user arrived at Flagsmith when a new user signs up. For example, when clicking on a google ad, the website receives a query string that the FE then adds to a cookie. 

This PR adds logic to retrieve the data from that cookie and store it in a new model which is then (TODO) used to send the correct attribution data to Hubspot. 

TODO


api/users/migrations/0037_add_user_sign_up_meta_model.py
```diff
@@ -0,0 +1,25 @@
...
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='UserSignUpMeta',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('created_at', models.DateTimeField(auto_now_add=True)),
+                ('updated_at', models.DateTimeField(auto_now=True)),
+                ('json_meta', models.TextField()),
```
zachaysan, line 21: It's too bad we can't use a json field here due to other databases

zachaysan (commented):
Overall approach looks great. I'd have marked this PR as approved but according to the PR description there are testing and other TODOs that need to land first.

matthewelwell:
Closing as this was a failed attempt at getting around Hubspots uselessness. We will need to develop a custom form in hubspot and use that to create the contacts instead of using their native APIs.

PR #4117, opened 2024-06-05
deps: bump workflows version

matthewelwell:
Thanks for submitting a PR! Please check the boxes below:

- [x] I have run [`pre-commit`](https://docs.flagsmith.com/platform/contributing#pre-commit) to check linting
- [ ] I have added information to `docs/` if required so people know about the feature!
- [x] I have filled in the "Changes" section below?
- [x] I have filled in the "How did you test this code" section below?
- [x] I have used a [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) title for this Pull Request

## Changes

Bump workflows version 2.3.4

## How did you test this code?

Tests in workflows repo

PR #4118, opened 2024-06-05
fix(versioning): use version live from

matthewelwell:

## Changes

Use the live_from value from the first environment feature version instead of the feature state if it exists. 

Need to first merge the code in #4117, then we can test this using the preview URL.

### DIFF
--- a/frontend/web/components/pages/ChangeRequestPage.js
+++ b/frontend/web/components/pages/ChangeRequestPage.js
@@ -151,9 +151,10 @@ const ChangeRequestsPage = class extends Component {
   publishChangeRequest = () => {
     const id = this.props.match.params.id
     const changeRequest = ChangeRequestStore.model[id]
-    const scheduledDate = changeRequest.environment_feature_versions
-      ? moment(changeRequest.environment_feature_versions[0].live_from)
-      : moment(changeRequest.feature_states[0].live_from)
+    const scheduledDate =
+      changeRequest.environment_feature_versions.length > 0
+        ? moment(changeRequest.environment_feature_versions[0].live_from)
+        : moment(changeRequest.feature_states[0].live_from)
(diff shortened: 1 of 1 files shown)


frontend/web/components/pages/ChangeRequestPage.js
```diff
@@ -154,7 +154,9 @@ const ChangeRequestsPage = class extends Component {
     const isScheduled =
       new Date(changeRequest.feature_states[0].live_from).valueOf() >
       new Date().valueOf()
-    const scheduledDate = moment(changeRequest.feature_states[0].live_from)
+    const scheduledDate = changeRequest.feature_states
+      ? moment(changeRequest.feature_states[0].live_from)
```
novakzaballa: Is this logic correct? I think it doesn't align with the PR description since it prioritizes the feature_states over the environment_feature_versions:
> Use the live_from value from the first environment feature version instead of the feature state if it exists.

novakzaballa (approved):
Approved but left a comment, please verify the comment.

frontend/web/components/pages/ChangeRequestPage.js
matthewelwell: A change request can only have either `feature_states` or `environment_feature_versions` so the result will be the same whichever way round we have it.

novakzaballa: Oh Ok, then it will work. Approved

matthewelwell: I actually did have to swap them around since the FE pulls the feature_states out of the environment_feature_versions but the feature_states do not have the live_from attribute.

novakzaballa: LGTM

PR #4119, opened 2024-06-05
fix(versioning): scheduled changes incorrectly considered live

matthewelwell:

## Changes

This PR fixes an issue where scheduled changes were incorrectly being considered live. 

Added a unit test.

### DIFF
--- a/api/features/versioning/managers.py
+++ b/api/features/versioning/managers.py
@@ -57,8 +57,11 @@ class EnvironmentFeatureVersionManager(SoftDeleteManager):
             params={
                 "environment_id": environment_id,
                 "api_key": environment_api_key,
-                # It seems as though there is a timezone issue when using postgres's
-                # built in now() function, so we pass in the current time from python.
+                # TODO:
+                #  It seems as though there is a timezone issue when using postgres's
+                #  built in now() function, so we pass in the current time from python.
+                #  Using <= now() in the SQL query returns incorrect results.
+                #  More investigation is needed here to understand the cause.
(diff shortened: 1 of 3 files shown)


api/features/versioning/managers.py
```diff
@@ -53,5 +54,11 @@ def _get_latest_versions(
 
         return self.raw(
             get_latest_versions_sql,
-            params={"environment_id": environment_id, "api_key": environment_api_key},
+            params={
+                "environment_id": environment_id,
+                "api_key": environment_api_key,
+                # It seems as though there is a timezone issue when using postgres's
+                # built in now() function, so we pass in the current time from python.
+                "live_from_before": timezone.now().isoformat(),
```
novakzaballa, line 65: I would prefer avoiding the use of Python's now() here. Maybe we can leave a TODO here to revisit it in the future?

novakzaballa (approved):
I Approved it to allow its deployment, though I left some comments.

api/features/versioning/managers.py
matthewelwell: I've updated the comment to use a TODO.

matthewelwell: Yeah, so this works (after updating the test to schedule the other version for 2 hours later): 

```sql
and efv2."live_from" <= now() + interval '90 minutes'
```

Indicating that it likely is some TZ issue here.

matthewelwell: Actually, that's not true because even this works: 

```sql
and efv2."live_from" <= now() + interval '1 second'
```

PR #4126, opened 2024-06-06
infra: task processor settings tweaks

matthewelwell:

## Changes

Tweak the task processor retention settings to: 

 1. reduce the retention period (only by 1 day as we want to limit the impact on the database that deleting multiple days would cause). We should ideally continue to decrement this number until we get down to 30. 
 2. increase the batch size in which tasks are deleted so that it's not having to run as many batches when deleting tasks. We should ideally continue to increment this number until we get to 500,000. 

Needs testing in production when the next scheduled task runs to delete tasks.


khvn26 (approved):
Let's try this!

PR #4136, opened 2024-06-10
fix(django-upgrade): upgrade django major version 

gagantrivedi:

## Changes

BREAKING CHANGE: Upgrade django major version

Updates tests


matthewelwell:
@gagantrivedi just a heads up that I've updated the title of this PR to mark it as breaking

api/tests/unit/features/test_unit_features_views.py
```diff
@@ -2588,6 +2589,10 @@ def test_list_features_n_plus_1(
 ) -> None:
     # Given
     with_project_permissions([VIEW_PROJECT])
+    num_queries = 16
+
+    if settings.IS_RBAC_INSTALLED:  # pragma: no cover
+        num_queries += 1
```
khvn26: I'd rather parametrise the test and mock the setting explicitly.

api/tests/unit/segments/test_unit_segments_views.py
```diff
@@ -375,9 +376,15 @@ def test_list_segments(
...
     num_queries: int,
     required_a_segment_metadata_field: MetadataModelField,
+    is_admin_master_api_key_client: bool,
 ):
     # Given
     num_segments = 5
+    if (
+        settings.IS_RBAC_INSTALLED and not is_admin_master_api_key_client
+    ):  # pragma: no cover
+        num_queries += 1
```
khvn26: This definitely should be parametrised.

khvn26 (changes requested):

api/tests/unit/features/test_unit_features_views.py
```diff
@@ -3092,6 +3101,10 @@ def test_feature_list_last_modified_values(
     environment_v2_versioning_2 = Environment.objects.create(
         name="environment 2", project=project, use_v2_feature_versioning=True
     )
+    num_queries = 18
+
+    if settings.IS_RBAC_INSTALLED:  # pragma: no cover
+        num_queries += 1
```
khvn26: See https://github.com/Flagsmith/flagsmith/pull/4136/files#r1673612832

api/tests/unit/features/test_unit_features_views.py
gagantrivedi: I'd rather not mock the settings and cover both the cases

khvn26: > I'd rather not mock the settings

I can't understand what is wrong with mocking the settings here?

> cover both the cases

We can still keep it DRY and cover both paths by same test code if we parametrise the mock.

On the contrary, the current approach requires to setup the test run twice with different setting modules to cover both paths (as evident by the `pragma: no cover` comment).

gagantrivedi: > I can't understand what is wrong with mocking the settings here?

Mocking the setting is irrelevant here... the setting is only being used to identify if the module is installed or not. If we have to mock something, it will be functions provided by rbac module

> On the contrary, the current approach requires to setup the test run twice with different setting modules to cover both paths (as evident by the pragma: no cover comment).

Different code path here are being covered by different workflows

api/environments/identities/models.py
```diff
@@ -163,7 +163,9 @@ def get_segments(
         :return: List of matching segments
         """
         matching_segments = []
-        traits = self.identity_traits.all() if traits is None else traits
+        traits = (
+            self.identity_traits.all() if (traits is None and self.id) else traits or []
```
matthewelwell, line 167: I'm curious to know which scenarios exist where we don't have `self.id` set and we're calling this method?

api/tests/unit/features/feature_segments/test_unit_feature_segments_views.py
```diff
@@ -28,29 +30,37 @@
...
+    project: Project,
+    django_assert_num_queries: DjangoAssertNumQueries,
+    client: APIClient,
+    feature_segment: FeatureSegment,
+    num_queries: int,
+    is_admin_master_api_key_client: bool,
 ):
     # Given
+    if (
+        settings.IS_RBAC_INSTALLED and not is_admin_master_api_key_client
```
matthewelwell: This feels slightly odd to me: 

 1. I'd rather not have to use another parametrized argument for the boolean if we can avoid it (although I have looked into it myself and can't see any obvious options). 
 2. If we do need to use the extra argument, why not invert it and use `is_admin_client` rather than `is_admin_master_api_key_client`? 

Either way, it'd be good to add a comment as to why we expect the extra query in this scenario.

api/tests/unit/sales_dashboard/test_unit_sales_dashboard_views.py
```diff
@@ -75,18 +75,16 @@ def test_get_organisation_info__get_event_list_for_organisation(
 
 def test_list_organisations_search_by_name(
     organisation: Organisation,
-    client: Client,
-    admin_user: FFAdminUser,
+    superuser_client: Client,
```
matthewelwell, line 78: What is this `superuser_client` fixture and why can't we use `admin_client` instead?

api/tests/unit/features/test_unit_features_views.py
matthewelwell: I guess what Gagan is trying to achieve here is a generic solution that works for both (one which includes all the private packages) and one which doesn't. 

I guess with @khvn26 's approach, we could add a `pytest.mark.skip` perhaps?

khvn26: A skip mark depending on setting value along with a separate private test would be preferable, agreed.

api/environments/identities/models.py
gagantrivedi: For transient identity, test in question: `test_get_identities__transient__no_persistence`

api/tests/unit/features/test_unit_features_views.py
gagantrivedi: I'd prefer to not duplicate and maintain two tests that pretty much test the same thing, but let me know if you feel strongly about this @khvn26

api/tests/unit/sales_dashboard/test_unit_sales_dashboard_views.py
gagantrivedi: Superuser_client is `django.test.client.Client` and admin_client is `rest_framework.test.APIClient` i,e: admin_client will not work with `@staff_member_required`

api/tests/unit/features/test_unit_features_views.py
matthewelwell: So, I guess to summarse, we have a few options here: 

 1. Keep it as it is, although I agree with @khvn26 that the pragma no cover is a bit of a code smell here. 
 2. Write 2 separate tests, one skipped if `settings.IS_RBAC_INSTALLED = False`. Note that we wouldn't necessarily need to maintain 2 separate tests here, right, the test code could all be in a central, private function, then we just call that from each separate test function. 

I think there probably is a third option (or set of options) where we do mock out the settings instead, right? And let the tests from the rbac module itself handle the code paths that we're trying to test here. I'm not sure how feasible this is?

matthewelwell (approved):
Looks good to me, but keen to get @khvn26 's thoughts on the solution to the main discussion thread.

khvn26 (approved):

PR #4138, opened 2024-06-10
feat: Create versioning for segments

zachaysan:

## Changes

This introduces versioning for segments by cloning their features into new objects and filtering out older versions of segment through a new manager class.

A few new unit tests around the models as well as updates to some existing tests to handle audit log cases.


api/segments/migrations/sql/0023_add_versioning_to_segments.sql
```diff
@@ -0,0 +1,2 @@
+UPDATE segments_segment SET version_of_id = id;
+UPDATE segments_historicalsegment SET version_of_id = id;
```
matthewelwell: Hmm... I'm not sure that we should necessarily operate on the historical segment table like this. I don't think we _need_ to since we never use the data in those tables anyway.

api/segments/models.py
```diff
@@ -84,6 +111,46 @@ def id_exists_in_rules_data(rules_data: typing.List[dict]) -> bool:
...
+            self.version = 1
+
+    @hook(AFTER_CREATE, when="version_of", is_now=None)
+    def set_version_of_to_self_if_none(self):
+        """
+        This allows the segment model to reference all versions of
+        itself including itself.
+        """
+        self.version_of = self
+        self.save()
```
matthewelwell, line 142: Hmm... I don't _love_ that it has to call save on itself again after create, but I'm not sure there is another option here since we only get the id after create.

api/segments/models.py
```diff
@@ -84,6 +111,46 @@ def id_exists_in_rules_data(rules_data: typing.List[dict]) -> bool:
...
+
+    def deep_clone(self) -> "Segment":
+        new_segment = Segment.objects.create(
+            name=self.name,
+            description=self.description,
+            project=self.project,
+            feature=self.feature,
+            version=self.version,
+            version_of=self,
+        )
```
matthewelwell: We have clone functions in other models in the code base where we take a slightly different approach, e.g.: 

```python
def deep_clone(self) -> "Segment":
    new_segment = copy.deepcopy(self)

    new_segment.id = None  # this is what forces django to think it's creating a new model object

    ...
```

The benefit of this approach is that it maintains itself when new fields are added to those model classes. With the approach that you're using here, if a new field is added to the segment, we'll need to remember to add it here I think?

api/segments/models.py
```diff
@@ -140,6 +207,51 @@ def get_segment(self):
             rule = rule.rule
         return rule.segment
 
+    def deep_clone(self, versioned_segment: Segment) -> "SegmentRule":
+        if self.rule:
+            assert False, "Unexpected rule, expecting segment set not rule"
```
matthewelwell, line 237: This might need a little bit more explanation. I think I understand (from reading the rest of the code in this method) but it's not immediately clear why we only expect this type of rule.

api/segments/serializers.py
```diff
@@ -81,11 +81,15 @@ def create(self, validated_data):
...
 
-    def update(self, instance, validated_data):
+    def update(self, instance: Segment, validated_data: dict[str, typing.Any]) -> None:
         # use the initial data since we need the ids included to determine which to update & which to create
         rules_data = self.initial_data.pop("rules", [])
         metadata_data = validated_data.pop("metadata", [])
         self.validate_segment_rules_conditions_limit(rules_data)
+
+        # Create a version of the segment now that we're updating.
+        instance.deep_clone()
```
matthewelwell: We might need to be more clever here... I'm worried about failure modes in the update that would mean we end up with phantom versions. 

We could wrap the below in a `try` / `except` perhaps, or we could move this logic to the model.

zachaysan: Do you mean a `try` / `except` in the logic that handles updates? If so, we'd be altering the segment before cloning it which would defeat the purpose, no?

api/segments/models.py
zachaysan: Ok I've added a comment to explain it.

api/segments/models.py
zachaysan: Ok I'll implement that strategy. Just to double check, the timestamps will still be set to `now` for the new model, right?

api/segments/models.py
zachaysan: I tried setting the model to itself before create and it threw an exception. I think our only option is to keep this behaviour.

api/segments/migrations/sql/0023_add_versioning_to_segments.sql
zachaysan: Ok I've removed the historical segment part. I can see it working either way, but if you want it removed that works for me.

api/segments/models.py
zachaysan: Ok this is done.

api/segments/models.py
matthewelwell: Yep, fair enough I guess.

api/segments/models.py
matthewelwell: Hmm... I'd argue that the comment doesn't explain it though, it just states the same thing that one can interpret from the code. What I'm looking for is the 'why' - why are we only expecting a rule here attached to a segment, not a rule.

api/segments/models.py
matthewelwell: >Just to double check, the timestamps will still be set to now for the new model, right?

Good question - I don't know. We should verify in a test. Although I guess the question is whether we want that to be the case. Since we're backfilling the history, wouldn't we _want_ e.g. the `created_at` not to change?

api/segments/serializers.py
matthewelwell: Well, yes, I guess in an ideal world, we'd create the clone without persisting it to the database before doing the update and then persist it once the update has happened successfully. 

Since that's not really possible, due to all the foreign key relationships, we could instead just handle the deletion (I guess we'd want to hard delete too) of the already created clone in the `except` - although this still isn't bulletproof.

api/tests/integration/audit/test_audit_logs.py
```diff
@@ -250,7 +250,7 @@ def test_retrieve_audit_log_includes_changes_when_segment_override_created_and_d
     get_audit_logs_response_2 = admin_client.get(get_audit_logs_url)
     assert get_audit_logs_response_2.status_code == status.HTTP_200_OK
     results = get_audit_logs_response_2.json()["results"]
-    assert len(results) == 5
+    assert len(results) == 6
```
matthewelwell: What's the audit log record that we've added here?

api/tests/unit/segments/test_unit_segments_models.py
```diff
@@ -167,3 +169,247 @@ def test_condition_get_update_log_message(segment, segment_rule, mocker):
...
+) -> None:
+    # Given
+    segment = Segment.objects.create(
+        name="SpecialSegment",
+        description="A lovely, special segment.",
+        project=project,
+        feature=feature,
+    )
+
+    # Check that the versioning is correct, since we'll testing
```
matthewelwell:
```suggestion
    # Check that the versioning is correct, since we'll be testing
```

api/tests/unit/segments/test_unit_segments_models.py
```diff
@@ -167,3 +169,247 @@ def test_condition_get_update_log_message(segment, segment_rule, mocker):
...
+    )
+    child_rule2 = SegmentRule.objects.create(
+        rule=parent_rule1, type=SegmentRule.ALL_RULE
+    )
+    child_rule3 = SegmentRule.objects.create(
+        rule=parent_rule2, type=SegmentRule.NONE_RULE
+    )
+    child_rule4 = SegmentRule.objects.create(
+        rule=parent_rule2, type=SegmentRule.ALL_RULE
+    )
```
matthewelwell: I think we can stick to the structure created by the dashboard here which is just: 

```
parent_rule = SegmentRule.objects.create(
    segment=segment, type=SegmentRule.ALL_RULE,
)

child_rule1 = SegmentRule.objects.create(
    rule=parent_rule, type=SegmentRule.ANY_RULE
)
child_rule2 = SegmentRule.objects.create(
    rule=parent_rule, type=SegmentRule.NONE_RULE
)
...
```

Note that the UI only allows 'any' and 'none' rules nested beneath a single 'all' rule.

api/tests/unit/segments/test_unit_segments_models.py
```diff
@@ -167,3 +169,247 @@ def test_condition_get_update_log_message(segment, segment_rule, mocker):
...
+        value="condition1",
+        created_with_segment=True,
+    )
+    condition2 = Condition.objects.create(
+        rule=parent_rule1,
+        property="parent_rule1",
+        operator=PERCENTAGE_SPLIT,
+        value="0.1",
+        created_with_segment=False,
+    )
```
matthewelwell: Again, the UI doesn't allow for conditions at the parent rule level.

matthewelwell: Again, the UI doesn't allow for conditions at the parent rule level.

api/tests/unit/segments/test_unit_segments_models.py
```diff
@@ -167,3 +169,247 @@ def test_condition_get_update_log_message(segment, segment_rule, mocker):
...
+    Condition.objects.create(
+        rule=child_rule4,
+        property="child_rule4",
+        operator=EQUAL,
+        value="condition7",
+        created_with_segment=True,
+    )
+
+    # When
+    new_segment = segment.deep_clone()
```
matthewelwell: I think we could perhaps improve the naming here and use `cloned_segment` instead of `new_segment` ?

api/tests/unit/segments/test_unit_segments_models.py
```diff
@@ -167,3 +169,247 @@ def test_condition_get_update_log_message(segment, segment_rule, mocker):
...
+
+def test_manager_returns_only_highest_version_of_segments(
+    segment: Segment,
+) -> None:
+    # Given
+    # The built-in segment fixture is pre-versioned already.
+    assert segment.version == 2
+    assert segment.version_of == segment
+
+    new_segment = segment.deep_clone()
```
matthewelwell: As above, let's use `cloned_segment` here?

api/tests/unit/segments/test_unit_segments_models.py
```diff
@@ -167,3 +169,247 @@ def test_condition_get_update_log_message(segment, segment_rule, mocker):
...
+) -> None:
+    # Given
+    segment = Segment.objects.create(
+        name="SpecialSegment",
+        description="A lovely, special segment.",
+        project=project,
+        feature=feature,
+    )
+
+    sub_rule = SegmentRule.objects.create(type=SegmentRule.ALL_RULE)
```
matthewelwell: I'm a little unsure what's going on in this test. This `sub_rule` is not valid since it doesn't have a parent segment or rule and `sub_rule` to me indicates that it's a rule underneath something, whereas below we're using it as a parent of what we define (via the comment) as a parent rule.

api/tests/unit/segments/test_unit_segments_views.py
```diff
@@ -175,11 +176,14 @@ def test_audit_log_created_when_segment_updated(project, segment, client):
 
     # Then
     assert res.status_code == status.HTTP_200_OK
+
+    # Three audit log records created, two for the primary
+    # segment and one for the newly versioned segment.
```
matthewelwell: Hmm... this feels odd. I don't think we should be creating new audit log records for older versions of segments - the operation of updating a segment should only create a single audit log record, the fact that a versioned segment is created is a byproduct of that operation. 

We can override the `get_skip_audit_log` method on the `Segment` (and maybe `SegmentCondition`) models to ensure we're not writing unnecessary audit log records.

matthewelwell: Hmm... this feels odd. I don't think we should be creating new audit log records for older versions of segments - the operation of updating a segment should only create a single audit log record, the fact that a versioned segment is created is a byproduct of that operation. 

We can override the `get_skip_audit_log` method on the `Segment` (and maybe `SegmentRule` / `Condition`?) model to ensure we're not writing unnecessary audit log records.

matthewelwell (changes requested):
I've added a few more comments but I also noticed that none of the `view` tests actually verify that the clone is created (perhaps because they rely on the number of audit records instead, which I'm suggesting that we remove).

api/tests/unit/segments/test_unit_segments_models.py
zachaysan: Yeah I had it backwards in my head when I was writing it. I've fixed this locally. Either way the test is of a misbehaving rule, but I've added the segment to the rule before also adding the rule which has two parents (needed for the test).

api/segments/serializers.py
zachaysan: Ok I've implemented this. Sorta feels a bit cludgy but I guess it's the only way forward.

api/segments/models.py
zachaysan: My understanding is that a rule either belongs to a segment xor it belongs to a rule. There are no exceptions. I've made it more explicit in the comment.

api/segments/models.py
zachaysan: Well it turns out that none of the models involved in this PR have `created_at` or `updated_at` timestamps, although they probably should, at least on the segments. Should I add them to this PR? If so, what should the default be for existing objects `None` or `now()`? Should I include conditions and rules in having timestamps as well?

api/tests/unit/segments/test_unit_segments_models.py
zachaysan: Sounds good. Fixed locally.

api/tests/unit/segments/test_unit_segments_models.py
zachaysan: Sounds good. Fixed this locally.

api/tests/integration/audit/test_audit_logs.py
zachaysan: I've fixed this back to the original value since our conversation about the versioned segments impact on the audit log

api/tests/unit/segments/test_unit_segments_views.py
zachaysan: Ok this is now done.

zachaysan: Ok this is now done.

zachaysan (commented):
One thing I found while I was programming this was that `SegmentRule` is not presently being saved to the audit log. I made methods for skipping the audit log creation in advance of anyone adding audit logging to segment rules, since they will be cloned many times as well. 

Should we make a ticket to add audit logging, including tests for different usage patters triggering audit log, or is it too low of a priority?

zachaysan:
> I've added a few more comments but I also noticed that none of the `view` tests actually verify that the clone is created (perhaps because they rely on the number of audit records instead, which I'm suggesting that we remove).

I've added view tests with segment checking including exception handling now.

api/segments/models.py
```diff
@@ -45,14 +57,44 @@ class Segment(
...
+    all_objects = SoftDeleteExportableManager()
+
     class Meta:
         ordering = ("id",)  # explicit ordering to prevent pagination warnings
 
     def __str__(self):
         return "Segment - %s" % self.name
 
+    def get_skip_create_audit_log(self) -> bool:
+        skip = segment_audit_log_helper.should_skip_audit_log(self.id)
```
zachaysan, line 86: After trying multiple different approaches including instance variables, `.initial_state`, `.state` all of which were wiped during the save process I finally resorted to using a helper to track the state of which audit logs can be created.

api/segments/models.py
matthewelwell: Let's add these fields to all models in the segments app, yes, but let's do it in a separate PR. Regarding the default, we _could_ be smart and look in the objects history but that might end up being a fairly mammoth migration.

api/segments/models.py
```diff
@@ -45,14 +57,44 @@ class Segment(
...
+        if skip is not None:
+            return skip
+
+        try:
+            if self.version_of and self.version_of != self:
+                return True
```
matthewelwell: Can we use `self.version_of_id` here instead to avoid the extra query?

api/segments/models.py
```diff
@@ -84,6 +126,70 @@ def id_exists_in_rules_data(rules_data: typing.List[dict]) -> bool:
...
+        """
+        This allows the segment model to reference all versions of
+        itself including itself.
+        """
+        segment_audit_log_helper.set_skip_audit_log(self.id)
+        self.version_of = self
+        self.save()
+        segment_audit_log_helper.unset_skip_audit_log(self.id)
+
+    def deep_delete(self, hard: bool) -> None:
```
matthewelwell: Maybe that this will become apparent later on but I'm not sure why we need this method. This should just be handled by cascading deletes.

api/segments/models.py
```diff
@@ -84,6 +126,70 @@ def id_exists_in_rules_data(rules_data: typing.List[dict]) -> bool:
...
+            self.delete()
+
+    def deep_clone(self) -> "Segment":
+        cloned_segment = deepcopy(self)
+        cloned_segment.id = None
+        cloned_segment.uuid = uuid.uuid4()
+        cloned_segment.version_of = self
+        cloned_segment.save()
+
+        segment_audit_log_helper.set_skip_audit_log(self.id)
```
matthewelwell, line 152: Couldn't (shouldn't) we resort to overriding the `save` method before doing this?

api/segments/models.py
matthewelwell: I'm still unsure why this is needed vs just the cascade delete. I think the soft delete package also offers a hard delete method as well?

api/segments/models.py
zachaysan: Ok sounds good I've made a new ticket for this.

https://github.com/Flagsmith/flagsmith/issues/4211

api/segments/models.py
zachaysan: Nice catch. Updated here and in the other models.

api/segments/models.py
zachaysan: As we discussed during standup, there are issues when running the code in hard delete mode. Waiting on you to take a look.

api/segments/models.py
zachaysan: We don't want to run the save hook every time with these changes so I think they are better in-place.

api/segments/models.py
zachaysan: The PR is up for the timestamps:

https://github.com/Flagsmith/flagsmith/pull/4236

matthewelwell (approved):

PR #4140, opened 2024-06-11
ci(tests/private-packages):  Add SAML, auth controller and workflow

gagantrivedi:

## Changes
Add SAML, workflow and auth controller to private package test workflow

The workflow ran successfully here: https://github.com/Flagsmith/flagsmith/actions/runs/9533292963/job/26276374433?pr=4140


.github/workflows/api-tests-with-private-packages.yml
```diff
@@ -42,7 +44,35 @@ jobs:
...
+          make install-packages opts="--with saml,auth-controller,workflows"
+          rm -rf ${HOME}/.git-credentials
+
+      - name: Install SAML Dependencies
+        run: sudo apt-get install -y xmlsec1
+
+      - name: Get workflows-logic package version
+        id: workflow-version
+        run: |
+          tag=$(grep -A 1 "\[tool.poetry.group.workflows.dependencies\]" pyproject.toml | awk -F '"' '{printf $4}')
```
khvn26: I'd rather use something like https://github.com/benc-uk/workflow-dispatch to trigger the workflow tests in their own repo. Actually, this might be a better solution for all the private modules overall

gagantrivedi: Because each package runs only its own tests, this will not help us cover the use case where updating package X breaks package Y.

.github/workflows/api-tests-with-private-packages.yml
```diff
@@ -11,7 +11,9 @@ defaults:
     working-directory: api
 
 env:
+  FLAGSMITH_SAML_REVISION: v1.6.1
   FLAGSMITH_RBAC_REVISION: v0.7.0
+  FLAGSMITH_AUTH_CONTROLLER_REVISION: v0.1.0
```
khvn26: Auth-controller version now has to be inferred from pyproject now that #4174 is merged.

khvn26 (changes requested):
One comment, otherwise LGTM.

There's a number of improvements to consider that fall out of scope of this PR:
- The private cloud build doesn't live in the workflow anymore so it can and should be used for tests too. Given there's a Docker-based build for private-cloud, we could explore something among the lines of how flagsmith-workflows executes its tests on top of an API Docker image.
- I tend to think that these tests, in fact, belong to the `flagsmith/flagsmith` repo simply based on the amount of Core API dependencies they currently have. It feels logical to run the full Core test suite, including the integration tests, for each PR to the upstream repos, instead of gathering each upstream test suite here. That might change if/when we assemble those dependencies to 1-2 larger packages, though.

.github/workflows/api-tests-with-private-packages.yml
gagantrivedi: This is done

.github/workflows/api-tests-with-private-packages.yml
```diff
@@ -11,6 +11,7 @@ defaults:
     working-directory: api
 
 env:
+  FLAGSMITH_SAML_REVISION: v1.6.1
```
khvn26: Already outdated (see [here](https://github.com/Flagsmith/flagsmith/blob/0354c264103409f5cb4c06db42b591fe236e42ea/api/Makefile#L14-L15))

.github/workflows/api-tests-with-private-packages.yml
```diff
@@ -42,7 +43,42 @@ jobs:
...
+          make install-packages opts="--with saml,auth-controller,workflows"
+          rm -rf ${HOME}/.git-credentials
+
+      - name: Install SAML Dependencies
+        run: sudo apt-get install -y xmlsec1
+
+      - name: Get workflows-logic package version
+        id: workflow-version
+        run: |
+          tag=$(make get-workflow-version)
```
khvn26: I think it's better to put all of this workflow, not just version retrieval, into a Makefile target so we're able to run `make test-private` locally. This way we won't have to specify the non-pip dependency versions twice, too.

khvn26 (approved):

khvn26 (changes requested):
Erroneously approved, still a couple outstanding comments to address.

.github/workflows/api-tests-with-private-packages.yml
gagantrivedi: Yep, good shout!  this is done

khvn26 (approved):

PR #4144, opened 2024-06-11
feat: Grafana integration

dabeeeenster:
Adds a Grafana integration, similar to New Relic and Datadog.

Relevant API is https://grafana.com/docs/grafana/latest/developers/http_api/annotations/#create-annotation


khvn26:
- Moved the integration to project level.
- Added tags to annotations, including boolean flag state and feature tags from Flagsmith.
- Added a `get_audited_instance_from_audit_log_record` service function to extract audited objects from an audit log in a generic way.
- Improved code coverage.
- Improved documentation.

api/integrations/grafana/mappers.py
```diff
@@ -0,0 +1,72 @@
...
+    FeatureStateValue,
+)
+from integrations.grafana.types import GrafanaAnnotation
+from segments.models import Segment
+
+
+def _get_feature_tags(
+    feature: Feature,
+) -> list[str]:
+    return list(feature.tags.values_list("label", flat=True))
```
matthewelwell, line 16: Is there a reason to use `"label"` here instead of `"tag"` - shouldn't we be consistent with the terminology that we use in the Flagsmith platform?

api/util/util.py
```diff
@@ -11,6 +11,7 @@ def decorator(*args, **kwargs):
         t = Thread(target=function, args=args, kwargs=kwargs)
         t.daemon = True
         t.start()
+        return t
```
matthewelwell: Maybe I missed it, but I'm not sure where/if this is used? If it is used, we should add it to the typing for the function.

matthewelwell (commented):
This looks good on the whole. Added a couple of minor comments.

api/integrations/grafana/mappers.py
khvn26: It's implied that the function result is a list of Grafana tags.

matthewelwell: Right, so we don't want to use `"tag"` again because it would be a tag within a tag? 🤔

api/util/util.py
khvn26: Not used; removed this.

api/integrations/grafana/mappers.py
khvn26: We only need tag labels, so we fetch those.

https://github.com/Flagsmith/flagsmith/blob/8582a661fb5a43ab230deeca796047c92bc27314/api/projects/tags/models.py#L13

matthewelwell: Sorry, brain not switched on yet today. Read this totally wrong! 

So, I guess my follow up question would be whether we need to namespace these at all like we are with the other Grafana tags?

khvn26: Good question. I think, as a user, I'd be annoyed by having to prefix those.

matthewelwell: Fair enough 👍

matthewelwell (approved):

PR #4151, opened 2024-06-12
deps: remove locally cached js deps

matthewelwell:

## Changes

Removes locally cached (and very outdated!) JS deps from the `frontend/web/static` dir. Also removes entirely the `msal` dependency as I couldn't see it being used anywhere. 

This PR is really a half-fix and the proper solution would be to use npm to correctly add the dependencies when bundling the FE assets to serve via the API. 

Ran `npm run bundledjango` locally and confirmed that the FE assets were correctly served.


frontend/web/index.handlebars
```diff
@@ -9,11 +9,10 @@
     <meta property="og:image" content="https://flagsmith.com/static/aa08b77a1a9bca99643945f1f3eeb027/b2d24/social-preview.jpg" />
     <meta property="og:title" content="Flagsmith - Feature Flags, Feature Toggles and Remote Config" />
     <meta name="robots" content="noindex, nofollow"/>
-    <script src="/static/javascript/highlight.min.js"></script>
-    <script src="/static/javascript/msal.min.js"></script>
```
matthewelwell, line 13: As mentioned in the description, I'm not sure if we should / can be removing the msal dependency here, but I can't see any reason that it exists. Maybe for LDAP authentication I guess? 

@kyle-ssg probably knows.

matthewelwell:
I can't approve this as it's my own PR, but I've tested it and all looks good to me 👍

kyle-ssg (approved):

Issue #3718, opened 2024-04-04
Minimise Docker builds in CI


khvn26:
The following Docker builds are unneccessary:

For E2E:
- https://github.com/Flagsmith/flagsmith/blob/1e75ae9b573063684c5935fd01ac1945bc869a93/.github/workflows/platform-pull-request.yml#L55-L56
- https://github.com/Flagsmith/flagsmith/blob/1e75ae9b573063684c5935fd01ac1945bc869a93/.github/workflows/platform-pull-request.yml#L84-L85
- https://github.com/Flagsmith/flagsmith/blob/1e75ae9b573063684c5935fd01ac1945bc869a93/.github/workflows/platform-pull-request.yml#L113-L114
- https://github.com/Flagsmith/flagsmith/blob/1e75ae9b573063684c5935fd01ac1945bc869a93/.github/workflows/platform-pull-request.yml#L142-L143
- https://github.com/Flagsmith/flagsmith/blob/1e75ae9b573063684c5935fd01ac1945bc869a93/.github/workflows/platform-pull-request.yml#L270-L271
- https://github.com/Flagsmith/flagsmith/blob/1e75ae9b573063684c5935fd01ac1945bc869a93/.github/workflows/platform-pull-request.yml#L293-L294
- https://github.com/Flagsmith/flagsmith/blob/1e75ae9b573063684c5935fd01ac1945bc869a93/.github/workflows/platform-pull-request.yml#L317-L318
- https://github.com/Flagsmith/flagsmith/blob/1e75ae9b573063684c5935fd01ac1945bc869a93/.github/workflows/platform-test-merge-to-main.yml#L32-L33

For Uffizzi:
- https://github.com/Flagsmith/flagsmith/blob/1e75ae9b573063684c5935fd01ac1945bc869a93/.github/workflows/uffizzi-build.yml#L40-L41

All of the above are building the same couple of images. We should build them once per CI run, push to [ephemeral registry](https://depot.dev/blog/depot-ephemeral-registry), and pass the reference around.

matthewelwell:
I think we should tie into this work, the addition of e.g. Docker Scout into our CI pipeline. 

https://github.com/docker/scout-action

khvn26:
https://github.com/quay/clair-action looks good too

khvn26:
trivy is pretty good — found vulns both from claire's (Quay) and Docker Scout outputs

matthewelwell:
[...]

PR #4165, opened 2024-06-14
ci(build): Reusable private cloud build

khvn26:

## Changes

Contributes to #3718.

This pull request improves our build toolchain to clarify our shipping targets and make building them more accessible. It integrates Private Cloud builds into our platform PR workflow, allowing us to run the E2E test suite against it. It streamlines our deployment workflows, allowing deploy jobs to benefit from Depot caching and cleanly segregates deployment configurations between GitHub environments. It sets up Trivy Docker image scan result uploads to the repo's [Security tab](https://github.com/Flagsmith/flagsmith/security), enabling us to address detected vulnerabilities ahead of releases. This PR is large, so I'll break down the changes below.

### Single Dockerfile for all targets

I consolidated multiple Dockerfiles into a single Dockerfile containing all build stages. We ship multiple Docker targets:

- Private cloud: Unified
- SaaS: API
- Open Source: API, Frontend, Unified

Previously, Private Cloud and SaaS were awkwardly distributed between API and Unified Dockerfiles and corresponding GHA workflows. The private backend code was manually pulled by the workflow, and a special build argument controlled whether the dependencies for it should be installed. SaaS deployments required specifying private backend code versions for each environment. Changes to the private side required touching two Dockerfiles and 2-3 workflows and hoping the build worked since it was only reproducible in GHA.

Now, every target, including Private Cloud and SaaS, can be built using `docker build`, provided the required secrets are mounted. The final stages are minimal, assembled from building blocks factored out of our former Dockerfiles and GHA workflows. All stages are documented for easy navigation.

The `CI_COMMIT_SHA`, `.versions.json`, and edition markers are now explicitly requested by the Docker build instead of being added by GHA workflows. Permanent caching for the backend's `get_version_info` utility has been added. 

To demonstrate improved build sanity, I've added a Private cloud API target. The stage's code is just 5 lines. I've also created `private-cloud-api` private repositories in our GitHub Container registry and Docker Hub organization.

The `docker build -t flagsmith:local .` command remains fully backward compatible due to the `oss-unified` stage being the last one.

### Private cloud images built for PRs and on merge to `main`

The "Platform Pull Requests" workflow now builds a Unified Private cloud image, enabling faster testing of enterprise features. E2E tests are now run against both Open Source and Private cloud API image versions. We can also deliver improvements and bug fixes to enterprise clients ahead of releases by offering a PR or a staging image.

### ECS deployments with less fat

Deployment configuration data values have been moved to their respective GHA environments, allowing a reusable ECS deployment workflow for both staging and production. The workflow uses Depot and the new Dockerfile to leverage cached build stages. The change also results in less infrastructure-specific data in our workflow code.

### Platform workflow publishes all Docker images

All Docker image building and publication have been consolidated in `.github/workflows/platform-docker-build-test-publish.yml`, triggered on merges to `main` and releases. Images are built for every merge and release, and published on release only if an E2E run is successful. After successful image publication, a PR to `flagsmith-charts` is created. Individual Docker image publication workflows have been removed.

### CodeQL reports

CodeQL has been enabled for the repository. In addition to free code scanning, Docker image scan results are now available as part of CodeQL reports. Alerts/CI failures will occur on newly detected critical issues.

### Bits and Pieces

- Added the `install-private-modules` Makefile target for use outside of Docker builds, i.e., by the "Run API tests with private package" workflow.
- Removed `.ubi` Dockerfiles as they are no longer needed.
- Removed E2E image publication to Docker Hub; those needing it can pull from `ghcr.io/flagsmith/flagsmith-e2e`.

### To do / to think on

- Consider making the SSE PGP key a runtime dependency to build the SaaS API target once per deployment, not per environment.
- Avoid invoking `docker build` on release trigger; re-tag the main branch images instead. To do that, ensure release jobs run after the latest main branch builds complete, e.g. using a concurrency group.
- Reevaluate Uffizzi's reusable workflow to remove boilerplate.
- Consider adding the E2E image to the mega-Dockerfile targets.
- Avoid duplicating the Private cloud build for the "Run API tests with private package" workflow.

Docker builds were tested locally with `docker build`.

We need a plan for testing the ECS deployment workflow.


.github/actions/api-deploy-ecs/action.yml
```diff
@@ -169,7 +71,7 @@ runs:
       with:
         task-definition: ${{ inputs.aws_task_definitions_directory_path }}/ecs-task-definition-migration.json
         container-name: flagsmith-api-migration
-        image: ${{ steps.build-image.outputs.image }}
+        image: ${{ inputs.api_ecr_image_url }}
```
matthewelwell, line 74: I'm sure this will become apparent later on, but at this stage, I'm not sure how the image gets to ECR.

.github/workflows/.reusable-docker-build.yml
```diff
@@ -6,30 +6,39 @@ on:
...
-        description: Whether to scan image for vulnerabilities
+        description: Whether to scan built image for vulnerabilities
         required: false
         default: true
     outputs:
       image:
         description: Resulting image specifier
         value: ${{ inputs.registry-url }}/flagsmith/${{ inputs.image-name }}:${{ jobs.build.outputs.version }}
+    secrets:
+      secrets:
```
matthewelwell, line 39: Perhaps this is some GH workflows oddness but it feels a little strange to define the secrets key twice here?

.github/workflows/.reusable-docker-build.yml
```diff
@@ -67,18 +76,38 @@ jobs:
...
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
 
       - name: Build and push image
         uses: depot/build-push-action@v1
         with:
+          context: .
           push: true
-          build-args: ${{ inputs.build-args }}
+          platforms: linux/amd64,linux/arm64
```
matthewelwell, line 87: Hmm... do we always want to build for both platforms? Maybe we do.

.github/workflows/.reusable-docker-build.yml
```diff
@@ -67,18 +76,38 @@ jobs:
...
           image-ref: ${{ inputs.registry-url }}/flagsmith/${{ inputs.image-name }}:${{ steps.meta.outputs.version }}
+          format: sarif
+          output: trivy-results.sarif
+        env:
+          TRIVY_USERNAME: ${{ github.actor }}
+          TRIVY_PASSWORD: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Upload Trivy scan results to GitHub Security tab
+        uses: github/codeql-action/upload-sarif@v2
+        if: inputs.scan && (success() || failure())
```
matthewelwell, line 112: what other options are there other than `success()` and `failure()` here? `error()` maybe I guess?

.github/workflows/.reusable-docker-publish.yml
```diff
@@ -0,0 +1,52 @@
...
+          images: ${{ inputs.target-images }}
+          tags: |
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+
+      - name: Publish Image
+        uses: contiamo/retag-push@main
+        with:
+          source: ${{ inputs.source-images }}
+          target: ${{ steps.meta.outputs.tags }}
```
matthewelwell: I'm a little confused as to how this works with multiple source images and multiple targets? Or is the use case only for one of the 2 lists to include more than one element?

.github/workflows/platform-docker-build-test-publish.yml
```diff
@@ -0,0 +1,161 @@
...
+
+  docker-publish-private-cloud:
+    needs: [docker-build-private-cloud, run-e2e-tests]
+    uses: ./.github/workflows/.reusable-docker-publish.yml
+    if: github.event_name == 'release'
+    with:
+      source-images: ${{ needs.docker-build-private-cloud.outputs.image }}
+      target-images: flagsmith/flagsmith-private-cloud
+
+  update-charts:
```
matthewelwell, line 114: If possible, I'd prefer to see this in a separate workflow, but I can't remember if you can pass variables / add dependencies between workflows?

.github/workflows/platform-pull-request.yml
```diff
@@ -60,23 +60,23 @@ jobs:
     name: Build Unified Image
     uses: ./.github/workflows/.reusable-docker-build.yml
     with:
-      file: Dockerfile
+      target: oss-unified
       image-name: flagsmith
```
matthewelwell, line 64: Shouldn't these be tagged with the PR somewhere so they don't pollute the main release images? Maybe that's happening somewhere else.

Dockerfile
```diff
@@ -14,64 +63,147 @@ COPY frontend/env/ ./frontend/env/
...
-RUN if [ "${SAML_INSTALLED}" = "1" ]; then apt-get update && apt-get install -y xmlsec1; fi;
+# Install private Python dependencies
+RUN make install-packages opts='--without dev --with saml,auth-controller,ldap,workflows'
 
-# arm architecture platform builds need postgres drivers installing via apt
-ARG TARGETARCH
-RUN if [ "${TARGETARCH}" != "amd64" ]; then apt-get update && apt-get install -y libpq-dev && rm -rf /var/lib/apt/lists/*; fi;
+# Integrate private modules
+ARG SAML_REVISION
+ARG AUTH_CONTROLLER_REVISION
```
matthewelwell: Sorry for the collision but I think that (based on #4174) this is actually no longer needed.

api/Makefile
```diff
@@ -11,6 +11,9 @@ POETRY_VERSION ?= 1.8.2
 
 GUNICORN_LOGGER_CLASS ?= util.logging.GunicornJsonCapableLogger
 
+SAML_REVISION ?= v1.6.0
+RBAC_REVISION ?= v0.7.0
```
matthewelwell: What about workflows, ldap, auth-controller?

matthewelwell (commented):
This looks great overall. 

I've added a few specific comments, but nothing too worrying. 

One general question I had though is around the Dockerfile and whether we're _always_ building all stages and then using the relevant builds to output the target image, or if docker is smart enough to know, for example, to not build the stages it doesn't need.

.github/actions/api-deploy-ecs/action.yml
khvn26: See [here](https://github.com/Flagsmith/flagsmith/blob/ec01d411b68a23aa61f60f33275dad6d95a93632/.github/workflows/.reusable-deploy-ecs.yml#L46-L68), and then [here](https://github.com/Flagsmith/flagsmith/blob/ec01d411b68a23aa61f60f33275dad6d95a93632/.github/workflows/.reusable-deploy-ecs.yml#L93) and [here](https://github.com/Flagsmith/flagsmith/blob/ec01d411b68a23aa61f60f33275dad6d95a93632/.github/workflows/.reusable-deploy-ecs.yml#L103).

.github/workflows/.reusable-docker-build.yml
khvn26: This is a coincidence. Bottom-level `secrets` key exposes the docker build and push action `secrets` argument. Since we're passing Github secrets as values to this argument, we have to use the top-level workflow `secrets` key.

.github/workflows/.reusable-docker-build.yml
khvn26: We ship both platforms and this workflow is intended for building our shippable targets.

.github/workflows/.reusable-docker-build.yml
khvn26: This condition prompts the results upload step to happend even if `trivy-action` fails (but not if `scan` argument is `false`). `trivy-action` might fail if the number of found critical/high vulnerabilities crosses a certain threshold.

.github/workflows/platform-pull-request.yml
khvn26: See [here](https://github.com/Flagsmith/flagsmith/blob/ec01d411b68a23aa61f60f33275dad6d95a93632/.github/workflows/.reusable-docker-build.yml#L69-L80). The tags generated are aware of the calling GHA event: pr-prefixed for PR events, version tag and `latest` for releases, `main` for the push to main event.

Dockerfile
khvn26: Removed.

api/Makefile
khvn26: Their versions are specified in pyproject.toml.

The default private module versions are here just because we needed a source of truth other than GHA workflows.

.github/workflows/platform-docker-build-test-publish.yml
khvn26: You can, but there's so much boilerplate required to support it I'm hesitant to do it.

I prefer to have all interdependent jobs in one workflow anyway.

.github/workflows/.reusable-docker-publish.yml
khvn26: That's awkward naming from the action author. `target` is, in fact, a list of tags too.

khvn26:
> One general question I had though is around the Dockerfile and whether we're always building all stages and then using the relevant builds to output the target image, or if docker is smart enough to know, for example, to not build the stages it doesn't need.

@matthewelwell Buildkit (which we use) does skip unneeded stages: https://docs.docker.com/build/building/multi-stage/#differences-between-legacy-builder-and-buildkit

.github/actions/api-deploy-ecs/action.yml
matthewelwell: Sorry, I should have deleted this one as it was fairly obvious from the rest of the PR.

.github/workflows/.reusable-docker-build.yml
matthewelwell: We don't currently do that for _all_ images though I didn't think. But I think we probably should.

.github/workflows/.reusable-docker-build.yml
matthewelwell: My question though was ... what other statuses are there? `success` and `failure` seem to cover all the bases so isn't this statement just the same as `if: inputs.scan` ?

.github/workflows/platform-docker-build-test-publish.yml
matthewelwell: Yep, sure 👍

.github/workflows/.reusable-docker-build.yml
khvn26: No, the default behaviour is equivalent to `if: inputs.scan && success()`.

Overall, besides `success()` and `failure()` there's also `skipped()`, which we don't want to ignore.

matthewelwell: Gotcha 👍

.github/workflows/platform-pull-request.yml
matthewelwell: Sorry, I completely trust that what you're saying is correct I'm just trying to understand how it works a little bit (but this shouldn't hold up the PR review so if everything else is resolved, I will approve). I can see (via the link you shared) that we define the following tags: 

```
            type=ref,event=branch
            type=ref,event=tag
            type=ref,event=pr
```

... but what I'm confused about I guess is why we define all 3 there. What determines which of the three it is, and what determines the actual tag name?

api/Makefile
matthewelwell: So we no longer have e.g. `inputs.flagsmith_rbac_revision` to manage? If so, 👌

khvn26: Exactly 👌

matthewelwell (commented):
This looks good now for sure but we should carefully think about how / when we want to release this as it's not unlikely we might need to add some fixes to the build processes for the deployment.

matthewelwell (approved):
Sorry, meant to approve with the last comment. Struggling with train wifi...

.github/workflows/platform-pull-request.yml
khvn26: The tags input is detailed [in the action docs](https://github.com/docker/metadata-action/tree/v5/?tab=readme-ov-file#tags-input).

According to the rules we define, the built images are tagged:
- On PR events: `pr-` prefixed.
- On merge to main: `main`.
- On tag: tag name, semver full version, semver major.minor version, `latest`.

When published to Docker Hub, the images are re-tagged:
- On tag: tag name, semver full version, semver major.minor version, `latest`.

PR #4173, opened 2024-06-17
test: seed data for permission and roles e2e tests

### LINKED ISSUE #3854: Add e2e tests for permissions and roles

Recently, https://github.com/Flagsmith/flagsmith/pull/3853 caused a regression whereby editing permissions stopped working. This should have been caught in e2e.

Add E2E that covers
- Creating a role
- Creating a group
- Testing variations of permissions

N/A

### Additional context

_No response_

novakzaballa:

## Changes

- Seed data for permission and roles e2e tests

- Unit tests added


api/e2etests/e2e_seed_data.py
```diff
@@ -35,7 +65,49 @@ def seed_data() -> None:
...
+    )
+    non_admin_user_with_env_permissions: FFAdminUser = FFAdminUser.objects.create_user(
+        email=settings.E2E_NON_ADMIN_USER_WITH_ENV_PERMISSIONS,
+        password=PASSWORD,
+        username=settings.E2E_NON_ADMIN_USER_WITH_ENV_PERMISSIONS,
+    )
+    non_admin_user_with_a_role: FFAdminUser = FFAdminUser.objects.create_user(
+        email=settings.E2E_NON_ADMIN_USER_WITH_A_ROLE,
+        password=PASSWORD,
+        username=settings.E2E_NON_ADMIN_USER_WITH_A_ROLE,
```
matthewelwell: We shouldn't need to set the `username` here I don't think?

api/e2etests/e2e_seed_data.py
```diff
@@ -58,10 +140,47 @@ def seed_data() -> None:
         project = Project.objects.create(
             name=project_info["name"], organisation=organisation
         )
+        if project_info["name"] == "My Test Project 5 Project Permission":
```
matthewelwell: Couldn't we define this as an extra attribute in the `project_test_data` dictionary instead of using the project name? 

If we have to use the project name, it should be defined as a constant.

api/organisations/models.py
```diff
@@ -363,6 +365,11 @@ def get_subscription_metadata(self) -> BaseSubscriptionMetadata:
             # plan.
             return FREE_PLAN_SUBSCRIPTION_METADATA
 
+        if self.plan == SCALE_UP and self.subscription_id == "test_subscription_id":
```
matthewelwell: I don't like having this test code here. Won't it, by default, return the data that we add to the subscription in the seed data anyway?

matthewelwell (changes requested):

api/e2etests/e2e_seed_data.py
novakzaballa: Username removed

api/e2etests/e2e_seed_data.py
novakzaballa: New constants created

api/organisations/models.py
novakzaballa: With regards to obtaining subscription data, this is the process followed by the e2e test:

https://github.com/Flagsmith/flagsmith/blob/7456fa320b9001613cf50145a53b0a957d3c9197/api/organisations/models.py#L408-L416

We need to support multiple projects for the test that creates a new project based on permissions.

api/e2etests/e2e_seed_data.py
```diff
@@ -35,7 +68,45 @@ def seed_data() -> None:
...
+    )
+    non_admin_user_with_env_permissions: FFAdminUser = FFAdminUser.objects.create_user(
+        email=settings.E2E_NON_ADMIN_USER_WITH_ENV_PERMISSIONS,
+        password=PASSWORD,
+    )
+    non_admin_user_with_a_role: FFAdminUser = FFAdminUser.objects.create_user(
+        email=settings.E2E_NON_ADMIN_USER_WITH_A_ROLE,
+        password=PASSWORD,
+    )
     org_admin.add_organisation(organisation, OrganisationRole.ADMIN)
```
zachaysan: nit: I would move this line of code to line 71 so it's more obvious that an admin was attached to the organisation before all the non-admin user definitions.

api/organisations/models.py
zachaysan: If I understand him correctly, I agree with Matt. Putting in test code in the Subscription model is a total code smell. Can you not mock `get_subscription_metadata` to return what you need in the test?

Edit: This wont work since the test *calls* `get_subscription_metadata` for the primary part of the test.

api/tests/unit/organisations/test_unit_organisations_models.py
```diff
@@ -339,6 +341,23 @@ def test_organisation_subscription_get_subscription_metadata_returns_free_plan_m
...
+) -> None:
+
+    # Given
+    Subscription.objects.filter(organisation=organisation).update(
+        max_seats=8, plan=SCALE_UP, subscription_id="test_subscription_id"
+    )
+    organisation.refresh_from_db()
+
+    # When
+    subscription_metadata = organisation.subscription.get_subscription_metadata()
```
zachaysan: Oh I see the issue now. Mocking `get_subscription_metadata` would be pointless since it's the point of the test. What's the purpose of this test if we're writing custom code just to return to it?

zachaysan (commented):
Left some comments. Overall things look good, but I'm confused about the purpose of the test I commented about.

api/tests/unit/organisations/test_unit_organisations_models.py
novakzaballa: Actually, this code is more for the e2e test, which creates a new project only if permission is granted. In the previous flow, the number of projects was `None`, causing an issue when trying to create a new one from the UI. With this code, I set the maximum number of projects to 7, ensuring there won't be any problem when creating the new project.

api/organisations/models.py
novakzaballa: That code is not for testing the API, it's for testing the E2E tests in the FE that call the get-subscription-metadata endpoint. Among the response values, it includes the number of projects we are allowed to have. Without that code, the response returns projects: 1, which prevents the frontend from creating a new project to test the CREATE_PROJECT permission.

zachaysan: I still don't like test code leaking into the actual application, but at least this makes some sense now. I'll leave it up to you to think of a better solution if you have one but otherwise I'll finish the review.

zachaysan (approved):

api/organisations/models.py
matthewelwell: Another option would be to make sure that we're testing against a container that returns `is_enterprise() == True` ? I think that would just be a case of adding a volume mapped to `/app/ENTERPRISE_VERSION` ? 

Otherwise, we definitely need to think of a better way to achieve this. The approach here is far too hacky - relying on a specific string. Can't we use the E2E_TEST_TOKEN somehow for example?

zachaysan (dismissed):

api/projects/permissions.py
```diff
@@ -31,11 +32,16 @@
...
+class ProjectPermissions(E2ETestPermissionMixin, IsAuthenticated):
     def has_permission(self, request, view):
         """Check if user has permission to list / create project"""
         if not super().has_permission(request, view):
             return False
+        elif (
+            super().has_permission(request, view)
+            and getattr(request, "is_e2e", False) is True
+        ):
+            return True
```
matthewelwell: @novakzaballa even if this was necessary, the call to `super()` is quite clearly unnecessary as we're early returning on the line above if it's not true. 

In general though, this is exactly the job that the `E2ETestPermissionMixin` is trying to achieve. Can you have another look at this please?

matthewelwell (changes requested):

api/projects/permissions.py
novakzaballa: You’re right, the call to super() was no longer necessary and has been corrected.

There was an issue with this part of the [code](https://github.com/Flagsmith/flagsmith/blob/6ef7a742f0f56aef2335da380770dc7f307d53c5/api/projects/permissions.py#L56) that was validating the number of projects and returning `False`, I moved part of the logic exclusively to this validation.

matthewelwell: In this case, I don't think the mixin is actually doing anything? I would rather see the mixin used, or removed if we can't get it working.

novakzaballa: I removed it since it doesn't do what we expect.

matthewelwell (approved):

PR #4184, opened 2024-06-18
feat: Add UI for SAML attribute mapping

### LINKED ISSUE #4167: Add UI for configuring SAML attribute mappings
As per the new endpoints added as part of [this issue](https://github.com/Flagsmith/flagsmith-saml/issues/37) in the SAML repository, we need to add a UI for configuring these attribute mappings.

novakzaballa:

## Changes

- Add a new tab for the SAML attribute mapping
- Improve SAML modal

1.  Go to Organisation settings -> SAML Tab
2. Create or select a SAML configuration
3. Go to the Attribute Mapping tab and create a new Attribute


matthewelwell:
@novakzaballa I've released the SAML fix that you created to staging so this can now be tested correctly against staging, but I've found a few issues: 

 1. It should be possible to add attributes as part of the creation workflow (or at least the modal shouldn't auto close)
 2. The FE is making an invalid request initially when opening the edit modal to retrieve the mappings for `saml_configuration=0` (see screenshot 1 below)
 3. The UI doesn't handle long IdP attributes (see screenshot 2 below - note that this is a very valid attribute and is used in a lot of our customer's configurations)
 4. The attributes aren't editable but appear like they are (as it shows a pointer when you hover over each row in the table). I guess making them editable would be tricky given the fact that we're already in a modal, so perhaps we should just remove the pointer on hover? 

image

image

novakzaballa:
> 1. It should be possible to add attributes as part of the creation workflow (or at least the modal shouldn't auto close)

Now the modal does not close when the configuration is created, and the SAML attribute tab displays immediately.

> 2. The FE is making an invalid request initially when opening the edit modal to retrieve the mappings for `saml_configuration=0` (see screenshot 1 below)

Corrected

> 3. The UI doesn't handle long IdP attributes (see screenshot 2 below - note that this is a very valid attribute and is used in a lot of our customer's configurations)

Done
Screenshot 2024-06-20 at 3 57 44 PM


> 4. The attributes aren't editable but appear like they are (as it shows a pointer when you hover over each row in the table). I guess making them editable would be tricky given the fact that we're already in a modal, so perhaps we should just remove the pointer on hover?

Removed.

matthewelwell:
The overflow issue still isn't resolved. The key thing we need to solve it for is not multiple words, but a single long URL. The test case you used is not really valid. 

image

matthewelwell:
I've fixed this [here](https://github.com/Flagsmith/flagsmith/pull/4184/commits/3c5cf0865b8054d535aa733fcd6bff676d06ee64) but I'm seeing weird behaviour on the tooltip. It looks like the container for the tooltip itself isn't wide enough for the string but I have no idea how to resolve that... 

image

novakzaballa:
> I've fixed this but I'm seeing weird behaviour on the tooltip. It looks like the container for the tooltip itself isn't wide enough for the string but I have no idea how to resolve that...
> 
Screenshot 2024-06-24 at 10 34 43 AM



Done

matthewelwell (approved):

PR #4189, opened 2024-06-19
refactor: replace task processor

matthewelwell:

## Changes

Replace the Task Processor code (which is currently split between this repo and https://github.com/Flagsmith/flagsmith-task-processor) with a direct poetry dependency instead. 

All the unit tests have been moved to the task-processor repository. There are tests through this repo that still test the integration with the task processor.


matthewelwell:
Converting to draft to: 

 1. ~Fix isort issues~
 2. ~Wait until I can bump to 1.0.0 (which requires an update in flagsmith-workflows)~

api/pyproject.toml
```diff
@@ -37,7 +37,7 @@ use_parentheses = true
 multi_line_output = 3
 include_trailing_comma = true
 line_length = 79
-known_first_party=['analytics','app','custom_auth','environments','integrations','organisations','projects','segments','tests','users','webhooks','api','audit','e2etests','features','permissions','util']
+known_first_party=['analytics','app','custom_auth','environments','integrations','organisations','projects','segments','tests','users','webhooks','api','audit','e2etests','features','permissions','task_processor','util']
```
khvn26: Just for the lean diff?

matthewelwell: Well, partly, but I also thought it logically made sense. It's still 'first party' in my mind. But I did switch back and forth on this so I could definitely be persuaded that it's 'third party'.

khvn26: Why isn't workflows_logic first party then?

matthewelwell: Good question. IMO it should be. At the moment it's not defined explicitly as either so isort is defaulting to third party, but I'd argue that's incorrect.

khvn26: I don't agree. 1st vs 3rd party import means package local to the codebase vs external dependency. IMO shouldn't add any more meaning to import order, it will only confuse people.

matthewelwell: Fair enough. I'll revert the latest commit then.

matthewelwell: Sorry, I thought I'd done this. I have now...

khvn26 (approved):

matthewelwell:
I have verified that there haven't been any more changes in the `task_processor` module in this repo. Once CI passes, I will merge.

PR #4193, opened 2024-06-19
docs: fix Edge Proxy docker compose ports definition

matthewelwell:
## Changes

Fixes the ports definition in the example docker compose for hosting the edge proxy. Note that I've also removed the deprecated `version` attribute. 

Ran the docker compose locally and confirmed it runs as expected and exposes the correct port.

### DIFF
--- a/docs/docs/deployment/hosting/locally-edge-proxy.md
+++ b/docs/docs/deployment/hosting/locally-edge-proxy.md
@@ -204,8 +204,7 @@ services:
      source: ./config.json
      target: /app/config.json
   ports:
-   - target: 8000
-     published: 8000
+    - "8000:8000"
 ```
 
 The Proxy is now running and available on port 8000.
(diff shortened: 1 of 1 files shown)


docs/docs/deployment/hosting/locally-edge-proxy.md
```diff
@@ -206,7 +205,7 @@ services:
      target: /app/config.json
   ports:
    - target: 8000
```
khvn26:
```suggestion
   - "8000:8000"
```

docs/docs/deployment/hosting/locally-edge-proxy.md
```diff
@@ -206,7 +205,7 @@ services:
...
-   - published: 8000
+     published: 8000
```
khvn26:
```suggestion
```

khvn26 (approved):
First time I've seen anyone use the long format to define ports. Would prefer a more traditional format but fine either way, I guess.

docs/docs/deployment/hosting/locally-edge-proxy.md
matthewelwell: Yeah, I tend to agree but (for some reason) I stuck as close to the previous definition as possible. I'll just update to the shorter form.

PR #4207, opened 2024-06-21
fix: oauth user case sensitivity

### LINKED ISSUE #4185: Duplicate user issue when using social auth
Sentry Issue: [FLAGSMITH-API-4Q3](https://flagsmith.sentry.io/issues/5504508599/?referrer=github_integration)

```
FFAdminUser.MultipleObjectsReturned: get() returned more than one FFAdminUser -- it returned 2!
(13 additional frame(s) were not displayed)
...
  File "custom_auth/views.py", line 42, in post
  File "users/models.py", line 93, in get_by_natural_key
```

It seems as though it is possible to create duplicate (when evaluated case insensitively) email addresses in the database by authenticating first with email password. 

Based on example, it seems that the steps to reproduce this issue are: 

 1. Register (using email and password) with an email address using all lower case e.g. billy.the.kid@gmail.com
 2. Register (using Google) with the same email address but with certain characters capitalised, e.g. Billy.The.Kid@gmail.com

matthewelwell:
## Changes

Resolves https://github.com/Flagsmith/flagsmith/issues/4185. 

The main goal of this PR is to ensure that users who authenticate with an OAuth provider after authenticating with email/password, another OAuth provider, or even SAML (although this use case is slim) get a consistent experience. Currently if the email differs by case, a new user will be created and the user will likely be confused as to why they can't see their projects, etc. 

As a side effect, we had to handle a few edge cases where a user might already have duplicate accounts and make sure that the user gets the same experience they do today (authenticating with method A gets them into user A, authenticating with user B gets them user B). 

We should, separate to this, get in touch with the users for whom this is an issue (~20) and try to merge their accounts. 

Added 3 new tests to cover all scenarios.

### DIFF
--- a/api/custom_auth/oauth/serializers.py
+++ b/api/custom_auth/oauth/serializers.py
@@ -1,3 +1,5 @@
+from abc import abstractmethod
+
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.contrib.auth.signals import user_logged_in
@@ -107,11 +109,12 @@ class OAuthLoginSerializer(serializers.Serializer):
 
         return existing_user
 
+    @abstractmethod
     def get_user_info(self):
         raise NotImplementedError("`get_user_info()` must be implemented.")
 
     def get_auth_type(self) -> AuthType:
-        if not self.auth_type:
+        if not self.auth_type:  # pragma: no cover
             raise NotImplementedError(
                 "`auth_type` must be set, or `get_auth_type()` must be implemented."
             )
--- a/api/custom_auth/oauth/serializers.py
+++ b/api/custom_auth/oauth/serializers.py
@@ -60,11 +60,23 @@ class OAuthLoginSerializer(serializers.Serializer):
     def _get_user(self, user_data: dict):
         email: str = user_data.pop("email")
 
+        # There are a number of scenarios that we're catering for in this
+        # query:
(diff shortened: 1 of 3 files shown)


api/custom_auth/oauth/serializers.py
```diff
@@ -65,20 +92,50 @@ def _get_user(self, user_data: dict):
...
 
+    def get_auth_type(self) -> AuthType:
+        if not self.auth_type:  # pragma: no cover
+            raise NotImplementedError(
+                "`auth_type` must be set, or `get_auth_type()` must be implemented."
+            )
+        return self.auth_type
+
+    def get_user_model_id_attribute(self) -> str:
+        return self.user_model_id_attribute
```
zachaysan: Not really sure why this method is necessary. Can't we just reference `self.user_model_id_attribute` directly?

zachaysan (approved):

api/custom_auth/oauth/serializers.py
matthewelwell: Yeah, true - I initially had it because it threw a NotImplementedError if the attribute was not set. But, to get the tests that test the base serializer class directly to pass, I had to set a default. 

I'll remove the method.

matthewelwell: Done

PR #4215, opened 2024-06-24
fix(ci): Secrets unavailable for deploy jobs

khvn26:

## Changes

Fixes deploy jobs. 

This is a CI change.

### DIFF
--- a/.github/workflows/.reusable-deploy-ecs.yml
+++ b/.github/workflows/.reusable-deploy-ecs.yml
@@ -85,7 +85,6 @@ jobs:
           aws_ecs_service_name: ${{ vars.AWS_ECS_SERVICE_NAME }}
           aws_vpc_subnet_id: ${{ vars.AWS_VPC_SUBNET_ID }}
           aws_ecs_security_group_id: ${{ vars.AWS_ECS_SECURITY_GROUP_ID }}
-          aws_ecr_repository_arn: ${{ vars.AWS_ECR_REPOSITORY_ARN }}
           aws_identity_migration_event_bus_name: ${{ vars.AWS_IDENTITY_MIGRATION_EVENT_BUS_NAME }}
           aws_identity_migration_event_bus_rule_id: ${{ vars.AWS_IDENTITY_MIGRATION_EVENT_BUS_RULE_ID }}
           aws_identity_migration_task_role_arn: ${{ vars.AWS_IDENTITY_MIGRATION_TASK_ROLE_ARN }}
--- a/.github/workflows/.reusable-deploy-ecs.yml
+++ b/.github/workflows/.reusable-deploy-ecs.yml
@@ -33,11 +33,12 @@ jobs:
         uses: depot/setup-action@v1
 
       - name: Configure AWS Credentials
-        uses: aws-actions/configure-aws-credentials@v1-node16
+        uses: aws-actions/configure-aws-credentials@v4
         with:
           aws-access-key-id: ${{ vars.AWS_ACCESS_KEY_ID }}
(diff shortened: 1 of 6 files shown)


.github/workflows/.reusable-deploy-ecs.yml
```diff
@@ -61,7 +62,7 @@ jobs:
           build-args: CI_COMMIT_SHA=${{ github.sha }}
           secrets: |
             github_private_cloud_token=${{ secrets.GH_PRIVATE_ACCESS_TOKEN }}
-            sse_pgp_pkey=${{ secrets.SSE_PGP_PRIVATE_KEY }}
+            "sse_pgp_pkey=${{ secrets.SSE_PGP_PRIVATE_KEY }}"
```
matthewelwell, line 65: Odd that it's only this one that needs quotations. Could / should we add a comment?

Dockerfile
```diff
@@ -167,8 +167,7 @@ FROM api-runtime-private as saas-api
 RUN --mount=type=secret,id=sse_pgp_pkey \
   apt-get update && apt-get install -y gnupg && \
   gpg --import /run/secrets/sse_pgp_pkey && \
-  mv /root/.gnupg /app/; \
-  chown -R nobody /app/.gnupg
+  mv /root/.gnupg/ /app/
```
matthewelwell: I'm keen to get @gagantrivedi's review here as well to make sure we know what we need to test.

khvn26: Sure. The image has been deployed to staging so we just need to verify SSE there.

gagantrivedi: Yeah, we have to make sure `sse` events are still reaching influx

.github/workflows/.reusable-deploy-ecs.yml
khvn26: The reason for the quotes is that the value is multiline.

khvn26: A comment would be tough because yaml `|` makes it impossible to comment the same line.

matthewelwell: Ok, yep, fair point. Hopefully this conversation will serve enough if people are searching for a reason in the future...

Dockerfile
```diff
@@ -167,8 +167,8 @@ FROM api-runtime-private as saas-api
 RUN --mount=type=secret,id=sse_pgp_pkey \
   apt-get update && apt-get install -y gnupg && \
   gpg --import /run/secrets/sse_pgp_pkey && \
-  mv /root/.gnupg /app/; \
-  chown -R nobody /app/.gnupg
+  mv /root/.gnupg/ /app/ && \
```
gagantrivedi, line 170: Are you able to build this image without any GPG key?

khvn26: > Are you able to build this image without any GPG key?

There's no need as `saas-api` target is only intended for building with the key.

gagantrivedi: > > Are you able to build this image without any GPG key?
> 
> There's no need as `saas-api` target is only intended for building with the key.

Hmm, I am little confused… Is this file not used for private cloud? What if we want to build an image locally to test something?

khvn26: This file is used for all of our targets now. To build the `private-cloud` target, you don't need the key; the GPG import is the only layer added by the `saas-api` target stage.

I'd love to make the key a runtime dependency and shave off one target, though! Let me know if you have an idea how to do that.

See Dockerfile comments and the [PR description](https://github.com/Flagsmith/flagsmith/pull/4165#issue-2353549270) for details.

matthewelwell (approved):

PR #4218, opened 2024-06-24
feat: Announcement feature flag per page

novakzaballa:

## Changes

- Now the UI can read a list of pages from the `announcement_per_page` feature flag and display the announcement only on those pages, this is the pages name list:

  login
  not-found
  signup
  home
  github-setup
  maintenance
  password-reset
  features
  change-requests
  scheduled-changes
  change-request
  scheduled-change
  widget
  invite
  invite-link
  broken
  oauth
  saml
  environment-settings
  sdk-keys
  integrations
  users
  user-id
  user
  create-environment
  project-settings-in-environment
  compare
  feature-history
  feature-history-detail
  project-settings
  permissions
  segments
  organisation-settings
  organisation-permissions
  organisation-usage
  organisation-settings-redirect
  organisation-projects
  account-settings
  audit-log
  organisations
  audit-log-item
  create-organisation
  project-redirect
  account

- Update the `announcement_per_page` FF with a new key named "pages", and add it the page names for example: ["segments", "users"]


matthewelwell:
@novakzaballa I think we discussed creating a different component for this. What made you change your mind? If we stick with the announcement feature, I think we want to consider a couple of extra things: 

 1. Should we allow the value to be a list as well as an object, and handle the typing in the code? This would be a bit ugly, but it would allow us flexibility in the future while maintaining backwards compatibility. 
 2. Should we add another new attribute which is e.g. `"allow_dismiss"` that we can set to false for announcements that we want to ensure are permanent?

frontend/web/components/App.js
```diff
@@ -350,12 +351,27 @@ const App = class extends Component {
       return <div>{this.props.children}</div>
     }
     const announcementValue = Utils.getFlagsmithJSONValue('announcement', null)
+    const announcementPerPageDismissed = flagsmith.getTrait(
```
kyle-ssg: No need to pollute app with this logic, can go in the component

frontend/web/components/App.js
```diff
@@ -409,6 +425,12 @@ const App = class extends Component {
                         </div>
                       </div>
                     )}
+                    {user && showAnnouncementPerPage && announcementInPage && (
+                      <AnnouncementPerPage
```
kyle-ssg: Might make more sense to just call this <Announcement

frontend/web/components/App.js
```diff
@@ -350,12 +351,27 @@ const App = class extends Component {
...
+      announcementPerPageDismissed !== announcementPerPageValue.id
+    Utils.getFlagsmithHasFeature('announcement_per_page') &&
+      announcementPerPageValue?.pages?.length > 0
     const dismissed = flagsmith.getTrait('dismissed_announcement')
     const showBanner =
       announcementValue &&
       (!dismissed || dismissed !== announcementValue.id) &&
       Utils.getFlagsmithHasFeature('announcement') &&
       this.state.showAnnouncement
+    const announcementInPage = announcementPerPageValue?.pages?.some((page) =>
```
kyle-ssg: Hmm, would this work for urls with ids or does this need to match in a more detailed way?

If it's the latter, note examples like this 

```
 matchPath(pathname, {
      exact: false,
      path: '/project/:projectId',
      strict: false,
    })
```

frontend/web/components/App.js
kyle-ssg: Also on this, does a page-specific announcement take priority over the announcement banner?  It makes sense to combine all of this into 1 component.

novakzaballa: I want to keep them separated so we can show both at the same time, a general announcement and a page-specific announcement. Of course, I'll resume as much logic as possible, however, these components are very simple, anyway.

frontend/web/components/App.js
novakzaballa: I understand this should target a specific page and not particular entities.
Maybe in the future if we need it at some point.

frontend/web/components/App.js
novakzaballa: As mentioned before I prefer to have 2 different components.

frontend/web/components/App.js
kyle-ssg: Hmm I don't understand how this will work, how would this target a page in remote config?

Nearly every page has ids in the pathname e.g. 
'/project/x/environment/y/features'

kyle-ssg (approved):

kyle-ssg (approved):

Issue #4145, opened 2024-06-11
Prevent versioned change requests from overwriting each other


matthewelwell:
The way in which change requests work when versioning is enabled is such that a new version is created as a snapshot of the current state of a feature in a given environment when the change request is opened. This is illustrated in the following process: 

Given Feature A, Segment X and Segment Y. 

1. Create CR1 which adds a segment override for Segment X 
2. Create CR2 which adds a segment override for Segment Y
3. Publish CR1
4. Publish CR2

The expected behaviour here is that both segment overrides should exist after following this process. Currently, the behaviour however, is that the end result is that only the override for Segment Y applied in CR2 will exist after following these steps because the override for Segment X did not exist when CR2 was created.

matthewelwell:
Some thoughts on possible (perhaps temporary) improvements: 

 1. We should show an alert when viewing / publishing a change request if another change request has been published between the time that this one was created and the current time. 
 2. We could perhaps impose a limit on the number of change requests for a given feature (so that only 1 change request can exist for a given feature)
 3. We should allow edits to be made to an existing change request (by it's author)

Some thoughts on a more permanent solution: 

1. Change requests should consist of a 'diff' only, e.g. `segment_overrides_to_add: [...]`, `segment_overrides_to_update: [...]`, `segment_overrides_to_delete: [...]`, `environment_feature_state_updates: {...}`. When a change request is published, these changes can be applied to a newly created version at that point (which will therefore include any changes made by other change requests up to that point). 
2. We could implement a workflow to handle 'merges' so that when you open a change request, it shows that other changes have been made to a given feature since the change request was created and it won't let you publish it until a merge is completed. This merge should step through any differences between the current version and the version in the CR and verify if they should be added to the CR's version. 

My preference here would be option (1) since I believe that will handle scheduled changes as well, but we will likely still need to consider how we deal with merge conflicts between multiple change requests. For example, if CR1 enables a feature for a segment, and CR2 adds a value for it, how would we handle this?

matthewelwell:
As discussed with @kyle-ssg the permanent solution will behave as follows. 

When creating a change request, a 'diff' will be provided to the API. This diff will consist of: 

 1. The updates to make to existing components of the feature in the given environment (the environment default state and value, and any modifications to existing segment overrides)
 2. Any new segment overrides to create
 3. Any segment overrides to remove

[...]

PR #4245, opened 2024-06-26
feat: Improve versioned change requests to handle multiple open CRs for single feature

matthewelwell:
## Changes

This PR is the 'Core' part of the resolution to #4145.  

In this PR, I've added most of the business logic which hinges on the `VersionChangeSet` model. This model is created alongside a change request (as per the code in [this workflows PR](https://github.com/Flagsmith/flagsmith-workflows/pull/41)) and holds the information about the changes that should be applied when the change request is published (or goes live if it's scheduled). 

The reason for this is to remove the need to duplicate the version when a change request is created, resulting in conflicts / clobbering of changes that happen between PR creation and publish. 

As part of this work, however, we need to account for the fact that there could still be conflicts, they're just less likely and less destructive. The conflict behaviour here checks if changes have been made to the same component of a feature in a given environment (e.g. a specific segment override, or the environment default value). These conflicts are then surfaced by the FE to provide visibility. 

I've added enough tests in this repository to satisfy the coverage requirements, but a lot of the testing has been done via full integration tests in the flagsmith-workflows PR [here](https://github.com/Flagsmith/flagsmith-workflows/pull/41).


api/features/versioning/models.py
```diff
@@ -163,3 +167,212 @@ def clone_to_environment(
...
+        from features.versioning.tasks import publish_version_change_set
+
+        kwargs = {"version_change_set_id": self.id, "user_id": user.id}
+        if not self.live_from or self.live_from < timezone.now():
+            publish_version_change_set(**kwargs)
+        else:
+            kwargs["is_scheduled"] = True
+            publish_version_change_set.delay(kwargs=kwargs, delay_until=self.live_from)
+
+    def get_conflicts(self) -> list[Conflict]:
```
matthewelwell, line 264: Note that I've added some unit tests for this method but I have covered a lot more use cases with some integration tests added to the `flagsmith-workflows` repository in the PR [here](https://github.com/Flagsmith/flagsmith-workflows/pull/41).

api/pyproject.toml
```diff
@@ -170,6 +170,7 @@ hubspot-api-client = "^8.2.1"
 djangorestframework-dataclasses = "^1.3.1"
 pyotp = "^2.9.0"
 flagsmith-task-processor = { git = "https://github.com/Flagsmith/flagsmith-task-processor", tag = "v1.0.0" }
+flagsmith-common = { git = "https://github.com/Flagsmith/flagsmith-common", branch = "feat/add-serializers-for-versioning-change-requests" }
```
matthewelwell: TODO: update to use a tag.

api/pyproject.toml
```diff
@@ -193,7 +194,7 @@ flagsmith-ldap = { git = "https://github.com/flagsmith/flagsmith-ldap", tag = "v
 optional = true
 
 [tool.poetry.group.workflows.dependencies]
-workflows-logic = { git = "https://github.com/flagsmith/flagsmith-workflows", tag = "v2.3.7" }
+workflows-logic = { git = "https://github.com/flagsmith/flagsmith-workflows", branch = "feature(versioning)/handle-multiple-change-requests-for-single-feature" }
```
matthewelwell: TODO: update to use a tag.

api/conftest.py
```diff
@@ -587,6 +592,19 @@ def segment_featurestate(feature_segment, feature, environment):
     )
 
 
+@pytest.fixture()
+def another_segment_featurestate(
+    feature: Feature, environment: Environment, another_segment: Segment
+):
```
zachaysan: Nit: Missing return type

api/features/versioning/views.py
```diff
@@ -211,14 +213,18 @@ def get_serializer_context(self):
         context["environment_feature_version"] = self.environment_feature_version
         return context
 
-    def perform_create(self, serializer: CreateSegmentOverrideFeatureStateSerializer):
+    def perform_create(
+        self, serializer: CustomCreateSegmentOverrideFeatureStateSerializer
+    ):
```
zachaysan: Nit: Missing return typing

api/features/versioning/views.py
```diff
@@ -211,14 +213,18 @@ def get_serializer_context(self):
...
         serializer.save(
             feature=self.feature,
             environment=self.environment,
             environment_feature_version=self.environment_feature_version,
         )
 
-    def perform_update(self, serializer: CreateSegmentOverrideFeatureStateSerializer):
+    def perform_update(
+        self, serializer: CustomCreateSegmentOverrideFeatureStateSerializer
+    ):
```
zachaysan: Nit: Missing return typing

api/tests/unit/features/versioning/test_unit_versioning_tasks.py
```diff
@@ -246,3 +258,209 @@ def test_enable_v2_versioning_for_scheduled_changes(
...
+    assert mailoutbox[0].subject == change_request_1.email_subject
+    assert mailoutbox[0].to == [staff_user.email]
+    assert mailoutbox[0].body == render_to_string(
+        "versioning/scheduled_change_failed_conflict_email.txt",
+        context={
+            "change_request": change_request_1,
+            "user": staff_user,
+            "feature": feature,
+        },
+    )
```
zachaysan, line 356: Ah I wish I knew about using the `render_to_string` function for the email testing code I had written for the API usage.

zachaysan (approved):
Overall looks great. I didn't find any major flaws and the comments, especially in the tests, made it easy to understand.

api/features/versioning/models.py
```diff
@@ -163,3 +167,210 @@ def clone_to_environment(
 
         _clone.save()
         return _clone
+
+
+class VersionChangeSet(LifecycleModelMixin, SoftDeleteObject):
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+    published_at = models.DateTimeField(blank=True, null=True)
+    published_by = models.ForeignKey(
```
gagantrivedi, line 176: How is this going to work if the change is made by `master_api_key`?

matthewelwell: Yeah, that's a good question. I did have a TODO in the code to handle API keys BUT the reason that I left it out in the end is because the only way to publish a version change set currently is through a Change Request and, at the moment, change requests themselves don't support API keys. 

I've just created an issue for this [here](https://github.com/Flagsmith/flagsmith-workflows/issues/42) but I do think that the use cases for using an API key in change request workflows are pretty slim.

gagantrivedi (approved):

Issue #1535, opened 2022-10-12
WIP: More granular feature level permissions


dabeeeenster:
# Overview

Where larger teams are working on a single Flagsmith Project, it would be useful for them to be able to assign RBAC privieldges based on a subset of Flags within the Project. 

# Potential Solution

The current proposed solution involves leveraging the tags that already exist within the platform. We would add the ability, when modifying the permissions for a group in an environment, to add a subset of tags that that permission / those permissions apply to. For example, you may have a group of users that should only have the 'Update feature state' permission against a subset of tags. One UI option to illustrate this is to add the ability to add tags below each permission, as per the following screenshot.

<img width="626" alt="Screenshot 2023-02-08 at 11 41 30" src="https://user-images.githubusercontent.com/14089968/217535912-e5a99345-e58a-45bb-8e31-1975902b032e.png">

# Detailed Example

TBC

# Issues/Open Questions

- Who defines what tag a Flag is associated with?
- Would this require a new concept of 'Roles'? Currently we only have the concept of groups which means that users will likely belong to more than one group. 
- Is there a way we can allow existing customers to migrate?
- Should we have a setting against the project which enables these tag-based permissions for the project?
- Should we only allow the management of tag based permissions for a group? This seems to make sense since customers that want to use permissions on such a granular level are likely using groups exclusively anyway.

# Other Possible Approaches

### Environment / Feature Combination Level Permissions

The tags approach makes the management of the permissions slightly simpler, however, it may not be the best solution for creating the most granular permissions. Another option here would be to allow admins to create permissions for groups (& users?) against a combination of a feature and an environment. The UI management of these permissions would be challenging but it would likely allow admins to create more granular permissions as needed.

matthewelwell:
Some thoughts based on a conversation with @gagantrivedi . 

1. Initial idea was to have 'tag based permission' be opt-in and, if enabled, the Update Feature State permission at the environment would be ignored. There would be View / Edit / Delete permissions on each tag. This is deemed to not be a viable solution as it would be hard to apply these permissions to individual environments, since the tags live at the project level. 
2. Second idea (which is as per the screenshot in the description) is to add tags to a role. These tags could be added to environment / project roles such that the permissions are only given to features with those tags applied. 

Next steps: start implementing option (2) above and see where it gets us :)

dabeeeenster:
Yep agree - leveraging roles seems the right path here

gagantrivedi:
[...]

PR #4250, opened 2024-06-26
feat: Add more granular permissions based on tags

novakzaballa:

## Changes

Based on the issue: https://github.com/Flagsmith/flagsmith/issues/1535

- Tags can now be added to roles to enhance granularity in permissions.
Screenshot 2024-06-26 at 3 24 26 PM

To properly test the functionality, this [issue](https://github.com/Flagsmith/flagsmith/issues/4287) must be resolved first.

- Go to the "users and permissions" tab
- Click the "roles" tab
- Select the project where you have the tags you want to manage.
- Select the tags, and click save/update button


frontend/web/components/modals/CreateRole.tsx
```diff
@@ -373,6 +371,44 @@ const CreateRole: FC<CreateRoleType> = ({
           id='description'
           placeholder='E.g. Some role description'
         />
+        <FormGroup className='mb-5 setting'>
+          <InputGroup
+            title={'Tags'}
+            tooltip={'Select the project where you want to manage your tags'}
```
matthewelwell: This looks wrong, we need to provide good context here about what adding tags to a role means. Can we link out to some documentation?

novakzaballa: I added documentation

frontend/web/components/modals/CreateRole.tsx
```diff
@@ -373,6 +371,54 @@ const CreateRole: FC<CreateRoleType> = ({
...
+                  >
+                    Learn more.
+                  </Button>
+                </p>
+                <div className='mb-2' style={{ width: 250 }}>
+                  <ProjectFilter
+                    organisationId={organisationId!}
+                    onChange={(p) => {
+                      setProject(p)
+                      getTags(
```
kyle-ssg: Is there a reason why this is called? seems wrong.

kyle-ssg: Looking at this more closely, it does look like this is unnecessary. AddEditTags uses useGetTagsQuery.

kyle-ssg:
I don't understand what this PR does, what does associating tags to a role do?

image

kyle-ssg:
I think I understand, so this lets us create a role with permissions but just targets a tagged set of flags?

If so I have further questions that are not indicated by this work, apologies if I'm assuming something wrong:

**Presumably only certain permissions are affected by the tags that are selected and the rest aren't?**
   - If this is the case we need to clearly show what setting these affects, it should be in the permissions tab clearly showing what selecting tags does.
   
**I'm going to assume that this affects Delete feature,  Create feature, Update feature state, Create change request and Approve change request?** 
   - If this is the case, isn't it quite a big assumption that all the permissions are for the same tags? It feels like this should be set for each permission otherwise you'd have to create multiple roles with the same users/groups to achieve what you want.

frontend/web/components/modals/CreateRole.tsx
novakzaballa: You're right, removed.

novakzaballa:
> I think I understand, so this lets us create a role with permissions but just targets a tagged set of flags?
> 
> If so I have further questions that are not indicated by this work, apologies if I'm assuming something wrong:
> 
> **Presumably only certain permissions are affected by the tags that are selected and the rest aren't?**
> 
> * If this is the case we need to clearly show what setting these affects, it should be in the permissions tab clearly showing what selecting tags does.
> 
> **I'm going to assume that this affects Delete feature, Create feature, Update feature state, Create change request and Approve change request?**
> 
> * If this is the case, isn't it quite a big assumption that all the permissions are for the same tags? It feels like this should be set for each permission otherwise you'd have to create multiple roles with the same users/groups to achieve what you want.

This PR aims to implement the backend created for this [issue](https://github.com/Flagsmith/flagsmith/issues/1535), This means that when a feature is tagged, the role to which a user belongs must have the same flag to delete and update a feature.

kyle-ssg:
I'm still basing this on assumptions about how this works, but I think either:

- The UI needs to be improved by moving this to the permissions tab and clearly indicating what it does. A user seeing this for the first time would have a lot of questions. I understand that there are docs, but that doesn't mean our UI should be so unclear.
  - The documentation needs to contain much more information.
  - Are only certain permissions affected by the selected tags while the rest aren't?
  - If it works the way I think it does, adding a tag to an existing role without understanding its impact is quite dangerous. We need to clearly state what that would do.

- Consider whether these tags should be at the permission level instead. This would automatically make it more obvious by how the front end would need to work.

novakzaballa:
> I'm still basing this on assumptions about how this works, but I think either:
> 
> * The UI needs to be improved by moving this to the permissions tab and clearly indicating what it does. A user seeing this for the first time would have a lot of questions. I understand that there are docs, but that doesn't mean our UI should be so unclear.
>   
>   * The documentation needs to contain much more information.

1. Added more information about the tags in the role.


>   * Are only certain permissions affected by the selected tags while the rest aren't?

2. Only the Delete feature and Update Feature state permissions are affected.

>   * If it works the way I think it does, adding a tag to an existing role without understanding its impact is quite dangerous. We need to clearly state what that would do.

3. Added a better description in the UI and the Docs.

> * Consider whether these tags should be at the permission level instead. This would automatically make it more obvious by how the front end would need to work.

4. Moved the tags into a new Tags tab inside the permissions tab

kyle-ssg:
Hmm I don't think we can provide a good solution to this without some rethinking of the API given the above.

We currently have a well defined use case provided to us:
A company has separate sections of their site managed by different teams that are a mix of developers, QAs etc. Whenever features are created they are tagged with the team e.g. "team_a".


Looking through our permissions it seems like the following are relevant for this use case

- Delete feature: a member of team A should not be able to delete a feature unless it's tagged with "team_a"
- Create Change Request: a member of team A should not be able to suggest changes for team_b's features
- Approve Change Request: a member of team A should not be able to approve changes for team_b's features
- Update Feature State: a member of team A should not be able to make updates to team_b's features
- Manage Segment Overrides: a member of team A should not be able to make updates to team_b's features

In addition to the above, the team would like to define the tag permissions in isolation. e.g. a QA can only approve change requests. 


Maybe we could have this for now, but would it mean that moving towards this would be difficult RE migration?

matthewelwell:
See comment [here](https://github.com/Flagsmith/flagsmith/issues/1535#issuecomment-2250458494) following an offline discussion which should address comments above.

kyle-ssg:
As discussed, let's hide irrelevant things when opting into tag based permissions

image

I think we can hide:
- The organisation tab
- Any project in the project tab that isn't the one selected
- Any project and environment features that aren't relevant, including View environment and View Project

novakzaballa:
> 
> I think we can hide:
> 
> * The organisation tab

This tab hides when a tag is saved in the role

> * Any project in the project tab that isn't the one selected

Projects that do not have the same tags as the role should not be displayed.

> * Any project and environment features that aren't relevant, including View environment and View Project

  All features that aren't relevant have been hidden.

PR #4265, opened 2024-06-28
feat: Create change requests for segments

### LINKED ISSUE #4044: Epic: Change requests for segments

When editing segments, I want to be able to request a review from other members of my organisation. 

Extend the Change Requests functionality to be able to operate on segments as well as features. 

N/a

### Additional context

_No response_

zachaysan:
## Changes

This PR works in concert with [this other PR](https://github.com/Flagsmith/flagsmith-workflows/pull/39) on the workflows repo.

This PR introduces many changes to support the addition of segments in change requests. The primary logic around transferring change request segment data into the target segment data involves re-using logic that was originally written for the segment versioning. 

Due to the primary manager (`.objects`) now including versioned or change requested segments copies, some amount of work was necessary to get all the aspects of the existing code to pass. The `.live_objects` manager is used for segments that are not merely versions or change request segments but in some places manually filtering segments was the only solution.

Added tests across the test suite, including new tests for duplicate segments to test filtering, as well as tests located in with [the other workflows PR](https://github.com/Flagsmith/flagsmith-workflows/pull/39) to get coverage of edge cases.


api/features/workflows/core/models.py
```diff
@@ -179,6 +180,30 @@ def _publish_change_sets(self, published_by: "FFAdminUser") -> None:
         for change_set in self.change_sets.all():
             change_set.publish(user=published_by)
 
+    def _publish_segments(self) -> None:
+        segments = list(self.segments.all())
+        if not segments:
+            return
+
+        for segment in segments:
```
matthewelwell: Is there a reason to do the empty check separately instead of just relying on the `for` loop? e.g. 

```suggestion
        for segment in self.segments.all():
```

api/features/workflows/core/models.py
```diff
@@ -179,6 +180,30 @@ def _publish_change_sets(self, published_by: "FFAdminUser") -> None:
...
+            target_segment = segment.version_of
+            assert target_segment != segment
+
+            # Deep clone the segment to establish historical version.
+            target_segment.deep_clone()
```
matthewelwell, line 204: Couldn't we optimise this by sending in an argument, or via a separate method, which doesn't clone the target_segments rules? Since we're just deleting them again below, it seems unnecessary to do this.

api/projects/models.py
```diff
@@ -142,19 +142,22 @@ def edge_v2_identity_overrides_migrated(self) -> bool:
         return self.edge_v2_migration_status == EdgeV2MigrationStatus.COMPLETE
 
     def get_segments_from_cache(self):
+        from segments.models import Segment
```
matthewelwell: Hmm... can we avoid this if we convert this method into a function and move it out of `models`? Since this method gets called in some pretty hot paths, I'd rather avoid local imports.

api/segments/models.py
```diff
@@ -60,24 +59,30 @@ class Segment(
...
         null=True,
         blank=True,
     )
+
+    change_request = models.ForeignKey(
+        "workflows_core.ChangeRequest",
+        on_delete=models.CASCADE,
+        null=True,
+        related_name="segments",
+    )
```
matthewelwell: let's add `blank=True` here too in case it appears in the django admin anywhere

api/segments/models.py
```diff
@@ -145,6 +150,26 @@ def set_version_of_to_self_if_none(self):
         self.save()
         segment_audit_log_helper.unset_skip_audit_log(self.id)
 
+    def shallow_clone(
```
matthewelwell, line 154: Ah, is this the method I described above?

api/tests/unit/features/workflows/core/test_unit_workflows_models.py
```diff
@@ -773,3 +795,61 @@ def test_change_request_live_from_for_change_request_with_change_set(
...
+    Condition.objects.create(
+        rule=child_rule2,
+        property="child_rule2",
+        operator=PERCENTAGE_SPLIT,
+        value="0.2",
+        created_with_segment=False,
+    )
+
+    # When
+    change_request._publish_segments()
```
matthewelwell: Why not just call `change_request.commit()` ?

api/tests/unit/projects/test_unit_projects_models.py
```diff
@@ -55,6 +56,35 @@ def test_get_segments_from_cache_set_not_called(project, segments, monkeypatch):
     mock_project_segments_cache.set.assert_not_called()
 
 
+@pytest.mark.django_db()
```
matthewelwell: We don't need this since the `project` and `segment` fixtures will require the `db` fixture already.

api/util/mappers/engine.py
```diff
@@ -194,7 +195,17 @@ def map_environment_to_engine(
...
 
     # Read relationships - grab all the data needed from the ORM here.
+
+    # Because of the queryset parameter of the calling code's
+    # Prefetch this queryset will actually load the live_objects
+    # manager for the Segment lookup.
     project_segments: List["Segment"] = project.segments.all()
+
+    # Even though the main calling code filters via a prefetch,
+    # we still want to gaurd this function in case there are other
```
matthewelwell:
```suggestion
    # we still want to guard this function in case there are other
```

🎃

matthewelwell (commented):
In general this looks really good. I've added a few minor comments. 

One other thing that stood out to me was around audit trail - are we happy that we're writing the correct audit logs here? I guess we're covered by those written for the change request itself?

api/tests/unit/features/workflows/core/test_unit_workflows_models.py
zachaysan: Sure that's no problem

api/features/workflows/core/models.py
zachaysan: Nice catch, implemented

api/features/workflows/core/models.py
zachaysan: This is for establishing a versioned history for the soon-to-be-changing segment. Without a deep clone we won't have a copy of a previous version of the segment with the associated rules and conditions. We can always skip doing the deep clone but then our version history would have gaps in it if change requests are used to change segments.

api/projects/models.py
zachaysan: I'll move it to a function, but I think we should keep the method on project even if it's just a reference to the function because it's a natural calling path.

zachaysan: Ok so I moved it to the function but I'm calling `apps.get_model("segments", "Segment")` in the function because of the circular import issue. Is that ok or does that still trigger your hot path sensitivity? If it does trigger I'll move it out of the method call on the `Project` model and update callers to just use the function directly.

api/segments/models.py
zachaysan: Good idea. Implemented.

api/segments/models.py
zachaysan: No the above method is necessary to preserve the rule versioning. This is used in the workflows process when creating a change request segment where we don't want a deep clone of the rules.

zachaysan:
> In general this looks really good. I've added a few minor comments.
> 
> One other thing that stood out to me was around audit trail - are we happy that we're writing the correct audit logs here? I guess we're covered by those written for the change request itself?

The audit trail tests have already been implemented when working on the segments versioned code. In addition to that, the audit trail logs are implemented in the workflows repository.

api/features/workflows/core/models.py
matthewelwell: I don't think that's true is it? I get that we want to keep the history, but we don't need to actually duplicate the rules and conditions, we just need to archive the ones from the current (outgoing) segment. We don't need to create them for the new segment because we're not going to use them - we just delete them below.

api/projects/models.py
matthewelwell: Hmm... interesting dilemma, I'm not sure. I think I prefer not using `apps.get_model()` as I suspect that also involves a local import. 

I'm keen to get other opinions on this - @gagantrivedi

api/features/workflows/core/models.py
matthewelwell: As discussed, this should be thought of like regular updates to a segment where we create the clone to create the version, then modifying the new 'draft' version with the data from the change request. 

Let's update the comments for this method to try and make this a little clearer.

zachaysan: Ok the comment has been updated.

api/app/urls.py
```diff
@@ -54,8 +55,24 @@
 
 if settings.WORKFLOWS_LOGIC_INSTALLED:
     workflow_views = importlib.import_module("workflows_logic.views")
+    from projects.views import ProjectViewSet  # pragma: no cover
```
matthewelwell: I think we can just move the `pragma: no cover` to the whole conditional block here?

matthewelwell: In fact, this whole section feels a little bit odd to me to be honest. Why not move this conditional logic to `projects/urls.py` and just add the `ProjectChangeRequestViewSet` there?

api/features/workflows/core/migrations/0011_add_project_to_change_requests.py
```diff
@@ -0,0 +1,62 @@
+# Generated by Django 4.2.15 on 2024-09-17 15:34
+
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+def set_project_for_existing_environments(apps, schema_model):
```
matthewelwell:
```suggestion
def set_project_for_existing_change_requests(apps, schema_model):
```

api/features/workflows/core/migrations/0011_add_project_to_change_requests.py
```diff
@@ -0,0 +1,62 @@
...
+    ChangeRequest = apps.get_model("workflows_core", "ChangeRequest")
+
+    for change_request in ChangeRequest.objects.filter(
+        environment_id__isnull=False
+    ).select_related("environment"):
+        change_request.project = change_request.environment.project  # pragma: no cover
+        change_request.save()  # pragma: no cover
```
matthewelwell: Again, the `pragma: no cover` here could be on the conditional block. That being said, I think we could write a short migration test for this?

api/features/workflows/core/models.py
```diff
@@ -77,10 +77,19 @@ class ChangeRequest(
         null=True,
     )
 
+    # Change requests get deleted in a delegated task when a project is deleted.
+    project = models.ForeignKey(
+        "projects.Project",
+        on_delete=models.SET_NULL,
+        related_name="change_requests",
+        null=True,
```
matthewelwell: Wouldn't things be easier here (and in certain situations below) if we made `project` not nullable? We can set it to not null after running the migration code you wrote. We can then use a BEFORE_CREATE hook to set the project if it's not already set.

api/features/workflows/core/models.py
```diff
@@ -209,9 +243,24 @@ def _get_environment(self) -> typing.Optional["Environment"]:
...
+        return self.project or (self.environment and self.environment.project)
 
     def is_approved(self):
+        if self.environment:
+            return self.is_approved_via_environment()
+        if self.project:
+            return self.is_approved_via_project()
+        raise RuntimeError(  # pragma: no cover
+            "Unable to approve change request without environment or project"
+        )
```
matthewelwell: If project was not nullable, we could get rid of this entirely (the `pragma: no cover` is a bit of a code smell to me)

api/features/workflows/core/models.py
```diff
@@ -228,8 +277,15 @@ def url(self):
...
-        url += f"/environment/{self.environment.api_key}"
+        if self.environment:
+            url += f"/project/{self.environment.project_id}"
+            url += f"/environment/{self.environment.api_key}"
+        elif self.project_id:
+            url += f"/projects/{self.project_id}"
+        else:
+            raise RuntimeError(  # pragma: no cover
+                "Change request missing fields for URL"
+            )
```
matthewelwell: Again, this could all be simplified if project was not nullable

api/features/workflows/core/tasks.py
```diff
@@ -0,0 +1,8 @@
+from task_processor.decorators import register_task_handler
+
+from features.workflows.core.models import ChangeRequest
+
+
+@register_task_handler()
+def delete_change_request(change_request_id: int) -> None:
+    ChangeRequest.objects.get(pk=change_request_id).delete()
```
matthewelwell: Leaving a comment here because this task seems slightly odd to me 🤔

api/projects/migrations/0025_add_change_request_project_permissions.py
```diff
@@ -0,0 +1,61 @@
...
+def remove_default_project_permissions(apps, schema_model):
+    PermissionModel = apps.get_model(  # pragma: no cover
+        "permissions", "PermissionModel"
+    )
+    PermissionModel.objects.get(  # pragma: no cover
+        key=MANAGE_PROJECT_LEVEL_CHANGE_REQUESTS
+    ).delete()
+    PermissionModel.objects.get(  # pragma: no cover
+        key=APPROVE_PROJECT_LEVEL_CHANGE_REQUESTS
+    ).delete()
```
matthewelwell:
```suggestion
def remove_default_project_permissions(apps, schema_model):  # pragma: no cover
    PermissionModel = apps.get_model(
        "permissions", "PermissionModel"
    )
    PermissionModel.objects.get(
        key=MANAGE_PROJECT_LEVEL_CHANGE_REQUESTS
    ).delete()
    PermissionModel.objects.get(
        key=APPROVE_PROJECT_LEVEL_CHANGE_REQUESTS
    ).delete()
```

api/projects/migrations/0025_add_change_request_project_permissions.py
```diff
@@ -0,0 +1,61 @@
...
+    manage_description = approve_description = None
+
+    for project_permission in PROJECT_PERMISSIONS:
+        if project_permission[0] == MANAGE_PROJECT_LEVEL_CHANGE_REQUESTS:
+            manage_description = project_permission[1]
+        elif project_permission[0] == APPROVE_PROJECT_LEVEL_CHANGE_REQUESTS:
+            approve_description = project_permission[1]
+
+    assert manage_description
+    assert approve_description
```
matthewelwell: This code took me a little while to understand. I don't think we've worried about doing this dynamically in the past. I don't mind having the duplication here really. 

Alternatively, if you really want to keep this, maybe let's add a comment. Or, just define the descriptions as constants alongside the permissions themselves.

api/projects/models.py
```diff
@@ -106,6 +106,7 @@ class Project(LifecycleModelMixin, SoftDeleteExportableModel):
         default=30,
         help_text="Number of days without modification in any environment before a flag is considered stale.",
     )
+    minimum_change_request_approvals = models.IntegerField(blank=True, null=True)
```
matthewelwell, line 109: This will also need adding to relevant serializer(s).

api/tests/unit/features/workflows/core/test_unit_workflows_models.py
```diff
@@ -897,3 +977,31 @@ def _create_segment(percentage_value: int) -> Segment:
...
+def test_approval_via_project(project_change_request: ChangeRequest) -> None:
+    # Given - The project change request fixture
+    assert project_change_request.environment is None
+    assert project_change_request.project.minimum_change_request_approvals is None
+
+    # When
+    is_approved = project_change_request.is_approved()
+
+    # Then
+    assert is_approved is True
```
matthewelwell, line 991: Is this a legit use case? For environments at least, `minimum_change_request_approvals = None` means that change requests aren't enabled at all.

api/tests/unit/features/workflows/core/test_unit_workflows_models.py
```diff
@@ -897,3 +977,31 @@ def _create_segment(percentage_value: int) -> Segment:
...
+
+def test_url_via_project(project_change_request: ChangeRequest) -> None:
+    # Given
+    assert project_change_request.environment is None
+
+    # When
+    url = project_change_request.url
+
+    # Then
+    project_id = project_change_request.project.id
```
matthewelwell: ... feel bad nitpicking after all the other comments, but this is a particularly bad habit to get into with django, so I'm calling it out here, even though it's just test code. 

```suggestion
    project_id = project_change_request.project_id
```

api/app/urls.py
zachaysan: Ok I've moved the urls to the projects urls module, but for some reason I'm dealing with a non-update of the image when I try to push the code so I haven't been able to test the change yet.

api/features/workflows/core/migrations/0011_add_project_to_change_requests.py
zachaysan: Good idea. I've written the migration test.

api/features/workflows/core/models.py
zachaysan: Sounds good. I've done these changes.

api/features/workflows/core/models.py
zachaysan: Yep. Agreed. Have removed the exception.

api/features/workflows/core/models.py
zachaysan: Yep. Simplified it.

api/features/workflows/core/tasks.py
zachaysan: Since project is now non-nullable I switched this to just use the cascading deletion by default.

api/projects/migrations/0025_add_change_request_project_permissions.py
zachaysan: Ok I've duplicated the descriptions directly in the code

api/projects/models.py
zachaysan: Alright I've added it to `ProjectListSerializer` which looks like it will include the field to `ProjectRetrieveSerializer` and other serializers.

api/tests/unit/features/workflows/core/test_unit_workflows_models.py
zachaysan: According to a pre-existing test `test_change_request_is_approved_returns_true_when_minimum_change_request_approvals_is_none` I believe your understanding is incorrect. I just mirrored the behavior from the environments definition for the project piece.

api/tests/unit/features/workflows/core/test_unit_workflows_models.py
zachaysan: No problem, I've updated the code to match

api/app/urls.py
zachaysan: Ok it looks like the images are being created again. Must have either been a merge conflict with `main` or an intermittent outage. I was able to use the new urls after requiring to update the namespacing.

api/projects/models.py
zachaysan: I believe the models are stored in a [registry](https://docs.djangoproject.com/en/5.1/ref/applications/#django.apps.AppConfig.get_model) that can be looked up quickly during a call to `apps.get_model()` so I think we are safe in using this here. It makes sense that Django would cache the models in, say a list or dictionary.

api/features/workflows/core/migrations/0011_add_project_to_change_requests.py
```diff
@@ -0,0 +1,72 @@
...
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+def set_project_for_existing_change_requests(apps, schema_model):
+    ChangeRequest = apps.get_model("workflows_core", "ChangeRequest")
+
+    for change_request in ChangeRequest.objects.filter(
+        environment_id__isnull=False
+    ).select_related("environment"):
```
matthewelwell:
```suggestion
    ).select_related("environment", "environment__project"):
```

api/features/workflows/core/migrations/0011_add_project_to_change_requests.py
```diff
@@ -0,0 +1,72 @@
...
+        change_request.project = change_request.environment.project
+        change_request.save()
```
matthewelwell: I think we could feasibly use `ChangeRequest.objects.bulk_update()` here instead, right?

api/features/workflows/core/models.py
```diff
@@ -77,10 +78,19 @@ class ChangeRequest(
         null=True,
     )
 
+    # Change requests get deleted in a delegated task when a project is deleted.
```
matthewelwell: I'm not sure this comment is still valid? Perhaps we weren't using `models.CASCADE` before?

matthewelwell (commented):
Added a couple of minor comments, but this looks good on the whole. I guess the only question I have is whether there should also be a bunch of removed code here based on the stuff we moved to the flagsmith-common repo?

api/features/workflows/core/migrations/0011_add_project_to_change_requests.py
zachaysan: Yep! Implemented.

api/features/workflows/core/models.py
zachaysan: Good catch! Removed.

zachaysan:
> Added a couple of minor comments, but this looks good on the whole. I guess the only question I have is whether there should also be a bunch of removed code here based on the stuff we moved to the flagsmith-common repo?

Ok it took a while, but I've updated the repo with references to the flagsmith-common repo and of course removed the now unnecessary code.

api/permissions/migrations/0001_initial.py
matthewelwell, line 1: It always worries me when we make changes like this to migrations. We've been burnt by it in the past. The black formatter shouldn't apply to migration files. Just be careful of this in the future, and try to avoid large formatting changes to migration files, unless it's really obvious that it's only formatting that's changed.

api/pyproject.toml
```diff
@@ -196,7 +196,7 @@ flagsmith-ldap = { git = "https://github.com/flagsmith/flagsmith-ldap", tag = "v
 optional = true
 
 [tool.poetry.group.workflows.dependencies]
-workflows-logic = { git = "https://github.com/flagsmith/flagsmith-workflows", tag = "v2.5.0" }
+workflows-logic = { git = "https://github.com/flagsmith/flagsmith-workflows", branch = "feat/add_change_requests_to_segments" }
```
matthewelwell: Just a reminder that we'll need to update this before the merge.

api/segments/views.py
```diff
@@ -39,7 +40,7 @@ def get_queryset(self):
         )
         project = get_object_or_404(permitted_projects, pk=self.kwargs["project_pk"])
 
-        queryset = project.segments.all()
+        queryset = Segment.live_objects.filter(project=project)
```
matthewelwell, line 48: I'm surprised to see this here - I thought we'd updated all of these in a previous PR where we added the versioning?

api/tests/unit/segments/test_unit_segments_views.py
```diff
@@ -619,7 +619,7 @@ def test_update_segment_versioned_segment(
 
     # Before updating the segment confirm pre-existing version count which is
     # automatically set by the fixture.
-    assert Segment.all_objects.filter(version_of=segment).count() == 2
+    assert Segment.objects.filter(version_of=segment).count() == 2
```
matthewelwell, line 692: As above, I'm not sure why these have changed here?

matthewelwell (commented):
Added a few minor comments

api/permissions/migrations/0001_initial.py
zachaysan: Ok thanks for letting me know, although I'm pretty sure the formatting changes here are fine.

api/pyproject.toml
zachaysan: Yup!

api/segments/views.py
zachaysan: I believe some were missed or maybe there was a refactor in this PR where calling sites had to be upgraded.

api/tests/unit/segments/test_unit_segments_views.py
zachaysan: This PR changed the way objects are accessed. As you can see [here](https://github.com/Flagsmith/flagsmith/pull/4265/files#diff-13990717d51e5dbadd003d4a2fe9bf926d5238232e80aa73c48f450e4b02719bR86)

api/segments/views.py
zachaysan: Yes it was a refactor, see my comment below.

api/tests/unit/segments/test_unit_segments_views.py
matthewelwell: Hmm... in that case, I want to be really sure that we've updated ALL usages. For example, I think there's still a few that might need updating? Some examples:

 * https://github.com/Flagsmith/flagsmith/blob/main/api/integrations/launch_darkly/services.py#L244
 * https://github.com/Flagsmith/flagsmith/blob/main/api/util/mappers/engine.py#L197
 * https://github.com/Flagsmith/flagsmith/blob/main/api/projects/models.py#L122
 * https://github.com/Flagsmith/flagsmith/blob/main/api/projects/models.py#L151

There are a few more that I found using this regex search: `((?<!\w)Segment\.objects|(?<!feature_)segments)\.(get|filter|all|count)` which should find all instances of `Segment.objects.get|filter` and `x.segments.all|filter|count`. 

I ignored the ones found in the test directory, but all others should be looked at.

zachaysan: Ok these are updated and I've pushed the fix to the conflicts below.

matthewelwell (commented):
This all looks good to me now - there's just the (new?) conflicts to resolve, and the CI failures.

zachaysan:
> This all looks good to me now - there's just the (new?) conflicts to resolve, and the CI failures.

Conflicts and CI failures have been resolved. Are we good to merge (other than the update to the workflows repo which needs to be merged itself)?

matthewelwell (approved):
I'm approving this now, but we need to consider the release plan as this is quite a meaty change. It will need a fair amount of regression testing in staging.

api/pyproject.toml
zachaysan: Ok this has been updated to the latest tag for workflows.

api/projects/migrations/0025_add_change_request_project_permissions.py
```diff
@@ -0,0 +1,48 @@
...
+    PermissionModel.objects.get_or_create(
+        key=MANAGE_PROJECT_LEVEL_CHANGE_REQUESTS,
+        description=manage_description,
+        type=PROJECT_PERMISSION_TYPE,
+    )
+    PermissionModel.objects.get_or_create(
+        key=APPROVE_PROJECT_LEVEL_CHANGE_REQUESTS,
+        description=approve_description,
+        type=PROJECT_PERMISSION_TYPE,
+    )
```
khvn26, line 34:
```python
    PermissionModel.objects.get_or_create(
        key=MANAGE_PROJECT_LEVEL_CHANGE_REQUESTS,
        type=PROJECT_PERMISSION_TYPE,
        defaults={"description": manage_description},
    )
    PermissionModel.objects.get_or_create(
        key=APPROVE_PROJECT_LEVEL_CHANGE_REQUESTS,
        type=PROJECT_PERMISSION_TYPE,
        defaults={"description": manage_description},
    )
```

PR #4266, opened 2024-06-28
chore(build): API test image

khvn26:

## Changes

This adds a `flagsmith-api-test` image to PR Docker build workflow.
The `flagsmith-api-test` image:
- Is based on the "fat" Debian image, so has additional binaries required for development, like `git`, pre-installed.
- Includes dev dependencies.
- Offers Bash as an entrypoint.
- Offers `make test` as the default command.
The image will be used for integration testing the private packages. It will also be published on merge to main.

Tested the build locally.


Dockerfile
```diff
@@ -133,6 +136,20 @@ FROM api-runtime as api-runtime-private
 # Install SAML binary dependency
 RUN apt-get update && apt-get install -y xmlsec1 && rm -rf /var/lib/apt/lists/*
 
+# - Internal stages
+# * api-test [build-python]
+FROM build-python AS api-test
```
gagantrivedi, line 146: We are not generating a private cloud image for tests, right? i.e: it won't help here: https://github.com/Flagsmith/flagsmith-auth-controller/pull/20/files

khvn26: There's two ways we run integration tests now:

1. In the upstream repos using Compose, mounting the package code in a Docker volume.
2. In the main repo using a Make target + Github workflow.

Currently, this PR is aimed at use case 1. It'd be trivial to use the new image for the private package tests by using the `make install-private-modules integrate-private-tests test` as the entrypoint, though. However, one thing this setup won't support is running tests on Python 3.12, which we're doing in CI to simplify the future upgrade.

matthewelwell (approved):

PR #4289, opened 2024-07-05
fix: Non-admin project Role request to /projects/ID/features/ID Causes Crash

### LINKED ISSUE #4286: Non-admin project Role request to /projects/ID/features/ID Causes Crash

- [ ] Self Hosted with Docker
- [X] SaaS at flagsmith.com
- [X] Some other way (add details in description below)

### Describe the bug

When you are part of a role without being a project admin, making a request /projects/ID/features/Id/ causes a crash.

It was tested locally and in staging.

### Steps To Reproduce

1. Create a role with View Project, View environment, and update feature state, and assign it to test user
2. Log in with the test user
3. Go to the project -> environment that you have permission to access.
4. Select a feature and update the feature state
5. See the error

### Expected behavior

When I change the feature state value, the request should proceed without a 403 error and not display an error in the app

### Screenshots

https://github.com/Flagsmith/flagsmith/assets/41410593/d94dfc43-4ae7-425d-bf3a-c975f2f31069

Screenshot 2024-07-05 at 3 41 18 PM

novakzaballa:

## Changes

Tag IDs are only sent when there is permission that have tags. For example DELETE_FEATURE.

Fixes #4286

1. Create a role with View Project, View environment, and update feature state, and assign it to test user
2. Log in with the test user
3. Go to the project -> environment that you have permission to access.
4. Select a feature and update the feature state

### DIFF
--- a/api/features/permissions.py
+++ b/api/features/permissions.py
@@ -67,9 +67,12 @@ class FeaturePermissions(IsAuthenticated):
             required_permission = ACTION_PERMISSIONS_MAP.get(view.action)
             if required_permission in TAG_SUPPORTED_PROJECT_PERMISSIONS:
                 tag_ids = list(obj.tags.values_list("id", flat=True))
+                return request.user.has_project_permission(
+                    ACTION_PERMISSIONS_MAP[view.action], obj.project, tag_ids=tag_ids
+                )
 
             return request.user.has_project_permission(
-                ACTION_PERMISSIONS_MAP[view.action], obj.project, tag_ids=tag_ids
+                ACTION_PERMISSIONS_MAP[view.action], obj.project
             )
 
         if view.action == "segments":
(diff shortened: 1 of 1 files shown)


matthewelwell (commented):
I don't quite understand how this is any different? From what I can tell, the `tag_ids` variable is set to an empty list above, which will evaluate as falsy, the same as the default on the `has_project_permission` method. 

Could you add more description and ideally add a test that would replicate the issue?

novakzaballa:
> I don't quite understand how this is any different? From what I can tell, the `tag_ids` variable is set to an empty list above, which will evaluate as falsy, the same as the default on the `has_project_permission` method.
> 
> Could you add more description and ideally add a test that would replicate the issue?


Change the code so that instead of the value of tag_ids being [], it will be None. This change is based on this line of code: https://github.com/Flagsmith/flagsmith/blob/cdece74b830f60716e19ac6a3939db51ad808edc/api/permissions/permission_service.py#L74
This adjustment is needed because when making the GET request to `api/v1/projects/id/features/id/`, tag-based permissions should not be applied (we should only send an array, in the case of using the DELETE_FEATURE permission)

matthewelwell (approved):

PR #4298, opened 2024-07-08
chore: Rename metadata to custom fields

### LINKED ISSUE #4293: Rename "metadata" to "custom fields" (UI and docs only)

"Metadata" means "data that provides information about other data". This name is technically correct for what the Metadata feature does in Flagsmith, but it also includes things that it shouldn't such as:

* who created/modified an entity
* when an entity was created/modified
* permissions related to an entity
* tags

"Metadata" is a vague term that does not lend itself well to searching and does not describe what the feature actually does, which is to define custom fields when creating or updating Flagsmith entities.

Rename all instances of "Metadata" to "Custom fields". This should be stylised as "custom fields" and not "Custom Fields".

It's not necessary to rename the API or database fields - we might choose to do that later but I doubt it. The docs should mention what the relevant API fields are.

I can't really think of a better name than "custom fields". What we currently call "metadata" lets customers define custom fields - that's exactly what it is.

### Additional context

_No response_

kyle-ssg:

## Changes

- Renames "Metadata" to "Custom Fields"
- Improves wording in some of the docs
- Creates a docs redirect since it was technically live prior to this

- Went through every usage of the word metadata in docs and frontend


frontend/web/components/metadata/AddMetadataToEntity.tsx
matthewelwell, line 1: Should we rename the actual types here (and in the rest of the edited files) as well?

kyle-ssg: Im not sure, it felt a bit meaningless to change anything other than copy if we're not changing the actual API, both changing and not changing would be confusing one way or another

matthewelwell: Yep, fair enough - I did think that after I posted the comment.

matthewelwell (approved):

PR #4299, opened 2024-07-08
fix: feature-specific segments link

### LINKED ISSUE #4277: Permalinks to feature-specific segments do not work

- [ ] Self Hosted with Docker
- [X] SaaS at flagsmith.com

### Describe the bug

Permalinks to feature-specific segments such as https://app.flagsmith.com/project/16945/segments?id=30243 only load the segments page and do not open the modal showing the segment detail. Enabling the "Include Feature-Specific" toggle causes the modal to open.

### Steps To Reproduce

1. Create a feature-specific segment
2. Open the link to that segment, such as https://app.flagsmith.com/project/16945/segments?id=30243
3. Note that the segment modal does not open
4. Click the "Include Feature-Specific" button
5. Note that only now the modal will open

### Expected behavior

Links to feature-specific segments should behave the same way as regular segments, with the difference that the segments view should have the "Include Feature-Specific" option enabled.

"Include Feature-Specific" should be enabled so that it's possible to find the same feature-specific segment that was just opened after closing the modal for whatever reason (accidental or intentional).

### Screenshots

_No response_

kyle-ssg:

## Changes
Fixes urls to a feature-specific segment.

- Refreshed a url to a feature-specific segment


novakzaballa (approved):

rolodato (commented):
This does fix links to feature-specific segments, but it does not enable the "Include Feature-Specific" option. Is that feasible to do as well?

My reasoning is that if you open a link to a segment, there's nothing in the segment itself or the URL that tells you it's feature-specific. If you close the modal intentionally or accidentally, it's more difficult to go back to where you were - especially considering that segment modals do not add browser history, so you can't use Back to re-open them after closing.

Also suggest a different title: `fix: Fix links to feature-specific segments not opening the segment details modal`

kyle-ssg:
Ok if that's the usecase then it can only be solved by the modal being url based in the history, since segments are paged.

I've adjusted the behaviour such that the user can go back and forward in browser history

rolodato:
@kyle-ssg thanks for the update, I'm able to use browser navigation to get back to a closed segment modal now.

One last thing - would it be feasible to have the "Include Feature-Specific" option pre-selected if they open a link to a feature-specific segment? My reasoning is that, to get to a feature-specific segment from the Segments page, you need to have selected the "Include Feature-Specific" option at some point. Having it enabled when you open a link to a feature-specific segment seems more consistent with the navigation behaviour.

kyle-ssg:
Done @rolodato

rolodato:
@kyle-ssg I see we're including the `featureSpecific` URL parameter but it's not automatically added when opening a feature-specific segment if the parameter isn't already there.

This is definitely better than what we have now, but I was hoping to have this parameter added automatically if it's not there. For example, this URL is for a feature-specific segment:

```
https://flagsmith-frontend-staging-gui9akkrb-flagsmith.vercel.app/project/6605/segments?id=9766
```

which ends up as:

```
https://flagsmith-frontend-staging-gui9akkrb-flagsmith.vercel.app/project/6605/segments?id=9766&featureSpecific=false
```

Instead, I was expecting:

```
https://flagsmith-frontend-staging-gui9akkrb-flagsmith.vercel.app/project/6605/segments?id=9766&featureSpecific=true
```

kyle-ssg:
> @kyle-ssg I see we're including the featureSpecific URL parameter but it's not automatically added when opening a feature-specific segment if the parameter isn't already there.

I'm not sure how they'd get the url with a missing parameter to begin with. Regardless, this will now update based on the fetched segment.

rolodato:
It works! One last thing which I'm not sure was present before, is that toggling "Include Feature-Specific" changes the URL and adds a history entry, but navigating back/forward does not actually reflect this change in the page.

IMO toggling "Include Feature-Specific" should not add a history change, whether that happens through interaction or not. Or, it should add a history entry but reflect the state change correctly, but I would prefer the former option.


https://github.com/user-attachments/assets/169df061-e8fe-40bd-b5f9-64d8785b0b76

kyle-ssg:
In the interest of not keeping PRs around for weeks on end, I've merged this for now. I can adjust the behaviour of this at a later date

PR #4301, opened 2024-07-08
feat: versioned change request change sets

### LINKED ISSUE #4145: Prevent versioned change requests from overwriting each other

### LINKED ISSUE #4240: Warn when creating empty change requests

It's possible to create change requests that do not make any changes, by editing a feature and clicking "Schedule Update" without making any other changes. This is error-prone and can cause a nasty surprise when the go-live date for a feature arrives and nothing happens.

The only valid use case I can think for this would be to ensure a feature gets modified at a certain date, while other changes are being made or scheduled before this cutoff date. This seems like a valid but very fringe use case.

Add a mandatory confirmation step when creating a change request or scheduled change that makes no modifications to the current environment.

1. Show a warning banner - this could get missed, and the potential impact of missing it IMO is far worse than the possible annoyance of adding one more click to this workflow. Also becomes annoying if we need other info/warning banners for anything else on the same screen
2. Don't allow creating change requests that don't make any changes to the environment, by e.g. disabling the "create change request" or "schedule change" button if no changes to the feature were made. I would be also be okay with this solution until we can prove that there's a real customer need for creating change requests that don't make any changes.

### Additional context

_No response_

kyle-ssg:

## Changes

Adjusts the way change requests are created in order to prevent versioned change requests from overwriting each other.

- Create a change request with a segment override
- Create a change request with a remote config update
- Publish the segment override CR
- Check the diff of the second CR, it should recognise segment overrides as unchanged
- Fixes an issue where creating / actioning CRs doesn't update the count


matthewelwell:
@kyle-ssg I'm testing this locally now and it doesn't seem to be working (at least not how I expected). 

Steps: 

 * create 2 change requests that both operate on the environment default feature state (one which enables it, and one which sets a value). 
 * publish the one that enables the feature
 * view the other change request (that adds a value)

At this point, I am expecting to see a warning that there's a conflict (as you can see from the retrieve request in the console screenshot), but instead, I'm just seeing that the change request will disable the feature and set the value. 

image
image

Edit: it looks like this is also the case with segment overrides. If I create 2 CRs to modify a segment override, then publish one and view the other, I don't see any conflicts.

matthewelwell:
@kyle-ssg some things I've seen in testing: 

 1. All change requests seem to be showing this alert in the publish modal

image

 2. As discussed on Slack, I think that new segment overrides should be created at the bottom of the list instead of the top to avoid conflicts being created by shunting all the segment overrides down.

matthewelwell:
Hmm... I'm not sure if this is related to item (2) above, but I've noticed an issue (that might be tricky to solve!). The steps to reproduce are: 

 1. Given a feature that already has 1 segment override (segment A)
 2. Create a change request to add another segment override (Segment B) and move it to the bottom of the list (to try and avoid the issue as described in (2) above)
 3. In the network console we see that the FE sends a request which includes the (unmodified) original segment override (segment A) in `feature_states_to_update`

Further steps to actually visually see the issue: 

 4. Create another change request to update segment A
 5. Publish either change request and notice that the other change request shows a conflict

This should not create a conflict because segment A should not be modified as part of the change request to create the override for segment B.

matthewelwell:
~I've found another issue related to scheduled changes. When I try to publish a scheduled change, I see the following error:~ 

```
TypeError: scheduledDate.format is not a function
    at eval (ChangeRequestPage.js:203:37)
    at HTMLUnknownElement.callCallback (react-dom.development.js:188:14)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:237:16)
    at invokeGuardedCallback (react-dom.development.js:292:31)
    at invokeGuardedCallbackAndCatchFirstError (react-dom.development.js:306:25)
    at executeDispatch (react-dom.development.js:389:3)
    at executeDispatchesInOrder (react-dom.development.js:414:5)
    at executeDispatchesAndRelease (react-dom.development.js:3278:5)
    at executeDispatchesAndReleaseTopLevel (react-dom.development.js:3287:10)
    at forEachAccumulated (react-dom.development.js:3259:8)
```

Edit: fixed [here](https://github.com/Flagsmith/flagsmith/pull/4301/commits/50999354e3dd1bfa7a6ae464aa804e0b9d47f5c4)

frontend/common/stores/feature-list-store.ts
```diff
@@ -458,135 +465,216 @@ const controller = {
...
-          ...changeRequestData
-        } = changeRequest
 
-        const userApprovals = approvals.filter((u) => {
-          const keys = Object.keys(u)
-          return keys.includes('user')
-        })
+      let feature_states_to_create:
+          | Req['createFeatureVersion']['feature_states_to_create']
+          | undefined = undefined,
```
matthewelwell, line 494: I don't quite understand what's happening here?

frontend/web/components/ErrorMessage.js
```diff
@@ -12,6 +12,7 @@ export default class ErrorMessage extends PureComponent {
     const errorMessageClassName = `alert alert-danger ${
       this.props.errorMessageClass || 'flex-1 align-items-center'
     }`
+    console.log(this.props.error)
```
matthewelwell: Should we remove this now?

matthewelwell (approved):
I've tested this manually via staging (as that's where the API code is currently) and all looks good to me. 

I've added a couple of minor comments on certain aspects of the code, but nothing major.

frontend/common/stores/feature-list-store.ts
kyle-ssg: feature_states_to_create etc are assigned a value if versioning is enabled, otherwise it doesn't send anything (undefined)

frontend/web/components/ErrorMessage.js
kyle-ssg: Removed

frontend/common/stores/feature-list-store.ts
matthewelwell: gotcha, sorry, I understand now 👍

matthewelwell (approved):

PR #4304, opened 2024-07-08
docs: tweak limits

dabeeeenster:


docs/docs/system-administration/system-limits.md
```diff
@@ -38,3 +30,29 @@ Please contact us if you want to override the current system limits.
...
+
+### SDK Limits
+
+Requests made by our SDKs are _not_ rate limited, by design; we can't predict what sort of profile your traffic will
+look like.
+
+#### Free Plan Limits
+
+If you are on our Free Plan, we will stop answering requests from our API if you go over your
+[Free Plan limit](https://www.flagsmith.com/pricing). We will send you a warning email 7 days before we block your
```
rolodato, line 56: When is the block automatically cleared? Is it when the next billing period starts (or when the account is upgraded) or something else?

Also, who receives the email? Is it all org admins? There's also in-app notifications about this right?

docs/docs/system-administration/system-limits.md
```diff
@@ -38,3 +30,29 @@ Please contact us if you want to override the current system limits.
...
+requests, to give you time to upgrade your account.
+
+#### Paid Plan Limits
+
+For paid plans, we'll give you a 30 day grace period and then charge you at $50 per million API calls for a calendar
```
rolodato: What does the grace period mean exactly? I've been trying to put it into words but not sure how it's supposed to behave.

Do you ever get more grace periods, or is it only one in the organisation's lifetime?

docs/docs/system-administration/system-limits.md
```diff
@@ -38,3 +30,29 @@ Please contact us if you want to override the current system limits.
 
 You can modify the system limits on a per-Project basis. These limits are defined in the database against the Project.
 The easiest way to modify them is with the [Django admin](/deployment/configuration/django-admin.md) interface.
+
+## Traffic Limits
+
+### SDK Limits
+
+Requests made by our SDKs are _not_ rate limited, by design; we can't predict what sort of profile your traffic will
```
rolodato, line 46: I can work on rewording this, but requests are technically rate limited. The limit is per month/billing period and not a smaller unit of time like seconds or minutes, so I would leave this out to avoid confusion.

docs/docs/system-administration/system-limits.md
rolodato: Is there an option for customers to prevent auto-charges, and just default to blocking? This could get uncontrollably expensive during DDoS attacks or bugs that cause high API usage, and it's scary to have auto-charges without offering something like Vercel's [Spend Management](https://vercel.com/docs/pricing/spend-management) (hard budget caps, among other things)

docs/docs/system-administration/system-limits.md
dabeeeenster: Hmm yeah -  I meant "not rate limited in the short term" sort of thing...

docs/docs/system-administration/system-limits.md
dabeeeenster: @matthewelwell or @zachaysan  to answer these? I'm not sure!

docs/docs/system-administration/system-limits.md
dabeeeenster: No - we cant react that quickly AFAIUI cc @matthewelwell or @zachaysan too

docs/docs/system-administration/system-limits.md
dabeeeenster: I think people understand rate limiting to be sort of "near real time" - going over a monthly limit I wouldnt regard as "rate limiting"

rolodato: Yes, I understood what you meant here. My point is that it's not necessary to mention that there are no short-term limits if customers don't need to do anything about that, which also helps prevent confusion between short-term and long-term limits.

docs/docs/system-administration/system-limits.md
matthewelwell: > When is the block automatically cleared?

30 days after the warning email was first sent, or when they upgrade. 

> Also, who receives the email?

Org admins receive the email if the usage is <100%, all users in the organisation receive it if usage is >= 100%

> There's also in-app notifications about this right?

Yes, there is a warning at the top of the dashboard showing the API usage.

docs/docs/system-administration/system-limits.md
matthewelwell: I don't actually know the answer to this. From looking at the code, it looks like there is no provision in the code for this but I think it needs updating. 

The behaviour should be such that everyone gets one month's grace (assuming their usage is below 200%). If they exceed the usage in any subsequent months, or the usage is greater than 200% in the first month then we charge them for all overages >100% (yes, that includes the first month if they exceed 200%, we will still charge for everything above 100%).

matthewelwell: No, we don't have any provisions set for this. We could perhaps add it, but it would be separate to this behaviour.

docs/docs/system-administration/system-limits.md
zachaysan: For free plans we do rate limit after a warning and grace period, so this should be added to this section.

zachaysan: Oh I see we've added it below. Could be still better to mention up here.

docs/docs/system-administration/system-limits.md
zachaysan: Yeah what Matt said covers everything.

matthewelwell (approved):

PR #4308, opened 2024-07-09
feat(pg-usage-data): Add cache to batch tracking data

### LINKED ISSUE #4241: Batch API usage raw write to reduce the load on database
More information here: https://flagsmith.slack.com/archives/C02HPR34L65/p1719305729415889?thread_ts=1719206868.235169&cid=C02HPR34L65

gagantrivedi:

## Changes

Add cache to batch usage tracking
fixes: https://github.com/Flagsmith/flagsmith/issues/4241

Adds unit tests


api/tests/unit/app_analytics/test_cache.py
matthewelwell, line 1: Let's update the name of the file to `test_unit_app_analytics_cache.py`

api/tests/unit/app_analytics/test_cache.py
```diff
@@ -0,0 +1,90 @@
...
+                        "host": host,
+                        "environment_key": environment_key_2,
+                        "count": 10,
+                    }
+                )
+            )
+        mocked_track_request_task.delay.assert_has_calls(expected_calls)
+
+
+def test_api_usage_cache__track_request_calls_flush_only_after_cache_flush_interval(
```
matthewelwell: Do we need this test at all. Doesn't the above test cover it?

gagantrivedi: Yeah, it needs a better name: it tests that flush resets the interval and the cache dict

matthewelwell: Ah, ok. Then let's rename it and I'll approve.

gagantrivedi: Done

api/tests/unit/app_analytics/test_unit_app_analytics_cache.py
```diff
@@ -0,0 +1,102 @@
...
+
+    with freeze_time(now) as frozen_time:
+        # Add some data
+        cache.track_request(Resource.FLAGS, "host", "environment_key")
+
+        # move the time forward
+        frozen_time.tick(CACHE_FLUSH_INTERVAL + 1)
+
+        # Next, let's call flush
+        cache.flush()
```
matthewelwell: I think we have this pattern elsewhere so we're probably just following that, but it's odd to me that we're calling `flush` directly here since I don't think any external calls are made to it. I think we should probably mark the `flush` method as private, and this test should only use the `track_request` method.

api/tests/unit/app_analytics/test_middleware.py
```diff
@@ -12,13 +15,56 @@
...
+        ("/api/v1/environment-document", Resource.ENVIRONMENT_DOCUMENT),
+    ],
+)
+def test_APIUsageMiddleware_calls_track_request_correctly_without_cache(
+    rf: RequestFactory,
+    mocker: MockerFixture,
+    path: str,
+    enum_resource_value: int,
+    settings: SettingsWrapper,
+):
```
zachaysan: nit: Missing typing for return value

api/tests/unit/app_analytics/test_middleware.py
```diff
@@ -12,13 +15,56 @@
...
 )
-def test_APIUsageMiddleware_calls_track_request_correctly(
-    rf, mocker, path, enum_resource_value
+def test_APIUsageMiddleware_calls_track_request_correctly_with_cache(
+    rf: RequestFactory,
+    mocker: MockerFixture,
+    path: str,
+    enum_resource_value: int,
+    settings: SettingsWrapper,
 ):
```
zachaysan: Nit: Missing typing for return value

api/tests/unit/app_analytics/test_middleware.py
```diff
@@ -39,13 +85,14 @@ def test_APIUsageMiddleware_calls_track_request_correctly(
 
 
 def test_APIUsageMiddleware_avoids_calling_track_request_if_resoure_is_not_tracked(
-    rf, mocker
+    rf: RequestFactory, mocker: MockerFixture, settings: SettingsWrapper
 ):
```
zachaysan: Nit: Missing typing for return value

api/tests/unit/app_analytics/test_unit_app_analytics_cache.py
```diff
@@ -0,0 +1,102 @@
+from app_analytics.cache import CACHE_FLUSH_INTERVAL, APIUsageCache
+from app_analytics.models import Resource
+from django.utils import timezone
+from freezegun import freeze_time
+from pytest_mock import MockerFixture
+
+
+def test_api_usage_cache(mocker: MockerFixture):
```
zachaysan: Nit: Missing typing for return value

api/tests/unit/app_analytics/test_unit_app_analytics_cache.py
```diff
@@ -0,0 +1,102 @@
...
+                        "count": 10,
+                    }
+                )
+            )
+        mocked_track_request_task.delay.assert_has_calls(expected_calls)
+
+
+def test_api_usage_cache__flush_resets_cache_and_interval(
+    mocker: MockerFixture,
+):
```
zachaysan: Nit: Missing typing for return value

zachaysan (approved):
Overall looks good to me, just a bunch of typing suggestions.

api/tests/unit/app_analytics/test_unit_app_analytics_cache.py
gagantrivedi: Although It's not being used, but I feel it makes the interface a bit more clean? Does that make sense?

matthewelwell: Hmm... maybe it does, yes, but the test feels a bit too 'internal' to me. It's not really testing the actual behaviour.

gagantrivedi: Yeah, that's a fair point! I have made it private

api/tests/unit/app_analytics/test_middleware.py
gagantrivedi: Done

matthewelwell (approved):

PR #4316, opened 2024-07-09
fix: incorrect negative value conversion

### LINKED ISSUE #4197: Negative number flag values are set to strings after updating

- [ ] Self Hosted with Docker
- [X] SaaS at flagsmith.com

### Describe the bug

See steps below.

### Steps To Reproduce

1. Create a new feature with a value of `-1`. Its value will be correctly saved as an integer
2. Update the feature; erase and type in `-1` again, and save
3. The value will be incorrectly saved as the string `"-1"`

### Expected behavior

Valid integers should always be saved as integers.

### Screenshots

https://github.com/Flagsmith/flagsmith/assets/829698/86d39885-f314-4c6f-b0ba-d4aee99494da

kyle-ssg:

## Changes

Negative ints were being sent as strings to the api when saving a feature.

Saved a feature with the value -1

### DIFF
--- a/frontend/web/components/Feature.js
+++ b/frontend/web/components/Feature.js
@@ -135,10 +135,10 @@ export default class Feature extends PureComponent {
             warningMessage={
               <div>
                 This feature currently has the value of{' '}
-                <strong>"{environmentFlag?.feature_state_value}"</strong> as a
-                string. As of version <strong>2.137.0</strong>, saving this
-                feature will convert its value from a string to a number. If you
-                wish to preserve this value as a string, please save it via our{' '}
+                <strong>"{environmentFlag?.feature_state_value}"</strong>.
+                Saving this feature will convert its value from a string to a
+                number. If you wish to preserve this value as a string, please
+                save it via our{' '}
                 <a href='https://api.flagsmith.com/api/v1/docs/#/api/api_v1_environments_featurestates_update'>
                   API
                 </a>
--- a/frontend/web/components/Feature.js
+++ b/frontend/web/components/Feature.js
@@ -138,7 +138,7 @@ export default class Feature extends PureComponent {
                 <strong>"{environmentFlag?.feature_state_value}"</strong> as a
                 string. As of version <strong>2.137.0</strong>, saving this
                 feature will convert its value from a string to a number. If you
-                wish to preserve this value as a string, please save it via our
+                wish to preserve this value as a string, please save it via our{' '}
                 <a href='https://api.flagsmith.com/api/v1/docs/#/api/api_v1_environments_featurestates_update'>
                   API
                 </a>
--- a/frontend/web/components/Feature.js
+++ b/frontend/web/components/Feature.js
@@ -7,10 +7,31 @@ import AddVariationButton from './mv/AddVariationButton'
 import ErrorMessage from './ErrorMessage'
 import Tooltip from './Tooltip'
 import Icon from './Icon'
+import Select from 'react-select'
+import InputGroup from './base/forms/InputGroup'
+import WarningMessage from './WarningMessage'
+function isNegativeNumberString(str) {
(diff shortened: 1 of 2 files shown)


matthewelwell (dismissed):

matthewelwell (commented):
Hmm... I tend to agree with this change... but I have a feeling it's a breaking change, right? If people re-save a feature with a negative value it will immediately start returning as an integer in the SDKs. 

For reference, I've just checked the production database and we have ~60 such cases.

kyle-ssg:
Hmm yeah, we've fixed this conversion before but I think that only affected people experiencing an issue. It depends on what language SDK they are using (we should maybe track this in headers).

JavaScript will actually work with both, because JavaScript.

rolodato:
> For reference, I've just checked the production database and we have ~60 such cases.

@matthewelwell Could you share that dataset/query? I'd like to evaluate our options to move forward based on that.

matthewelwell:
> > For reference, I've just checked the production database and we have ~60 such cases.
> 
> @matthewelwell Could you share that dataset/query? I'd like to evaluate our options to move forward based on that.

I've done this in Slack FYI @rolodato .

rolodato:
I've reviewed the data and there's about 10 organisations, viewable here: https://metabase.flagsmith.com/question/9-features-with-negative-numbers-as-strings. Some of them do have these values set in production so we can't just ship this change as-is, unfortunately.

I can think of a few paths forward:

1. Abandon this PR and ignore this issue until we can make feature value types explicit, which sidesteps this problem entirely (assuming this is planned/desired)
2. Add more functionality to this PR to enable users affected by this issue to migrate to the new behaviour at their own pace:
  2.1. If a feature is currently a negative number string and a user tries to update it to a negative number, offer a choice to save the value as string (default) or as integer (secondary choice)
  2.2. If someone does need to create a new feature with a negative string number, they can use the API manually.
3. Add more functionality to this PR to display a warning banner when editing a feature with a negative number string, saying that saving changes to the value will change its type to integer, with instructions on how to manually save the value as string if needed.
4. Set some deadline in the future and notify affected customers that they need to migrate before we ship this PR as-is. CS can create documentation on how SaaS customers need to migrate, how they can tell if they're affected, etc. For on-prem Enterprise customers, we can let them know before they upgrade. On-prem Free users are a bit more tricky to find and notify.

IMO, option 2 is ideal UX until if/when we ship explicit feature value types, but option 3 might be a better effort/reward compromise based on how few SaaS users are actually affected by this. Option 3 is not as much implementation effort and is less risky (no user interaction or manipulating feature values on behalf of users). I'll leave this decision to @kyle-ssg since you'll probably write the code for the actual solution :)

kyle-ssg:
I've been reviewing/attempting 2 and I think it's a big change with lots of code considering the amount of people this will temporarily affect. Especially when this considers change requests and feature versioning.

I'd be inclined to go with 3.

kyle-ssg:
Ok yep after discussion this is now done, 
image

rolodato (approved):

frontend/web/components/Feature.js
```diff
@@ -102,13 +123,30 @@ export default class Feature extends PureComponent {
...
+        {this.state.isNegativeNumberString && (
+          <WarningMessage
+            warningMessage={
+              <div>
+                This feature currently has the value of{' '}
+                <strong>"{environmentFlag?.feature_state_value}"</strong>.
+                Saving this feature will convert its value from a string to a
+                number. If you wish to preserve this value as a string, please
+                save it via our{' '}
+                <a href='https://api.flagsmith.com/api/v1/docs/#/api/api_v1_environments_featurestates_update'>
```
rolodato, line 142: Minor nit - suggest wording this as `please save it using the Admin API` for consistency with other docs. LGTM, good stuff 🙏

PR #4329, opened 2024-07-12
fix: frontend fails to load when announcement flag isn't set

matthewelwell:
## Changes

Fixes an error which prevents the FE from loading when the `announcement` flag isn't set. 

Ran the FE locally pointing to a flagsmith on flagsmith environment that doesn't have the `announcement` flag set.

### DIFF
--- a/frontend/web/components/Announcement.tsx
+++ b/frontend/web/components/Announcement.tsx
@@ -20,6 +20,11 @@ const Announcement: FC<AnnouncementType> = () => {
   }
 
   const announcementValue = Utils.getFlagsmithJSONValue('announcement', null)
+
+  if (!announcementValue) {
+    return null
+  }
+
   const { buttonText, description, id, isClosable, title, url } =
     announcementValue as AnnouncementValueType
   const dismissed = flagsmith.getTrait('dismissed_announcement')


frontend/web/components/Announcement.tsx
```diff
@@ -20,32 +20,37 @@ const Announcement: FC<AnnouncementType> = () => {
...
-        >
-          <div>
-            <div>{description}</div>
-          </div>
-        </InfoMessage>
-      )}
-    </>
-  )
+
+  if (announcementValue) {
```
kyle-ssg: I think this might be better if we just did 

```
if(!announcementValue) {
  return null
}
```
Above this

matthewelwell: Well, I tried that (or similar at least) but then my IDE complained that I shouldn't be returning `null` from the component.

kyle-ssg: Hmm no null is valid, maybe a strange ide typescript setup 
image

matthewelwell: Ok, I'll just do this then 👍

matthewelwell: Done

novakzaballa (approved):

PR #4338, opened 2024-07-15
feat: Send users notification when api flags have been blocked

zachaysan:
## Changes

Previously, users using the free Flagsmith plan were blocked from using the feature flagging API once they had breach the limit and grace period. With this change, users now get an email notification that their usage is now being blocked.

Added tests for the functionality to an existing long test which covered multiple of the possible scenarios.


api/organisations/tasks.py
```diff
@@ -105,6 +105,25 @@ def finish_subscription_cancellation() -> None:
         subscription.save_as_free_subscription()
 
 
+def send_api_flags_blocked_notification(organisation: Organisation) -> None:
```
matthewelwell: I'd rather keep the `tasks` module reserved for tasks. This module is already _slightly_ confused by the fact that we conditionally add these tasks, but I think with this one we should either: 

 1. Move the code to another module
 2. (probably the easiest option, but makes testing slightly harder) Make this function private

api/organisations/templates/organisations/api_flags_blocked_notification.txt
```diff
@@ -0,0 +1,7 @@
+Hi there,
+
+As per previous warnings, we have blocked organisation {{ organisation.name }}. Please visit app.flagsmith.com to upgrade your account and re-enable service.
+
+Thank you!
+
+The Flagsmith Team
```
matthewelwell: Note that this template will likely need updating as per [this conversation](https://docs.google.com/document/d/1nqNj5kdqCbcyc29v0FmMORt0eZCY5hIxgzwxJLttPi8/edit?disco=AAABRskXWjg).

matthewelwell:
```suggestion
Hi there,

This is a system generated notification related to your Flagsmith API Usage. As per previous warnings, we have had to blocked your companyorganisation {{ organisation.name }} after the 7 day grace period. Flags are not currently being served for your organization, and will continue to be blocked until your billing period resets or you upgrade your account. You can upgrade your account at app.flagsmith.com.

Thank you!

The Flagsmith Team
```

(Note that this will also need to be changed in the HTML template as well)

zachaysan: Ok this has been added

api/organisations/tasks.py
zachaysan: As discussed, moved this and other functions into a `task_helpers.py` module

api/organisations/tasks.py
```diff
@@ -47,6 +42,10 @@
     STARTUP_V2,
     SubscriptionCacheEntity,
 )
+from .task_helpers import (
+    _handle_api_usage_notifications,
```
matthewelwell: This shouldn't still be a private function since we're calling it from another module now.

api/organisations/task_helpers.py
```diff
@@ -0,0 +1,137 @@
...
+        subject="Flagsmith API use has been blocked due to overuse",
+        message=render_to_string(message, context),
+        from_email=settings.DEFAULT_FROM_EMAIL,
+        recipient_list=list(recipient_list.values_list("email", flat=True)),
+        html_message=render_to_string(html_message, context),
+        fail_silently=True,
+    )
+
+
+def send_api_usage_notification(
```
matthewelwell: This probably _can_ be a private function now?

matthewelwell (changes requested):

api/organisations/tasks.py
zachaysan: Ok I've renamed it

api/organisations/task_helpers.py
zachaysan: Sure, I've renamed it.

matthewelwell (approved):

PR #4340, opened 2024-07-15
feat: Support Aptible deployments

rolodato:

## Changes

This PR adds functionality to enable deploying to Aptible, which an Enterprise prospect is currently evaluating. They have been able to work around these changes by forking open-source Flagsmith and making similar changes (i.e. removing `waitfordb` instead of making it optional). They would however need the changes to be upstreamed to deploy the Flagsmith Enterprise image.

### Add `/healthcheck` as a health check route

Exposes the same health check endpoint on `/healthcheck` in addition to `/health`. Aptible (surprisingly) does not allow configuring a custom health check route and forces it to be `/healthcheck`. [Strict health checks](https://www.aptible.com/docs/core-concepts/apps/connecting-to-apps/app-endpoints/https-endpoints/health-checks) also do not accept redirects as responses.

There are some alternatives to directly adding `/healthcheck` as an alias of `/health`:

1. Add an option to make the health check route configurable. This would require a more thorough documentation review and is potentially error prone. YAGNI also applies since we have never needed to customise the health check route at the API level (i.e. not using a proxy in front of it) until now.
2. Add an option to opt in to the `/healthcheck` route, using some sort of "Aptible mode". I don't see a downside to always aliasing `/health` so this seems like extra complexity for no benefit.
3. Add an option for additional configurable health check routes, i.e. in addition to `/health`. I believe YAGNI also applies here and we can add it later if needed.

### Add `SKIP_WAIT_FOR_DB` environment variable

When Aptible is deploying an application, it first checks for successful health checks before it is added to the same network as an Aptible-hosted database. This means that trying to deploy Flagsmith to Aptible without these changes will always fail when health checks fail while `waitfordb` is hanging.

This PR adds an environment variable that skips waiting for the database before running migrations or starting the API.

It seemed to me that an environment variable is the least intrusive and Aptible-specific way we could support this - happy to consider different approaches and variable/parameter names.

Deployed Flagsmith to a trial Aptible account here using the steps described in the docs from this PR: https://app-76164.on-aptible.com/. I'll add a link in the docs to the actual Flagsmith release that contains these changes once ready.

_Please describe._

### DIFF
--- a/api/app/urls.py
+++ b/api/app/urls.py
@@ -14,8 +14,8 @@ urlpatterns = [
     re_path(r"^api/v1/", include("api.urls.v1", namespace="api-v1")),
     re_path(r"^api/v2/", include("api.urls.v2", namespace="api-v2")),
     re_path(r"^admin/", admin.site.urls),
+    re_path(r"^health", include("health_check.urls", namespace="health")),
     path("healthcheck", include("health_check.urls", namespace="health")),
-    re_path(r"^health/", include("health_check.urls", namespace="health")),
     re_path(r"^version", views.version_info, name="version-info"),
     re_path(
         r"^sales-dashboard/",
--- a/api/app/urls.py
+++ b/api/app/urls.py
@@ -15,6 +15,7 @@ urlpatterns = [
     re_path(r"^api/v2/", include("api.urls.v2", namespace="api-v2")),
     re_path(r"^admin/", admin.site.urls),
     path("healthcheck", include("health_check.urls", namespace="health")),
+    re_path(r"^health/", include("health_check.urls", namespace="health")),
     re_path(r"^version", views.version_info, name="version-info"),
     re_path(
         r"^sales-dashboard/",
--- a/api/app/urls.py
+++ b/api/app/urls.py
@@ -15,6 +15,8 @@ urlpatterns = [
     re_path(r"^api/v2/", include("api.urls.v2", namespace="api-v2")),
     re_path(r"^admin/", admin.site.urls),
     re_path(r"^health", include("health_check.urls", namespace="health")),
+    # Aptible health checks must be on /healthcheck and cannot redirect
+    # see https://www.aptible.com/docs/core-concepts/apps/connecting-to-apps/app-endpoints/https-endpoints/health-checks
     path("healthcheck", include("health_check.urls", namespace="health")),
     re_path(r"^version", views.version_info, name="version-info"),
     re_path(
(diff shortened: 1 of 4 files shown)


dabeeeenster:
Annoying about the healthcheck, but I don't see an issue with just adding another route.

docs/docs/deployment/hosting/aptible.md
```diff
@@ -0,0 +1,49 @@
+---
+title: Aptible
+---
+
+## Prerequisites
+
+The options and health check routes described in this document are available from Flagsmith 2.130.0.
```
dabeeeenster, line 7:
```suggestion
The options and health check routes described in this document are available from Flagsmith `2.130.0`.
```

dabeeeenster (approved):

api/app/urls.py
```diff
@@ -15,6 +15,7 @@
     re_path(r"^api/v2/", include("api.urls.v2", namespace="api-v2")),
     re_path(r"^admin/", admin.site.urls),
     re_path(r"^health", include("health_check.urls", namespace="health")),
+    path("healthcheck", include("health_check.urls", namespace="health")),
```
matthewelwell, line 20: Maybe we should add a short comment explaining that this was added to support aptible deployments?

matthewelwell (approved):
All seems reasonable to me. Add one minor comment.

khvn26 (approved):

docs/docs/deployment/hosting/aptible.md
rolodato: There's no right or wrong way to do this, but I prefer the advice from Google's technical documentation style guide - version numbers are typically not styled in code font. See https://developers.google.com/style/code-in-text

api/app/urls.py
rolodato: Added in e78cf072b130b9312e99c823525b18a89727c621 ✅

matthewelwell (approved):

rolodato:
Thanks all for the reviews! I've shared this with the prospect and am waiting to get their feedback before merging.

rolodato:
Got feedback from the prospect that this is working for them. The only thing that I haven't got working yet on Aptible is the task processor. It is possible to run it but need to keep experimenting with different settings.

matthewelwell:
> Got feedback from the prospect that this is working for them. The only thing that I haven't got working yet on Aptible is the task processor. It is possible to run it but need to keep experimenting with different settings.

Should we mark this PR as draft until this is done then? Or merge this and submit another PR for the task processor?

dabeeeenster:
I vote to merge and then additional PRs if there's a need

PR #4363, opened 2024-07-22
docs: Transient Traits and Identities

dabeeeenster:


docs/docs/advanced-use/transient-traits.md
```diff
@@ -0,0 +1,587 @@
...
+
+## Overview and Use Cases
+
+Transient Identities and Traits are particularly useful when you need to run flag evaluations without storing specific
+data in Flagsmith. Below are common scenarios where transient Traits or Identities come in handy.
+
+### Anonymous Identities
+
+If you choose to provide an empty or blank (`""`) identifier is provided, Flagsmith will automatically treat the
+Identity as transient. In this case, an auto-generated identifier is returned, based on a hash of the provided Traits.
```
matthewelwell: What if no traits are provided?

docs/docs/advanced-use/transient-traits.md
```diff
@@ -0,0 +1,587 @@
...
+});
+let identifierToUseLater = flagsmith.getContext().identity.identifier;
+```
+
+This approach is useful for scenarios such as A/B testing in e-commerce, where you want to include users who haven't
+registered or logged in.
+
+:::info
+
+Server-generated transient identifiers are consistent across flag evaluations, meaning the same identifier will be
```
matthewelwell: I'm a little unsure what 'server-generated' means here - is it relating to the Flagsmith API, or server side clients (or both)?. I think the confusion stems from giving the example only in JS above, but then talking about server-generated in the note below. Can we clarify this any further?

docs/docs/advanced-use/transient-traits.md
```diff
@@ -0,0 +1,587 @@
...
+
+If you want to switch the `screenOrientation` to `portrait` for a specific session without overwriting the existing
+value, you can mark the Trait as transient:
+
+```javascript
+flagsmith.setTrait('screenOrientation', { value: 'portrait', transient: true });
+```
+
+In this case, `portrait` will be used for one specific flag evaluation, but the stored `landscape` value will remain
+unchanged.
```
matthewelwell: > one specific flag evaluation

I guess this will vary for different SDKs but this is a little unclear to me. Can we reword it to e.g. 'the next flag evaluation in the SDK, subsequent evaluations after that will use the previously stored value?' - is that even true? Wouldn't it remain as the transient value for the remainder of the 'session' ?

docs/docs/advanced-use/transient-traits.md
```diff
@@ -0,0 +1,587 @@
...
+        true,
+    ),
+    SDKTrait::new(
+        "my_other_key".to_string(),
+        FlagsmithValue {
+            value: "true".to_string(),
+            value_type: FlagsmithValueType::Bool,
+        },
+        true,
+    ),
```
matthewelwell: I'm slightly confused here. It looks like we're using 2 different constructors, but then also sending a third argument to both constructors - which I assume is the `transient` argument?

docs/docs/advanced-use/transient-traits.md
```diff
@@ -0,0 +1,587 @@
...
+            value_type: FlagsmithValueType::Float,
+        },
+    ),
+    SDKTrait::new(
+        "my_other_key".to_string(),
+        FlagsmithValue {
+            value: "true".to_string(),
+            value_type: FlagsmithValueType::Bool,
+        },
+        true,
```
matthewelwell: Maybe I'm wrong above - what is this argument for?

matthewelwell (commented):
Looks good overall, but I've added a few comments / questions.

docs/docs/advanced-use/transient-traits.md
khvn26: Typo; this is fixed.

docs/docs/advanced-use/transient-traits.md
khvn26: This was incorrect; fixed.

docs/docs/advanced-use/transient-traits.md
khvn26: > the next flag evaluation in the SDK, subsequent evaluations after that will use the previously stored value

TBH this sounds more confusing to me than what we have now

> Wouldn't it remain as the transient value for the remainder of the 'session' ?

How so? What even constitutes a session when we're talking about all possible SDKs?

docs/docs/advanced-use/transient-traits.md
khvn26: Reworded.

docs/docs/advanced-use/transient-traits.md
khvn26: Added a clarification.

docs/docs/advanced-use/transient-traits.md
matthewelwell: > > the next flag evaluation in the SDK, subsequent evaluations after that will use the previously stored value
> 
> TBH this sounds more confusing to me than what we have now
>
> > Wouldn't it remain as the transient value for the remainder of the 'session' ?
> 
> How so? What even constitutes a session when we're talking about all possible SDKs?

Well, my biggest issue is that the example you've given is using the `setTrait` method in the JS SDK. This doesn't directly retrieve the flags so I have no idea how it works. In other SDKs, it's pretty clear because you provide the traits directly when retrieving the flags.

docs/docs/advanced-use/transient-traits.md
```diff
@@ -0,0 +1,585 @@
...
+Traits and Identities as transient — meaning they are used for flag evaluation but are not stored.
+
+## Overview and Use Cases
+
+Transient Identities and Traits are particularly useful when you need to run flag evaluations without storing specific
+data in Flagsmith. Below are common scenarios where transient Traits or Identities come in handy.
+
+### Anonymous Identities
+
+If you choose to provide an empty or blank (`""`) identifier is provided, Flagsmith will automatically treat the
```
dabeeeenster: this doesnt make sense? 

> an empty or blank identifier is provided

docs/docs/advanced-use/transient-traits.md
```diff
@@ -0,0 +1,585 @@
...
+with an Identity to evaluate flags. This default behavior works for most use cases. However, there are scenarios where
+you may want more control over which Traits or Identities are stored. Using the Flagsmith API and SDKs, you can mark
+Traits and Identities as transient — meaning they are used for flag evaluation but are not stored.
+
+## Overview and Use Cases
+
+Transient Identities and Traits are particularly useful when you need to run flag evaluations without storing specific
+data in Flagsmith. Below are common scenarios where transient Traits or Identities come in handy.
+
+### Anonymous Identities
```
dabeeeenster, line 39: I feel like it would be better to start with "Skipping Personal Identifiable Information" as it doesn't introduce the blank identifier but just a blank trait, which is simpler to understand.

docs/docs/advanced-use/transient-traits.md
```diff
@@ -0,0 +1,585 @@
...
+
+This approach is useful for scenarios such as A/B testing in e-commerce, where you want to include users who haven't
+registered or logged in.
+
+:::info
+
+Server-generated transient identifiers are consistent across flag evaluations, meaning the same identifier will be
+returned for identical sets of Traits.
+
+To generate truly unique identifiers, include additional unique Trait values.
```
dabeeeenster, line 67: Do we have an answer to the question "What happens when the user authenticates" yet or not?

docs/docs/advanced-use/transient-traits.md
khvn26: Fair enough. I've tried to improve the wording a bit.

docs/docs/advanced-use/transient-traits.md
khvn26: The answer would be "nothing, use the identifier returned to you by Flagsmith".

docs/docs/advanced-use/transient-traits.md
khvn26: Changed to "null or blank".

docs/docs/advanced-use/transient-traits.md
khvn26: Agreed. Changed the order.

docs/docs/advanced-use/transient-traits.md
```diff
@@ -14,14 +14,35 @@ Traits and Identities as transient — meaning they are used for flag evaluation
...
+
+After making the SDK call, if you check the Identities tab, you'll see the `test-user-with-transient-email` Identity
+without the `email` Trait being stored.
+
 ### Anonymous Identities
 
-If you choose to provide an empty or blank (`""`) identifier is provided, Flagsmith will automatically treat the
-Identity as transient. In this case, an auto-generated identifier is returned, based on a hash of the provided Traits.
-This ensures consistent flag evaluations without persisting the Identity in the Flagsmith dashboard.
+If you choose to provide a blank (`""`) or null identifier is provided, Flagsmith will automatically treat the Identity
```
matthewelwell:
```suggestion
If you choose to provide a blank (`""`) or null identifier, Flagsmith will automatically treat the Identity
```

matthewelwell (approved):

Issue #4278, opened 2024-07-03
Implement transient traits and identities


khvn26:
### Goal

To support additional use cases, such as evaluating identities in different contexts or managing anonymous identities, we aim to enable our SDK users to:

1. Evaluate flags against identities without persisting them.
2. Evaluate flags against identities while persisting only a subset of identity traits.

### Solution Proposal

Make /api/v1/identities endpoints support optional boolean `"transient"` attributes both for the top-level identity object and a trait object.

Don't persist identities and traits marked as transient.

In SDKs, support `transient: bool` and `transientTraits: list[str]` optional keywords for the `getIdentityFlags` interface.

### Related issues and discussions
Customer requests and use case discussions:
- #168, #77
- #1840, #1839 

API tasks:
- https://github.com/Flagsmith/flagsmith/issues/4279
- https://github.com/Flagsmith/flagsmith/issues/4448
- https://github.com/Flagsmith/edge-api/issues/271
- https://github.com/Flagsmith/edge-api/issues/275

SDK tasks:
- https://github.com/Flagsmith/flagsmith-python-client/issues/88
- https://github.com/Flagsmith/flagsmith-js-client/issues/236
- https://github.com/Flagsmith/flagsmith-kotlin-android-client/issues/52
- https://github.com/Flagsmith/flagsmith-flutter-client/issues/62
- https://github.com/Flagsmith/flagsmith-php-client/issues/74
- https://github.com/Flagsmith/flagsmith-elixir-client/issues/35
- https://github.com/Flagsmith/flagsmith-ruby-client/issues/63
- https://github.com/Flagsmith/flagsmith-go-client/issues/132
- https://github.com/Flagsmith/flagsmith-rust-flag-engine/issues/14 + https://github.com/Flagsmith/flagsmith-rust-client/issues/22
- https://github.com/Flagsmith/flagsmith-java-client/issues/152
- https://github.com/Flagsmith/flagsmith-dotnet-client/issues/104
- https://github.com/Flagsmith/flagsmith-nodejs-client/issues/156
- https://github.com/Flagsmith/flagsmith-ios-client/issues/63

novakzaballa:
@khvn26 I understand that we should implement transient traits because it is a quick and cost-effective way to cover some of the use cases related to contexts. While it could also help with anonymous identities or other use cases, we should keep the scope restricted to transient traits since that is the reason we decided to move ahead with its implementation.
[...]

PR #4375, opened 2024-07-23
feat: Return transient traits explicitly

khvn26:

## Changes

As discussed offline, this makes the `/api/v1/identities` SDK route return transient traits explicitly. This helps simplify SDK caching logic — i.e. see [this commit](https://github.com/Flagsmith/flagsmith-js-client/pull/240/commits/628e45b3afe9cc06b80c3059b54c907b8a34382e#diff-0b9b511548a3e7571b5371b6ad98facf6955df9dfa855af7107b182e04bc524bL114-L115) to JS SDK.

Contributes to #4278.

Modified an existing unit test.

### DIFF
--- a/api/environments/identities/traits/models.py
+++ b/api/environments/identities/traits/models.py
@@ -58,8 +58,11 @@ class Trait(models.Model):
         return self.get_trait_value()
 
     @property
-    def transient(self):
-        return not self.pk
+    def transient(self) -> bool:
+        try:
+            return self._is_transient
+        except AttributeError:
+            return self.pk is None
 
     def get_trait_value(self):
         try:
--- a/api/environments/identities/traits/models.py
+++ b/api/environments/identities/traits/models.py
@@ -59,10 +59,11 @@ class Trait(models.Model):
 
     @property
     def transient(self) -> bool:
-        try:
-            return self._is_transient
-        except AttributeError:
-            return self.pk is None
+        return getattr(self, "_transient", False)
+
+    @transient.setter
+    def transient(self, transient: bool) -> None:
+        self._transient = transient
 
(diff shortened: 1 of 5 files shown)


api/environments/identities/traits/models.py
```diff
@@ -57,6 +57,13 @@ def natural_key(self):
     def trait_value(self):
         return self.get_trait_value()
 
+    @property
+    def transient(self) -> bool:
```
matthewelwell, line 61: I think some documentation on this property will probably help in the future.

api/environments/identities/traits/models.py
```diff
@@ -57,6 +57,13 @@ def natural_key(self):
...
+        try:
+            return self._is_transient
+        except AttributeError:
+            return self.pk is None
```
matthewelwell: Possible not an issue, but I'm a _little_ concerned that this will bleed into other logic. A trait that just hasn't been saved yet isn't really _transient_, right? But it would be according to this logic. 

Maybe this should actually only be defined as a `SerializerMethodField`? (and maybe `_is_transient` should be a public attribute?)

khvn26: I've reimplemented the property to be more cleaner and decoupled it from model state.

matthewelwell (approved):

api/environments/identities/traits/models.py
matthewelwell: Nice use of a `setter` 👌

api/environments/identities/traits/models.py
matthewelwell: I think this is probably fine now as the code is cleaner.

PR #4383, opened 2024-07-24
fix: feature segments created with priority 0 are sent to bottom

matthewelwell:
## Changes

The previous PR [here](https://github.com/Flagsmith/flagsmith/pull/4381) actually causes an issue because the FE always defaults to sending the priority as 0. 

This PR is in part accounting for that, but also accounting for the fact that the `django-ordered-model` library that we are using doesn't enforce uniqueness on the priority so, when the priority is set in the save, any subsequent calls to `to()` don't work because it early returns that it already exists at that priority. The issue is that we now likely have 2 feature segments at that priority. To resolve this, we first move the 'collision' feature segment down and out of the way before completing the save. 

Added a bunch more tests for all scenarios.

### DIFF
--- a/api/features/feature_segments/serializers.py
+++ b/api/features/feature_segments/serializers.py
@@ -1,8 +1,7 @@
-import typing
-
 from common.features.serializers import (
     CreateSegmentOverrideFeatureSegmentSerializer,
 )
+from django.db import transaction
 from rest_framework import serializers
 from rest_framework.exceptions import PermissionDenied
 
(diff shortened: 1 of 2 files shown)


api/features/feature_segments/serializers.py
```diff
@@ -49,17 +48,33 @@ class CustomCreateSegmentOverrideFeatureSegmentSerializer(
...
+                    feature=kwargs["feature"],
+                    priority=priority,
+                ).first()
+            )
+            is not None
+        ):
+            # Since there is no unique clause on the priority field, if a priority
+            # is set, it will just save the feature segment and not move others
+            # down. This ensures that the incoming priority space is 'free'.
+            collision.to(priority + 1)
```
khvn26: Why not just sort all segments?

khvn26: As discussed elsewhere, the `collision.to(priority + 1)`, in fact, does this.

khvn26 (approved):
The new tests look exhaustive enough to me. 👍

PR #4397, opened 2024-07-26
docs: better openapi spec

dabeeeenster:
- Moved docusaurus to Typescript
- Move the API auto-generation to https://github.com/PaloAltoNetworks/docusaurus-openapi-docs


docs/static/openapi/edge-api.yaml
```diff
@@ -0,0 +1,225 @@
+openapi: 3.0.3
```
kyle-ssg: Do we need a note saying this is for interaction with our SDK endpoints / link to admin API swagger for other functions? I think as a first time user clicking API from docs it looks like these are the only apis we offer.

dabeeeenster: I've added an overview page that makes things clearer here

docs/static/openapi/edge-api.yaml
```diff
@@ -0,0 +1,225 @@
...
+          type: array
+          $ref: '#/components/schemas/FeatureState'
+    FeatureState:
+      type: object
+      properties:
+        enabled:
+          type: boolean
+        feature_state_value:
+          type: Can be any of the following - string, integer, boolean, float
+          example: 'a-remote-config-value'
```
khvn26: Strictly, this should be a [`oneOf`](https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/).

docs/static/openapi/edge-api.yaml
```diff
@@ -0,0 +1,225 @@
...
+      tags: ['Edge API']
+      summary: Get flags
+      description: Retrieve the flags for an environment.
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Flags'
```
khvn26: I'd love this to be the reality but unfortunately this schema is not correct.

Currently the `/flags` endpoint returns either:
- an array of flag objects
- a single flag object — if `feature` query parameter provided to the endpoint

khvn26 (changes requested):
Sorry for the long wait — dropped a couple comments.

.github/workflows/docs-pull-request.yml
```diff
@@ -29,8 +29,6 @@ jobs:
 
       - name: npm install
         run: npm ci --include=dev --ignore-scripts
+
       - name: Run docusaurus build to check for broken links etc
-        run: |
-          npx prettier --version
-          npx prettier --check docs
```
matthewelwell, line 35: This is being done by pre-commit now

docs/static/openapi/edge-api.yaml
matthewelwell: Fixed.

docs/static/openapi/edge-api.yaml
matthewelwell: Fixed (unfortunately).

khvn26 (approved):

docs/docs/edge-api/edge-api.info.mdx
```diff
@@ -0,0 +1,76 @@
...
+    as={"h2"}
+    className={"openapi-tabs__heading"}
+    children={"Authentication"}
+  >
+  </Heading><SchemaTabs
+    className={"openapi-tabs__security-schemes"}
+  >
+    <TabItem
+      label={"API Key: EnvironmentKeyAuth"}
+      value={"EnvironmentKeyAuth"}
```
khvn26: Judging from site content we should change "EnvironmentKeyAuth" to a human-readable value, but I don't know exactly how to do it.

matthewelwell: From what I can tell, there's no real option to do this in the OpenAPI specification. See [here](https://swagger.io/docs/specification/v3_0/authentication/api-keys/).

The best we can do perhaps, is to just drop the `Auth` from the end? Like this: 

image

What do you think?

khvn26: That looks good, let's do it.

matthewelwell: Done

Issue #4413, opened 2024-07-29
Introduce the SDK Evaluation Context object


khvn26:
## Abstract

While implementing #4278, a [discussion](https://github.com/Flagsmith/flagsmith-go-client/pull/133#discussion_r1694892206) has occurred about how we should approach SDK changes in the future.

The consensus is we should take after OpenFeature and implement a common _evaluation context object_ that a single `getFlags` interface should accept as its only input.

I'm creating this issue as a hub for further implementation discussion and planning.

## Schema proposal

You can observe and review the proposed schema in #4414. 

## Implementation plan

For implementation phase 1, We should settle on one of the following:
- Implement it in the Golang SDK as part of #4278, and evolve the existing interfaces for the rest of the SDKs.
- Implement it for all SDKs as part of #4278.

For phase 2, we should consider a single `/api/v2/flags` API with a schema identical to the Evaluation Context schema. Should we go forward with this, it'll allow us to fully code generate remote evaluation parts of the SDKs!

matthewelwell:
My preference for the implementation is (1) as I am keen to get transient traits out there ASAP. It also gives us some time to perfect the implementation and, if we decide it's not quite right, we would only need a breaking change in a single SDK. 

The only concern here would be that the documentation becomes slightly more difficult, but I think we can get around that.

matthewelwell:
A few comments on the schema: 

 1. I see that `Traits` is an object. Should this be a list?
 2. I see that `environment` is required, since the evaluation context (as I understand it), is provided when requesting the flags, shouldn't the environment default to what was provided when initialising the client, or are we removing that entirely from init? 
 3. Let's make sure that we're consistent with our casing. Maybe I'm missing a pattern, but it's odd to me that e.g. `environment`, `feature`, etc. are lower case, but `Traits` is upper case.

zachaysan:
I mirror @matthewelwell's comments on the approach. But other than that the overall approach seems sensible to me.

matthewelwell:
> A few comments on the schema:
> 
> 1. I see that `Traits` is an object. Should this be a list?
> 2. ~I see that `environment` is required, since the evaluation context (as I understand it), is provided when requesting the flags, shouldn't the environment default to what was provided when initialising the client, or are we removing that entirely from init?~
[...]

PR #4414, opened 2024-07-29
feat: Introduce the SDK Evaluation Context schema

khvn26:

Contributes to #4413.

## Changes

The following is a proposed Flagsmith Evaluation Context schema JSON Schema definition.

### Notes

- `IdentityEvaluationContext.traits` is a mapping of trait keys to a `TraitEvaluationContext | null` union. In case a `null` is provided, we assume intent to delete the previously persisted trait.
- `TraitEvaluationContext.value` can only be a `string` now.
- We now have a `TraitEvaluationContext.type` enum field designating the trait value type. For now it will support an SDK-side mapping layer. It's expected that the v2 SDK API should support it in the future.

This is a documentation change, but I'm verifying the schema against a validator.

### DIFF
--- a/sdk/evaluation-context.json
+++ b/sdk/evaluation-context.json
@@ -66,9 +66,7 @@
                     "title": "Transient"
                 }
             },
-            "required": [
-                "traits"
-            ],
+            "required": [],
             "title": "IdentityEvaluationContext",
             "type": "object"
         },
--- a/sdk/evaluation-context.json
+++ b/sdk/evaluation-context.json
@@ -69,42 +69,19 @@
         "TraitEvaluationContext": {
             "properties": {
                 "value": {
-                    "title": "Value",
-                    "type": "string"
-                },
-                "type": {
-                    "$ref": "#/$defs/TraitValueType"
+                    "title": "Value"
                 },
                 "transient": {
-                    "anyOf": [
-                        {
-                            "type": "boolean"
-                        },
-                        {
-                            "type": "null"
-                        }
-                    ],
-                    "default": null,
+                    "type": "boolean",
+                    "default": false,
                     "title": "Transient"
                 }
             },
             "required": [
-                "value",
-                "type"
+                "value"
             ],
             "title": "TraitEvaluationContext",
             "type": "object"
-        },
-        "TraitValueType": {
-            "enum": [
-                "String",
-                "SemverString",
-                "Integer",
-                "Float",
-                "Boolean"
-            ],
-            "title": "TraitValueType",
-            "type": "string"
         }
     },
     "properties": {
--- a/sdk/evaluation-context.json
+++ b/sdk/evaluation-context.json
@@ -77,10 +77,7 @@
                             "type": "string"
                         },
                         {
-                            "type": "integer"
-                        },
-                        {
-                            "type": "float"
+                            "type": "number"
                         }
                     ],
                     "title": "Value"
--- a/sdk/evaluation-context.json
+++ b/sdk/evaluation-context.json
@@ -109,7 +109,15 @@
     },
     "properties": {
         "environment": {
-            "$ref": "#/$defs/EnvironmentEvaluationContext"
+            "anyOf": [
+                {
+                    "$ref": "#/$defs/EnvironmentEvaluationContext"
+                },
+                {
+                    "type": "null"
+                }
+            ],
+            "default": null
         },
         "identity": {
             "anyOf": [
@@ -134,9 +142,6 @@
             "default": null
         }
     },
-    "required": [
-        "environment"
-    ],
     "title": "FlagsmithEvaluationContext",
     "type": "object"
 }
--- a/sdk/evaluation-context.json
+++ b/sdk/evaluation-context.json
@@ -69,6 +69,20 @@
         "TraitEvaluationContext": {
             "properties": {
                 "value": {
+                    "anyOf": [
+                        {
+                            "type": "boolean"
+                        },
+                        {
+                            "type": "string"
+                        },
+                        {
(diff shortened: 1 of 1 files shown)


zachaysan (approved):

sdk/evaluation-context.json
```diff
@@ -0,0 +1,147 @@
...
+                            {
+                                "$ref": "#/$defs/TraitEvaluationContext"
+                            },
+                            {
+                                "type": "null"
+                            }
+                        ]
+                    },
+                    "title": "Traits",
+                    "type": "object"
```
matthewelwell, line 54: Should this be a list?

khvn26: The idea is that we expect a map of strings (trait keys) to `TraitEvaluationContext | null` (trait values):

1. The schema ensures the trait data is being sent once per trait key.
2. We don't need a nullable `TraitEvaluationContext.value` field — just use `{"trait_key": null}` to delete a persisted trait.

sdk/evaluation-context.json
```diff
@@ -0,0 +1,147 @@
...
+                "type": {
+                    "$ref": "#/$defs/TraitValueType"
+                },
+                "transient": {
+                    "anyOf": [
+                        {
+                            "type": "boolean"
+                        },
+                        {
+                            "type": "null"
```
gagantrivedi: Why do we need null? We can just default to false

sdk/evaluation-context.json
matthewelwell: I see, ok. I guess what I don't necessarily like about that is that the keys aren't well defined, but perhaps that's acceptable with a schema such as this one. It will be harder to model with API documentation (certainly with DRF at least).

sdk/evaluation-context.json
khvn26: To distinguish between a default value and a non-present field. We have a little more flexibility this way. This approach is taken for all "optional" fields here.

sdk/evaluation-context.json
khvn26: Doesn't DRF have `DictField`?

khvn26: > the keys aren't well defined

You don't have much choice in JSON — keys are strings (which is what we need for traits)

matthewelwell: > Doesn't DRF have `DictField`?

It does, but then I don't think there's a way (without manual intervention) to define the schema of that dict field, so you just end up with a dict-like black hole! 

> You don't have much choice in JSON — keys are strings (which is what we need for traits)

Well, the other choice is still to use a list, right? But I agree that using an object here at least gives us a better way of validating uniqueness.

khvn26: If we use a list, in addition to not having uniqueness enforced, we'll have poorer deletion semantics — e.g. sending null `TraitEvaluationContext.value` along with `TraitEvaluationContext.type` won't make much sense.

matthewelwell: Yep, fair enough. I'm not necessarily advocating that we should definitely use a list, I'm mostly just raising the concern that the API documentation will be slightly more challenging.

khvn26: > It does, but then I don't think there's a way (without manual intervention) to define the schema of that dict field, so you just end up with a dict-like black hole!

DRF docs state the opposite: https://www.django-rest-framework.org/api-guide/fields/#dictfield

Basically, `child = fields.Nested(TraitEvaluationContextSchema)` will do the trick — unless `drf_yasg` does not support it.

khvn26: Besides the above, we could probably use a Pydantic model or validate directly against the json schema.

matthewelwell: This is all fine, and think I do agree that it's the best option here, but it's worth noting that it's not as simple as `child = fields.Nested(TraitEvaluationContextSchema)` because of the fact that the keys are dynamic.

khvn26: Sorry for dragging this on, but it is, unless I'm massively overlooking something?

According to docs, `child` is not a key name, it's a meta field representing every dict value.

As for the keys, the docs state that

> The keys in `DictField` are always assumed to be string values.

matthewelwell: ... oh, my bad! Sounds good then :)

matthewelwell (approved):

sdk/evaluation-context.json
khvn26: For example, we might consider adding a per-org/per-project option to regard all incoming traits as transient by default. If we offered a default `false` value, this would render the schema semantics inaccurate.

kyle-ssg (approved):

gagantrivedi (approved):

novakzaballa:
@khvn26 I love the idea of a schema.
I would like to see examples of how this schema would be used. While I think the current schema is complete and clear, I think we could simplify the structure of some of the contexts, but that would be clearer if we could have some implementation examples of the schema.

khvn26:
@novakzaballa One thing I could do to demonstrate is to go ahead and add it for JS and Golang SDK transient ids/traits PRs. You'll be able to review the schema with those implementations in consideration.

novakzaballa:
> @novakzaballa One thing I could do to demonstrate is to go ahead and add it for JS and Golang SDK transient ids/traits PRs. You'll be able to review the schema with those implementations in consideration.

Let's do that @khvn26 . Some examples I generated look good to me and we can always optimize if we see an opportunity while we move forward. I'll approve the PR. Nice job!

novakzaballa (approved):
LGTM

PR #4415, opened 2024-07-29
feat: Add command for Edge V2 migration

khvn26:

## Changes

This adds a `migrate_to_edge_v2` management command that migrated all unmigrated projects to Edge V2.

Added a unit test.

### DIFF
--- a/api/edge_api/apps.py
+++ b/api/edge_api/apps.py
@@ -1,0 +1,5 @@
+from django.apps import AppConfig
+
+
+class EdgeAPIAppConfig(AppConfig):
+    name = "edge_api"
--- a/api/edge_api/management/commands/migrate_to_edge_v2.py
+++ b/api/edge_api/management/commands/migrate_to_edge_v2.py
@@ -1,0 +1,17 @@
+from typing import Any
+
+from django.core.management import BaseCommand
+
+from projects.models import EdgeV2MigrationStatus, Project
+from projects.tasks import migrate_project_environments_to_v2
+
+
+class Command(BaseCommand):
+    def handle(self, *args: Any, **options: Any) -> str | None:
+        for project_id in Project.objects.filter(
+            edge_v2_migration_status__in=(
+                EdgeV2MigrationStatus.NOT_STARTED,
+                EdgeV2MigrationStatus.INCOMPLETE,
+            )
+        ).values_list("id", flat=True):
+            migrate_project_environments_to_v2(project_id)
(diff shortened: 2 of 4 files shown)


api/edge_api/apps.py
```diff
@@ -0,0 +1,5 @@
+from django.apps import AppConfig
```
gagantrivedi, line 1: Don't we already have an app named `edge_api`? :confused:

api/edge_api/management/commands/migrate_to_edge_v2.py
```diff
@@ -0,0 +1,17 @@
+from typing import Any
+
+from django.core.management import BaseCommand
+
+from projects.models import EdgeV2MigrationStatus, Project
+from projects.tasks import migrate_project_environments_to_v2
+
+
+class Command(BaseCommand):
```
gagantrivedi, line 9: Should this be an admin view rather than a command? How are we planning to trigger this?

api/edge_api/management/commands/migrate_to_edge_v2.py
```diff
@@ -0,0 +1,17 @@
...
+    def handle(self, *args: Any, **options: Any) -> str | None:
+        for project_id in Project.objects.filter(
+            edge_v2_migration_status__in=(
+                EdgeV2MigrationStatus.NOT_STARTED,
+                EdgeV2MigrationStatus.INCOMPLETE,
+            )
+        ).values_list("id", flat=True):
+            migrate_project_environments_to_v2(project_id)
```
gagantrivedi, line 17: Looks like it is going to create a task for every project? Can we verify that this won't overwhelm the task processor?

api/edge_api/apps.py
khvn26: Nope — we have `"edge_api.identities"` which, amusingly, cannot be added to `INSTALLED_APPS` because it collides with `"environments.identities"`.

api/edge_api/management/commands/migrate_to_edge_v2.py
khvn26: It would if it used the `.delay` method, but we're using `TaskHandler.__call__` here, which prompts the task to run synchronously.

api/edge_api/management/commands/migrate_to_edge_v2.py
khvn26: Create an ECS task, either manually or by adding it to deployment definition. It's a one-off task so either way is fine by me.

api/edge_api/management/commands/migrate_to_edge_v2.py
gagantrivedi: Ah got it

gagantrivedi (approved):

PR #4418, opened 2024-07-30
feat(app_analytics): Add cache for feature evaluation

gagantrivedi:

## Changes

Create and use cache to batch feature evaluation calls to reduce load on task processor

Added unit test


api/app_analytics/cache.py
```diff
@@ -31,3 +33,51 @@ def track_request(self, resource: int, host: str, environment_key: str):
             self._cache[key] += 1
         if (timezone.now() - self._last_flushed_at).seconds > CACHE_FLUSH_INTERVAL:
             self._flush()
+
+
+class FeatureEvaluationCache:
+    def __init__(self):
+        self._cache = {}
```
khvn26, line 42: Maybe use a Django cache here so self-hosting users could swap backends potentially.

api/app_analytics/cache.py
```diff
@@ -31,3 +33,51 @@ def track_request(self, resource: int, host: str, environment_key: str):
...
+        self._last_flushed_at = timezone.now()
+
+    def _flush(self):
+        evaluation_data = {}
+        for (environment_id, feature_name), eval_count in self._cache.items():
+            if environment_id in evaluation_data:
+                evaluation_data[environment_id][feature_name] = eval_count
+            else:
+                evaluation_data[environment_id] = {feature_name: eval_count}
```
khvn26: nit: Use `defaultdict(dict)`

api/app_analytics/cache.py
```diff
@@ -31,3 +33,51 @@ def track_request(self, resource: int, host: str, environment_key: str):
...
+        key = (environment_id, feature_name)
+        if key not in self._cache:
+            self._cache[key] = evaluation_count
+        else:
+            self._cache[key] += evaluation_count
+
+        if (
+            timezone.now() - self._last_flushed_at
+        ).seconds > settings.FEATURE_EVALUATION_CACHE_SECONDS:
+            self._flush()
```
khvn26, line 82: Just to clarify: since the cache is in-memory by default, we have to be ready to lose unsent evaluations.

If it's not an option, we probably should add a destructor that calls the `_flush` method, or come up with a better solution that guarantees the flush in case the app is shut down with a populated cache.

api/app_analytics/cache.py
```diff
@@ -31,3 +33,51 @@ def track_request(self, resource: int, host: str, environment_key: str):
...
+                    }
+                )
+
+            elif settings.INFLUXDB_TOKEN:
+                track_feature_evaluation_influxdb.delay(
+                    kwargs={
+                        "environment_id": environment_id,
+                        "feature_evaluations": feature_evaluations,
+                    }
+                )
```
khvn26, line 65: nit: I feel that the postgres/influx discrimination belongs inside the task itself. Moreover, this should improve analytics backend migration for self-hosting customers.

khvn26 (changes requested):
Two questions and two nitpicks. Otherwise, looks good.

api/app_analytics/cache.py
gagantrivedi: I  think we are okay with that @matthewelwell ?

api/app_analytics/cache.py
gagantrivedi: I don't really think anything other than memory makes much sense here(and we don't really "support" redis/memcache at the moment)

api/app_analytics/cache.py
gagantrivedi: Yeah, I agree with you, but that's a different pull request I think(because there are two v2 tasks as well)

api/app_analytics/cache.py
matthewelwell: Yeah, these figures don't need to be super accurate. I think that's fine.

khvn26 (approved):

PR #4422, opened 2024-07-30
fix: Check API usage before restricting serving flags and admin

zachaysan:
## Changes

Added a check for the latest API usage since the 7 day grace period could have been quieter, reducing the usage to within limits.

See this issue for more context and information: https://github.com/Flagsmith/flagsmith-private/issues/14

Altered an existing test.


api/organisations/tasks.py
```diff
@@ -290,6 +290,17 @@ def restrict_use_due_to_api_limit_grace_period_over() -> None:
         if not stop_serving and not block_access:
             continue
 
+        if not organisation.has_subscription_information_cache():
+            continue
+
+        subscription_cache = organisation.subscription_information_cache
+        api_usage = get_current_api_usage(organisation.id, "30d")
```
khvn26, line 297: Do you think `"30d"` should be a constant?

api/organisations/tasks.py
```diff
@@ -290,6 +290,17 @@ def restrict_use_due_to_api_limit_grace_period_over() -> None:
...
+        if api_usage / subscription_cache.allowed_30d_api_calls < 1.0:
+            logger.info(
+                f"API use for organisation {organisation.id} has fallen to below limit, so not restricting use."
+            )
```
khvn26, line 301: Do we _need_ to log that? I'd rather log the other branch?

khvn26 (approved):
LGTM overall with a couple of questions.

api/organisations/tasks.py
zachaysan: In an upcoming branch the calls that say "30d" or similar will be replaced with datetimes so I think we can just leave it for now.

api/organisations/tasks.py
zachaysan: This is the less frequently occurring case where we may need to debug when the task didn't select them, so I think it's a good candidate for inclusion.

PR #4423, opened 2024-07-30
ci: reduce size of custom runner for API unit tests

matthewelwell:
## Changes

Testing out the performance impact by reducing the size of the runner that's running the unit tests. 

CI change.


matthewelwell:
Closing because this reduces the time taken for the unit tests drastically so it won't actually save us any money since we'd just be paying for more time.

PR #4433, opened 2024-07-31
feat(versioning): limit returned number of versions by plan

matthewelwell:
## Changes

This PR primarily sets out to implement a limit on the number of versions returned for a given subscription plan. As part of the discussions related to this functionality, however, I have implemented (in this same PR) a similar limit on the audit log records. 

Some things we'll need to do outside of the deployment of this: 

 1. Add the limit information to the metadata for the relevant plans

Added / updated unit tests.


gagantrivedi (dismissed):

matthewelwell:
This may need to change based on a discussion [here](https://flagsmith.slack.com/archives/C01NC7EN7DE/p1723028676546999). Moving this PR to draft for now.

api/organisations/subscriptions/metadata.py
```diff
@@ -9,38 +11,44 @@ def __init__(
...
-        elif other.projects is None and self.projects is not None:
-            projects = self.projects
-        else:
-            projects = None
-
         return self.__class__(
             seats=self.seats + other.seats,
             api_calls=self.api_calls + other.api_calls,
-            projects=projects,
+            projects=_add_nullable_limit_attributes(self.projects, other.projects),
```
matthewelwell, line 35: Note: this does technically change the behaviour since, previously, if we added 2 metadata objects together where one had `project=None` and the other had `projects=1` the result would be `projects=1` but (a) this didn't make sense because `None` implied no limit so adding 1 should have kept the 'no limit', and (b) we don't actually use this functionality because we don't have any add ons for extra projects (certainly not that we use in SaaS).

api/tests/unit/organisations/chargebee/test_unit_chargebee_metadata.py
```diff
@@ -12,7 +12,7 @@ def test_add_chargebee_object_meta_data():
     # Then
     assert added_chargebee_obj_metadata.seats == 30
     assert added_chargebee_obj_metadata.api_calls == 300
-    assert added_chargebee_obj_metadata.projects == 100
+    assert added_chargebee_obj_metadata.projects is None
```
matthewelwell, line 15: See explanation [here](https://github.com/Flagsmith/flagsmith/pull/4433#discussion_r1718731826).

matthewelwell:
@kyle-ssg you can use the uffizzi environment to develop against here but you won't be able to alter the value of the `feature_history_visibility_days` attribute that is now returned on the `/organisations/:id/get-subscription-metadata` endpoint. 

The default for all organisations though is 7 days so hopefully that at least gives you something to work with. Really, I don't think there's anything for you to do though as the API only returns the correct data. Maybe you just want to add a UI hint somewhere based on this value (bear in mind that `null` means no limit, 0 means no data). 

It's possibly worth mentioning that this PR also applies the same logic to the Audit Log (`audit_log_visibility_days` also added to the same endpoint as above). 

If you want to be able to manipulate these values, it's probably best if you run the API via docker (using the `ghcr.io/flagsmith/flagsmith-api:pr-4433` image) and let me know and I can help update those values.

gagantrivedi:
This is ready for review, right?

matthewelwell:
> This is ready for review, right?

Yep.

api/audit/views.py
```diff
@@ -44,18 +48,42 @@ def get_queryset(self) -> QuerySet[AuditLog]:
...
-    def _get_base_filters(self) -> Q:
-        return Q()
+        return self._apply_visibility_limits(queryset)
 
     def get_serializer_class(self):
         return {"retrieve": AuditLogRetrieveSerializer}.get(
             self.action, AuditLogListSerializer
         )
 
+    def _get_base_filters(self) -> Q:  # pragma: no cover
```
gagantrivedi, line 62: Why are we keeping this?

api/audit/views.py
```diff
@@ -64,16 +92,53 @@ def _get_base_filters(self) -> Q:
...
     def _get_base_filters(self) -> Q:
         return Q(project__organisation__id=self.kwargs["organisation_pk"])
 
+    def _get_organisation(self) -> Organisation | None:
+        return (
+            Organisation.objects.select_related(
+                "subscription", "subscription_information_cache"
+            )
+            .filter(pk=self.kwargs["organisation_pk"])
+            .first()
```
gagantrivedi, line 127: nit: since this is always going to return 1 organisation, can we change this to `get()` to  make it a bit easier to read

api/features/versioning/views.py
```diff
@@ -139,6 +146,27 @@ def publish(self, request: Request, **kwargs) -> Response:
...
+            and (
+                version_limit_days := subscription_metadata.feature_history_visibility_days
+            )
+            is not None
+        ):
+            limited_queryset = queryset.filter(
+                Q(live_from__gte=timezone.now() - timedelta(days=version_limit_days))
+                | Q(live_from__isnull=True)
+            )
+            return limited_queryset if limited_queryset.exists() else queryset[:1]
```
gagantrivedi: Question: Why are we doing `queryset[:1]` instead of returning `queryset` Ah, it's because it's current version?

gagantrivedi (approved):
Few minor changes/questions;

api/audit/views.py
matthewelwell: This is essentially just an abstract method. I did think about whether I should mark it as an abstract method and/or use `NotImplementedError()` but I decided that it was fine to just keep it as it is since there could in the future be a use case where we don't need to override this method.

api/audit/views.py
matthewelwell: The reason I did this is that I preferred this code [here](https://github.com/Flagsmith/flagsmith/pull/4433/files#diff-4e1f59ff543830cbdbcd9a52331502f15bef30c1b71fa2b20e666996f25da7c5R68-R70) vs a `try` / `except` but maybe your suggestion is better tbh.

api/features/versioning/views.py
matthewelwell: Yeah, maybe I should add a comment here.

api/audit/views.py
matthewelwell: Hmm... actually, it's also because [this implementation](https://github.com/Flagsmith/flagsmith/pull/4433/files#diff-4e1f59ff543830cbdbcd9a52331502f15bef30c1b71fa2b20e666996f25da7c5R95-R112) _has_ to be `.first()` so I thought it was better to keep it consistent rather than having to to a try / except _and_ a null check.

api/features/versioning/views.py
matthewelwell: Done

matthewelwell:
@gagantrivedi I've responded to your questions and implemented a change for one of them. I also had to make a change that I thought about in my own head after you reviewed as well so could you take another quick look at those changes?

gagantrivedi (approved):

PR #4440, opened 2024-07-31
fix: Create a check for billing started at in API usage task helper

zachaysan:
## Changes

Check when `billing_starts_at` is `None` and log it and return from the function.

One new test which checks the error behaviour. Existing tests can cover normal operation.


novakzaballa (approved):
LGTM

PR #4449, opened 2024-08-02
feat: Support blank identifiers, assume transient

### LINKED ISSUE #4448: Support blank identifiers, assume transient
Contributes to #4278.

Evolve the API schemas as per the following:

#### `#/definitions/IdentifyWithTraits`

```diff
{
    "required": [
        "identifier"
    ],
    "type": "object",
    "properties": {
        "identifier": {
            "title": "Identifier",
+           "x-nullable": true,
            "type": "string",
-           "minLength": 1
        },
        "traits": {
            "type": "array",
            "items": {
                "$ref": "#/definitions/TraitSerializerBasic"
            }
        },
        "flags": {
            "type": "array",
            "items": {
                "$ref": "#/definitions/SDKFeatureState"
            },
            "readOnly": true
         },
         "transient": {
             "type": "boolean",
             "x-nullable": true
        }
    }
}
```
If null/empty identifier provided, assume "transient": true. Supply `identifier=""` to the evaluation engine.

khvn26:

## Changes

Closes #4448.

This PR adds the following for `POST` `/api/v1/identities`:
- Allow users to provide null/blank identifiers. The identity is assigned a UUID identifier to evaluate against and is not persisted in this case.
- If an identity is provided with `"transient": true`, all provided traits are automatically considered transient.
- If an identifier is provided, the existing persisted traits and overrides are loaded for evaluation. In case of collisions, transient traits take precedence during evaluation but are not persisted.
- Identifiers are returned to keep consistency with Edge API.

Updated integration and unit tests with new cases.

### DIFF
--- a/api/environments/identities/models.py
+++ b/api/environments/identities/models.py
@@ -199,7 +199,7 @@ class Identity(models.Model):
     def generate_traits(
         self,
         trait_data_items: list[SDKTraitData],
-        persist=False,
+        persist: bool = False,
     ) -> list[Trait]:
         """
         Given a list of trait data items, validated by TraitSerializerFull, generate
--- a/api/environments/identities/models.py
+++ b/api/environments/identities/models.py
@@ -4,12 +4,12 @@ from itertools import chain
 from django.db import models
 from django.db.models import Prefetch, Q
 from django.utils import timezone
-from flag_engine.identities.traits.types import TraitValue
 from flag_engine.segments.evaluator import evaluate_identity_in_segment
 
 from environments.identities.managers import IdentityManager
 from environments.identities.traits.models import Trait
 from environments.models import Environment
+from environments.sdk.types import SDKTraitData
 from features.models import FeatureState
 from features.multivariate.models import MultivariateFeatureStateValue
 from segments.models import Segment
@@ -196,7 +196,11 @@ class Identity(models.Model):
     def __str__(self):
         return "Account %s" % self.identifier
 
-    def generate_traits(self, trait_data_items, persist=False):
+    def generate_traits(
+        self,
+        trait_data_items: list[SDKTraitData],
+        persist=False,
+    ) -> list[Trait]:
         """
         Given a list of trait data items, validated by TraitSerializerFull, generate
         a list of TraitModel objects for the given identity.
@@ -232,7 +236,7 @@ class Identity(models.Model):
 
     def update_traits(
         self,
-        trait_data_items: list[dict[str, TraitValue]],
+        trait_data_items: list[SDKTraitData],
     ) -> list[Trait]:
         """
         Given a list of traits, update any that already exist and create any new ones.
--- a/api/environments/sdk/serializers.py
+++ b/api/environments/sdk/serializers.py
@@ -131,7 +131,6 @@ class IdentifyWithTraitsSerializer(
     HideSensitiveFieldsSerializerMixin, serializers.Serializer
 ):
     identifier = serializers.CharField(
-        write_only=True,
         required=False,
         allow_blank=True,
         allow_null=True,
@@ -185,6 +184,7 @@ class IdentifyWithTraitsSerializer(
 
         return {
             "identity": identity,
+            "identifier": identity.identifier,
             "traits": traits,
             "flags": all_feature_states,
         }
--- a/api/environments/sdk/serializers.py
+++ b/api/environments/sdk/serializers.py
@@ -159,8 +159,8 @@ class IdentifyWithTraitsSerializer(
             )
 
         elif transient:
-            # Get presently stored traits and identity overrides
-            # but don't persist incoming data.
+            # Don't persist incoming data but load presently stored
+            # overrides and traits, if any.
             identity, traits = get_identified_transient_identity_and_traits(
                 environment=environment,
                 identifier=identifier,
@@ -168,8 +168,8 @@ class IdentifyWithTraitsSerializer(
             )
 
         else:
-            # Persist the identity in accordance with non-local settings
-            # and individual trait transiency.
+            # Persist the identity in accordance with individual trait transiency
+            # and persistence settings outside of request context.
             identity, traits = get_persisted_identity_and_traits(
                 environment=environment,
                 identifier=identifier,
--- a/api/environments/sdk/serializers.py
+++ b/api/environments/sdk/serializers.py
@@ -125,7 +125,12 @@ class IdentitySerializerWithTraitsAndSegments(serializers.Serializer):
 class IdentifyWithTraitsSerializer(
     HideSensitiveFieldsSerializerMixin, serializers.Serializer
 ):
-    identifier = serializers.CharField(write_only=True, required=True)
+    identifier = serializers.CharField(
+        write_only=True,
+        required=False,
+        allow_blank=True,
+        allow_null=True,
+    )
     transient = serializers.BooleanField(write_only=True, default=False)
     traits = TraitSerializerBasic(required=False, many=True)
     flags = SDKFeatureStateSerializer(read_only=True, many=True)
@@ -137,22 +142,23 @@ class IdentifyWithTraitsSerializer(
         Create the identity with the associated traits
         (optionally store traits if flag set on org)
         """
+        identifier = self.validated_data.get("identifier") or ""
         environment = self.context["environment"]
 
         transient = self.validated_data["transient"]
         trait_data_items = self.validated_data.get("traits", [])
 
-        if transient:
+        if transient or not identifier:
             identity = Identity(
                 created_date=timezone.now(),
-                identifier=self.validated_data["identifier"],
+                identifier=identifier,
                 environment=environment,
             )
             trait_models = identity.generate_traits(trait_data_items, persist=False)
 
         else:
             identity, created = Identity.objects.get_or_create(
-                identifier=self.validated_data["identifier"], environment=environment
+                identifier=identifier, environment=environment
             )
 
             if not created and environment.project.organisation.persist_trait_data:
--- a/api/environments/sdk/serializers.py
+++ b/api/environments/sdk/serializers.py
@@ -2,7 +2,6 @@ import typing
 from collections import defaultdict
 
 from core.constants import BOOLEAN, FLOAT, INTEGER, STRING
-from django.utils import timezone
 from rest_framework import serializers
 
 from environments.identities.models import Identity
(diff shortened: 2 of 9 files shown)


api/environments/identities/models.py
```diff
@@ -196,7 +196,11 @@ def get_all_user_traits(self):
     def __str__(self):
         return "Account %s" % self.identifier
 
-    def generate_traits(self, trait_data_items, persist=False):
+    def generate_traits(
+        self,
+        trait_data_items: list[SDKTraitData],
+        persist=False,
```
matthewelwell:
```suggestion
        persist: bool = False,
```

api/environments/sdk/serializers.py
```diff
@@ -137,44 +146,46 @@ def save(self, **kwargs):
         Create the identity with the associated traits
         (optionally store traits if flag set on org)
         """
+        identifier = self.validated_data.get("identifier")
         environment = self.context["environment"]
-
         transient = self.validated_data["transient"]
```
matthewelwell, line 151: I wonder if it might help to be more explicit here and call this variable `transient_identity` or `identity_is_transient`?

api/environments/sdk/serializers.py
```diff
@@ -137,44 +146,46 @@ def save(self, **kwargs):
...
+                sdk_trait_data=sdk_trait_data,
             )
-            trait_models = identity.generate_traits(trait_data_items, persist=False)
 
-        else:
-            identity, created = Identity.objects.get_or_create(
-                identifier=self.validated_data["identifier"], environment=environment
+        elif transient:
+            # Get presently stored traits and identity overrides
+            # but don't persist incoming data.
```
matthewelwell: I'm sure I'll see this later, but I'm just commenting here as a reminder. 

This confused me a _little_ bit because it seems to talk explicitly about the case where the identity does already exist (which I think is a smaller use case for transient identities). I wonder if the comment should be reworded a little bit - perhaps an '(if exists)' clause somewhere would be useful.

api/environments/sdk/serializers.py
```diff
@@ -137,44 +146,46 @@ def save(self, **kwargs):
...
-                # partially update any traits and return the full list
-                trait_models = identity.update_traits(trait_data_items)
-            else:
-                # generate traits for the identity and store them if configured to do so
-                trait_models = identity.generate_traits(
-                    trait_data_items,
-                    persist=environment.project.organisation.persist_trait_data,
-                )
+        else:
+            # Persist the identity in accordance with non-local settings
```
matthewelwell: What do we mean by 'non-local settings' here? I guess this is e.g. `Organisation.persist_trait_data` ?

api/environments/sdk/serializers.py
```diff
@@ -137,44 +146,46 @@ def save(self, **kwargs):
...
+            traits=traits,
             additional_filters=self.context.get("feature_states_additional_filters"),
         )
-        identify_integrations(identity, all_feature_states, trait_models)
+        identify_integrations(identity, all_feature_states, traits)
 
         return {
             "identity": identity,
-            "traits": trait_models,
+            "identifier": identity.identifier,
```
matthewelwell, line 187: Maybe this is to simplify some logic later, but I'm not sure why we explicitly need to return the identifier here if we're already returning the whole identity.

api/environments/sdk/services.py
```diff
@@ -0,0 +1,86 @@
...
+
+from environments.identities.models import Identity
+from environments.identities.traits.models import Trait
+from environments.models import Environment
+from environments.sdk.types import SDKTraitData
+
+IdentityAndTraits: TypeAlias = tuple[Identity, list[Trait]]
+
+
+def _get_transient_identity(
```
matthewelwell: nit: I'd rather see private functions at the bottom of the module.

api/environments/sdk/services.py
```diff
@@ -0,0 +1,86 @@
...
+            identity := _get_transient_identity(
+                environment=environment,
+                identifier=str(uuid.uuid4()),
+            )
+        ),
+        identity.generate_traits(sdk_trait_data, persist=False),
+    )
+
+
+def get_identified_transient_identity_and_traits(
```
matthewelwell, line 36: I think it would be good to add a docstring to this function to help explain some of the assumptions. For example, that the identity has been marked as transient.

api/tests/integration/environments/identities/test_integration_identities.py
```diff
@@ -224,13 +227,65 @@ def test_get_feature_states_for_identity_only_makes_one_query_to_get_mv_feature_
...
+        pytest.param(
+            {"identifier": None},
+            lazy_fixture("transient_random_identifier"),
+            id="null-identifier",
+        ),
+        pytest.param(
+            {}, lazy_fixture("transient_random_identifier"), id="missing-identifier"
+        ),
+    ],
+)
```
matthewelwell: This is quite complicated but comprehensive! 👌

api/tests/unit/environments/identities/test_unit_identities_views.py
```diff
@@ -1166,10 +1179,68 @@ def test_post_identities__transient__no_persistence(
...
+    # Then
+    assert response.status_code == status.HTTP_200_OK
+    response_json = response.json()
+
+    assert response_json["flags"][0]["feature_state_value"] == feature_state_value
+    assert response_json["traits"][0]["trait_key"] == trait.trait_key
+    assert not response_json["traits"][0].get("transient")
+
+    assert response_json["traits"][1]["trait_key"] == trait_key
+    assert response_json["traits"][1]["transient"]
```
matthewelwell, line 1239: This confused me for a while but I think this is because, when the identity is transient, we mark all traits as transient too? I think it might be worth adding a comment here to confirm that.

matthewelwell (commented):
I've added some comments to some specific areas in the code, but I also had one more general comment regarding the PR description. 

> Identifiers are returned to keep consistency with Edge API.

I'm not quite clear on this? I think perhaps it's just related to including the identifier in the API response, as we do in the Edge API.

api/environments/sdk/serializers.py
khvn26: I think it would only increase cognitive load for those reading the code if I'm entirely honest

api/environments/sdk/serializers.py
khvn26: This does simplify things, absurdly so if you start thinking of a "correct" implementation.

khvn26: (Have I mentioned my disdain for DRF serializers recently?)

api/environments/sdk/services.py
khvn26: Added docstrings for every service function.

api/environments/sdk/services.py
khvn26: Done.

api/environments/sdk/serializers.py
khvn26: I've improved the comment, and added the service docstring.

api/tests/unit/environments/identities/test_unit_identities_views.py
khvn26: Added more comments to the test, hopefully it's clearer now.

api/environments/sdk/serializers.py
khvn26: This is hopefully more clearly stated in the comment.

khvn26:
Note: changed transient identifier generation to a hash of traits as advised by [Gagan's review of the Edge API PR](https://github.com/Flagsmith/edge-api/pull/276#discussion_r1708375522).

api/environments/sdk/serializers.py
matthewelwell: My concern is that we have transient identities and transient traits in the same area of the code so I wanted to ensure there was a clear definition as to which entity this `transient` variable related to... but, this isn't a sticking point if you don't agree. I'll leave it up to you.

api/environments/sdk/services.py
```diff
@@ -0,0 +1,117 @@
...
+        }.values()
+    )
+
+
+def get_transient_identifier(sdk_trait_data: list[SDKTraitData]) -> str:
+    return hashlib.sha256(
+        "".join(
+            f'{trait["trait_key"]}{trait["trait_value"]["value"]}'
+            for trait in sorted(sdk_trait_data, key=itemgetter("trait_key"))
+        ).encode(),
```
matthewelwell: This is a good approach but I'm not sure it works how we want it to if no traits are provided? I think that would then mean that the same identifier is used every time. Maybe that's what we want, but I'm not sure it is?

khvn26: The concern is valid. I say let's fall back to uuid4 in case traits are empty.

matthewelwell: Agreed.

gagantrivedi: Isn't that just a call to `/flags`? We should definitely not allow that?

matthewelwell: No, I don't think so. Think about percentage split / MV flags.

khvn26: > Isn't that just a call to /flags?

Essentially, yes, but 

> We should definitely not allow that?

I don't think we _definitely_ should. The identifier is being returned now, so people are able to make use of it in potential subsequent calls with traits.

khvn26: The fallback is now implemented both here and in the Edge API PR.

api/environments/sdk/serializers.py
khvn26: > have transient identities and transient traits in the same area of the code

I would argue it's not exactly the same area. Individual trait transiency is resolved in `Identity.generate_traits`/`Identity.update_traits` methods.

matthewelwell (approved):

api/environments/sdk/services.py
matthewelwell: I'm happy with this. @gagantrivedi I'm keen to know your thoughts too?

gagantrivedi: I think I'd prefer mv evaluation to be stable, i.e: if the identifier doesn't exists maybe use a hash of empty string or environment key or something

khvn26: @gagantrivedi When designing this, I assumed the following API semantics for the MV evaluation:

1. If you want a guaranteed stable MV evaluation, provide an identifier.
2. If you don't provide an identifier, you're not entitled to stable MV evaluation. If you don't offer an identifier, the API will generate one for you that you'll be able to reuse along with the MV bucket that it's being put into.

(There's a [Slack thread](https://flagsmith.slack.com/archives/CTF0THS2D/p1722616527027659) with more context)

That being said, IMO, using a trait hash to stabilise MV for null identifiers is a good idea. However, placing all of the identifier-less + trait-less transient identities in the same bucket is definitely not desirable given the semantics listed above, as most of these requests are going to serve distinct anonymous users.

gagantrivedi: Sounds good!

PR #4450, opened 2024-08-02
refactor: Use datetimes for InfluxDB queries

zachaysan:
## Changes

This PR changes the former calling procedure from specifying strings for ranges to use datetimes for more exact matches. There is one function that was used by a view with serializer data sent directly to it which was avoided for the change but all other functions have been updated to the new approach. I tried to cover every caller but it is possible that something slipped through the cracks.

I ran many, many tests as I was going through and I updated scores of them to get the coverage we need to be confident that the code will pass in production.


novakzaballa (approved):
LGTM

PR #4451, opened 2024-08-05
refactor: reduce DB updates by traits update requests without real changes in the data

devapro:

## Changes

It is common case for mobile app to set traits on the app start, because some properties can be changed offline or by request from another device with the same account etc. However in this case db is update even if no changes in the data. This change should reduce updates by getting list of traits from db than compares the existing and incoming data, and only add/update when the data actually changes. This strategy drastically reduces unnecessary UPDATE operations in case when the clients send requests with the same data.

manually

_Please describe._

### DIFF
--- a/api/environments/identities/traits/views.py
+++ b/api/environments/identities/traits/views.py
@@ -238,6 +238,58 @@ class SDKTraits(mixins.CreateModelMixin, viewsets.GenericViewSet):
 
         return Response(serializer.data, status=200)
 
+    def _update_traits(self, request):
+
+        identities = {trait["identity"]["identifier"] for trait in request.data}
+
+        existing_traits = Trait.objects.filter(
+            identity__identifier__in=identities,
+            identity__environment=request.environment,
+        )
+
+        # Map to easily access existing traits
+        existing_traits_map = {
+            (trait.identity.identifier, trait.trait_key): trait
+            for trait in existing_traits
+        }
+
+        updated_traits = []
+        delete_filter_query = Q()
+
+        for trait in request.data:
+            trait_key = trait.get("trait_key")
+            identifier = trait["identity"]["identifier"]
+
+            if trait.get("trait_value") is None:
+                delete_filter_query = delete_filter_query | Q(
+                    trait_key=trait_key,
+                    identity__identifier=identifier,
+                    identity__environment=request.environment,
+                )
+                continue
+
+            existing_trait = existing_traits_map.get((identifier, trait_key))
+            if not existing_trait or existing_trait.trait_value != trait["trait_value"]:
+                updated_traits.append(trait)
+
+        if delete_filter_query:
+            Trait.objects.filter(delete_filter_query).delete()
+
+        if len(updated_traits) > 0:
+            serializer = self.get_serializer(data=updated_traits, many=True)
+            serializer.is_valid(raise_exception=True)
+            serializer.save()
+
+            if settings.EDGE_API_URL and request.environment.project.enable_dynamo_db:
+                forward_trait_requests.delay(
+                    args=(
+                        request.method,
+                        dict(request.headers),
+                        request.environment.project.id,
+                        updated_traits,
+                    )
+                )
+
     @swagger_auto_schema(request_body=SDKCreateUpdateTraitSerializer(many=True))
     @action(detail=False, methods=["PUT"], url_path="bulk")
     def bulk_create(self, request):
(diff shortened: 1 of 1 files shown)


zachaysan (dismissed):
Looks good. Though it seems like CI has failed for this test, perhaps try re-running the test suite with an empty commit?

api/environments/identities/traits/views.py
```diff
@@ -245,39 +245,65 @@ def bulk_create(self, request):
...
+                            request.environment.project.id,
+                            traits,
+                        )
                     )
-                )
 
-            return Response(serializer.data, status=200)
+                return Response(serializer.data if traits else [], status=200)
+
+            return Response(traits, status=200)
```
matthewelwell: Hmm... this actually looks like it changes the behaviour of the endpoint which could have an effect on clients. Previously, we are always returning the traits that were sent (assuming they were valid). Now we're sending an empty list if none of the traits are actually updated / created. 

Perhaps we could do something like the following: 

```python
all_traits = [*traits, *SDKCreateUpdateTraitSerializer(instance=existing_traits, many=True).data]
return Response(all_traits)
```

matthewelwell (commented):
> Though it seems like CI has failed for this test, perhaps try re-running the test suite with an empty commit?

This is actually an issue in our CI workflows that don't work correctly for external contributors. We are working to resolve this. 

I've added a comment to a specific part of the diff but, in general, I'd also like to see a test / some tests written for this logic.

api/environments/identities/traits/views.py
devapro: Thank you for review.
I made changes, now traits always added to the response.
Moved all logic of updating traits to new method, I think it will be a bit easy to read an understand in this case.

api/environments/identities/traits/views.py
```diff
@@ -273,11 +286,37 @@ def bulk_create(self, request):
...
-            return Response(serializer.data, status=200)
+    @swagger_auto_schema(request_body=SDKCreateUpdateTraitSerializer(many=True))
+    @action(detail=False, methods=["PUT"], url_path="bulk")
+    def bulk_create(self, request):
+        try:
+            if not request.environment.trait_persistence_allowed(request):
+                raise BadRequest("Unable to set traits with client key.")
+
+            # endpoint allows users to delete existing traits by sending null values
+            # for the trait value so we need to filter those out here
```
matthewelwell: This comment is now stale. It probably needs moving elsewhere.

api/environments/identities/traits/views.py
```diff
@@ -273,11 +286,37 @@ def bulk_create(self, request):
...
+                identity__identifier__in=identities,
+                identity__environment=request.environment,
+            )
+
+            return Response(
+                [
+                    *SDKCreateUpdateTraitSerializer(
+                        instance=all_traits, many=True
+                    ).data,
+                ],
```
matthewelwell:
```suggestion
            return Response(
                SDKCreateUpdateTraitSerializer(
                    instance=all_traits, many=True
                ).data,
```

matthewelwell (commented):
Thanks for the changes @devapro . I've added a couple more comments. 

As per my previous review though, it's still missing any sort of testing coverage.

matthewelwell:
@devapro just checking in to see if you had time to look at my recent review comments and whether you might have time to update the PR?

devapro:
Sorry for long response, will check your comments next week.

api/environments/identities/traits/views.py
devapro: Moved comment closer to related code.

devapro:
> Thanks for the changes @devapro . I've added a couple more comments.
> 
> As per my previous review though, it's still missing any sort of testing coverage.

I checked existing unit tests and looks like it is covered all logic that I refactored.
I think will be good to add test that checks case when db is not called, but my knowledge in python very limited, I didn't find an example of how to do it in current code.

api/environments/identities/traits/views.py
```diff
@@ -238,32 +238,47 @@ def increment_value(self, request):
 
         return Response(serializer.data, status=200)
 
-    @swagger_auto_schema(request_body=SDKCreateUpdateTraitSerializer(many=True))
-    @action(detail=False, methods=["PUT"], url_path="bulk")
-    def bulk_create(self, request):
-        try:
-            if not request.environment.trait_persistence_allowed(request):
-                raise BadRequest("Unable to set traits with client key.")
+    def _update_traits(self, request):
```
matthewelwell:
```suggestion
    def _update_traits(self, request: Request) -> None:
```

api/environments/identities/traits/views.py
```diff
@@ -238,32 +238,47 @@ def increment_value(self, request):
...
+
+        identities = {trait["identity"]["identifier"] for trait in request.data}
```
matthewelwell, line 246: We're unnecessarily doing this twice. Once here, and once in the remaining `bulk_create` view method. Why not pass in the `identities`, or return them from this method?

matthewelwell (commented):
This is looking better. I've left 2 fairly minor additional comments, but I think it's pretty close.

matthewelwell:
>I think will be good to add test that checks case when db is not called, but my knowledge in python very limited, I didn't find an example of how to do it in current code.

You can just copy one of the existing tests, manipulate the data to be what you want and then use the `django_assert_num_queries` fixture to do something like: 

```python
with django_assert_num_queries(0):
    ... 
```

You'll find plenty of usages of this fixture throughout the code base.

api/environments/identities/traits/views.py
devapro: fixed, thanks

api/environments/identities/traits/views.py
devapro: :heavy_check_mark:

devapro:
> > I think will be good to add test that checks case when db is not called, but my knowledge in python very limited, I didn't find an example of how to do it in current code.
> 
> You can just copy one of the existing tests, manipulate the data to be what you want and then use the `django_assert_num_queries` fixture to do something like:
> 
> ```python
> with django_assert_num_queries(0):
>     ... 
> ```
> 
> You'll find plenty of usages of this fixture throughout the code base.

I can't find existing test for method that I changed. It is not simple for me to understand how to write new test for specific method.

Didn't manage to run tests locally, for running project I am using docker + mounting specific files with changes, because can't install all required dependencies.

PR #4454, opened 2024-08-06
fix: don't allow bypassing `ALLOW_REGISTRATION_WITHOUT_INVITE` behaviour

matthewelwell:
## Changes

This PR adds a new `invite_hash` field to the payload on the register endpoint to ensure that it's not possible to bypass the `ALLOW_REGISTRATION_WITHOUT_INVITE` setting. 

Note: this will need to be added on the frontend. It's likely that the E2E tests will fail until that is done.  

Added new tests and updated an existing test.

### DIFF
--- a/api/custom_auth/oauth/serializers.py
+++ b/api/custom_auth/oauth/serializers.py
@@ -83,10 +83,9 @@ class OAuthLoginSerializer(InviteLinkValidationMixin, serializers.Serializer):
 
         if not existing_user:
             sign_up_type = self.validated_data.get("sign_up_type")
-            if not settings.ALLOW_REGISTRATION_WITHOUT_INVITE:
-                self._validate_registration_invite(
-                    {"email": email, **self.validated_data}
-                )
+            self._validate_registration_invite(
+                email=email, sign_up_type=self.validated_data.get("sign_up_type")
+            )
 
             return UserModel.objects.create(
                 **user_data, email=email.lower(), sign_up_type=sign_up_type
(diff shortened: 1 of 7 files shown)


matthewelwell:
FE component has been merged and tested [here](https://github.com/Flagsmith/flagsmith/pull/4458)

api/custom_auth/oauth/serializers.py
```diff
@@ -85,12 +83,10 @@ def _get_user(self, user_data: dict):
 
         if not existing_user:
             sign_up_type = self.validated_data.get("sign_up_type")
-            if not (
-                settings.ALLOW_REGISTRATION_WITHOUT_INVITE
-                or sign_up_type == SignUpType.INVITE_LINK.value
-                or Invite.objects.filter(email=email).exists()
-            ):
-                raise PermissionDenied(USER_REGISTRATION_WITHOUT_INVITE_ERROR_MESSAGE)
+            if not settings.ALLOW_REGISTRATION_WITHOUT_INVITE:
```
khvn26: Do you think it makes sense to move the setting check inside the `_validate_registration_invite` method?

api/custom_auth/oauth/serializers.py
```diff
@@ -85,12 +83,10 @@ def _get_user(self, user_data: dict):
...
+                self._validate_registration_invite(
+                    {"email": email, **self.validated_data}
```
khvn26: I'd rather provide dependencies explicitly to avoid ambiguity
```suggestion
                    {"email": email, "sign_up_type": self.validated_data["sign_up_type"]}
```

api/custom_auth/oauth/serializers.py
matthewelwell: Yeah, I debated this myself. I've moved it into the method as suggested.

api/custom_auth/oauth/serializers.py
matthewelwell: I've implemented this in a very similar way, but made it even more explicit.

khvn26 (approved):

PR #4459, opened 2024-08-07
fix: protect get environment document endpoint

matthewelwell:
## Changes

Ensures that `get_document` endpoint is correctly protected by permissions. 

Added a new unit test and updated existing one to ensure that the correct permission check is performed.

### DIFF
--- a/api/environments/permissions/permissions.py
+++ b/api/environments/permissions/permissions.py
@@ -45,6 +45,8 @@ class EnvironmentPermissions(IsAuthenticated):
     def has_object_permission(self, request, view, obj):
         if view.action == "clone":
             return request.user.has_project_permission(CREATE_ENVIRONMENT, obj.project)
+        elif view.action == "get_document":
+            return request.user.has_environment_permission(VIEW_ENVIRONMENT, obj)
 
         return request.user.is_environment_admin(obj) or view.action in [
             "user_permissions"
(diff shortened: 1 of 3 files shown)


api/environments/permissions/permissions.py
```diff
@@ -45,6 +45,8 @@ def has_permission(self, request, view):
     def has_object_permission(self, request, view, obj):
         if view.action == "clone":
             return request.user.has_project_permission(CREATE_ENVIRONMENT, obj.project)
+        elif view.action == "get_document":
+            return request.user.has_environment_permission(VIEW_ENVIRONMENT, obj)
 
         return request.user.is_environment_admin(obj) or view.action in [
```
khvn26, line 51: Am I right to assume `request.user.is_environment_admin(obj)` was always returning `True` before? If that's the case, why?

matthewelwell: No, the issue was that `has_object_permission` was not being called at all. That was added [here](https://github.com/Flagsmith/flagsmith/pull/4459/files#diff-47c3475f05ad0d8647b602c5c331798d8fe03e4ac587ecffadf67e609c566a3dR228-R230), so then I needed to add this to ensure that the correct response was given based on permissions.

khvn26 (approved):

PR #4462, opened 2024-08-07
fix: show correct SAML Frontend URL on edit

matthewelwell:
## Changes

Ensures that the current stored Frontend URL is shown when editing, instead of using the current URL. 

Also includes a bonus fix for a typo. 

Ran the FE against production and tested create / edit.


kyle-ssg (approved):

kyle-ssg:
Let's convet this to done

PR #4466, opened 2024-08-08
feat(versioning): add version rollback

matthewelwell:
## Changes

Adds a new endpoint to rollback to a previous version. 

Note that there are some implementation decisions that I made here, which are: 

 1. Rolling back to a given version just 're-instates' it to be the latest version. This does mean that we're essentially picking up the version and dropping it at the top of the list again. There's an argument to say that we should leave the version as it was and just create a new cloned version from the one that we're choosing to rollback to. I'm keen to get some feedback on this. 
 2. Rolling back creates 2 audit log records. One for the version that we're rolling back to, and one for the version that we're rolling back from. Again, there's an argument to say that this should be a single record, but I wanted to err on the side of more records, particularly to make the `related_object_uuid` logic easier. 

Added a new test to verify the logic end to end.


api/audit/tasks.py
```diff
@@ -184,3 +185,46 @@ def create_segment_priorities_changed_audit_log(
...
+@register_task_handler()
+def create_version_rolled_back_audit_log_records(
+    rolled_back_to_uuid: str,
+    rolled_back_from_uuid: str,
+    user_id: int | None,
+    api_key_id: int | None,
+) -> None:
+    rolled_back_from = EnvironmentFeatureVersion.objects.select_related(
+        "feature", "feature"
+    ).get(uuid=rolled_back_from_uuid)
```
zachaysan: Why is feature repeated twice here?

zachaysan (dismissed):
Overall approach looks good

api/audit/tasks.py
matthewelwell: Good catch, this was because I originally intended to add `"feature__project"` as well, but decided against it... apparently I made that decision halfway through writing the argument 😄

matthewelwell:
Converting this to draft while I work out how to handle this better with regards to change requests in particular.

matthewelwell:
Closing as PR will change signficantly

PR #4469, opened 2024-08-09
docs: Deprecate group-level permissions. Rewrite RBAC docs and add Mermaid diagrams

rolodato:
Deprecates group-level permissions. This is only a UI change and no action is required from customers. Added `target=_blank` to all links in the deprecation notice to avoid accidentally discarding any unsaved changes.

Can probably make more improvements to the docs here, but I'd be happy to merge it as-is.

The part about provisioning users and permissions is not strictly related to RBAC so it could go in its own document later on. This is probably the best place for it to be at the moment.

https://github.com/user-attachments/assets/a6c21c71-b741-4ff8-b52a-c12d41dfafa1

### DIFF
--- a/docs/docs/system-administration/rbac.md
+++ b/docs/docs/system-administration/rbac.md
@@ -77,6 +77,13 @@ Users can also join your organisation directly by logging in to Flagsmith using
 
 ### Email invites
 
+:::info
+
+If you are self-hosting Flagsmith, you must
+[configure an email provider](/deployment/hosting/locally-api#email-environment-variables) before using email invites.
+
+:::
+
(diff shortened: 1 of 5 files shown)


rolodato:
Reminder before merging: add the actual Flagsmith version when this is released here https://github.com/Flagsmith/flagsmith/pull/4469/files#diff-a272079eeec1855f9514e0c2254805b5cc9dcc82b3e09f6e5d4a26cda188c4b1R123

matthewelwell (approved):
Excellent work @rolodato. Love the initiative here, and the docs are very nice and clear!

PR #4481, opened 2024-08-12
fix: Add decorator for running task every hour

### LINKED ISSUE #4473: Add recurring task to update organisationsubscriptioninformationcache records with influx data instead of relying on pressing a button from the admin dashboard

zachaysan:
## Changes

Fixes https://github.com/Flagsmith/flagsmith/issues/4473

Added a decorator to run this task once an hour.

I called the `.delay()` method to ensure that the task is still callable.


api/organisations/tasks.py
```diff
@@ -84,6 +84,9 @@ def update_organisation_subscription_information_influx_cache():
     subscription_info_cache.update_caches((SubscriptionCacheEntity.INFLUX,))
 
 
+@register_recurring_task(
+    run_every=timedelta(hours=1),
+)
```
matthewelwell: Hmm... I think we should add this to the task above. We don't need to call chargebee as the data should be updated via the webhook (in fact we could perhaps even remove this functionality in a future PR). 

I think we should also reduce the frequency. I'd say either 6 or 12 hours would be sufficient realistically.

zachaysan: Ok I've switched the target for the decorator and set the frequency lower.

matthewelwell (approved):

PR #4486, opened 2024-08-13
refactor(get_permitted_environments): get rid of .distinct()

gagantrivedi:

## Changes

Refactor `get_permitted_environments_for_user` to not use distinct

Unit tests

### DIFF
--- a/api/permissions/permission_service.py
+++ b/api/permissions/permission_service.py
@@ -134,12 +134,15 @@ def get_permitted_environments_for_user(
             return queryset.prefetch_related("metadata")
         return queryset
 
-    base_filter = get_base_permission_filter(
+    environment_ids_from_base_filter = get_object_id_from_base_permission_filter(
         user, Environment, permission_key, tag_ids=tag_ids
     )
-    filter_ = base_filter & Q(project=project)
+    environment_ids = project.environments.filter(
+        id__in=environment_ids_from_base_filter
+    ).values_list("id", flat=True)
 
-    queryset = Environment.objects.filter(filter_)
+    queryset = Environment.objects.filter(id__in=environment_ids)
+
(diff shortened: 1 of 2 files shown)


api/permissions/permission_service.py
```diff
@@ -134,12 +134,15 @@ def get_permitted_environments_for_user(
...
-    base_filter = get_base_permission_filter(
+    environment_ids_from_base_filter = get_object_id_from_base_permission_filter(
         user, Environment, permission_key, tag_ids=tag_ids
     )
-    filter_ = base_filter & Q(project=project)
+    environment_ids = project.environments.filter(
+        id__in=environment_ids_from_base_filter
+    ).values_list("id", flat=True)
+
+    queryset = Environment.objects.filter(id__in=environment_ids)
```
matthewelwell: Unless I'm missing something, I think we can simplify this to remove _one_ query? 

```suggestion
    queryset = Environment.objects.filter(
        id__in=environment_ids_from_base_filter, project=project
    )
```

I can't see any reason that we'd still need to use distinct after this? This code I ran in the dango shell seems to suggest that's correct. 

```
>>> Environment.objects.filter(id__in=[1, 2, 2, 1]).count()
2
```

If we did still have an issue, maybe we could just do: 

```suggestion
    queryset = Environment.objects.filter(
        id__in=set(environment_ids_from_base_filter), project=project
    )
```

matthewelwell (changes requested):

api/permissions/permission_service.py
gagantrivedi: I have updated it because it looks a tad bit more clean, but it doesn't change the query count because the part that we removed was being evaluated lazily

matthewelwell (approved):

PR #4491, opened 2024-08-13
fix: add reverse sql to versioning migration

matthewelwell:
## Changes

This fixes a previously added migration to include a reverse script to ensure that duplicate segments are not shown after a rollback due to the versioning system added past this point. 

Added 2 new migration tests and tested the sql script manually on my local database.


api/tests/unit/segments/test_unit_segments_migrations.py
```diff
@@ -0,0 +1,239 @@
...
+from django_test_migrations.migrator import Migrator
+from flag_engine.segments import constants
+from pytest_django.fixtures import SettingsWrapper
+
+
+@pytest.mark.skipif(
+    test_settings.SKIP_MIGRATION_TESTS is True,
+    reason="Skip migration tests to speed up tests where necessary",
+)
+def test_create_whitelisted_segments_migration(
```
matthewelwell, line 14: Hmm... this test was not modified. The diff is showing that a file was deleted / created vs the renaming action that actually happened. This test function does not need to be reviewed.

api/segments/migrations/0023_add_versioning_to_segments.py
```diff
@@ -5,6 +5,16 @@
...
 
 
+def _get_sql_file_content(filename: str) -> str:
+    return open(
+        os.path.join(
+            os.path.dirname(__file__),
+            "sql",
+            filename,
+        )
+    ).read()
```
khvn26: Please be consistent with [other migrations that do this](https://github.com/Flagsmith/flagsmith/blob/36e634ca057e6aa55ffed41686a05df0883a1062/api/features/versioning/managers.py#L12-L13).

khvn26: The above is a nitpick, though.

khvn26 (approved):

api/segments/migrations/0023_add_versioning_to_segments.py
matthewelwell: This is done.

PR #4494, opened 2024-08-13
feat: Copy ACS URL for SAML configurations to clipboard. Disable editing SAML configuration names

rolodato:

### Disable editing SAML configuration name

Changing a SAML configuration name has the following effects:

* "Remember this SAML organisation" no longer works for users that previously logged in with this configuration, locking users out until they can figure this out
* ACS URL changes, breaking any previously-configured IdPs which have whitelisted this URL
* SP metadata XML changes, which can cause confusion if it was already downloaded before the name changed

Preventing changes to the configuration name from the UI prevents these footguns, while still leaving an escape hatch open to change names from the API or Django Admin. A better solution would be to use fixed SAML configuration IDs (e.g. numbers) in the ACS URLs instead of the configuration name, but this is a good stopgap until we can implement that. For example:

```
https://flagsmith.example.com/api/v1/auth/saml/1234/response/
```

instead of

```
https://flagsmith.example.com/api/v1/auth/saml/my_config_name/response/
```

### Add button to copy ACS URL

The ACS URL is only currently available in documentation and requires figuring out the API URL and manually building the URL, remembering to include trailing slashes, etc. This PR adds a button to copy the correct URL.

I decided to add a button in the list of configurations instead of showing the full URL inside the edit modal because it's easier to implement.

**Unfinished**: The API URL used is just whatever URL is configured for Flagsmith-on-Flagsmith. For example, in my dev environment this resolves to `https://api.flagsmith.com/api/v1/` but should be `http://localhost:8000/api/v1/`. I haven't figured out how to get the correct URL yet.

This still looks bad but works:

image

Manually - create a SAML configuration and look at it from the frontend.


kyle-ssg (approved):

rolodato:
Moved the ACS URL into the edit modal - not too happy with the styling but it's an improvement over what we have now in prod for sure



https://github.com/user-attachments/assets/2ee744d7-933e-434e-a4ca-fb2faa175a11

rolodato:
@kyle-ssg could I bother you again with a quick review on this? 🙏

zachaysan (approved):

PR #4495, opened 2024-08-14
fix: Make influx cache task recurring

zachaysan:
## Changes

Stacking the task decorators doesn't work. I looked into the task processor implementation and I think it's because the wrapped function can't be called. So instead I just redefined the function with an extended name.

I did not test the code since the change is simple.

### DIFF
--- a/api/organisations/tasks.py
+++ b/api/organisations/tasks.py
@@ -82,6 +82,10 @@ def send_org_subscription_cancelled_alert(
 @register_recurring_task(
     run_every=timedelta(hours=6),
 )
+def update_organisation_subscription_information_influx_cache_recurring():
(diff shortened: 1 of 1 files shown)


api/organisations/tasks.py
```diff
@@ -82,6 +82,10 @@ def send_org_subscription_cancelled_alert(
 @register_recurring_task(
     run_every=timedelta(hours=6),
 )
+def update_organisation_subscription_information_influx_cache_recurring():
+    subscription_info_cache.update_caches((SubscriptionCacheEntity.INFLUX,))
```
matthewelwell:
```suggestion
    update_organisation_subscription_information_influx_cache()
```

matthewelwell (approved):

PR #4498, opened 2024-08-14
feat: Improve Github integration

### LINKED ISSUE #4349: GitHub integration - styling issues

There are a few stylistic changes that would improve the integration

These will look better if they are inline rather than on separate rows
image

Replace loading state with a  center <Loader/>
image

Remove margin bottom from select options, just use standard styles
image

We are showing a no external resources linked as well as the ui to add in the center, this looks quite strange I think we have 2 options

1 - Left align the "You have no" text
2 - Remove the messaging

Linking an additional item UI should be above existing, it looks a bit strange it being underneath.

The save button feels unnecessary, we may as well just apply it when an item is selected 
image

 
Link icon overflowing to new line
image

N/A

### Additional context

_No response_

kyle-ssg:

## Changes

The current GitHub integration needed neatening up, there was also unnecessary code. 

**Before**

image

**After**

image

In addition, this will also:

- Preselect a repository
- Preselect "Issue"
- Adding an item doesn't require clicking a save button

Currently it looks like adding an integration throws a 500, but I was able to validate that it sends the same payload.


frontend/web/components/GitHubResourcesSelect.tsx
```diff
@@ -1,147 +1,61 @@
...
-    useGitHubResourceSelectProvider()
+  const githubTypes = Object.values(Constants.resourceTypes).filter(
+    (v) => v.type === 'GITHUB',
+  )
+  const [repo, setRepo] = useState('')
+  const { repoName, repoOwner } = useMemo(() => {
+    if (!repo) {
+      return {}
+    }
+    const [repoOwner, repoName] = repo.split('/')
```
novakzaballa: There seems to be an error here. The `repo` is structured as repo-name/repo-owner, which is causing an issue with the API.

Screenshot 2024-08-15 at 9 55 21 AM

frontend/web/components/GitHubResourcesSelect.tsx
```diff
@@ -153,43 +67,74 @@ const GitHubResourcesSelect: FC<GitHubResourcesSelectType> = ({
...
+            })}
+          />
+        </div>
+      </div>
+      {!!repoName && !!repoOwner && (
+        <div>
+          <label className='cols-sm-2 control-label'>
+            Link new Issue / Pull Request
+          </label>
+          <Select
```
novakzaballa, line 97: An issue that is already linked is still appearing in the list, which will cause an error with the backend when trying to select it again.

Screenshot 2024-08-15 at 10 01 56 AM

Screenshot 2024-08-15 at 10 02 03 AM

novakzaballa (commented):
Most of the code looks good, but I left some comments that could cause errors.

frontend/web/components/GitHubResourcesSelect.tsx
kyle-ssg: ah, that's what I get for testing with flagsmith/flagsmith

frontend/web/components/GitHubResourcesSelect.tsx
kyle-ssg: Ah didn't spot this in the original provider. Sorted

novakzaballa (approved):

PR #4504, opened 2024-08-15
feat: Added a Streaming identity debugger to Identity tab(UsersPage.tsx)

### LINKED ISSUE #3459: Implement a streaming Identity Debugger

Sometimes it can be hard to grok why identity data/sdk implementations are not working as expected

Add a page in the dashboard that shows a real time stream of identities being sent to the API from SDKs. This is similar to real time debugger pages in analytics tools (https://www.docs.developers.amplitude.com/data/debugger/#ingestion-debugger)

None

### Additional context

We could reuse the identity webhook to stream content back into the app.

SebastianRamos3:

## Changes
Solve for[ 3459](https://github.com/Flagsmith/flagsmith/issues/3459)
-Added graph in identities tab(UsersPage.tsx) on home aside.
-Graph uses useOrganisationUsage to retrieve usage data for identities.
-use useOrganisation to get organizationID, as it is not in route parameters for UsersPage.tsx
-Tested how graph looked when it received data.
-Tested how the graph looks when it doesn't receive data.

I would like to highlight two points:

GitHub Services: My request to delete the other pull request was unfortunately denied by GitHub services.
Clarification on API Tracking: During our review of the initial pull request, we were informed that the API does not track new identities but rather the calls made to identities and we may need a new api to achieve that. From our understanding, our goal was to create a graph depicting a "real-time stream of identities being sent to the API from SDKs," rather than tracking new identities. Please let us know if we have misunderstood any aspect of this requirement.


kyle-ssg:
Hey, thanks for the PR. From what I see this PR copies a similar graph to the organisation usage one and just restricts to identity calls, I'm not sure that this adds much / solves the original issue.

Looking at the related issue I think this is quite different to the problem we'd want to solve. To validate implementation of identity calls we'd want to see individual events in realtime with data of each call rather than a graph. I don't believe we have any api right now to achieve the desired outcome.

PR #4514, opened 2024-08-19
fix: Catch API billing errors

### LINKED ISSUE #4515: Catch the error that add_100k_api_calls_... in order to allow processing to continue for API billing code

zachaysan:
## Changes

Fixes https://github.com/Flagsmith/flagsmith/issues/4515

Added a try-except block around the functions that apply charges to organisations.

One new test to verify the caught exception is logged and that no billing data is created.


novakzaballa (approved):
LGTM, but it looks like there are lines missing coverage.

PR #4518, opened 2024-08-19
fix: subscription info cache race condition

matthewelwell:
## Changes

Fixes race condition(s) related to the cancellation of a subscription. Related sentry issue [here](https://flagsmith.sentry.io/issues/5517108730). 

As far as I can tell from doing some manual testing locally by running the API and pointing it to the staging database, there were likely 2 issues in the code here: 

 1. There was a `BEFORE_DELETE` hook on the organisation model which cancelled the subscription if it was paid. This involved deleting the related `OrganisationSubscriptionInformationCache` object, but the execution path then continued to delete the organisation, and subsequently try and cascade delete the OSIC object. This would fail (seemingly only some of the time). 
 2. When a subscription is cancelled in chargebee, that also triggers a webhook which will execute the same code and try to delete the OSIC object which possibly won't exist depending on when the webhook is triggered / received. 

I have updated a unit test to test the new behaviour, however, (perhaps more importantly) I have also tested this locally to confirm I can delete an organisation that I was previously unable to delete due to the related sentry issue.

### DIFF
--- a/api/organisations/models.py
+++ b/api/organisations/models.py
@@ -458,6 +458,10 @@ class OrganisationSubscriptionInformationCache(LifecycleModelMixin, models.Model
         self.organisation.api_usage_notifications.all().delete()
 
     def reset_to_defaults(self):
+        """
+        Resets all limits and CB related data to the defaults, leaving the
+        usage data intact.
+        """
         self.current_billing_term_starts_at = None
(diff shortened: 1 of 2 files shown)


novakzaballa (approved):
LGTM

PR #4522, opened 2024-08-20
chore: update org cancelled alert to only send to customersuccess@flagsmith.com

matthewelwell:
## Changes

Updates email alert when a SaaS organisation cancels their subscription to only send to specific email addresses. Note that this information is already available via Chargebee alerts in Slack. 

Updated existing tests


novakzaballa (commented):
LGTM, but there are a couple of tests that should no longer validate emails.
For example:
https://github.com/Flagsmith/flagsmith/blob/33074f349e24ea06cede711e797d941c0bc042c4/api/tests/unit/organisations/test_unit_organisations_views.py#L620

matthewelwell:
@novakzaballa I updated this based on the conversation in slack [here](https://flagsmith.slack.com/archives/CTH7DCCQL/p1724171107341469?thread_ts=1724161406.470739&cid=CTH7DCCQL), please can you re-review?

novakzaballa (approved):

PR #4525, opened 2024-08-21
fix: audit logs generation for feature state value

gagantrivedi:

## Changes

- Catch ObjectDoesNotExists if the feature state is deleted
- Don't pass null to diff_history if the prev history record does not exists

Updates test cases

### DIFF
--- a/api/tests/unit/features/test_unit_features_models.py
+++ b/api/tests/unit/features/test_unit_features_models.py
@@ -698,7 +698,7 @@ def test_feature_state_value_get_skip_create_audit_log_if_environment_feature_ve
 
 
 def test_feature_state_value__get_skip_create_audit_log_for_deleted_feature_state(
-    feature: Feature, feature_segment, environment
+    feature: Feature, feature_segment: FeatureSegment, environment: Environment
 ):
     # Give
     feature_state = FeatureState.objects.create(
--- a/api/tests/unit/features/test_unit_features_models.py
+++ b/api/tests/unit/features/test_unit_features_models.py
@@ -11,7 +11,12 @@ from pytest_mock import MockerFixture
 from environments.identities.models import Identity
 from environments.models import Environment
 from features.constants import ENVIRONMENT, FEATURE_SEGMENT, IDENTITY
-from features.models import Feature, FeatureSegment, FeatureState
+from features.models import (
+    Feature,
+    FeatureSegment,
+    FeatureState,
+    FeatureStateValue,
+)
 from features.versioning.models import EnvironmentFeatureVersion
 from features.workflows.core.models import ChangeRequest
 from projects.models import Project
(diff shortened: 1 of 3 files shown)


api/tests/unit/features/test_unit_features_views.py
```diff
@@ -2322,6 +2322,39 @@ def test_cannot_update_environment_of_a_feature_state(
...
+        "environment": environment.id,
+        "identity": None,
+        "feature_segment": None,
+    }
+    # When
+    response = admin_client_new.put(
+        url, data=json.dumps(data), content_type="application/json"
+    )
+
+    assert response.status_code == status.HTTP_200_OK
```
```suggestion

    # Then
    assert response.status_code == status.HTTP_200_OK
```

api/tests/unit/features/test_unit_features_models.py
```diff
@@ -692,6 +697,24 @@ def test_feature_state_value_get_skip_create_audit_log_if_environment_feature_ve
     assert feature_state.feature_state_value.get_skip_create_audit_log() is True
 
 
+def test_feature_state_value_get_skip_create_audit_log_for_deleted_feature_state(
```
matthewelwell: Based on the codecov report, it looks like this isn't actually testing the codepath that you expect.

api/tests/unit/features/test_unit_features_views.py
```diff
@@ -2322,6 +2322,39 @@ def test_cannot_update_environment_of_a_feature_state(
     )
 
 
+def test_update_feature_state_without_history_of_fsv(
```
matthewelwell, line 2325: Where was the error coming from previously? In the audit log generation? And this test is sufficient I guess because `TASK_RUN_METHOD` is `SYNCHRONOUS`?

gagantrivedi: > Where was the error coming from previously

The error was coming from ` get_update_log_message`

> And this test is sufficient I guess because TASK_RUN_METHOD is SYNCHRONOUS?

Yep, that is correct

api/tests/unit/features/test_unit_features_models.py
```diff
@@ -692,6 +697,28 @@ def test_feature_state_value_get_skip_create_audit_log_if_environment_feature_ve
     assert feature_state.feature_state_value.get_skip_create_audit_log() is True
 
 
+def test_feature_state_value__get_skip_create_audit_log_for_deleted_feature_state(
+    feature: Feature, feature_segment, environment
+):
```
matthewelwell:
```suggestion
def test_feature_state_value__get_skip_create_audit_log_for_deleted_feature_state(
    feature: Feature, feature_segment: FeatureSegment, environment: Environment
) -> None:
```

matthewelwell (approved):

PR #4526, opened 2024-08-21
feat: usage period filter

### LINKED ISSUE #3406: Organisation usage charts should show data for current billing period
Conversation for context: https://github.com/Flagsmith/flagsmith/pull/3340#pullrequestreview-1872606131. 

Currently the Organisation usage charts show the numbers for the previous 30 days. We are making changes in the PR linked above to send customers alerts when they hit certain thresholds of their API calls limits. Based on this, we'll need to add information to this screen which shows the data for the current billing period. 

image

I suggest that we add some kind of switcher which allows you to switch between current billing period and then the previous 30 days. The charts should display exactly the same, but the numbers should change (in 99% of cases).

kyle-ssg:

## Changes

image

- For paid projects, add a usage filter.
- Add ability to select api call types

As above

### DIFF
--- a/frontend/common/types/requests.ts
+++ b/frontend/common/types/requests.ts
@@ -32,6 +32,11 @@ export const billingPeriods = [
     value: 'previous_billing_period',
   },
   { label: 'Last 90 days', value: '90_day_period' },
+  { label: 'Last 30 days', value: undefined },
+]
+export const freePeriods = [
+  { label: 'Last 90 days', value: '90_day_period' },
+  { label: 'Last 30 days', value: undefined },
 ]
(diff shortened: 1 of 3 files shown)


matthewelwell:
We should get rid of / update this alert too

image

matthewelwell:
Since each of these requests triggers a request to influx (that we pay for), we should also optimise this: 

image

We're making a request to the root endpoint, and then with the `period=current_billing_period` query parameter set when we first load the page (where the default is current_billing_period).

novakzaballa (approved):

matthewelwell (approved):

PR #4529, opened 2024-08-21
fix: usage and analytics data duplicates the current day

matthewelwell:
## Changes

This code fixes an issue where the current day was duplicated in the usage / analytics graphs and all other days seemed to be off by one. 

The reason for this was that the influx query was using the end of the aggregation window to use as the key for the data. This PR updates all uses of `aggregateWindow` in the code to use the start of the window as the key. 

Note that this will mean that we get e.g. 91 results when asking for the last 90 days because it's essentially requesting the exact last (90 x 24h), so it will include today and the same chunk of the first day in the period. Without overcomplicating the query in Influx this would be difficult to solve and I think is perfectly acceptable for now. 

Reference: https://docs.influxdata.com/flux/v0/stdlib/universe/aggregatewindow/#timesrc

I have updated the (mocked) tests, but obviously this isn't very useful. 

@zachaysan and I ran the code against influx directly from my laptop and verified that the data was correct.

### DIFF
--- a/api/organisations/views.py
+++ b/api/organisations/views.py
@@ -146,7 +146,7 @@ class OrganisationViewSet(viewsets.ModelViewSet):
 
     @swagger_auto_schema(
         deprecated=True,
-        operation_description="Please use ​​/api​/v1​/organisations​/{organisation_pk}​/usage-data​/total-count​/",
+        operation_description="Please use /api/v1/organisations/{organisation_pk}/usage-data/total-count/",
     )
     @action(
         detail=True,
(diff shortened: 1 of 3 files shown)


api/organisations/views.py
```diff
@@ -221,7 +221,7 @@ def get_hosted_page_url_for_subscription_upgrade(self, request, pk):
 
     @swagger_auto_schema(
         deprecated=True,
-        operation_description="Please use ​​/api​/v1​/organisations​/{organisation_pk}​/usage-data​/",
+        operation_description="Please use /api/v1/organisations/{organisation_pk}/usage-data/",
```
khvn26, line 224: What are these changes?

khvn26 (approved):
LGTM apart from a couple of noop changes (whitespace?)

api/organisations/views.py
matthewelwell: Ah yes, some sneaky secret characters that I wanted to get rid of so snuck into this PR. 

This is the diff in my IDE

image

PR #4535, opened 2024-08-23
fix(ldap-login): create custom serializer to fix login field

gagantrivedi:

## Changes
- Add a custom login serializer to make login field configurable because LDAP supports login using username/password instead of email/password 

Adds a unit test

### DIFF
--- a/api/custom_auth/serializers.py
+++ b/api/custom_auth/serializers.py
@@ -1,5 +1,5 @@
 from django.conf import settings
-from django.contrib.auth import authenticate
+from djoser.conf import settings as djoser_settings
 from djoser.serializers import TokenCreateSerializer, UserCreateSerializer
 from rest_framework import serializers
 from rest_framework.authtoken.models import Token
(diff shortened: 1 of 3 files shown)


api/custom_auth/serializers.py
```diff
@@ -17,6 +18,23 @@
 )
 
 
+class CustomTokenCreateSerializer(TokenCreateSerializer):
+    def validate(self, attrs):
```
matthewelwell, line 41: We seem to have missed [these lines](https://github.com/sunscrapers/djoser/blob/master/djoser/serializers.py#L124-L127) in overriding this function, was that intentional?

matthewelwell: Wouldn't a better way to do this be something like the following?  

```python
    def validate(self, attrs):
        if settings.LOGIN_FIELD != FFAdminUser.USERNAME_FIELD:
            attrs[settings.LOGIN_FIELD] = attrs.pop(FFAdminUser.USERNAME_FIELD]
        return super().validate(attrs)
```

Maybe we can even be more specific and check for LDAP being installed or something? This serializer could also technically live in the `flagsmith-ldap` module and we just update `REST_FRAMEWORK["SERIALIZERS"]["token_create"]` if LDAP is installed (although this would mean having to add a dependency on djoser to `flagsmith-ldap` I guess).

The main reason that I prefer the approach above though is that it means we don't have to worry if we upgrade djoser in the future for example as we're still calling the super validate method.

gagantrivedi: I agree with everything except the statement, "This serializer could also technically live in the `flagsmith-ldap` module." While I agree that we only have to do this because of LDAP, I don't think it's something that LDAP should concern itself with. It should be the responsibility of the application to match the interface of the library (LDAP) and not the other way around.
It also allows us to add more authentication backends in the future that use a username or something else.

gagantrivedi: > We seem to have missed [these lines](https://github.com/sunscrapers/djoser/blob/master/djoser/serializers.py#L124-L127) in overriding this function, was that intentional?

Yep, I don't think that part is relevant for us

matthewelwell: Yep, fair enough 👍

gagantrivedi: > Wouldn't a better way to do this be something like the following?

No, that won't work because of [this line](https://github.com/sunscrapers/djoser/blob/master/djoser/serializers.py#L116) if you change `settings.LOGIN_FIELD`(in djoser settings) that serializer will just ignore the email field 
ps: In the current implementation, we are overriding djoser settings instead we are updating our own `settings`

gagantrivedi: > The main reason that I prefer the approach above though is that it means we don't have to worry if we upgrade djoser in the future for example as we're still calling the super validate method.

I think we still have to, e.g: if they change validate method? And we are overriding djoser at a lot of places, so we would absolutely have to check every update

matthewelwell: Ok, so just override `__init__` as well?

gagantrivedi: > Ok, so just override `__init__` as well?

Then we pretty much loose all of this? 

```
The main reason that I prefer the approach above though is that it means we don't have to worry if we upgrade djoser in the future for example as we're still calling the super validate method.
```

matthewelwell: Well, no, we can just call super?

gagantrivedi: But, there is not much value left in doing that anymore?

matthewelwell: I don't really see how [this](https://github.com/Flagsmith/flagsmith/compare/fix/ldap-login...fix/ldap-login-v2) isn't better than what we have here?

Yes, djoser changes may cause issues in the future, but I think we'd know about them via our tests vs just re-implementing the code as we have here. In this implementation, I'm worried that e.g. security changes would slip through the net.

api/custom_auth/serializers.py
```diff
@@ -17,6 +18,33 @@
 )
 
 
+class CustomTokenCreateSerializer(TokenCreateSerializer):
+    """
+    NOTE: Some authentication backends (e.g., LDAP) support only
+    username and password authentication. However, the front-end
+    currently sends the email as the login key. To accommodate
+    this, we override the serializer to rename the email field
+    to the username(or any other field configurable using djoser settings) field.
```
matthewelwell: Shouldn't this be the other way around? 

```suggestion
    this, we override the serializer to rename the username field
    to the email (or any other field configurable using djoser settings) field.
```

matthewelwell (approved):

PR #4539, opened 2024-08-26
feat: Add hubspot cookie tracking

### LINKED ISSUE #3855: Original source > HubSpot
We send contacts to HubSpot via the API, so they will always come up as original source = Offline. This means that all app signups don't have original source attribution in HubSpot, which affects tracking and ad performance for our paid ads. 

Can we either:
1. Implement tracking in the app domain so that their source can send across (right now when users reach the app they become "invisible" to HubSpot) OR
2. Capture and send their GCLID and their UTM values to HubSpot

More info and a Slack thread: https://flagsmith.slack.com/archives/C06D7F82WF7/p1714390077596589?thread_ts=1714058518.960549&cid=C06D7F82WF7 

Another Slack thread for more info:
https://flagsmith.slack.com/archives/C06D7F82WF7/p1713357279287279

zachaysan:
## Changes

Fixes https://github.com/Flagsmith/flagsmith/issues/3855

This change adds in a form in the background of a Hubspot contact creation to add in the special Hubspot tracking cookie to give extra visibility in our user's journey in places like the activity tab of the Hubspot dashboard.

Manually tested it with my own Hubspot cookie and I added programatic tests after gaining confidence that it worked.


api/integrations/lead_tracking/hubspot/client.py
```diff
@@ -42,6 +49,51 @@ def get_contact(self, user: FFAdminUser) -> None | dict:
 
         return results[0]
 
+    def create_lead_form(self, user: FFAdminUser, hubspot_cookie: str) -> dict:
```
khvn26: nit: we should avoid empty `dict` annotations. Depending on our needs, we should either come up with a `TypedDict` or use `dict[str, Any]`.

api/organisations/invites/views.py
```diff
@@ -44,6 +46,15 @@ def join_organisation_from_email(request, hash):
...
 
+    hubspot_cookie = request.COOKIES.get(HUBSPOT_COOKIE_NAME)
+
+    if hubspot_cookie:
+        if not HubspotTracker.objects.filter(user=request.user).exists():
+            HubspotTracker.objects.create(
+                user=request.user,
+                hubspot_cookie=hubspot_cookie,
+            )
+
```
khvn26: I'd like to avoid code duplication, which occurs further in `join_organisation_from_link`. This code can be extracted to either of the following:
- A `HubspotTracker` class method
- A function in `services.py` — that's how I'd do it, see existing `services.py` modules for reference

api/organisations/invites/views.py
```diff
@@ -62,6 +73,15 @@ def join_organisation_from_link(request, hash):
...
         raise InviteExpiredError()
 
+    hubspot_cookie = request.COOKIES.get(HUBSPOT_COOKIE_NAME)
+
+    if hubspot_cookie:
+        if not HubspotTracker.objects.filter(user=request.user).exists():
+            HubspotTracker.objects.create(
+                user=request.user,
+                hubspot_cookie=hubspot_cookie,
+            )
```
khvn26: See [here](https://github.com/Flagsmith/flagsmith/pull/4539/files#r1735097706).

api/organisations/views.py
```diff
@@ -109,6 +111,15 @@ def create(self, request, **kwargs):
...
         Override create method to add new organisation to authenticated user
         """
+        hubspot_cookie = request.COOKIES.get(HUBSPOT_COOKIE_NAME)
+
+        if hubspot_cookie:
+            if not HubspotTracker.objects.filter(user=request.user).exists():
+                HubspotTracker.objects.create(
+                    user=request.user,
+                    hubspot_cookie=hubspot_cookie,
+                )
```
khvn26: See [here](https://github.com/Flagsmith/flagsmith/pull/4539/files#r1735097706).

api/organisations/invites/views.py
```diff
@@ -44,6 +46,15 @@ def join_organisation_from_email(request, hash):
         error_data = {"detail": str(e)}
         return Response(data=error_data, status=status.HTTP_400_BAD_REQUEST)
 
+    hubspot_cookie = request.COOKIES.get(HUBSPOT_COOKIE_NAME)
+
+    if hubspot_cookie:
+        if not HubspotTracker.objects.filter(user=request.user).exists():
+            HubspotTracker.objects.create(
```
khvn26: Do we need to update the stored cookie in case it's regenerated for some reason?

khvn26 (changes requested):
I've got a nitpick, a code improvement proposal, and a question.

api/integrations/lead_tracking/hubspot/client.py
zachaysan: Good idea. Implemented.

api/organisations/invites/views.py
zachaysan: Sure this is easy.

api/organisations/invites/views.py
zachaysan: Sure, I've switched the code to an `update_or_create` method to handle that if it ever comes up.

api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -58,6 +58,12 @@ def create_lead(self, user: FFAdminUser, organisation: Organisation = None) -> N
 
         HubspotLead.objects.create(user=user, hubspot_id=response["id"])
 
+        if HubspotTracker.objects.filter(user=user).exists():
```
gagantrivedi: nit: we can use walrus here and get rid of the exists

gagantrivedi (approved):

api/integrations/lead_tracking/hubspot/lead_tracker.py
zachaysan: Nice catch 👍🏼

api/integrations/lead_tracking/hubspot/client.py
```diff
@@ -42,7 +50,56 @@ def get_contact(self, user: FFAdminUser) -> None | dict:
...
+            "Content-Type": "application/json",
+        }
+        url = f"{HUBSPOT_ROOT_FORM_URL}/{HUBSPOT_PORTAL_ID}/{HUBSPOT_FORM_ID}"
+
+        response = requests.post(url, headers=headers, data=json.dumps(payload))
+
+        if response.status_code not in {200, 201}:
+            logger.error(
+                f"Problem posting data to Hubspot's form API due to {response.status_code} "
+                f"status code and following form data {response.text}"
```
khvn26:
```suggestion
                f"status code and following response: {response.text}"
```

khvn26 (approved):

PR #4547, opened 2024-08-28
feat: restrict versioning by days

kyle-ssg:

## Changes

Adjusts limiting of version history to date rather than number of versions

Manually set limitation in code against https://pr-4433-deployment-54847-flagsmith.app.uffizzi.com/

Disabled links in audit log 

image

Disabled versions in history 

image


novakzaballa (approved):
LGTM from a code perspective

matthewelwell:
I guess it's good to have the code there, but in theory we should never need the logic to blank out the link from the audit log, because the audit log visibility is also restricted for non-enterprise plans.

matthewelwell:
Some issues I've found in testing:

1. Since I'm running this self hosted, I would have expected the plan next to 'Version History' to say 'Enterprise' here, like it does for Audit Logs. 

image

image

2. The only time that the API will return versions older than the limit is when the current live version is older than the limit. We shouldn't really blur this out. 

image

If we want to blur things out, I could perhaps return a field that states that there are older versions, but I don't think this is necessary. I think we should just let the API handle the restrictions here.

matthewelwell (changes requested):
See feedback in comment above.

kyle-ssg:
@matthewelwell this has been resolved

kyle-ssg:
Note: due to recent PR merge this can now be found within the feature modal 
image

frontend/web/components/AuditLog.tsx
```diff
@@ -96,17 +103,47 @@ const AuditLog: FC<AuditLogType> = (props) => {
...
+              />
+            </div>
+          }
         >
-          <Button theme='text'>View version</Button>
-        </Link>
+          {isOverLimit
+            ? `<div>
+              Unlock your feature's entire history.<br/>Currently limited to${' '}
+              <strong>${versionLimitDays} days</strong>.
```
rolodato, line 145: The wording here could be interpreted as unlocking this specific feature's history and not history on all features - maybe could be reworded as:

> Upgrade to access more than 7 days of feature history.

This avoids using negative terms (lock, limit) in this CTA and is generic enough to work for potential plans that might have more generous, but not unlimited, limits.

frontend/web/components/AuditLog.tsx
```diff
@@ -96,17 +103,47 @@ const AuditLog: FC<AuditLogType> = (props) => {
...
+              {isOverLimit ? (
+                VersionButton
+              ) : (
+                <Link
+                  to={`/project/${project.id}/environment/${environment.api_key}/history/${related_object_uuid}/`}
+                >
+                  {VersionButton}
+                </Link>
+              )}
+              <PlanBasedBanner
```
rolodato, line 134: I know it's not related to this PR, but this button has too low contrast in dark mode:

image

rolodato (commented):
I briefly played around with this but wasn't able to set up a fully representative local environment to test looking at older audit log entries or version history - is there a quick way I could do that?

kyle-ssg:
> I briefly played around with this but wasn't able to set up a fully representative local environment to test looking at older audit log entries or version history - is there a quick way I could do that?

q for @matthewelwell

kyle-ssg:
@rolodato resolved button colour in dark mode. I think we're good to go on this ?

frontend/web/components/AuditLog.tsx
```diff
@@ -96,17 +103,47 @@ const AuditLog: FC<AuditLogType> = (props) => {
     })
     const colour = index === -1 ? 0 : index
     let link: ReactNode = null
-    if (
-      related_object_uuid &&
-      related_object_type === 'EF_VERSION' &&
-      environment
-    ) {
+    const date = moment(created_date)
```
rolodato, line 106: We're parsing `created_date` into a `moment` date here, and also here in `isVersionOverLimit`: https://github.com/Flagsmith/flagsmith/pull/4547/files#diff-e181306c527188403c35874ca85e95550ddaefd2c59a0b3f71dc4603e1a1fd1cR305

matthewelwell:
> > I briefly played around with this but wasn't able to set up a fully representative local environment to test looking at older audit log entries or version history - is there a quick way I could do that?
> 
> q for @matthewelwell

Sorry @rolodato - I missed this one. The BE has now been merged if that's helpful. In order to manually change the limit though you'll need to directly edit the `organisations_organisationsubscriptioninformationcache` table in the database and set the relevant fields (which should be fairly self explanatory). 

Note that if you're using an organisation on the scale up plan you will also need to set the `VERSIONING_RELEASE_DATE` environment variable in the API to a date in the past to work around the logic which grandfathers unlimited audit log history for existing scale up customers.

matthewelwell (approved):
This looks good to me from a versioning perspective. 

The only thing I would add is that we should include an upgrade banner on the Audit Log page stating how much Audit Log history the organisation has access to. This is also available in the `audit_log_visibility_days` attribute from the subscription metadata endpoint.

matthewelwell (approved):
Re-tested manually. Looks good to me.

PR #4554, opened 2024-08-28
feat: Add sane defaults for segment_operators, integration_data, saml flags

rolodato:

## Changes

Add default values to the FoF flags `segment_operators`, `integration_data`, and `saml`. This avoids the following scenario:

1. Customer is setting up self-hosted environment, everything works fine
2. FoF environment is created with only a few specific flags that the customer wants to modify
3. Creating segments and trying to use the "Single Sign-On" button at the login page no longer works, unless the `segment_operators` and `saml` flags are created with the correct values.

The flags all still work as usual, but don't break anything if they don't exist in a FoF environment.

`integration_data` is a bonus to have integrations work out of the box.

In a local environment, set up FoF but do not create any flags. These should still work:

* Creating segments
* Viewing and creating integrations
* Clicking the "Single Sign-On" button in the home page


frontend/web/components/pages/HomePage.js
```diff
@@ -221,7 +221,10 @@ const HomePage = class extends React.Component {
         )
       }
 
-      if (Utils.getFlagsmithHasFeature('saml')) {
+      if (
+        !Utils.flagsmithFeatureExists('saml') ||
```
matthewelwell, line 225: Hmm... maybe this is fine, but I guess this will enable the 'Single Sign-On' button (admittedly it won't work) for pretty much all self hosted customers that _have_ set up Flagsmith on Flagsmith (I'm intrigued to know how many this actually is). 

I'm a little concerned that the experience for any curious open source self hosters will be pretty bad as well, because I don't think the endpoint will even exist for the call made when adding in the organisation name via 'Single Sign-On'.

matthewelwell (commented):
I've added one specific comment, but I'm also interested to know if / how we plan to keep the defaults in sync with any changes to the flags in production. 

I'm also keen to know @kyle-ssg 's opinion on this of course.

frontend/web/components/pages/HomePage.js
rolodato: This is the current experience:

* Out of the box, "Single Sign-On" is shown, since all self-hosted images connect to SaaS FoF by default, which has the `saml` flag enabled
* If you set up FoF with no flags, "Single Sign-On" disappears until you create and enable the `saml` flag

After this change:

* Out of the box, "Single Sign-On" is shown as before
* If you set up FoF with no flags, "Single Sign-On" is still shown until you create and disable the `saml` flag

In any case, the error that you get when trying to use SSO is "Please check your organisation name and try again." despite the 405 response, so I'm not too worried about this. We could use the SSO button as an upsell opportunity, the same way we do now for features that are gated behind plans. Even without adding an explicit upsell action item, it still advertises that SSO exists is good enough without being confusing IMO.

rolodato:
> I'm also interested to know if / how we plan to keep the defaults in sync with any changes to the flags in production.

* `segment_operators`: I see 2 options here:
  1. Keep making manual changes to the hardcoded defaults at the same time we make changes to production. Changes to segment operators are rare enough that this is probably fine
  2. Deprecate this flag and plan to remove it. This seems awkward to do given the dependency on production SaaS - one way I could see this happening is by passing in a transient `version` trait while fetching FoF flags, and creating a segment override on `segment_operators` that removes its value if this trait `version` is set. This keeps old instances working, lets users customise `segment_operators` on their own FoF, and frees us from needing to make changes to this flag again.
* `integration_data`: Same options as for `segment_operators`
* `saml`: This flag will probably never be changed in production

matthewelwell (approved):

PR #4561, opened 2024-08-30
fix(github-4555): use api_key name for changed_by

### LINKED ISSUE #4555: AttributeError: 'APIKeyUser' object has no attribute 'id'
Sentry Issue: [FLAGSMITH-API-4S6](https://flagsmith.sentry.io/issues/5628601594/?referrer=github_integration)

```
AttributeError: 'APIKeyUser' object has no attribute 'id'
(6 additional frame(s) were not displayed)
...
  File "edge_api/identities/serializers.py", line 192, in save
    "changed_by_user_id": request.user.id,
```

gagantrivedi:

## Changes

Fix the issue with updating feature states for edge identities using the API key by setting the changed_by webhook field to the name of the key.

Updates unit tests

### DIFF
--- a/api/edge_api/identities/models.py
+++ b/api/edge_api/identities/models.py
@@ -6,7 +6,7 @@ from django.db.models import Prefetch, Q
 from flag_engine.features.models import FeatureStateModel
 from flag_engine.identities.models import IdentityFeaturesList, IdentityModel
 
-from api_keys.models import MasterAPIKey
+from api_keys.user import APIKeyUser
 from edge_api.identities.tasks import (
     generate_audit_log_records,
     sync_identity_document_features,
@@ -165,26 +165,22 @@ class EdgeIdentity:
         with suppress(ValueError):  # ignore if feature state didn't exist
             self._engine_identity_model.identity_features.remove(feature_state)
 
-    def save(self, user: FFAdminUser = None, master_api_key: MasterAPIKey = None):
+    def save(self, user: FFAdminUser | APIKeyUser = None):
         self.dynamo_wrapper.put_item(self.to_document())
         changeset = self._get_changes()
         self._update_feature_overrides(
             changeset=changeset,
             user=user,
-            master_api_key=master_api_key,
         )
         self._reset_initial_state()
 
-    def delete(
-        self, user: FFAdminUser = None, master_api_key: MasterAPIKey = None
-    ) -> None:
+    def delete(self, user: FFAdminUser | APIKeyUser = None) -> None:
         self.dynamo_wrapper.delete_item(self._engine_identity_model.composite_key)
         self._engine_identity_model.identity_features.clear()
         changeset = self._get_changes()
         self._update_feature_overrides(
             changeset=changeset,
             user=user,
-            master_api_key=master_api_key,
         )
         self._reset_initial_state()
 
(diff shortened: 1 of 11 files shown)


api/edge_api/identities/tasks.py
```diff
@@ -40,10 +41,14 @@ def call_environment_webhook_for_feature_state_change(
         return
 
     feature = Feature.objects.get(id=feature_id)
-    changed_by = FFAdminUser.objects.get(id=changed_by_user_id)
+    match changed_by_user_id:
+        case str():
+            changed_by = MasterAPIKey.objects.get(id=changed_by_user_id).name
+        case _:
+            changed_by = FFAdminUser.objects.get(id=changed_by_user_id).email
```
khvn26: This is too hacky for my taste.

Is there a reason we can't add a new `changed_by_user_type` task argument?

zachaysan: I agree with @khvn26 on this.

api/tests/unit/edge_api/identities/test_edge_api_identities_serializers.py
```diff
@@ -51,15 +54,27 @@ def test_edge_identity_feature_state_serializer_save_allows_missing_mvfsvs(
...
+        lazy_fixture("admin_user"),
+    ],
+)
 def test_edge_identity_feature_state_serializer_save_calls_webhook_for_new_override(
-    mocker, identity, feature, admin_user
+    mocker,
+    identity,
+    feature,
+    user,
+    rf: RequestFactory,
```
zachaysan: Missing typing

api/edge_api/identities/tasks.py
gagantrivedi: This code doesn't exist anymore

zachaysan (approved):
Overall looks good, just a couple of comments.

api/tests/unit/edge_api/identities/test_edge_api_identities_serializers.py
gagantrivedi: Done

api/edge_api/identities/serializers.py
```diff
@@ -189,7 +186,7 @@ def save(self, **kwargs):
                 "environment_api_key": identity.environment_api_key,
                 "identity_id": identity.id,
                 "identity_identifier": identity.identifier,
-                "changed_by_user_id": request.user.id,
+                "changed_by": str(request.user),
```
matthewelwell, line 189: @gagantrivedi changing the signature here will be a problem during the deployment because the old task processor instances will not be able to execute the function with the updated arguments. We need to resolve this please in a new PR. 

I suggest that we just add a new argument `changed_by` to the task function and update the calling code to only send `changed_by_user_id` if `request.user` is an `FFAdminUser` and then ignore that field in the actual task handler. 

Note: we should probably consider adding some kind of versioning logic to the task processor to handle this in the future too.

api/edge_api/identities/models.py
```diff
@@ -200,23 +196,21 @@ def to_document(self) -> dict:
...
-                    "user_id": getattr(user, "id", None),
-                    "changes": changeset,
-                    "identity_uuid": str(self.identity_uuid),
-                    "master_api_key_id": getattr(master_api_key, "id", None),
-                }
-            )
+            kwargs = {
+                "environment_api_key": self.environment_api_key,
+                "identifier": self.identifier,
+                "user_id": getattr(user, "id", None),
```
matthewelwell, line 206: This will start to fail in the future if someone adds the `id` attribute to the `APIKeyUser` class. Will a test pick it up if so? Or should we be more explicit with this check? e.g. 

```
"user_id": user.id if not getattr(user, "is_master_api_key_user", False) else None
```

PR #4566, opened 2024-08-31
docs: Add docs on pricing and API requests. Improve real-time docs

rolodato:
Now with Visuals™

https://docs-git-docs-api-usage-flagsmith.vercel.app/pricing/

https://docs-git-docs-api-usage-flagsmith.vercel.app/advanced-use/real-time-flags


docs/docs/pricing/index.mdx
```diff
@@ -0,0 +1,105 @@
...
+
+From this page, you can break down usage per project and per environment.
+
+The following screenshot shows an example usage graph from the Flagsmith dashboard:
+
+![img.png](usage-graph.png)
+
+## What counts as an API request?
+
+Any operation that calls Flagsmith's API to retrieve information about the current state of flags counts as 1 billable
```
matthewelwell: Hmm... this isn't completely true. We should also include something about identifying an identity and/or updating an identity's traits.

docs/docs/pricing/index.mdx
```diff
@@ -0,0 +1,105 @@
...
+1. A user visits the application for the first time. The user's device calls Flagsmith to get the flags for the
+   production environment (1 API request).
+2. The user logs in to their account. Now that we know the user's identity, their device calls Flagsmith to get the
+   flags for this specific user, taking into account their
+   [A/B tests or progressive rollouts](/advanced-use/ab-testing), [segments](/basic-features/segments) and user-specific
+   flags (1 API request). Any user data sent by your application is stored by Flagsmith as traits.
+3. The user performs some action that should update their stored data in Flagsmith. The device calls Flagsmith to update
+   this user's traits and receives their latest flags in return (1 API request).
+
+Because these API requests were made using Remote Evaluation, they would all send data to any analytics
```
matthewelwell: This isn't necessarily true. Probably the first request when the user is unauthenticated will not send any event.

matthewelwell (changes requested):
A couple of minor comments.

docs/docs/pricing/index.mdx
rolodato: Identifying an identity still returns the current state of flags - it's the current state of flags as seen by a specific user.

Updating an identity's traits also returns the current state of flags for the identity that is being updated.

I tried to make this as generic as possible while not overwhelming with technical details, considering the audience for this document includes people who have not used Flagsmith at all.

rolodato: I also tried to make this explicit in the example below:

> The device calls Flagsmith to update this user's traits and receives their latest flags in return (1 API request).

docs/docs/pricing/index.mdx
rolodato: True - looking back, I don't know why I've included this, it's not relevant to knowing how API usage is calculated. I'll remove it

docs/docs/pricing/index.mdx
matthewelwell: I think we can solve any ambiguity here with a minor change to the wording: 

```suggestion
Any operation that calls Flagsmith's API to retrieve information about the current state of flags and/or persist an identity or their traits counts as 1 billable
```

rolodato: I think having this level of technical detail in the first sentence does more harm than good. We don't need to introduce the concepts of identity or trait at this point - I've specifically avoided using the word "identity" as a Flagsmith concept in this doc to make it accessible for a wider audience, especially less-technical decision makers who don't care about these details.

In a separate PR, I could add a dedicated section with details on what actual endpoints are billed, things like `/flags`, `/identities`, `/environment-document` etc to detail the actual specifics.

rolodato: Another way to word this that is technically more accurate and doesn't add complexity could be:

```suggestion
Any operation that calls Flagsmith's API and retrieves information about the current state of flags counts as 1 billable
```

matthewelwell: Yep, good with me, apart from one minor typo: `persist` -> `persists`

matthewelwell (approved):

PR #4569, opened 2024-09-02
feat: search identities by dashboard alias

matthewelwell:
## Changes

Adds functionality to search edge identities by a dashboard alias. 

Note: requires the relevant index to be created on the `flagsmith_identities` table in dynamodb. 

I have manually tested this PR by running the API pointed to staging DB and dynamo tables, running the FE and testing searching for `dashboard_alias:...`. 

I have also added some new tests to the python unit tests for specific use cases.


api/edge_api/identities/views.py
```diff
@@ -97,44 +95,43 @@ def initial(self, request, *args, **kwargs):
...
-                "EQUAL"
-            ], search_query.replace('"', "")
-        return self.dynamo_identifier_search_functions["BEGINS_WITH"], search_query
+    def get_serializer_class(self):
+        if self.action in ("update", "partial_update"):
+            return EdgeIdentityUpdateSerializer
+        return EdgeIdentitySerializer
 
     def get_object(self):
+        # TODO: should this return an EdgeIdentity object instead of a dict?
```
matthewelwell, line 105: I'm leaving this here as a todo because I didn't want to bloat this PR with this work, but I do think this should be refactored.

zachaysan: Maybe make a GitHub issue for it? I find code TODOs often get forgotten about over time.

api/pyproject.toml
```diff
@@ -142,7 +142,7 @@ environs = "~9.2.0"
 django-lifecycle = "~1.0.0"
 drf-writable-nested = "~0.6.2"
 django-filter = "~2.4.0"
-flagsmith-flag-engine = "^5.1.1"
+flagsmith-flag-engine = { git = "https://github.com/Flagsmith/flagsmith-engine", branch = "feat/add-dashboard-alias"}
```
zachaysan: Is this change meant for the merge with main or is it just a WIP change?

zachaysan (approved):
Looks great 🥇

api/pyproject.toml
matthewelwell: Sorry, no, good catch. I need [this PR](https://github.com/Flagsmith/flagsmith-engine/pull/210) merged and released before we merge this one.

api/edge_api/identities/views.py
matthewelwell: #4578

api/pyproject.toml
matthewelwell: This is done now.

zachaysan (approved):

kyle-ssg:
@matthewelwell Is there a way to make this an or filter (maybe it is and I'm missing how it works)? The reason I ask is that it'd be super clunky to have to specify whether we're searching by id or alias especially in the case where I only have a few aliases and the rest of the time I search by id.

image

matthewelwell:
> @matthewelwell Is there a way to make this an or filter (maybe it is and I'm missing how it works)? The reason I ask is that it'd be super clunky to have to specify whether we're searching by id or alias especially in the case where I only have a few aliases and the rest of the time I search by id.
> 
> image

We'll, yes, technically I could. But I think the UX _should_ force the user to choose what they're searching on. The two things will always be very different fields anyway, so I'm not sure what your concern is?

kyle-ssg:
I think it's just annoying/clunky to have to select what to search on vs just typing a value, I don't think they should really have to tell flagsmith they are wanting to search for an alias vs id especially given that they'd have very different values. 

I.e. if I type kyle id always want to be searching for an alias, if I paste an id I'd always want to search by id. In both those cases I think I'd often see no results then realise I have to change a filter.

matthewelwell:
@kyle-ssg as discussed on slack, adding search across both entities is technically challenging on the API (since it requires sending 2 separate search requests to dynamo). Let's implement the selector for now and add this as an improvement.

PR #4580, opened 2024-09-04
docs: edge proxy health check configuration

matthewelwell:
## Changes

The main purpose of this PR is to add the corresponding documentation for [this edge proxy PR](https://github.com/Flagsmith/edge-proxy/pull/122). 

I have also slightly tweaked the formatting of the documentation (although it could still use some finessing at some point). 

Ran the docs locally and verified the changes.

### DIFF
--- a/docs/docs/deployment/hosting/locally-edge-proxy.md
+++ b/docs/docs/deployment/hosting/locally-edge-proxy.md
@@ -238,27 +238,32 @@ specified by the [`"logging.log_format"`](#logginglog_format) setting.
 ### Health Check
 
 The Edge Proxy exposes a health check endpoint at `/proxy/health` that responds with a 200 status code if it was able to
-fetch all its configured environment documents. By default, if any environment document could not be fetched during the
-latest poll, it will respond with a 500 status code. In some cases, you may want the Edge Proxy to succeed its health
-checks even if it failed to fetch one or more environment documents, but only if it these documents were successfully
-fetched at some point in the past. You can achieve this using the `environment_update_grace_period_seconds` setting
-defined below.
+fetch all its configured environment documents. By default, if any update of the configured environment documents takes
+longer than the allowed grace period (see below), then the health check will return with a 500 status code. In some
+cases, you may want the Edge Proxy to succeed its health checks even if it failed to fetch one or more environment
+documents, but only if it these documents were successfully fetched at some point in the past. You can achieve this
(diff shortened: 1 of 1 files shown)


docs/docs/deployment/hosting/locally-edge-proxy.md
```diff
@@ -229,7 +235,39 @@ Or, log access logs to file in generic format while logging everything else to s
...
+
+The health check can be configured depending on the use case for the Edge Proxy by adding the `health_check` object to
+the root of the settings file.
+
+```json
+{
+  ...
+  "health_check": {
+    "count_stale_documents_as_failing": true,
+    "grace_period_seconds": 30
```
rolodato: Are these the new defaults? Looking at https://github.com/Flagsmith/edge-proxy/pull/122/files#diff-4573c040951bc5c2676139dcf260c4a96b0f9b830f15227322c0a1f8545da004R104-R105

In which case these docs should mention how to change the defaults, only if needed

docs/docs/deployment/hosting/locally-edge-proxy.md
```diff
@@ -229,7 +235,39 @@ Or, log access logs to file in generic format while logging everything else to s
 When adding logger configurations, you can use the `"default"` handler which writes to stdout and uses formatter
 specified by the [`"logging.log_format"`](#logginglog_format) setting.
 
-### `config.json` example
+### Health Check
+
+The health check can be configured depending on the use case for the Edge Proxy by adding the `health_check` object to
```
rolodato: Suggest using active voice instead of passive voice https://developers.google.com/style/voice, and clarifying what the default behaviour is. Something like this:

```
The Edge Proxy exposes a health check endpoint at `/proxy/health` that responds with a 200 status code if was able to fetch all its configured environment documents. If any environment document could not be fetched during the latest poll, it will respond with a 500 status code.

In some cases, you may want the Edge Proxy to succeed its health checks even if it failed to fetch any environment document, but only if it these documents were successfully fetched at some point in the past. You can achieve this by...
```

(I was typing this suggestion until I saw this: https://github.com/Flagsmith/flagsmith/pull/4580/files#r1743739010 - this would need to be changed depending on what the new defaults are)

matthewelwell: Good feedback, thanks. I've updated the PR with changes that (I think) resolve both of your comments.

rolodato (approved):

PR #4586, opened 2024-09-04
fix: GitHub integration tagging issues

novakzaballa:

## Changes

Solve https://github.com/Flagsmith/flagsmith/issues/4585 and https://github.com/Flagsmith/flagsmith/issues/4584

Unit tests updated


api/integrations/github/github.py
```diff
@@ -55,18 +57,28 @@ def tag_feature_per_github_event(
         | Q(external_resources__type="GITHUB_ISSUE"),
         external_resources__url=metadata.get("html_url"),
     ).first()
+    repo: list[str] = repo_full_name.split("/")
```
matthewelwell: I think this is probably cleaner? Assuming that an organisation can't have a slash in it's name? 

```suggestion
    repository_owner, repository_name = repo_full_name.split("/", max_split=1)
```

api/integrations/github/github.py
```diff
@@ -55,18 +57,28 @@ def tag_feature_per_github_event(
...
+    tagging_enabled = (
+        GitHubRepository.objects.filter(
+            project=feature.project, repository_owner=repo[0], repository_name=repo[1]
+        )
+        .first()
+        .tagging_enabled
+    )
```
matthewelwell: Why use `.first()` here instead of `.get()` ? This could throw an `AttributeError` if a matching `GithubRepository` isn't returned - is this a valid use case? If so, we should actually make sure that we handle it correctly.

api/integrations/github/github.py
```diff
@@ -100,8 +112,10 @@ def handle_github_webhook_event(event_type: str, payload: dict[str, Any]) -> Non
         handle_installation_deleted(payload)
     elif event_type in tag_by_event_type:
         action = str(payload.get("action"))
-        metadata = payload.get("issue", {}) or payload.get("pull_request", {})
-        tag_feature_per_github_event(event_type, action, metadata)
+        if action in tag_by_event_type[event_type]:
+            repo_full_name = payload.get("repository", {}).get("full_name")
+            metadata = payload.get("issue", {}) or payload.get("pull_request", {})
+            tag_feature_per_github_event(event_type, action, metadata, repo_full_name)
```
matthewelwell: There's a possibility here that we send `repo_full_name` as `None` which will then throw an attribute error in the `tag_feature_per_github_event` function. If we know that `repository.full_name` will always exist, why not just do `repo_full_name = payload["repository"]["full_name"]` ?

matthewelwell (changes requested):

api/integrations/github/github.py
novakzaballa: Done.

api/integrations/github/github.py
novakzaballa: `.filter `was changed to `.get`.
It should not raise an AttributeError because this function is only triggered when the webhook is called, which only happens when there is an integration with the issues/PRs of a GitHub repo with Flagsmith. Therefore, using .get should be appropriate in this context.

api/integrations/github/github.py
novakzaballa: GitHub always sends the full_name, changed.

zachaysan (approved):
I'm going to be testing this on staging post merge.

api/integrations/github/github.py
```diff
@@ -100,8 +116,10 @@ def handle_github_webhook_event(event_type: str, payload: dict[str, Any]) -> Non
         handle_installation_deleted(payload)
     elif event_type in tag_by_event_type:
```
matthewelwell: This is pointless, if `event_type` isn't in `tag_by_event_type` then we'll get a `KeyError` [here](https://github.com/Flagsmith/flagsmith/pull/4586/files?diff=split&w=0#diff-9571c944c3603fa5041c2bbb3bfc36614b1a102e1d22e7552b6f943c7cb24f9dR84). I assume based on that, there is no way for `event_type` to not exist in `tag_by_event_type` but we should check on this. 

If that is the case, I think this can just be changed to (and we can remove one layer of nesting):

```python 
elif action in tag_by_event_type[event_type]
```

matthewelwell (changes requested):

api/integrations/github/github.py
zachaysan: I checked and it was already filtered in the view code, so the check here is unnecessary. I've updated the relevant code.

matthewelwell (approved):

PR #4589, opened 2024-09-04
fix: Handle null cancellation dates

zachaysan:
## Changes

Trials mean we don't always get cancellation dates in the response from Chargebee. This fixes this.

Edit: There is now testing in place. The one path where `current_term_end` is set was covered by an existing test, so I've added one new one for the case where it is unset.

### DIFF
--- a/api/tests/unit/organisations/test_unit_organisations_views.py
+++ b/api/tests/unit/organisations/test_unit_organisations_views.py
@@ -688,6 +688,9 @@ def test_when_subscription_is_set_to_non_renewing_then_cancellation_date_set_and
 
     # Then
     chargebee_subscription.refresh_from_db()
+    # Cancellation date is set to None because the missing current_term_end field
+    # means that the cancellation is processed immediately and the subscription
+    # reverts to being a free plan, so there's no more cancelation date set.
     assert chargebee_subscription.cancellation_date is None
     assert chargebee_subscription.plan == FREE_PLAN_ID
     assert len(mail.outbox) == 1
--- a/api/tests/unit/organisations/test_unit_organisations_views.py
+++ b/api/tests/unit/organisations/test_unit_organisations_views.py
@@ -37,6 +37,7 @@ from organisations.permissions.models import UserOrganisationPermission
 from organisations.permissions.permissions import CREATE_PROJECT
 from organisations.subscriptions.constants import (
     CHARGEBEE,
+    FREE_PLAN_ID,
     MAX_API_CALLS_IN_FREE_PLAN,
     MAX_SEATS_IN_FREE_PLAN,
     SUBSCRIPTION_BILLING_STATUS_ACTIVE,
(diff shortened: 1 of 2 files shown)


matthewelwell:
@zachaysan I'm not sure why testing is N/A here? Let's make sure that we add some test coverage please.

zachaysan:
> @zachaysan I'm not sure why testing is N/A here? Let's make sure that we add some test coverage please.

This is done now @matthewelwell

api/tests/unit/organisations/test_unit_organisations_views.py
```diff
@@ -657,6 +659,44 @@ def test_when_subscription_is_set_to_non_renewing_then_cancellation_date_set_and
...
+    url = reverse("api-v1:chargebee-webhook")
+
+    settings.ORG_SUBSCRIPTION_CANCELLED_ALERT_RECIPIENT_LIST = ["foo@bar.com"]
+
+    # When
+    staff_client.post(url, data=json.dumps(data), content_type="application/json")
+
+    # Then
+    subscription.refresh_from_db()
+    assert subscription.cancellation_date is None
```
matthewelwell: This feels like the opposite of what we want to test?

api/tests/unit/organisations/test_unit_organisations_views.py
```diff
@@ -657,6 +659,44 @@ def test_when_subscription_is_set_to_non_renewing_then_cancellation_date_set_and
     mocked_cancel_chargebee_subscription.assert_not_called()
 
 
+@pytest.mark.freeze_time("2023-01-19T09:09:47.325132+00:00")
+@mock.patch("organisations.models.cancel_chargebee_subscription")
+def test_when_subscription_is_set_to_non_renewing_then_cancellation_date_set_and_current_term_end_is_missing(
+    mocked_cancel_chargebee_subscription: MagicMock,
+    subscription: Subscription,
```
matthewelwell: Shouldn't we use e.g. `chargebee_subscription` here?

api/tests/unit/organisations/test_unit_organisations_views.py
zachaysan: It gets set to `None` because if a canceled subscription isn't slated in the future it reverts to being a free plan, so the cancelation date is set to `None`.

api/tests/unit/organisations/test_unit_organisations_views.py
zachaysan: Sure that's a good idea, I've added it.

api/tests/unit/organisations/test_unit_organisations_views.py
zachaysan: I've added a comment to the assertion to clarify why it is set to `None`.

matthewelwell (approved):

PR #4590, opened 2024-09-05
feat: allow ignore conflicts on scheduled change

matthewelwell:
## Changes

This PR sets out to resolve an issue where it's not possible to schedule consecutive change requests on the same feature. For example, in the case where you want to schedule a slow roll out by changing the segment override on a feature to a different percentage value. 

To achieve this, I have added a value to the ChangeRequest model which allows the creator to ignore conflicts. 

Currently the use case given in the test will only really be useful when using the API directly but this is the use case for a current customer. 

Added one long integration style test to test the exact use case for a customer.


api/features/workflows/core/models.py
```diff
@@ -91,6 +91,8 @@ class ChangeRequest(
         null=True,
     )
 
+    ignore_conflicts = models.BooleanField(default=False)
```
gagantrivedi, line 94: Do we need to add this to some serializer to make it editable from the UI?

gagantrivedi (commented):
One question

api/features/workflows/core/models.py
matthewelwell: Yes, in the workflows repo which I will do after merging this PR so I can test it there properly.

gagantrivedi (approved):

PR #4598, opened 2024-09-06
fix: Don't include null traits in transient identifier

### LINKED ISSUE #4597: TypeError: 'NoneType' object is not subscriptable
Sentry Issue: [FLAGSMITH-API-4YX](https://flagsmith.sentry.io/issues/5801518743/?referrer=github_integration)

```
TypeError: 'NoneType' object is not subscriptable
(8 additional frame(s) were not displayed)
...
  File "environments/identities/views.py", line 240, in post
    instance = serializer.save()
  File "environments/sdk/serializers.py", line 156, in save
    identity, traits = get_transient_identity_and_traits(
  File "environments/sdk/services.py", line 29, in get_transient_identity_and_traits
    identifier=get_transient_identifier(sdk_trait_data),
  File "environments/sdk/services.py", line 97, in get_transient_identifier
    "".join(
  File "environments/sdk/services.py", line 98, in <genexpr>
    f'{trait["trait_key"]}{trait["trait_value"]["value"]}'
```

khvn26:

## Changes

Fixes #4597.

This clarifies typing and discards null traits from the transient identifier generation.

Added a null trait to corresponding integration test.

### DIFF
--- a/api/environments/sdk/services.py
+++ b/api/environments/sdk/services.py
@@ -95,8 +95,9 @@ def get_transient_identifier(sdk_trait_data: list[SDKTraitData]) -> str:
     if sdk_trait_data:
         return hashlib.sha256(
             "".join(
-                f'{trait["trait_key"]}{trait["trait_value"]["value"]}'
+                f'{trait["trait_key"]}{trait_value["value"]}'
                 for trait in sorted(sdk_trait_data, key=itemgetter("trait_key"))
+                if (trait_value := trait["trait_value"])
             ).encode(),
             usedforsecurity=False,
         ).hexdigest()
--- a/api/environments/sdk/services.py
+++ b/api/environments/sdk/services.py
@@ -97,7 +97,7 @@ def get_transient_identifier(sdk_trait_data: list[SDKTraitData]) -> str:
             "".join(
                 f'{trait["trait_key"]}{trait_value["value"]}'
                 for trait in sorted(sdk_trait_data, key=itemgetter("trait_key"))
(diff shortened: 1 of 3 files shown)


api/environments/sdk/services.py
```diff
@@ -95,8 +95,9 @@ def get_transient_identifier(sdk_trait_data: list[SDKTraitData]) -> str:
     if sdk_trait_data:
         return hashlib.sha256(
             "".join(
-                f'{trait["trait_key"]}{trait["trait_value"]["value"]}'
+                f'{trait["trait_key"]}{trait_value["value"]}'
                 for trait in sorted(sdk_trait_data, key=itemgetter("trait_key"))
+                if (trait_value := trait["trait_value"])
```
matthewelwell: Shouldn't this be the following?

```python
if (trait_value := trait["trait_value"]) is not None
``` 

Otherwise, e.g. `0` would not be included (I think)?

matthewelwell (changes requested):

api/environments/sdk/services.py
khvn26: Here, `"trait_value"` is a structure enveloping the value, not the value itself.

matthewelwell: Ah, I see 👍

matthewelwell (approved):

api/environments/sdk/services.py
matthewelwell: I've approved, but I actually think it would still be better to be explicit here.

PR #4603, opened 2024-09-09
fix(versioning): fix issue creating duplicate priority segment overrides

matthewelwell:
## Changes

This tightens up the API to avoid situations where existing segment overrides are sent as updates even though nothing actually changes. Previously this caused an issue where 2 segments would end up with the same priority. 

Related issue #4604

Added a specific test for the issue found by a customer, ensured that the test failed as seen in the behaviour reported by the customer, and subsequently fixed the test.


matthewelwell:
API component of #4604

gagantrivedi (approved):

PR #4609, opened 2024-09-10
feat: Detect unchanged feature states when saving versions

kyle-ssg:

## Changes

Saving a versioned feature / segment overrides will now correctly detect changes before saving. This also means we can show an error to the user if their save contains no changes 

image
image

Attempted to save segment overrides / features without changes


matthewelwell (approved):
I've tested this manually using the preview, and given the code a quick look. All looks good to me.

PR #4611, opened 2024-09-10
fix: Set organisation api usage to zero

zachaysan:
## Changes

The original code involved in setting `api_calls_24h`, `api_calls_7d`, and `api_calls_30d` for the `OrganisationSubscriptionInformationCache` only updated values that were known to InfluxDB, this led to an issue where once an organisation left Flagsmith the code updating the values for `0` level API responses were instead left their previous value. This PR fixes that by checking for organisations that have no values and sets them to `0`.

Partially manually against production InfluxDB and then added a test for the empty caches case.

### DIFF
--- a/api/organisations/subscription_info_cache.py
+++ b/api/organisations/subscription_info_cache.py
@@ -85,12 +85,21 @@ def _update_caches_with_influx_data(
 
         org_calls = get_top_organisations(date_start, limit)
 
+        covered_orgs = set()
+
         for org_id, calls in org_calls.items():
             subscription_info_cache = organisation_info_cache_dict.get(org_id)
+            covered_orgs.add(org_id)
+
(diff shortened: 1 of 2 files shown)


gagantrivedi:
What if Postgres is being used to store usage data instead of influx?

zachaysan:
> What if Postgres is being used to store usage data instead of influx?

Presently the `update_caches` function does not handle that use-case. The caches that are generated are mostly used for areas of the application which don't really have any use outside of the Flagsmith staff. For example, the sales dashboard is one of the uses of the cache.

That said, if we want to support the Postgres way of going about things then I'd argue we should create a new GitHub issue instead of lumping it in with this PR.

gagantrivedi (approved):

PR #4612, opened 2024-09-10
feat: Add use_identity_overrides_in_local_eval setting

### LINKED ISSUE #4607: UI(environment): Add toggle for use_identity_override_in_local_eval

kyle-ssg:

## Changes

image

Toggled setting

### DIFF
--- a/frontend/web/components/pages/EnvironmentSettingsPage.js
+++ b/frontend/web/components/pages/EnvironmentSettingsPage.js
@@ -168,6 +168,7 @@ const EnvironmentSettingsPage = class extends Component {
         allow_client_traits: !!this.state.allow_client_traits,
         banner_colour: this.state.banner_colour,
         banner_text: this.state.banner_text,
+        use_identity_overrides_in_local_eval: this.state.use_identity_overrides_in_local_eval,
         description: description || env.description,
         hide_disabled_flags: this.state.hide_disabled_flags,
         hide_sensitive_data: !!this.state.hide_sensitive_data,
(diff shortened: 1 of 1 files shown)


rolodato (approved):
image

image

Other than that looks good, but I'd like to have some explanation on why users would want to have this enabled or not. There are also docs like https://docs.flagsmith.com/clients/#local-evaluation-1 that need to be updated to take into account this new setting, but we can add this in a different PR

kyle-ssg:
Good question as to why they'd want this enabled or not, I get the feeling that we'd start enabling this by default @gagantrivedi ? 

The only reason I see this being useful as an option is if a project uses flags on the backend and frontend and want to reserve identity overrides for just development and never want it hitting their backend.

gagantrivedi:
> Good question as to why they'd want this enabled or not, I get the feeling that we'd start enabling this by default @gagantrivedi ?
> 
> The only reason I see this being useful as an option is if a project uses flags on the backend and frontend and want to reserve identity overrides for just development and never want it hitting their backend.


The reasoning is captured here https://flagsmith.slack.com/archives/CTF0THS2D/p1725270568904189?thread_ts=1724865246.072169&cid=CTF0THS2D

gagantrivedi (approved):

kyle-ssg:
Ok yep, bringing here for visibility 

> This will mean that a release on our side might (probably will) change the value for flags in our customer’s environments. I actually think this should be an opt in flag on the environment, defaulted to true for all new environments, but off for any 
 existing ones.

So essentially this is to avoid any unexpected changes

kyle-ssg:
@rolodato based on this, should we just have a small note somewhere on this rather than having it be too prominent in docs ?

rolodato:
Let's merge this as-is and improve the docs later. Getting more people onboarded to this is more important than having perfect docs.

We'd need to share more detailed context in the docs by explaining that this is only for SaaS, was/will be enabled by default for all new environments since a specific Flagsmith version - too much to fit in a small note IMO.

PR #4615, opened 2024-09-10
feat: Manage tags permission

kyle-ssg:

## Changes

Based on the feature flag manage_tags_permission, this will require MANAGE_TAGS rather than ADMIN permissions for creating,editing and removing tags

Validate that the permission variables are set correctly, permission doesn't exist yet


matthewelwell:
The PR to add the API logic is [here](https://github.com/Flagsmith/flagsmith/pull/4628). We should make sure that the API PR is merged and released first as it doesn't change the current behaviour (since it will fall back to just using project admin is a user doesn't have the MANAGE_TAGS permission).

kyle-ssg:
@matthewelwell Actually that's fine since it's under the flag manage_tags_permission. Plus frontend would do the same too so I guess it's actually backwards compatible

matthewelwell:
@kyle-ssg the tooltip on the 'Add Tag' button still refers to the 'Admin' permission. 

image

kyle-ssg:
@matthewelwell done

matthewelwell (approved):

PR #4618, opened 2024-09-10
fix: ignore old versions when validating segment override limit

matthewelwell:
## Changes

Ensure that old versions aren't included when validating segment override limit. 

Added tests.


api/features/feature_segments/limits.py
```diff
@@ -0,0 +1,52 @@
...
+
+SEGMENT_OVERRIDE_LIMIT_EXCEEDED_MESSAGE = (
+    "The environment has reached the maximum allowed segments overrides limit."
+)
+
+
+def exceeds_segment_override_limit(
+    environment: Environment,
+    segment_ids_to_create_overrides: list[int] = None,
+    segment_ids_to_delete_overrides: list[int] = None,
```
zachaysan: I understand it's setting them equal to empty list below, but I still think the typing should match the default arguments.
```suggestion
    segment_ids_to_create_overrides: list[int] | None = None,
    segment_ids_to_delete_overrides: list[int] | None = None,
```

api/features/feature_segments/limits.py
```diff
@@ -0,0 +1,52 @@
...
+        )
+    )
+
+    if _check(
+        segment_override_count + extra,
+        environment.project.max_segment_overrides_allowed,
+    ):
+        return True
+
+    return False
```
zachaysan:
```suggestion
    return _check(
        segment_override_count + extra,
        environment.project.max_segment_overrides_allowed,
    )
```

api/tests/unit/features/versioning/test_unit_versioning_views.py
```diff
@@ -1198,3 +1202,312 @@ def generate_segment_override_fs_payload(
...
+        ],
+    }
+    create_version_url = reverse(
+        "api-v1:versioning:environment-feature-versions-list",
+        args=[environment_v2_versioning.id, feature.id],
+    )
+
+    # When
+    # We create a new version, without making any changes, we shouldn't receive
+    # any errors.
```
zachaysan: I'm a little confused by this comment because a `status.HTTP_400_BAD_REQUEST` is returned below. Is a `400` response not considered an error?

zachaysan (approved):
Overall looks great, just a few small questions / comments.

api/tests/unit/features/versioning/test_unit_versioning_views.py
matthewelwell: Sorry, this was a copy pasta error.

api/features/feature_segments/limits.py
matthewelwell: Good catch, updated.

api/features/feature_segments/limits.py
matthewelwell: Yep, that's a fair point, updated.

zachaysan (approved):

PR #4620, opened 2024-09-11
feat: Identity alias

### LINKED ISSUE #4016: Epic: Improve search and display capabilities for identities in the Flagsmith UI
Currently, due to the large quantities of data involved in identity storage, and the way in which that data is stored in our SaaS platform to support the Edge API, searching and displaying additional data about identities can be very difficult. 

Some of the main problems are: 

 1. It is not possible to search on anything other than the identifier. This is problematic, particularly when introducing non-engineering users to Flagsmith since the identifier is often a unique key such as a `uuid` or similar which most users will not have access to. 
 2. Similar to the above, it is not possible to see at a quick glance from the list of identities which identities are which because we only show the identifier. 
 3. We do not show the total number of identities (only applicable to SaaS). 

Note that this issue combines both #444 and #290.

kyle-ssg:

## Changes

- Allows you to specify an alias for identities
- The flagsmith app will now start identifying by id rather than email

image

GIF Recording 2024-09-17 at 4 36 58 PM
_


kyle-ssg:
Related https://github.com/Flagsmith/flagsmith/pull/4569

matthewelwell:
@kyle-ssg as discussed, let's force the input to lower case. 

image

kyle-ssg:
@matthewelwell done

matthewelwell:
@kyle-ssg this still feels a little bit unpolished to me. The main 2 issues I have are: 

 1. I'm not sure why we can't apply the same logic that we have in for e.g. feature names which doesn't allow you to input upper case characters at all, and instead automatically lower cases them as you type. 
 2. It's quite annoying that it clears the whole alias when you click 'Edit' - what happens if I just realised that I made a typo and I want to quickly fix it? I now have to re-enter the whole thing. Can we get the edit to just 'edit in place' as it were?

matthewelwell (approved):
I'm approving this. The only remaining thing I noticed was that we still show the alias on the list page even when the flag is disabled. 

I don't see this as a particular issue, since it's obviously not possible to set an alias without the flag being enabled, but it might be odd if we subsequently want to turn the flag off again in a future situation.

PR #4625, opened 2024-09-12
fix: Add logging to segments code

zachaysan:
## Changes

Adds logging to help diagnose a strange bug that came up in production around conditions being created in large numbers.

The existing test suite should cover the code running ok.

### DIFF
--- a/api/segments/models.py
+++ b/api/segments/models.py
@@ -246,6 +246,9 @@ class SegmentRule(SoftDeleteExportableModel):
         cloned_rule.uuid = uuid.uuid4()
         cloned_rule.id = None
         cloned_rule.save()
+        logger.info(
+            f"Deep copying rule {self.id} for cloned rule {cloned_rule.id} for cloned segment {cloned_segment.id}"
+        )
 
         # Conditions are only part of the sub-rules.
         assert self.conditions.exists() is False
@@ -259,6 +262,10 @@ class SegmentRule(SoftDeleteExportableModel):
             cloned_sub_rule.uuid = uuid.uuid4()
             cloned_sub_rule.id = None
             cloned_sub_rule.save()
+            logger.info(
+                f"Deep copying sub rule {sub_rule.id} for cloned sub rule {cloned_sub_rule.id} "
+                f"for cloned segment {cloned_segment.id}"
+            )
 
             cloned_conditions = []
             for condition in sub_rule.conditions.all():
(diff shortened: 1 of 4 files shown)


api/segments/models.py
```diff
@@ -246,6 +247,10 @@ def deep_clone(self, cloned_segment: Segment) -> "SegmentRule":
         cloned_rule.uuid = uuid.uuid4()
         cloned_rule.id = None
         cloned_rule.save()
+        if is_saas():
```
matthewelwell: I don't think we need to worry about only logging in SaaS - most people won't have the log level set to info anyway ... in fact, I'm not sure that we do?

zachaysan: 1. I sorta agree with you but this was the feedback from @rolodato so I deferred to his judgement.
2. We don't log info in production? That's really surprising. How should I diagnose issues like this where we want to track down problems?

zachaysan: Ok I've reverted the commit that adds the `is_saas` check, though I'm not really sure if this PR is useful if we're not logging info in production so I still need guidance on that one.

matthewelwell: (2) based on [this](https://github.com/Flagsmith/flagsmith/blob/main/api/app/settings/common.py#L572) and [this](https://github.com/Flagsmith/flagsmith/blob/main/infrastructure/aws/production/ecs-task-definition-web.json), I don't think we do. 

You'll need to propose a solution :)

zachaysan: Two solutions come to mind:
1. Enable logging info in production. I don't really see the downside of logging info. It's the log level that's been set at every other company I've been at and I can't be the only engineer that's surprised to see that it's not logged. Disabling debug level logging makes sense though.
2. Switch the logging level to warning which sorta rubs me the wrong way since these aren't really warnings, but I can add comments next to each logging warning call that says basically "these aren't really warnings but we want to log them in production and production is currently set at the warning level" but something more pithier.

matthewelwell: Yep, I'm happy with (1) - can you add that to this PR? It should just be a case of setting that environment variable in the task definition JSON file (+ the one for staging too for consistency).

zachaysan: Ok I think I've set the environment variables for staging and production properly, but I'm not really sure how to test the setting. Maybe we can verify once this is merged into `main` on staging?

matthewelwell: Yes, exactly.

infrastructure/aws/production/ecs-task-definition-web.json
```diff
@@ -19,6 +19,10 @@
             ],
             "essential": true,
             "environment": [
+                {
+                    "name": "LOG_LEVEL",
+                    "value": "info"
```
matthewelwell: It's probably fine, but for consistency, I think this should be `INFO` ?

zachaysan: Sounds good. Updated.

matthewelwell (approved):

matthewelwell:
Tested in staging

image

Issue #4600, opened 2024-09-09
Add MANAGE_TAGS permission


matthewelwell:
Currently the UI requires the user to have project admin privileges to manage the tags for a feature, whereas the API only requires `CREATE_FEATURE` permission.

kyle-ssg:
As discussed I think this was more that ``CREATE_FEATURE`` users cant create new tags, we should create a new permission ``MANAGE_TAGS``

PR #4628, opened 2024-09-13
feat: add MANAGE_TAGS permission

matthewelwell:
## Changes

Adds the MANAGE_TAGS permission as per #4600 

Added new unit test.


api/permissions/migrations/0010_add_manage_tags_permission.py
```diff
@@ -0,0 +1,32 @@
...
+def add_manage_tags_permission(apps: Apps, schema_editor: BaseDatabaseSchemaEditor) -> None:
+    permission_model_class = apps.get_model("permissions", "permissionmodel")
+    permission_model_class.objects.get_or_create(
+        key=MANAGE_TAGS,
+        description="Allows the user to manage tags in the given project.",
+        type=PROJECT_PERMISSION_TYPE,
+    )
+
+
+def reverse(apps: Apps, schema_editor: BaseDatabaseSchemaEditor) -> None:  # pragma: no cover
```
matthewelwell, line 19: Justification for adding `pragma: no cover` here:

Considering the complexity of the code, and the very limited scope in which this code will ever be executed, I figured that it wasn't worth writing a migration test for it (since they are quite arduous to write). 

I have tested the migration forwards and backwards manually.

khvn26 (approved):

PR #4643, opened 2024-09-18
feat: Add tag based permissions

### LINKED ISSUE #1535: WIP: More granular feature level permissions

### LINKED ISSUE #4424: Restrict Permissions Based on Tags in Roles
**FE Restrictions:**

- When a tag is added to a role, the FE should automatically disable any other permissions that are not compatible with the tag.

**API Validation:**

- The API should reject the creation of any role that has a tag associated with unsupported permissions.

- The API response for the /{entity}/permissions/ endpoints (where entity can be projects or environments) should include information about these validation errors.

kyle-ssg:

## Changes

Supersedes https://github.com/Flagsmith/flagsmith/pull/4250

Users can now create tag-based roles
image

A tag based role will let you specify relevant permissions

image

image

The frontend will check for any tag-based permissions if it cannot find any standard permissions. In this example my user can only update feature states for the marketing tag.

https://github.com/user-attachments/assets/9b8fe88d-8848-46bf-915b-b98ba7eea851

Needs https://github.com/Flagsmith/flagsmith/issues/4799 to be implemented in order for it to work fully.


rolodato:
I don't understand how this is supposed to be used based on the docs description.

From what I understand, this functionality allows defining feature-level permissions so that only specific users can modify certain sensitive features. Happy to collaborate on these docs but I'd like to sync and understand how this functionality actually works first.

kyle-ssg:
> I don't understand how this is supposed to be used based on the docs description.
> 
> From what I understand, this functionality allows defining feature-level permissions so that only specific users can modify certain sensitive features. Happy to collaborate on these docs but I'd like to sync and understand how this functionality actually works first.

It's quite hard to describe but it does work as updated in the docs. 

If permissions are tagged, the following permissions will only be granted on features with matching tags:
- Create feature, Update Feature, Segment Overrides, and Change Requests

If a user has admin permissions from another role or group it'll grant the user the missing permissions.

kyle-ssg:
@gagantrivedi cc just to confirm, were the missing permissions added? This used to be just update feature and segment override

gagantrivedi:
> @gagantrivedi cc just to confirm, were the missing permissions added? This used to be just update feature and segment override

No, they are not added yet

PR #4654, opened 2024-09-24
feat(export): Add support for edge identities data

gagantrivedi:

## Changes

Add support  for exporting edge-identities data(i.e: identities, traits, and feature states) 

Adds unit test and test manually by exporting(from staging) and loading an organisation locally


api/edge_api/identities/export.py
```diff
@@ -0,0 +1,181 @@
+import typing
+import uuid
+from decimal import Decimal
+from typing import Dict, List, Tuple
```
zachaysan: Is there a reason for using the typing module instead of just using `dict`, `list`, and `tuple`?

api/edge_api/identities/export.py
```diff
@@ -0,0 +1,181 @@
...
+        feature_id_to_uuid[feature_id] = _get_feature_uuid(feature_id)
+    feature_uuid = feature_id_to_uuid[feature_id]
+
+    return {
+        "model": "features.featurestate",
+        "fields": {
+            "uuid": featurestate_uuid,
+            "created_at": timezone.now().isoformat(),
+            "updated_at": timezone.now().isoformat(),
+            "live_from": timezone.now().isoformat(),
```
zachaysan, line 163: It's a little surprising that these timestamps are all set to `now()` maybe add a comment explaining why?

api/edge_api/identities/export.py
```diff
@@ -0,0 +1,181 @@
...
+        )
+    mv_feature_option_uuid = mv_feature_option_id_to_uuid[mv_feature_option_id]
+
+    return {
+        "model": "multivariate.multivariatefeaturestatevalue",
+        "fields": {
+            "uuid": uuid.uuid4(),
+            "feature_state": [featurestate_uuid],
+            "multivariate_feature_option": [mv_feature_option_uuid],
+            "percentage_allocation": 100.0,
```
zachaysan: Why is this hard coded to `100.0`?

api/tests/unit/import_export/test_unit_import_export_export.py
```diff
@@ -278,6 +283,181 @@ def test_export_features_with_environment_feature_version(
     # TODO: test whether the export is importable
 
 
+def test_export_edge_identities(
+    flagsmith_identities_table: Table,
+    project: Project,
+    environment: Environment,
+    multivariate_feature: Feature,
+    multivariate_options: typing.List[MultivariateFeatureOption],
+):
```
zachaysan: Missing return type

zachaysan (approved):
Overall looks good. I just have a couple of small suggestions / questions.

api/edge_api/identities/export.py
gagantrivedi: It's because an identity override can only be 100 percent(for mv)

api/edge_api/identities/export.py
gagantrivedi: Mostly to support things like these

`    feature_id_to_uuid: Dict[int, str] = {}
`

api/edge_api/identities/export.py
gagantrivedi: Yeah, not a bad idea

api/edge_api/identities/export.py
```diff
@@ -0,0 +1,205 @@
...
+            for trait in item["identity_traits"]:
+                traits_export.append(
+                    export_edge_trait(trait, identifier, environment_api_key)
+                )
+            for override in item["identity_features"]:
+                featurestate_uuid = override["featurestate_uuid"]
+                feature_id = override["feature"]["id"]
+                if feature_id not in feature_id_to_uuid:
+                    try:
+                        feature_id_to_uuid[feature_id] = _get_feature_uuid(feature_id)
```
matthewelwell: Seems like it would perhaps be worthwhile just grabbing all the features up front and creating a dictionary or similar to grab the uuid here.

api/edge_api/identities/export.py
```diff
@@ -0,0 +1,205 @@
...
+                    )
+                )
+                featurestate_value = override["feature_state_value"]
+                if featurestate_value is not None:
+                    # export feature state value
+                    identity_override_export.append(
+                        export_featurestate_value(featurestate_value, featurestate_uuid)
+                    )
+                if mvfsv := override["multivariate_feature_state_values"]:
+                    # only one mvfs override can exists
```
matthewelwell:
```suggestion
                    # only one mvfs override can exist
```

matthewelwell: I actually don't think this is necessarily true. e.g. I suggested to a direct API user to not do this [here](https://app.crisp.chat/website/8857f89e-0eb5-4263-ab49-a293872b6c19/inbox/session_0ca20b09-38a7-49d3-afc4-6991895f2e43/).

api/edge_api/identities/export.py
```diff
@@ -0,0 +1,205 @@
...
+                    mv_feature_option_id = mvfsv[0]["multivariate_feature_option"]["id"]
+                    if mv_feature_option_id not in mv_feature_option_id_to_uuid:
+                        try:
+                            mv_feature_option_id_to_uuid[mv_feature_option_id] = (
+                                _get_mv_feature_option_uuid(mv_feature_option_id)
```
matthewelwell: As per the above, maybe we should grab the options up front and use it here, instead of always calling `.get` in the function.

api/edge_api/identities/export.py
zachaysan: This runs just fine unless I'm missing something
```python
def example_function(feature_id_to_uuid: dict[int, str] = {}) -> None:
    print(feature_id_to_uuid)
```

api/edge_api/identities/export.py
matthewelwell: As per my comment above, I'm not sure this is strictly true.

api/import_export/export.py
```diff
@@ -166,6 +188,20 @@ def export_identities(organisation_id: int) -> typing.List[dict]:
...
+    identities = []
+    traits = []
+    identity_overrides = []
+    for environment in Environment.objects.filter(
+        project__organisation__id=organisation_id, project__enable_dynamo_db=True
+    ):
+        result = export_edge_identity_and_overrides(environment.api_key)
+        identities.extend(result[0])
+        traits.extend(result[1])
+        identity_overrides.extend(result[2])
```
matthewelwell: This is probably just preference, but I much prefer something like the following: 

```suggestion
        exported_identities, exported_traits, exported_overrides = export_edge_identity_and_overrides(environment.api_key)
        identities.extend(exported_identities)
        ...
```

api/import_export/export.py
```diff
@@ -115,13 +117,25 @@ def export_projects(organisation_id: int) -> typing.List[dict]:
...
-            Q(rule__segment__project__organisation__id=organisation_id)
-            | Q(rule__rule__segment__project__organisation__id=organisation_id),
+            Q(
+                rule__segment__project__organisation__id=organisation_id,
+                rule__segment_id=F("rule__segment__version_of"),
+            )
+            | Q(
+                rule__rule__segment__project__organisation__id=organisation_id,
+                rule__rule__segment_id=F("rule__rule__segment__version_of"),
+            ),
```
matthewelwell, line 138: I'm surprised that we don't see a specific test for this?

api/edge_api/identities/export.py
gagantrivedi: We are calling get only if the `id` is not in our cache (mv_feature_option_id_to_uuid), so the only difference is that we are populating the cache lazily

api/edge_api/identities/export.py
gagantrivedi: > setting the percentage_allocation variation that you want to use for the given identity to 100%, and the others to 0%. 

I don't understand why, though? If you override from the UI, it only sends one option with 100 percent?

api/edge_api/identities/export.py
matthewelwell: But if there are 600 features, each with at least 1 override, we are making 600 queries... vs just one to grab all the features up front?

api/edge_api/identities/export.py
matthewelwell: Yeah, I'm not sure, but the customer in crisp suggested that they got unexpected behaviour without it. I don't think we can rule it out as a use case, as strictly that array should always contain a value for each MV option.

api/edge_api/identities/export.py
gagantrivedi: Same as https://github.com/Flagsmith/flagsmith/pull/4654/files#r1778686641

api/edge_api/identities/export.py
gagantrivedi: Oh! I wasn't aware of that! thanks

api/import_export/export.py
gagantrivedi: Yep, added

api/edge_api/identities/export.py
gagantrivedi: Done

api/edge_api/identities/export.py
gagantrivedi: Yep, Done

api/import_export/export.py
gagantrivedi: Done

api/edge_api/identities/export.py
gagantrivedi: Done, it is not hardcoded anymore

api/tests/unit/import_export/test_unit_import_export_export.py
```diff
@@ -96,6 +102,36 @@ def test_export_project(organisation):
     # TODO: test whether the export is importable
 
 
+def test_export_project__only_live_segments_are_imported(
```
matthewelwell:
```suggestion
def test_export_project__only_live_segments_are_exported(
```

matthewelwell (approved):
Approved with a minor suggested naming improvement.

PR #4656, opened 2024-09-24
fix: diff check for versioned segment overrides and MV

kyle-ssg:

## Changes

- Fixes an issue whereby the frontend doesn't detect changes when adjusting only a segment override value or % weight for versioned change requests.
- Fixes an issue where creating a mv feature then attempting to edit without refreshing didnt work

This was due to the frontend having access to the raw feature state value rather than the object version ``{string_value:"eg",...}``

Likely related to this PR: #4609. 

- Published versions without change requests
- Updated feature values
- Updated segment override values
- Published versions with change requests
- Created change requests changing feature values and segment override values


matthewelwell:
It looks like there's still an issue when updating the value of a variation, see [here](https://www.loom.com/share/f315ee27bef54655ae464e4913f503d8?sid=4db1fc07-4646-4d88-a5ea-0ffb298ab129).

frontend/web/components/FeatureVersion.tsx
```diff
@@ -29,7 +29,7 @@ const FeatureVersion: FC<VersionDiffType> = ({
     environmentId,
     featureId: `${featureId}`,
     uuid: oldUUID,
-  })
+  })``
```
matthewelwell: This looks like a mistake?

frontend/web/components/diff/diff-utils.ts
```diff
@@ -117,6 +122,9 @@ export const getSegmentDiff = (
         : '',
     )
 
+    const multivariateDiffs = getVariationDiff(oldFeatureState, newFeatureState)
+    debugger
```
matthewelwell: Should this be removed?

matthewelwell (approved):

PR #4662, opened 2024-09-26
feat: Support cookie authentication

khvn26:

## Changes

This adds a basic JWT cookie support. As long as `COOKIE_AUTH_ENABLED` setting is set to `True`: 
- The HTTP-only cookie is returned from registration and login endpoints. It then can be used to authenticate requests to endpoints that expect a token.
- The classic token is not being returned from registration and login endpoints. Login endpoint produces a 204 response.
- The CORS setting defaults are modified to allow credentials and the FE domain origin. All of the CORS settings are still settable individually for special cases / local development.

The JWT lifetime is set to be 10 hours by default. No refresh endpoints have been implemented at this point; when we do implement refreshing, token lifetime should be shortened significantly.

It's possible to invalidate the cookie via a `POST /api/v1/auth/logout/` request. The token is added to blacklist and the cookie is unset.

A signal handler is added to set `Access-Control-Allow-Origin` based on `get_current_site_url` utility function. The function is modified to properly use the Django sites framework cache.

Added an integration test.

BE (including SAML) + FE has been tested in staging.


api/custom_auth/jwt_cookie/serializers.py
matthewelwell, line 1: Why do we need this empty file?

api/custom_auth/jwt_cookie/views.py
```diff
@@ -0,0 +1,11 @@
...
+from rest_framework.request import Request
+from rest_framework.response import Response
+from rest_framework_simplejwt.tokens import SlidingToken
+
+
+class TokenDestroyJWTSlidingBlacklistView(TokenDestroyView):
+    def post(self, request: Request) -> Response:
+        if isinstance(request.auth, SlidingToken):
+            request.auth.blacklist()
+        return super().post(request)
```
matthewelwell: I'm interested to know when this endpoint should be called? It feels like this should just be done on the generic logout endpoint?

api/custom_auth/urls.py
```diff
@@ -26,7 +26,11 @@
         CustomAuthTokenLoginWithMFACode.as_view(),
         name="mfa-authtoken-login-code",
     ),
-    path("logout/", TokenDestroyView.as_view(), name="authtoken-logout"),
+    path(
+        "logout/",
+        TokenDestroyJWTSlidingBlacklistView.as_view(),
+        name="authtoken-logout",
+    ),
```
matthewelwell: Hmm... shouldn't we only add this if the setting is enabled?

api/custom_auth/views.py
```diff
@@ -74,6 +79,8 @@ def post(self, request: Request) -> Response:
                 ephemeral_token=serializer.validated_data["ephemeral_token"],
             )
             serializer.user = user
+            if settings.AUTH_JWT_COOKIE_ENABLED:
+                return authorise_response(user, Response())
```
matthewelwell: Right, I was typing out some questions, because I didn't fully understand the behaviour here (and in the similar conditional statement above), but I think I understand it more now, although I still have some questions. 

As I understand it, we're probably adding some headers to the response so that the client has the JWT? What I'm still not sure about then is why we're returning an empty `Response` object, instead of doing something like: 

```python
response = self._action(serializer)
if settings.AUTH_JWT_COOKIE_ENABLED:
    return authorise_response(user, response)
return response
```

khvn26:
```
response = self._action(serializer)
```

creates the classic token and returns it. We don't need that for the cookie auth.

api/custom_auth/jwt_cookie/views.py
khvn26: >  the generic logout endpoint?

Which is...?

khvn26: I guess the real question here is why am I adding support for classic token deletion on logout. The answer is "no idea, it shouldn't be there."

I'm modifying it to only invalidate the JWT if it's present.

api/custom_auth/jwt_cookie/serializers.py
khvn26: Good catch, removed.

api/custom_auth/urls.py
khvn26: Good point, done.

api/custom_auth/jwt_cookie/views.py
khvn26: This is done.

api/custom_auth/urls.py
khvn26: Reversed this to simplify testing. The endpoint is a no-op for classic token auth anyway now.

api/custom_auth/views.py
matthewelwell: I guess this means that the API changes depending on an environment variable, which is slightly odd? But maybe it's ok.

matthewelwell: I guess we have similar cases elsewhere so yeah, this is fine.

api/custom_auth/urls.py
```diff
@@ -26,7 +26,11 @@
         CustomAuthTokenLoginWithMFACode.as_view(),
         name="mfa-authtoken-login-code",
     ),
-    path("logout/", TokenDestroyView.as_view(), name="authtoken-logout"),
+    path(
+        "logout/",
+        JWTSlidingBlacklistView.as_view(),
+        name="authtoken-logout",
+    ),
```
matthewelwell: So, I guess what you're saying in other conversations is that `TokenDestroyView` does nothing, is that right? I guess I'm concerned about a couple of things here: 

 1. Are we removing functionality here for regular tokens? I assume not, but just checking. 
 2. Shouldn't we just have a `Logout` view class where we can internalise the conditional logic depending on the setting?

api/custom_auth/jwt_cookie/views.py
matthewelwell: Sorry, I think I misunderstood this code before. I didn't really notice that we were subclassing the `TokenDestroyView`. I guess this now relates to the question that I've asked [here](https://github.com/Flagsmith/flagsmith/pull/4662#discussion_r1777540740) in my second review.

khvn26: Yeah I got misled by your comment as well. Will revert tomorrow.

api/custom_auth/urls.py
khvn26: `TokenDestroyView`, in fact, does remove the token. However, the endpoint has never been used by the frontend; to log the user out, the application simply removed the `t` cookie.

Now that I expect the endpoint to be actually used, I ultimately decided to remove function related to classic tokens and only implement the JWT logout.

I have renamed the view and added a docstring to hopefully be more clear.

api/custom_auth/jwt_cookie/views.py
khvn26: Responded [here](https://github.com/Flagsmith/flagsmith/pull/4662#discussion_r1777697979).

rolodato:
Nit: I don't like the name `AUTH_JWT_COOKIE_ENABLED` - the fact that this cookie is a JWT is an implementation detail not relevant to any customers. We could have implemented an equivalent solution that uses opaque cookie tokens and not JWTs. In the context of OAuth/OIDC, the format of access tokens is only relevant to resource servers, i.e. Flagsmith API. Access tokens are meant to be opaque to clients (i.e. API-consuming applications, including the Flagsmith frontend).

It's likely that many customers will need to opt-in to this option if we decide to migrate everyone, so I'd like to have a useful name here. Nothing as permanent as a temporary solution etc.

How about `COOKIE_AUTH`, `COOKIE_AUTH_ENABLED` or similar instead?

khvn26:
@rolodato Good point, changing the name to `COOKIE_AUTH_ENABLED`.

@kyle-ssg FYI

api/custom_auth/urls.py
matthewelwell: Maybe the FE never used this functionality, but that's not to say that it's never been used by someone integrating with the API. It's obviously very unlikely, but not totally out of the question since we only provided API access via these user tokens for a long time. 

I don't see the harm in either: 

 1. Reverting to what you had initially and subclassing `TokenDestroyView` and calling `super()`
 2. Having 2 separate views that we add into urlpatterns depending on the existence of the environment variable.

khvn26: > but that's not to say that it's never been used by someone integrating with the API

They will still have the `DELETE /api/v1/auth/token/` endpoint that serves this exact function, and arguably has better semantics as we do not delete classic tokens on logout.

I can add the notion to OpenAPI docs.

api/Makefile
```diff
@@ -11,7 +11,7 @@ POETRY_VERSION ?= 1.8.3
 
 GUNICORN_LOGGER_CLASS ?= util.logging.GunicornJsonCapableLogger
 
-SAML_REVISION ?= v1.6.3
+SAML_REVISION ?= feat/jwt-cookie-auth
```
khvn26: Remove this once [this PR](https://github.com/Flagsmith/flagsmith-saml/pull/47) is merged.

kyle-ssg:
In order to test this I had to add the following to common.py when testing against localhost:8000 from localhost:8080

```
# CORS_ORIGIN_ALLOW_ALL = True - Remove
CORS_ALLOW_CREDENTIALS = True
CORS_ALLOWED_ORIGINS = [
    "http://localhost:8080",
]
``` 

In addition, if we were to have this run in production, I believe the cookie domain would have to be ".flagsmith.com".

Do we need to make all of these things configurable? I don't think we can have ``CORS_ORIGIN_ALLOW_ALL = True`` if we run with this?

api/custom_auth/urls.py
```diff
@@ -26,7 +26,11 @@
         CustomAuthTokenLoginWithMFACode.as_view(),
         name="mfa-authtoken-login-code",
     ),
-    path("logout/", TokenDestroyView.as_view(), name="authtoken-logout"),
+    path(
```
kyle-ssg, line 29: I think we need to delete the JWT_SLIDING_COOKIE_KEY cookie here, it looks to just invalidate rather than remove it too.

kyle-ssg: I also noticed as a result, /auth/login/ wouldn't work until I force deleted the cookie. This makes me wonder if the backend needs to ignore the cookie on certain endpoints. 

I think this will cause an issue for invite workflows where an already "logged in" user registers / logs in with another account in order to accept an invite they created.

khvn26: This is done.

khvn26:
@kyle-ssg I have modified the CORS defaults so users won't have to set CORS settings manually for this.

api/custom_auth/urls.py
matthewelwell: >They will still have the DELETE /api/v1/auth/token/ endpoint that serves this exact function, and arguably has better semantics as we do not delete classic tokens on logout.

But it's technically a breaking API change, right? We're removing functionality from an existing endpoint, that existed, and was documented for people to use.

Dockerfile
```diff
@@ -106,7 +106,7 @@ FROM build-python AS build-python-private
 
 # Authenticate git with token, install private Python dependencies,
 # and integrate private modules
-ARG SAML_REVISION
+ARG SAML_REVISION=feat/jwt-cookie-auth
```
matthewelwell: Remove this too before merge.

api/custom_auth/jwt_cookie/signals.py
```diff
@@ -0,0 +1,20 @@
...
+
+from core.helpers import get_current_site_url
+from corsheaders.signals import check_request_enabled
+from django.dispatch import receiver
+from django.http import HttpRequest
+
+
+@receiver(check_request_enabled)
+def cors_allow_current_site(request: HttpRequest, **kwargs: Any) -> bool:
+    # The signal is expected to only be dispatched:
```
matthewelwell, line 12: Is there some documentation that we can link to for this?

api/custom_auth/urls.py
khvn26: It seems you are correct here:

image

Still, I'd prefer to deprecate this functionality and direct people to `DELETE /api/v1/auth/token/` (as it will result in "logout" too).

matthewelwell: We can't just remove it though. I don't mind adding a warning that it'll be deprecated in the next version (and let's add it to the V2 API issue).

frontend/common/stores/account-store.js
```diff
@@ -186,7 +186,7 @@ const controller = {
           return
         }
 
-        data.setToken(res.key)
+        data.setToken(Project.cookieAuthEnabled?'true' : res.key)
```
matthewelwell: nit but there's an inconsistency in the formatting of some of these ternary statements.

```javascript
Project.cookieAuthEnabled?'true' : res.key
```

here vs the following in subsequent lines


```javascript
Project.cookieAuthEnabled ? 'true' : data.token
```

infrastructure/aws/staging/ecs-task-definition-web.json
```diff
@@ -19,6 +19,10 @@
             ],
             "essential": true,
             "environment": [
+                {
+                    "name": "COOKIE_AUTH_ENABLED",
+                    "value": "1"
+                },
```
matthewelwell: Do we also need to remove this before merge @khvn26 ?

api/custom_auth/jwt_cookie/signals.py
khvn26: Not really, you have to read django-cors-headers source for this knowledge.

matthewelwell: Understood.

api/custom_auth/urls.py
khvn26: The problem is that it's being called in FE now. We need to either add conditional logic to FE or explain to users that their tokens get invalidated if they logout explicitly. The latter is fine by me as using an admin API key is the clear alternative.

infrastructure/aws/staging/ecs-task-definition-web.json
khvn26: Yes.

frontend/common/stores/account-store.js
khvn26: Fixed.

api/custom_auth/urls.py
matthewelwell: > their tokens get invalidated if they login explicitly

I assume here you meant to say 'logout explicitly' ? 

It doesn't seem like it'd be hard to add conditional logic to the frontend to only call the logout endpoint if the jwt cookie exists, and seems to be the right thing to do to maintain compatibility, but I can't speak for @kyle-ssg here. 

Alternatively, we do what you suggested, and just show a warning when the user hits the logout button to say that any direct API integrations they have using the user token will stop functioning.

Another alternative (that doesn't involve any more FE work) is still to conditionally change the view class used by the `logout` route depending on the existence of environment variable, but we may then need to think about whether we can actually run both in parallel.

khvn26: As discussed offline, the FE conditional logic was added and token deletion function restored in the BE.

frontend/common/stores/account-store.js
```diff
@@ -328,12 +329,20 @@ const controller = {
...
-      API.setCookie('t', '')
-      data.setToken(null)
-      API.reset().finally(() => {
-        store.model = user
-        store.organisation = null
-        store.trigger('logout')
+      if (!data.token) {
+        return
+      }
+      ;(Project.cookieAuthEnabled
```
matthewelwell, line 335: ... is the semi-colon here some horrible JS syntax that I'm not aware of, or a typo? @kyle-ssg

matthewelwell (approved):
Approving with one minor comment

PR #4667, opened 2024-09-27
fix: update permissions on classes with missing / unclear permissions

### LINKED ISSUE #4532: Add missing `permission_classes` / `get_permissions` to views
The views listed below need to be updated with `permission_classes` / `get_permissions`.

- `api/integrations/slack/views.py:SlackGetChannelsViewSet` — potentially leaks Slack channels accessible to previously created integration

matthewelwell:
## Changes

1. Adds the required permission class to the SlackGetChannelsViewSet
2. Moves the `staff_member_required` from urls.py to decorate the methods on the class for consistency with other permission definitions

Added new tests for permissions failure use cases. Existing tests cover happy paths.

### DIFF
--- a/api/integrations/slack/views.py
+++ b/api/integrations/slack/views.py
@@ -6,13 +6,12 @@ from django.core.signing import TimestampSigner
 from django.shortcuts import redirect
 from rest_framework import status
 from rest_framework.decorators import action
-from rest_framework.permissions import BasePermission, IsAuthenticated
+from rest_framework.permissions import BasePermission
 from rest_framework.response import Response
 from rest_framework.viewsets import GenericViewSet
 from slack_sdk.oauth import AuthorizeUrlGenerator
 
 from environments.models import Environment
-from environments.permissions.permissions import NestedEnvironmentPermissions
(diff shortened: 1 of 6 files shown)


khvn26 (approved):

api/integrations/slack/views.py
```diff
@@ -36,6 +37,7 @@
 class SlackGetChannelsViewSet(GenericViewSet):
     serializer_class = SlackChannelListSerializer
     pagination_class = None  # set here to ensure documentation is correct
+    permission_classes = [IsAuthenticated, NestedEnvironmentPermissions]
```
khvn26: According to default permissions set in `NestedEnvironmentPermissions`, anybody who can view the environment will be able to list Slack channels using credentials stored with Slack integration settings. Is this the desired behaviour or the permissions could be narrowed further to only allow this for project admins?

matthewelwell: Yeah, I think this is a fair suggestion. I did consider this, but the functionality didn't seem to exist 'off the shelf' in the codebase (which I was surprised by). I'll add it here though.

matthewelwell: Done

PR #4672, opened 2024-09-30
fix: Handle cancellation date for api usage

zachaysan:
## Changes

Two small updates to the API usage alerting code that checks for subscription being cancelled and treating them as free accounts.

Two new tests.


khvn26 (approved):
Looks good.

PR #4684, opened 2024-10-02
feat: Set base url for segment

### LINKED ISSUE #4675: Add support for EU region for Segment
Currently the Segment integration only supports the default (US) region. We should allow users to select their region when configuring their integrations with Segment. 

Note that as part of this, we will also need to update the Edge API since the code [here](https://github.com/Flagsmith/edge-api/blob/main/src/integrations/identity_integrations/external/segment.py#L22-L24) is not sending the `host` argument when instantiating the client.

zachaysan:
## Changes

Introduce a change to the `host` parameter of the Segment client in order to support non-default regions. Also includes a migration to set existing segment clients to the default `host` via the `base_url` field of the configuration model.

Altered existing tests.


api/integrations/segment/serializers.py
```diff
@@ -1,10 +1,23 @@
...
 from integrations.segment.models import SegmentConfiguration
 
 
 class SegmentConfigurationSerializer(BaseEnvironmentIntegrationModelSerializer):
+
+    base_url = serializers.ChoiceField(
+        choices=[
+            constants.DEFAULT_BASE_URL,
+            constants.DUBLIN_BASE_URL,
+        ],
```
khvn26, line 16: I don't enjoy the idea of having to generate a migration each time this list changes on Segment side.

zachaysan: Migration? This is just in the serializers code.

khvn26: Apologies — I stand corrected.

khvn26 (approved):

PR #4694, opened 2024-10-03
docs: Update Rust SDK example

khvn26:

## Changes

This updates the Rust example in accordance with breaking changes introduced in https://github.com/Flagsmith/flagsmith-rust-client/pull/23.

This PR is to be merged after the next Rust SDK release.

Observed the docs preview build.

### DIFF
--- a/docs/docs/clients/server-side.md
+++ b/docs/docs/clients/server-side.md
@@ -515,21 +515,21 @@ buttonData, _ := flags.GetFeatureValue("secret_button")
 <TabItem value="rust" label="Rust">
 
 ```rust
-use flagsmith_flag_engine::identities::Trait;
+use flagsmith::models::SDKTrait;
 use flagsmith_flag_engine::types::{FlagsmithValue, FlagsmithValueType};
 
 let identifier = "delboy@trotterstraders.co.uk";
 
-let traits = vec![Trait {
-            trait_key: "car_type".to_string(),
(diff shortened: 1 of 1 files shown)


docs/docs/clients/server-side.md
```diff
@@ -515,21 +515,21 @@ buttonData, _ := flags.GetFeatureValue("secret_button")
...
+            FlagsmithValue {
                 value: "robin_reliant".to_string(),
                 value_type: FlagsmithValueType::String,
             },
-        }];
+        )];
 
 // The method below triggers a network request
-let identity_flags = flagsmith.get_identity_flags(identifier, Some(traits)).unwrap();
+let identity_flags = flagsmith.get_identity_flags(identifier, Some(traits), None).unwrap();
```
matthewelwell, line 532: `None` seems odd here? Shouldn't this be a Boolean?

khvn26: Argument type is `Option<bool>`, so it should be one of:
- `None`
- `Some(true)`
- `Some(false)`

`Option` type is used to convey that the argument is non-essential, and the user can avoid making an explicit decision by passing `None`.

matthewelwell: Fair enough, it just seems a bit odd to explicitly set `None` instead of `false`. It's not important though, so I'll just approve.

matthewelwell (approved):

PR #4705, opened 2024-10-08
fix: Remove mailer lite

zachaysan:
## Changes

Removing mailer lite from the codebase since it has been deprecated.

Relying on existing test suite to handle failures.


matthewelwell (commented):
Looks good, but I think there's an environment variable that we can still remove in the production task definition [here](https://github.com/Flagsmith/flagsmith/blob/main/infrastructure/aws/production/ecs-task-definition-web.json#L118-L121).

zachaysan:
> Looks good, but I think there's an environment variable that we can still remove in the production task definition [here](https://github.com/Flagsmith/flagsmith/blob/main/infrastructure/aws/production/ecs-task-definition-web.json#L118-L121).

Nice catch @matthewelwell I've removed the env var.

matthewelwell (approved):

PR #4710, opened 2024-10-09
fix: subscription state

### LINKED ISSUE #3952: Navigation issues post plan change in organisation settings
When completing a change of plan in the organisation settings (note that this happens for upgrades and downgrades), the FE shows an error message regarding permissions and the navigation does not show all the options. 

https://www.loom.com/share/3f06d7304a7c477c86412b26349822be?sid=bf0245b7-e21e-4973-9de7-bb78bb934d41

kyle-ssg:

## Changes

Closes https://github.com/Flagsmith/flagsmith/issues/3952

Fixes state when subscription by refreshing the organisations.

Subscribed on staging.

### DIFF
--- a/frontend/common/stores/account-store.js
+++ b/frontend/common/stores/account-store.js
@@ -241,12 +241,12 @@ const controller = {
       .post(`${Project.api}auth/users/`, {
         email,
         first_name,
+        invite_hash: API.getInvite() || undefined,
         last_name,
         marketing_consent_given,
         password,
         referrer: API.getReferrer() || '',
         sign_up_type: API.getInviteType(),
-        invite_hash: API.getInvite() || undefined,
       })
       .then((res) => {
         data.setToken(res.key)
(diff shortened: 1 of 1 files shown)


frontend/common/stores/account-store.js
```diff
@@ -367,20 +367,15 @@ const controller = {
...
       )
       .then((res) => {
-        const idx = _.findIndex(store.model.organisations, {
-          id: store.organisation.id,
+        try {
+          if (res && res.subscription && res.subscription.plan) {
+            API.trackEvent(Constants.events.UPGRADE(res.subscription.plan))
+            API.postEvent(res.subscription.plan, 'chargebee')
+          }
+        } catch (e) {}
```
zachaysan, line 375: I don't really love the empty catch here. Can we silently swallow error conditions that we've allowlisted ahead of time?

zachaysan (approved):
Provisionally approved with one improvement suggested.

frontend/common/stores/account-store.js
kyle-ssg: This is just in case Analytics tools are broke for whatever reason, in which case we'd still see something in sentry. Just needs to ensure the later more important things happen regardless.

PR #4720, opened 2024-10-11
chore: expose timezone setting

matthewelwell:
## Changes

Due to configuration from a specific customer, it might be necessary to expose the timezone setting to work around a misconfiguration of their postgres database(s). 

NOTE: this is a change that will need to be tested by the customer so shouldn't yet be merged. 

Ran the application in docker compose against a database with it's timezone set to something other than UTC, configured the Flagsmith application's TIME_ZONE setting to match the database and verified that times were shown correctly in the dashboard.


matthewelwell:
The issue was instead resolved with postgres configuration.

Issue #3600, opened 2024-03-12
Trying to edit a segment to put a 50% split AFTER some other conditions but after save it reverts to being the first condition


dabeeeenster:
### How are you running Flagsmith

- [ ] Self Hosted with Docker
- [ ] Self Hosted with Kubernetes
- [X] SaaS at flagsmith.com
- [ ] Some other way (add details in description below)

### Describe the bug

This was a customer issue reported over chat. 

@matthewelwell reported:

> It seems as though, in production at least, the API is returning the nested rules in reverse order. I have tried to reproduce this in a unit test using the exact data that reproduces the issue in the production but I have so far not been able to (branch here: https://github.com/Flagsmith/flagsmith/tree/fix/show-segment-rules-in-the-correct-order). More testing in production needed to see if this is also an issue when evaluating segments or not.

### Steps To Reproduce

Unclear

### Expected behavior

Results come back in the correct order/

nathHy:
This is affecting our production instance (SaaS) as well.  We were attempting to target a segment via a trait condition and then do split % on that. 
When creating the segment the UI looks fine, but when hitting save it returns back with the split first.
Let me know if you want any more info to help debug. This is a needed feature for us.

EDIT: A colleague used an existing segment and managed to get the ordering to work? 🤷  My example above was on a brand new segment.

dabeeeenster:
cc @rolodato @kyle-ssg can we investigate pls!

rolodato:
Looking into this now, I'll write back ASAP with details.

rolodato:
I've confirmed this is a bug. To reproduce, create a segment like this:

![Screenshot 2024-05-10 at 15 42 29](https://github.com/Flagsmith/flagsmith/assets/829698/36c349cc-dbfb-4aaf-af52-390c8cadc5e1)

The UI correctly creates the segment with this payload:

<details>

<summary>POST https://api.flagsmith.com/api/v1/projects/{project_id}/segments</summary>

```json
{
  "description": "",
  "name": "debug_3600",
  "project": "16945",
  "rules": [
    {
      "conditions": [],
      "rules": [
        {
          "conditions": [
            {
              "operator": "EQUAL",
              "property": "trait_1",
              "value": "1"
            }
          ],
          "rules": [],
          "type": "ANY"
        },
        {
          "conditions": [
            {
              "operator": "EQUAL",
              "property": "trait_2",
              "value": "2"
            }
          ],
          "rules": [],
[...]

PR #4726, opened 2024-10-12
fix: Duplicated segment conditions on save

### LINKED ISSUE #4725: Segment conditions are duplicated when saving multiple updates and not refreshing

- [ ] Self Hosted with Docker
- [x] SaaS at flagsmith.com

### Describe the bug

https://github.com/user-attachments/assets/3fb16e67-f2b4-459a-8da4-4b6877e5d9eb

### Steps To Reproduce

1. From an existing segment, add a condition `A`
2. Save changes, don't close the modal
3. Add another OR condition `B`
4. Save changes
5. Refresh the page - condition `A` is now duplicated

This happens because after the initial save, the frontend does not save the new ID that the API assigned to condition `A`. When saving condition `B`, it sends condition `A` without an ID, which the API interprets to be an entirely new condition.

First save - in this case, condition `1` existed in the segment before loading. Condition `2`was added here.

Image

Second save - condition `3` was added here, but condition `2` is sent again without an ID. This causes condition `2` to get duplicated.

Image

### Expected behavior

Don't duplicate conditions.

### Screenshots

_No response_

kyle-ssg:

## Changes

Updates segments when they are saved, fixing an issue whereby rules were duplicated after multiple saves.

- Create a segment and save it then save it again
- Update an already created segment then save it again


rolodato (approved):
LGTM, but noting this will make https://github.com/Flagsmith/flagsmith/issues/3600 a lot more evident since it happens immediately on saving without needing to refresh.

PR #4735, opened 2024-10-16
fix(versioning): handle versioned environments for associated-features endpoint

matthewelwell:
## Changes

Adds a new query parameter to the `segments/:id/associated-features/` endpoint which allows us to filter segment overrides by the current environment, and hence we can filter by latest versions. 

Added unit test for the API. 

Tested FE changes locally.


api/features/serializers.py
```diff
@@ -593,6 +593,10 @@ class Meta:
         fields = ("id", "feature", "environment")
 
 
+class AssociatedFeaturesQuerySerializer(serializers.Serializer):
+    environment = serializers.IntegerField(required=False)
```
khvn26, line 597: Just a note to self — we do seem to have a convention to use `entity`, not `entity_id`, for query filters. The rule is only broken for the `evaluation-data` endpoint, which has an `environment_id` filter.

khvn26 (approved):

PR #4738, opened 2024-10-16
feat: combine value + segment change requests for versioned environments

kyle-ssg:

## Changes

This allows segment and environment changes to be created under 1 change requests for versioned environments.

Created change requests changing both the value and the segment overrides and individually.


frontend/web/components/pages/ChangeRequestPage.js
```diff
@@ -371,7 +371,7 @@ const ChangeRequestsPage = class extends Component {
                         >
                           Delete
                         </Button>
-                        {!isVersioned && !changeRequest?.committedAt && (
```
matthewelwell: Was this change supposed to slip in here?

matthewelwell (approved):
Manual testing shows that this works as expected. Add one minor comment.

frontend/web/components/pages/ChangeRequestPage.js
kyle-ssg: Yep looks like it, reverting

PR #4750, opened 2024-10-18
fix: Prevent newlines in environment variables from causing frontend syntax errors

rolodato:

## Changes

If a frontend environment variable such as `FLAGSMITH_ON_FLAGSMITH_API_KEY` includes a trailing newline, the response from `/config/project-overrides` will be invalid JavaScript. For example, running the frontend like this causes a syntax error on the login page:

```
rolodato@rolosmith frontend % FLAGSMITH_ON_FLAGSMITH_API_KEY='ENktaJnfLVbLifybz34JmX
' npm run dev
```

results in this response, which is invalid and causes a frontend syntax error:

```
window.projectOverrides = {
            preventSignup: false,
                    preventEmailPassword: false,
                    preventForgotPassword: false,
                    superUserCreateOnly: false,
                    flagsmith: 'ENktaJnfLVbLifybz34JmX
',
            maintenance: false,
                    disableAnalytics: false,
                    flagsmithAnalytics: true,
                flagsmithRealtime: false,
                    hideInviteLinks: false,
                    linkedinPartnerTracking: false,
                    useSecureCookies: true,
            
    };
```

A customer ran into this when using a manually base64-encoded Kubernetes secret without realising it had a trailing newline, which broke their frontend.

`String.prototype.trim` removes all surrounding whitespace, not just newlines, but I'm 99% certain we will never have environment variables that depend on trailing or leading whitespace.

This can still break if environment variables contain `'`, `/*` or other strings that interfere with JS. We should use a smarter mechanism to inject configuration that cannot cause a syntax error in the frontend, such as injecting a JSON payload and trying to decode that from JS, instead of injecting JS directly.

Manually by running the FE locally, and confirming this works as expected with this change:

```
FLAGSMITH_ON_FLAGSMITH_API_KEY='ENktaJnfLVbLifybz34JmX
' npm run dev
```

### DIFF
--- a/frontend/api/index.js
+++ b/frontend/api/index.js
@@ -40,7 +40,7 @@ app.get('/config/project-overrides', (req, res) => {
             `
     }
 
-    return `    ${name}: '${value}',
+    return `    ${name}: '${value.trim()}',
         `
   }
   const envToBool = (name, defaultVal) => {


matthewelwell (approved):
This seems reasonable to me. 

Minor question: have we checked the behaviour of the python logic that also generates the same file in the single container image set up? It looks to me (based on [this](https://github.com/Flagsmith/flagsmith/blob/main/api/app/views.py#L63)) like it would depend on how the settings attributes are generated, and whether there's any trimming built in there.

rolodato:
This bug only affects the FE image because it's trying to manually generate valid JavaScript. The backend instead generates a JSON payload, which is guaranteed to be valid JS and no trimming is necessary: https://github.com/Flagsmith/flagsmith/blob/fbfe31726b7bee889bca20a2d298edffe4b31019/api/app/views.py#L69.

The failure mode of having unwanted trailing/leading whitespace would be slightly different if using the unified image, as API requests could possibly fail, but at least it won't cause the whole FE to crash.

The FE should probably scrap the current approach of generating JS and instead do the same thing the API is doing, but that would be a more intrusive fix.

PR #4751, opened 2024-10-18
fix: Frontend error when creating SAML configuration if API URL is relative

rolodato:

## Changes

If `Project.api` points to a relative URL such as `/api/v1`, clicking on "Create SAML Configuration" or opening an existing SAML configuration results in an error:

```
Uncaught TypeError: URL constructor: /api/v1/ is not a valid URL.
    CreateSAML CreateSAML.tsx:58
```

This does not happen with the default `npm run dev` because we're setting an absolute URL of `http://localhost:8000/api/v1/`. To reproduce this bug locally, have the FE proxy requests to the BE by running like this:

```
FLAGSMITH_PROXY_API_URL=http://127.0.0.1:8000 npm run dev
```

Manually - using `flagsmith/flagsmith-private-cloud` as the API, try to create a SAML configuration or edit an existing one.

### DIFF
--- a/frontend/web/components/modals/CreateSAML.tsx
+++ b/frontend/web/components/modals/CreateSAML.tsx
@@ -55,7 +55,10 @@ const CreateSAML: FC<CreateSAML> = ({ organisationId, samlName }) => {
     { skip: !samlName },
   )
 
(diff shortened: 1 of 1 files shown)


khvn26 (approved):
LGTM

PR #4753, opened 2024-10-20
fix: Disable is_admin switcher in Organization API Keys

### LINKED ISSUE #3644: When creating an Organisation API key, Free plans should not show the roles toggle
image

The Free plan should always just create a full permission API key.

dougfabris:

## Changes

It proposes to disabled the "is admin" switcher in order to not display roles toggle for free plans

Closes #3644

- Go to organization API Keys
- Create or edit an api key
- Is admin should be disabled for free plans

### DIFF
--- a/frontend/web/components/AdminAPIKeys.js
+++ b/frontend/web/components/AdminAPIKeys.js
@@ -178,9 +178,6 @@ export class CreateAPIKey extends PureComponent {
     const { expiry_date, is_admin, roles, showRoles } = this.state
     const buttonText = this.props.isEdit ? 'Update' : 'Create'
     const buttonSavingText = this.props.isEdit ? 'Updating' : 'Creating'
-    const isFreePlan =
-      AccountStore.getOrganisation().subscription.plan === 'free'
-
     return (
       <>
         <div className='modal-body flex flex-column flex-fill px-4'>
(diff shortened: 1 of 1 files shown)


frontend/web/components/AdminAPIKeys.js
```diff
@@ -213,6 +216,7 @@ export class CreateAPIKey extends PureComponent {
                       })
                     }}
                     checked={is_admin}
+                    disabled={isFreePlan && is_admin}
```
matthewelwell: I think the following is a slightly neater way of doing this, but I appreciate that this is internal pricing logic that isn't necessarily something I expect the open source community to know about. Since the use of non-admin keys is only really useful when RBAC is available, I think we can use the 'RBAC' permission. 

```suggestion
                    disabled={Utils.getPlansPermission('RBAC') && is_admin}
```

By doing this, I think we can then remove the `isFreePlan` const defined above.

matthewelwell (commented):
Hi @dougfabris, thanks very much for this - it's a very good catch. 

I think we can also extend it to add a banner that shows it is only available in our paid plans, but we can handle that internally. For now, getting the switcher disabled would be important for our customers. 

I've added one suggested improvement, but I'd happily merge it and take on the improvement internally if needed.

frontend/web/components/AdminAPIKeys.js
dougfabris: Hey @matthewelwell Nice to know about the plans permissions, thanks for reviewing it! 
I just changed a little bit to `!Utils.getPlansPermission('RBAC')` since the function will return if the plan has the permission

matthewelwell: Good catch adding the leading `!` too, instead of my suggestion!

matthewelwell (approved):
Great, thanks @dougfabris !

PR #4755, opened 2024-10-22
chore(dev): set `DEBUG=True` when enabling DB logging

matthewelwell:
## Changes

Raises an exception on startup if `ENABLE_DB_LOGGING == True and DEBUG != True` since DB logging functionality does not work without the DEBUG environment variable set to true. 

This caught me out for a while so I figured it would be good to ensure it doesn't affect others in the future. 

Ran the application locally with all combinations of `DJANGO_ENABLE_DB_LOGGING` and `DEBUG` environment variables.

### DIFF
--- a/api/app/settings/common.py
+++ b/api/app/settings/common.py
@@ -616,7 +616,7 @@ if APPLICATION_INSIGHTS_CONNECTION_STRING:
     LOGGING["loggers"][""]["handlers"].append("azure")
 
 ENABLE_DB_LOGGING = env.bool("DJANGO_ENABLE_DB_LOGGING", default=False)
-if ENABLE_DB_LOGGING:
+if ENABLE_DB_LOGGING:  # pragma: no cover
     if not DEBUG:
         warnings.warn("Setting DEBUG=True to ensure DB logging functions correctly.")
         DEBUG = True
--- a/api/app/settings/common.py
+++ b/api/app/settings/common.py
@@ -617,6 +617,8 @@ if APPLICATION_INSIGHTS_CONNECTION_STRING:
 
 ENABLE_DB_LOGGING = env.bool("DJANGO_ENABLE_DB_LOGGING", default=False)
 if ENABLE_DB_LOGGING:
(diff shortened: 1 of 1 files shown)


api/app/settings/common.py
```diff
@@ -617,6 +617,8 @@
 
 ENABLE_DB_LOGGING = env.bool("DJANGO_ENABLE_DB_LOGGING", default=False)
 if ENABLE_DB_LOGGING:
+    if not DEBUG:  # pragma: no cover
+        raise ImproperlyConfigured("DB logging will not work without DEBUG=True")
```
khvn26: why not set `DEBUG` to `True` if `DJANGO_ENABLE_DB_LOGGING` is set instead?

matthewelwell: Yeah, I did consider that, and in hindsight I think that's probably better

api/app/settings/common.py
```diff
@@ -617,8 +617,10 @@
 
 ENABLE_DB_LOGGING = env.bool("DJANGO_ENABLE_DB_LOGGING", default=False)
 if ENABLE_DB_LOGGING:
-    if not DEBUG:  # pragma: no cover
-        raise ImproperlyConfigured("DB logging will not work without DEBUG=True")
+    if not DEBUG:
+        warnings.warn("Setting DEBUG=True to ensure DB logging functions correctly.")
+        DEBUG = True
```
khvn26, line 622: Not a fan of re-setting `DEBUG`. I think this is better:
 
```python
DEBUG = env.bool("DEBUG", default=False) or env.bool("DJANGO_ENABLE_DB_LOGGING", default=False)
```

Honestly thinking about it more shouldn't we merge the two and leave just the `DEBUG` environment variable?

gagantrivedi: I don't think merging them is a good idea... `DB_LOGGING` is very verbose

khvn26: Agreed

khvn26 (approved):

PR #4377, opened 2024-07-24
chore(sales-dashboard): optimise sales dashboard search query

matthewelwell:
## Changes

Fix badly performing query in the sales dashboard for certain search terms. 

## How did you test this code?

Ran the sales dashboard locally and grabbed the resulting SQL query from the queryset, ran this against the production database to confirm that the query returned within a reasonable time period.

PR #4757, opened 2024-10-22
fix(sales-dashboard): prevent 500 error when user doesn't exist on sales dashboard search

matthewelwell:
## Changes

Prevents a 500 error when searching the sales dashboard for a user email address that doesn't exist.

Also adds functionality to search by a domain, which will now search across all users with that domain and return any organisations that belong to users with that domain. 

Note that the reason this can't be simplified to e.g. `users__email__endswith` is because of this PR: #4377. 

Added a new unit test.

### DIFF
--- a/api/sales_dashboard/views.py
+++ b/api/sales_dashboard/views.py
@@ -137,8 +137,11 @@ class OrganisationList(ListView):
             query = Q()
 
             if domain_regex.match(search_term):
+                matching_users = FFAdminUser.objects.filter(
+                    email__iendswith=search_term
+                )
                 org_ids = UserOrganisation.objects.filter(
-                    user__email__iendswith=search_term
+                    user__in=matching_users
                 ).values_list("organisation_id", flat=True)
                 query = query & Q(id__in=org_ids)
 
--- a/api/sales_dashboard/views.py
+++ b/api/sales_dashboard/views.py
@@ -33,6 +33,7 @@ from organisations.chargebee.tasks import update_chargebee_cache
 from organisations.models import (
     Organisation,
     OrganisationSubscriptionInformationCache,
+    UserOrganisation,
 )
 from organisations.tasks import (
     update_organisation_subscription_information_cache,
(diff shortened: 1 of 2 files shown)


rolodato:
> grabs the domain and searches for organisations belonging to users with that domain.

This behaviour is super useful, but I don't like it as a fallback, especially for public domains such as `gmail.com`, `icloud.com`, etc. Returning 0 results when searching for a specific user in this case is desirable.

Instead, we could make this explicit when searching for a domain, like `@example.com`, which would return all organisations that have users with a domain of `example.com`.

matthewelwell:
> > grabs the domain and searches for organisations belonging to users with that domain.
> 
> This behaviour is super useful, but I don't like it as a fallback, especially for public domains such as `gmail.com`, `icloud.com`, etc. Returning 0 results when searching for a specific user in this case is desirable.
> 
> Instead, we could make this explicit when searching for a domain, like `@example.com`, which would return all organisations that have users with a domain of `example.com`.

Yeah, I tend to agree. I was just being lazy. I'll update.

api/sales_dashboard/views.py
```diff
@@ -53,6 +54,7 @@
 DEFAULT_ORGANISATION_SORT_DIRECTION = "DESC"
 
 email_regex = re.compile(r"^[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,}$")
+domain_regex = re.compile(r"^[a-z0-9.-]+\.[a-z]{2,}$")
```
rolodato, line 57: I'd prefer if this domain lookup was made explicit, for example by searching for `@example.com` instead of `example.com`.

Contrived example: imagine you're looking for an organisation named `apple.com` and the users all have email domains of `icloud.com`. You would have to search for `apple` or a similar query instead.

By making this explicit we can arbitrarily search for email domains without caring about possible conflicts with organisation names.

rolodato (approved):
Left a minor suggestion, either way it's an improvement and happy to ship it as-is 👍

gagantrivedi (approved):

api/sales_dashboard/views.py
matthewelwell: Let's see if that becomes an issue. I think the current implementation is better UX than what you're suggesting because I imagine people would want to search just the domain, and would likely forget they have to prepend `@` to search on the domain.

PR #4766, opened 2024-10-23
ci: replace trivy scan action with one in Flagsmith/actions repo

matthewelwell:
## Changes

Replace trivy-scan-action with the one added to Flagsmith/actions repo as per [this PR](https://github.com/Flagsmith/actions/pull/4). 

Workflows on this PR should test it (once the new version has been released)


khvn26 (changes requested):
We should also change the cron `.github/workflows/platform-docker-trivy-scan.yml` workflow (or move it to the actions repo).

matthewelwell:
> We should also change the cron `.github/workflows/platform-docker-trivy-scan.yml` workflow (or move it to the actions repo).

Good catch. For now, I've just updated the workflow in this repository. In response to the point about moving it to the actions repository, I added a comment [here](https://github.com/Flagsmith/actions/pull/4#issuecomment-2442134034) to discuss further.

khvn26 (approved):

PR #4767, opened 2024-10-23
docs: Resolve latest SDK versions at build time

rolodato:

## Changes

Dynamically fetch the latest SDK versions at docs build time, instead of manually having to update the docs when any SDK version is released.

Most of the line changes in this PR are only formatting, forcefully introduced by prettier in pre-commit hooks because of the format change from `.md` to `.mdx`.

The way this works is:

* Once, during build time, we fetch all the latest SDK versions from their corresponding package manager APIs and make them available for any doc page to consume from a Docusaurus plugin hook
* If we fail to fetch any version during a production build, the build fails. We might want to mitigate this by adding hardcoded fallback values to avoid blocking the docs build if any remote API is unavailable
* If we fail to fetch any version during a non-production build, we display a placeholder version of `x.y.z`
* If we try to display the version for an invalid SDK, the build fails
* Known versions are pinned to a semver specification such as `~2`, which means we only need to update 1 file whenever any new major SDK version is released.

If the docs are never built, the SDK versions are never updated. This is still much better than the current situation as the docs are generally updated at a reasonable cadence, but could be further mitigated by allowing manual redeploys of the docs from a GitHub Action, or having each SDK repository trigger a docs rebuild after a new version is published.

Downsides of this approach:

* Docusaurus does not fully support `<CodeBlock>` elements if the content is dynamic, so we lose out on syntax highlighting, the "copy to clipboard" button on hover, and the ability to show a title/filename header on any code blocks with dynamic content such as version numbers
* Using `<` characters and sensitive whitespace within MDX code blocks is much more complicated than in plain Markdown fenced code blocks
* We are adding a hard dependency on remote package manager APIs for the docs build. If any of these APIs fail, the docs build will fail. We might want to persist the "latest fetched versions" in this same repository, and read the known versions from that file instead.
* ~It might be difficult to coordinate releasing new SDK versions and updating the corresponding docs at the same time. We'll need to become much more diligent at this if we choose this approach. This could be mitigated by "pinning" the rendered versions in docs to specific major versions, which lets us decouple new version releases from the corresponding docs changes, at the cost of more complexity in the docs build logic.~ solved by https://github.com/Flagsmith/flagsmith/pull/4767/commits/ed4365d958cc51baa1f20a44573146222a45e97e

* https://docs-git-docs-build-time-sdk-versions-flagsmith.vercel.app/clients/server-side?language=dotnet
* https://docs-git-docs-build-time-sdk-versions-flagsmith.vercel.app/clients/server-side?language=java
* https://docs-git-docs-build-time-sdk-versions-flagsmith.vercel.app/clients/server-side?language=rust
* https://docs-git-docs-build-time-sdk-versions-flagsmith.vercel.app/clients/server-side?language=elixir
* https://docs-git-docs-build-time-sdk-versions-flagsmith.vercel.app/clients/android
* https://docs-git-docs-build-time-sdk-versions-flagsmith.vercel.app/clients/ios
* https://docs-git-docs-build-time-sdk-versions-flagsmith.vercel.app/quickstart


matthewelwell (commented):
This looks great to me. 

I don't think any of the downsides are strong enough for me that they outweigh the benefits here. I think we can accept the lack of syntax highlighting for the blocks that actually need the version.

khvn26:
I'd rather invest in a centralised Docusaurus build that pulls SDK docs from their respective repos.

An ability to ship SDK documentation together with SDK code seems logical to me, and a better alternative to what we have now. I believe the documented latest version problem can be solved by this in a more precise manner, too.

IMO the SDK documentation is more valuable than the example projects that are currently maintained with SDK repos.

rolodato:
Pulling SDK docs from each SDK's repo is a much more intrusive change and it's more difficult to guarantee consistency across docs (things like linting, broken link/anchor checking, using Docusaurus modules if we need more than plain Markdown etc) if the content is spread out across multiple repositories. It might be desirable to eventually go there, but I still think this PR is an incremental step in the right direction (pulling some information at build from an external source).

The source of truth for latest versions is still the actual package manager APIs - whatever we pull from our repositories is not guaranteed to be actually downloadable by package managers. So even if we do pull docs from each SDK's repo, I still see a use case for pulling in the latest (relevant) SDK versions from package manager APIs. This is mostly helpful for package managers that do not have an "install latest and save to dependency manifest" command such as Maven, Gradle, Hex, Cargo, or JS CDNs.

I think a necessary condition before merging a change like this is being able to "pin" whatever we're fetching to a specific major version. That way we only need to update the docs manually whenever we release an SDK major version. I'll look into adding this before making this PR ready to review.

docs/docs/clients/client-side/android.md
```diff
@@ -5,35 +5,18 @@ description: Manage your Feature Flags and Remote Config in your Android applica
...
-In the new Gradle version 7+ update your `settings.gradle` file to include JitPack if you haven't already
-
-```groovy
-repositories {
-    google()
-    mavenCentral()
-
-    maven("https://jitpack.io")
-}
-```
```
matthewelwell: We can't remove this yet, unless we change the below to `io.flagsmith` - I'm still working on finalising the deprecation of jitpack.

docs/src/components/SdkVersions.js
```diff
@@ -0,0 +1,26 @@
+import { usePluginData } from '@docusaurus/useGlobalData';
+import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
+import { maxSatisfying } from 'semver';
+
+export const Version = ({ sdk, spec = '*', options = {} }) => {
```
matthewelwell: Based on the below, do we actually need to export this function anymore?

docs/docs/clients/client-side/android.md
rolodato: Changed to use `io.flagsmith` - jitpack is more friction and can be a blocker as we've seen in the past

docs/src/components/SdkVersions.js
rolodato: Fixed

docs/docs/clients/client-side/android.md
matthewelwell: Hmm... but I'm hoping that `io.flagsmith` is temporary and we will (in the not too distant future, when I find time to understand the migration process for our java client) move to `com.flagsmith`. 

I think the move from jitpack is a bigger change than just hiding it in with these changes.

rolodato: The [`com.flagsmith` release](https://mvnrepository.com/artifact/io.flagsmith.flagsmith-kotlin-android-client/flagsmith-kotlin-android-client) is still on 1.6.1 which does not include `getIdentityFlags`. I'd prefer to ship with `io.flagsmith` even if it's temporary since it has all the latest features and customers are not blocked if they need a bugfix before we can sort out publishing to `com.flagsmith`. We can figure out how to publish to `com.flagsmith` later and it's a trivial change for consumers if they eventually want to migrate.

matthewelwell: > The [`com.flagsmith` release](https://mvnrepository.com/artifact/io.flagsmith.flagsmith-kotlin-android-client/flagsmith-kotlin-android-client) is still on 1.6.1 which does not include getIdentityFlags

Sorry, I'm confused. The link you included isn't to the `com.flagsmith` (because there isn't one in maven central yet!), it's to the one where I had incorrectly included the name of the package in the namespace (see [here](https://github.com/Flagsmith/flagsmith-kotlin-android-client/pull/57/commits/ef6c30034063dc808649927481d987aa757ae4f7#diff-458e7f80b7f7dc9707ace9d8e9e81c049f8ea426e1bffa418e5b084c29a9dffeR199)). 

To be clear, there is only the io.flagsmith releases in maven central [here](https://central.sonatype.com/artifact/io.flagsmith/flagsmith-kotlin-android-client/versions) and the jitpack releases are just pulled straight from Github releases.

I'm not sure it's worth continuing to argue this though, we can go with what you have here if you think it's the right thing to do.

matthewelwell (approved):

dabeeeenster:
We could cronjob (probably weekly to maintain sanity) the docusaurus builds that pulls in SDK versions at compile time?

dabeeeenster:
This is great btw @rolodato ! Way above my React paygrade.

PR #4769, opened 2024-10-24
docs: add rollback instructions

matthewelwell:
## Changes

Adds documentation for a manual rollback process. 

Ran the query, which outputs something like the following: 

image

and subsequently ran the python commands that were obtained. 

image

### DIFF
--- a/docs/docs/deployment/hosting/locally-api.md
+++ b/docs/docs/deployment/hosting/locally-api.md
@@ -451,8 +451,8 @@ The Django `collectstatic` command then copies all the additional static assets 
 
 :::warning
 
-These steps may result in data loss in the scenario where new models have been added to the database. We recommend
-taking a full backup of the database before completing the rollback.
+These steps may result in data loss in the scenario where new models or fields have been added to the database. We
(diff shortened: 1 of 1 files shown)


docs/docs/deployment/hosting/locally-api.md
```diff
@@ -447,6 +447,38 @@ Webpack compiles a front end build, sourcing `api/app/templates/index.html`. It
 The Django `collectstatic` command then copies all the additional static assets that Django needs, including
 `api/app/templates/webpack/index.html`, into `api/static`.
 
+## Rolling Back to a Previous Version of Flagsmith
```
rolodato:
```suggestion
## Rolling back to a previous version
```

rolodato (approved):

docs/docs/deployment/hosting/locally-api.md
matthewelwell: I've done this to be consistent, but this is a title... it should be title case. But I'm not about to start updating all the titles!

PR #4770, opened 2024-10-24
fix: Fix stored XSS when rendering tooltips

rolodato:

## Changes

Fixes a stored XSS when rendering tooltips. This can be triggered with the following actions:

* Enabling "Compact view" in the Features list and hovering over a feature with an XSS payload in its description
* Comparing two environments or identities in a project that has a feature with an XSS payload in its description
* Visiting the identity details page for any identity in a project containing a feature with an XSS payload in its description

Example description payload:

```
<img src=x onerror=alert('hello')>
```

Locally by performing the above actions and comparing them to the current release version of the frontend.

### DIFF
--- a/frontend/web/components/Tooltip.tsx
+++ b/frontend/web/components/Tooltip.tsx
@@ -2,6 +2,7 @@ import React, { FC, ReactElement, ReactNode, useRef } from 'react'
 import ReactTooltip, { TooltipProps as _TooltipProps } from 'react-tooltip'
 import Utils from 'common/utils/utils'
 import classNames from 'classnames'
+import { sanitize } from 'dompurify'
 
 export type TooltipProps = {
   title: ReactNode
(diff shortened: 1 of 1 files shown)


frontend/web/components/Tooltip.tsx
```diff
@@ -44,7 +45,7 @@ const Tooltip: FC<TooltipProps> = ({
           ) : (
             <div
               style={{ wordBreak: 'break-word' }}
-              dangerouslySetInnerHTML={{ __html: children }}
+              dangerouslySetInnerHTML={{ __html: sanitize(children) }}
```
matthewelwell, line 48: I don't really understand why we're even using `dangerouslySetInnerHTML` here at all? Shouldn't we be able to just render out the text as plain text?

matthewelwell (approved):
I'm approving this as I've tested all of the use cases described in the PR (and mentioned by a customer) so I'm happy that it fixes the immediate issue but I do think that the question I've added is a valid one that should be addressed at some point.

PR #4781, opened 2024-10-29
fix: associated segment override check

### LINKED ISSUE #4772: Users with permissions to manage segments and segment overrides cannot create segment overrides from the Segment

- [ ] Self Hosted with Docker
- [x] SaaS at flagsmith.com

### Describe the bug

Users with the project-level permission `MANAGE_SEGMENTS` and the environment-level permission `MANAGE_SEGMENT_OVERRIDES` should be able to add, and modify segment overrides when viewing the 'Features' tab from a segment. Currently they cannot do so. 

### Steps To Reproduce

1. Log into the Flagsmith dashboard as a user with permission to manage segments and segment overrides
2. Navigate to any segment, and click on the 'Features' tab
3. Notice that the option to create a segment override does not exist

### Expected behavior

The option to create a segment override does exist

### Screenshots

User with correct permissions:

Image

User with admin permissions:

Image

kyle-ssg:

## Changes

Fixes associated segment override permission checks to use the selected environment.

- As a user with manage segment override permissions on the demo environment but not production we see the following 

image
image

### DIFF
--- a/frontend/web/components/modals/AssociatedSegmentOverrides.js
+++ b/frontend/web/components/modals/AssociatedSegmentOverrides.js
@@ -14,6 +14,7 @@ import { getStore } from 'common/store'
 import { getEnvironment } from 'common/services/useEnvironment'
 import { saveFeatureWithValidation } from 'components/saveFeatureWithValidation'
 import Utils from 'common/utils/utils'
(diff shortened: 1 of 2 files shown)


matthewelwell (approved):
Tested manually - looks good to me.

PR #4805, opened 2024-11-05
feat: Scheduled segment overrides

### LINKED ISSUE #4492: Allow scheduling changes to segment overrides without change requests enabled

In https://github.com/Flagsmith/flagsmith/pull/3790, we implemented change requests for versioned segment overrides.

Using this new functionality requires change requests to be enabled, which is not always desirable.

For comparison, changes to versioned features in environments without change requests can be scheduled by using the "Schedule Update" button:

image

Allow scheduling changes to versioned change requests in environments that do not have change requests enabled.

As a workaround, it's possible to:

1. Enable change requests with 0 approvals
2. Create the versioned segment override's change request
3. Ship the change request you just created

### Additional context

_No response_

kyle-ssg:

## Changes

Adds the ability to schedule segment overrides. 

image

Schedule segment overrides, view scheduled changes

image


matthewelwell:
@kyle-ssg I'm a little confused by this one: 

Firstly, I don't understand how the following can be true since scheduled updates rely on change requests under the hood, and we don't support change requests for segments in non-versioned environments. 

> This applies to both versioned and non-versioned environments

Secondly, when I tried to test it, the first thing that happened was that I was just redirected to the billing page directly which confused me. I think it's because scheduled flags aren't available for non-paid plans, but I'm unsure why we're not using the pattern we have elsewhere that shows a banner that it's only available in another plan. 

Then, when I tested it in an organisation that is using the enterprise plan, using a non-versioned environment, it just appears to immediately apply the change, I don't see the scheduling modal pop up.

kyle-ssg:
> Firstly, I don't understand how the following can be true since scheduled updates rely on change requests under the hood, and we don't support change requests for segments in non-versioned environments. 

Ah yep correct, for some reason I thought this worked for both.


>  Secondly, when I tried to test it, the first thing that happened was that I was just redirected to the billing page directly which confused me. 

I've updated the UI to be as such, I don't think we'd want to show a banner here, a simple lock badge should be fine 
image


> Then, when I tested it in an organisation that is using the enterprise plan, using a non-versioned environment, it just appears to immediately apply the change, I don't see the scheduling modal pop up. 

I believe this was due to being tested on a non-versioned project which as you said doesn't work.

matthewelwell (approved):

PR #4807, opened 2024-11-05
chore: Use local api for self hosted snippets

### LINKED ISSUE #4130: All code examples should be instance-aware
Thanks to @rolodato for reporting this.

For every Flagsmith deployment besides the default Flagsmith SaaS environment, a correct base URL should be added to all SDK init code in all of the code examples.

kyle-ssg:

## Changes

Uses admin api for SDK snippets instead of flagsmith on flagsmith url. Closes https://github.com/Flagsmith/flagsmith/issues/4130

Viewed code snippets in all languages


matthewelwell (dismissed):

matthewelwell:
I'm not sure if this is a symptom of using the preview URL, but it's showing the API URL being overridden here which I Don't think is correct? 

image

kyle-ssg:
> I'm not sure if this is a symptom of using the preview URL, but it's showing the API URL being overridden here which I Don't think is correct?

I suppose in this exact case where they are self hosting but the admin api is still api.flagsmith.com, I think I'll exclude it just for that case.

kyle-ssg:
This should be resolved now the api url for the sdk docs will adhere to the following 

- If using saas or the frontend is configured to point to our production api it'll use the edge api
- Otherwise use wherever the frontend is hitting for admin endpoints


```
 getFlagsmithSDKUrl: () =>
    Utils.isSaas() || Project.api.startsWith('https://api.flagsmith.com')
Dismissed
      ? Project.flagsmithClientEdgeAPI
      : Project.api,
```

matthewelwell:
Sorry, I think it also doesn't work very well in the unified image? 

Screenshot 2024-11-27 at 11 21 42

frontend/common/constants.ts
```diff
@@ -441,6 +441,10 @@ export default {
       'TRAITS_ID': 150,
     },
   },
+  getFlagsmithSDKUrl: () =>
+    Utils.isSaas() || Project.api.startsWith('https://api.flagsmith.com')
+      ? Project.flagsmithClientEdgeAPI
+      : Project.api,
```
matthewelwell: I don't really know what the solution is here, but we need to account for when `Project.api` is just `/api/v1` as per the above screenshot.

matthewelwell (changes requested):

frontend/common/constants.ts
kyle-ssg: Ah I didn't expect that to be relative, should be fixed

rolodato: If it helps, this is how I solved a similar problem for displaying SAML URLs based on absolute or relative API URLs: https://github.com/Flagsmith/flagsmith/blob/bb7849c799fff98c1750adf1d5a057479994d41b/frontend/web/components/modals/CreateSAML.tsx#L58

matthewelwell:
I'm going to make a couple of tweaks to this PR: 

 1. Sort out the formatting for the python snippet because it should conform to `black`
 2. Fix errors like this one that are technically unrelated to this PR but would cause awful conflicts if I implement elsewhere

Screenshot 2024-12-04 at 13 26 18

matthewelwell (approved):

PR #4809, opened 2024-11-05
fix: Edit identity override with prevent flag defaults enabled

### LINKED ISSUE #4771: Prevent Flag Defaults settings makes identity flags unusable

- [x] Self Hosted with Docker
- [ ] SaaS at flagsmith.com

### Describe the bug

in [CreateFlag.js](https://github.com/Flagsmith/flagsmith/blob/main/frontend/web/components/modals/CreateFlag.js#L1905-L1910) the `prevent_flag_defaults` is passed as a  parameter to the `<Value />` component, in which basically the input for `hideValue` is set to true. This makes an identity override unusable unless this project setting is disabled

### Steps To Reproduce

1. Go to project Settings
2. Enable `Prevent Flag Defaults` 
3. Go on an environment 
4. Select a flag
5. Attempt to override the value for a particular user

### Expected behavior

As a user I should still be able to override a flag for a particular identity.

### Screenshots

_No response_

dougfabris:

## Changes

It adds the identity verification in `hideValue` param on the `Value` component in order to render the input fields, where wasn't possible to edit under identity override

Closes: #4771

- Go to project settings
- Enable the prevent flag defaults
- Create a feature 
- Go to identities
- Clicking on the feature you created you should be able to edit it

**Note to the reviewer**: I was looking into it and indeed removing the parameter mentioned to the `Value` component seems to solve the issue. But I'm afraid I'm missing something because the verification using `prevent_flag_defaults` was added 2 years ago here #1453

### DIFF
--- a/frontend/web/components/modals/CreateFlag.js
+++ b/frontend/web/components/modals/CreateFlag.js
@@ -721,7 +721,7 @@ const CreateFlag = class extends Component {
       </>
     )
 
-    const Value = (error, projectAdmin, createFeature) => {
+    const Value = (error, projectAdmin, createFeature, hideValue) => {
       const { featureError, featureWarning } = this.parseError(error)
       return (
         <>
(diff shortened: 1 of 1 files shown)


kyle-ssg:
Hey @dougfabris, thanks for the PR!

Prevent_flag_defaults aims to stop from defining a default remote configuration when creating a flag (which sets a default for all environments), forcing people to set it later per environment. Although your PR does fix an issue where it prevents user overrides, it also stops prevent_flag_defaults from working - we can define on a remote config value on flag creation with this change.

Instead this should do something along the lines of ``project.prevent_flag_defaults && !identity``

dougfabris:
So in fact I was missing something then. I didn't knew about the remote config you mentioned! 
I've applied your suggestion. Thanks for reviewing it @kyle-ssg

matthewelwell (approved):
Thanks @dougfabris !

PR #4816, opened 2024-11-06
fix: Avoid a long table lock for the `first_name` migration

khvn26:

## Changes

This replaces `users.0039` migration with a manual one that aims to avoid long locking the `users_ffadminusers` table during migration execution.

A new `users_ffadminusers.first_name_v2` column is added and is backing the `FFAdminUser.first_name` field. 

Notes:
- Should we replace the migraton or rather add one? Currently, to deploy this on top of the previous version, a `migrations` table mofification is in order.
- Alternatively, we could probably do away with adding `ALTER TABLE {table_name} RENAME COLUMN first_name_v2 TO first_name` statement to the forward migration.
- This is most likely not compatible with Oracle due to syntax differences.

Added two unit tests for reverse and forward migrations.


matthewelwell:
Alternate implementation here: #4817

matthewelwell:
Superseded by #4817

PR #4817, opened 2024-11-06
fix: replace alter field with adding a new field

matthewelwell:
## Changes

Alternate implementation of #4816

To test the updates to migration 0039, I ran the migration locally. Verified that `python manage.py makemigrations users` doesn't think there are any new migrations to make. 

To test the new migration 0040, I ran it in an environment which had run both implementations of 0039 and observed that the behaviour was correct. 

In both cases, I ran the API locally post migration and verified that the application runs as expected.

Note that I was able to also add a migration test for 0039 but this would be quite challenging to achieve for 0040.


api/users/migrations/0039_alter_ffadminuser_first_name.py
```diff
@@ -1,15 +1,22 @@
...
+def populate_new_first_name_field(apps: Apps, schema_editor: BaseDatabaseSchemaEditor) -> None:
     FFAdminUser = apps.get_model("users", "FFAdminUser")
 
     FFAdminUser.objects.update(first_name_v2=F("first_name"))
 
 
+def populate_old_first_name_field(apps: Apps, schema_editor: BaseDatabaseSchemaEditor) -> None:
+    FFAdminUser = apps.get_model("users", "FFAdminUser")
+
+    FFAdminUser.objects.update(first_name=F("first_name_v2"))
```
khvn26:
```suggestion
    FFAdminUser.objects.update(first_name=F("first_name_v2")[:30])
```

api/users/migrations/0040_alter_ffadminuser_first_name_fix.py
```diff
@@ -0,0 +1,42 @@
...
+This migration exists because 0039 was retrospectively fixed to prevent locking issues
+on the users table. This migration ensures that:
+
+ 1. in environments where 0039 was run before the fix was added, the new column is
+    correctly added to the database to match what the ORM expects and the data
+    is correctly migrted to the new column.
+ 2. in environments where 0039 was run after the fix was added, nothing happens.
+
+Note that we only need to care about Postgres databases here because we did not release
+the bad migration for 0039 in a format that can be consumed by oracle / mysql.
```
khvn26, line 11: Nice 👍

khvn26 (approved):

PR #4822, opened 2024-11-08
fix: Handle invalid colour codes on tags, allow default colours

rolodato:

## Changes

Fixes the frontend breaking when rendering tags that have an invalid colour.

To reproduce, create a tag with a colour code that is not accepted by `color`. Example (missing `#`):

```console
curl 'http://localhost:8000/api/v1/projects/34/tags/' -X POST -H 'authorization: Token ...' -H 'content-type: application/json' --data-raw '{"color":"ff0000","label":"my-tag"}'
```

Trying to render a tag with this colour crashes the frontend:

```
Error: Unable to parse color from string: ff0000
    at new c (index.js:53:10)
    at c (index.js:28:10)
    at TagContent.tsx:98:38
    at t.default (TagContent.tsx:137:8)
    at Ja (react-dom.production.min.js:153:146)
    at ks (react-dom.production.min.js:261:496)
    at Cl (react-dom.production.min.js:246:265)
    at Sl (react-dom.production.min.js:246:194)
    at gl (react-dom.production.min.js:239:172)
    at react-dom.production.min.js:123:115
```

The reporter of this bug found it when creating a tag using our Terraform provider: https://registry.terraform.io/providers/Flagsmith/flagsmith/latest/docs/resources/tag#tag_colour-1.

The API currently requires forces tags to be created with a colour. Tag colours are entirely visual so there is no reason to force users to provide a value. This change also allows creating tags in the API without specifying a colour.

https://github.com/Flagsmith/terraform-provider-flagsmith/pull/166 makes tag colours optional in the Terraform provider.

Manually by looking at tags rendered in different parts of the app.

<details>
<summary>
</summary>

</details>


frontend/common/utils/utils.tsx
```diff
@@ -110,11 +111,20 @@ const Utils = Object.assign({}, require('./base/_utils'), {
     return typeof value === 'number'
   },
 
+  colour(c: string, fallback = '#6837FC'): InstanceType<typeof Color> {
```
kyle-ssg: Maybe fallback=Constants.tagColors[0]

kyle-ssg (approved):

PR #4840, opened 2024-11-18
fix: Speed up identity overrides

### LINKED ISSUE #4639: Identity override count retrieval is slow for large numbers of overrides
For instance, for an environment with ~50k overrides, retrieving features with override counters takes 15 to 80 seconds.

Things to explore:
1. Look into DynamoDB access pattern and ways to optimise it
1. Cache/store identity counters elsewhere

Related PR that likely introduced the issue is here #3164. Note that this only then became an issue once we migrated customers with large numbers of identity overrides as part of the work for local identity overrides.

zachaysan:
## Changes

This changes the way that identity overrides are requested for the features page in two ways. First, instead of the entire document getting paginated and returned to the client, only one page per feature is returned. Second, instead of every feature being returned in the network call, we only serve up the features for the current page.

I've written one new test, although I'm awaiting codecov to let me know where else needs testing, since I'm sure some are missing.


api/environments/dynamodb/wrappers/environment_wrapper.py
```diff
@@ -66,23 +67,70 @@ def get_identity_overrides_by_environment_id(
...
         try:
-            return list(
-                self.query_get_all_items(
-                    KeyConditionExpression=Key(ENVIRONMENTS_V2_PARTITION_KEY).eq(
-                        str(environment_id),
+            if not limit_feature_identities_to_one_page:
+                if feature_ids is not None:
+                    raise NotImplementedError(
+                        "Multiple feature ids is currently not supported "
+                        "when not limiting to a single page of features"
```
khvn26: It seems like `feature_id`, `feature_ids` and `limit_feature_identities_to_one_page` could be compressed to a single `feature_ids` argument. This would significantly simplify the logic too.

api/environments/dynamodb/wrappers/environment_wrapper.py
```diff
@@ -66,23 +67,70 @@ def get_identity_overrides_by_environment_id(
...
+
+                futures = []
+                with ThreadPoolExecutor() as executor:
+                    for feature_id in feature_ids:
+                        futures.append(
+                            executor.submit(
+                                self.get_page_of_feature_identities,
+                                environment_id,
+                                feature_id,
+                            )
```
khvn26: This certainly looks like a nice solution — I haven't thought of running several queries in parallel when thinking of solving the issue.

Have you considered cases when one feature has >1MB of identity overrides data? The original issue mentioned introducing approximate counters for those (e.g. `"1000+"`) but it doesn't seem to be implemented here.

api/environments/dynamodb/wrappers/environment_wrapper.py
```diff
@@ -66,23 +67,70 @@ def get_identity_overrides_by_environment_id(
...
+                    result = future.result()
+                    for item in result:
+                        results.append(item)
+
+                return results
+
         except KeyError as e:
             raise ObjectDoesNotExist() from e
 
+    def get_page_of_feature_identities(
```
khvn26: Not a fan of introducing `feature_identities` name while having established `identity_overrides`. Also, not a fan of using prepositions in function names.

```suggestion
    def get_identity_overrides_page(
```

api/tests/unit/features/test_unit_features_views.py
```diff
@@ -2062,6 +2063,38 @@ def test_list_features_provides_segment_overrides_for_dynamo_enabled_project(
...
+
+    # When
+    response = admin_client_new.get(url)
+
+    # Then
+    assert response.status_code == status.HTTP_200_OK
+    mock_get_overrides_data.assert_called_once_with(
+        dynamo_enabled_project_environment_one,
+        [feature.id],
+    )
```
khvn26, line 2095: This test seems to fail in CI. Also, in addition to a mocked unit test, I would expect to see a functional test that places actual data in fixtured Dynamo tables.

api/environments/dynamodb/wrappers/environment_wrapper.py
```diff
@@ -66,23 +67,70 @@ def get_identity_overrides_by_environment_id(
...
+        query_response = self.table.query(
+            KeyConditionExpression=Key(ENVIRONMENTS_V2_PARTITION_KEY).eq(
+                str(environment_id),
+            )
+            & Key(ENVIRONMENTS_V2_SORT_KEY).begins_with(
+                get_environments_v2_identity_override_document_key(
+                    feature_id=feature_id,
+                ),
+            )
+        )
```
khvn26: nit: The query is identical to the one used in `get_identity_overrides_by_environment_id` method — maybe it makes sense to DRY this out.

api/environments/dynamodb/wrappers/environment_wrapper.py
```diff
@@ -66,23 +67,70 @@ def get_identity_overrides_by_environment_id(
...
+                        futures.append(
+                            executor.submit(
+                                self.get_page_of_feature_identities,
+                                environment_id,
+                                feature_id,
+                            )
+                        )
+
+                results = []
+                for future in as_completed(futures):
```
khvn26: I don't think I understand the need to use `as_completed` here.

khvn26 (changes requested):
Added a bunch of questions + not sure if the original issue is actually solved by this.

api/environments/dynamodb/wrappers/environment_wrapper.py
zachaysan: It's just to process results as they become available rather than waiting for all tasks to finish before proceeding.

khvn26: This doesn't do anything as all the submitted tasks are completed by the time of this call. 

The tasks are done by implicitly calling `ThreadPoolExecutor.__exit__` as you exit from the `with ThreadPoolExecutor()` block.

api/environments/dynamodb/wrappers/environment_wrapper.py
zachaysan: Sure, sounds good. Updated.

api/environments/dynamodb/wrappers/environment_wrapper.py
zachaysan: Good idea. Updated.

api/tests/unit/features/test_unit_features_views.py
zachaysan: I've created a functional test with the Dynamo fixtured tables.

api/environments/dynamodb/wrappers/environment_wrapper.py
zachaysan: I've added a new field to mark identity overrides as being over the limit but this will need to be supported by the frontend with a new field.

api/environments/dynamodb/wrappers/environment_wrapper.py
zachaysan: We need to keep the distinction between `feature_id` and `feature_ids` to support existing logic, but I was able to remove the `limit_feature_identities_to_one_page` and rely on `feature_ids` for the pre-existing logic.

api/tests/unit/features/test_unit_features_views.py
```diff
@@ -2719,6 +2719,10 @@ def test_list_features_n_plus_1_without_rbac(
     django_assert_num_queries: DjangoAssertNumQueries,
     environment: Environment,
 ) -> None:
+    """
+    NOTE: When running locally, this test can come up with an extra query.
+          It should be tested against CI to ensure it passes.
+    """
```
khvn26: A flaky test should be fixed. I'll appreciate a TODO and an issue link here.

api/environments/dynamodb/wrappers/environment_wrapper.py
zachaysan: Oh of course. Silly me. I've removed the `as_completed` code.

api/tests/unit/environments/dynamodb/wrappers/test_unit_dynamodb_environment_v2_wrapper.py
```diff
@@ -59,6 +59,45 @@ def test_environment_v2_wrapper__get_identity_overrides_by_environment_id__retur
...
+    # When
+    results = wrapper.get_identity_overrides_by_environment_id(
+        environment_id=environment.id,
+        feature_ids=[feature.id],
+    )
+
+    # Then
+    assert len(results) == 1
+    override_document["more_identity_overrides"] = False
+    assert results[0] == override_document
```
khvn26: Assignments in the part that's supposed to only contain assertions don't sit well with me.

```suggestion
    assert results[0] == {**override_document, "more_identity_overrides": False}
```

api/environments/dynamodb/wrappers/environment_wrapper.py
```diff
@@ -116,21 +102,33 @@ def get_identity_overrides_by_environment_id(
...
-                    feature_id=feature_id,
-                ),
+            KeyConditionExpression=self.get_identity_overrides_key_condition_expression(
+                environment_id=environment_id,
+                feature_id=feature_id,
             )
         )
+        last_evaluated_key = query_response.get("LastEvaluatedKey")
+        for item in query_response["Items"]:
+            item["more_identity_overrides"] = last_evaluated_key is not None
```
khvn26: I understand the desire to preserve the function return type, but this attribute does not belong to individual items.

Let's return a typed dict with `items: list[dict[str, Any]` and a `more_identity_overrides: bool` keys.

I would also suggest to think on the naming some more. I prefer the boolean attribute names to be more apparent, e.g. `is_num_identity_overrides_complete`.

api/environments/dynamodb/types.py
```diff
@@ -88,6 +88,7 @@ class IdentityOverrideV2(BaseModel):
     identifier: str
     identity_uuid: str
     feature_state: FeatureStateModel
+    more_identity_overrides: bool = False
```
khvn26: I didn't check thoroughly, but this added field might prompt the backend to index this field in DynamoDB! It doesn't belong here.

api/features/serializers.py
```diff
@@ -145,6 +145,11 @@ class CreateFeatureSerializer(DeleteBeforeUpdateWritableNestedModelSerializer):
         "in the environment provided by the `environment` query parameter. "
         "Note: will return null for Edge enabled projects."
     )
+    more_identity_overrides = serializers.SerializerMethodField(
+        help_text="A boolean that indicates whether there are more"
+        " identity overrides than are being listed in order to "
+        "improve performance of the features list endpoint."
+    )
```
khvn26: Yeah, the name I suggested looks better to me. Additionally, I updated the comment to reflect the specific use case more:

```suggestion
    is_num_identity_overrides_complete = serializers.SerializerMethodField(
        help_text="A boolean that indicates whether there are more"
        " identity overrides than are being listed. This field might be"
        " `True` when querying overrides data for a features list page."
    )
```

api/tests/unit/environments/dynamodb/wrappers/test_unit_dynamodb_environment_v2_wrapper.py
```diff
@@ -59,6 +59,45 @@ def test_environment_v2_wrapper__get_identity_overrides_by_environment_id__retur
...
+
+
```
khvn26, line 137: I appreciate the coverage, but I still would like to see a test for a case when the new boolean field should be `True`.

khvn26 (changes requested):

api/environments/dynamodb/types.py
matthewelwell: I'm not 100% what you mean by 'index' here, but yes, I believe that adding this field here would mean that it gets written to each record. Hence, it would be included in the index that we have on that table which projects all attributes. 

I agree with Kim that this is not desirable for sure. 

If this approach simplifies the code though, we could just sub class `IndentityOverrideV2`, add this new field and use the sub class for whatever logic relies on the field.

khvn26: Another problem, which I pointed out [here](https://github.com/Flagsmith/flagsmith/pull/4840/files#r1868126098), is that this field doesn't belong to an individual override document.

matthewelwell: Got it, yep, agreed.

api/tests/unit/features/test_unit_features_views.py
zachaysan: Sounds good. I've added a `TODO` and a link to a GitHub issue to the docstring.

api/environments/dynamodb/wrappers/environment_wrapper.py
zachaysan: As discussed over a call with @khvn26 we're going with a `dataclass` over a `TypedDict` for the return type.

api/tests/unit/environments/dynamodb/wrappers/test_unit_dynamodb_environment_v2_wrapper.py
zachaysan: This ended up being a boot point since the refactor brought in dataclass responses. That said, I appreciate your approach over mine for whatever its worth.

api/tests/unit/environments/dynamodb/wrappers/test_unit_dynamodb_environment_v2_wrapper.py
zachaysan: Ok I've made a new DynamoDB-related test with a bunch of identities to force it to page the request. The test now passes.

api/environments/dynamodb/types.py
zachaysan: Ok this has now been removed.

api/features/serializers.py
zachaysan: I tweaked your wording to be more explicit on the usecase of the field but overall kept your suggestions intact.

khvn26 (approved):

api/environments/dynamodb/wrappers/environment_wrapper.py
```diff
@@ -66,23 +74,66 @@ def get_identity_overrides_by_environment_id(
         self,
         environment_id: int,
         feature_id: int | None = None,
-    ) -> typing.List[dict[str, Any]]:
+        feature_ids: None | list[int] = None,
+    ) -> list[dict[str, Any]] | list[IdentityOverridesQueryResponse]:
```
khvn26, line 78: Since the PR has been approved this is a non-blocking comment but the polymorphic return type does look like a code smell.

At this point, factoring the `feature_ids is not None` branch to a separate method looks like a good solution.

zachaysan: Talked with @matthewelwell about it and we're going to leave that change to a subsequent PR.

PR #4863, opened 2024-11-25
feat: Add mypy type checking

### LINKED ISSUE #4708: Enforcing typing
Question is whether we can use mypy and whitelist the current (vast number of) offending files

~Blocked by django 4 upgrade~

Current proposed plan: 

 1. Run mypy across the project to (hopefully) generate a whitelist
 2. Enforce typing on all new code additions (using pre-commit)
 3. Prosper

zachaysan:
## Changes

This adds mypy type checking. Existing failures have been grandfathered in and a helper script is run to compare failures against the predefined list. Mypy was also added into the `pre-commit` config and will run on any commits to python code.

I altered the predefined mypy list and saw a failure crop up when running `pre-commit`.


.pre-commit-config.yaml
```diff
@@ -40,5 +40,13 @@ repos:
       - id: poetry-check
         args: ['-C', 'api']
 
+  - repo: local
+    hooks:
+      - id: mypy-check
+        name: mypy check
+        entry: bash -c "cd api && poetry run python scripts/mypy_check.py"
```
matthewelwell: Since I was looking at similar functionality yesterday (see [here](https://github.com/Flagsmith/flagsmith/pull/4867)) , I am somewhat intrigued by this.

My understanding of pre-commit is that it passes the modified files into the command specified in `entry`, so the resulting command would look something like: 

```
bash -c "cd api && poetry run python scripts/mypy_check.py" file1.py file2.py 
```

... which I would expect to just not work at all, or at least throw some sort of error?

api/mypy.ini
matthewelwell, line 1: Unless there's a good reason not to, I think I prefer to use pyproject.toml for this instead?

https://mypy.readthedocs.io/en/stable/config_file.html#using-a-pyproject-toml-file

matthewelwell (commented):
This seems like a decent enough solution to me but I have some general questions: 

 1. What happens if we (perhaps inadvertently) resolve an existing issue. Based on the logic in your script, I guess we wouldn't know about it, right? Because we're just seeing what's left after we subtract the original errors from the new ones which wouldn't tell us anything about any that we've solved. It would obviously be great to incentivise resolving existing issues where we can. 

2. This approach works fine I guess, but another solution would have been to just add all of the distinct files into the exclude. Then every new file would need to conform to mypy, but changes to existing files wouldn't. I'm not sure how we could incentivise fixing the existing errors. (Note: I'm certainly not suggesting this is a better option, I'm just playing devil's advocate).

api/util/mappers/sdk.py
```diff
@@ -48,5 +48,5 @@ def map_environment_to_sdk_document(environment: "Environment") -> SDKDocument:
     ]
 
     return engine_environment.model_dump(
-        exclude=SDK_DOCUMENT_EXCLUDE,
+        exclude=SDK_DOCUMENT_EXCLUDE,  # type: ignore
```
matthewelwell: I'm not sure why we've added specific ignores here vs just letting the `mypy_baseline.txt` account for these errors?

.pre-commit-config.yaml
matthewelwell: @zachaysan I'm not sure the latest changes make any reference to this?

zachaysan: I've altered the script to run the files one at a time. I had to use `git diff` to list the target files of the commit because no matter what I tried to pass into the bash script it just wouldn't run with the automatic listed files. The good news is that we are now doing a `mypy` check on only the committed files.

api/mypy.ini
zachaysan: I've switched to `pyproject.toml` good call.

api/util/mappers/sdk.py
zachaysan: I was dealing with a bug. I've reverted the change and just added them to the baseline.

zachaysan:
> This seems like a decent enough solution to me but I have some general questions:
> 
> 1. What happens if we (perhaps inadvertently) resolve an existing issue. Based on the logic in your script, I guess we wouldn't know about it, right? Because we're just seeing what's left after we subtract the original errors from the new ones which wouldn't tell us anything about any that we've solved. It would obviously be great to incentivise resolving existing issues where we can.

I think we should treat the baseline as something we can devote engineering time towards. Delete a chunk of the baseline then re-run mypy see the failures that you eliminated from the baseline then handle them. Over time, I think we can get to a point where the baseline file is deleted and we have full coverage.

> 2. This approach works fine I guess, but another solution would have been to just add all of the distinct files into the exclude. Then every new file would need to conform to mypy, but changes to existing files wouldn't. I'm not sure how we could incentivise fixing the existing errors. (Note: I'm certainly not suggesting this is a better option, I'm just playing devil's advocate).

I think the approach I've taken is actually better for coverage because new errors in existing files will be surfaced. As for incentivizing fixing the existing errors, I really think we should just have engineering time to delete things from the baseline then fix them up locally.

api/scripts/mypy_check.py
```diff
@@ -0,0 +1,74 @@
...
+        sys.exit(0)
+
+    # Mypy failed in some other way other than listing failing file lines
+    if result.returncode != 1:
+        print(command)
+        print(
+            f"Error running mypy with return code {result.returncode} and error:",
+            result.stderr,
+        )
+        print(result.stdout.strip())
```
matthewelwell: These 3 print statements seem overkill? Particularly the first one that prints out the command - I suspect that is left from the debugging that you did around the pre-commit files being passed in? 👀

matthewelwell (approved):
Approving with a minor comment, but as discussed in the engineering meeting, I'd like to get everyone's eyes on this.

api/scripts/mypy_check.py
zachaysan: Yeah I was thinking that it could be useful for someone to debug the mypy error without running pre-commit but in retrospect that probably isn't necessary. I've removed the `print(command)` line. The other two should still live there though.

gagantrivedi (approved):

api/pyproject.toml
```diff
@@ -99,6 +99,16 @@ skip = ['migrations', '.venv', '.direnv']
 addopts = ['--ds=app.settings.test', '-vvvv', '-p', 'no:warnings']
 console_output_style = 'count'
 
+[tool.mypy]
+exclude = "^(tests/)$"
```
khvn26, line 103: Is there a reason to exclude the test code?

api/scripts/mypy_check.py
```diff
@@ -0,0 +1,73 @@
...
+        sys.exit(1)
+
+    current_output = result.stdout.strip().splitlines()
+
+    with open(baseline_path, "r") as f:
+        baseline = set(line.strip() for line in f if line.strip())
+
+    # Detect new errors and remove information line filtering out third party packages
+    current_errors = {line for line in current_output if "site-packages" not in line}
+    new_errors = current_errors - baseline
```
khvn26, line 50: I think we should check for `baseline - current_errors` as well to make sure the baseline file contains up to date errors. Rarely, but things may change implicitly (e.g. with a dependency update).

api/scripts/mypy_check.py
```diff
@@ -0,0 +1,73 @@
...
+        print(result.stdout.strip())
+        sys.exit(1)
+
+    current_output = result.stdout.strip().splitlines()
+
+    with open(baseline_path, "r") as f:
+        baseline = set(line.strip() for line in f if line.strip())
+
+    # Detect new errors and remove information line filtering out third party packages
+    current_errors = {line for line in current_output if "site-packages" not in line}
```
khvn26, line 49: I appreciate it's going to be slightly more difficult to compare against, but why not just have third party errors in the baseline as well?

khvn26:
[pre-commit.ci](http://pre-commit.ci/) run

khvn26:
pre-commit.ci run

.pre-commit-config.yaml
```diff
@@ -40,5 +40,13 @@ repos:
       - id: poetry-check
         args: ['-C', 'api']
 
+  - repo: local
+    hooks:
+      - id: mypy-check
+        name: mypy check
+        entry: bash -c "cd api && poetry run python scripts/mypy_check.py $(git diff --name-only --cached -- '*.py')"
```
khvn26, line 52: As seen from pre-commit.ci run [here](https://results.pre-commit.ci/run/github/136163130/1733506020.vLWIKPvMRNi862KNAN5dkA), this hook won't work when poetry isn't present globally.

Maybe try adding

```
        additional_dependencies: ['poetry']
```

or using `repo: local` (see docs [here](https://pre-commit.com/#repository-local-hooks)).

If all else fails, moving the hook to its own repo could be an option.

khvn26 (changes requested):

matthewelwell:
> pre-commit.ci run

@zachaysan I think you'll need to rebase onto the main branch to get pre-commit to pass after I merged [this PR](https://github.com/Flagsmith/flagsmith/pull/4902).

api/pyproject.toml
zachaysan: Mypy throws an error when we try to include the `tests` module in the coverage. We tried hard to figure out how to solve this bug but it was outside of our grasp.

api/scripts/mypy_check.py
zachaysan: Great idea. I've implemented this with a helpful error message.

api/scripts/mypy_check.py
zachaysan: Since the line numbers change between installs it becomes difficult to handle it that way. Plus, `site-packages` is going to be only ever present in the installation path for non-local modules.

api/scripts/mypy_check.py
zachaysan: Just realized that this doesn't actually work since `mypy` only checks errors for the current working module. I've reverted my change.

.pre-commit-config.yaml
zachaysan: The `additional_dependencies` seems to have worked once I changed the `language` to `python`.

khvn26:
Superseded by #5111

PR #4875, opened 2024-11-29
feat(environment/views): Add get-by-uuid action 

gagantrivedi:

## Changes

-    Add `uuid` to the environment model.
 - Add `get-by-uuid` action to support fetching environment using UUID

Add unit tests


api/environments/migrations/0037_add_uuid_field.py
```diff
@@ -0,0 +1,30 @@
+# Generated by Django 3.2.23 on 2024-11-27 08:46
+
+from django.db import migrations, models
+import uuid
+
+from core.migration_helpers import AddDefaultUUIDs
+
+
+class Migration(migrations.Migration):
+    atomic = False  # avoid long running transaction
```
gagantrivedi: Because we are not running all of these operations in a single transaction, there is a small probability that the last operation will fail. This could happen if the environment is created between step 2 and step 3 with the old code, which does not include the UUID.

api/features/versioning/managers.py
```diff
@@ -20,13 +20,15 @@ def get_latest_versions_by_environment_id(self, environment_id: int) -> RawQuery
         """
         return self._get_latest_versions(environment_id=environment_id)
 
-    def get_latest_versions_by_environment_api_key(
-        self, environment_api_key: str
+    def get_latest_versions(
+        self, environment_api_key: str = None, environment_uuid: str = None
```
matthewelwell: This feels a bit odd / inconsistent to me, why not just add a new `get_latest_versions_by_uuid` method?

api/environments/views.py
```diff
@@ -69,7 +71,7 @@
     ),
 )
 class EnvironmentViewSet(viewsets.ModelViewSet):
-    lookup_field = "api_key"
+    lookup_field = "api_key"  # This is deprecated please uuid
```
matthewelwell: I don't really like this approach - the documentation is going to be incorrect. Shouldn't we just add a new viewset? Can the django urls routing logic tell the difference between our API key and a uuid? It should be able to, right?

matthewelwell: In general, I'm also not sure why we've gone with this approach vs the approach on other models of adding the `get-by-uuid` route?

gagantrivedi: > In general, I'm also not sure why we've gone with this approach vs the approach on other models of adding the get-by-uuid route?

Because we will need this viewset as part of `allow rotation of client side environmen key `

matthewelwell (approved):

PR #4876, opened 2024-11-29
fix: audit log integrations for versioned environments

matthewelwell:
## Changes

Fixes audit log based integrations when running in a versioned environment. 

Updated existing tests.


api/audit/signals.py
```diff
@@ -61,6 +61,7 @@ def signal_wrapper(sender, instance, **kwargs):
             RelatedObjectType.FEATURE.name,
             RelatedObjectType.FEATURE_STATE.name,
             RelatedObjectType.SEGMENT.name,
+            RelatedObjectType.EF_VERSION.name,
```
matthewelwell, line 64: This is essentially the only thing that needed to change to fix the issue. The rest of the changes are related to the tests, improving the integration for EF versions, or refactoring.

api/tests/unit/audit/test_unit_audit_signals.py
```diff
@@ -99,15 +108,16 @@ def test_send_audit_log_event_to_grafana__project_grafana_config__calls_expected
...
-    project.grafana_config = grafana_config
     audit_log_record = AuditLog.objects.create(
         project=project,
         related_object_type=RelatedObjectType.FEATURE.name,
     )
     grafana_wrapper_mock = mocker.patch("audit.signals.GrafanaWrapper", autospec=True)
     grafana_wrapper_instance_mock = grafana_wrapper_mock.return_value
 
+    grafana_config = GrafanaProjectConfiguration(base_url="test.com", api_key="test")
+    project.grafana_config = grafana_config
```
matthewelwell, line 119: Having added the logic to prevent the postpone decorator from spawning unmanaged threads we need to move this here to prevent the generation of the audit log record itself from triggering an external request to Grafana (or Dynatrace as the case may be below).

api/tests/unit/integrations/dynatrace/test_unit_dynatrace.py
```diff
@@ -44,22 +51,33 @@ def test_dynatrace_initialized_correctly():
...
+        environment=getattr(related_object, "environment", None),
+        project=project,
         related_object_type=related_object_type,
         related_object_id=related_object.id,
     )
 
+    mocker.patch(
+        "integrations.dynatrace.dynatrace.get_audited_instance_from_audit_log_record",
+        return_value=related_object,
+    )
```
matthewelwell, line 79: I don't love having to introduce a mock here, but this test is all a little bit superficial anyway. I'd rather replace all of these parameterised tests with more end-to-end style tests like the new one I added for EF version [here](https://github.com/Flagsmith/flagsmith/pull/4876/files#diff-9326d6ea8a6975db03e80a54706f25bcf90f6f49557c20fa54e5e60733100589R264).

api/util/util.py
```diff
@@ -3,14 +3,19 @@
...
+
 T = TypeVar("T")
 
 
 def postpone(function):
     def decorator(*args, **kwargs):
-        t = Thread(target=function, args=args, kwargs=kwargs)
-        t.daemon = True
-        t.start()
+        if settings.ENABLE_POSTPONE_DECORATOR:
```
matthewelwell: This is somewhat ugly, and I'd rather remove the `postpone` decorator entirely to rely instead upon the `run_in_thread` method exposed by registered tasks. But, I'd still then want to be able to disable the threading in a test environment because it's all a bit difficult to test anyway (and it's how we sometimes end up with errors in the logs, but not in the tests).

khvn26 (approved):

PR #4883, opened 2024-12-03
chore: Send hubspot token to register endpoints

### LINKED ISSUE #4671: Add Hubspot tracking code to other domains
Based on an investigation into Hubspot not properly tracking usage I believe we need to add the below to our primary domains (definitely [app.flagsmith.com](http://app.flagsmith.com/) and [docs.flagsmith.com](http://docs.flagsmith.com/), but check if we have other subdomains.) The script should be running on every page for flagsmith.com
```html
<!-- Start of HubSpot Embed Code -->
<script type="text/javascript" id="hs-script-loader" async defer src="//js-eu1.hs-scripts.com/143451822.js"></script>
<!-- End of HubSpot Embed Code -->
```

kyle-ssg:

## Changes

- Attempts to read hubspotutk and send it to create organisation and join endpoints

staging.flagsmith.com has this deployed

image

### DIFF
--- a/frontend/common/stores/account-store.js
+++ b/frontend/common/stores/account-store.js
@@ -14,7 +14,9 @@ const controller = {
     API.setInvite('')
     API.setInviteType('')
     return data
-      .post(`${Project.api}users/join/link/${id}/`)
+      .post(`${Project.api}users/join/link/${id}/`, {
+        hubspotutk: API.getCookie('hubspotutk'),
+      })
       .catch((error) => {
         if (
           Utils.getFlagsmithHasFeature('verify_seats_limit_for_invite_links') &&
@@ -23,7 +25,9 @@ const controller = {
           API.ajaxHandler(store, error)
           return
         }
-        return data.post(`${Project.api}users/join/${id}/`)
+        return data.post(`${Project.api}users/join/${id}/`, {
+          hubspotutk: API.getCookie('hubspotutk'),
+        })
       })
       .then((res) => {
         store.savedId = res.id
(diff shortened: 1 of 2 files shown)


frontend/common/stores/account-store.js
```diff
@@ -85,24 +89,29 @@ const controller = {
...
-        { ...res, role: 'ADMIN' },
-      ])
-      AsyncStorage.setItem('user', JSON.stringify(store.model))
-      store.savedId = res.id
-      store.saved()
+    data
+      .post(`${Project.api}organisations/`, {
+        hubspotutk: API.getCookie('hubspotutk'),
+        name,
+      })
```
matthewelwell, line 96: This makes sense I guess because we're triggering the call to hubspot when a user joins an organisation, which is obviously done implicitly when they create an organisation. I'm not sure the backend is set up to handle this scenario though @zachaysan ?

zachaysan: I'm not sure I fully understand what you're saying. Is the user present in this request? (I assume it is.) If so, passing the `hubspotutk` value up will optionally trigger the `HubspotTracker.objects.update_or_create` if the `hubspotutk` value is present. So unless I'm misunderstanding something this code should work fine.

frontend/common/stores/account-store.js
```diff
@@ -85,24 +89,29 @@ const controller = {
...
-      if (sendRelayEvent) {
-        dataRelay.sendEvent(AccountStore.getUser(), {
-          apiKey: relayEventKey,
-        })
-      }
-    })
+        const relayEventKey = Utils.getFlagsmithValue('relay_events_key')
+        const sendRelayEvent =
+          Utils.getFlagsmithHasFeature('relay_events_key') && !!relayEventKey
+        window.lintrk?.('track', { conversion_id: 16798338 })
```
zachaysan: What is `conversion_id` here? If it's a magic number we should assign it to a meaningful variable name like `google_account_tracking_id = 16798338` before using it in the track action.

zachaysan (approved):
Approving because the code looks good to me, but there are a couple of small comments to be addressed before merging.

frontend/common/stores/account-store.js
matthewelwell: Sorry, yep, just looked at the API code - all looks good to me.

frontend/common/stores/account-store.js
kyle-ssg: Yep good point, done

PR #4904, opened 2024-12-09
docs(android): remove jitpack from kotlin docs

matthewelwell:
## Changes

Removes jitpack from the docs in favour of maven. 

Note - relies on the following PR (and subsequent release)

https://github.com/Flagsmith/flagsmith-kotlin-android-client/pull/61

The docs build will continue to fail until that is completed

n/a

### DIFF
--- a/docs/docs/clients/client-side/android.md
+++ b/docs/docs/clients/client-side/android.md
@@ -25,12 +25,6 @@ In your project path `app/build.gradle` add a new dependency:
 
 <CodeBlock>{`implementation("com.flagsmith:flagsmith-kotlin-android-client:`}<AndroidVersion />"{`)`}</CodeBlock>
 
-:::info
-
-Note: jitpack support has been deprecated as of `v1.7.1`.
-
-:::
-
 ## Basic Usage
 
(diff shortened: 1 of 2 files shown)


docs/docs/clients/client-side/android.md
```diff
@@ -14,20 +14,23 @@ This SDK can be used for Android applications written in Kotlin. The source code
...
 }
 ```
 
 In your project path `app/build.gradle` add a new dependency:
 
 <CodeBlock>{`implementation("com.flagsmith:flagsmith-kotlin-android-client:`}<AndroidVersion />"{`)`}</CodeBlock>
 
+:::info
+
+Note: jitpack support has been deprecated as of `v1.7.1`.
```
rolodato: I think we can leave this warning out. It's not relevant to new users, and new users don't need to do anything in response to it. Maybe they could remove JitPack as a repository if they're not using it anywhere else in their project, but it's not important either way.

rolodato (approved):

docs/docs/clients/client-side/android.md
matthewelwell: Done

PR #4905, opened 2024-12-09
fix: Relax Hubspot cookie tracking

### LINKED ISSUE #4900: Hubspot cookie data handled incorrectly for user that already has one
Sentry Issue: [FLAGSMITH-API-51F](https://flagsmith.sentry.io/issues/6121478470/?referrer=github_integration)

```
HubspotTracker.DoesNotExist: HubspotTracker matching query does not exist.
  File "django/db/models/query.py", line 916, in get_or_create
    return self.get(**kwargs), False
  File "django/db/models/query.py", line 637, in get
    raise self.model.DoesNotExist(

UniqueViolation: duplicate key value violates unique constraint "users_hubspottracker_hubspot_cookie_key"
DETAIL:  Key (hubspot_cookie)=(d4b2f190eb2d2072670294dda413d8e2) already exists.

  File "django/db/backends/utils.py", line 89, in _execute
    return self.cursor.execute(sql, params)

IntegrityError: duplicate key value violates unique constraint "users_hubspottracker_hubspot_cookie_key"
DETAIL:  Key (hubspot_cookie)=(d4b2f190eb2d2072670294dda413d8e2) already exists.

(21 additional frame(s) were not displayed)
...
  File "organisations/views.py", line 116, in create
    register_hubspot_tracker(request)
  File "integrations/lead_tracking/hubspot/services.py", line 24, in register_hubspot_tracker
    HubspotTracker.objects.update_or_create(
```

zachaysan:
## Changes

Fixes https://github.com/Flagsmith/flagsmith/issues/4900

Since Hubspot cookies are used to merge identities, it is preferable to ignore a Hubspot token for different users on the same machine.

Proved that the error was raised without it then wrote a test to verify that multiple users can supply the same Hubspot cookie without an error being raised.


api/integrations/lead_tracking/hubspot/services.py
```diff
@@ -17,13 +18,18 @@ def register_hubspot_tracker(request: Request) -> None:
...
+            },
+        )
+        logger.info(
+            f"Created HubspotTracker instance for user {request.user.email} with cookie {hubspot_cookie}"
+        )
+    except IntegrityError:
+        logger.info(
+            f"HubspotTracker could not be created for user {request.user.email}"
+            f" due to cookie conflict with cookie {hubspot_cookie}"
+        )
```
matthewelwell: I'm not sure if I'm missing something here as to why we're using `update_or_create` here, but I don't love the implementation - mostly because `IntegrityError` might not always indicate what we're assuming here. 

I think we should just be more explicit, and do something like: 

```python
if HubspotTracker.objects.filter(hubspot_cookie=hubspot_cookie).exists():
    logger.info(
        f"HubspotTracker could not be created for user {request.user.email}"
        f" due to cookie conflict with cookie {hubspot_cookie}"
    )
else:
    HubspotTracker.objects.create(...)
    logger.info(...)
```

zachaysan: Ok I've switched to a filter and exclude for the check. I'm keeping the `update_or_create` here because if a users hubspot cookie is changed we want to include that change in their tracking data.

api/integrations/lead_tracking/hubspot/services.py
```diff
@@ -17,13 +17,23 @@ def register_hubspot_tracker(request: Request) -> None:
...
-    logger.info(
-        f"Creating HubspotTracker instance for user {request.user.email} with cookie {hubspot_cookie}"
-    )
+    if (
+        HubspotTracker.objects.filter(hubspot_cookie=hubspot_cookie)
+        .exclude(user=request.user)
+        .exists()
+    ):
+        logger.info(
+            f"HubspotTracker could not be created for user {request.user.email}"
```
matthewelwell: Sorry, I'd actually rather that we just log the user id here (and below) - I don't like the idea of logging emails if we can avoid it.

zachaysan: Ok that's no problem. I've switched the loggers to use user ids.

matthewelwell (approved):

PR #4907, opened 2024-12-10
fix: Restore `make docker-up`

khvn26:

## Changes

Restores the `docker-up` Make target removed in #4747. Adds conventional file name and a bit of documentation.

Ran `make docker-up test` to verify the local Postgres db is created correctly.

### DIFF
--- a/docker/api/docker-compose.local.yml
+++ b/docker/api/docker-compose.local.yml
@@ -1,0 +1,19 @@
+# A Compose file with minimal dependencies to be able to run Flagsmith, including its test suite, locally.
+
+name: flagsmith
+
+volumes:
+  pg_11_data:
+
+services:
+  db:
+    image: postgres:15.5-alpine
+    pull_policy: always
+    restart: unless-stopped
+    volumes:
+      - pg_11_data:/var/lib/postgresql/data
+    ports:
+      - 5432:5432
+    environment:
+      POSTGRES_DB: flagsmith
+      POSTGRES_PASSWORD: password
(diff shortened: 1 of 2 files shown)


docker/api/docker-compose.local.yml
```diff
@@ -0,0 +1,19 @@
+# A Compose file with minimal dependencies to be able to run Flagsmith, including its test suite, locally.
+
+name: flagsmith
```
matthewelwell: This comment, and name seem a little unrelated to the actual content of the docker compose file?

khvn26: How so? The goal is to be able to run e.g. `python manage.py runserver` or `pytest` with environment from `.env-local`, which this Compose file allows.

The name is kept from previous `db.yaml` so the volumes are carried over.

matthewelwell: But this docker compose file only provides the database component. The comment, and the name (to me) indicate that there should be more services configured here. 

It's not really a huge deal, but it is somewhat confusing to me.

matthewelwell (approved):

docker/api/docker-compose.local.yml
khvn26: > only provides the database component

Which is the only dependency required to run Flagsmith locally, _outside_ of Docker.

matthewelwell: > _outside_ of Docker

That piece is not clear though. It just states "to run Flagsmith, [...], locally.". 'locally' doesn't imply outside of docker.

khvn26: Agreed. I'll add a comment to clarify.

khvn26: Done.

PR #4908, opened 2024-12-10
feat: Heal all identities with blank traits

khvn26:

## Changes

Adds a management command that scans through all identities and restores previously deleted blank traits.

Additionally, updates local pre-commit version to play nice with recently updated hooks.

Additionally, adds `structlog` and logs the command progress with structured events.

The code is covered by a new unit test. The test ensures only the fixed identities get written.

### DIFF
--- a/api/edge_api/management/commands/ensure_identity_traits_blanks.py
+++ b/api/edge_api/management/commands/ensure_identity_traits_blanks.py
@@ -1,0 +1,28 @@
+import logging
+from typing import Any
+
+from django.core.management import BaseCommand
+
+from environments.dynamodb import DynamoIdentityWrapper
+
+identity_wrapper = DynamoIdentityWrapper()
+
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    def handle(self, *args: Any, **options: Any) -> None:
+        for identity_document in identity_wrapper.query_get_all_items():
+            should_write_identity_document = False
+            if identity_traits_data := identity_document.get("identity_traits"):
+                for trait_data in identity_traits_data:
+                    if "trait_value" not in trait_data:
+                        should_write_identity_document = True
+                        trait_data["trait_value"] = ""
+            if should_write_identity_document:
+                identity_wrapper.put_item(identity_document)
+                self.stdout.write(
+                    f"Fixed identity id={identity_document['identity_uuid']}",
+                )
+        self.stdout.write(self.style.SUCCESS("Finished."))
--- a/api/edge_api/management/commands/ensure_identity_traits_blanks.py
+++ b/api/edge_api/management/commands/ensure_identity_traits_blanks.py
@@ -48,4 +48,11 @@ class Command(BaseCommand):
                     f"scanned={scanned_count}/{total_count}"
                     f"percentage={scanned_count/total_count*100:.2f}"
                 )
-        self.stdout.write(self.style.SUCCESS("Finished."))
+
+        self.stdout.write(
+            self.style.SUCCESS(
+                "finished"
+                f"scanned={scanned_count}/{total_count}"
+                f"percentage={scanned_count/total_count*100:.2f}"
+            )
+        )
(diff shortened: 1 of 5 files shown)


api/edge_api/management/commands/ensure_identity_traits_blanks.py
```diff
@@ -0,0 +1,28 @@
...
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    def handle(self, *args: Any, **options: Any) -> None:
+        for identity_document in identity_wrapper.query_get_all_items():
+            should_write_identity_document = False
+            if identity_traits_data := identity_document.get("identity_traits"):
+                for trait_data in identity_traits_data:
+                    if "trait_value" not in trait_data:
```
matthewelwell: This is just personal preference but I'd rather see a `filter()` statement here. 

e.g. 

```python
for trait_data in filter(lambda td: "trait_value" not in td, identity_traits_data):
```

api/edge_api/management/commands/ensure_identity_traits_blanks.py
```diff
@@ -0,0 +1,28 @@
...
+                        should_write_identity_document = True
+                        trait_data["trait_value"] = ""
+            if should_write_identity_document:
+                identity_wrapper.put_item(identity_document)
+                self.stdout.write(
+                    f"Fixed identity id={identity_document['identity_uuid']}",
+                )
```
matthewelwell: Maybe we should add some more log messages like this one to show some indication of progress?

I guess we don't want to write a message for every identity we iterate over, but perhaps we should add a counter, and log a message every 100k or something?

matthewelwell (approved):
Approved with a couple of minor comments. 

We should also remove the unrelated code (from the other PR [here](https://github.com/Flagsmith/flagsmith/pull/4907))

api/edge_api/management/commands/ensure_identity_traits_blanks.py
khvn26: Honestly, out of the three options

1. 
```python
for trait_data in filter(lambda td: "trait_value" not in td, identity_traits_data):
```

2. 
```python
for trait_data in (td in identity_traits_data if "trait_value" not in td):
```

3. 
```python
for trait_data in identity_traits_data:
    if "trait_value" not in trait_data:
```

the current one feels the most readable to me.

api/edge_api/management/commands/ensure_identity_traits_blanks.py
khvn26: If we had standardised stats export like Prometheus, sure.

At this point in time I prefer not to do this.

Besides, I don't expect that many corrupted identities.

api/edge_api/management/commands/ensure_identity_traits_blanks.py
matthewelwell: I certainly see your point of view, I just really dislike multiple levels of nested loops / conditionals - but I certainly don't feel strongly enough about it to disagree with you here.

api/edge_api/management/commands/ensure_identity_traits_blanks.py
matthewelwell: > Besides, I don't expect that many corrupted identities.

This is exactly the reason that I think we _should_ do this. Because it might be the case that, when it's running, it looks like nothing is happening.

api/edge_api/management/commands/ensure_identity_traits_blanks.py
khvn26: Okay, added some additional counting logging.

api/edge_api/management/commands/ensure_identity_traits_blanks.py
khvn26: Okay, added some additional counting logging.

api/edge_api/management/commands/ensure_identity_traits_blanks.py
khvn26: Fair enough, removed the extra nesting level.

matthewelwell (approved):

zachaysan (approved):
Looks good though you're missing some code coverage for a small section of the command.

api/edge_api/management/commands/ensure_identity_traits_blanks.py
khvn26: NB: I've decided to re-implement it using structlog.

PR #4917, opened 2024-12-12
fix: description state management in EnvironmentSettingsPage

tiagoapolo:

## Changes

- Checks for roles empty list
- Fix description field state management, conflicting value from what comes from the backend and what is stored in the component's state

Before

https://github.com/user-attachments/assets/d7265d85-de52-4644-98df-84712ddcf8ec

After

https://github.com/user-attachments/assets/faf59e6c-3521-48e0-80d6-0a845a704e5a

_Please describe._

### DIFF
--- a/frontend/web/components/pages/EnvironmentSettingsPage.js
+++ b/frontend/web/components/pages/EnvironmentSettingsPage.js
@@ -67,6 +67,8 @@ const EnvironmentSettingsPage = class extends Component {
       { organisation_id: AccountStore.getOrganisation().id },
       { forceRefetch: true },
     ).then((roles) => {
+      if (!roles?.data?.results?.length) return
+
       getRoleEnvironmentPermissions(
         getStore(),
         {
@@ -76,7 +78,6 @@ const EnvironmentSettingsPage = class extends Component {
         },
         { forceRefetch: true },
       ).then((res) => {
-        debugger
         const matchingItems = roles.data.results.filter((item1) =>
           res.data.results.some((item2) => item2.role === item1.id),
         )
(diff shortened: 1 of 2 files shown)


frontend/web/components/pages/EnvironmentSettingsPage.js
```diff
@@ -61,7 +61,7 @@ const EnvironmentSettingsPage = class extends Component {
     const env = ProjectStore.getEnvs().find(
       (v) => v.api_key === this.props.match.params.environmentId,
     )
-    this.setState({ env })
+    this.setState({ env, description: env.description ?? "" })
```
matthewelwell: I'm interested to know why `description` has to be handled differently? Shouldn't it just be updated as part of the environment object? 🤔

zachaysan (approved):

frontend/web/components/pages/EnvironmentSettingsPage.js
tiagoapolo: My approach here was to make sure we're consistent in how we're getting and updating the data from, previously we were mixing internal states with context one, which in my opinion is more confusing and prone to bugs. So, either using environment object or description internal states is fine.
```
  value={
    typeof this.state.description === 'string'
      ? this.state.description
      : env.description
  }
```

matthewelwell (approved):
This looks good to me now. 

The only think that I noticed which might be relevant to think about (in a separate issue) is the fact that there is no warning that any in progress edits will be lost when switching between environments. 

So, for example, if you're editing the description of your development environment, then switch to your production environment to check something, your edits from the development environment will be lost. I guess the slightly interesting quirk of the previous (buggy) behaviour is that this wasn't an issue! 

@tiagoapolo can you create a separate issue to tackle this? I think we should just add a warning when you navigate away if the environment settings haven't been saved.

PR #4928, opened 2024-12-18
docs: Document HTTP proxy configuration for Edge Proxy

rolodato:
See https://github.com/Flagsmith/edge-proxy/pull/134

Changes can be seen at https://docs-git-docs-edge-proxy-http-proxy-flagsmith.vercel.app/deployment/hosting/locally-edge-proxy#environment-variables

Confirmed with:

```
% cat tinyproxy.conf
Port 8889
Timeout 600
Listen 0.0.0.0
```
```
% tinyproxy -d -c tinyproxy.conf
```

```
% docker run \
    -e ALL_PROXY="host.docker.internal:8889" \
    -p 8000:8000 \
    flagsmith/edge-proxy:latest
```

Result:

```
CONNECT   Dec 18 13:22:25.809 [22667]: Connect (file descriptor 4): 127.0.0.1
CONNECT   Dec 18 13:22:25.809 [22667]: Request (file descriptor 4): CONNECT edge.api.flagsmith.com:443 HTTP/1.1
INFO      Dec 18 13:22:25.809 [22667]: No upstream proxy for edge.api.flagsmith.com
INFO      Dec 18 13:22:25.809 [22667]: opensock: opening connection to edge.api.flagsmith.com:443
INFO      Dec 18 13:22:25.817 [22667]: opensock: getaddrinfo returned for edge.api.flagsmith.com:443
CONNECT   Dec 18 13:22:25.823 [22667]: Established connection to host "edge.api.flagsmith.com" using file descriptor 8.
INFO      Dec 18 13:22:25.823 [22667]: Not sending client headers to remote machine
INFO      Dec 18 13:22:27.485 [22667]: Closed connection between local client (fd:4) and remote client (fd:8)
```

PR #4932, opened 2024-12-19
ci: fix prettier ci issues

matthewelwell:
## Changes

 * Add ignore records for python and image files. 
 * Move remaining exclusions from pre-commit to prettier config

I have run `prettier run --all-files` locally and I have also tested these changes against the changes in #4928 (by cherry picking them across to this branch, then reverting) which seem to be the ones bringing out the issue we're trying to solve.


.prettierignore
```diff
@@ -1,10 +1,13 @@
...
-*.json
-*.handlebars
-*.css
-.bablerc
-**/CHANGELOG.md
+# Ignore everything by default
+**
+
+# Include only specific extensions
+!*.{md,mdx,js,jsx,ts,tsx,json,css,html}
```
matthewelwell: I see that we're technically adding html, css and json by making this change, but since prettier is passing, I can't see any reason to keep them ignored.

.pre-commit-config.yaml
```diff
@@ -36,7 +36,7 @@ repos:
           - prettier@3.4.2
         args:
           - --check
-        exclude: ^(frontend/|CHANGELOG.md|.github/docker_build_comment_template.md)
+        types: [javascript,jsx,ts,tsx,markdown,mdx,html,css,json,yaml]
```
khvn26, line 39: Nice!

khvn26 (approved):

Issue #4859, opened 2024-11-21
Updating a segment to add additional rules does not update the users tab of the segment


zachaysan:
This can be seen from users tab below:

![Image](https://github.com/user-attachments/assets/e46d0cc8-1b54-4906-8f8f-15255c40f421)

Lucifer4255:
@matthewelwell Can you tell me how to reproduce the issue?

matthewelwell:
Hi @Lucifer4255 apologies for not getting back to you. We're going to work on this item internally to help us on board a new engineer on the team.

PR #4950, opened 2025-01-03
fix: Updates user segment tab when a segment is created/updated

tiagoapolo:

## Changes

Ref: [#4859](https://github.com/Flagsmith/flagsmith/issues/4859)

* Replaced the inline form and user tab content with the new `CreateSegmentRulesTabForm` and `CreateSegmentUsersTabContent` components. 
* Adds refresh button to refetch users list
* Updates user identity list after segment update/creation

1. In segments page create/update a segment rule adding a user to that segment
2. Go to users tab and make sure an environment is selected
3. Check if user/identity is in that segment
4. If not try clicking refresh and verify if user is now in that segment

Screenshot 2025-01-08 at 08 51 31

### DIFF
--- a/frontend/web/components/modals/CreateSegment.tsx
+++ b/frontend/web/components/modals/CreateSegment.tsx
@@ -240,6 +240,13 @@ const CreateSegment: FC<CreateSegmentType> = ({
     }
   }
 
+  const fetchUserIdentityList = () => {
+    if (!environmentId) return
+    identities?.results.forEach((identity) =>
+      AppActions.getIdentitySegments(projectId, identity.id),
+    )
+  }
+
   const [valueChanged, setValueChanged] = useState(false)
   const [metadataValueChanged, setMetadataValueChanged] = useState(false)
   const onClosing = useCallback(() => {
@@ -281,6 +288,7 @@ const CreateSegment: FC<CreateSegmentType> = ({
     if (createSuccess && createSegmentData) {
       setSegment(createSegmentData)
       onComplete?.(createSegmentData)
+      fetchUserIdentityList()
     }
     //eslint-disable-next-line
   }, [createSuccess])
@@ -288,17 +296,10 @@ const CreateSegment: FC<CreateSegmentType> = ({
     if (updateSuccess && updateSegmentData) {
       setSegment(updateSegmentData)
       onComplete?.(updateSegmentData)
+      fetchUserIdentityList()
     }
     //eslint-disable-next-line
   }, [updateSuccess])
-
-  useEffect(() => {
-    if (tab === UserTabs.USERS && environmentId) {
-      identities?.results.forEach((identity) =>
-        AppActions.getIdentitySegments(projectId, identity.id),
-      )
-    }
-  }, [tab, identities, projectId, environmentId])
 
   const operators: Operator[] | null = _operators || Utils.getSegmentOperators()
   if (operators) {
--- a/frontend/web/components/modals/CreateSegment.tsx
+++ b/frontend/web/components/modals/CreateSegment.tsx
@@ -52,6 +52,7 @@ import AddMetadataToEntity, {
 } from 'components/metadata/AddMetadataToEntity'
 import { useGetSupportedContentTypeQuery } from 'common/services/useSupportedContentType'
 import { setInterceptClose } from './base/ModalDefault'
+import AppActions from 'common/dispatcher/app-actions'
 
 type PageType = {
   number: number
@@ -127,6 +128,7 @@ const CreateSegment: FC<CreateSegmentType> = ({
   const [description, setDescription] = useState(segment.description)
   const [name, setName] = useState<Segment['name']>(segment.name)
   const [rules, setRules] = useState<Segment['rules']>(segment.rules)
+  const [shouldUpdateIdentities, setShouldUpdateIdentities] = useState(false)
   useEffect(() => {
     if (segment) {
       setRules(segment.rules)
@@ -282,6 +284,7 @@ const CreateSegment: FC<CreateSegmentType> = ({
     if (createSuccess && createSegmentData) {
       setSegment(createSegmentData)
       onComplete?.(createSegmentData)
+      setShouldUpdateIdentities(true)
     }
     //eslint-disable-next-line
   }, [createSuccess])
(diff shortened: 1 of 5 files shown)


frontend/web/components/modals/CreateSegment.tsx
```diff
@@ -282,16 +276,33 @@ const CreateSegment: FC<CreateSegmentType> = ({
     if (createSuccess && createSegmentData) {
       setSegment(createSegmentData)
       onComplete?.(createSegmentData)
+      setShouldUpdateIdentities(true)
```
tiagoapolo: Adding this logic was necessary due to a data propagation delay. After successfully updating/creating a segment it takes hundreds of milliseconds for `/segments/?identity={ID}` to return identities with the latest segment changes.

matthewelwell: This is likely due to the time it takes to update the environment document in DynamoDB (which is handled by the task processor). In normal circumstances, this should be less than 1 second, yep.

frontend/web/components/modals/CreateSegment.tsx
```diff
@@ -282,16 +276,33 @@ const CreateSegment: FC<CreateSegmentType> = ({
...
     }
     //eslint-disable-next-line
   }, [updateSuccess])
+
+  useEffect(() => {
+    if (
+      shouldUpdateIdentities &&
+      tab === 2 &&
+      environmentId &&
+      environmentId !== 'ENVIRONMENT_API_KEY'
```
matthewelwell: I'm interested to know why we have to cater for this case. I've certainly seen in the network console that there are often failed requests made to `/api/v1/environments/ENVIRONMENT_API_KEY/whatever/` but shouldn't we fix the cause, rather than accounting for it here?

frontend/web/components/modals/CreateSegment.tsx
matthewelwell: I have a feeling maybe this isn't completely bulletproof? I was able to get the tabs into a state out of sync by flicking between the two tabs a few times. Normally I'd say 'well, that's not really usual user behaviour', but in the case, it does feel like a valid use case. 

https://www.loom.com/share/764bfa8d16254fb094711943980589fc?sid=17b1d5dc-42a6-4697-84d4-40e895187ce1

tiagoapolo: I guess in this case the easy approach is fetching `/segments/?identity={ID}` every time the user tab is rendered, being more intuitive for the user to understand that they can "refresh" the status by switching tabs

matthewelwell: Yeah, that makes sense.

frontend/web/components/modals/CreateSegment.tsx
```diff
@@ -80,7 +75,14 @@ type CreateSegmentType = {
   segment?: Segment
 }
 
+enum UserTabs {
+  RULES = 0,
+  FEATURES = 1,
+  USERS = 2,
+}
```
matthewelwell, line 82: ❤️

matthewelwell (commented):
Hmm... @tiagoapolo this still doesn't seem to work correctly. It seems like it doesn't update each time, only after you update the segment twice? See recording here:

https://github.com/user-attachments/assets/4d6fd18b-a36d-41c4-a799-c133815f39c2

matthewelwell:
As discussed on the standup with @tiagoapolo, this is actually caused by the fact that the logic to get the segments for a given identity uses the environment document, which is updated asynchronously by the task processor. See the code [here](https://github.com/Flagsmith/flagsmith/blob/main/api/segments/views.py#L71), and subsequently [here](https://github.com/Flagsmith/flagsmith/blob/main/api/environments/dynamodb/wrappers/identity_wrapper.py#L187-L190). 

Thus, the issue raised by myself above actually requires an improvement to be implemented in the API. Some options for this are: 

 1. Allow certain endpoints to rebuild the environment document synchronously
 2. Add a new endpoint to rebuild the environment document synchronously
 3. Update the logic linked above to use ORM models, mapped accordingly and passed into the engine

My preference would likely be (3) but will probably require changes to the engine as well. 

Another, simpler issue would be to add a 'reload' option to the 'Users' tab so that users can manually reload the list which would allow them to reload after making a change to the segment. What do you think about this approach @tiagoapolo ?

tiagoapolo:
> As discussed on the standup with @tiagoapolo, this is actually caused by the fact that the logic to get the segments for a given identity uses the environment document, which is updated asynchronously by the task processor. See the code [here](https://github.com/Flagsmith/flagsmith/blob/main/api/segments/views.py#L71), and subsequently [here](https://github.com/Flagsmith/flagsmith/blob/main/api/environments/dynamodb/wrappers/identity_wrapper.py#L187-L190).
> 
> Thus, the issue raised by myself above actually requires an improvement to be implemented in the API. Some options for this are:
> 
> 1. Allow certain endpoints to rebuild the environment document synchronously
> 2. Add a new endpoint to rebuild the environment document synchronously
> 3. Update the logic linked above to use ORM models, mapped accordingly and passed into the engine
> 
> My preference would likely be (3) but will probably require changes to the engine as well.
> 
> Another, simpler issue would be to add a 'reload' option to the 'Users' tab so that users can manually reload the list which would allow them to reload after making a change to the segment. What do you think about this approach @tiagoapolo ?

Makes sense! Also by having a refresh btn we don't need to refetch the list when the tab is loaded.

frontend/web/components/modals/CreateSegmentUsersTabContent.tsx
```diff
@@ -0,0 +1,157 @@
...
+import Utils from 'common/utils/utils'
+import { Segment } from 'common/types/responses'
+import Icon from 'components/Icon'
+
+interface CreateSegmentUsersTabContentProps {
+  projectId: string | number
+  environmentId: string
+  setEnvironmentId: (environmentId: string) => void
+  identitiesLoading: boolean
+  identities: any
```
kyle-ssg: This can be Res['identities'] rather than any

frontend/web/components/modals/CreateSegmentUsersTabContent.tsx
```diff
@@ -0,0 +1,157 @@
...
+  page: any
+  setPage: (page: any) => void
```
kyle-ssg, line 19: The way the CreatesSegment is paging seems very strange rather than it just being a number. Probably one to fix as a future PR - may as well leave as any for now.

matthewelwell (approved):
Looks good to me now

PR #4952, opened 2025-01-03
fix: handle fetch github stars errors

tiagoapolo:

## Changes

Ref: [#4945](https://github.com/Flagsmith/flagsmith/issues/4945)

- Don't show github star count when gh api fails
- bypass error by adding a `.catch` handler

1. Access flagsmith dashboard, on the top right, stars count should display
2. Access flagsmith dashboard, on the top right, if gh api fails then stars count will not display 

### Success
Screenshot 2025-01-03 at 17 03 03

### GitHub API Failure
Screenshot 2025-01-03 at 17 03 21

### DIFF
--- a/frontend/web/components/GithubStar.tsx
+++ b/frontend/web/components/GithubStar.tsx
@@ -21,6 +21,9 @@ const GithubStar: FC<GithubStarType> = ({}) => {
         .then(function (res) {
           setStars(res.stargazers_count)
         })
+        .catch(() => {
+          setStars(undefined)
+        })
     }
   }, [planName])
 
@@ -28,8 +31,8 @@ const GithubStar: FC<GithubStarType> = ({}) => {
     return <></>
   }
 
-  return (
-    <>
+  if (!stars) {
+    return (
       <a
         style={{ width: 90 }}
         target='_blank'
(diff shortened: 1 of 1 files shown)


frontend/web/components/GithubStar.tsx
```diff
@@ -21,28 +21,41 @@ const GithubStar: FC<GithubStarType> = ({}) => {
...
+      target='_blank'
+      href='https://github.com/flagsmith/flagsmith'
+      className='btn btn-sm btn-with-icon text-body'
+      rel='noreferrer'
+    >
+      <div className='d-flex flex-row justify-content-center align-items-center'>
+        <IonIcon style={{ fontSize: 16 }} icon={logoGithub} />
+        <div className='ms-1'>{stars}</div>
+      </div>
+    </a>
```
matthewelwell: Can we make this more DRY? I'm sure it's more complicated than this, but would something _like_ the following work? 

```javascript
  return (
    <a
      style={{ width: 90 }}
      target='_blank'
      href='https://github.com/flagsmith/flagsmith'
      className='btn btn-sm btn-with-icon text-body'
      rel='noreferrer'
    >
      <div className='d-flex flex-row justify-content-center align-items-center'>
        <IonIcon style={{ fontSize: 16 }} icon={logoGithub} />
        { !stars ? <div className='ms-1'>{stars}</div> : '' }
      </div>
    </a>
  )
```

tiagoapolo: Normally I prefer the early return because it's easier to read but in this case I agree we can make it less repetitive and smaller

matthewelwell (approved):

PR #4977, opened 2025-01-08
ci: fix frontend deploy workflow

matthewelwell:
## Changes

Fixes frontend deployment workflow, and refactors for code reuse. 

Triggered the workflow (without actually deploying) for the new branch (see [here](https://github.com/Flagsmith/flagsmith/actions/runs/12668896418/job/35305234168))

### DIFF
--- a/.github/workflows/.reusable-frontend-deploy.yml
+++ b/.github/workflows/.reusable-frontend-deploy.yml
@@ -7,6 +7,10 @@ on:
         type: string
         description: Github environment to use for secret / variable access
         required: true
+      npm_build_env:
+        type: string
+        description: The environment set as the ENV environment variable when running npm run env
+        required: true
 
 defaults:
   run:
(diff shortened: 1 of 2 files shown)


.github/workflows/.reusable-frontend-deploy.yml
```diff
@@ -0,0 +1,50 @@
...
+        with:
+          path: ~/.npm
+          key: npm-${{ hashFiles('package-lock.json') }}
+          restore-keys: npm-
+
+      - name: Deploy to Vercel
+        env:
+          VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
+          VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
+          ENV: prod
```
kyle-ssg, line 43: I think it might be worth signifying this is the production workflow in the filename if this isn't an input.

kyle-ssg: I think it might be worth mentioning this is a production workflow in the filename if this isn't an input

matthewelwell: Yeah, fair point. I've added an input for this as I hope that we can use this for staging in the future rather than having different workflows for staging / production and demo.

kyle-ssg (approved):

PR #4983, opened 2025-01-09
fix(edge-migrator): add call to migrate to v2

matthewelwell:

## Changes

Creating a draft PR for now based on [this Crisp chat](https://app.crisp.chat/website/8857f89e-0eb5-4263-ab49-a293872b6c19/inbox/session_aab6b104-bda7-4f26-aec2-84fdb5ce5427/). This needs more work, but this is a start. 

_Please describe._


api/environments/dynamodb/migrator.py
```diff
@@ -81,4 +82,5 @@ def migrate(self):
             )
         )
         identity_wrapper.write_identities(iterator_with_prefetch(identities))
+        migrate_project_environments_to_v2(project_id)
```
matthewelwell, line 127: This will need more work since it won't work unless we set the read capacity budget against the project. We should therefore just write the identity overrides directly to the `flagsmith_environments_v2` table but this requires a larger refactor of the logic.

matthewelwell:
Not worth investing time in right now since there are very few active customers that still need to migrate to the Edge API. We can handle this additional step manually in the meantime which I have documented [here](https://www.notion.so/flagsmith/Support-Engineer-Runbook-2fba2b634bf98073894cfb4ab1ef00aa?source=copy_link#33aa2b634bf980ccbd63e1caab9e123b).

PR #4984, opened 2025-01-09
fix: Remove is_enterprise check

zachaysan:
## Changes

This enables allowing licences without having to check the `is_enterprise` function.

Simple change. Did not test.

### DIFF
--- a/api/organisations/models.py
+++ b/api/organisations/models.py
@@ -415,9 +415,7 @@ class Subscription(LifecycleModelMixin, SoftDeleteExportableModel):
         return cb_metadata
 
     def _get_subscription_metadata_for_self_hosted(self) -> BaseSubscriptionMetadata:
-        if is_enterprise() and hasattr(
-            self.organisation, "licence"
-        ):  # pragma: no cover
+        if hasattr(self.organisation, "licence"):  # pragma: no cover
             licence_information = self.organisation.licence.get_licence_information()
             return BaseSubscriptionMetadata(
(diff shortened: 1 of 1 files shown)


api/organisations/models.py
```diff
@@ -415,9 +415,7 @@ def _get_subscription_metadata_for_chargebee(self) -> ChargebeeObjMetadata:
         return cb_metadata
 
     def _get_subscription_metadata_for_self_hosted(self) -> BaseSubscriptionMetadata:
-        if is_enterprise() and hasattr(
-            self.organisation, "licence"
-        ):  # pragma: no cover
+        if hasattr(self.organisation, "licence"):  # pragma: no cover
```
matthewelwell, line 418: Sorry, I'm super confused now. The `is_enterprise()` function is unrelated to the data in the database - it is just checking whether the image that's being used is the enterprise version. 

This change will not resolve the issue highlighted by @rolodato [here](https://github.com/Flagsmith/licensing/issues/2) (if indeed it is an issue, or maybe this whole thing was just a miscommunication?).

zachaysan: I believe this was a whole chain of miscommunication. The steps @rolodato outlined were a miscommunication by me since I *thought* that the `is_enterprise()` check hit the database and mentioned it to him.

zachaysan: Ok I'm going to close this PR now, since we want to keep the `is_enterprise()` check.

PR #4988, opened 2025-01-10
feat: Add `version_of` and matches to rules and conditions

zachaysan:
## Changes

When creating a new change request for segments we need a way of tracing which provided rules and conditions match the candidate change request so that the frontend can more visually associate which rules and conditions match between the change request and the targeted segment (the one that will be updated).

To do so I've reused the concept of `version_of` which was originally used for segments and I associate a rule to another rule (or sub-rule to sub-rule) and in that process also match up conditions at the same layer. If at least one condition matches, then the sub-rule and rule match and future matching for the subsequent sub-rule is guided by holding onto the match between the parent rule and its sister rule. This way matches are only ever tree like in nature.

### Remaining work

Since the matching code is ready in this PR another PR will need to be opened to tie in these changes into the relevant code. Probably both the Workflows and Commons will need to be updated.

Lots of new tests. Some are simple, like checking that `version_of` gets correctly set, which is necessary due to an audit log bug we had when `version_of` was setup for the segments work.


api/pyproject.toml
```diff
@@ -170,7 +170,7 @@ hubspot-api-client = "^8.2.1"
 djangorestframework-dataclasses = "^1.3.1"
 pyotp = "^2.9.0"
 flagsmith-task-processor = { git = "https://github.com/Flagsmith/flagsmith-task-processor", tag = "v1.1.1" }
-flagsmith-common = { git = "https://github.com/Flagsmith/flagsmith-common", tag = "v1.4.2" }
+flagsmith-common = { git = "https://github.com/Flagsmith/flagsmith-common", branch = "feat/add_version_of_to_rules_and_conditions" }
```
matthewelwell, line 174: Let's remember to update these once we've got builds in those packages.

api/segments/models.py
```diff
@@ -198,6 +198,38 @@ def deep_clone(self) -> "Segment":
...
+        _rules = self.rules.all()
+        matched_rules = set()
+        matched_sub_rules = set()
+
+        for rule in segment.rules.all():
+            sub_rules = rule.rules.all()
+            for sub_rule in sub_rules:
+                sub_rule_matched = False
+                for _rule in _rules:
+                    if sub_rule_matched:
```
matthewelwell: As far as I can tell, the only time we set `sub_rule_matched = True`, we already `break` so do we actually need this, or is it just belt and safeguarding?

api/segments/models.py
```diff
@@ -198,6 +198,38 @@ def deep_clone(self) -> "Segment":
 
         return cloned_segment
 
+    def match_rules_to_segment(self, segment: "Segment") -> None:
+        _rules = self.rules.all()
```
matthewelwell: I think we should come up with a better way of differentiating 'self' and 'other' segment entities here rather than just the leading underscore.

api/segments/models.py
```diff
@@ -198,6 +198,38 @@ def deep_clone(self) -> "Segment":
...
+        matched_rules = set()
+        matched_sub_rules = set()
+
+        for rule in segment.rules.all():
+            sub_rules = rule.rules.all()
+            for sub_rule in sub_rules:
```
matthewelwell:
```suggestion
            for sub_rule in rule.rules.all():
```

api/segments/models.py
```diff
@@ -198,6 +198,38 @@ def deep_clone(self) -> "Segment":
...
+                    if sub_rule_matched:
+                        break
+
+                    if _rule in matched_rules and _rule.version_of != rule:
+                        continue
+
+                    if rule.type != _rule.type:
+                        continue
+                    for _sub_rule in _rule.rules.all():
+                        if _sub_rule in matched_sub_rules:  # pragma: no cover
```
matthewelwell: Why do we need the pragma no cover here?

api/segments/models.py
```diff
@@ -198,6 +198,38 @@ def deep_clone(self) -> "Segment":
 
         return cloned_segment
 
+    def match_rules_to_segment(self, segment: "Segment") -> None:
```
matthewelwell: This code is really hard to follow. I've added some specific suggestions below, but we definitely need a docstring here, especially as this code isn't even (seemingly at least) called from this repository.

api/segments/models.py
```diff
@@ -261,13 +300,44 @@ def get_segment(self):
...
+        if not conditions and not _conditions:
+            # Empty rule with the same type matches.
+            return True
+
+        matched_conditions = set()
+
+        for condition in conditions:
+            for _condition in _conditions:
+                if _condition in matched_conditions:
+                    continue
```
matthewelwell: I don't really understand the need for this check?

api/segments/models.py
```diff
@@ -261,13 +300,44 @@ def get_segment(self):
             rule = rule.rule
         return rule.segment
 
+    def matches_rule(self, rule: "SegmentRule") -> bool:
```
matthewelwell: Again, let's add a docstring here please. Also, why does this method return a bool, but the parent one doesn't?

api/segments/models.py
zachaysan: We break from a more nested part of the codebase. The block that's named 

```python
for _sub_rule in _rule.rules.all():
```

So we need the `sub_rule_matched` bool check to break the parent

```python
for _rule in _rules:
```

Part of the codebase.

matthewelwell: Ah yes, ok, sorry.

api/segments/models.py
zachaysan: Because this part here:

```python
for _condition in _conditions:
```

Is subnested in the parent 

```python
for condition in conditions:
```

So if we have already matched a `_condition` we do not want to evaluate it again.

api/segments/models.py
zachaysan: No problem on the docstring. Added locally. As for returning bool here and not returning a bool in the calling method, it's just about the callers needing a bool or not. The callers of `match_rules_to_segment` don't need a bool as to whether or not there was a match. But it is easy enough to add so I've altered the `match_rules_to_segment` to return a bool.

api/segments/models.py
zachaysan: Ok I used the word `self` instead of the leading underscore for this method's logic as well as the `matches_rule` method as well.

api/segments/models.py
zachaysan: 1. Yeah the code is being called from the workflows repository (where there is more testing logic there as well). 
2. I've added a docstring to this method.

api/segments/models.py
zachaysan: I've altered a test to expand coverage to tackle it.

api/pyproject.toml
zachaysan: Good idea.

api/segments/models.py
```diff
@@ -198,6 +198,71 @@ def deep_clone(self) -> "Segment":
...
+                 with those of the current object's sub-rules.
+               - If a match is found:
+                 - Update the `version_of` field for the matched sub-rule and rule.
+                 - Track the matched rules and sub-rules to avoid duplicate processing.
+            3. Perform a bulk update on matched rules and sub-rules to persist
+               versioning changes.
+
+        Side Effects:
+            - Updates the `version_of` field for matched rules and sub-rules.
+        """
```
matthewelwell: This is good, but it doesn't really explain the 'why'. With enough time, I can glean all of this information from reading the code, what I'm struggling to get from reading the code is what this method is actually used for?

api/segments/models.py
matthewelwell: As above, I'd like more of the 'why' here.

api/pyproject.toml
```diff
@@ -197,7 +197,7 @@ flagsmith-ldap = { git = "https://github.com/flagsmith/flagsmith-ldap", tag = "v
 optional = true
 
 [tool.poetry.group.workflows.dependencies]
-workflows-logic = { git = "https://github.com/flagsmith/flagsmith-workflows", tag = "v2.7.5" }
+workflows-logic = { git = "https://github.com/flagsmith/flagsmith-workflows", branch = "feat/add_matches_to_workflows" }
```
zachaysan, line 201: Need to fix this one as well before merging

api/segments/models.py
zachaysan: Good point. I've added a "why" paragraph in order to answer that question.

api/segments/models.py
zachaysan: Ok I've added more to the docstring here as well.

api/segments/models.py
```diff
@@ -198,6 +198,76 @@ def deep_clone(self) -> "Segment":
...
+                    for self_sub_rule in self_rule.rules.all():
+                        if self_sub_rule in matched_sub_rules:
+                            continue
+                        if self_sub_rule.assign_conditions_if_matching_rule(sub_rule):
+                            self_sub_rule.version_of = sub_rule
+                            sub_rule_matched = True
+                            matched_sub_rules.add(self_sub_rule)
+                            self_rule.version_of = rule
+                            matched_rules.add(self_rule)
+                            break
```
matthewelwell: Sorry @zachaysan , this code is still unreadable to me. I was hoping that the docstring and renaming of the variables would help, but it hasn't really. Can we add some inline comments on each of the conditional statements perhaps?

matthewelwell: It feels to me like this is perhaps evidence that we're thinking about this wrong? Could we instead just store a 'diff' object (probably in JSON), or set of objects, which look(s) something like: 

```python
class SegmentConditionChanges:
    condition: Condition

    new_operator: str
    new_property: str
    new_value: str

class SegmentRuleChanges:
    rule: SegmentRule

    new_type: str

    new_rules: list[SegmentRule]
    deleted_rule_ids: list[int]
    updated_rules: list[SegmentRuleChanges]

    new_conditions: list[Condition]
    delete_condition_ids: list[int]
    updated_conditions: list[SegmentConditionChanges]

class SegmentChangeSet:
    segment: Segment

    new_name: str
    new_description: str

    new_rules: list[SegmentRule]
    deleted_rule_ids: list[int]
    updated_rules: list[SegmentRuleChanges]
```

... maybe it's a bit late in the day for this, but I feel like this might be more flexible?

api/tests/unit/segments/test_unit_segments_models.py
```diff
@@ -451,3 +451,268 @@ def test_segment_get_skip_create_audit_log_when_exception(
...
+    # When
+    condition2 = Condition.objects.create(
+        operator=PERCENTAGE_SPLIT, value=0.1, rule=rule2, version_of=condition1
+    )
+
+    # Then
+    assert condition2.version_of == condition1
+
+
+def test_assign_matching_rules_to_segment_simple(project: Project) -> None:
```
matthewelwell: Can we be more explicit with the test names? What does `_simple` mean? Alternatively (or maybe as well as), let's add more comments / docstrings to these tests.

api/segments/models.py
```diff
@@ -261,13 +338,74 @@ def get_segment(self):
...
+        matched_conditions = set()
+
+        for condition in conditions:
+            for self_condition in self_conditions:
+                if self_condition in matched_conditions:
+                    continue
+                if (
+                    condition.operator == self_condition.operator
+                    and condition.property == self_condition.property
+                ):
```
matthewelwell: Shouldn't we only care about the property here, rather than the operator as well? What if they want to change it from e.g. `organisation_id EQUALS 123` to `organisation_id IN 123,456` - that seems like a legitimate use case that we want to cater for, right?

matthewelwell: Ah, but there's perhaps an issue with operators like percentage split in that case? But I think the logic should be something like: 

```python
if not condition.property: 
    if condition.operator == self_condition.operator:
        # do valid match logic
elif condition.property == self_condition.property:
    # do valid match logic
```

My concern here as well is that I don't think there is anyone stopping someone from adding e.g. 2 percentage split conditions to a segment. It would be a stupid idea, but there's nothing stopping it from happening as far as I know, and there are likely other scenarios where this is the case. 

For example, there are a lot of segments that look something like: 

```
organisation_id IN 123,456,...
OR 
organisation_id IN abc,def,...
```

This is to get around our limitation on the size of the value in a segment condition. I don't think your logic here (or mine suggested above for that matter) will handle this scenario very well, unless I'm missing something?

api/tests/unit/segments/test_unit_segments_models.py
zachaysan: Ok I've updated the test name and have added comments locally.

api/segments/models.py
zachaysan: Ok I've added some inline comments as we discussed.

api/segments/models.py
zachaysan: I've updated the condition matching logic to now include operator differences as well as adding a section to the code to look for exact condition matches first.

api/segments/models.py
```diff
@@ -198,6 +198,93 @@ def deep_clone(self) -> "Segment":
...
+                # Set a bool flag to be used to break out of the
+                # below self_rules iteration if a sub_rule matches.
+                sub_rule_matched = False
+                for self_rule in self_rules:
+                    if sub_rule_matched:
+                        break
+
+                    # If a self_rule has already been matched then
+                    # the only time we care about it is when we
+                    # further iterate on the the same sub_rules
```
matthewelwell:
```suggestion
                    # further iterate on the same sub_rules
```

api/segments/models.py
```diff
@@ -261,13 +355,107 @@ def get_segment(self):
...
+            return True
+
+        matched_conditions = set()
+
+        # In order to provide accurate diffs we first go through the conditions
+        # and collect conditions that have matching values (property, operator, value).
+        for condition in conditions:
+            for self_condition in self_conditions:
+                if self_condition in matched_conditions:
+                    continue
```
matthewelwell: I really can't see a scenario where this could be the case?

api/segments/models.py
```diff
@@ -198,6 +198,93 @@ def deep_clone(self) -> "Segment":
...
+
+        self_rules = self.rules.all()
+        matched_rules = set()
+        matched_sub_rules = set()
+
+        for rule in segment.rules.all():
+            for sub_rule in rule.rules.all():
+
+                # Set a bool flag to be used to break out of the
+                # below self_rules iteration if a sub_rule matches.
```
matthewelwell: Why? Please _always_ try to phrase comments about the why, not the what. I can infer exactly what this comment says from a quick glance at the variable name, and its use below, but I'm still none the wiser really as to _why_.

api/segments/models.py
```diff
@@ -198,6 +198,93 @@ def deep_clone(self) -> "Segment":
...
+                sub_rule_matched = False
+                for self_rule in self_rules:
+                    if sub_rule_matched:
+                        break
+
+                    # If a self_rule has already been matched then
+                    # the only time we care about it is when we
+                    # further iterate on the the same sub_rules
+                    # that are present in the matching parent rule.
```
matthewelwell: Again, why? This logic is super hard to follow and all these comments are doing is just rephrasing the code that I can read already. It's super hard to follow what self_rule, sub_rule and 'parent rule' here are. Can we reword this to talk about the business logic rather than the code?

api/segments/models.py
```diff
@@ -198,6 +198,93 @@ def deep_clone(self) -> "Segment":
...
+                    if self_rule in matched_rules and self_rule.version_of != rule:
+                        continue
+
+                    # To eliminate false matches we force the types
+                    # to be the same for the rules.
```
matthewelwell: What is a 'false match' ?

api/segments/models.py
```diff
@@ -198,6 +198,93 @@ def deep_clone(self) -> "Segment":
...
+               - A match is determined if the sub-rule's type and properties align
+                 with those of the current object's sub-rules.
+               - If a match is found:
+                 - Update the `version_of` field for the matched sub-rule and rule.
+                 - Track the matched rules and sub-rules to avoid duplicate processing.
+            3. Perform a bulk update on matched rules and sub-rules to persist
+               versioning changes.
+
+        Side Effects:
+            - Updates the `version_of` field for matched rules and sub-rules.
```
matthewelwell: Which segment does it modify though?

api/segments/models.py
```diff
@@ -198,6 +198,93 @@ def deep_clone(self) -> "Segment":
...
+        Assign matching rules of the current object to the rules of the given segment.
+
+        This method iterates through the rules of the provided `Segment` and
+        attempts to match them to the current object's rules and sub-rules,
+        updating versioning where matches are found.
+
+        This is done in order to set the `version_of` field for matched rules
+        and conditions so that the frontend can more reliably diff rules and
+        conditions between a change request for a segment and the target segment
+        itself.
```
matthewelwell: This still doesn't really talk about when it should be used? One of the biggest things I'm finding is that I'm consistently struggling to follow which segment is which. Like should we be more explicit here (and in all of the rule / condition variables) to say something like `current_segment`, `modified_segment` or something? 

Maybe the method could then be more explicit like: `update_segment_with_matches_from_current_segment(self, current_segment: "Segment")`? 

This feels like it's pigeon-hoing us a little bit into _only_ using this functionality for change requests, but since we're modifying the segments, it feels like this functionality can only ever be used for that purpose anyway?

api/segments/models.py
zachaysan: Ok I've updated this to explicitly say that it calls the segment `self`.

api/segments/models.py
zachaysan: > This still doesn't really talk about when it should be used?

Ok I've updated the introductory paragraph to include that `self` is the target of the changes and that this is called in the context of a change request.

> One of the biggest things I'm finding is that I'm consistently struggling to follow which segment is which. Like should we be more explicit here (and in all of the rule / condition variables) to say something like current_segment, modified_segment or something?

I've updated the code to reference `current_[segment,rule,sub_rule,condition]` and I've added references to which object is being updated by referencing them as `modified_[rule,sub_rule,condition]` to make it clear which objects are optionally being modified.

> Maybe the method could then be more explicit like: update_segment_with_matches_from_current_segment(self, current_segment: "Segment")?

Ok I've updated the method name and args.

> This feels like it's pigeon-hoing us a little bit into only using this functionality for change requests, but since we're modifying the segments, it feels like this functionality can only ever be used for that purpose anyway?

Yes this functionality is only going to be used for change requests.

api/segments/models.py
zachaysan: If the condition had previously matched another condition, then when the outer loop starts again we don't want to match it again. This is useful for conditions that have the same property and could match different conditions.

api/segments/models.py
zachaysan: Ok I've tried to lean in with a `Because` to guide my writing, but it's hard not to explain code when writing about why a variable exists.

api/segments/models.py
zachaysan: Ok I've tried to do this one as well.

api/segments/models.py
zachaysan: When the match shouldn't go through.

zachaysan (commented):
Since I'm on my last day I'd like to add some comments to this PR about where we are at.

Overall this work should be considered to be feature complete. Change requests for segments can be created, updated, approved, and committed in order to update the underlying segment. The attribute `version_of` was added to rules and conditions to allow for the frontend to assign diffs in conjunction with what a likely diff should be.

In order to properly merge this code the following updates need to happen:
1. `flagsmith-common` in `pyproject.toml` must be updated to a **new** tag that include previously merged work, which includes https://github.com/Flagsmith/flagsmith-common/pull/11.
2. This PR https://github.com/Flagsmith/flagsmith-workflows/pull/63 needs updating as well, both for the same `flagsmith-common` library as well as for the image in the `docker-compose` file which must be set to `ghcr.io/flagsmith/flagsmith-api-test:main`. I'm not sure how to solve the circular dependency because we also must...
3. Update the `workflows-logic` in `pyproject.toml` in this repo once step 2 can be merged into the workflows repository and tagged as a release.

When in doubt about how to run some chunk of functionality, look into the tests in the workflows PR here: https://github.com/Flagsmith/flagsmith-workflows/pull/63 including tests that pre-exist that PR.

emyller:
Moving the effort to #5768.

PR #4989, opened 2025-01-10
chore: MFA Method debugging / support improvements

matthewelwell:
## Changes

Add created_at and updated_at to MFAMethod class. Note that I chose not to use auto_now and auto_now_add because I wanted existing records to only be marked as updated once they have in fact been updated to avoid confusion. 

N/a - trivial changes


rolodato (approved):
I don't expect us to actually ever modify existing MFA methods, but ideally we shouldn't allow editing the created/updated_at fields from the details page:

image

Not a blocker since this still solves the immediate issue, but otherwise LGTM 👍

PR #5004, opened 2025-01-15
fix: Don't display password managers on irrelevant input fields

rolodato:

## Changes

All Input components by default have `one-time-code` autocomplete. This causes password managers such as 1Password to display a prompt on all kinds of inputs where it makes no sense and is annoying:

image

image

This change removes this default.

The only place where `one-time-code` should be used is in the MFA code input, which is added by this change:

image

Note that at MFA autocomplete still works (at least in 1Password) without setting `autocomplete=one-time-code`.

Lastly, this change also sets explicit `autocomplete` values for changing passwords. These also still work in 1Password currently, but is good practice to have them.

Manually tested different input fields in the frontend.

### DIFF
--- a/frontend/web/components/SimpleTwoFactor/prompt.js
+++ b/frontend/web/components/SimpleTwoFactor/prompt.js
@@ -28,6 +28,7 @@ export default class TheComponent extends PureComponent {
         >
           <InputGroup
             inputProps={{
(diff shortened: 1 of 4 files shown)


tiagoapolo:
This is great!

tiagoapolo (approved):

PR #5006, opened 2025-01-15
docs: Improve import/export docs

rolodato:
This change depends on https://github.com/Flagsmith/flagsmith-charts/pull/305, which allows mounting additional volumes into API containers when using Helm.

Adds an index page for import/export to explain the different options and how to choose. See https://docs-git-docs-import-export-flagsmith.vercel.app/system-administration/importing-and-exporting/

Removes references to localstack since they are not very actionable - we can add more examples later.


docs/docs/system-administration/importing-and-exporting/organisations.md
```diff
@@ -33,150 +36,213 @@ We **will** export the following entities:
...
+following methods:
 
-![Image](/img/organisations-admin.png)
+* From the Flagsmith dashboard, click your organisation name in the top left. The organisation ID is displayed in 
+  the URL bar: `https://flagsmith.example.com/organisation/YOUR_ORGANISATION_ID/...`.
+* From [Django Admin](/deployment/configuration/django-admin), browse to the Organisations section in the sidebar. 
+  Here you can see all of your organisations and their IDs. 
+* If you have an Admin API key, call the
+  [List Organisations API endpoint](https://api.flagsmith.com/api/v1/docs/#/api/api_v1_organisations_list). This 
+  returns all the organisations that your API key is a member of.
```
matthewelwell: Technically you can use the 'show JSON references' here too, but maybe that's just more complicated to explain here than needs be?

docs/docs/system-administration/importing-and-exporting/organisations.md
```diff
@@ -33,150 +36,213 @@ We **will** export the following entities:
...
+
+To provide credentials, set the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment variables before running
+any commands. For example:
+
+```
+export AWS_ACCESS_KEY_ID='abc123'
+export AWS_SECRET_ACCESS_KEY='xyz456'
+```
+
+By default, all commands will interact with buckets hosted on AWS S3. To use other S3-compatible services such as Google
```
matthewelwell, line 250: > other S3-compatible services such as Google

TIL

matthewelwell (approved):

docs/docs/system-administration/importing-and-exporting/organisations.md
rolodato: It's 2 more clicks than looking at the URL bar (open Org Settings, then expand the JSON data), not everyone has it enabled, and users that don't want it will need to disable it again. We also display the org ID in the Billing section but this is not shown onprem:

image

IMO we should just display the org ID somewhere easier to find, which I didn't want to think about for this PR.

rolodato:
Thanks @matthewelwell - going to wait until https://github.com/Flagsmith/flagsmith-charts/pull/305 is released before merging, and adding a note on the minimum required Helm chart version.

docs/docs/system-administration/importing-and-exporting/organisations.md
matthewelwell: Yeah, agreed. I think in the past I've seen scenarios where like clicking on the organisation name in the top left copies the ID to your clipboard or something. Agreed that it's not in the scope of this PR, and I'm happy to just suggest the URL approach.

PR #5011, opened 2025-01-17
fix: (WIP) Disable Secure cookies by default, enable when using HTTPS

### LINKED ISSUE #4940: Missing authentication token in cookies

- [x] Self Hosted with Docker
- [ ] SaaS at flagsmith.com

### Describe the bug

I have deployed Flagsmith on my server. After logging in, if I refresh the page, I am redirected to the login page. But in my localhost setup it works fine.
After some research I found that Flagsmith on my server has no auth token in cookies.

### Steps To Reproduce

1. Login
2. Click refresh in browser

### Expected behavior

Stay on page

### Screenshots

Before refresh:
Image

After refresh:
Image

On localhost:
Image

rolodato:
(Not ready to merge yet - only marked as not draft to trigger image builds)

This change possibly fixes https://github.com/Flagsmith/flagsmith/issues/4940, though is a breaking change.

Secure cookies are now opt-in instead of opt-out. By default, secure cookies are now automatically enabled if the request is [secure](https://docs.djangoproject.com/en/5.1/ref/request-response/#django.http.HttpRequest.is_secure). Users can still force secure cookies if needed.

Having secure cookies enabled by default means that Flagsmith must be used on HTTPS or localhost exclusively. Deploying our provided Docker Compose definition or Helm chart and trying to access it over anything except HTTPS or localhost will fail, unless secure cookies are disabled. Specifically this will fail on any origin that is not [potentially trustworthy](https://www.w3.org/TR/secure-contexts/#is-origin-trustworthy), which is essentially everything except HTTPS, `127.0.0.1`, `::1`, and localhost.

This is particularly annoying when testing Kubernetes or Docker deployments, where it's often easier to access services using service URLs and not port forwarding to localhost. The behaviour when secure cookies fail is also confusing, where users can log in but are immediately logged out with no visible errors when reloading the page.

I consider these new defaults to be more secure for most users - they allow testing and experimenting over HTTP when it is not a concern, and they provide secure behaviour when using HTTPS in production. With the current situation, users might forget to opt back into secure cookies after needing to disable them for testing over HTTP, and carry a less secure configuration into a production deployment.

There are legitimate use cases for using Flagsmith in production over HTTP such as having a VPN deal with TLS instead of the application, or local networks.

The Helm chart could also be made to force HTTPS depending on values or ingress configuration.

## Testing

Running the `ghcr.io/flagsmith/flagsmith-private-cloud:pr-5011` image on Kubernetes, we can successfully log in and access the cluster over HTTP using the default settings:

image

Connecting with the same settings over HTTPS sets a secure cookie - the load balancer here is OrbStack using Docker Compose:

image


kyle-ssg:
I suppose this should be considered a major update since people's current environment variables may rely on the old defaults to achieve what they want.

PR #5022, opened 2025-01-20
fix: Disable invite button when email config is not set

### LINKED ISSUE #4036: Don't show 'Invite members' if email is not set up
Currently the 'Users and permissions' screen in organisation settings will show the 'Invite members' button (which triggers an email invite) regardless of whether email configuration has been added to the API. 

We have a couple of options to solve this: 

 1. If the FE receives an error when sending an invite, then show a dialog recommending that they use invite links instead
 2. The API can add an attribute to e.g. the `version` endpoint which indicates whether email configuration has been added or not. If not, the FE can disable the button and show a tooltip. 

The preference here is likely (2), but I did enjoy the experience from Metabase which is more along the lines of (1).

tiagoapolo:

## Changes

- Disables button when email provider is not set
- Shows tooltip text
- adds `has_email_provider` field to /version, that checks if email config was set.
Screenshot 2025-01-20 at 18 14 58

- API test for /version
- Manually tested in UI, access `/account` with no email config set in API, invite member button should be disabled and tooltip is shown when hovering.

### DIFF
--- a/api/tests/unit/app/test_unit_app_utils.py
+++ b/api/tests/unit/app/test_unit_app_utils.py
@@ -137,15 +137,30 @@ def test_get_version_info_with_email_config_ses(mocker: MockerFixture) -> None:
 
 
 def test_get_version_info_without_email_config(mocker: MockerFixture) -> None:
-
-    # When
-    result = get_version_info()
-
-    # Then
-    assert result == {
+    expected = {
         "ci_commit_sha": "unknown",
         "image_tag": "unknown",
         "has_email_provider": False,
         "is_enterprise": False,
         "is_saas": False,
     }
+
+    mocker.patch(
+        "app.utils.EMAIL_BACKEND", "django.core.mail.backends.smtp.EmailBackend"
+    )
+    mocker.patch("app.utils.EMAIL_HOST_USER", None)
+
+    result = get_version_info()
+    assert result == expected
+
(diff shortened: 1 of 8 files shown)


kyle-ssg (approved):
FE looks fine but looks like there's some api unit test failures

api/tests/unit/app/test_unit_app_utils.py
```diff
@@ -76,6 +77,90 @@ def path_side_effect(file_path: str) -> mocker.MagicMock:
...
         "is_saas": False,
     }
+
+
+def test_get_version_info_with_email_config_smtp(mocker: MockerFixture) -> None:
+
+    mocker.patch(
+        "app.utils.EMAIL_BACKEND", "django.core.mail.backends.smtp.EmailBackend"
+    )
+    mocker.patch("app.utils.EMAIL_HOST_USER", "user")
```
zachaysan: Instead of using `mocker.patch()` I'd point you towards using the fixture `settings: SettingsWrapper` which will override these constants all at once and saving the test suite from getting into an inconsistent state where they are sometimes set and other times unset.

Of course this suggestion applies to all of the tests.

zachaysan (approved):
Overall looks good, but I have one major point of feedback for the tests.

matthewelwell (approved):

Issue #4032, opened 2024-05-27
Possible to create multiple environments with the same name in a project


rolodato:
### How are you running Flagsmith

- [ ] Self Hosted with Docker
- [ ] Self Hosted with Kubernetes
- [X] SaaS at flagsmith.com
- [ ] Some other way (add details in description below)

### Describe the bug

Two environments with the same name in the same project can only be distinguished by their ID, which is not human-readable. This is error-prone and can lead to data loss by accidentally deleting the wrong environment.

For example, it's impossible to tell these environments apart:

![image](https://github.com/Flagsmith/flagsmith/assets/829698/44cbef50-557b-421b-b30e-cc5478596077)

![image](https://github.com/Flagsmith/flagsmith/assets/829698/1f6e4a9e-740f-4786-ab36-3a3aac61eeb3)

When deleting an environment, we make users type the target environment name to confirm they are deleting the correct one, but this is not enough to identify it:

![image](https://github.com/Flagsmith/flagsmith/assets/829698/30f3aed3-5cf1-4d34-84f1-17b796bdace9)

### Steps To Reproduce

1. Create a new environment within a project
2. Give the environment the same name as the one you are cloning from
3. You now have two environments with the same name

### Expected behavior

Trying to create an environment with the same name within a project should fail with a 400 error.

I see two possible solutions:

1. Deduplicate environment names within projects by renaming them, e.g. add `(1)`, `(2)`, etc suffixes to duplicates based on when they were created, and then enforce this unique constraint at the model/database level
2. Keep all environment names as they are, and only prevent new duplicates from being created

matthewelwell:
As per the solution in #5038 , I think we should go with option 2 (at least for now) from the above since environment names are also used to e.g. tag data entities in other platforms that we integrate with.

matthewelwell:
Resolved in https://github.com/Flagsmith/flagsmith/pull/5058

PR #5023, opened 2025-01-21
feat: Backend support for feature health

khvn26:

## Changes

- Auditable feature health events
- Sample feature health event provider
- Feature health event ingestion webhook with signed paths
- "Unhealthy" system tag automatically added to features with corresponding feature health events

To create a Sample feature health provider, perform the following request:

```sh
curl -x POST http://<base-api-url>/api/v1/projects/<project-id>/feature-health/providers -H 'authorization: Token <user token>' --data-raw '{"type": "Sample"}'
```

or use the admin interface.

Request the resulting webhook URL:

```sh
curl -x POST http://<base-api-url>/api/v1/feature-health/<webhook-hash> --data-raw '{"environment": "<your-env-name>", "status": "unhealthy", "feature": "<your-feature-name>"}'
```

Observe the new system tag being added to your feature of choice.

Get a list of latest feature events per feature-provider pair:

```sh
curl http://<base-api-url>/api/v1/projects/<project-id>/feature-health/events -H 'authorization: Token <user token>'
```

_Please describe._


api/features/feature_health/constants.py
```diff
@@ -0,0 +1,13 @@
...
+FEATURE_HEALTH_EVENT_CREATED_MESSAGE = "Health status changed to %s for feature %s."
+FEATURE_HEALTH_EVENT_CREATED_FOR_ENVIRONMENT_MESSAGE = (
+    "Health status changed to %s for feature %s in environment %s."
+)
+FEATURE_HEALTH_EVENT_CREATED_PROVIDER_MESSAGE = "\n\nProvided by %s"
+FEATURE_HEALTH_EVENT_CREATED_REASON_MESSAGE = "\n\nReason:\n%s"
+
+UNHEALTHY_TAG_COLOUR = "#FFC0CB"
+
+FEATURE_HEALTH_WEBHOOK_PATH_PREFIX = "/api/v1/feature-health/"
```
matthewelwell: I don't love that this is a constant vs using `reverse` but I'll read on...

api/api/urls/v1.py
```diff
@@ -49,6 +50,12 @@
     # GitHub integration webhook
     re_path(r"github-webhook/", github_webhook, name="github-webhook"),
     re_path(r"cb-webhook/", chargebee_webhook, name="chargebee-webhook"),
+    # Feature health webhook
+    re_path(
+        r"feature-health/(?P<path>.{0,100})$",
+        feature_health_webhook,
+        name="feature-health-webhook",
+    ),
```
zachaysan, line 58: Why is the path limited to 100 characters?

api/features/feature_health/models.py
```diff
@@ -0,0 +1,163 @@
...
+    abstract_base_auditable_model_factory(["uuid"]),
+):
+    history_record_class_path = (
+        "features.feature_health.models.HistoricalFeatureHealthProvider"
+    )
+    related_object_type = RelatedObjectType.FEATURE_HEALTH
+
+    name = models.CharField(max_length=50, choices=FeatureHealthProviderName.choices)
+    project = models.ForeignKey("projects.Project", on_delete=models.CASCADE)
+    created_by = models.ForeignKey("users.FFAdminUser", on_delete=models.CASCADE)
```
zachaysan: Shouldn't this be set to null if the user is deleted?

api/features/feature_health/services.py
```diff
@@ -0,0 +1,115 @@
...
+        )
+        return None
+    project = provider.project
+    if feature := Feature.objects.filter(
+        project=provider.project, name=response.feature_name
+    ).first():
+        if response.environment_name:
+            environment = Environment.objects.filter(
+                project=project, name=response.environment_name
+            ).first()
```
zachaysan, line 60: I know this is an annoying question, but given that we have some customers that use the same name for multiple environments are we at least reliably mapping to the same environment between invocations?

api/tests/integration/features/feature_health/test_views.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+@pytest.mark.freeze_time("2023-01-19T09:09:47.325132+00:00")
+def test_webhook__unhealthy_feature__post__expected_feature_health_event_created__expected_tag_removed(
+    unhealthy_feature: int,
+    project: int,
+    feature_name: str,
+    sample_feature_health_provider_webhook_url: str,
+    api_client: APIClient,
+    admin_client: APIClient,
```
zachaysan: Here and elsewhere consider using `admin_client_new` for this fixture.

zachaysan (approved):
Overall the code looks great. I have a few small suggestions or questions but nothing is a showstopper.

api/api/urls/v1.py
khvn26: This is to prevent the business logic from parsing webhook hashes that are unreasonably long. Valid hashes are not expected to be longer than 100 characters.

api/features/feature_health/models.py
```diff
@@ -0,0 +1,163 @@
...
+    environment = models.ForeignKey(
+        "environments.Environment",
+        on_delete=models.CASCADE,
+        related_name="feature_health_events",
+        null=True,
+    )
+
+    created_at = models.DateTimeField(auto_now_add=True)
+    type = models.CharField(max_length=50, choices=FeatureHealthEventType.choices)
+    provider_name = models.CharField(max_length=255, null=True, blank=True)
```
matthewelwell, line 131: Shouldn't we link the `FeatureHealthProvider` here?

api/features/feature_health/models.py
```diff
@@ -0,0 +1,163 @@
...
+    feature = models.ForeignKey(
+        "features.Feature",
+        on_delete=models.CASCADE,
+        related_name="feature_health_events",
+    )
+    environment = models.ForeignKey(
+        "environments.Environment",
+        on_delete=models.CASCADE,
+        related_name="feature_health_events",
+        null=True,
```
matthewelwell, line 126: I'm interested to know in which scenarios a feature is marked as unhealthy without the context of an environment?

api/features/feature_health/providers/sample/mappers.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def map_sample_event_status_to_feature_health_event_type(
+    status: SampleEventStatus,
+) -> FeatureHealthEventType:
+    return (
+        FeatureHealthEventType.UNHEALTHY
+        if status == "unhealthy"
+        else FeatureHealthEventType.HEALTHY
+    )
```
matthewelwell, line 18: Was a decision made to not have a `FeatureHealthEventType.UNKNOWN` ?

api/features/feature_health/providers/sample/mappers.py
```diff
@@ -0,0 +1,33 @@
...
+def map_payload_to_provider_response(
+    payload: str,
+) -> FeatureHealthProviderResponse:
+    event_data: SampleEvent = json.loads(payload)
+
+    return FeatureHealthProviderResponse(
+        feature_name=event_data["feature"],
+        environment_name=event_data.get("environment"),
+        event_type=map_sample_event_status_to_feature_health_event_type(
+            event_data["status"]
```
matthewelwell, line 30: This (combined with line 24 above) means (I think, although maybe a serializer handles this) that we could have an unhandled exception here if "status" is not provided.

api/features/feature_health/models.py
khvn26: Excellent question! Amended the model.

api/features/feature_health/services.py
khvn26: The answer is complicated. Formally, we expect the environment names to be unique and the [current predicament](https://github.com/Flagsmith/flagsmith/issues/4032) to be fixed eventually. The real solution would be to start exporting entity slugs/external uuids specifically for integrations. Both of these points are out of the scope of this PR, and I'm reluctant to accommodate new code for current erroneous behaviour. But maybe we should do it.

api/tests/integration/features/feature_health/test_views.py
khvn26: Thanks — this actually exposed a bug when creating feature providers via master API keys 🚀

api/features/feature_health/providers/services.py
```diff
@@ -0,0 +1,33 @@
...
+
+logger = structlog.get_logger("feature_health")
+
+_provider_webhook_signer = signing.Signer(sep="/", salt="feature_health")
+
+
+def get_webhook_path_from_provider(
+    provider: FeatureHealthProvider,
+) -> str:
+    return FEATURE_HEALTH_WEBHOOK_PATH_PREFIX + _provider_webhook_signer.sign_object(
```
matthewelwell: As per my earlier comment, could we do something like: 

```python
return reverse("api-v1:feature-health", args=(_provider_webook_signer.sign_object(...),) 
```

?

api/features/feature_health/providers/services.py
```diff
@@ -0,0 +1,33 @@
...
+def get_provider_from_webhook_path(path: str) -> FeatureHealthProvider | None:
+    try:
+        hex_string = _provider_webhook_signer.unsign_object(path)
+    except signing.BadSignature:
+        logger.warning("invalid-feature-health-webhook-path-requested", path=path)
+        return None
+    feature_health_provider_uuid = uuid.UUID(hex_string)
+    return FeatureHealthProvider.objects.filter(
+        uuid=feature_health_provider_uuid
+    ).first()
```
matthewelwell, line 35: Are we just using `.first()` here to return `None` in the case that it doesn't exist? We don't ever expect there to be more than one, right?

api/features/feature_health/serializers.py
```diff
@@ -0,0 +1,46 @@
...
+            "environment",
+            "feature",
+            "provider_name",
+            "reason",
+            "type",
+        )
+
+
+class FeatureHealthProviderSerializer(serializers.ModelSerializer):
+    created_by = serializers.SlugRelatedField(slug_field="email", read_only=True)
```
matthewelwell, line 27: Why use this SlugRelatedField vs just providing the whole user object?

matthewelwell: Won't this also cause issues with the api keys we just talked about too?

api/features/feature_health/services.py
```diff
@@ -0,0 +1,88 @@
...
+    return None
+
+
+def create_feature_health_event_from_provider(
+    provider: FeatureHealthProvider,
+    payload: str,
+) -> FeatureHealthEvent | None:
+    try:
+        response = get_provider_response(provider, payload)
+    except Exception as exc:
```
matthewelwell: Why do we need to catch all exceptions here?

api/features/feature_health/services.py
matthewelwell: The issue we might have is that I'm not sure that we can necessarily ever fix the duplicate data that exists right now for similar reasons I have called out on my PR [here](https://github.com/Flagsmith/flagsmith/pull/5039). 

I agree that it's beyond the scope of this PR, but I don't want to end up having to handle multiple identifiers in the future if we can avoid it. Are we cornering ourselves if we go with this `name` approach for now?

api/features/feature_health/services.py
```diff
@@ -0,0 +1,88 @@
...
+        )
+    return None
+
+
+def update_feature_unhealthy_tag(feature: "Feature") -> None:
+    if feature_health_events := [
+        *FeatureHealthEvent.objects.get_latest_by_feature(feature)
+    ]:
+        unhealthy_tag, _ = Tag.objects.get_or_create(
+            label="Unhealthy",
```
matthewelwell: This should be a constant

api/features/feature_health/views.py
```diff
@@ -0,0 +1,111 @@
...
+
+        created_by = None
+        if isinstance(self.request.user, FFAdminUser):
+            created_by = self.request.user
+
+        instance = self.model_class.objects.create(
+            project=project,
+            name=request_serializer.validated_data["name"],
+            created_by=created_by,
+        )
```
matthewelwell, line 89: Can't we replace this with `self.perform_create(request_serializer, project=project, created_by=created_by)` (or similar?)

matthewelwell (commented):
Reviewed sans tests

api/features/feature_health/services.py
khvn26: > Are we cornering ourselves if we go with this name approach for now?

No we're not, IMO. As soon as we have public human readable identifiers we'll be able to migrate the webhooks while retaining backwards compat with environment names.

matthewelwell: Then I'm happy accepting this restriction. There aren't any valid use cases I can think of for having multiple environments with the same name anyway.

api/features/feature_health/services.py
khvn26: Good point, clarified. Also changed the event name a bit to better reflect the expected mapping errors.

api/features/feature_health/services.py
khvn26: Added.

api/features/feature_health/views.py
khvn26: We would need to rewrite `CreateFeatureHealthProviderSerializer` to be a `ModelSerializer` for that and, to be honest, I don't feel it's worth the time. Let me know if you feel strongly about this.

api/features/feature_health/models.py
khvn26: I erred on the side of not having a foreign key relation here since it's time series data intended to be kept forever. I couldn't justify having a constraint for this column.

api/features/feature_health/constants.py
khvn26: How would you use `reverse` to generate webhook paths?

api/features/feature_health/serializers.py
khvn26: > Why use this SlugRelatedField vs just providing the whole user object?

Do we employ a convention to pass full user objects in views like this one? We'll want to change LaunchDarkly import request viewset too, in that case.

khvn26: > Won't this also cause issues with the api keys we just talked about too?

No it won't. `SlugRelatedField` respects nullable relation fields (and I decided to write null `created_by` for master api keys as there's really no existing solution for this in the codebase).

matthewelwell: My understanding was that we did, yes. It seems more extensible that way. I'm not going to die on this hill, but just an observation.

api/features/feature_health/views.py
matthewelwell: Nope, all good. Sorry, didn't notice it wasn't a model serializer.

api/features/feature_health/providers/sample/mappers.py
khvn26: Thanks! `create_feature_health_event_from_provider` now handling `KeyError` as well.

api/features/feature_health/providers/sample/mappers.py
khvn26: Yes. We're keeping it as simple as possible for the time being.

api/features/feature_health/models.py
khvn26: E.g. in order to use Grafana integration, you _must_ label your alert with a feature name. The environment label, however, is optional. If you do provide an environment label, the UI will direct you there; otherwise, the feature will be regarded as unhealthy for the whole project.

api/features/feature_health/constants.py
khvn26: Figured it out, added it.

api/features/feature_health/providers/services.py
khvn26: That's correct.

api/features/feature_health/models.py
matthewelwell: Seems like an odd behaviour imo, considering you _have_ to work in the context of an environment in Flagsmith, even if you only have a single environment, but I do concede that it's probably better to be less restrictive than more, so let's see how people want to use it :)

api/features/feature_health/providers/services.py
matthewelwell: Ok, minor suggestion, but we could add a comment to this effect.

matthewelwell (approved):
Approved with a few remaining minor comments.

api/features/feature_health/providers/services.py
khvn26: It's kind of obvious given we're selecting by uuid?

matthewelwell: Yeah, fair enough.

PR #5026, opened 2025-01-21
feat: multiple environment flag analytics

kyle-ssg:

## Changes

Prior to this change, the feature analytics tab would only let you view data from the environment you're in context of. Whilst this makes sense, we often want to check that a feature is no longer in use over all environments so that we can safely delete it.

Opening the analytics tab will preselect the environment you are viewing but allow you to view others.

image

As above viewed production feature usage.


matthewelwell:
I agree that this is useful, but it really highlights a flaw in the current UI architecture (which is definitely already highlighted elsewhere, e.g. in the settings tab). In this tab, you're able to see data from multiple environments, but to get to this feature, and subsequent tab, you have to select an environment. 

I'm not sure what the solution is other than a large overhaul of the UI to be able to view a feature in the context of the project, not a specific environment, but I'm just highlighting that it makes me feel a little uneasy.

kyle-ssg:
Yeah I think we could get a redesign concept around the hierarchy, however all the changes I can think of to this would involve making the product more difficult to navigate when creating a feature, adjusting it for an environment and updating any settings that belong to the project level. 

I guess we'd have a features page at the project level containing create feature, edit settings + analytics then have replace the environment level page something like "feature states".

I think regardless of this though, seeing analytics for all environments might be necessary to make the final decision as to whether you can delete it.

matthewelwell:
Yep, I agree, but is there a half step that we can take here perhaps? Like, the analytics _tab_ in the feature modal just shows you data for the specific environment, but it also has a link out to a project-level screen (maybe specifically for analytics, or maybe for the feature itself)?

tiagoapolo:
Can we have an environment selector in the analytics tab? It would be defaulted to the current environment but you could change the environment just for pulling analytics, it would not change the app selected environment

kyle-ssg:
@tiagoapolo that's what this pr does 

@matthewelwell hmm yeah I suppose we could have a feature analytics nav item in the project level. I'd actually be in favour of removing it from the feature modal if we did that.

kyle-ssg:
@matthewelwell added an alternative here https://github.com/Flagsmith/flagsmith/pull/5096

kyle-ssg:
Closed in favour of https://github.com/Flagsmith/flagsmith/pull/5100

PR #5028, opened 2025-01-21
fix: Raise if `environment_feature_version` is `None`

zachaysan:
## Changes

[Solves this Sentry alert](https://flagsmith.sentry.io/issues/6234460650/?alert_rule_id=14542360&alert_type=issue&environment=production&notification_uuid=4c43be3a-55a9-48a9-8374-cb8654767711&project=5544478&referrer=slack). Elsewhere in this part of the codebase we raise value errors to deal with bad input. Rather than erroring out with a type error that looks like a bug, we should raise an error of our own to deal with it.

One new test.

### DIFF
--- a/api/conftest.py
+++ b/api/conftest.py
@@ -478,7 +478,7 @@ def environment_v2_versioning(environment: Environment):
 
 
 @pytest.fixture()
-def identity(environment: Environment):
+def identity(environment: Environment) -> Identity:
     return Identity.objects.create(identifier="test_identity", environment=environment)
 
 
--- a/api/conftest.py
+++ b/api/conftest.py
@@ -471,14 +471,14 @@ def with_project_permissions(
 
 
 @pytest.fixture()
-def environment_v2_versioning(environment):
+def environment_v2_versioning(environment: Environment):
     enable_v2_versioning(environment.id)
     environment.refresh_from_db()
     return environment
 
 
 @pytest.fixture()
-def identity(environment):
+def identity(environment: Environment):
     return Identity.objects.create(identifier="test_identity", environment=environment)
 
 
--- a/api/conftest.py
+++ b/api/conftest.py
@@ -471,7 +471,7 @@ def with_project_permissions(
 
 
 @pytest.fixture()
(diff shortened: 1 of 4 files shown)


api/conftest.py
```diff
@@ -471,14 +471,14 @@ def _with_project_permissions(
 
 
 @pytest.fixture()
-def environment_v2_versioning(environment):
+def environment_v2_versioning(environment: Environment):
```
khvn26:
```suggestion
def environment_v2_versioning(environment: Environment) -> Environment:
```

api/conftest.py
```diff
@@ -471,14 +471,14 @@ def _with_project_permissions(
...
     enable_v2_versioning(environment.id)
     environment.refresh_from_db()
     return environment
 
 
 @pytest.fixture()
-def identity(environment):
+def identity(environment: Environment):
```
khvn26:
```suggestion
def identity(environment: Environment) -> Identity:
```

khvn26 (approved):

khvn26:
Do we have any idea how we reached the faulty state in the first place?

zachaysan:
> Do we have any idea how we reached the faulty state in the first place?

No but we did somehow.

api/tests/unit/users/test_unit_users_views.py
```diff
@@ -926,7 +926,13 @@ def test_list_user_groups(
...
+
+    for user in group_2_users:
+        if user["id"] == user1.pk:
+            assert user["group_admin"] is False
+            continue
+        if user["id"] == user2.pk:
+            assert user["group_admin"] is True
+            continue
+        # We should only match the two users.
+        assert False
```
khvn26: nit: I'm pretty sure using sets will solve this in a neater way:

```python
    assert set((user["id"], user["group_admin"]) for user in group_2_users) == {
        (user1.pk, False),
        (user2.pk, True),
    }
```

You won't have to maintain the uncovered `assert False` line too.

zachaysan: Oh this is a much better approach. Thank you!!

PR #5041, opened 2025-01-27
fix(webhook/signal): Optimise historical record writes 

### LINKED ISSUE #5020: Deleted feature throws FLAG_UPDATED event

- [ ] Self Hosted with Docker
- [x] SaaS at flagsmith.com

### Describe the bug

I'm using the webhook for create, update and delete features, 
I noticed that after the delete event which has `event_type: "FLAG_DELETED"`,
i get after another event of `event_type: "FLAG_UPDATED"` of the same event with the same `previous_state` and `new_state`. 
I should not get the updated event 

### Steps To Reproduce

1. In Environment setting page, Create Feature Webhook and connect it local server.
2. In Features page, create feature.
3. click on the 3 dots , on the new feature, click Remove feature, approve the removal.
4. a `FLAG_DELETED` event will be sent into the server, and immediately a `FLAG_UPDATED` will be sent as well, the `FLAG_UPDATED` will have the same previous and new state data.

### Expected behavior

Only the `FLAG_DELETED` event should be sent.

### Screenshots

_No response_

gagantrivedi:

## Changes

- Subclass `HistoricalRecords` to not write `~` historical records on soft delete
- Add `get_skip_create_audit_log`  to the signal method to avoid creating tasks that don't do anyting
- Add history to `SegmentRule`
- Update `django-lifecycle ` to support more complex hook conditions

Adds and updates unit tests


api/users/models.py
```diff
@@ -134,6 +135,18 @@ def schedule_hubspot_tracking(self) -> None:
                 ),
             )
 
+    @hook(AFTER_SAVE, condition=(WhenFieldHasChanged("email", has_changed=True)))
+    def send_warning_email(self):
+        from users.tasks import send_email_changed_notification_email
```
gagantrivedi, line 140: We can certainly get rid of this local import by doing some refactoring, but I will create a different pull request for that to avoid polluting this one

khvn26: I'm so used to importing tasks this way I regard it basically as a pattern now.

Globally, we could think on implementing a task dispatcher that accepts a string and does this under the hood so the task users don't have to.

khvn26 (approved):

api/users/models.py
gagantrivedi: Yeah, that's a good point

PR #5053, opened 2025-01-29
feat: Update identity override styles

### LINKED ISSUE #5051: UI showing flag states for an Identity with different overrides is confusing
Image

The above screenshot shows an identity with each variation of a flag state. In order, the flags are as follows: 

1. Flag **on** at environment level, no overrides
2. Flag **off** at environment level, no overrides
3. Flag **off** at environment level, **on** for segment override 
4. Flag **on** at environment level, **off** for segment override
5. Flag **off** at environment level, **on** for identity override
6. Flag **on** at environment level, **off** for identity overrides

These visuals aren't clear enough and some users are confused, particularly when looking at a segment override that sets the value to **off**. We should work on the UI here to make it clearer. 

Some thoughts / suggestions: 

 * Should we add some text to each toggle, e.g. 'ON' and 'OFF'? 
 * Should we add a colour coding to identity overrides? Or, should we remove the colour coding from segment overrides? 
 * Could we add more context closer to the toggle to indicate that there's an override in place? Some kind of icon next to each switcher (like the ones that we have showing that there are overrides in place on the environment UI? (see screenshot below)

Image

kyle-ssg:

## Changes

- Removes colour coded styles for identity/segment overrides
- Adds identity/segment override icons next to features that have overrides

image

As above


matthewelwell:
@kyle-ssg I was suggesting these icons could be used, but perhaps not exactly as they are in the environment UI. For example, the number will always just be 1, right? Also, I think it makes more sense in this context for it to be located nearer to the switch / value?

tiagoapolo (approved):

PR #5056, opened 2025-01-29
feat: Add self-hosted user data to version endpoint

matthewelwell:
## Changes

Add user data to version endpoint to support UI for self hosted onboarding workflow. 

Added unit tests


gagantrivedi (approved):

api/app/utils.py
```diff
@@ -48,21 +55,37 @@ def has_email_provider() -> bool:
...
         "ci_commit_sha": _get_file_contents("./CI_COMMIT_SHA"),
         "image_tag": UNKNOWN,
         "has_email_provider": has_email_provider(),
         "is_enterprise": is_enterprise(),
-        "is_saas": is_saas(),
+        "is_saas": _is_saas,
+        "self_hosted_data": None,
     }
 
+    _is_saas = is_saas()
```
khvn26: I think this line is not needed?

khvn26 (changes requested):

api/app/utils.py
gagantrivedi: Yep

api/app/utils.py
```diff
@@ -63,6 +72,18 @@ def get_version_info() -> VersionInfo:
         version_json["package_versions"] = manifest_versions
         version_json["image_tag"] = manifest_versions["."]
 
+    if not _is_saas:
+        from users.models import FFAdminUser
+
+        version_json["self_hosted_data"] = {
+            "has_users": FFAdminUser.objects.count() > 0,
+            "has_logins": FFAdminUser.objects.filter(last_login__isnull=False).exists(),
+            "is_bootstrapped": (
```
khvn26: For clarity and visibility, this endpoint is currently being used as liveness probe precisely because it does not perform any database queries.

Consider caching the results, or rethink the approach entirely (e.g., we could add a distinct `/version/self-hosted` endpoint, or use the upcoming prometheus integration).

CC @rolodato

gagantrivedi: Woah that should not happen

gagantrivedi: > Consider caching the results, or rethink the approach entirely (e.g., we could add a distinct /version/self-hosted endpoint, or use the upcoming prometheus integration).

IMO, I think we should add a different endpoint for the liveness check. The current implementation of relying on the version feels hacky to me.

khvn26: Agreed — can we do it as part of this PR or would you want this work in a separate one?

gagantrivedi: No, that should be a different pull request

khvn26: Opened https://github.com/Flagsmith/flagsmith/pull/5221.

api/app/utils.py
```diff
@@ -63,6 +72,18 @@ def get_version_info() -> VersionInfo:
...
+                settings.ALLOW_ADMIN_INITIATION_VIA_CLI is True
+                and FFAdminUser.objects.filter(email=settings.ADMIN_EMAIL).exists()
```
rolodato: Is checking for `ADMIN_EMAIL` reliable? It's not required when deploying with Docker Compose or Helm.

Maybe we could check for an active superadmin/staff user? Or store some data that indicates the bootstrap process was completed? Sorry, don't have much context to suggest something concrete, it's just `ADMIN_EMAIL` that caught my eye.

gagantrivedi: Is there a way to bootstrap without setting `ADMIN_EMAIL`?  I think that's through the UI, right?

rolodato: Yes, or by running `python manage.py createsuperuser` from a shell, or by signing up and making yourself a superuser from the database.

gagantrivedi: I don't think we really care about that. We are only trying to check the bootstrap method that we support, which will be deprecated (or made redundant) once the new API for onboarding is ready. @matthewelwell  is this correct?

gagantrivedi: Thinking about this a bit more, I don't think is_bootstrapped makes sense or is needed.

khvn26 (approved):

PR #5058, opened 2025-01-29
fix: Force environments to be different names

zachaysan:
## Changes

This forces new environments to have unique names. This was not intentionally done at the model level because some existing integrations rely on environment names and could break if they were updated to different names to solve existing conflicts.

One new test.

### DIFF
--- a/api/environments/views.py
+++ b/api/environments/views.py
@@ -140,6 +140,11 @@ class EnvironmentViewSet(viewsets.ModelViewSet):
         return queryset
 
     def perform_create(self, serializer):
+        existing_environment = Environment.objects.filter(
+            name=serializer.validated_data["name"]
+        )
+        if existing_environment:
+            raise ValidationError("Existing environment for given name.")
         environment = serializer.save()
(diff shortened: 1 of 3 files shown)


api/environments/views.py
```diff
@@ -140,6 +140,12 @@ def get_queryset(self):
         return queryset
 
     def perform_create(self, serializer):
+        existing_environment = Environment.objects.filter(
+            name=serializer.validated_data["name"],
+            project=serializer.validated_data["project"],
+        )
+        if existing_environment:
+            raise ValidationError("Existing environment for given name.")
```
matthewelwell: I think we can do this in a much simpler way using the serializer as per the PR I linked on the issue (see [here](https://github.com/Flagsmith/flagsmith/pull/5039))

matthewelwell (changes requested):

api/environments/views.py
zachaysan: Sorry I missed the linked PR! I've updated the code to handle this in a similar manner to what you posted.

matthewelwell (approved):

Issue #5009, opened 2025-01-16
Link from segment override / identity override icons no longer take user to the correct tab


matthewelwell:
Previously, clicking on the segment / identity override icons in a feature row would take you straight to the relevant tab in the feature slide out. This is no longer the case and needs fixing. 

See recording here:

https://github.com/user-attachments/assets/8fa68586-41d4-4c45-b6f5-bff4e44489c4

matthewelwell:
Resolved by https://github.com/Flagsmith/flagsmith/pull/5067

PR #5067, opened 2025-01-30
fix: link to segment and identity override

tiagoapolo:

Ref: #5009

## Changes

- Fix links to segment and identity overrides tab

Click on a segment override or identity override icon and check if the feature pane opens in the respective tab

https://github.com/user-attachments/assets/9fb1da57-3f61-40bd-ab29-a95161ceb499

### DIFF
--- a/frontend/web/components/FeatureRow.js
+++ b/frontend/web/components/FeatureRow.js
@@ -21,6 +21,17 @@ import { getTags } from 'common/services/useTag'
 import { getStore } from 'common/store'
 
 export const width = [200, 70, 55, 70, 450]
+
+export const TABS = {
+  ANALYTICS: 'analytics',
+  HISTORY: 'history',
+  IDENTITY_OVERRIDES: 'identity-overrides',
+  LINKS: 'links',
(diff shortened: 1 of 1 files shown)


frontend/web/components/FeatureRow.js
```diff
@@ -310,14 +310,22 @@ class TheComponent extends Component {
                 <SegmentOverridesIcon
                   onClick={(e) => {
                     e.stopPropagation()
-                    this.editFeature(projectFlag, environmentFlags[id], 1)
+                    this.editFeature(
+                      projectFlag,
+                      environmentFlags[id],
+                      'segment-overrides',
```
matthewelwell: I thought that we had enums for these tab values now?

tiagoapolo: We don't have enums for that one yet, I'll add some constants but probably would be more useful if that method was typed

tiagoapolo: It should be good to go for now

matthewelwell (approved):

PR #5079, opened 2025-02-06
fix: service to use default credentials mode when cookie auth disabled

blackjid:

## Changes

There was a problem when the cookie authentication is disabled and we expect some cookies to be send on the requests. This was an issue when using Cloudflare Access, where the authentication sets the `CF_Authentication` cookie that is needed. The refactor add `credentials: omit` to the fetch method, which cause all requests to fail before reaching the flagsmith api.

The change of behaviour was introduced [here](https://github.com/Flagsmith/flagsmith/compare/v2.143.0...v2.144.0#diff-4ae7df25ecd9cba5842dad530754eb9d27309186f9abc82677695d317ca05716R19)

To fix it the service that call api endpoints from the frontend now uses the default value for `credentials` when the cookie authentication is disabled. This is the same behaviour it used to have before the cookie authentication refactor.

I test it manually using the content override feature of the chrome developer tools.

### DIFF
--- a/frontend/common/service.ts
+++ b/frontend/common/service.ts
@@ -16,7 +16,7 @@ export const baseApiOptions = (queryArgs?: Partial<FetchBaseQueryArgs>) => {
     | 'extractRehydrationInfo'
   > = {
     baseQuery: fetchBaseQuery({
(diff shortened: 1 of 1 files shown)


matthewelwell:
Thanks @blackjid , one of the team will take a look at this and get back to you soon.

tiagoapolo (approved):

PR #5080, opened 2025-02-06
fix: delete identity overrides from environments_v2 table on feature delete

matthewelwell:
## Changes

Adds logic to delete identity override documents from flagsmith_environments_v2 table when a feature is deleted. 

Added an end-to-end style test to delete a feature from the view. 

Note: the test isn't passing and I can't understand why - I can see that the code is being hit in the task, but the item is still in the table when we query it in the test code

### DIFF
--- a/api/edge_api/identities/tasks.py
+++ b/api/edge_api/identities/tasks.py
@@ -162,7 +162,7 @@ def update_flagsmith_environments_v2_identity_overrides(
 
 
 @register_task_handler()
-def delete_environments_v2_identity_overrides_from_for_feature(feature_id: int) -> None:
+def delete_environments_v2_identity_overrides_by_feature(feature_id: int) -> None:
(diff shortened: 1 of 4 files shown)


api/edge_api/identities/tasks.py
```diff
@@ -159,3 +159,14 @@ def update_flagsmith_environments_v2_identity_overrides(
         identifier=identifier,
     )
     dynamodb_wrapper_v2.update_identity_overrides(identity_override_changeset)
+
+
+@register_task_handler()
+def delete_environments_v2_identity_overrides_from_for_feature(feature_id: int) -> None:
```
khvn26:
```suggestion
def delete_environments_v2_identity_overrides_by_feature(feature_id: int) -> None:
```

khvn26 (approved):

PR #5082, opened 2025-02-07
fix: format-create-segment-errors-from-payload

### LINKED ISSUE #5071: Bad error message rendering when failing to update segment

- [x] Self Hosted with Docker
- [x] SaaS at flagsmith.com

### Describe the bug

See reproduction steps.

### Steps To Reproduce

Create a segment rule with over 1000 bytes - here is a random 1001 byte string:

```
1W086HAAm5j95zXtexaJwGOVsBaxvg1+/dDxkkOsdZIN294cvkgH/+QbuYtcuI4Hg3d13NhAcJZjZQjBhs/CMNEWnAogRfQMZKpRBxU2+2sboq8SjafonDxw+Cc0JWmPK2fraQ6CyemmgkEGjF5WuiJZP6VCUTqLHin3OIilYFraLfwnS9GaMYg/KIl3KjS21GuMxeoxezOTCKbjDW2MWeUUUDDuMKQZVHC0FX/QJ+lXtBjzQ/A3IwJIVm+y8k1LsQqdgG+3QgYAiG4RAIVqXjhtcSs4l9zq5dJiOIHoGOH+J6mGI7MnzVaGH1uGHJH71KbHozlT3wztAvS/lL9tttzD/SmaK29k4+jbAhrXavbD99NJtm+GvWq+vx3sJ21nZcunyuE+u0NPqQL0+NCzDFIyhVuEezirPpvnWXRMbUE9It1lxG4rrqfy7twOLnVEV04B7W938gSDbOF42kpFyXc/LROu3bukUBMEfCSukZtLvzQBJPYVn3WH/6QATPKO/1rSgT7CsKD4sJYzyRfiejk1YB7MwOKKZGLnUk+ixTrRbFY+Lypf+C9fSuL0pdMKLBqXZ30eaVjIwe1dzZE6tuke7ZJqVKTjd73YvbHNn+MFCD6uzAHGDBJv/5IL4ImPLTYFaPMntYRcIjpTUugIG65x2vE5rT20Vmw1MjnY08eCsEjd79HjggqfLnHUs2ujNrDHv606PSnQJNLw9f2bskIR60YkdvM0xepVkgIHftN7uCSvWRYRhIVEimnC9+UVY7h3Mjdxzgvr8JQgyRrbc370G+mH0xICYrcI/fas5fdyd01r60nr3H4YGVmqo/UqlTDk8HaQUFe5VYKGv5RODPL6tz61AYM48T79XhXzbEkbQpAQDHncDvJFo+C0yP8nHGoBHXPYQk9hkRWyMpYaGgHkZgClRHV+/gZBZ3MVlvn9acXv/ojz65Mk2hwkhTWk0uEO9eiotG+aYH3aSI33BX1izJ9hhz2W6c6/fB6R1
```

The API error response is:

```json
{
  "rules": [
    {
      "rules": [
        {
          "conditions": [
            {
              "value": [
                "Ensure this field has no more than 1000 characters."
              ]
            }
          ]
        }
      ]
    }
  ]
}
```

This is rendered as `[object Object]` like this:

Image

### Expected behavior

Render the correct error message.

### Screenshots

_No response_

Zaimwa9:

## Changes
Issue: https://github.com/Flagsmith/flagsmith/issues/5071
Fixes #5071 

_Please describe._
The payload was passed as of from the request response. 
- Added typing of error
- Implemented a formatError method in `CreateSegment`
- Prefixed the error with `value/property`
- If single error, only returns a string (to keep `Rules: XXXXX`), if multiple errors, returns an array that is natively handled by the error component (but displays them as indexed list)

⚠️  This is more a hotfix than a permanent solution. The payload seems to be redundant with rules, cf type being, could be fixed in the backend:
```
type CreateSegmentError = {
  status: number,
  data: {
    rules: [{
      rules: Array<{
        conditions: SegmentConditionsError[]
      }>
    }]
  }
}
```

Also please let me know if you have a better solution to handle multiple errors, the best would be something like the following but maybe you have other plans/prio:
```
Rule 1:
 - value XXX
Rule 4: 
 - property: YYY
 - value: XXX
```

**Single error case**
image

**Multiple error case (to be improved)**
image

_Please describe._

### DIFF
--- a/frontend/env/project_dev.js
+++ b/frontend/env/project_dev.js
@@ -1,12 +1,12 @@
 const globalThis = typeof window === 'undefined' ? global : window
 module.exports = global.Project = {
-  api: 'https://api-staging.flagsmith.com/api/v1/',
+  api: 'http://localhost:8000/api/v1/',
 
   chargebee: {
     site: 'flagsmith-test',
   },
 
-  env: 'staging',
+  env: 'dev',
 
   // This is our Bullet Train API key - Bullet Train runs on Bullet Train!
   flagsmith: 'ENktaJnfLVbLifybz34JmX',
--- a/frontend/env/project_dev.js
+++ b/frontend/env/project_dev.js
@@ -1,12 +1,12 @@
 const globalThis = typeof window === 'undefined' ? global : window
 module.exports = global.Project = {
-  api: 'http://localhost:8000/api/v1/',
+  api: 'https://api-staging.flagsmith.com/api/v1/',
 
   chargebee: {
     site: 'flagsmith-test',
   },
 
-  env: 'dev',
+  env: 'staging',
 
   // This is our Bullet Train API key - Bullet Train runs on Bullet Train!
   flagsmith: 'ENktaJnfLVbLifybz34JmX',
--- a/frontend/web/components/modals/CreateSegment.tsx
+++ b/frontend/web/components/modals/CreateSegment.tsx
@@ -17,6 +17,7 @@ import {
   Operator,
   Segment,
   SegmentRule,
+  SegmentConditionsError,
 } from 'common/types/responses'
 import { Req } from 'common/types/requests'
 import { useGetIdentitiesQuery } from 'common/services/useIdentity'
@@ -73,6 +74,16 @@ type CreateSegmentType = {
   onComplete?: (segment: Segment) => void
   readOnly?: boolean
   segment?: Segment
+}
+type CreateSegmentError = {
+  status: number,
+  data: {
+    rules: [{
+      rules: Array<{
+        conditions: SegmentConditionsError[]
+      }>
+    }]
+  }
 }
 
 enum UserTabs {
@@ -160,7 +171,7 @@ const CreateSegment: FC<CreateSegmentType> = ({
   const [tab, setTab] = useState(UserTabs.RULES)
   const [metadata, setMetadata] = useState<Metadata[]>(segment.metadata)
   const metadataEnable = Utils.getPlansPermission('METADATA')
-  const error = createError || updateError
+  const error: CreateSegmentError = createError || updateError
   const totalSegments = ProjectStore.getTotalSegments() ?? 0
   const maxSegmentsAllowed = ProjectStore.getMaxSegmentsAllowed() ?? 0
   const isLimitReached = totalSegments >= maxSegmentsAllowed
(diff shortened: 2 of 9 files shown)


frontend/web/components/modals/CreateSegment.tsx
```diff
@@ -430,6 +441,26 @@ const CreateSegment: FC<CreateSegmentType> = ({
     </FormGroup>
   )
 
+  const formatError = (error: CreateSegmentError): string | string[] => {
```
kyle-ssg: Hmm, if we're receiving the errors in the same order as the current rules, I wonder if it's better to highlight the inputs with a tooltip of the error and have a generic error underneath.

If this is not feasible we should at least indicate which rule it is

kyle-ssg: It looks like we'd be able to infer this from the api, non-error conditions seem to just have {}
image

Zaimwa9: Capture d’écran 2025-02-12 à 15 46 38
Capture d’écran 2025-02-12 à 15 45 31

@kyle-ssg, wdyt ?

tiagoapolo: That way the UI looks a bit more lean o but IMO having the error message next to the input is more direct for the user to validate what is wrong and fix it. like this

PS.: That's just a mockup so consider that the input field should be highlighted when there's an error
Screenshot 2025-02-13 at 11 11 20

Zaimwa9: Alright, I have limited access to my computer for the next weeks, i'll rework it this way when back

Zaimwa9:
@tiagoapolo Here is the version based on your mockup 👍 
I didn't change the legacy behavior in this PR (clicking away without value from Name/Description/Value highlights the inputs) and the errors are purely based on the api payload

image

## Changes
- Errors are displayed at rule level
- Removed `ErrorMessage` and `errors` props in CreateSegmentRulesTabForm 
- Low hanging fruit: recentered unread badge css

I think next step if ever it becomes prio would be to clean up errors on edition, but from what I see it might require some form validators and refactoring (including the inputs)

## QA
Success on all type of operators + update with failing + update success
 
https://www.loom.com/share/99157d3cf7094113808e97727e54bc25

frontend/env/project_dev.js
```diff
@@ -1,12 +1,12 @@
 const globalThis = typeof window === 'undefined' ? global : window
 module.exports = global.Project = {
```
kyle-ssg: To revert

kyle-ssg (approved):

kyle-ssg:
Ok tested and all looks great!

PR #5084, opened 2025-02-07
docs: correct node cache example

hammerOwen:

## Changes

The cache implementation for node needs to return a Flags object not a JSON object.

No testing this is only documentation update

### DIFF
--- a/docs/docs/clients/server-side.mdx
+++ b/docs/docs/clients/server-side.mdx
@@ -1892,8 +1892,8 @@ interface.
 For example, this cache implementation uses Redis as a backing store:
 
 ```typescript
-import { Flagsmith } from 'flagsmith-nodejs';
-import type { BaseOfflineHandler, EnvironmentModel, Flags, FlagsmithCache } from 'flagsmith-nodejs';
+import { Flagsmith, Flags } from 'flagsmith-nodejs';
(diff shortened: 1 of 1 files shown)


rolodato (changes requested):
Hi @hammerOwen, thanks for sending this over. You're right that this example is not correct, but I think neither is your suggested change:

image

There are several things wrong with the original example, which I'm surprised we didn't notice before. This looks like a correct example to me:

```js
import { Flagsmith, Flags } from 'flagsmith-nodejs';
import type { BaseOfflineHandler, EnvironmentModel, FlagsmithCache } from 'flagsmith-nodejs';
import * as redis from 'redis';

const redisClient = redis.createClient({
    url: 'localhost:6379',
});

const redisFlagsmithCache = {
    async get(key: string): Promise<Flags | undefined> {
        const cachedValue = await redisClient.get(key);
        if (cachedValue) {
            return new Flags(JSON.parse(cachedValue));
        }
    },
    async set(key: string, value: Flags): Promise<void> {
        await redisClient.set(key, JSON.stringify(value), { EX: 60 });
    },
} satisfies FlagsmithCache;
```

Another change is that we're now using `import` instead of `import type` for `Flags`, since you can't instantiate classes imported with `import type`.

Let me know if that looks correct to you as well, and if it does add it to this PR and we'll get it merged.

hammerOwen:
Updated the example with your corrections.

rolodato (approved):

PR #5096, opened 2025-02-12
feat: feature analytics

kyle-ssg:

## Changes

Alternate to https://github.com/Flagsmith/flagsmith/pull/5026, moves feature analytics to separate nav. This may end up being a stop gap to where we have a project level "Features" navigation for create,settings and analytics vs the environment feature states.

- Cleans up feature actions
- Migrates the feature select to rtk / typescript
- Moves feature usage to RTK + TypeScript
- Creates reusable environment tag filter component
- Removes analytics tab from feature modal

image
image

Viewed feature analytics in  s


matthewelwell (commented):
This mostly looks good to me but I've added some thoughts / findings below. 

1. Should we keep the 'analytics' tab for now at least, perhaps with a note saying that it has moved? We could even add a link to the new screen, with a query parameter that pre-selects the correct feature. I suspect that @rolodato will advocate for something along these lines to avoid unnecessary support. 
2. I noticed an error in the tooltip that we're displaying when hovering over the data in the graph. Looks like we're showing the environment id instead of the environment name? 

image

3. Super minor, but I'd suggest that we show an empty graph when all environments are deselected, instead of no graph at all. 

image

4. I have a feeling that this is covered elsewhere, alongside some other functionality, but adding a 'clear all' option here I feel will be important. Without it, we're not only moving the functionality but actually making it harder for them to see the data they had previously, particularly for projects that have lots of environments, they might find it important to see the view of a single environment only.

matthewelwell:
5. It looks like we're also showing archived features in the list? Perhaps we should exclude those unless explicitly requested, like we do in the main features list in each environment?

kyle-ssg:
Moved to https://github.com/Flagsmith/flagsmith/pull/5096, pr was coupled with segment change request changes

PR #5098, opened 2025-02-13
feat: Grafana feature health provider

khvn26:

## Changes

This PR implements the Grafana (or Prometheus) health provider. Additionally, the feature health framework is refactored:
 - Feature health events can be associated with an external ID. When retrieving events for UI, latest events are now selected per project per feature per environment per provider per external ID. This is done to allow tracking multiple different alerts targeting the same feature.
 - `FeatureHealthEvent.created_at` can now be set to arbitrary datetime. This is done so we can use datetimes provided by alertmanager.
 - `FeatureHealthEvent.reason` is now a `JSONField` with predetermined schema common for all providers. This enables supplying arbitrary number of text and URL data with the feature health event.
 - Feature health providers are now allowed to produce multiple feature health events from one webhook call.
 - A service layer is added to individual providers to potentially allow for more than simple mapping logic.

Added integration tests:
 - For a basic scenario.
 - For a case when different alerts hit the same webhook.


api/features/feature_health/providers/grafana/types.py
```diff
@@ -0,0 +1,26 @@
...
+
+from pydantic import BaseModel
+
+
+class GrafanaFeatureHealthEventReason(typing.TypedDict):
+    alert_name: str
+    generator_url: str
+    description: typing.NotRequired[str]
+    runbook_url: typing.NotRequired[str]
+    summary: typing.NotRequired[str]
```
tiagoapolo: Could we maybe have a BaseFeatureHealthEventReason that would be reused for other providers?

khvn26: All of the fields here are pretty specific to Alertmanager. Do you want us to figure out a reason card layout with common elements?

tiagoapolo: Got it, maybe could we have `description` and `url` (a link to the provider issue)  fields as common fields for every provider?

khvn26: As discussed offline, we now use a common schema for all providers. See [here](https://github.com/Flagsmith/flagsmith/blob/6c121bc188410139a7f751082d69466977669a49/api/features/feature_health/types.py#L10-L22).

matthewelwell:
> FeatureHealthEvent.reason is now a JSONField with predetermined schema common for all providers. This enables supplying arbitrary number of text and URL data with the feature health event.

We might need to consider a couple of things here: 

 1. We'll need to remember to replace this with `OracleSafeJSONField` in the ee repo as per [this](https://github.com/Flagsmith/flagsmith-ee/blob/main/api/core/fields.py#L5). 
 2. We should ideally make sure that this field isn't needed in the 'list' endpoint as I believe we'll have to `defer` it in the queryset, resulting in an N+1 problem.

api/Makefile
```diff
@@ -73,12 +73,21 @@ django-make-migrations:
 	poetry run python manage.py waitfordb
 	poetry run python manage.py makemigrations $(opts)
 
+.PHONY: django-squash-migrations
+django-squash-migrations:
+	poetry run python manage.py waitfordb
+	poetry run python manage.py squashmigrations $(opts)
```
matthewelwell, line 83: Why was this added here?

api/features/feature_health/providers/grafana/mappers.py
```diff
@@ -0,0 +1,83 @@
...
+    alert_instance: GrafanaAlertInstance,
+) -> FeatureHealthEventData | None:
+    labels = alert_instance.labels
+    if feature_name := labels.get(GRAFANA_FEATURE_LABEL_NAME):
+        if alert_instance.status == "firing":
+            created_at = alert_instance.startsAt
+            event_type = FeatureHealthEventType.UNHEALTHY
+        else:
+            created_at = alert_instance.endsAt
+            event_type = FeatureHealthEventType.HEALTHY
```
matthewelwell, line 73: It's a little odd that we directly set the event_type to the enum object here, but later for the provider we have to reference the string value. Can we make this consistent?

docs/docs/advanced-use/feature-health.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Sample Provider
+
+We provide a Sample Provider for your custom integrations. To create a Sample Feature Health webhook:
+
+1. Go to Project Settings > Feature Health.
+2. Select "Sample" from the Provider Name drop-down menu.
+3. Click Create and copy the Webhook URL.
+
+You can use the webhook in your custom integration. Refer to the payload schema below:
```
matthewelwell, line 29: I wonder if we should actually add this schema as a separate file to the codebase and link to it, so people can point to it if they want to, for generation / validation purposes?

api/Makefile
khvn26: I'm used to interacting with Django environment through make. The Makefile loads the appropriate environment for me. During this PR development, I added some new migrations, but I only want to merge one, so I'm squashing them. See https://github.com/Flagsmith/flagsmith/blob/6c121bc188410139a7f751082d69466977669a49/api/features/feature_health/migrations/0002_alter_featurehealthevent_external_id_and_more_squashed_0003_alter_featurehealthevent_reason_and_more.py#L9

api/features/feature_health/providers/grafana/mappers.py
khvn26: Fixed in 47319c5ffd60a3040e85b9203d37d5cf489cda8a

api/Makefile
matthewelwell: If they were both added here though, why not delete them both, and just recreate a single migration?

docs/docs/advanced-use/feature-health.md
khvn26: Nice idea. I guess we could embed it in docs using something like https://github.com/peterrhodesdev/react-embed-github-file-content too.

api/Makefile
khvn26: Because I care about other engineers that play with the PR's Docker images against their local databases. With squashed migrations,  they won't have to recreate their schemas.

matthewelwell: Hmm... I don't think we should, but ok.

docs/docs/advanced-use/feature-health.md
khvn26: Okay, I've added it as a file, we'll be able to embed it once this PR is merged.

matthewelwell:
> 2. We should ideally make sure that this field isn't needed in the 'list' endpoint as I believe we'll have to defer it in the queryset, resulting in an N+1 problem.

@khvn26 this is an example of the logic I'm referring to with my comment here

https://github.com/Flagsmith/flagsmith/blob/main/api/permissions/permission_service.py#L147-L152

api/features/feature_health/migrations/0002_alter_featurehealthevent_external_id_and_more.py
matthewelwell, line 1: Somewhat related to our ongoing discussion around the squashed migration, I'm not particularly happy with the naming of this migration (and 0003 for that matter). 

We should ensure that all migrations are manually named - imo `alter_featurehealthevent_external_id_and_more` is not a good enough name to identify what the migration is doing at a quick glance. 

I think therefore, we'll have to break migrations for anyone that used the image locally because these migration names will have to change regardless.

khvn26: I've removed the squashed migrations in 7ffe3b6.

api/Makefile
khvn26: This thread is now irrelevant as per 7ffe3b6.

khvn26:
@matthewelwell Reverted to TextField to avoid `JSONField` usage, and removed the squashed migrations.

matthewelwell (approved):

tiagoapolo (approved):

PR #5100, opened 2025-02-13
feat: Move feature analytics to project level

kyle-ssg:

## Changes

Supersedes https://github.com/Flagsmith/flagsmith/pull/5096

Moves feature analytics to separate nav. This may end up being a stop gap to where we have a project level "Features" navigation for create, settings and analytics vs the environment feature states.

- Cleans up feature actions
- Migrates the feature select to rtk / typescript
- Moves feature usage to RTK + TypeScript
- Creates reusable environment tag filter component
- Removes analytics tab from feature modal

image
image

Viewed feature analytics in production


kyle-ssg:
Moving comments from Matt here:

This mostly looks good to me but I've added some thoughts / findings below. 

1. Should we keep the 'analytics' tab for now at least, perhaps with a note saying that it has moved? We could even add a link to the new screen, with a query parameter that pre-selects the correct feature. I suspect that @rolodato will advocate for something along these lines to avoid unnecessary support. 
2. I noticed an error in the tooltip that we're displaying when hovering over the data in the graph. Looks like we're showing the environment id instead of the environment name? 

image

3. Super minor, but I'd suggest that we show an empty graph when all environments are deselected, instead of no graph at all. 

image

4. I have a feeling that this is covered elsewhere, alongside some other functionality, but adding a 'clear all' option here I feel will be important. Without it, we're not only moving the functionality but actually making it harder for them to see the data they had previously, particularly for projects that have lots of environments, they might find it important to see the view of a single environment only. 
5. It looks like we're also showing archived features in the list? Perhaps we should exclude those unless explicitly requested, like we do in the main features list in each environment?

kyle-ssg:
1 - I'm not sure, let's have a discussion about this since one of our overall aims was to remove this confusion of project level feature things from environment feature value config
2 - Hmm yeah I imagine this was a pre-existing issue
3 - Yeah that would make sense
4 - Yeah a clear all would make sense, the feature actions menu also links to this page with the environment selected
5 - Yes, we use this feature select throughout the app and I wonder if we can add more filters to the menu e.g. tags

rolodato:
> Should we keep the 'analytics' tab for now at least, perhaps with a note saying that it has moved? We could even add a link to the new screen, with a query parameter that pre-selects the correct feature. I suspect that @rolodato will advocate for something along these lines to avoid unnecessary support.

Instead of keeping the current Analytics tab, I would prefer to have another way of accessing what's currently only inside the "⋮" dropdown button from the feature details modal as well. We already do this for "Copy feature name", which can be done by clicking the button next to the feature name, whether that's in the features list or a feature's details.

IMO it's instinctive to click on a feature to do anything with it (like seeing analytics, history etc) and it's frustrating to have to remember that certain things are accessed from inside the modal, and others are from the features list dropdowns.

I don't have a UX suggestion for how to achieve this (talk is cheap, I know).

kyle-ssg:
> IMO it's instinctive to click on a feature to do anything with it (like seeing analytics, history etc) and it's frustrating to have to remember that certain things are accessed from inside the modal, and others are from the features list dropdowns.

This train of thought is why I think keeping everything in the same place has the fewest downsides (i.e. [this](https://github.com/Flagsmith/flagsmith/pull/5026)).

 The other options:
- Move feature creation, settings and analytics to separate nav to environment level settings

As you said, this means people are remembering its in a different page to where they are viewing the values of features.

-  Move these but keep a reference within the feature modal. 

This means everything is repeated, we introduce bloat, people will unnecessarily be opening another link from the feature modal to do what they need.

PR #5107, opened 2025-02-14
chore: enforce `--name/-n` argument in makemigrations management command

matthewelwell:
## Changes

Enforces the use of `--name/-n` when creating migrations. 

Ran `makemigrations` with and without `--name/-n` and verified the correct result.

### DIFF
--- a/api/core/management/commands/makemigrations.py
+++ b/api/core/management/commands/makemigrations.py
@@ -4,7 +4,7 @@ from django.core.management.commands.makemigrations import (
 )
 
 
-class Command(MakeMigrationsCommand):
+class Command(MakeMigrationsCommand):  # pragma: no cover
     """
     Customise the makemigrations command to enforce use of `--name/-n` argument.
     """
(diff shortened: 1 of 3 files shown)


khvn26:
Looks like this breaks the `--check` check in CI.

matthewelwell:
> Allow dry-run / check

Yeah, fixed in https://github.com/Flagsmith/flagsmith/pull/5107/commits/2689d5e773b9c06d3740ccc6e6e15f844bf08f80

api/core/management/commands/makemigrations.py
```diff
@@ -0,0 +1,18 @@
+from django.core.management import CommandError
+from django.core.management.commands.makemigrations import (
+    Command as MakeMigrationsCommand,
+)
+
+
+class Command(MakeMigrationsCommand):  # pragma: no cover
```
khvn26: You have to move that `# pragma: no cover` to the method, looks like.

matthewelwell: I don't love that I've ended up having to add an omit in pyproject.toml... I think I'll just write a darn test.

khvn26 (approved):

PR #5108, opened 2025-02-14
feat: enable feature versioning for new environments

matthewelwell:
## Changes

Enable feature versioning for environments in newly created projects (based on a flag). 

Added unit tests.


api/integrations/flagsmith/client.py
```diff
@@ -25,13 +25,15 @@
...
-    try:
-        _flagsmith_client = _flagsmith_clients[name]
-    except (KeyError, TypeError):
-        kwargs = _get_client_kwargs()
-        kwargs["enable_local_evaluation"] = local_eval
-        _flagsmith_client = Flagsmith(**kwargs)
-        _flagsmith_clients[name] = _flagsmith_client
+    if settings.USE_GLOBAL_FLAGSMITH_CLIENTS and (
+        client := _flagsmith_clients.get(name)
+    ):
```
khvn26: Why _not_ use global Flagsmith clients? If it's just for tests, I'd prefer a fixture mocking them than a Django setting. This will also solve the coverage issue.

api/tests/unit/conftest.py
```diff
@@ -232,3 +240,53 @@ def populate_environment_document_cache(
...
+                        id=1,
+                        name="flagsmith-organisation",
+                        feature_analytics=False,
+                        stop_serving_flags=False,
+                        persist_trait_data=False,
+                    ),
+                    hide_disabled_flags=False,
+                ),
+                feature_states=[],
+            )
```
khvn26: Is this really necessary to build an offline handler? Looks really clunky (and we should do more than just https://github.com/Flagsmith/flagsmith-engine/pull/213 to fix that, IMO).

api/tests/unit/conftest.py
```diff
@@ -232,3 +240,53 @@ def populate_environment_document_cache(
     persistent_environment_document_cache.get.return_value = (
         map_environment_to_environment_document(environment)
     )
+
+
+@pytest.fixture()
+def set_flagsmith_client_flags(
+    mocker: MockerFixture,
+) -> typing.Callable[[list[TestFlagData]], None]:
```
khvn26: Note: I'm curious to see this fixture used and evolved. IMO we can include it as a pytest plugin along with the Python client when it's mature enough.

api/environments/models.py
```diff
@@ -195,6 +198,22 @@ def delete_environment_document_from_cache(self) -> None:
...
+        organisation = self.project.organisation
+        flag = flagsmith_client.get_identity_flags(
+            organisation.flagsmith_identifier,
+            traits={"organisation_id": organisation.id},
+        ).get_flag("enable_feature_versioning_for_new_projects")
+
+        if (
+            flag.enabled
+            and self.project.created_date.date()
+            >= datetime.date.fromisoformat(str(flag.value))
```
khvn26: 1. I don't think I clearly understand the reason to grandfather in existing projects vs. just scheduling a change request for this flag. What are the implications of having versioned and non-versioned environments in the same project? If we want to avoid it, why not have a project-level setting to enable v2 versioning?
2. The flag value usage here is questionable. Enabling the flag without setting a correct value will basically break Flagsmith. Can we not make FoF more fragile please?

Side note: to me, this code is a good example of current limitations in Flagsmith's segmenting; if the engine supported datetime comparisons, this could've been 

```python
        flag = flagsmith_client.get_identity_flags(
            organisation.flagsmith_identifier,
            traits={
                "organisation_id": organisation.id,
                "project_created_date": {
                    "value": self.project.created_date,
                    "transient": True,
                 }
            },
        ).get_flag("enable_feature_versioning_for_new_projects")
```

or, if we had multi-context support, an even better

```python
        flag = flagsmith_client.get_identity_flags(
            organisation.flagsmith_identifier,
            contexts={
                "default": {"organisation_id": organisation.id},
                "project": {
                    "created_date": self.project.created_date,
                },
            },
        ).get_flag("enable_feature_versioning_for_new_projects")
```

— both of the above are much safer and clearer than the code I'm reviewing.

api/tests/unit/environments/test_unit_environments_models.py
```diff
@@ -1208,3 +1209,66 @@ def test_environment_metric_query_helpers_match_expected_counts(
     assert change_request_count_result == change_request_count
     assert scheduled_count_result == scheduled_change_count
     assert identity_override_count == 0
+
```
khvn26, line 1211: I guess I'd like to see

a) a test with project with `created_date` < flag value
b) a test with an empty/non-date flag value

khvn26 (changes requested):
Left some comments.

api/integrations/flagsmith/client.py
matthewelwell: Yeah, TBH, I have no idea why this logic exists. I wrote it far too long ago to remember the context. When I came back to it yesterday, I also hated it :)

api/environments/models.py
matthewelwell: Yeah, I couldn't really see a reason for only enabling this based on the project creation, so I've simplified it to just checking if it should be enabled for new environments

api/integrations/flagsmith/client.py
matthewelwell: So, in revisiting this, the reason for this being added as part of this PR is that we are now accessing the flagsmith client as part of the `Environment` model creation lifecycle. As such, mocking in a fixture is not sufficient since (as far as I'm aware), we cannot control the order in which fixtures are loaded so when e.g. the `environment` fixture is used, the flagsmith client is already instantiated. 

I'm still looking into this, but just adding this comment here for documentation and possible outside opinion / help.

khvn26: Since flagsmith is called when the model is created, you're supposed to test model creation, right? It makes sense to instantiate `Environment` manually in the test body instead of relying on a fixture.

matthewelwell: Yeah, good point - I'm not sure why my test [here](https://github.com/Flagsmith/flagsmith/pull/5108/files#diff-20538f1a0b1c78b087d26dfe7f43e2e8f7485071b1e47eb31c77044a56f22fa8R1216) was loading the environment fixture. I've removed that (and the unexpected reference to the environment fixture on the feature fixture [here](https://github.com/Flagsmith/flagsmith/blob/main/api/conftest.py#L571)). 

I'm still not getting the mocking behaviour I'm expecting, but I think I can move past it. Let's see.

api/tests/unit/environments/test_unit_environments_models.py
matthewelwell: I've removed the date logic entirely so I'll resolve this comment.

api/tests/unit/conftest.py
matthewelwell: I think this fixture needs more thought, and as you mentioned, it should be included as a pytest plugin and live in the flagsmith-python-client repository. 

I've created an issue for this [here](https://github.com/Flagsmith/flagsmith-python-client/issues/135) as I think it lives outside the scope of this PR.

api/tests/unit/conftest.py
matthewelwell: This code was removed in favour of looking into a better solution as per the issue [here](https://github.com/Flagsmith/flagsmith-python-client/issues/135).

api/integrations/flagsmith/client.py
matthewelwell: As mentioned in other comments, I've completely simplified the mocking logic here to focus on the actual code paths that I care about. 

I'd love to create a more generic solution as discussed elsewhere, but in order to move this PR forwards, I have descoped it and implemented the simple solution.

api/tests/unit/environments/conftest.py
```diff
@@ -12,3 +13,33 @@ def mock_dynamo_env_wrapper(mocker: MockerFixture) -> Mock:
...
+    mocker: MockerFixture,
+) -> typing.Callable[[], None]:
+    """
+    This fixture returns a callable that allows us to enable the flag such that new environments
+    are created with feature versioning v2 enabled. It returns a callable to ensure that we can
+    control when the flag is enabled in the test code, to ensure that other fixtures are unaffected.
+
+    TODO: How can we make this more generic so that we don't need to mock individual calling
+     locations and instead we can have a fixture that somehow mocks a given flag or set of flags.
+     This should probably be a part of the flagsmith-python-client repository itself.
```
khvn26: Add the [issue URL](https://github.com/Flagsmith/flagsmith-python-client/issues/135) please

matthewelwell: I wonder if I should just remove the TODO altogether now that we have the issue?

api/environments/models.py
```diff
@@ -195,6 +197,24 @@ def delete_environment_document_from_cache(self) -> None:
         ):
             environment_document_cache.delete(self.api_key)
 
+    @hook(BEFORE_CREATE)  # type: ignore[misc]
+    def enable_v2_versioning(self) -> None:
+        if self.enable_v2_versioning:
```
khvn26: Must be a typo?
```suggestion
        if self.use_v2_feature_versioning:
```

matthewelwell: Ugh, hoisted by my own petard yet again

api/tests/unit/environments/conftest.py
matthewelwell: I've removed the TODO, but still left a reference to the issue.

khvn26: Well, the TODO is to switch to the fixture once we have it in place?

matthewelwell: Yes, true, but I think that the description on the issue (which references this PR) is sufficient to point us towards the code that needs changing. Not a hill I'll die on though so I can re-add the TODO if you'd prefer.

matthewelwell:
Ok, I think this PR is ready for review now @khvn26 . 

I've made a few simplifications and I think it's a lot clearer now what the purpose of the PR is. I've removed some of the code which was resolving unrelated bad test code to [another PR](https://github.com/Flagsmith/flagsmith/pull/5723), and created a [separate issue](https://github.com/Flagsmith/flagsmith-python-client/issues/135) in the flagsmith-python-client repository to handle improving our testing capabilities.

khvn26 (approved):
Nice 👌

PR #5111, opened 2025-02-14
ci: Add incremental Python type checking

khvn26:

## Changes

In this PR, I refined the idea first realised in #4863 with a number of improvements.

Pre-commit's idea of tool caching is not really compatible with Mypy, so the hook is re-written with `language: system` to rely on Poetry's virtual environment. The hook is skipped in pre-commit.ci. In CI, the script is more appropriately run in the unit test pipeline, along with the migration check.

The Mypy output is set to be in JSON format, which makes issue comparison more reliable. The file paths are normalised regardless of how the script is run. The issues are compared based on line/column numbers and error codes; messages and hints, shown to be inconsistent across Mypy runs, are disregarded during the baseline comparison, but still presented in stdout.

When running against all files by running `make typecheck` locally or in CI, the baseline is verified for stale issues.  

The tests are now properly checked — the issue preventing that from happening is fixed in 154ea5ca7291a3915a43f01261ce5d734f4a39ab, which took me embarrassingly long to figure out.

## Notes

- Mypy skips too many issues when fed individual files. It's still possible to use the script for increments, but for the time being I set `pass_filenames: false` in pre-commit configuration to improve the hook's precision. Disappointed with performance, I considered trying Pyright instead but wasnt't sure about how good it handled Django codebases. 
- It's very nice to have a centrally kept list of issues, but there's a number of disadvantages that come with it. It's certainly more fragile than a bunch of `# type: ignore` comments; e.g., we'll have to edit the baseline when refactoring or even when formatting changes call for line rearrangement in a file. Some issues are presented differently between Python 3.11 and 3.12, and I had to fix those as part of this PR. In a codebase like ours, fixing typing errors is not trivial and almost always leads down some kind of a rabbit hole, so we should expect the fixes to modify the baseline file with "new" new issues instead of simply removing entries. We might be entering a world of pain (but fixing these issues does feel really good).
- The `# type: ignore` comments I added relate to the way we import our private/optional packages:
   1. `import-untyped` is ignored when the packages are installed.
   2. `import-not-found` is ignored when the packages are not present (i.e. in public CI)
   3. `unused-ignore` is ignored because it's one or the other of the above, but never both.

Ran both manually and as part of pre-commit hook against generated changes in Python files.


.pre-commit-config.yaml
```diff
@@ -36,13 +36,21 @@ repos:
...
   - repo: https://github.com/python-poetry/poetry
     rev: 2.0.1
     hooks:
       - id: poetry-check
-        args: ['-C', 'api']
+        args: ["-C", "api"]
 
 ci:
-  autoupdate_commit_msg: 'ci: pre-commit autoupdate'
+  skip: [python-typecheck]
```
matthewelwell, line 55: useful to know

api/app/settings/test.py
```diff
@@ -63,3 +63,5 @@
 """
 
 ENABLE_POSTPONE_DECORATOR = False
+
+DEBUG = True
```
matthewelwell, line 67: Do we still need this?

api/scripts/typecheck.py
```diff
@@ -0,0 +1,169 @@
...
+    # We want to ignore the path to the virtual environment
+
+    @property
+    def normalised_file(self) -> str:
+        if "site-packages" in (_file := self.file):
+            return _file.split("site-packages")[-1]
+        return _file
+
+    # Apparently, "message" and "hint" can differ between runs
+    # so ignore them for comparison and hashing
```
matthewelwell: It's unclear to me what this comment relates to.

api/scripts/typecheck.py
```diff
@@ -0,0 +1,169 @@
...
+@dataclass(frozen=True, eq=False)
+class MypyIssue:
+    file: str
+    line: int
+    column: int
+    code: int
+    severity: str
+    original_line: str
+
+    # We want to ignore the path to the virtual environment
```
matthewelwell: It's unclear to me what this comment relates to.

api/scripts/typecheck.py
```diff
@@ -0,0 +1,169 @@
...
+import os
+import sys
+from dataclasses import dataclass, field
+
+from mypy.api import run as run_mypy
+
+
+@dataclass
+class Config:
+    pre_commit_mode: bool = bool(os.environ.get("PRE_COMMIT"))
```
matthewelwell, line 23: Just checking that this is definitely the behaviour you were expecting? 

image

api/app/settings/test.py
khvn26: This is added to add test coverage to this bit: https://github.com/Flagsmith/flagsmith/blob/06e95fcc416f6e3361ad91324dd38571e7a0e4e3/api/app/urls.py#L46-L50

It doesn't break any tests, so we're gtg with keeping it set.

api/scripts/typecheck.py
khvn26: Yes. `PRE_COMMIT` is expected to be unset outside of pre-commit environment.

api/scripts/typecheck.py
khvn26: Improved in 0a5cd448c4cfb2945ab49ae84e884fac1eebe693.

api/scripts/typecheck.py
khvn26: Improved in 0a5cd448c4cfb2945ab49ae84e884fac1eebe693.

api/scripts/typecheck.py
```diff
@@ -0,0 +1,173 @@
...
+    line: int
+    column: int
+    code: int
+    severity: str
+    original_line: str
+
+    @property
+    def normalised_file(self) -> str:
+        if "site-packages" in (_file := self.file):
+            # For issues detected in dependecies' code,
```
matthewelwell:
```suggestion
            # For issues detected in dependencies' code,
```

matthewelwell (approved):
Approved with a minor typo fix suggestion.

PR #5113, opened 2025-02-17
fix: Feature tags not showing up in tag selection UI 

kyle-ssg:

## Changes

The following change caused a regression where tags were not shown here:

image

The code in question looks to try to hide unhealthy tags since they are automated however:

- Tags come through as an array of ids so the code would have never worked
- The condition is the wrong way around this would have only shown the unhealthy tags

View a feature that has tags and go to the settings tab

### DIFF
--- a/frontend/web/components/modals/CreateFlag.js
+++ b/frontend/web/components/modals/CreateFlag.js
@@ -72,7 +72,7 @@ const CreateFlag = class extends Component {
           multivariate_options: [],
         }
     const { allowEditDescription } = this.props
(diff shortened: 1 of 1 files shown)


matthewelwell:
This definitely fixes the immediate issue, but it does reveal the unhealthy tag when the `feature_health` flag is turned off. I think, based on the description of the PR, that is expected, but I'm just mentioning it here for confirmation. We should perhaps work out how to correctly hide this tag when the flag is off (in a separate PR). 

image

khvn26 (approved):

matthewelwell (approved):

PR #5119, opened 2025-02-17
ci: Use Mypy directly

khvn26:

## Changes

This removes the baseline file used for incremental type checking added in #5111, and adds 5000+ `# type: ignore` comments instead. All the comments are added with error codes. Comments added to imports are broken by isort (see https://github.com/PyCQA/isort/issues/2244), so I have to manually skip it, but black and flake8 hooks are passing.

The PR is essentialy a 1:1 conversion from the baseline file to `# type: ignore` comments. We'll be able to be more flexible with our changes, avoid bulky baseline file edits when e.g. changing order of existing lines, and clearly distinguish new issues vs the pre-existing ones. 

Note: after merging this, we will need to add a `.git-blame-ignore-revs` file containing the merge commit SHA to avoid polluting git blames with this.

This is a CI change + locally.


.pre-commit-config.yaml
```diff
@@ -52,5 +52,5 @@ repos:
         args: ["-C", "api"]
 
 ci:
-  skip: [python-typecheck]
+  skip: [python-typecheck, isort]
```
matthewelwell, line 55: Won't this still fire an error in our local dev environments then though?

khvn26: Not if problematic imports are not included in the pre-commit diffs...

matthewelwell (approved):

PR #5151, opened 2025-02-21
feat: Add separate liveness and readiness checks

rolodato:
This PR adds separate liveness and readiness checks, to be added later in the Helm charts. The existing `/health` endpoint is maintained with no changes for backwards compatibility.

The liveness check is the same as calling `/version`, which has no external dependencies. It does check that the filesystem is readable, since the version data is read from the filesystem. This is just testing that the API can respond to requests.

The readiness check is the same as the current `/health` endpoint, which checks that all migrations have been applied, and the database is available. We add a separate `/health/readiness` endpoint to make this obvious.

We also modify the `api/scripts/healthcheck.py` script, which is used by the default Docker Compose setup. Docker Compose health checks should effectively be liveness and not readiness health checks.

Tested manually by making requests to all the endpoints, and confirming there are no warnings about URL namespaces.

### DIFF
--- a/api/app/urls.py
+++ b/api/app/urls.py
@@ -14,6 +14,8 @@ urlpatterns = [
     re_path(r"^api/v1/", include("api.urls.v1", namespace="api-v1")),
     re_path(r"^api/v2/", include("api.urls.v2", namespace="api-v2")),
     re_path(r"^admin/", admin.site.urls),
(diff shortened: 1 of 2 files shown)


khvn26 (approved):
LGTM 👍

PR #5153, opened 2025-02-21
fix: Catch errors when failing to verify JWTs

### LINKED ISSUE #4982: Do not check cookie authentication for `/auth/saml/{configuration}/request/`

The Django `SECRET_KEY` is used as the JWT signing secret when `COOKIE_AUTH_ENABLED` is set but no `COOKIE_AUTH_JWT_SIGNING_KEY` was provided: https://github.com/Flagsmith/flagsmith/blob/2fe4e02be5a70c7a55205aed870db2b8d22fb361/api/app/settings/common.py#L837

Our default Helm values does not define a secret key, so one gets randomly generated on each deployment: https://github.com/Flagsmith/flagsmith-charts/blob/65ce6fd59b9dc3189e44cc8f4766aa5d8e468724/charts/flagsmith/values.yaml#L39-L44

Because the `/auth/saml/{configuration}/request/` endpoint validates cookies, the following can happen:

1. Log in via SSO - a JWT cookie gets set
2. Redeploy Flagsmith without a secret key - another secret key gets generated
3. Try to log in via SSO again - because the user still has a cookie set with the old secret key, this gets rejected by `/auth/saml/{configuration}/request` and the user cannot log in until they clear their cookies

This is technically not a bug since users should be setting secret keys, but it is a quality-of-life improvement to avoid confusing behaviour when testing Flagsmith infrastructure.

`/auth/saml/{configuration}/request` is a public endpoint, so it should not validate cookies at all.

Alternatively, clear cookies even before login if they were rejected by having an invalid signature. This is a possibly more intrusive change, and is already handled correctly by the frontend after login.

### Additional context

_No response_

rolodato:
Fixes https://github.com/Flagsmith/flagsmith/issues/4982.

The issue as described is incorrect, as the SAML routes already use [`AllowAny`](https://github.com/Flagsmith/flagsmith-saml/blob/cd925492ce8bf76a0bd46d5c36755b8118eb20a4/saml/views.py#L164). The problem is actually caused by `get_validated_token` returning an exception when called with an invalid JWT. We now catch any of these expected errors and gracefully fail the authentication instead.

This is completely untested.

### DIFF
--- a/api/tests/unit/custom_auth/jwt_cookie/test_unit_jwt_cookie_authentication.py
+++ b/api/tests/unit/custom_auth/jwt_cookie/test_unit_jwt_cookie_authentication.py
@@ -1,7 +1,7 @@
 from typing import Type
-from unittest.mock import MagicMock, patch
 
 import pytest
+from pytest_mock import MockerFixture
 from rest_framework.request import Request
 from rest_framework_simplejwt.exceptions import (
     AuthenticationFailed,
@@ -15,10 +15,10 @@ from custom_auth.jwt_cookie.constants import JWT_SLIDING_COOKIE_KEY
 from users.models import FFAdminUser
 
 
-def test_authenticate_without_cookie() -> None:
+def test_authenticate_without_cookie(mocker: MockerFixture) -> None:
     # Given
     auth = JWTCookieAuthentication()
-    request = MagicMock(spec=Request)
+    request = mocker.MagicMock(spec=Request)
     request.COOKIES = {}
 
     # When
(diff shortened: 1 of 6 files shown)


khvn26:
I am surprised that the new code path is covered by tests and no existing tests are modified, which suggests the backend behaviour is not changed by this PR?

rolodato:
@khvn26 I don't see any tests for this code, but in any case you can test this by trying to fetch a non-existent SAML configuration. Even if you are using an invalid JWT cookie, this should return 404 when using this fix:

```
curl -i -X POST http://localhost:8000/api/v1/auth/saml/foo/request/ -d{} --cookie jwt='badtoken'

HTTP/1.1 404 Not Found
Date: Fri, 07 Mar 2025 02:28:52 GMT
Server: WSGIServer/0.2 CPython/3.12.2
Content-Type: application/json
Vary: Accept, Origin
Allow: POST, OPTIONS
Expires: Fri, 07 Mar 2025 02:28:52 GMT
Cache-Control: max-age=0, no-cache, no-store, must-revalidate, private
Pragma: no-cache
X-Frame-Options: DENY
Content-Length: 58
X-Content-Type-Options: nosniff
Referrer-Policy: same-origin
Cross-Origin-Opener-Policy: same-origin

{"detail":"No SamlConfiguration matches the given query."}⏎
```

Before this change, the request fails and prevents users with expired or invalid tokens from logging in via SAML:

```
HTTP/1.1 401 Unauthorized
Date: Fri, 07 Mar 2025 02:28:23 GMT
Server: WSGIServer/0.2 CPython/3.12.2
Content-Type: application/json
WWW-Authenticate: Cookie realm="api"
Vary: Accept, Origin
Allow: POST, OPTIONS
Expires: Fri, 07 Mar 2025 02:28:23 GMT
Cache-Control: max-age=0, no-cache, no-store, must-revalidate, private
Pragma: no-cache
X-Frame-Options: DENY
Content-Length: 185
X-Content-Type-Options: nosniff
Referrer-Policy: same-origin
Cross-Origin-Opener-Policy: same-origin

{"detail":"Given token not valid for any token type","code":"token_not_valid","messages":[{"token_class":"SlidingToken","token_type":"sliding","message":"Token is invalid or expired"}]}⏎
```

api/tests/unit/custom_auth/jwt_cookie/test_unit_jwt_cookie_authentication.py
```diff
@@ -0,0 +1,71 @@
...
+    TokenError,
+)
+from rest_framework_simplejwt.tokens import Token
+
+from custom_auth.jwt_cookie.authentication import JWTCookieAuthentication
+from custom_auth.jwt_cookie.constants import JWT_SLIDING_COOKIE_KEY
+from users.models import FFAdminUser
+
+
+class TestJWTCookieAuthentication:
```
khvn26: We don't use classes for new tests we write. We rely on modules and [markers](https://docs.pytest.org/en/stable/example/markers.html) for grouping tests.

api/tests/unit/custom_auth/jwt_cookie/test_unit_jwt_cookie_authentication.py
```diff
@@ -0,0 +1,71 @@
...
+
+                # Then
+                assert result == (user, validated_token)
+                mock_validate.assert_called_once_with(raw_token)
+                mock_get_user.assert_called_once_with(validated_token)
+
+    @pytest.mark.parametrize(
+        "exception_class", [InvalidToken, TokenError, AuthenticationFailed]
+    )
+    def test_authenticate_invalid_cookie(self, exception_class) -> None:
```
khvn26:
```suggestion
    def test_authenticate_invalid_cookie(self, exception_class: Type[Exception]) -> None:
```

khvn26:
```suggestion
    def test_authenticate_invalid_cookie(self, exception_class: Type[Exception]) -> None:
```

khvn26 (approved):

PR #5161, opened 2025-02-25
feat: Switch existing task processor health checks to new liveness probe

khvn26:

## Changes

This accommodates for changes introduced in https://github.com/Flagsmith/flagsmith-task-processor/pull/25 by intercepting `python manage.py checktaskprocessorthreadhealth` command and redirecting it to new liveness probe introduced in #5151.

Ran `python manage.py checktaskprocessorthreadhealth` and made sure the correct HTTP call is made.


api/scripts/healthcheck.py
```diff
@@ -2,7 +2,13 @@
 
 import requests
 
-url = "http://localhost:8000/health/liveness"
-status = requests.get(url).status_code
 
-sys.exit(0 if 200 >= status < 300 else 1)
+def main() -> None:
```
rolodato, line 12: Can we add a deprecation warning when this is being used? In case we need to troubleshoot other customer deployments that have similar issues, this could tell us immediately if they are using the old style healthchecks or not.

I understand this code is only ever called by [Docker Compose healthchecks](https://github.com/Flagsmith/flagsmith/blob/46570719d361f1cc9c4fbf789f9f2bbfbf30256e/docker-compose.yml#L61), which should be removed.

Otherwise LGTM 👍

khvn26: Done.

rolodato:
@khvn26 Added https://github.com/Flagsmith/flagsmith/pull/5161/commits/236df8663bbb494b9e0c56ebd412ee17ff710436 to remove healthchecks from Docker Compose, and bake them directly into the Dockerfile. I understand this should only be used automatically by Compose, and not ECS, Kubernetes or other orchestrators.

rolodato (approved):

Dockerfile
```diff
@@ -139,6 +139,9 @@ ENTRYPOINT ["/app/scripts/run-docker.sh"]
 
 CMD ["migrate-and-serve"]
 
+HEALTHCHECK --interval=2s --timeout=2s --retries=3 --start-period=20s \
+  CMD curl -f http://localhost:8000/health/liveness || exit 1
```
rolodato, line 143: @khvn26 I only just now noticed the Wolfi Docker images don't contain curl. Don't merge this until it's corrected, sorry about that!

khvn26:
In 121128e3e1d71e01447b145f07a6f3025261dedb, I added support for [the new Task processor endpoint](https://github.com/Flagsmith/flagsmith-task-processor/pull/25/commits/43387cbaa0ad6cec6e48b729e2a741a2e0004013) that now accepts command line arguments for Gunicorn as well. This enables HTTP healthchecks directly on Task processor containers.

We'll need to merge https://github.com/Flagsmith/flagsmith-task-processor/pull/25 and change the dependency reference before we merge this.

I also refreshed documentation related to Task processor operation.

api/pyproject.toml
```diff
@@ -146,7 +146,7 @@ pygithub = "2.1.1"
 hubspot-api-client = "^8.2.1"
 djangorestframework-dataclasses = "^1.3.1"
 pyotp = "^2.9.0"
-flagsmith-task-processor = { git = "https://github.com/Flagsmith/flagsmith-task-processor", tag = "v1.2.1" }
+flagsmith-task-processor = { git = "https://github.com/Flagsmith/flagsmith-task-processor", rev = "2f4e95787d15f3c0fc86ecd0fc7036b353fbeba6" }
```
gagantrivedi: This needs to be updated once task processor is released?

gagantrivedi (approved):

api/pyproject.toml
khvn26: That's correct.

api/scripts/healthcheck.py
matthewelwell: I am slightly concerned with the deprecation warning that's been added here since, if I understand correctly, anyone invoking `python manage.py checktaskprocessorthreadhealth` will also see this confusing deprecation warning. 

I _think_ that could be anyone who explicitly uses the management command, but more likely anyone that updates the flagsmith version without updating their chart version?

matthewelwell (approved):

api/scripts/healthcheck.py
khvn26: Valid point. I've reworked the warning to be more descriptive.

Issue #4384, opened 2024-07-24
Move custom field settings into org settings


rolodato:
### How are you running Flagsmith

- [ ] Self Hosted with Docker
- [ ] Self Hosted with Kubernetes
- [X] SaaS at flagsmith.com
- [ ] Some other way (add details in description below)

### Describe the bug

Custom fields are defined per organisation, but the UI to configure them lives under Project Settings. This should be under Organisation Settings instead.

### Steps To Reproduce

Go to Project Settings > Custom Fields. The fields shown here are global to the org, not specific to the current project

### Expected behavior

Fixing this will probably break existing links like https://app.flagsmith.com/project/16945/settings?tab=custom-fields. I think this is okay given how little Custom Fields are currently used, and because the links are project-specific and not global to all users.

kyle-ssg:
I've taken a look at this, and for some reason it looks like creating/editing the fields is done on an organisations level but making a field "required" is done on the project level via a is_required_for "object_id" which is a project id. I don't think this makes sense, we should either be creating a field at the project level or making the required attribute organisation wide.

It seems in order to move this to organisation settings the easiest approach would be to just remove the object_id parameter.

tiagoapolo:
> I've taken a look at this, and for some reason it looks like creating/editing the fields is done on an organisations level but making a field "required" is done on the project level via a is_required_for "object_id" which is a project id. I don't think this makes sense, we should either be creating a field at the project level or making the required attribute organisation wide.
> 
I drafted a PR moving custom fields to org settings. In addition to Kyle's comment, I tested setting a field as required and that field was set as required for every project meaning that `object_id` should not be necessary and the required attribute is already set at organisation level.

The next step is removing `object_id` field from `/metadata-model-fields/` endpoint

https://github.com/user-attachments/assets/497082c7-f34b-40e7-bdb5-ea26a584d034

tiagoapolo:
That [sub-task](https://github.com/Flagsmith/flagsmith/issues/5167) is a requirement for this this to be completed

Zaimwa9:
I'm answering here but it mostly concerns the [sub-task](https://github.com/Flagsmith/flagsmith/issues/5167).

I wanted to confirm some of your above points (correct me if I am wrong or if it's used differently elsewhere): 
- The whole project_id/object_id field seems unused at the moment
- As stated, creating a custom field is already behaving as if it was at the org level and it will appear in every projects.
[...]

PR #5163, opened 2025-02-27
fix: move custom fields to org settings

tiagoapolo:

## Changes

Ref: #4384

- Moving custom fields config to org settings because they are org level changes
- `object_id` needs to be removed because when a field is set as required it's set at org level configuration, we don't need to specify the project id 

_Please describe._

### DIFF
--- a/frontend/web/components/pages/ProjectSettingsPage.js
+++ b/frontend/web/components/pages/ProjectSettingsPage.js
@@ -24,6 +24,7 @@ import Setting from 'components/Setting'
 import PlanBasedBanner from 'components/PlanBasedAccess'
 import classNames from 'classnames'
 import EditHealthProvider from 'components/EditHealthProvider'
+import WarningMessage from 'components/WarningMessage'
 
 const ProjectSettingsPage = class extends Component {
   static displayName = 'ProjectSettingsPage'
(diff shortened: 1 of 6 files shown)


emyller (approved):

matthewelwell (commented):
This looks good to me from a code perspective, but I wonder if there's a way that we can redirect any users that are expecting to navigate to the custom fields tab in the project settings to the organisation settings?

tiagoapolo:
> but I wonder if there's a way that we can redirect any users that are expecting to navigate to the custom fields tab in the project settings to the organisation settings?

Sure thing!

tiagoapolo:
> > but I wonder if there's a way that we can redirect any users that are expecting to navigate to the custom fields tab in the project settings to the organisation settings?
> 
> Sure thing!
Screenshot 2025-04-24 at 14 02 49

frontend/web/components/pages/ProjectSettingsPage.js
```diff
@@ -604,9 +604,28 @@ const ProjectSettingsPage = class extends Component {
...
+                            page.{' '}
+                            <a
+                              href={`/organisation/${
+                                AccountStore.getOrganisation()?.id
+                              }/settings?tab=custom-fields`}
+                              rel='noreferrer'
+                            >
+                              Click here to go there
+                            </a>
+                            .
```
matthewelwell: I think we can be slightly less verbose here. 

```suggestion
                            Custom fields have been moved to {' '}
                            <a
                              href={`/organisation/${
                                AccountStore.getOrganisation()?.id
                              }/settings?tab=custom-fields`}
                              rel='noreferrer'
                            >
                              Organisation Settings
                            </a>
                            .
```

matthewelwell (approved):
Looks good to me now. I've added a small suggestion for the copy of the relocation warning.

Zaimwa9 (approved):
Tested and approved after committing suggestion 👏

PR #5187, opened 2025-03-05
feat: add permanent environment document cache

matthewelwell:
## Changes

This PR sets out to add a permanent cache of the environment document, similar to how our SaaS environment works where it writes the document to dynamodb. This updates the current logic for caching the environment document, but hooks it into the same logic that updates the document in dynamodb for our SaaS platform. It also makes the cache backend customisable to allow self hosted installs to use, e.g. redis to cache the document. 

This PR also adds a new metric called `flagsmith_environment_document_cache_result` which tracks the number of cache hits and misses for the environment document cache. 

Added a new test to verify the flow end-to-end. I also tested it manually, as can be seen in [this comment](https://github.com/Flagsmith/flagsmith/pull/5187#issuecomment-2744020451).


api/app/settings/common.py
```diff
@@ -679,8 +679,24 @@
     "django.core.cache.backends.locmem.LocMemCache",
 )
 
-CACHE_ENVIRONMENT_DOCUMENT_SECONDS = env.int("CACHE_ENVIRONMENT_DOCUMENT_SECONDS", 0)
 ENVIRONMENT_DOCUMENT_CACHE_LOCATION = "environment-documents"
+ENVIRONMENT_DOCUMENT_CACHE_BACKEND = env(
+    "ENVIRONMENT_DOCUMENT_CACHE_BACKEND", "django.core.cache.backends.db.DatabaseCache"
+)
```
khvn26: Have we discussed trade-offs and advantages of using database as the default cache backend for environments? Do you think an in-memory cache is more appropriate as a default?

api/app/settings/common.py
```diff
@@ -679,8 +679,24 @@
...
+    if "CACHE_ENVIRONMENT_DOCUMENT_SECONDS" in os.environ:
+        warnings.warn(
+            "Ignoring CACHE_ENVIRONMENT_DOCUMENT_SECONDS variable "
+            "since EXPIRE_ENVIRONMENT_DOCUMENT_CACHE is False"
+        )
+    CACHE_ENVIRONMENT_DOCUMENT_TIMEOUT = None
+else:
+    CACHE_ENVIRONMENT_DOCUMENT_TIMEOUT = env.int(
+        "CACHE_ENVIRONMENT_DOCUMENT_SECONDS", 0
+    )
```
khvn26: What's the difference between `EXPIRE_ENVIRONMENT_DOCUMENT_CACHE=False` and `CACHE_ENVIRONMENT_DOCUMENT_TIMEOUT=0`? Also, why change a good name (_SECONDS suffix) to a worse one?

api/environments/models.py
```diff
@@ -288,10 +291,21 @@ def write_environments_to_dynamodb(
...
+                and environment_v2_wrapper.is_enabled
+            ):  # type: ignore[union-attr]
+                environment_v2_wrapper.write_environments(environments)
+        elif not settings.EXPIRE_ENVIRONMENT_DOCUMENT_CACHE:
+            environment_document_cache.set_many(
+                {
+                    e.api_key: map_environment_to_environment_document(e)
+                    for e in environments
+                }
+            )
```
khvn26: I don't believe the `is_saas()` check belongs here. I'd rather allow our users to employ their Dynamo in the future, e.g., by writing a Dynamo cache backend.

api/app/settings/common.py
matthewelwell: The only reason is backwards compatibility. I agree that it's not the most appropriate, but it's not worth breaking any existing behaviour for.

api/app/settings/common.py
matthewelwell: > What's the difference between EXPIRE_ENVIRONMENT_DOCUMENT_CACHE=False and CACHE_ENVIRONMENT_DOCUMENT_TIMEOUT=0?

Well, technically, lots! See [here](https://docs.djangoproject.com/en/5.1/topics/cache/#cache-arguments) - they're distinct opposites. If we were to assume that `0` meant cache indefinitely then we'd be completely changing the default behaviour (which is to not cache at all). Admittedly, we have logic elsewhere in the app that only inspects the cache if the value is greater than 0 (see [here](https://github.com/Flagsmith/flagsmith/blob/647712caf6e4e61eac18fca0a38ca7ee2beaadef/api/environments/models.py#L367)), so we could modify that to our will, but I'd rather keep it such that the settings make sense on their own. 

> Also, why change a good name (_SECONDS suffix) to a worse one?

Yeah, this is a good question. My logic here was that `None` is now a valid value for the variable which, to me, meant that `_TIMEOUT` made more sense - I don't feel particularly strongly about this though. Note that I did not change the environment variable name, only the setting.

api/environments/models.py
matthewelwell: Hmm... I somewhat agree with this. Let me see if I can refactor this (and the environment variables as per the above conversation).

api/app/settings/common.py
khvn26: I've come up with the following options to consider:
1. Accept `-1` as a `CACHE_ENVIRONMENT_DOCUMENT_SECONDS ` value that means "never expire the cache". This will simplify business logic and hopefully ease cognitive load. We have a lot of settings already!
2. Add an `CACHE_ENVIRONMENT_DOCUMENT_MODE: Literal["PERSISTENT", "EXPIRING"] = env.str(default="EXPIRING")`. This introduces a new setting but it's more expressive and easy to understand.

In any case, I expect the new/modified settings to be reflected in the docs.

> My logic here was that None is now a valid value for the variable

I think it can be just left at 0 (or -1 if we go with option 1) .

matthewelwell: Updated the logic in line with your second suggestion 👍

api/environments/models.py
matthewelwell: Looking into this a bit more, I think my concern here is that this implies a much larger refactor.

I guess the ideal solution would be to refit our SaaS platform to essentially have something like: 

```
ENVIRONMENT_DOCUMENT_CACHE_MODE: PERSISTENT
ENVIRONMENT_DOCUMENT_CACHE_BACKEND: cache.backends.custom.dynamodb
```

where `cache.backends.custom.dynamodb` is our own implementation of the django cache backend which handles the interactions with dynamodb.

While the prospect of this certainly excites me, I don't think it should necessarily be in scope for this piece of work. 

An option to solve your direct comment here would be to remove the call to `is_saas()` and instead use some other setting, perhaps derived from whether e.g. `ENVIRONMENTS_TABLE_NAME_DYNAMO` is set.

Thoughts @khvn26?

matthewelwell: I actually found a half decent halfway solution as there was already a conditional check in the code that was only relevant to the dynamodb path, so I've just used that as the check instead of `is_saas` for now. I still think the custom cache backend would be the best option, and may well simplify a lot of the dynamo integration code.

matthewelwell:
More testing is needed here, but initial signs are good regarding performance. I tested this code by building the docker image locally and using the following (simplified for brevity) docker compose file. 

```yml
volumes:
  pgdata:

services:
  postgres:
    image: postgres:15.5-alpine
    environment:
      POSTGRES_PASSWORD: password
      POSTGRES_DB: flagsmith
    volumes:
      - pgdata:/var/lib/postgresql/data
    healthcheck:
      test: ['CMD-SHELL', 'pg_isready -d flagsmith -U postgres']
      interval: 2s
      timeout: 2s
      retries: 20
      start_period: 20s

  flagsmith:
    image: flagsmith/flagsmith:local
    environment:

      ENVIRONMENT: production # set to 'production' in production.
      DJANGO_ALLOWED_HOSTS: '*' # Change this in production
      ALLOW_ADMIN_INITIATION_VIA_CLI: 'true' # Change this in production
      FLAGSMITH_DOMAIN: localhost:8000 # Change this in production
      DJANGO_SECRET_KEY: secret

      # Enable Task Processor
      TASK_RUN_METHOD: TASK_PROCESSOR

      ENVIRONMENT_DOCUMENT_CACHE_MODE: PERSISTENT
      ENVIRONMENT_DOCUMENT_CACHE_BACKEND: django.core.cache.backends.redis.RedisCache
      ENVIRONMENT_DOCUMENT_CACHE_LOCATION: redis://redis_db:6379/1

    ports:
      - 8000:8000
    healthcheck:
      test: ['CMD-SHELL', 'python /app/scripts/healthcheck.py']
      interval: 2s
      timeout: 2s
      retries: 20
      start_period: 20s
    depends_on:
      postgres:
        condition: service_healthy
      redis_db:
        condition: service_started

  flagsmith_processor:
    image: flagsmith/flagsmith:local
    environment:
      USE_POSTGRES_FOR_ANALYTICS: 'true'

      LOG_LEVEL: DEBUG

      ENVIRONMENT_DOCUMENT_CACHE_MODE: PERSISTENT
      ENVIRONMENT_DOCUMENT_CACHE_BACKEND: django.core.cache.backends.redis.RedisCache
      ENVIRONMENT_DOCUMENT_CACHE_LOCATION: redis://redis_db:6379/1
    depends_on:
      flagsmith:
        condition: service_healthy
    command: run-task-processor

  redis_db:
    image: redis:latest
```

In order to compare performance, I simply removed the `ENVIRONMENT_DOCUMENT_X` environment variables from the `flagsmith` container and restarted the docker compose service. 

Here are the before and afters: 

**Before**

```
❯ wrk http://localhost:8000/api/v1/environment-document/ -H 'X-Environment-Key: ser.XXX'

Running 10s test @ http://localhost:8000/api/v1/environment-document/
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   154.93ms  338.61ms   1.69s    89.80%
    Req/Sec   128.30     21.90   151.00     91.67%
  2179 requests in 10.09s, 3.93MB read
Requests/sec:    216.05
Transfer/sec:    399.19KB
```

**After**

```
❯ wrk http://localhost:8000/api/v1/environment-document/ -H 'X-Environment-Key: ser.XXX'

Running 10s test @ http://localhost:8000/api/v1/environment-document/
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    15.68ms  115.10ms   1.61s    98.57%
    Req/Sec     1.64k   366.56     2.22k    56.00%
  32616 requests in 10.01s, 63.83MB read
Requests/sec:   3259.35
Transfer/sec:      6.38MB
```

api/app/settings/common.py
```diff
@@ -679,8 +680,27 @@
     "django.core.cache.backends.locmem.LocMemCache",
 )
 
+ENVIRONMENT_DOCUMENT_CACHE_LOCATION = env(
+    "ENVIRONMENT_DOCUMENT_CACHE_LOCATION", default="environment-documents"
```
rolodato: Could we use a common prefix on the caching-related env vars? Example:

```
CACHE_ENVIRONMENT_DOCUMENT_LOCATION
CACHE_ENVIRONMENT_DOCUMENT_BACKEND
CACHE_ENVIRONMENT_DOCUMENT_MODE
```

matthewelwell: Yep, good call, not sure why I hadn't done this... Is done [here](https://github.com/Flagsmith/flagsmith/pull/5187/commits/6e09d73bec7593b7db4fd27e69016509f4b74fdf).

api/environments/models.py
```diff
@@ -386,9 +427,15 @@ def _get_environment_document_from_cache(
...
         environment_document = environment_document_cache.get(api_key)
-        if not environment_document:
+        if not (cache_hit := environment_document is not None):
             environment_document = cls._get_environment_document_from_db(api_key)
             environment_document_cache.set(api_key, environment_document)
+
+        flagsmith_environment_document_cache_result.labels(
+            api_key=api_key,
+            result=CACHE_HIT if cache_hit else CACHE_MISS,
+        ).inc()
```
matthewelwell: There's an argument to say that this metric incrementing should be done in `Environment.get_environment_document` above but I decided that it didn't really make sense to track cache hits / misses in an instance that isn't configured to use the cache.

api/environments/metrics.py
```diff
@@ -0,0 +1,10 @@
+import prometheus_client
+
+CACHE_HIT = "CACHE_HIT"
+CACHE_MISS = "CACHE_MISS"
+
+flagsmith_environment_document_cache_result = prometheus_client.Counter(
+    "flagsmith_environment_document_cache_result",
```
khvn26: Please check with the [metric naming convention](https://prometheus.io/docs/practices/naming/#metric-names) when coming up with metric names.

Since this is a unit-less accumulating count, I'd suggest `"flagsmith_environment_document_cache_operations_total"`.

api/environments/metrics.py
```diff
@@ -0,0 +1,10 @@
...
+    "Results of cache retrieval for environment document (hit or miss)",
+    ["api_key", "result"],
```
khvn26: While it being clearly valuable, I am not sure about the `"api_key"` label. It's unbound, which is a recipe for cardinality explosion.

If we want to keep this label, consider adding a configurable allow list for environment keys to track — we could use it for any other metrics we'd like to attribute to environment keys as well.

matthewelwell: Yep, fair point. I stuck with `results` instead of `operations` as I'm not sure I'd consider a 'CACHE_HIT' to be an operation. I'd argue that operations would be e.g. `get|set|delete|...` ?

khvn26: I believe your comment is for the [other thread](https://github.com/Flagsmith/flagsmith/pull/5187#discussion_r2039187959)

matthewelwell: Ugh, yep, sorry.

api/environments/metrics.py
matthewelwell: Yep, fair point. I stuck with `results` instead of `operations` as I'm not sure I'd consider a 'CACHE_HIT' to be an operation. I'd argue that operations would be e.g. `get|set|delete|...` ?

api/environments/models.py
```diff
@@ -386,9 +427,15 @@ def _get_environment_document_from_cache(
...
         environment_document = environment_document_cache.get(api_key)
-        if not environment_document:
+        if not (cache_hit := environment_document is not None):
             environment_document = cls._get_environment_document_from_db(api_key)
             environment_document_cache.set(api_key, environment_document)
+
+        flagsmith_environment_document_cache_results_total.labels(
+            api_key=api_key,
+            result=CACHE_HIT if cache_hit else CACHE_MISS,
+        ).inc()
```
khvn26: Please cover the increment with tests. You can use the `assert_metric` fixture.

api/environments/metrics.py
khvn26: `results` sounds a bit obscure to me. Do you think it would make sense to use two distinct counters here?
- `flagsmith_environment_document_cache_hits_total`
- `flagsmith_environment_document_cache_misses_total`

api/environments/models.py
matthewelwell: [I have](https://github.com/Flagsmith/flagsmith/pull/5187/files#diff-20538f1a0b1c78b087d26dfe7f43e2e8f7485071b1e47eb31c77044a56f22fa8R1059-R1087)? Admittedly using mocks, but I couldn't see any examples or documentation on the `assert_metric` fixture outside of the common repository. 

I'll happily update it though to avoid the use of mocks if you can point me in the right direction.

api/environments/metrics.py
matthewelwell: > While it being clearly valuable, I am not sure about the "api_key" label. It's unbound, which is a recipe for cardinality explosion.

Fair point! I'll just remove for now.

api/environments/models.py
khvn26: Ah, sorry, missed that. 

Yeah, `assert_metric` is usable outside of `flagmith-common` repo; it's installed as part of the pytest plugin. [This](https://github.com/Flagsmith/flagsmith-common/blob/1d35c43870028885373b5ffcbcd07c41c8a94b29/tests/unit/task_processor/test_unit_task_processor_processor.py#L579-L636) should be a good usage example.

I'll work on adding this info to common's README asap.

api/environments/metrics.py
matthewelwell: > results sounds a bit obscure to me.

That's a fair assessment. 

> Do you think it would make sense to use two distinct counters here?

I did originally have it that way. The reason I went with the single metric was that I assumed it would be easier to perform maths on a single metric with multiple labels to determine hits vs misses, but I assume it's easy enough to use 2 separate metrics? If so, I'll just do this.

khvn26: Okay, let's just come up with a better name for the single metric. [Prometheus docs](https://prometheus.io/docs/practices/instrumentation/#caches) suggest `queries` which sounds good to me.

api/environments/models.py
matthewelwell: Done [here](https://github.com/Flagsmith/flagsmith/pull/5187/commits/3db98279bea21f8b5c7bbc8a5d70d10b5aa5226f)

api/environments/metrics.py
matthewelwell: Done [here](https://github.com/Flagsmith/flagsmith/pull/5187/commits/9f4851449912a099f3fd4550c0850a2ff91bebec).

api/environments/metrics.py
matthewelwell: Yep, agreed. Have updated.

khvn26 (approved):

PR #5192, opened 2025-03-05
feat: Enhance feature segment permissions with tag validation

### LINKED ISSUE #5132: Tag-level permissions for segment overrides

Per https://docs.flagsmith.com/system-administration/rbac#environment, we currently support tag-level permissions for updating feature states, and creating/approving change requests.

We are missing the capability to have tag-level permissions for managing segment overrides as well. This is key for allowing Enterprise customers to control their flag rollout in a safe way without affecting unrelated flags.

Implement tag-based permissions for the "Manage segment overrides" permission.

🤷 

### Additional context

_No response_

francescolofranco:

## Changes

This PR partially enables a check against tags for the management of segments overrides. Testing is limited as this would require the settings.IS_RBAC_INSTALLED to be true, forcing it in tests causes an obvious error.

What remains to do is some additional testing, I think.

Added a couple of automated tests, which are not really checking much of the logic, since rbac repo is needed for this.

## More info

I'm afraid this PR is only partial as to fully complete this work I would need access to the private rbac repo. This PR also requires the addition of the MANAGE_SEGMENT_OVERRIDES permission to the TAG_SUPPORTED_PERMISSIONS in the common repo. PR -> https://github.com/Flagsmith/flagsmith-common/pull/14

### DIFF
--- a/api/features/feature_segments/permissions.py
+++ b/api/features/feature_segments/permissions.py
@@ -27,14 +27,18 @@ class FeatureSegmentPermissions(IsAuthenticated):
                 environment = request.data.get("environment")
                 environment = Environment.objects.get(id=int(environment))
 
-                feature_id = request.data.get("feature") or view.kwargs.get("feature_pk")
+                feature_id = request.data.get("feature") or view.kwargs.get(
+                    "feature_pk"
+                )
                 feature = Feature.objects.get(
                     id=feature_id, project=environment.project
                 )
                 tag_ids = list(feature.tags.values_list("id", flat=True))
 
                 return request.user.has_environment_permission(
-                    permission=MANAGE_SEGMENT_OVERRIDES, environment=environment, tag_ids=tag_ids
+                    permission=MANAGE_SEGMENT_OVERRIDES,
+                    environment=environment,
+                    tag_ids=tag_ids,
                 )
 
         return False
--- a/api/features/feature_segments/permissions.py
+++ b/api/features/feature_segments/permissions.py
@@ -6,6 +6,7 @@ from common.environments.permissions import (  # type: ignore[import-untyped]
 from rest_framework.permissions import IsAuthenticated
 
 from environments.models import Environment
+from features.models import Feature
 
 
 class FeatureSegmentPermissions(IsAuthenticated):
@@ -17,7 +18,7 @@ class FeatureSegmentPermissions(IsAuthenticated):
         if view.detail:
             return True
 
-        # handle by the view
+        # handled by the view
         if view.action in ["list", "get_by_uuid", "update_priorities"]:
             return True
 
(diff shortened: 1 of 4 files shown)


api/features/feature_segments/permissions.py
```diff
@@ -26,13 +27,27 @@ def has_permission(self, request, view):  # type: ignore[no-untyped-def]
                 environment = request.data.get("environment")
                 environment = Environment.objects.get(id=int(environment))
 
+                feature_id = request.data.get("feature") or view.kwargs.get(
+                    "feature_pk"
+                )
+                feature = Feature.objects.get(
+                    id=feature_id, project=environment.project
+                )
+                tag_ids = list(feature.tags.values_list("id", flat=True))
```
matthewelwell: From what I can see, this class is only ever used on the `FeatureSegmentViewSet` which lives at `/features/feature-segments` so I can't see a valid reason to check `view.kwargs` here. 

In fact, the serializer for the `create` action (defined [here](https://github.com/Flagsmith/flagsmith/blob/main/api/features/feature_segments/serializers.py#L18) and referenced [here](https://github.com/Flagsmith/flagsmith/blob/main/api/features/feature_segments/views.py#L71)) enforces that the feature is provided. 

That being said, since the permission check is performed prior to the serializer validation, I would suggest that we write this code in a defensive manner, and only receive the feature (and tags) if we have the feature id. 

Something like the following perhaps?

```suggestion
                if feature_id := request.data.get("feature"):
                    feature = Feature.objects.get(
                        id=feature_id, project=environment.project
                    )
                    tag_ids = list(feature.tags.values_list("id", flat=True))
                else:
                    tag_ids = []
```

api/features/permissions.py
```diff
@@ -211,9 +211,15 @@ def has_permission(self, request, view):  # type: ignore[no-untyped-def]
             Environment, api_key=view.kwargs["environment_api_key"]
         )
 
+        feature_id = request.data.get("feature") or view.kwargs.get("feature_pk")
```
matthewelwell: Similar to the comment I left on the other similar piece of logic, here the `"feature_pk"` view kwarg is required so we can just grab that and use the same (or similar) logic as to what I suggested above.

matthewelwell (commented):
Thanks @francescolofranco. I've added a couple of suggestions for code improvements, but on the whole, I think it looks good. I've also approved the linked PR in the flagsmith-common repository which we'll need to release and bump the version of here before release. 

Regarding the RBAC repository, I _think_ there are no changes actually needed. Your comment about testing though is a good one, I'm not sure how we've handled this so far - perhaps @gagantrivedi knows? Based on the testing in the RBAC repo though, I suggest we can get away with the testing you've already added here @francescolofranco. Perhaps, if we wanted to go slightly further, we could mock the call to the function in the `rbac_wrapper` module and verify that the tags are being correctly passed?

api/features/feature_segments/permissions.py
francescolofranco: > From what I can see, this class is only ever used on the FeatureSegmentViewSet which lives at /features/feature-segments so I can't see a valid reason to check view.kwargs here.

Ah ok, that's what I was missing, I tried to use the same approach in both permissions cases but obviously tests were failing because the different views.

api/features/permissions.py
francescolofranco: Yep, updated

francescolofranco (commented):
I'm interested in knowing how do you feel about these types of tests -> [d0aca97](https://github.com/Flagsmith/flagsmith/pull/5192/commits/d0aca97bf45f2dfd6c1a748c3bfa5e1523a18695) ?

I know this is a diversion from the usual way of testing in this project, so I'm ok if you decide to discard this commit.

If you find those useful, I can also cover the `CreateSegmentOverridePermissions`, just let me know 🙏🏻

PR #5199, opened 2025-03-07
feat: Support passing in replica URLs in JSON format

rolodato:
When setting `REPLICA_DATABASE_URLS` or `CROSS_REGION_REPLICA_DATABASE_URLS`, users need to know if their passwords/URLs contain a comma, and set `REPLICA_DATABASE_URLS_DELIMITER` accordingly. This is annoying and not always possible, especially in environments where users might not ever have access to the raw database password such as Kubernetes secrets.

Instead of having users that don't know their passwords go through trial and error, we now accept the URLs as a JSON-formatted array and avoid needing to specify a delimiter. The plan is to use this new variable to make it easier to pass arbitrary replica URLs from Helm values.


api/poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 2.0.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.1.1 and should not be changed by hand.
```
gagantrivedi: This should be removed from this pull request

api/app/settings/common.py
```diff
@@ -173,21 +173,39 @@
...
     }
     REPLICA_DATABASE_URLS_DELIMITER = env("REPLICA_DATABASE_URLS_DELIMITER", ",")
-    REPLICA_DATABASE_URLS = env.list(
-        "REPLICA_DATABASE_URLS", default=[], delimiter=REPLICA_DATABASE_URLS_DELIMITER
+    # Allow specifying replica database URLs as a JSON array to avoid delimiter conflicts
+    REPLICA_DATABASE_URLS_JSON = json.loads(
+        env.str("REPLICA_DATABASE_URLS_JSON", "null")
     )
+
+    if REPLICA_DATABASE_URLS_JSON:
```
gagantrivedi: nit: `REPLICA_DATABASE_URLS_JSON` is not JSON but python here.. Maybe we can change it something like:

```
REPLICA_DATABASE_URLS_JSON = env.str("REPLICA_DATABASE_URLS_JSON")
if REPLICA_DATABASE_URLS_JSON:
  REPLICA_DATABASE_URLS= json.loads(REPLICA_DATABASE_URLS)

```

gagantrivedi (changes requested):

gagantrivedi (approved):

rolodato:
Thanks for the review! I'll merge this once I can generate these variables directly from Helm.

rolodato:
I spent way too much time trying to use this from a Helm chart and decided it's not a feasible approach. I'll open a separate PR with a different method.

Issue #4688, opened 2024-10-02
Support JWT refresh flow for cookie authentication


khvn26:
Most probably it'll include
- reducing token TTL to very short time, e.g. 5-15 minutes
- introducing a separate http-only cookie for the refresh token
- exposing the refresh endpoint, provided by the simplejwt library
- adding retry/refresh dance to FE fetchers

Spawned out of #4646 & #4647

kyle-ssg:
This will be quite simple to implement in service.ts prepareHeaders and _request in data.js

tiagoapolo:
@khvn26 I'll need some help reviewing this and probably fixing types.

rolodato:
Reducing the token TTL is a good goal to have, but introducing refresh tokens while treating them the same as access tokens will introduce more problems than they solve. I'll recap the current situation to understand the whole problem.

## Current situation

We store a session cookie that lasts for 30 days. The value of this cookie is valid indefinitely, since it corresponds to the current user's personal API access token. This API access token lasts until forever until revoked, and is stored in the database in plain text in the `authtoken_token` table. The cookie is only cleared client-side after the 30 days have passed, which logs the user out.

If a user's session cookie is stolen, or the database is leaked, the attacker can impersonate the user indefinitely until their token is manually rotated. Rotating the Django session secret will not help, as this token is not signed.

## JWT authentication

To mitigate the impact of stolen cookies or a leaked token database, we introduced the `COOKIE_AUTH_ENABLED` and `COOKIE_AUTH_JWT_ACCESS_TOKEN_LIFETIME_MINUTES` options. These make it so that:

* We issue a temporary access token instead of storing a permanent token in the authentication cookie
* This access token has a finite lifespan, which is enforced by the JWT's signature

With this approach, stolen cookies are still a threat, but the impact window is limited to the lifetime of the token (default 10 hours). The only downside is that users must now authenticate roughly every workday on every device when they want to use Flagsmith.

In addition to session cookies now actually expiring, we can also end sessions early by using SimpleJWT's [`token_blacklist` app](https://django-rest-framework-simplejwt.readthedocs.io/en/latest/blacklist_app.html). However, using a blacklist this way negates some of the benefit of using JWT as tokens, since we now need to keep some state about JWTs on the backend.

There are ways to mitigate the impact of stolen tokens further:

1. Detect session forking, which requires server-side state and logic, and is not 100% reliable. For example: https://slack.engineering/catching-compromised-cookies/
[...]

PR #5200, opened 2025-03-07
feat: JWT token refresh

tiagoapolo:

## Changes

Ref: [#4688](https://github.com/Flagsmith/flagsmith/issues/4688)

This pull request replaces Sliding Tokens with Access and Refresh Tokens.
- Exposes /auth/token/refresh/ for refreshing access tokens using a valid refresh token.
- Updated `authorise_response` to set cookies for both `AccessToken` and `RefreshToken`
- Introduced `JWTCookieTokenRefreshView` for handling token refresh and updated logout view to handle `RefreshToken` blacklisting.
- Implemented response interceptors in api calls to handle automatic token refresh on 401 responses.

Test Settings:
- cookie_auth_enabled=True
- cors_origin_allow_all=True
- frontend => Project.cookieAuthEnabled: true in `project_local.js`

Regression tests:
- Authentication
- Refresh token strategy forcing an expired access token
- Forced logout removing both access and refresh tokens
- Regular logout

Test run in usage page with hard refresh

Only additional calls are `/refresh` calls, proving that there were no impacts in the existing api calls

- Current
Screenshot 2025-03-12 at 17 07 57

- Using refresh token flow 
Screenshot 2025-03-12 at 16 59 05

_Please describe._


khvn26:
@tiagoapolo Added typing fixes — some tests need fixing now.

khvn26 (changes requested):
Having added Python typing fixes, I now see test failures in CI.

Added a couple comments, will dig into the meaty parts of simplejwt integration as soon as the tests start passing.

frontend/common/data/base/_data.js
```diff
@@ -73,9 +73,39 @@ module.exports = {
...
+
+    return originalFetch(url, options).then((response) => {
+      if (
+        isCookieAuthEnabled &&
+        response.status === 401 &&
+        !isExternal &&
+        !isExcluded
+      ) {
+        return this.post(`${Project.api}auth/token/refresh/`)
+          .then(() => {
```
khvn26: I appreciate this might be more work than it's worth, but could we implement some sort of shared state here to avoid excess `token/refresh` calls?

api/app/settings/common.py
```diff
@@ -821,13 +821,9 @@
...
-    "AUTH_TOKEN_CLASSES": ["rest_framework_simplejwt.tokens.SlidingToken"],
-    "SLIDING_TOKEN_LIFETIME": timedelta(
-        minutes=env.int(
-            "COOKIE_AUTH_JWT_ACCESS_TOKEN_LIFETIME_MINUTES",
-            default=10 * 60,
-        )
-    ),
+    "AUTH_TOKEN_CLASSES": ["rest_framework_simplejwt.tokens.AccessToken"],
+    "ACCESS_TOKEN_LIFETIME": timedelta(hours=1),
+    "REFRESH_TOKEN_LIFETIME": timedelta(days=1),
```
khvn26: I'm pretty sure we will want to make these configurable.

Consider adding support for `COOKIE_AUTH_JWT_REFRESH_TOKEN_LIFETIME_MINUTES` environment variable, with validation ensuring `REFRESH_TOKEN_LIFETIME > ACCESS_TOKEN_LIFETIME`.

frontend/common/data/base/_data.js
tiagoapolo: Yes, initially I tried a naive solution to have a shared state so we would not have an excess of `token/refresh`  api calls but had no success, I'll give it a new shot.

kyle-ssg: Yeah we shouldn't be relying on a 401 to do this, can probably just have a global rtk state, you can generate a slice with ``npx ssg action`` - the JWT can be decoded to figure out the expiry I imagine?

frontend/common/service.ts
```diff
@@ -1,49 +1,93 @@
...
+    ...queryArgs,
+  })
+
+  const baseQueryWithInterceptor = async (
+    args: string | FetchArgs,
+    api: BaseQueryApi,
+    extraOptions: {},
+  ) => {
+    const result = await baseQuery(args, api, extraOptions)
+    if (
```
kyle-ssg, line 53: Same as above

frontend/common/data/base/_data.js
khvn26: > the JWT can be decoded to figure out the expiry I imagine?

Probably not since we're required to use the HTTPOnly cookie. One thing we could do is to return the JWT in the body as well, should be fine as long as FE does not store it anywhere.

kyle-ssg: Ah yes right ofcourse, hmm yeah I think we'd need it returned somehow then. 

One alternative, have we actually just considered the backend doing the refreshing then in this case? It seems redundant for the FE to be involved in this at all. @khvn26

Zaimwa9: I see a couple of issues in moving the refreshing flow in the backend (assuming it would be in the validation service):
- There would still be the concurrency issues if the requests are triggered at the +/- same time as they would send the same cookie with same token. Just moved the problem there
- Cleaner and easier to observe/maintain distinct flows between validation and refreshing.

A lead could be some sort of Token manager class that would wrap a lock ? Triggered on 401
Something like:
```
TokenManager {
  static refreshLock: Promise<void> = null
  
  refreshJWTToken = async (client | fn): Promise {
    if (!refreshLock) {
      this.refreshLock = client.fetch(`${Project.api}auth/token/refresh/`, ...).finally(() => this.refreshTokenPromise = null)
    }
    return this.refreshLock
  }
}
```

```
if (response.status === 401) {
  await TokenManager.refreshJWTToken(this)
}
```

The concurrent calls should we waiting for refresh before triggering the original request or proceed if done (to be tested)

api/custom_auth/jwt_cookie/authentication.py
```diff
@@ -1,25 +1,40 @@
...
 
 class JWTCookieAuthentication(JWTAuthentication):
     def authenticate_header(self, request: Request) -> str:
         return f'Cookie realm="{self.www_authenticate_realm}"'
 
-    def authenticate(self, request: Request) -> tuple[FFAdminUser, Token] | None:
-        if raw_token := request.COOKIES.get(JWT_SLIDING_COOKIE_KEY):
+    def authenticate(self, request: Request) -> tuple[AuthUser, Token] | None:
+        raw_access_token = request.COOKIES.get(ACCESS_TOKEN_COOKIE_KEY)
+        raw_refresh_token = request.COOKIES.get(REFRESH_TOKEN_COOKIE_KEY)
```
matthewelwell:
```suggestion
```

api/custom_auth/jwt_cookie/authentication.py
```diff
@@ -1,25 +1,40 @@
...
+
+        if raw_access_token:
```
matthewelwell:
```suggestion
        if raw_access_token := request.COOKIES.get(ACCESS_TOKEN_COOKIE_KEY):
```

api/custom_auth/jwt_cookie/authentication.py
```diff
@@ -1,25 +1,40 @@
...
+            try:
+                # TODO https://github.com/jazzband/djangorestframework-simplejwt/pull/889
+                validated_access_token = self.get_validated_token(raw_access_token)  # type: ignore[arg-type]
+                # TODO https://github.com/jazzband/djangorestframework-simplejwt/pull/890
+                return self.get_user(validated_access_token), validated_access_token  # type: ignore[return-value]
+            except (InvalidToken, TokenError, AuthenticationFailed):
+                pass
+
+        if raw_refresh_token:
```
matthewelwell:
```suggestion
        if raw_refresh_token := request.COOKIES.get(REFRESH_TOKEN_COOKIE_KEY):
```

api/custom_auth/jwt_cookie/constants.py
```diff
@@ -1 +1,2 @@
-JWT_SLIDING_COOKIE_KEY = "jwt"
```
matthewelwell, line 1: I guess that removing this will mean that the authentication will fail for all users after they deploy the new version? The slightly bigger concern (but probably not worth worrying about) is that authentication will consistently fail for anyone that deploys the new version of the API without deploying the new version of the FE, right?

api/custom_auth/jwt_cookie/views.py
```diff
@@ -1,15 +1,63 @@
...
+        response.delete_cookie(ACCESS_TOKEN_COOKIE_KEY)
+        response.delete_cookie(REFRESH_TOKEN_COOKIE_KEY)
+        return response
+
+
+class JWTCookieTokenRefreshView(TokenRefreshView):
+    def post(self, request: Request, *args: Any, **kwargs: Any) -> Response:
+        raw_refresh_token = request.COOKIES.get(REFRESH_TOKEN_COOKIE_KEY)
+
+        if not raw_refresh_token:
```
matthewelwell:
```suggestion
        if not (raw_refresh_token := request.COOKIES.get(REFRESH_TOKEN_COOKIE_KEY)):
```

api/tests/unit/users/test_unit_users_custom_serializers.py
```diff
@@ -0,0 +1,22 @@
+from unittest.mock import MagicMock
+
+import pytest
+from rest_framework.exceptions import NotAuthenticated
+
+from users.models import FFAdminUser
+from users.serializers import CustomCurrentUserSerializer
+
+
+class TestCustomCurrentUserSerializer:
```
matthewelwell: Why are we using class based testing here?

api/users/serializers.py
```diff
@@ -161,6 +165,12 @@ class Meta(DjoserUserSerializer.Meta):  # type: ignore[misc]
             "uuid",
         )
 
+    def to_representation(self, instance: FFAdminUser) -> Any:
+        if not instance.is_authenticated:
+            raise NotAuthenticated("User is not authenticated.")
```
matthewelwell: Why do we need to raise this in the serializer?

frontend/common/data/base/_data.js
kyle-ssg: Yep ok, agree, if the backend isn't sending token expiry then above looks to be the best. And yeah good shout making the single static refreshLock promise

rolodato: We're not currently not using OIDC, but if we were, the backend could respond with an `expires_in` property as part of the authentication response to tell clients when a refresh is due. This is a performance optimisation to let clients preemptively refresh tokens before they expire, but it's not necessary and responding to a 401 as we're doing here is the preferred approach.

`expires_in` can be returned by any authentication method including using a refresh token. Search for `expires_in` here: https://openid.net/specs/openid-connect-core-1_0.html

The frontend should not decode the token because it can't validate it without having access to the shared signing secret, and because it's stored in an HttpOnly cookie anyway.

api/custom_auth/jwt_cookie/constants.py
tiagoapolo: Yes, that's right we need to deploy both frontend and backend. Also, I confirmed with Rodrigo and perhaps there's only 1 customer besides the one requiring this change that uses `COOKIE_AUTH_ENABLED`. Which leads me to believe deploying these changes won't have a big impact.

api/app/settings/common.py
```diff
@@ -845,14 +845,17 @@
...
-        )
+    "AUTH_TOKEN_CLASSES": ["rest_framework_simplejwt.tokens.AccessToken"],
+    "ACCESS_TOKEN_LIFETIME": timedelta(
+        hours=env.int("COOKIE_AUTH_JWT_ACCESS_TOKEN_LIFETIME", default=1)
+    ),
+    "REFRESH_TOKEN_LIFETIME": timedelta(
+        hours=env.int("COOKIE_AUTH_JWT_REFRESH_TOKEN_LIFETIME", default=24)
     ),
     "SIGNING_KEY": env.str("COOKIE_AUTH_JWT_SIGNING_KEY", default=SECRET_KEY),
+    "ROTATE_REFRESH_TOKENS": env.bool(
```
rolodato: Why is this an option instead of being always enabled?

api/app/settings/common.py
```diff
@@ -845,14 +845,17 @@
...
 SIMPLE_JWT = {
-    "AUTH_TOKEN_CLASSES": ["rest_framework_simplejwt.tokens.SlidingToken"],
-    "SLIDING_TOKEN_LIFETIME": timedelta(
-        minutes=env.int(
-            "COOKIE_AUTH_JWT_ACCESS_TOKEN_LIFETIME_MINUTES",
-            default=10 * 60,
-        )
+    "AUTH_TOKEN_CLASSES": ["rest_framework_simplejwt.tokens.AccessToken"],
+    "ACCESS_TOKEN_LIFETIME": timedelta(
+        hours=env.int("COOKIE_AUTH_JWT_ACCESS_TOKEN_LIFETIME", default=1)
```
rolodato, line 850: Do we need to make this breaking change or can we keep the existing option name?

api/custom_auth/jwt_cookie/authentication.py
```diff
@@ -1,25 +1,37 @@
...
-                return self.get_user(validated_token), validated_token  # type: ignore[return-value]
+                # TODO https://github.com/jazzband/djangorestframework-simplejwt/pull/889
+                validated_access_token = self.get_validated_token(raw_access_token)  # type: ignore[arg-type]
+                # TODO https://github.com/jazzband/djangorestframework-simplejwt/pull/890
+                return self.get_user(validated_access_token), validated_access_token  # type: ignore[return-value]
             except (InvalidToken, TokenError, AuthenticationFailed):
-                return None
+                pass
+
+        if raw_refresh_token := request.COOKIES.get(REFRESH_TOKEN_COOKIE_KEY):
```
rolodato: Clients should only ever send refresh tokens explicitly when wanting to obtain a refreshed access token. Using refresh tokens this way gives no security benefit over sliding tokens, since they're effectively the same thing. There's some possible security gain if we rotate refresh tokens but not enough to be meaningful.

The backend should not know anything about refresh tokens, except in the endpoints that accept and issue refresh tokens. All other API endpoints should only ever look at access tokens. For example, if we were building a pure OpenID Connect application, we would never implement anything about refresh tokens, because that's entirely the responsibility of the authorisation server and clients to deal with.

I can think of a few alternatives with their own tradeoffs to deal with this, but I'd like to get a shared understanding on the security goals/implications of this change first.

api/app/settings/common.py
tiagoapolo: It's already a breaking change because we're replace that auth method, so keeping the old token name won't help much

api/custom_auth/permissions.py
```diff
@@ -10,7 +10,7 @@ class CurrentUser(IsAuthenticated):
     """
 
     def has_permission(self, request, view):  # type: ignore[no-untyped-def]
-        return view.action == "me"
```
tiagoapolo, line 13: @matthewelwell just checking for `me` seems like it was implemented for a particular reason, do we not want to use the parent's `has_permission` check?

matthewelwell: image

Oof, it's kind of wild that I'm having to try and make sense of commits I made 5 years ago! 😄 

Err... looking at the behaviour here though, I'd say that it definitely should check the `super()` method though, yep.

api/users/serializers.py
```diff
@@ -6,13 +8,15 @@
...
-
-from .models import (
+from users.models import (
     FFAdminUser,
     UserPermissionGroup,
     UserPermissionGroupMembership,
 )
 
+if TYPE_CHECKING:
+    DjoserUserSerializer = DjoserUserSerializer[FFAdminUser]
```
matthewelwell, line 18: I'm intrigued as to why this has been added? I assume mypy threw some error without it?

api/app/settings/common.py
matthewelwell: I think we probably need to have a discussion between me, @tiagoapolo and @rolodato to understand the impact on existing customers before we release this PR then. @tiagoapolo can you arrange that please?

frontend/common/auth/TokenManager.ts
matthewelwell, line 1: Based on the fact that this code isn't used in SaaS (I don't think), we are probably safe to include both the FE and BE code in the same PR, but I'm just double checking that?

api/users/serializers.py
tiagoapolo: This was added by @khvn26 but I assume it was a mypy issue

frontend/common/data/base/_data.js
```diff
@@ -73,9 +75,38 @@ module.exports = {
         return response
       })
   },
+
   delete(url, data, headers) {
     return this._request('delete', url, data, headers)
   },
+  fetchWithInterceptor(url, options) {
+    const isExternal = !url.startsWith(Project.api)
+    const excludedUrls = [
```
kyle-ssg, line 84: Wouldn't auth endpoints be excluded too, e.g. oauth / login ?

api/custom_auth/jwt_cookie/services.py
```diff
@@ -1,18 +1,31 @@
...
+    REFRESH_TOKEN_COOKIE_KEY,
+)
 from users.models import FFAdminUser
 
 
 def authorise_response(user: FFAdminUser, response: Response) -> Response:
-    sliding_token = SlidingToken.for_user(user)
+    access_token = AccessToken.for_user(user)
+    refresh_token = RefreshToken.for_user(user)
     response.set_cookie(
```
Zaimwa9, line 15: The set_cookie block is similar to the one in the refreshView. It could be worth mutualizing them to avoid config discrepancies

https://github.com/Flagsmith/flagsmith/blob/7727cad76f7e2ebd646546e71fe1dc542a5baab2/api/custom_auth/jwt_cookie/views.py#L36-L57

api/custom_auth/jwt_cookie/views.py
```diff
@@ -1,15 +1,59 @@
...
+
+        response.delete_cookie(ACCESS_TOKEN_COOKIE_KEY)
+        response.delete_cookie(REFRESH_TOKEN_COOKIE_KEY)
+        return response
+
+
+class JWTCookieTokenRefreshView(TokenRefreshView):
+    def post(self, request: Request, *args: Any, **kwargs: Any) -> Response:
+        if not (raw_refresh_token := request.COOKIES.get(REFRESH_TOKEN_COOKIE_KEY)):
+            raise InvalidToken("No valid refresh token found in cookie")
```
Zaimwa9, line 34: Here we could mimic a logout and delete the cookie. As of we get stuck in a loop with an error (that I think would be silent on prod build) that needs a re-login

image

frontend/common/data/base/_data.js
```diff
@@ -73,9 +75,38 @@ module.exports = {
...
+        response.status === 401 &&
+        !isExternal &&
+        !isExcluded
+      ) {
+        return TokenManager.refreshJWTToken(fetch)
+          .then(() => {
+            return fetch(url, options)
+          })
+          .catch((error) => {
+            AppActions.setUser(null)
```
Zaimwa9, line 103: Based on the above error `no valid refresh token found in cookie` as a defensive move we could also call logout.
Because the kind of errors we could have:
- Non 401 auth errors => irrelevant because handled in the initial request
- 401 then refresh token error (inexistant or invalid) => logout
- 401 then refresh then 500  => In a perfect world we retry up to 3 times then I guess logout to unlock the user (but in case of massive outage we force re-login of all users)

frontend/common/service.ts
```diff
@@ -1,49 +1,76 @@
-import { createApi, fetchBaseQuery } from '@reduxjs/toolkit/dist/query/react'
+import {
```
Zaimwa9, line 1: Not critical and not related to patched files but I'm curious to know why and potential impacts

In account-store we have
```
  onLogin: (skipCaching) => {
    if (!skipCaching) {
      API.setCookie('t', Project.cookieAuthEnabled ? 'true' : data.token)
    }
    return controller.getOrganisations()
  },
```
(with a skipCaching that seems unused)

Now with access_token and refresh_token keys, it is set to true. Would it be worth to set the `access_token` in that case for consistency?

Only place it is used is in `WidgetPage` (no real impact)
```
  if (!API.getCookie('t')) {
    resolveAuthFlow({
      isAuthenticated: false,
    })
    return null
  }
```

and in `datadog-client` (same no big impact it seems)
```
const client: () => DDClient = () =>
  init({
    authProvider: {
      authStateCallback: async () => {
        const user = API.getCookie('t')
        return {
          isAuthenticated: !!user,
        }
      },
      resolution: 'message',

      /**
       * This where we want Datadog to direct users to authenticate.
       */
      url: '/',
    },
  })
  ```

Zaimwa9 (commented):
Couple of comments before hearing this topic was to be closed

matthewelwell:
Closing this based on comment [here](https://github.com/Flagsmith/flagsmith/issues/4688#issuecomment-2840124585).

PR #5222, opened 2025-03-12
feat: Each replica database URL can now be configured with an individual environment variable, using `REPLICA_DATABASE_URL_0`, `REPLICA_DATABASE_URL_1`, etc

rolodato:
In https://github.com/Flagsmith/flagsmith/pull/5199, I tried to improve the UX for configuring replica database URLs by accepting a JSON-encoded list of database URLs. This approach is not suitable for use with Helm.

In our [Helm values](https://github.com/Flagsmith/flagsmith-charts/blob/f7b6d91bbc326ba134602aa6a99ce3afaf3d182c/charts/flagsmith/values.yaml#L213), there are several ways of accepting database URLs:

* Plaintext URL
* Plaintext key/value config
* Secret reference

The problem is dealing with secret references. At deploy time, we cannot access the contents of Kubernetes secrets, so we cannot directly create a JSON-encoded list of URLs. We could imagine some workaround such as having an init container that dynamically creates this list, but this is more trouble than it's worth.

Instead, from Helm we can define each environment variable so that it gets its value from a secret. We can do this for any replica URLs that are passed in via secrets. See https://github.com/Flagsmith/flagsmith-charts/pull/337 for details.

We now accept multiple environment variables such as `REPLICA_DATABASE_URL_0`, `REPLICA_DATABASE_URL_1`, etc by starting from `0` and incrementing until we can't find any more variables.

No docs changes for now, as this is only meant to be used by the Helm chart for now.

Unit tests.


khvn26 (approved):
LGTM 👍

PR #5236, opened 2025-03-19
fix: Frontend can crash if query string is invalid

rolodato:
Trying to load a URL with a malformed query string currently crashes the frontend. For example (note the extra `=` at the end):

https://app.flagsmith.com/project/16945/environment/ihAHot7taBsh6hbHJKdYHs/features?is_archived=false&tag_strategy=INTERSECTION&page=1&search=&sortBy=name&sortOrder=asc=

Fails with:

```
Uncaught SyntaxError: JSON.parse: expected ',' or '}' after property value in object at line 1 column 110 of the JSON data
    fromParam 
```

At least one customer ran into this when using the ClearURLs Firefox add-on: https://addons.mozilla.org/en/firefox/addon/clearurls/

It's probably better to use a dedicated hook for accessing the query string instead of parsing it on every `fromParam` call, but this is a stop-gap solution to at least stop the frontend crashing.

Tested by trying to load an equivalent URL locally.


kyle-ssg:
I believe this was fixed in https://github.com/Flagsmith/flagsmith/pull/5235. Though if it's exactly the url mentioned 

> https://app.flagsmith.com/project/16945/environment/ihAHot7taBsh6hbHJKdYHs/features?is_archived=false&tag_strategy=INTERSECTION&page=1&search=&sortBy=name&sortOrder=asc=

Fails on both main and this PR since that would parse as sort_direction: ASC=
image

Issue #5224, opened 2025-03-12
Deprecate user-level permissions


rolodato:
### Is your feature request related to a problem? Please describe.

In 2.137.0, we deprecated group-level permissions for several reasons, which are described here: https://docs.flagsmith.com/system-administration/rbac#deprecated-features

For the same reasons, we should also deprecate user-level permissions. This is especially important now since tag-level permissions can only be applied to roles and not users. For example, user-level permissions look like this:

![Image](https://github.com/user-attachments/assets/71a4e682-5084-484e-804a-24da017b5263)

whereas editing role permissions offers more options:

![Image](https://github.com/user-attachments/assets/723b16a6-ee99-4e8c-a847-2bd589bc548d)

### Describe the solution you'd like.

Deprecate user-level permissions, add warnings in the UI when using them, and make them less prominent than role-level permissions.

### Describe alternatives you've considered

🤷

matthewelwell:
I think this needs discussions. I'm not currently convinced that deprecating user level permissions is the right idea. Perhaps we can encourage people to use roles, but I'm not sure we'd want to completely deprecate user level permissions.

rolodato:
The only benefit I can see for having user-level permissions is that it's easier to do ad-hoc/one-off permissions. This is also the thing that makes these permissions difficult to audit, since there's no single place to see which users have ad-hoc permissions outside of roles.

We could add more tooling to improve this situation like a detailed permissions debugger, or a single place to see all user-specific permissions, but this seems like effort that we could avoid by removing user-level permissions.

Roles also centralise the audit history for permissions, since you can (with some effort today, but definitely improvable) look at the history of what changed over time. With user-level permissions this is basically impossible or requires lots of tooling for it to be feasible.

IMO, being able to regularly review and baseline permissions is necessary for any enterprise-ready access control system. If you have 100+ users, and you need to audit every single user's permissions, that is too much effort and it does not scale.

PR #5247, opened 2025-03-25
feat: User permissions summary

### LINKED ISSUE #5237: UI improvements in project / environment permission configuration
Currently in the screen for managing user permissions for environments and projects, we show all users in the organisation, regardless of whether they have permissions to the environment / project. This can be confusing, and difficult to work with in organisations with larger numbers of users. 

In order to improve this permanently, we would likely need some UX design work, but it would be good to at least add some clarity in the UI on what 'Regular user' means. 

Please bear in mind that this may also be obsolete depending on the decision made in #5224 .

kyle-ssg:

## Changes

Whilst we decide on https://github.com/Flagsmith/flagsmith/issues/5224, this adds a stop-gap improvement where we summarise user-level permissions in the environment and project settings pages.

image

Caveats:
- This shows a tooltip explaining what No permissions means (i.e. there could be some at the group/role level)
- This will not show at the organisation settings level since here we show a list of projects / environments.

View permissions from project/environment settings


frontend/web/components/EditPermissions.tsx
```diff
@@ -1176,29 +1189,51 @@ const EditPermissions: FC<EditPermissionsType> = (props) => {
...
+                                      </Tooltip>
+                                    </div>
+                                  ) : (
+                                    <div onClick={onClick}>
+                                      {!matchingPermissions?.permissions
+                                        ?.length ? (
+                                        <Tooltip
+                                          title={
+                                            <div className='text-body gap-1 align-items-center d-flex'>
+                                              No Permissions{' '}
```
matthewelwell: Maybe we can be more explicit here and say something like 'No User-level Permissions' ? I'm conscious that the content in the tooltip (copied below) still sounds like we're suggesting that permissions do exist for the user, but then the table data suggests that they have no permissions. 

> Permissions for this user may exist at the Group or Role Level

Maybe we could change the tooltop content to

```
This user has no permissions assigned directly, but they may belong to a group that has permissions on this environment|project. 
```

kyle-ssg: Yep agree, it's a confusing area without us having a permissions debugger so being more explicit makes sense.

matthewelwell:
I can't see where the extra commas are coming from but there's some issue with the content in the tooltip: 

image

Also, there's an example [here](https://flagsmith-frontend-preview-git-feat-user-permi-2fe507-flagsmith.vercel.app/project/12/settings?tab=permissions) of a user (Daniel) that is marked as having no user-level permissions, but is seemingly an administrator when you click into his user permissions? 

image

image

tiagoapolo (approved):

frontend/web/components/EditPermissions.tsx
```diff
@@ -1176,29 +1190,63 @@ const EditPermissions: FC<EditPermissionsType> = (props) => {
...
+                                <div className='d-flex  gap-2 table-column fs-small'>
+                                  {role === 'ADMIN' ? (
+                                    <div className=''>
+                                      <Tooltip
+                                        title={
+                                          <PermissionSummaryItem isAdmin />
+                                        }
+                                      >
+                                        {
+                                          'Organisation administrators have all permissions enabled.<br/>To change the role of this user, visit Organisation Settings.'
```
matthewelwell: This tool tip looks a little odd with the single line break. We should either remove it, or add another imo. 

```suggestion
                                          'Organisation administrators have all permissions enabled. To change the role of this user, visit Organisation Settings.'
```

matthewelwell (approved):
Approved with a minor formatting comment.

gagantrivedi:
Is this same as: https://github.com/Flagsmith/flagsmith/pull/5375?

kyle-ssg:
Closing as yes I think we should remove these now we have the permissions inspect

PR #5238, opened 2025-03-20
fix: Syntax error in Python "Try it" code snippets

rolodato:
Tested locally:

image

image

PR #5252, opened 2025-03-26
fix: Syntax error in NodeJS "Try it" code snippets

joscha:
See:

SCR-20250326-orlg

Similar to #5238, but for NodeJS // cc @rolodato

### DIFF
--- a/frontend/common/code-help/create-user/create-user-node.js
+++ b/frontend/common/code-help/create-user/create-user-node.js
@@ -15,12 +15,12 @@ const ${LIB_NAME} = new Flagsmith({${
 });
 
 // Optional - set traits for this identity
-const traitList = { ${TRAIT_NAME}: 42 };
+const traits = { ${TRAIT_NAME}: 42 };
 
 // Identify the user
 const flags = await flagsmith.getIdentityFlags('${
   userId || USER_ID
-}', traitList);
+}', traits);
 
 // get the state / value of the user's flags 
 const isEnabled = flags.isFeatureEnabled('${FEATURE_NAME}');
--- a/frontend/common/code-help/create-user/create-user-node.js
+++ b/frontend/common/code-help/create-user/create-user-node.js
@@ -2,17 +2,20 @@ import Constants from 'common/constants'
 
 module.exports = (
   envId,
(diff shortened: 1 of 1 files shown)


frontend/common/code-help/create-user/create-user-node.js
```diff
@@ -2,17 +2,20 @@ import Constants from 'common/constants'
...
+    : ''
 }
     environmentKey: '${envId}'
 });
 
+// Optional - set traits for this identity
+const traitList = { ${TRAIT_NAME}: 42 };
 
 // Identify the user
 const flags = await flagsmith.getIdentityFlags('${
```
joscha, line 21: further down here, there are two `var` statements, which should probably at the very least be `let` but most likely `const`

rolodato: Using `const` would be fine - could you please add that to this PR? 🙏

joscha: Changed in 6bcad7268b4d7f99af192d697575a486c93eea9b

frontend/common/code-help/create-user/create-user-node.js
```diff
@@ -2,24 +2,27 @@ import Constants from 'common/constants'
...
-  `\n    apiUrl: '${Constants.getFlagsmithSDKUrl()}',`
+  Constants.isCustomFlagsmithUrl()
+    ? `\n    apiUrl: '${Constants.getFlagsmithSDKUrl()}',`
+    : ''
 }
     environmentKey: '${envId}'
 });
 
+// Optional - set traits for this identity
+const traitList = { ${TRAIT_NAME}: 42 };
```
matthewelwell: Being nitpicky now, but can we rename this variable since that's not a list.

matthewelwell (changes requested):

frontend/common/code-help/create-user/create-user-node.js
joscha: d2452bec8918ab89a7ddf6617775c91a25cb00a5

matthewelwell (approved):

matthewelwell:
Thanks @joscha !

PR #5254, opened 2025-03-26
feat: Initial Prometheus support

khvn26:

## Changes

This enables Prometheus support with new settings:
- `PROMETHEUS_ENABLED` controls existence of the `/metrics` endpoint. `False` by default.
- `PROMETHEUS_HISTOGRAM_BUCKETS` controls sizes and count of Prometheus histogram buckets. This is always a highly demanded customisation as default Prometheus buckets rarely fit to real world use cases.

Refer to `flagsmith-common` [documentation](https://github.com/Flagsmith/flagsmith-common?tab=readme-ov-file#metrics) to learn about existing metrics and how to add new ones. 

We're going to use this PR's Docker image to compile a reference Grafana dashboard. The main functionality is unit-tested in the `flagsmith-common` repo.

To test manually, run `flagsmith start api` and `flagsmith start task-processor`; HTTP server and build info metrics are not currently supported when running the Django development server.


api/app/settings/common.py
```diff
@@ -1348,5 +1350,10 @@
 if LICENSING_INSTALLED:  # pragma: no cover
     INSTALLED_APPS.append("licensing")
 
-PROMETHEUS_ENABLED = True
-PROMETHEUS_HISTOGRAM_BUCKETS = prometheus_client.Histogram.DEFAULT_BUCKETS
+PROMETHEUS_ENABLED = env.bool("PROMETHEUS_ENABLED", True)
```
rolodato: > PROMETHEUS_ENABLED controls existence of the /metrics endpoint. True by default. Maybe we should make it False by default to cater for existing insecure deployments out there.

Came here to comment this - this should be disabled by default. It'll be enabled by default every else for sure (Docker Compose, Helm chart etc)

khvn26: Fixed in 13bbef1ca2e93dae4df7034f95993fae369666b8.

api/scripts/run-docker.sh
```diff
@@ -25,34 +24,33 @@ serve() {
 
     waitfordb
 
-    exec gunicorn --bind 0.0.0.0:8000 \
+    exec flagsmith start \
```
rolodato, line 27: Should we adapt the Makefile as well to use `flagsmith start` instead of `gunicorn`?

rolodato (commented):
LGTM, but I actually can't get this to work locally. Installed dependencies using:

```
poetry install --with dev --no-root
```

then running the API with:

```
PROMETHEUS_MULTIPROC_DIR=/tmp PROMETHEUS_ENABLED=true python manage.py runserver
```

The `/metrics` endpoint always returns 200 with no body, even after I hit an endpoint like `/flags`.

Also, if I do not set `PROMETHEUS_MULTIPROC_DIR=/tmp`, the API still starts but `/metrics` fails with a 500:

```
django.request ERROR    Internal Server Error: /metrics
  File "/Users/rolodato/source/flagsmith/flagsmith/api/.venv/lib/python3.12/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/rolodato/source/flagsmith/flagsmith/api/.venv/lib/python3.12/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/rolodato/source/flagsmith/flagsmith/api/.venv/lib/python3.12/site-packages/common/core/views.py", line 22, in metrics
    registry = get_registry()
               ^^^^^^^^^^^^^^
  File "/Users/rolodato/source/flagsmith/flagsmith/api/.venv/lib/python3.12/site-packages/common/prometheus/utils.py", line 17, in get_registry
    MultiProcessCollector(registry)  # type: ignore[no-untyped-call]
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/rolodato/source/flagsmith/flagsmith/api/.venv/lib/python3.12/site-packages/prometheus_client/multiprocess.py", line 30, in __init__
    raise ValueError('env PROMETHEUS_MULTIPROC_DIR is not set or not a directory')
```

khvn26:
@rolodato Thanks for catching this — I fixed the `manage.py` usage in 52bdb239a490028e2e23e832f482feb932408261 (`flagsmith runserver` should work as well). 

However, the `http_server_` metrics will not show up when using the development server since they rely on instrumented Gunicorn. The `flagsmith_build_info` metric will not show up as well because it relies on the new `flagsmith start` entrypoint. Please use `flagsmith start api` when testing out the HTTP metrics.

Updated the PR description to hopefully make it clearer.

api/scripts/run-docker.sh
khvn26: Addressed in 3e29cffe52aba4d2cd46b38181321f98573f5c85.

api/scripts/run-docker.sh
```diff
@@ -25,34 +24,33 @@ serve() {
...
 run_task_processor() {
     waitfordb --waitfor 30 --migrations
     if [ -n "$ANALYTICS_DATABASE_URL" ] || [ -n "$DJANGO_DB_NAME_ANALYTICS" ]; then
         waitfordb --waitfor 30 --migrations --database analytics
     fi
-    RUN_BY_PROCESSOR=1 python manage.py runprocessor \
+    exec flagsmith start \
+             --bind 0.0.0.0:8000 \
+             --access-logfile $ACCESS_LOG_LOCATION \
+             --access-logformat "$ACCESS_LOG_FORMAT" \
```
matthewelwell, line 47: I see that this was there before, but I'm not sure why we necessarily need access logs for the task processor?

docker-compose.yml
matthewelwell, line 1: Looks like there are some temporary changes to revert in this file?

matthewelwell (changes requested):

docker-compose.yml
khvn26: Fixed in cc83ce5d4efa3e2cdd6cf9f814ac8f43c387566d.

api/scripts/run-docker.sh
khvn26: Task proccessor now serves `/version`, /health endpoints and, if enabled, the `/metrics` endpoint. The access log covers these.

docker-compose.yml
matthewelwell: Great, I'll approve now, but we'll also need to revert the images still.

matthewelwell (approved):
Approved but image references in compose need reverting still.

docker-compose.yml
khvn26: Additionally, 4f675e7f554ae08a5b752e98fd596a79e3054049 reverts image names to docker.flagsmith.com based ones.

docker-compose.yml
```diff
@@ -21,7 +21,7 @@ services:
       start_period: 20s
 
   flagsmith:
-    image: docker.flagsmith.com/flagsmith/flagsmith:latest
+    image: gdocker.flagsmith.com/flagsmith/flagsmith:latest
```
matthewelwell:
```suggestion
    image: docker.flagsmith.com/flagsmith/flagsmith:latest
```

matthewelwell (changes requested):

docker-compose.yml
matthewelwell: Not quite 😄 https://github.com/Flagsmith/flagsmith/pull/5254#discussion_r2034858055

docker-compose.yml
```diff
@@ -63,11 +64,15 @@ services:
 
   # The flagsmith_processor service is only needed if TASK_RUN_METHOD set to TASK_PROCESSOR
   # in the application environment
-  flagsmith_processor:
+  flagsmith-task-processor:
     image: docker.flagsmith.com/flagsmith/flagsmith:latest
     environment:
       DATABASE_URL: postgresql://postgres:password@postgres:5432/flagsmith
       USE_POSTGRES_FOR_ANALYTICS: 'true'
+      DJANGO_ALLOWED_HOSTS: '*'
```
matthewelwell, line 72: I agree that putting this in for now is reasonable, but I am hoping there is probably a better way to handle it such that we don't need to define it for the task processor?

khvn26: I believe the best way is to use `["localhost:8000", "127.0.0.1:8000"]` as the default value for this setting.

Let me know if you want it as part of this PR.

docker-compose.yml
khvn26: Thanks! 84ad821fcd5274a6ba48080b43cb059114599963.

docker-compose.yml
matthewelwell: No, let's do that separately.

matthewelwell (approved):

PR #5057, opened 2025-01-29
feat: self hosted onboarding

kyle-ssg:
Thanks for submitting a PR! Please check the boxes below:

- [ ] I have added information to `docs/` if required so people know about the feature!
- [x] I have filled in the "Changes" section below?
- [x] I have filled in the "How did you test this code" section below?
- [x] I have used a [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) title for this Pull Request

## Changes

https://github.com/user-attachments/assets/2788fc29-90e8-4582-8b6f-90f7bd26a70e

Adds an onboarding flow for new self-hosted instances.

- Registers the user and creates the organisation in 1 workflow,
- Migrates version endpoint and homepage to RTK in order to determine if onboarding is necessary
- Adds resources panel for company self-hosted emails, allows initialising crisp
- Sends more information to crisp

- Todo: Add an approach to detecting onboarding needs to happen in the versions endpoint
- If we are to link to getting started docs, we just need a url for them to live in
- If there's any other info we might want to gather it'd need specifying 

## How did you test this code?

- Run against the latest docker image with clean data
- Delete the auto-created superuser (FFAdminUser.objects.all().delete()), this will be removed from docker compose prior to the next release
- Go to localhost:8080
- Validate superuser is created and contact consent is sent over (this posts to onboarding/request/send, pending https://github.com/Flagsmith/flagsmith/pull/5331)

PR #5266, opened 2025-03-31
feat(superuser): Allow super user creation using signup endpoint

gagantrivedi:

## Changes

Allow superuser creation through signup endpoint if no other user exists 
part of [onboarding flow](https://github.com/Flagsmith/flagsmith/pull/5057)

Adds unit tests

### DIFF
--- a/api/custom_auth/serializers.py
+++ b/api/custom_auth/serializers.py
@@ -7,6 +7,7 @@ from rest_framework.authtoken.models import Token
 from rest_framework.exceptions import PermissionDenied
 from rest_framework.validators import UniqueValidator
 
+from common.core.utils import is_saas
 from organisations.invites.models import Invite, InviteLink
 from users.auth_type import AuthType
 from users.constants import DEFAULT_DELETE_ORPHAN_ORGANISATIONS_VALUE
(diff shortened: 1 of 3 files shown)


api/custom_auth/serializers.py
```diff
@@ -57,6 +57,8 @@ class Meta(UserCreateSerializer.Meta):  # type: ignore[misc]
             "is_active",
             "marketing_consent_given",
             "uuid",
+            "is_superuser",
+            "is_staff",
```
matthewelwell: In my head we were going to implicitly do this if no other user existed in the database. What's the logic behind exposing these fields?

gagantrivedi: > In my head we were going to implicitly do this if no other user existed in the database.

No, the plan was to create another endpoint explicitly for superuser creation via the UI... Making the first user a superuser implicitly feels a bit odd

matthewelwell: Fair enough. I think I'd rather though handle the additional complexity on the BE side, instead of exposing the django fields themselves. Perhaps it's overkill, but I think we could do something like: 

```python
# serializers.py

class CustomerUserCreateSerializer(...):
    ...

    superuser = serializers.BooleanField(write_only=True)
```

```python
# models.py

class FFAdminUser(...):
    ...
    
    @property
    def superuser(self) -> bool:
        return self.is_staff and self.is_superuser

    @superuser.setter
    def set_superuser(self, value: bool) -> None:
        self.is_staff = value
        self.is_superuser = value
```

What do you think?

gagantrivedi: Yeah, I agree

matthewelwell (approved):

PR #5283, opened 2025-03-31
feat: If-Modified-Since for /environment-document

rolodato:
Adds support for `If-Modified-Since` to `GET /environment-document`. This makes it so that if clients already have a recent environment document, the API can respond with an empty 304 instead of serving the whole document.

Tests are not yet passing. I understand this is because the environment document cache uses SDK keys and not environment IDs as its cache keys, which means that calling `clear_environment_cache` has no effect if the environment was only ever fetched using server-side SDK keys.


api/environments/authentication.py
```diff
@@ -41,3 +42,19 @@ def authenticate(self, request):  # type: ignore[no-untyped-def]
...
+
+class AuthenticatedRequest(HttpRequest):
+    _environment: Environment
+
+    @property
+    def environment(self) -> Environment:
+        if not self._environment:
+            raise AssertionError(
+                "Tried to access environment on an authenticated request, but was None"
+            )
```
matthewelwell: Maybe this is intentionally verbose, but this can be rewritten as 

```suggestion
        assert self._environment, (
            "Tried to access environment on an authenticated request, but was None"
        )
```

api/tests/unit/environments/test_unit_environments_views_sdk_environment.py
```diff
@@ -162,3 +163,48 @@ def test_get_environment_document_fails_with_invalid_key(
     # We get a 403 since only the server side API keys are able to access the
     # environment document
     assert response.status_code == status.HTTP_403_FORBIDDEN
+
+
+def test_environment_document_caching(
```
matthewelwell: I'd argue that this isn't really 'caching' ?

api/tests/unit/environments/test_unit_environments_views_sdk_environment.py
```diff
@@ -162,3 +163,48 @@ def test_get_environment_document_fails_with_invalid_key(
...
+
+    # When - second request with If-Modified-Since header
+    client.credentials(
+        HTTP_X_ENVIRONMENT_KEY=api_key,
+        HTTP_IF_MODIFIED_SINCE=http_date(environment.updated_at.timestamp()),
+    )
+    response2 = client.get(url)
+
+    # Then - second request should return 304 Not Modified
+    assert response2.status_code == status.HTTP_304_NOT_MODIFIED
```
matthewelwell, line 212: Could we also make a request from a client that doesn't have the `If-Modified-Since` header to make sure that it still returns the environment document?

matthewelwell (commented):
The reason that the tests are failing is because the `updated_at` value on the environment is updated by a hook on an audit log record getting created (it is not using the usual `auto_now` logic from django). Since you're updating the name of the environment manually, an audit log record isn't being created (because there's no user associated with the change). 

Some options: 

 1. Just manually update the `updated_at` value on the environment instance
 2. Convert this test into more of an integration test and actually make an update to the environment via the PUT endpoint for example

I'm happy with either option really so I'd recommend just going with option 1 as it's the easiest. As per my comment on the specific lines in the code though, we should add a test to verify that not passing the if-modified-since header still returns the whole document.

api/tests/unit/environments/test_unit_environments_views_sdk_environment.py
rolodato: Added in https://github.com/Flagsmith/flagsmith/pull/5283/commits/957e6ff66cccc667fe6203e628648806fe0e02be

api/tests/unit/environments/test_unit_environments_views_sdk_environment.py
rolodato: Fixed in https://github.com/Flagsmith/flagsmith/pull/5283/commits/4084a466852ad88f0a212e2798762d1fae1d8dc0

matthewelwell (approved):

Issue #5280, opened 2025-03-31
Styling issue for flags with longer string values


matthewelwell:
As can be seen in the following screenshot, the toggle overlaps with the string value in the features table. 

<img width="1223" alt="Image" src="https://github.com/user-attachments/assets/84bde86b-545b-4de4-b20d-e36dde349eb4" />

PR #5295, opened 2025-04-03
fix: Feature value overflowing container

tiagoapolo:

## Changes

ref: [#5280](https://github.com/Flagsmith/flagsmith/issues/5280)

- Fixes FeatureValue overflowing container

#### Features Page
Before
Screenshot 2025-04-03 at 16 10 13

After
Screenshot 2025-04-03 at 16 09 55

#### Compare Environments
Before
Screenshot 2025-04-03 at 12 06 48

After
Screenshot 2025-04-03 at 13 31 01

#### Compare Identities
Before
Screenshot 2025-04-03 at 16 06 16

After
Screenshot 2025-04-03 at 16 06 32

Tested in all places that implement FeatureValue component

### DIFF
--- a/frontend/web/components/FeatureValue.tsx
+++ b/frontend/web/components/FeatureValue.tsx
@@ -1,6 +1,5 @@
 import { FC } from 'react'
 import { FlagsmithValue } from 'common/types/responses'
-import Format from 'common/utils/format'
 import Utils from 'common/utils/utils'
 import { getViewMode } from 'common/useViewMode'
 import classNames from 'classnames' // we need this to make JSX compile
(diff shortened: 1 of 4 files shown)


frontend/web/components/FeatureValue.tsx
```diff
@@ -23,22 +23,25 @@ const FeatureValue: FC<FeatureValueType> = (props) => {
...
-        )}
+      <span style={{ overflow: 'hidden', textOverflow: 'ellipsis' }}>
+        {type == 'string' && <span className='quot'>"</span>}
+        <span className='feature-value'>
+          {Format.truncateText(
+            `${Utils.getTypedValue(props.value)}`,
+            isCompact ? 24 : 20,
+          )}
+        </span>
+        {type == 'string' && <span className='quot'>"</span>}
```
Zaimwa9: It looks like this end quote gets ellipsed too when there is no separator in the value.
I think that moving the overflow in the feature-value span along with a `flexWrap: nowrap` in the div and removing the extra-span could fix it (and maybe the little width gap we see in between the 2 ellipsed values in the screenshot).

Something like
```
<div ... >
    {type == 'string' && <span className='quot'>"</span>}
    <span className='feature-value' style:{{ ellipsis }}>
      ....
      </span>
    {type == 'string' && <span className='quot'>"</span>}
```

Wdyt ?

Zaimwa9 (changes requested):
Small detail to check, let me know if worth otherwise lgtm 👍

Zaimwa9 (approved):
Thanks for the update 🙏

PR #5302, opened 2025-04-04
fix: Prevent the frontend from crashing if Chargebee.js failed to load

rolodato:

## Changes

To reproduce this issue, block the `https://js.chargebee.com/v2/chargebee.js` request from your browser and try to load the Projects listing page:

image

We now log an error and don't crash the frontend if this request fails.

Tested by blocking the same request and verifying an error is logged, and the frontend doesn't crash.


tiagoapolo (approved):
nice!

PR #5306, opened 2025-04-07
docs: Add Edge Proxy liveness and readiness checks

rolodato:


docs/docs/deployment/hosting/locally-edge-proxy.md
```diff
@@ -245,14 +245,17 @@ Or, log access logs to file in generic format while logging everything else to s
...
-fetch all its configured environment documents. If any environment document could not be fetched during a configurable
-grace period, the health check will fail with a 500 status code. This allows the Edge Proxy to continue reporting as
-healthy even if the Flagsmith API is temporarily unavailable.
+The Edge Proxy exposes two health check endpoints:
 
-You can point your orchestration health checks to this endpoint.
+* `/proxy/health/liveness`: Always responds with a 200 status code. Use this health check to determine if the Edge
+  Proxy is alive and able to respond to requests.
+* `/proxy/health/readiness`: Responds with a 200 status if the Edge Proxy was able to fetch all its configured
+  environment documents within a configurable grace period. This allows the Edge Proxy to continue reporting as healthy
```
Zaimwa9, line 255: What do you think about keeping the "otherwise the health check will fail with a 500 status code" just to be exhaustive @rolodato ?

rolodato: I intentionally left this out, as IMO customers should not intentionally expect a 500 or any other status code. Any non-200 response should be considered a failure.

In https://github.com/Flagsmith/edge-proxy/pull/151, we changed the health checks to return a 503 instead of 500. This can help distinguish between an error happening in the health check itself, or the health check completing but failing. It's a subtle difference but can help distinguish is the Edge Proxy is the one actually responding with an expected failure, or if it's an unexpected failure from an upstream component such a proxy or load balancer.

Zaimwa9: Noted!

Zaimwa9 (approved):
All good for me then, thanks for the explanation!

Issue #123, opened 2021-06-16
Integer Traits and Feature State Values should be stored at least 64bit


dabeeeenster:
Currently integer traits and Feature State Values in postgres top out at 2147483647. We need to change the data type that is used to store the value in postgres. 

In addition, the front end should bubble up these error messages to the user if the values cannot be stored for whatever reason.

kyle-ssg:
As it stands the API does not return an error that can be parsed by the frontend, assign back to me when it can. 

The frontend will also now start treating numbers outside of the safe range as a string, however the API fails with a numeric string as it seems to cast it to an integer, resulting in a similar "integer out of range" error.

![image](https://user-images.githubusercontent.com/8608314/129481574-877e444a-c95a-4463-b3d6-4229044d4490.png)

![image](https://user-images.githubusercontent.com/8608314/129481599-21983f22-d555-4b30-894a-83b3047eba82.png)

matthewelwell:
I have been through and tested all of the scenarios on 2.169.2 in which using a numeric value greater than the maximum size of a 32-bit integer causes an error to be thrown: 

 1. When creating a new feature with an initial value

Cases I tested where the FE automatically converts the value to a string (and the BE leaves it as a string): 

 1. Updating a feature value in a versioned or a non-versioned environment
 2. Updating the value of a segment override in a versioned or non-versioned environment
 3. Update the value of an identity override in a versioned or a non-versioned environment, in SaaS (i.e. edge)
 4. Creating and publishing a change request for the environment value in a versioned or a non-versioned environment
 5. Creating a publishing a change request for a segment override in a versioned environment
 6. Creating an identity override in a versioned or a non-versioned environment, self-hosted

I have therefore created #5314 to resolve the one outstanding case. As per the description on that PR, the end goal is still to move away from storing values as anything other than strings, but this is a temporary solution that will work until we build API V2 and move to using strings and explicit type definitions everywhere.

PR #5314, opened 2025-04-08
fix: convert large integers to strings

matthewelwell:
## Changes

Related issue: https://github.com/Flagsmith/flagsmith/issues/123

This PR is a step in the right direction, but still needs further resolution as defined in https://github.com/Flagsmith/flagsmith/issues/1667. This simply replicates the behaviour already implemented by the FE in all cases except this one (because the conversion is happening on the backend). 

Updated an existing unit test.


api/features/utils.py
```diff
@@ -8,34 +11,38 @@ def get_value_type(value: str) -> str:
...
     e.g. "12" -> "int", "12.34" -> "float", etc.
     """
-    if is_integer(value):  # type: ignore[no-untyped-call]
+    if is_32_bit_integer(value):
         return INTEGER
-    elif is_boolean(value):  # type: ignore[no-untyped-call]
+    elif is_boolean(value):
         return BOOLEAN
     else:
         return STRING
```
Zaimwa9, line 19: We could use an enum for the feature value types available for stronger typing maybe with the ones in `value_types.py`
```
INTEGER = "int"
STRING = "unicode"
BOOLEAN = "bool"
FLOAT = "float"
```

matthewelwell: Yeah, agreed. I kind of hate enums in python but let me see what it looks like if I replace these strings with them.

api/tests/unit/features/test_unit_features_utils.py
```diff
@@ -8,6 +9,7 @@
     "value, expected_type",
     (
         ("1", INTEGER),
+        (f"{MAX_32_BIT_INTEGER + 1}", STRING),
```
Zaimwa9, line 12: Maybe super edge case but are we safe with the negative limit (-MAX_32_BIT_INTEGER) ?
I ran the tests with 
```
        (f"{-MAX_32_BIT_INTEGER - 100}", STRING),
```
It fails (returning integer) and I wonder if it would break ?

Zaimwa9 (changes requested):
One question to address, once resolved (needed or not) looks good to go

api/tests/unit/features/test_unit_features_utils.py
matthewelwell: Good catch! Fixed [here](https://github.com/Flagsmith/flagsmith/pull/5314/commits/7595c67a3a93ff164e426deae23167f5cd454644), and test updated.

api/features/utils.py
matthewelwell: This turned into a pretty large refactor that I'm not particularly keen on including in this PR (see [here](https://github.com/Flagsmith/flagsmith/pull/5314/commits/8193baa423fcc8aa14a9681bd16132656cc604b2)). It might be something to add in a separate PR, however.

Zaimwa9: Alright, just to confirm, do you need to revert the commit with enum or can I test the PR with it included ?

matthewelwell: Yeah, I'll revert that commit, and probably submit a separate PR with just that commit in it.

matthewelwell: I mostly just pushed it here as a bit of a 'show and tell' 🙂

Zaimwa9 (approved):
Thanks for the changes

matthewelwell:
Note:

I have manually tested the following scenarios in staging and validated that it works as expected: 

 1. Create a feature with a numeric value larger than the max size and verify that it is created successfully, and the value is cast to a string
 2. Create a feature with a numeric value smaller than the max size and verify that it is created successfully, and the value is not cast to a string (it is an integer)
 3. Create a new environment in a project that has a feature with an initial value which is numeric and larger than the max size, and verify that the new environment is correctly created, with the feature state, and the feature state's value is cast to a string

PR #5325, opened 2025-04-09
feat: MetadataModelRequirement can accept organization as content-type

Zaimwa9:
Thanks for submitting a PR! Please check the boxes below:

- [ ] I have added information to `docs/` if required so people know about the feature!
- [x] I have filled in the "Changes" section below?
- [x] I have filled in the "How did you test this code" section below?
- [x] I have used a [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) title for this Pull Request

## Changes
The goal of this PR is to prepare the API to handle organisation level custom metadata fields

- implemented organisation content type logic in `MetadataModelField.validate` to allow organisation level requirements
- Made object_id optional

## How did you test this code?
- Added 3 new tests on the serializer
- Ran test
- Conducted tests with related frontend (#5326) 

Pre-requisite for #5326

PR #5326, opened 2025-04-09
feat: MetadataModelRequirement uses organisation ID as objectID  (✅ relies on #5325 - deployed)

### LINKED ISSUE #5167: Project ID should not be required when creating/updating /metadata-model-fields

- [ ] Self Hosted with Docker
- [x] SaaS at flagsmith.com

### Describe the bug

When creating or updating a custom field through the /metadata-model-fields endpoint, the project_id parameter should not be required. This is because custom fields with a required entity are automatically enforced at the organization level. The object_id field in the request body already holds the project ID, as shown [here](https://github.com/Flagsmith/flagsmith/blob/7b9182dc75f63739fd85dc6cdc6af26ecba3bfb5/api/metadata/serializers.py#L53). This change is necessary to complete [issue #4384](https://github.com/Flagsmith/flagsmith/issues/4384).

### Steps To Reproduce

1. Go to project settings
2. Create custom field with a required entity

### Expected behavior

When creating/updating a custom field we don't need to provide the project id 

### Screenshots

https://github.com/user-attachments/assets/270335f3-01cb-4e00-ae15-f6522c6c6353

Zaimwa9:

## Changes
⚠️ Relies on #5325
Related to issue #5167

@tiagoapolo , following discussion with @kyle-ssg , we went for an in-between solution so that the required_for points to an organization while leaving the project implementation open for later. In this case it makes to relate the `required_for` field to the given organisation. Let me know your thoughts

- Use organisation `content_type` in `required_for` 
- Pass organisation_id as `object_id`

https://www.loom.com/share/f17075de90744c64bc64554629ed3eca

- Create custom field features in project settings
- Switch projects to see them everywhere

## Additional considerations
There are some points to keep in mind in the current metadata fields implementation. Just listing and depending on prio because quite some refacto.
As of Fields and Model fields are fetched with 2 different requests `GET /fields` and `GET /metadata-model-fields` then reconciliated in the frontend ( `MetadataPage` ) and it brings several problems long term:
- The 2 endpoints have their own pagination. If ever we reach the limits, it will be very hard to synchronize the 2 lists and be sure there are no misses. So no clean pagination possible as of
- It would be better to let the API handle the joins and the frontend just consumes them
- 4 requests are made to create a single field `create metadata field -> invalidated GET list -> create metadata model field -> invalidated GET list`, it could be reduced imo
- It's prone to error. It can happen that the metadata field is created but the second POST fails (metadata-model-field) and we end up with half of the data without a cleaning mechanism. Same I think we could have a single request creating both


tiagoapolo:
Great! 🎉

frontend/web/components/modals/CreateMetadataField.tsx
```diff
@@ -88,9 +86,11 @@ const CreateMetadataField: FC<CreateMetadataFieldType> = ({
   const [updateMetadataModelField] = useUpdateMetadataModelFieldMutation()
 
   const [deleteMetadataModelField] = useDeleteMetadataModelFieldMutation()
-  const projectContentType: ContentType =
-    supportedContentTypes &&
-    Utils.getContentType(supportedContentTypes, 'model', 'project')
+  const contentTypeKey = 'organisation'
```
tiagoapolo: Do you think it makes sense to make `'organisation'` a constant [here]?(https://github.com/Flagsmith/flagsmith/blob/main/frontend/common/constants.ts)

tiagoapolo (approved):
I left a suggestion but overall looks good to me and I'm aligned with that approach.

frontend/web/components/modals/CreateMetadataField.tsx
Zaimwa9: You're right, it adds no value here. I started an implem passing the scope (project or organisation) but as it's not implemented now I removed so it's properly done once needed.

Just pushed using an enum. I left it in the file, is there any other location that would make more sense ?

PR #5330, opened 2025-04-10
fix: `RouteLoggerMiddleware` breaks threads accessing Django request object

### LINKED ISSUE #5327: `RouteLoggerMiddleware` breaks threads accessing Django request object
Sentry Issue: [FLAGSMITH-API-54W](https://flagsmith.sentry.io/issues/6531478558/?referrer=github_integration)

```
RuntimeError: dictionary changed size during iteration
(1 additional frame(s) were not displayed)
...
  File "app_analytics/track.py", line 41, in track_request_influxdb_async
    return track_request_influxdb(request)  # type: ignore[no-untyped-call]
  File "app_analytics/track.py", line 109, in track_request_influxdb
    request.headers.get("X-Environment-Key")
```

The above is happening due to `RouteLoggerMiddleware`  modifying WSGI `environ`.
The solution is to avoid passing the request object directly to threaded code.

khvn26:

## Changes

Fixes #5327.

In this PR, we refactor the tracking middleware to avoid accessing the `request` object from a thread. InfluxDB request tracking is performed from a `track_request` async task instead, with all the required data extracted from the request synchronously. `InfluxDBMiddleware` is removed, and `APIUsageCache` is acommodated to cache Influx-tracked requests as well as Postgres-tracked ones. As a side effect, we might see less rate limits from Influx Cloud.

Numerous typing errors are fixed as well. 

Existing unit tests are modified to accommodate for new behaviour.
A new `test_track_request__influx__calls_expected` test is added to verify that the `track_request` task supports writing to InfluxDB as expected.

### DIFF
--- a/api/app_analytics/middleware.py
+++ b/api/app_analytics/middleware.py
@@ -1,4 +1,7 @@
+from typing import Callable
+
 from django.conf import settings
+from django.http import HttpRequest, HttpResponse
 
 from app_analytics.cache import APIUsageCache
 from app_analytics.tasks import track_request
@@ -26,16 +29,22 @@ class GoogleAnalyticsMiddleware:
 
 
 class APIUsageMiddleware:
-    def __init__(self, get_response):  # type: ignore[no-untyped-def]
+    def __init__(
+        self,
+        get_response: Callable[[HttpRequest], HttpResponse],
+    ) -> None:
         self.get_response = get_response
 
-    def __call__(self, request):  # type: ignore[no-untyped-def]
-        resource = get_resource_from_uri(request.path)  # type: ignore[no-untyped-call]
-        if resource in TRACKED_RESOURCE_ACTIONS:
+    def __call__(self, request: HttpRequest) -> HttpResponse:
+        environment_key = request.headers.get("X-Environment-Key")
+        if environment_key and (
+            (resource := get_resource_from_uri(request.path))
+            in TRACKED_RESOURCE_ACTIONS
+        ):
             kwargs = {
                 "resource": resource,
                 "host": request.get_host(),
-                "environment_key": request.headers.get("X-Environment-Key"),
+                "environment_key": environment_key,
             }
             if settings.USE_CACHE_FOR_USAGE_DATA:
                 api_usage_cache.track_request(**kwargs)
--- a/api/app_analytics/middleware.py
+++ b/api/app_analytics/middleware.py
@@ -3,15 +3,13 @@ from django.conf import settings
 from app_analytics.cache import APIUsageCache
 from app_analytics.tasks import track_request
 
-from .models import Resource
 from .track import (
     TRACKED_RESOURCE_ACTIONS,
     get_resource_from_uri,
     track_request_googleanalytics_async,
-    track_request_influxdb_async,
 )
 
-api_usage_cache = APIUsageCache()  # type: ignore[no-untyped-call]
+api_usage_cache = APIUsageCache()
 
 
 class GoogleAnalyticsMiddleware:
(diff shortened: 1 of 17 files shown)


api/environments/models.py
```diff
@@ -213,36 +213,41 @@ def get_environment_from_request(request):  # type: ignore[no-untyped-def]
...
+                *IDENTITY_INTEGRATIONS_RELATION_NAMES,
+            )
+            base_qs = cls.objects.select_related(*select_related_args).defer(
+                "description"
+            )
+            qs_for_embedded_api_key = base_qs.filter(api_key=api_key)
+            qs_for_fk_api_key = base_qs.filter(api_keys__key=api_key)
 
+            try:
                 environment = qs_for_embedded_api_key.union(qs_for_fk_api_key).get()
```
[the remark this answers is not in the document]
khvn26: I'd rather see an explicit `MultipleObjectsReturned` raised, because this is not expected here.

api/app_analytics/middleware.py
```diff
@@ -27,30 +28,23 @@ def __call__(self, request):  # type: ignore[no-untyped-def]
...
+        ):
             kwargs = {
-                "resource": Resource.get_from_resource_name(resource),
+                "resource": resource,
                 "host": request.get_host(),
-                "environment_key": request.headers.get("X-Environment-Key"),
+                "environment_key": environment_key,
             }
             if settings.USE_CACHE_FOR_USAGE_DATA:
                 api_usage_cache.track_request(**kwargs)
```
Zaimwa9, line 50: That used to be the APIUsage tracking entrypoint. Given the modification below, it wraps both Postgres_analytics tracking and influxDb tracking right?

That's where you made the whole middleware redundant ?

api/app_analytics/tasks.py
```diff
@@ -104,16 +108,27 @@ def track_feature_evaluation(
...
+                environment_id=environment.id,
+                count=count,
+            )
+        elif settings.INFLUXDB_TOKEN:
+            track_request_influxdb(
+                resource=resource,
+                host=host,
+                environment=environment,
+                count=count,
+            )
```
Zaimwa9, line 131: Stupid question here. Are the 2 of them exclusive 100% cases ? InfluxDB by default (if token ofc), otherwise Postgres with priority to the later ?

Zaimwa9 (approved):
Makes complete sense to me, nice one.
It could be overkilled and tricky to add cleanly but I could see a test verifying that an environ update wouldn't throw an error. Something like 2 mockeds request with threadA delayed while request B hits with extra headers ?
I let you decide the relevancy, otherwise looking good 👍

api/app_analytics/tasks.py
khvn26: Yeah, so far they are exclusive.

There's nothing preventing us from using both storage backends at the same time but that would require an additional PR (and probably an executive decision by @matthewelwell).

api/app_analytics/middleware.py
khvn26: Yeah, before it was two different middleware pointing to different analytics storage, now it's one (caching if so configured) middleware triggering a task which supports both storage backends.

khvn26:
> a test verifying that an environ update wouldn't throw an error.

Yeah this is tricky to reproduce but, IMO, also redundant since we're getting rid of both culprits here:
1. Threaded execution is discarded, task processor is being used instead.
2. Even if we end up using threading (say, by setting `TASK_RUN_METHOD` to `"SEPARATE_THREAD"`), the Django request object is not being passed around anymore — all required data is extracted in the middleware.

PR #5331, opened 2025-04-11
feat(onboarding): Add API to support onboarding support request

gagantrivedi:

## Changes
Create two APIs: one for sending onboarding support requests from self-hosted instances, and another for receiving those requests and pushing the data to HubSpot.

Adds unit tests


gagantrivedi:
Related: 
https://github.com/Flagsmith/flagsmith-private/issues/16

api/onboarding/views.py
```diff
@@ -0,0 +1,73 @@
...
+logger = logging.getLogger(__name__)
+
+
+SEND_SUPPORT_REQUEST_URL = "https://api.flagsmith.com/api/v1/onboarding/request/send/"
+
+
+@api_view(["POST"])
+@permission_classes([IsAdminUser])
+def send_onboarding_request_to_saas_flagsmith_view(request: Request) -> Response:
+    admin_user: FFAdminUser = request.user  # type: ignore[assignment]
```
gagantrivedi, line 32: @khvn26 Did you run into this typing error as well?

khvn26: Yeah, it's a known (and annoying) issue described [here](https://github.com/typeddjango/django-stubs?tab=readme-ov-file#how-can-i-create-a-httprequest-thats-guaranteed-to-have-an-authenticated-user)

api/app/settings/common.py
```diff
@@ -257,6 +258,7 @@
 LOGIN_THROTTLE_RATE = env("LOGIN_THROTTLE_RATE", "20/min")
 SIGNUP_THROTTLE_RATE = env("SIGNUP_THROTTLE_RATE", "10000/min")
 USER_THROTTLE_RATE = env("USER_THROTTLE_RATE", "500/min")
+ONBOARDING_REQUEST_THROTTLE_RATE = env("ONBOARDING_REQUEST_THROTTLE_RATE", "10/hour")
```
matthewelwell: Does this even need to be configurable?

api/integrations/lead_tracking/hubspot/services.py
```diff
@@ -35,3 +39,21 @@ def register_hubspot_tracker(request: Request) -> None:
     logger.info(
         f"Created HubspotTracker instance for user {request.user.id} with cookie {hubspot_cookie}"
     )
+
+
+def create_self_hosted_onboarding_lead(
```
matthewelwell, line 44: Why not add this as a method to `HubspotLeadTracker` ?

api/onboarding/apps.py
matthewelwell, line 1: It feels like there's an argument that this should be in a separate repo because we don't want to include this in our open source image, right?

api/onboarding/urls.py
```diff
@@ -0,0 +1,25 @@
...
+    ),
+]
+if is_oss():
+    urlpatterns.append(
+        path(
+            "request/send/",
+            send_onboarding_request_to_saas_flagsmith_view,
+            name="send-onboarding-request",
+        ),
+    )
```
matthewelwell, line 27: Right, I guess this negates my comment above. I suppose the logic here is that we want the FE to make an API request to the local API, which makes the request out to Flagsmith SaaS. I guess that makes sense.

api/onboarding/urls.py
```diff
@@ -0,0 +1,25 @@
...
+)
+
+app_name = "onboarding"
+
+urlpatterns = [
+    path(
+        "request/receive/",
+        ReceiveSupportRequestFromSelfHosted.as_view(),
+        name="receive-onboarding-request",
+    ),
```
matthewelwell: Shouldn't this only be added if `if_saas()` is true?

api/onboarding/views.py
```diff
@@ -0,0 +1,73 @@
...
+    organisation = admin_user.organisations.first()
+
+    if not organisation:
+        return Response(
+            {"error": "Please create an organisation before requesting support"},
```
matthewelwell: We have an ErrorSerializer that uses `"message"` as the key in the response [here](https://github.com/Flagsmith/flagsmith/blob/main/api/api/serializers.py#L4). I'm not sure how heavily it's used throughout the API, but we should at least try not to make things worse.

api/onboarding/views.py
```diff
@@ -0,0 +1,73 @@
...
+        response = requests.post(SEND_SUPPORT_REQUEST_URL, data=data, timeout=30)
+        response.raise_for_status()
+    except RequestException as e:
+        logger.error("Failed to send support request to flagsmith: %s", e)
+        return Response(
+            {"error": "Failed to send support request."},
+            status=status.HTTP_502_BAD_GATEWAY,
+        )
+
+    return Response(status=status.HTTP_204_NO_CONTENT)
```
matthewelwell: It feels to me like we shouldn't necessarily care about the response to this, and we should handle it with an asynchronous task?

api/onboarding/views.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class ReceiveSupportRequestFromSelfHosted(GenericAPIView):  # type: ignore[type-arg]
+    serializer_class = SelfHostedOnboardingSupportSerializer
+    authentication_classes = ()
```
matthewelwell, line 53: I wonder if there's any form of authentication we can put on this? At the moment, anyone that finds this endpoint in our API docs can spam it, and blow out our hubspot API quota. 

Some options I think that we have: 

 1. Just bundle in some kind of authentication token in the open source image
 2. Use some kind of signature validation on the payload data

We could also still technically make all of this code private, and just bundle the relevant endpoint in open source image at build time. Assuming that we're still going to use python, you'd still be able to find the code in the docker image, but it would at least be slightly harder to reverse engineer. 

Some other things that I'm thinking about: 

 1. Should we wrap this logic in a feature flag so that we can kill it if needs be?

api/tests/unit/integrations/lead_tracking/hubspot/test_unit_hubspot_client.py
```diff
@@ -145,3 +149,80 @@ def test_create_company_without_organisation_information(
...
     }
+
+
+def test__create_contact(
+    hubspot_client: HubspotClient, admin_user: FFAdminUser
+) -> None:
+    # Given
+    # email = "user@flagsmith.com"
+    # first_name = "test"
+    # last_name = "user"
```
matthewelwell: Are these needed?

api/tests/unit/integrations/lead_tracking/hubspot/test_unit_hubspot_client.py
```diff
@@ -145,3 +149,80 @@ def test_create_company_without_organisation_information(
...
+                        }
+                    ],
+                    "to": {"id": hubspot_company_id},
+                }
+            ],
+        )
+    )
+
+
+def test__create_self_hosted_contact(hubspot_client: HubspotClient) -> None:
```
matthewelwell: I'm not sure I like the `test__foo_bar` approach - it's not really in-keeping with the rest of the code base. Is there a reason to do this?

api/onboarding/apps.py
matthewelwell: I left this comment in because there are later comments that reference it, but I appreciate that this specific question is not really helpful - you can ignore it for now.

api/tests/unit/integrations/lead_tracking/hubspot/test_unit_hubspot_client.py
gagantrivedi: It's a bit easier to read imo.. But I don't have very strong feelings about this

api/onboarding/urls.py
gagantrivedi: Yeah, I didn't do that because it would just return a 400 in all the other environments anyway, but I don't see any reason to include it either.

api/onboarding/views.py
gagantrivedi: I'm not a huge fan of security by obscurity, but I guess in this case we could create a model and only allow one entry per IP (assuming we can store the IP). It's basically throttling, but with an infinite window.

api/integrations/lead_tracking/hubspot/services.py
gagantrivedi: I think there's too much business logic here for this to live on a type — separation of concerns and all that.

api/onboarding/views.py
matthewelwell: Hmm... yeah, interesting. Maybe we can just store a hash of the IP?

gagantrivedi: Yeah, that's a good idea. I will do that

matthewelwell: Couldn't we develop our own custom throttle class instead of a model?

gagantrivedi: Yeah, but we will need a (permanent) place to store IP hashes

matthewelwell: Well, if we use e.g. redis, I'd say that's permanent enough, right?

gagantrivedi: Yeah, I guess it's not the end of the world if we lose that data — and it'll also finally force us to use Redis in production.

api/app/settings/common.py
```diff
@@ -715,6 +716,18 @@
     "USER_THROTTLE_CACHE_OPTIONS", default={}
 )
 
+ONBOARDING_REQUEST_THROTTLE_CACHE_NAME = "onboarding-request-throttle"
+ONBOARDING_REQUEST_THROTTLE_CACHE_BACKEND = env.str(
+    "ONBOARDING_REQUEST_THROTTLE_CACHE_BACKEND",
+    "django.core.cache.backends.db.DatabaseCache",
```
gagantrivedi, line 722: Using the DB backend as the default here so we can deploy it before testing Redis in production

api/integrations/lead_tracking/hubspot/services.py
matthewelwell: Hmm... I'd argue that this business logic is exactly what that class is concerned about? But I don't have strong enough feelings about it to die on this hill.

api/integrations/lead_tracking/hubspot/tasks.py
```diff
@@ -62,3 +62,20 @@ def track_hubspot_lead_without_organisation(user_id: int) -> None:
...
+    from integrations.lead_tracking.hubspot.services import (
+        create_self_hosted_onboarding_lead,
+    )
+
+    create_self_hosted_onboarding_lead(
+        first_name=first_name,
+        last_name=last_name,
+        email=email,
+        organisation_name=organisation_name,
+    )
```
matthewelwell, line 81: What's the reason for not just defining the logic in the task itself? Just for code organisation?

api/integrations/lead_tracking/hubspot/services.py
gagantrivedi: I think that class is already trying to do a lot of different things; I'd prefer to keep this in service

api/integrations/lead_tracking/hubspot/tasks.py
gagantrivedi: Yeah, pretty much

matthewelwell (approved):

PR #5332, opened 2025-04-11
fix(n+1): fix environment-document n+1 for env_feature_version

gagantrivedi:

## Changes

Prefetch `environment_feature_version`  to avoid n+1 

Unit test updated to cover this change

### DIFF
--- a/api/environments/models.py
+++ b/api/environments/models.py
@@ -405,6 +405,7 @@ class Environment(
                         "feature",
                         "feature_state_value",
                         "identity",
+                        "environment_feature_version",
                         "identity__environment",
                     ).prefetch_related(
                         Prefetch(
--- a/api/tests/unit/environments/test_unit_environments_views_sdk_environment.py
+++ b/api/tests/unit/environments/test_unit_environments_views_sdk_environment.py
@@ -18,6 +18,7 @@ from features.models import (  # type: ignore[attr-defined]
     FeatureStateValue,
 )
 from features.multivariate.models import MultivariateFeatureOption
+from features.versioning.tasks import enable_v2_versioning
 from projects.models import Project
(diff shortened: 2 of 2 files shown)


api/environments/models.py
```diff
@@ -405,6 +405,7 @@ def _get_environment_document_from_db(
                         "feature",
                         "feature_state_value",
                         "identity",
+                        "environment_feature_version",
```
gagantrivedi, line 408: "We can include or exclude this based on whether `environment.use_v2_feature_versioning` is enabled, but I don't really see any downside to including it for environments where` use_v2_feature_versioning` is not enabled."

api/tests/unit/environments/test_unit_environments_views_sdk_environment.py
```diff
@@ -26,19 +27,28 @@
...
     project2 = Project.objects.create(
         name="standin_project", organisation=organisation_two
     )
 
-    environment = Environment.objects.create(name="Test Environment", project=project)
+    environment = Environment.objects.create(
+        name="Test Environment",
+        project=project,
+        use_v2_feature_versioning=use_v2_feature_versioning,
+    )
```
matthewelwell: I don't think this will matter, because we haven't yet created any features, but just for completeness, I recommend doing: 

```suggestion
    environment = Environment.objects.create(
        name="Test Environment",
        project=project,
    )
    enable_v2_versioning(environment.id)
```

gagantrivedi: Done

api/environments/models.py
matthewelwell: agreed.

matthewelwell (approved):

PR #5343, opened 2025-04-16
feat(billing): Remove `AUTO_SEAT_UPGRADE_PLANS` setting

khvn26:

## Changes

In this PR, we remove the `AUTO_SEAT_UPGRADE_PLANS` setting in favour of centralised `SubscriptionPlanFamily.get_by_plan_id` logic.
 
Made sure all relevant unit tests pass without the setting present.

### DIFF
--- a/api/organisations/models.py
+++ b/api/organisations/models.py
@@ -138,10 +138,7 @@ class Organisation(LifecycleModelMixin, SoftDeleteExportableModel):  # type: ign
         self.save()
 
     def is_auto_seat_upgrade_available(self) -> bool:
-        return (
-            len(settings.AUTO_SEAT_UPGRADE_PLANS) > 0
-            and self.subscription.can_auto_upgrade_seats
-        )
+        return self.subscription.can_auto_upgrade_seats
 
     @hook(BEFORE_DELETE)
     def cancel_subscription(self):  # type: ignore[no-untyped-def]
@@ -262,7 +259,7 @@ class Subscription(LifecycleModelMixin, SoftDeleteExportableModel):  # type: ign
 
     @property
     def can_auto_upgrade_seats(self) -> bool:
-        return self.plan in settings.AUTO_SEAT_UPGRADE_PLANS
+        return self.subscription_plan_family == SubscriptionPlanFamily.SCALE_UP
 
     @property
     def is_free_plan(self) -> bool:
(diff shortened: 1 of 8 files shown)


api/organisations/models.py
```diff
@@ -262,7 +259,7 @@ def update_plan(self, plan_id):  # type: ignore[no-untyped-def]
 
     @property
     def can_auto_upgrade_seats(self) -> bool:
-        return self.plan in settings.AUTO_SEAT_UPGRADE_PLANS
+        return self.subscription_plan_family == SubscriptionPlanFamily.SCALE_UP
```
matthewelwell: Just to future proof this, I wonder if we should add a check on SaaS? 

```suggestion
        return is_saas() and self.subscription_plan_family == SubscriptionPlanFamily.SCALE_UP
```

I don't see a scenario in which scale up is used in any environment other than SaaS, but perhaps worth belt and braces approach?

khvn26: > future proof

Surely we'll have to modify more that this code path if we decide to charge by seats on enterprise?

api/organisations/models.py
```diff
@@ -262,7 +259,10 @@ def update_plan(self, plan_id):  # type: ignore[no-untyped-def]
 
     @property
     def can_auto_upgrade_seats(self) -> bool:
-        return self.plan in settings.AUTO_SEAT_UPGRADE_PLANS
+        return (
+            is_saas()
+            and self.subscription_plan_family == SubscriptionPlanFamily.SCALE_UP
```
Zaimwa9, line 264: So i'm seeing 
```
class SubscriptionPlanFamily(Enum):
    FREE = "FREE"
    START_UP = "START_UP"
    SCALE_UP = "SCALE_UP"
    ENTERPRISE = "ENTERPRISE"
```

But I don't see the plan on the pricing page. Is it a custom in between start-up and enterprise?

Zaimwa9 (approved):
The tests seem to cover the different cases. If I summarize how we do now:
If subscription is scale-up-xx
Organisation don't have fixed user limits but when adding a new one:
- We charge the seat
- if payment fails => failure and user can't join
- otherwise +1 user ?

Could you let me know when it's on staging to use it to start manipulating chargebee 👍 ?

api/organisations/models.py
khvn26: Good question 👍 Our plan lineup has been revised to exclude scale-up for new customers, but we still support existing scale-up users.

api/organisations/models.py
khvn26: Explained offline, resolving this (basically, it's possible to trigger this code path for self-hosted installations by assigning a scale-up plan to the org subscription).

PR #5346, opened 2025-04-16
fix: `track_request` task not backwards compatible

khvn26:

## Changes

In this PR, we make sure the `track_request` task is backwards compatible by accepting the `resource: int` argument. 
All usage tracking functions are streamlined to use the `Resource` enum.
`TRACKED_RESOURCE_ACTIONS` constant is retired in favour of the `Resource.is_tracked` property.

The change that broke backwards compatibility was introduced in https://github.com/Flagsmith/flagsmith/pull/5330.

Modified the related unit tests.
Will test in staging.

### DIFF
--- a/api/app_analytics/tasks.py
+++ b/api/app_analytics/tasks.py
@@ -109,15 +109,16 @@ def track_feature_evaluation(
 
 @register_task_handler()
 def track_request(
-    resource: str,
+    resource: int,
     host: str,
     environment_key: str,
     count: int = 1,
 ) -> None:
     if environment := Environment.get_from_cache(environment_key):
+        resource = Resource(resource)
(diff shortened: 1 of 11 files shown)


Zaimwa9 (approved):
Discussed with Kim, lgtm

api/app_analytics/tasks.py
```diff
@@ -109,15 +109,16 @@ def track_feature_evaluation(
 
 @register_task_handler()
 def track_request(
-    resource: str,
+    resource: int,
```
matthewelwell, line 112: Just highlighting that this is the crux of the fix since it reverts the change of type as implemented [here](https://github.com/Flagsmith/flagsmith/pull/5330).

matthewelwell (approved):

PR #5354, opened 2025-04-17
feat: Test webhook from backend 

Zaimwa9:

## Changes
The goal of this PR is to move the webhook testing logic from the frontend (with custom method to mimic backend signature) to the backend
Frontend payload exemple (audit log)
```
{
    "payload": {
        "created_date": "2020-02-23T17:30:57.006318Z",
        "log": "New Flag / Remote Config created: my_feature",
        "author": {
            "id": 3,
            "email": "user@domain.com",
            "first_name": "john",
            "last_name": "doe"
        },
        "environment": null,
        "project": {
            "id": 6,
            "name": "Project name",
            "organisation": 1
        },
        "related_object_id": 6,
        "related_object_type": "FEATURE"
    },
    "scope": {
        "id": 2,
        "type": "organisation"
    },
    "secret": "my-secret",
    "webhookUrl": "http://localhost:3001/webhook"
}
```

- New `WebhookViewSet` with `test` endpoint
- Re-uses Organisation/Environment permissions based on `scope` payload object (on top of existing org/env pk from url)
- Reviewed errors handling: Success = 200 (unchanged), any error from end-user will return status 502 bad gateway and a text message with status and raw error (e.g `Webhook returned error status: 400`)
- Couple of UI fixes (see screenshots)

- Added unit tests
- Spawned 2 webservers (python / node) testing against them 

1. Go to environment => webhooks or organisation settings => webhooks
2. Prepare a webhook endpoint
3. Test with / without different signatures
4. Try different error responses

example code:
`pip install flask flask-cors`
`python server.py`

```
# server.py
from flask import Flask, request, jsonify
import hmac
import hashlib
import json
from functools import wraps
import os
import datetime
from flask_cors import CORS

app = Flask(__name__)

WEBHOOK_SECRET = "your-secret"
CORS(app, resources={
    r"/*": {
        "origins": "*", 
        "methods": ["GET", "POST", "PUT", "DELETE", "OPTIONS"],
        "allow_headers": ["Content-Type", "Authorization", "x-flagsmith-signature"]
    }
})

@app.route('/webhook', methods=['POST'])
def webhook():
    received_signature = request.headers.get('x-flagsmith-signature')
    
    if not received_signature and WEBHOOK_SECRET:
        return jsonify({
            'status': 'error',
            'message': 'Missing signature header'
        }), 401

    if WEBHOOK_SECRET:
        request_body = json.dumps(request.json)     
        expected_signature = hmac.new(
            key=WEBHOOK_SECRET.encode(),
            msg=request_body.encode(),
            digestmod=hashlib.sha256
        ).hexdigest()

        print(f"Expected signature: {expected_signature}")
        print(f"Received signature: {received_signature}")
        if not hmac.compare_digest(expected_signature, received_signature):
            return jsonify({
                'status': 'error',
                'message': 'Invalid signature'
            }), 401

    print("Webhook received and verified:", {
        'headers': dict(request.headers),
        'body': request.json,
        'timestamp': datetime.datetime.utcnow().isoformat()
    })

    return jsonify({
        'status': 'success',
        'message': 'Webhook received and verified successfully'
    })

if __name__ == '__main__':
    port = int(os.environ.get('PORT', 3001))
    app.run(host='0.0.0.0', port=port)
```


api/environments/views.py
```diff
@@ -55,6 +57,14 @@
...
 )
 
+from typing import TypeVar, runtime_checkable, Protocol
+
+T = TypeVar('T', bound=Model)
+
+@runtime_checkable
+class HasObjects(Protocol[T]):
+    objects: Manager[T]
+
```
Zaimwa9: Curious to have your opinion on this typing and cast @khvn26 before using it more (it's to tackle this `environments/views.py:291: error: "type[T]" has no attribute "objects"  [attr-defined]`)

api/webhooks/views.py
```diff
@@ -0,0 +1,50 @@
...
+                )
+            return Response(
+                {
+                    "detail": f"Webhook test successful. Response status: {response.status_code}"
+                },
+                status=status.HTTP_200_OK,
+            )
+        except requests.exceptions.RequestException as e:
+            return Response(
+                {"detail": f"Could not connect to webhook URL: {e}"},
```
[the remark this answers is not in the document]
Zaimwa9: I personally don't think it is an issue

frontend/common/services/useWebhooks.ts
```diff
@@ -32,6 +32,20 @@ export const webhookService = service
         }),
       }),
 
+      testWebhook: builder.mutation<void, Req['testWebhook']>({
+        query: (query) => {
+          return ({
+          body: {
+            payload: { ...query.body },
```
khvn26: I am surprised we don't hardcode/generate payload in the backend — seems a bit of a half-solution to me right now.

api/environments/views.py
```diff
@@ -55,6 +57,14 @@
     WebhookSerializer,
 )
 
+T = TypeVar("T", bound=Model)
+
+
+@runtime_checkable
```
khvn26: What's the reason to use `@runtime_checkable` here?

api/webhooks/views.py
```diff
@@ -0,0 +1,50 @@
...
+        try:
+            assert isinstance(webhook_url, str)
+            assert isinstance(payload, dict)
+            response = send_test_request_to_webhook(webhook_url, secret, payload)
+            if response.status_code >= 400:
+                return Response(
+                    {
+                        "detail": f"Webhook returned error status: {response.status_code}{', ' + response.text if response.text else ''}"
+                    },
+                    status=status.HTTP_502_BAD_GATEWAY,
```
khvn26: Maybe it's bikeshedding, but using semantics intended for gateways/proxies for testing webhooks seems awkward to me. I think returning 400 or even a 200 with status in body is more suitable for the context; we're not proxying this response, just checking it.

api/webhooks/views.py
```diff
@@ -0,0 +1,50 @@
...
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+        try:
+            assert isinstance(webhook_url, str)
+            assert isinstance(payload, dict)
+            response = send_test_request_to_webhook(webhook_url, secret, payload)
+            if response.status_code >= 400:
+                return Response(
+                    {
+                        "detail": f"Webhook returned error status: {response.status_code}{', ' + response.text if response.text else ''}"
```
khvn26: Consider extending the response schema to include `code` and `body` attributes instead of formatting them in a string message.

api/webhooks/views.py
```diff
@@ -0,0 +1,50 @@
...
+                    },
+                    status=status.HTTP_502_BAD_GATEWAY,
+                )
+            return Response(
+                {
+                    "detail": f"Webhook test successful. Response status: {response.status_code}"
```
khvn26: This makes me think if we're actually supposed to follow redirects when calling webhooks.

If yes: 
 - Are we doing it IRL?
 - Are we supposed to accept other finite status codes than 2xx?

If no:
- We should document this behaviour so it's clear to the users.

I'm leaning towards following redirects, maybe with some guardrails to avoid redirection loops, tarpits and such.

khvn26 (changes requested):
Came up with some comments/questions.

api/environments/views.py
khvn26: I understand why it's done here, but I don't think this is necessarily a good solution.

I'd try [following the official way to deal with this](https://github.com/typeddjango/django-stubs?tab=readme-ov-file#how-to-use-typemodel-annotation-with-objects-attribute) first by using the `._default_manager` attribute.

Zaimwa9: Ah perfect thanks I will look into it

frontend/common/services/useWebhooks.ts
Zaimwa9: True, and it adds unnecessary checks and cause of errors. will change and avoid using real names

api/webhooks/views.py
Zaimwa9: Actually it's a miss from my side. We do expect them to respond with a 200 only. And following redirects could expose us to some malicious attempts (or setup a whole security layer)

image

api/webhooks/webhooks.py
```diff
@@ -303,9 +303,15 @@ def _get_failure_email_template_data(  # type: ignore[no-untyped-def]
 
 
 def send_test_request_to_webhook(
-    url: str, secret: str | None, payload: dict[str, Any]
+    url: str, secret: str | None, scope: str
 ) -> requests.models.Response:
-    json_data = json.dumps(payload, sort_keys=True, cls=DjangoJSONEncoder)
+    testData = environment_webhook_data if scope == "environment" else organisation_webhook_data
```
khvn26: This 🐫 got lost

khvn26 (changes requested):
I believe the recent changes are a massive improvement, good job 👍 

There's still a couple comments I'd like you to address, though:

1. I'm not particularly fond of using static sample data, though, as it can get stale easily. Ideally, we should be able to reuse existing webhook tasks in the check.

2. Skimming through the code, I've found (currently unused?) `trigger_sample_webhook` utility function and `TriggerSampleWebhookMixin` that we need to either refactor and integrate into your work, or get rid of.

api/webhooks/webhooks.py
Zaimwa9: oupsie. Removed

Zaimwa9:
@khvn26 made the changes. To answer your questions:
1. I went half-way by using the models to generate the sample data so they are closer to reality while remaining predictable (for testing). They are using the existing methods so not much more maintainance, thanks for raising it
2. Indeed, it seems the purpose was similar, I could have caught it. I removed it as it was unused. Anticipating the question. I voluntarily didn't use `_call_webhook` there because as of it doesn't pass the error through which is imo helpful for users setting up the webhook and modifying could be a breaking change

api/tests/unit/webhooks/test_unit_webhooks.py
```diff
@@ -345,3 +322,159 @@ def test_call_integration_webhook_does_not_raise_error_on_backoff_give_up(
...
+    assert response.status_code == expected_final_status
+    mock_post.assert_called_once()
+    if expected_final_status == 200:
+        assert response.json() == {"detail": "Webhook test successful", "status": 200}
+    else:
+        assert response.json() == {
+            "detail": "Webhook returned error status",
+            "status": external_api_response_status,
+            "body": external_api_error_text,
+        }
```
khvn26: Please parametrise response json instead of introducing branching logic in the test body.

api/tests/unit/webhooks/test_unit_webhooks.py
```diff
@@ -345,3 +322,159 @@ def test_call_integration_webhook_does_not_raise_error_on_backoff_give_up(
...
+    )
+
+    # Then
+    call_kwargs = mock_post.call_args.kwargs
+    if should_have_signature:
+        assert FLAGSMITH_SIGNATURE_HEADER in call_kwargs["headers"]
+        assert call_kwargs["headers"][FLAGSMITH_SIGNATURE_HEADER] == expected_signature
+    else:
+        assert FLAGSMITH_SIGNATURE_HEADER not in call_kwargs["headers"]
+    assert response.status_code == 200
```
khvn26: Same as [here](https://github.com/Flagsmith/flagsmith/pull/5354/files#r2089293367) — please parametrise headers.

api/environments/views.py
```diff
@@ -276,43 +284,45 @@ def disable_v2_versioning(self, request: Request, api_key: str) -> Response:
...
 
-    def perform_update(self, serializer):  # type: ignore[no-untyped-def]
-        serializer.save(environment=self._get_environment())  # type: ignore[no-untyped-call]
+    def perform_update(self, serializer: BaseSerializer[T]) -> None:
+        serializer.save(environment=self._get_environment())
 
-    def _get_environment(self):  # type: ignore[no-untyped-def]
-        return Environment.objects.get(api_key=self.kwargs.get("environment_api_key"))
+    def _get_environment(self) -> Environment:
+        return cast(
```
emyller: Ideally casting should not be needed here. 🤔

api/environments/views.py
```diff
@@ -276,43 +284,45 @@ def disable_v2_versioning(self, request: Request, api_key: str) -> Response:
...
-class NestedEnvironmentViewSet(viewsets.GenericViewSet):  # type: ignore[type-arg]
-    model_class = None
+class NestedEnvironmentViewSet(Generic[T], viewsets.GenericViewSet[T]):
+    model_class: Type[T]
     webhook_type = WebhookType.ENVIRONMENT
 
-    def get_queryset(self):  # type: ignore[no-untyped-def]
-        return self.model_class.objects.filter(  # type: ignore[attr-defined]
+    def get_queryset(self) -> QuerySet[T]:
+        model_cls = cast(HasObjects[T], self.model_class)
```
emyller: @Zaimwa9 I'd love to understand the intent here and discuss whether this is the optimal approach to it.

Zaimwa9: It's definitely not the optimal approach but removing the `type: ignore` raises a typing challenge with Django models I haven't solved (and circling in loop trying to avoid refactoring).

I'm using the cast with `HasObjects` to help mypy understand Django's behavior.
The issue is that while every Django model does have an objects manager at runtime, this attribute is added dynamically. Mypy being conservative, it doesn't accept that the generic `Type[T]` will have a .objects attribute, unless explicitly told.
The cast is my way of saying "trust me on this one". Here, the challenge comes from the fact that `NestedEnvironmentViewSet` does require it to be a generic

```
class WebhookViewSet(
    NestedEnvironmentViewSet[Webhook],
    mixins.ListModelMixin,
    mixins.CreateModelMixin,
    mixins.UpdateModelMixin,
    mixins.DestroyModelMixin,
):
    serializer_class = WebhookSerializer
    pagination_class = None
    permission_classes = [IsAuthenticated, NestedEnvironmentPermissions]
    model_class: Type[Webhook] = Webhook
    webhook_type: WebhookType = WebhookType.ENVIRONMENT


class EnvironmentAPIKeyViewSet(
    NestedEnvironmentViewSet[EnvironmentAPIKey],
    mixins.ListModelMixin,
    mixins.CreateModelMixin,
    mixins.UpdateModelMixin,
    mixins.DestroyModelMixin,
):
    serializer_class = EnvironmentAPIKeySerializer
    pagination_class = None
    permission_classes = [IsAuthenticated, EnvironmentAdminPermission]
    model_class: Type[EnvironmentAPIKey] = EnvironmentAPIKey

```

I'm also not at all convinced with it. I just gave it another try and ended up in a maze of unfolded type errors. So far I haven't solved it and I plan to try other approaches with a simpler case (without a base view set). If ever you have an idea or a lead, i'd be more than happy to explore

api/webhooks/views.py
```diff
@@ -0,0 +1,62 @@
...
+        scopeType: str = scope.get("type", "")
+        if not all([webhook_url, scopeType]):
+            return Response(
+                {"detail": "webhookUrl is required"},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+        try:
+            webhook_type = (
+                WebhookType.ORGANISATION
+                if scopeType == "organisation"
```
khvn26:
```suggestion
                if scope_type == "organisation"
```

api/webhooks/views.py
```diff
@@ -0,0 +1,62 @@
...
+                else WebhookType.ENVIRONMENT
+            )
+            response = send_test_request_to_webhook(webhook_url, secret, webhook_type)
+            if response.status_code != 200:
+                return Response(
+                    {
+                        "detail": "Webhook returned error status",
```
khvn26: Since we don't accept redirect statuses as well:

```suggestion
                        "detail": "Webhook returned invalid status",
```

api/webhooks/views.py
```diff
@@ -0,0 +1,62 @@
...
+            "secret",
+        )
+        scope: dict[str, Any] = request.data.get("scope", {})
+        webhook_url: str = request.data.get("webhookUrl", "")
+        scopeType: str = scope.get("type", "")
+        if not all([webhook_url, scopeType]):
+            return Response(
+                {"detail": "webhookUrl is required"},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
```
khvn26: Can we use a serialiser/pydantic schema to validate the request body? + use it to include proper OpenAPI docs

Example [here](https://github.com/Flagsmith/flagsmith/blob/365ad46717cc4177936c5839eba272150e5a650c/api/segments/views.py#L128-L142)

khvn26 (changes requested):
Sorry, one last round — just a couple comments though, LGTM otherwise 👍

api/webhooks/views.py
Zaimwa9: Good call, upgraded with serializers, thanks for reminding

api/webhooks/serializers.py
```diff
@@ -1,10 +1,31 @@
...
+class WebhookURLSerializer(serializers.Serializer[None]):
     url = serializers.URLField()
+
+
+class ScopeSerializer(serializers.Serializer[None]):
+    type = serializers.ChoiceField(choices=["organisation", "environment"])
+
+
+class TestWebhookSerializer(serializers.Serializer[None]):
+    webhookUrl = serializers.URLField(required=True)
```
khvn26: I don't think we use camel case anywhere in our [OpenAPI schema](https://api.flagsmith.com/api/v1/docs/?format=openapi).

khvn26 (approved):

PR #5356, opened 2025-04-18
fix: enable usage cache by default

gagantrivedi:

## Changes

Enable USE_CACHE_FOR_USAGE_DATA by default (we always recommend enabling this; it's kind of a breaking change, but not really — let me know what you think).

Rename PG_API_USAGE_CACHE_SECONDS to API_USAGE_CACHE_SECONDS, since the cache now also covers Influx usage.

_Please describe._


api/app/settings/common.py
```diff
@@ -340,8 +340,14 @@
...
 
 USE_POSTGRES_FOR_ANALYTICS = env.bool("USE_POSTGRES_FOR_ANALYTICS", default=False)
-USE_CACHE_FOR_USAGE_DATA = env.bool("USE_CACHE_FOR_USAGE_DATA", default=False)
-PG_API_USAGE_CACHE_SECONDS = env.int("PG_API_USAGE_CACHE_SECONDS", default=60)
+USE_CACHE_FOR_USAGE_DATA = env.bool("USE_CACHE_FOR_USAGE_DATA", default=True)
+
+API_USAGE_CACHE_SECONDS = env.int("API_USAGE_CACHE_SECONDS", default=0)
+
+if not API_USAGE_CACHE_SECONDS:
+    # Fallback to the old variable name, which is deprecated
```
khvn26, line 348: Any reference in the docs we need to update?

khvn26 (approved):

api/app/settings/common.py
gagantrivedi: I don't think it's documented

gagantrivedi: I have added some docs now

PR #5362, opened 2025-04-21
fix: metadata-feature-changes-erases-existing-value

Zaimwa9:

## Changes
- Overriden `update` to use `update_metadata` method from `SerializerWithMetadata` inherited from common

- Running tests against common
- Manually (creating and updating custom metadata fields in env/segments/features)
- Created a new test in `segments_views` failing on previous `flagsmith-common` version, passing now
```
FAILED tests/unit/segments/test_unit_segments_views.py::test_update_segment_metadata_create_correct_number_of_metadata[admin_client] - assert 1 == 2
FAILED tests/unit/segments/test_unit_segments_views.py::test_update_segment_metadata_create_correct_number_of_metadata[admin_master_api_key_client] - assert 1 == 2
```
 
https://www.loom.com/share/f980f034f6144c548a2f30efd809f828


api/environments/serializers.py
```diff
@@ -97,6 +98,16 @@ def get_project(
...
         )
 
+    def update(
+        self, instance: models.Model, validated_data: dict[str, typing.Any]
+    ) -> Environment:
+        metadata_items = validated_data.pop("metadata", [])
+        environment = super().update(instance, validated_data)
+        self.update_metadata(environment, metadata_items)  # type: ignore[no-untyped-call]
+        environment.refresh_from_db()
+        assert isinstance(environment, Environment)
```
matthewelwell: This assertion feels unnecessary?

api/features/serializers.py
```diff
@@ -344,6 +345,16 @@ def get_project(
...
             )
 
+    def update(
+        self, instance: models.Model, validated_data: dict[str, typing.Any]
+    ) -> Feature:
+        metadata_items = validated_data.pop("metadata", [])
+        feature = super().update(instance, validated_data)  # type: ignore[no-untyped-call]
+        self.update_metadata(feature, metadata_items)
+        feature.refresh_from_db()
+        assert isinstance(feature, Feature)
```
matthewelwell: As above, this seems unnecessary?

matthewelwell (commented):
In general, I'm a little confused by this PR (and the one in common). I'm not sure exactly what we're solving. Based on the changes that I See here, it feels like the metadata logic never would have previously worked for environments and features, but I don't think that's true, right? 

Also, I don't see any additional tests added to this PR, or the one in the flagsmith-common repository. we should definitely add a test (or tests) somewhere.

api/environments/serializers.py
Zaimwa9: Actually as of it is "necessary" for mypy because `super().update` is not fully typed and returns a `models.Model`. I can type it in a separate PR because it might need to type a couple of used functions inside

api/features/serializers.py
Zaimwa9: Same typing issue

api/environments/serializers.py
khvn26: I'd at least hide it behind `if typing.TYPE_CHECKING:` to clearly state its purpose.

Ideally, i'd try annotating the var like so:

```python
environment: Environment = super().update(instance, validated_data)
```

and see if mypy complains about it.

Zaimwa9:
Added new test case to assert the correct number of metadata is created.
Failing on previous `common` version (only one created), success now
```
FAILED tests/unit/segments/test_unit_segments_views.py::test_update_segment_metadata_create_correct_number_of_metadata[admin_client] - assert 1 == 2
FAILED tests/unit/segments/test_unit_segments_views.py::test_update_segment_metadata_create_correct_number_of_metadata[admin_master_api_key_client] - assert 1 == 2
```

api/features/serializers.py
```diff
@@ -344,6 +345,15 @@ def get_project(
...
             )
 
+    def update(
+        self, instance: models.Model, validated_data: dict[str, typing.Any]
+    ) -> Feature:
+        metadata_items = validated_data.pop("metadata", [])
+        feature = typing.cast(Feature, super().update(instance, validated_data))
+        self.update_metadata(feature, metadata_items)
+        feature.refresh_from_db()
+        return feature
```
khvn26, line 355: The typing here is really awkward. 

Created https://github.com/Flagsmith/flagsmith-common/issues/71, no action needed here.

khvn26 (approved):
Looks good 👍

Issue #4253, opened 2024-06-26
Add Flagsmith release version to all API response headers


rolodato:
### Is your feature request related to a problem? Please describe.

Today I spent way more time than I'd like to admit troubleshooting a self-hosted customer issue that was caused by not running on the latest Flagsmith version. Having the current Flagsmith version be returned as part of the response headers might have saved me hours of troubleshooting.

### Describe the solution you'd like.

Add a response header like `X-Flagsmith-Version: 2.123.0` to all responses, no matter the status code or error condition.

### Describe alternatives you've considered

Don't really care about the specific header, just want to see the version in there somewhere.

I know we have the `/version` endpoint as well, but that requires 1) making an explicit request to it, and 2) knowing/remembering that it exists, which customers usually don't. Usually the information we get for troubleshooting is in the form of a failed curl command or similar, which if we're lucky contains the response headers as well.

emyller:
Confirmed it works in the SaaS API:

```sh
$ http -h https://api.flagsmith.com/robots.txt
HTTP/1.1 200 OK
Cache-Control: max-age=0, no-cache, no-store, must-revalidate, private
Connection: keep-alive
Content-Length: 26
Content-Type: text/plain
Cross-Origin-Opener-Policy: same-origin
Date: Mon, 19 May 2025 13:52:35 GMT
Expires: Mon, 19 May 2025 13:52:35 GMT
Flagsmith-Version: 2.177.1
Pragma: no-cache
Referrer-Policy: same-origin
Server: gunicorn
Vary: Origin
Via: 1.1 f5375a573bc3d2aad9cc6e2a6aa5aec0.cloudfront.net (CloudFront)
X-Amz-Cf-Id: D_5CprWuKKWXIF_k8F-O-rqY2iQYDEcbaegNLW_EwirvCFqaNgGgPw==
X-Amz-Cf-Pop: GRU1-C2
X-Cache: Miss from cloudfront
X-Content-Type-Options: nosniff
X-Frame-Options: DENY
```

PR #5366, opened 2025-04-22
feat: Add product version to API response headers

emyller:

## Changes

ref: #4253

- Read the release Git tag into an environment variable, then into Django settings.
- Introduce a middleware that adds `X-Flagsmith-Version` to **all** responses.

Functional tests.


emyller:
Messed up GitHub references.

api/tests/unit/api/test_unit_api.py
```diff
@@ -20,3 +21,40 @@
...
+        "/robots.txt",
+        # TODO: Add other GET-only URLs
+    ],
+)
+def test_api_version_is_added_to_success_response_headers(
+    url: str,
+    client: APIClient,
+    settings: SettingsWrapper,
+) -> None:
+    settings.VERSION = "v1.2.3"  # NOTE: Maybe read from Git tag into an env var?
```
matthewelwell: I think that this is a fine approach. I feel like reading the git tag could result in some flakiness. 

We can then just unit test the logic that reads the git tag into settings using mocking perhaps.

matthewelwell (commented):
Approach and tests look fine to me.

api/tests/unit/api/test_unit_api.py
khvn26: @matthewelwell Will we have to modify the `/version` endpoint to read from the setting as well?

matthewelwell: Yes, I believe so.

khvn26:
I have a hunch this belongs to https://github.com/flagsmith/flagsmith-common, as the implementation is likely going to be a global middleware

api/tests/unit/api/test_unit_api.py
emyller: @matthewelwell My first instinct was to read the Git tag into a Docker arg when building the image, then passing it through an env var to the code at build time, then finally into a new Django setting `VERSION`, as my code suggests so far.

However, I just learned that `flagsmith` will already inherit version information from release-please:

https://github.com/Flagsmith/flagsmith/blob/647712caf6e4e61eac18fca0a38ca7ee2beaadef/Dockerfile#L128

Then as the next step, `flagsmith-common` reads from that file into the `/version` endpoint ([ref](https://github.com/Flagsmith/flagsmith-common/blob/13c2016f5b384818e9e796dea226d045d8769094/src/common/core/utils.py#L69-L74)).

I'd lean to agreeing with @khvn26's [comment](https://github.com/Flagsmith/flagsmith/pull/5366#issuecomment-2820585967) below but I'm not a fan of the two context being coupled: whatever code implemented to `flagsmith-common` in order to display version info can only work if the effort starts in the `flagsmith` context — there's an implicit dependency there.

So according to the team's opinion, what should be the best path?

1. Implement a new middleware to `flagsmith-common` reusing the version info machinery that already exists in it. Add this middleware to `flagsmith` settings. Call it a day.
2. Move `/version` to `flagsmith` so all the implementation lives in it, removing the coupled code and implicit dependency. I don't know whether that endpoint is reused across other applications so that might not be the optimal solution.
3. Something else I can't see?

matthewelwell: Thanks for the write up. I think, based on these options, that (1) is the best option.

khvn26: `flagsmith-common` serves multiple roles at the moment:
- Code that needs to be shared between public (flagsmith/flagsmith) and private (saml, rbac, workflows-logic) codebases.
- Task processor library.
- "Platform library", i.e. common endpoints (healthchecks, `/metrics` `/version`), entrypoints (`flagsmith start`), common middleware, metrics tooling, and test helpers.

Taking the above into account, I believe option 1 makes sense.

api/tests/unit/api/test_api_version_http_header.py
```diff
@@ -0,0 +1,68 @@
...
+def clear_lru_caches() -> None:
+    get_file_contents.cache_clear()
+    get_versions_from_manifest.cache_clear()
+
+
+@pytest.mark.parametrize(
+    "url",
+    [
+        "/robots.txt",
+        # TODO: Add other GET-only URLs
```
matthewelwell: The only other unauthenticated endpoint that I can think of is `/version` but maybe that's a little too meta? Perhaps we should remove this parameterisation and add another test for an authenticated endpoint (e.g. `/users/me/`)? Or just use an authenticated client regardless and keep the parameterisation?

api/pyproject.toml
```diff
@@ -154,7 +154,7 @@ pygithub = "2.1.1"
 hubspot-api-client = "^8.2.1"
 djangorestframework-dataclasses = "^1.3.1"
 pyotp = "^2.9.0"
-flagsmith-common = "^1.10.0"
+flagsmith-common = { git = "https://github.com/flagsmith/flagsmith-common", branch = "feat/add-version-to-response-headers" }
```
matthewelwell: Just adding a comment here to remind us to update this once flagsmith-common is released.

matthewelwell: I wonder if we could have also deployed an alpha version of flagsmith-common to test it too, rather than reverting to a git dependency. Not too sure if we'd really benefit from doing that though.

api/tests/unit/api/test_api_version_http_header.py
emyller: Yeah also `/version` comes from `flagsmith-common` so it feels a bit unfair. I like using an authenticated client even for `/robots.txt` — after all this specific test isn't concerned with authentication capabilities or lack thereof.

api/pyproject.toml
emyller: Pinned to 1.11.0.

Regarding deploying an alpha: it may be a good idea but it feels like it adds too much bureaucracy for just one PR. Definitely the way to go when dealing with a bigger project IMHO.

matthewelwell (approved):

matthewelwell (approved):

PR #5369, opened 2025-04-22
feat: Add GIN index on `Identity.identifier`

khvn26:

## Changes

This PR improves `/api/v1/environments/{environment_api_key}/identities/` when searching through large quantities of identities.

Made sure the migration works as expected locally.

The index has been tested in staging and is working in prod for one of the enterprise customers.


api/environments/identities/models.py
```diff
@@ -40,6 +43,12 @@ class Meta:
         # avoid any downtime. If people using MySQL / Oracle have issues with poor performance on the identities table,
         # we can provide them the SQL to add it manually in a small window of downtime.
         index_together = (("environment", "created_date"),)
+        indexes = [
+            UnrestrictedGinIndex(
+                OpClass(Upper("identifier"), name="gin_trgm_ops"),
+                name="environments_identity_identifier_gin_idx",
```
gagantrivedi: Nit: Can we shorten the index name to just `identifier_gin_idx` to avoid having to subclass? It's not immediately clear what `UnrestrictedGinIndex` means just from the class name. If we do keep that class, could we add a docstring to explain its purpose?

gagantrivedi (changes requested):

api/environments/identities/migrations/0003_add_idenity_identifier_gin_index.py
```diff
@@ -0,0 +1,47 @@
...
+from django.contrib.postgres.operations import TrigramExtension
+from django.db import migrations
+import django.contrib.postgres.indexes
+import django.db.models.functions.text
+
+from core.migration_helpers import PostgresOnlyRunSQL
+import core.db
+
+
+_create_index_sql = "CREATE INDEX CONCURRENTLY IF NOT EXISTS environments_identity_identifier_gin_idx ON environments_identity USING GIN (UPPER(identifier) gin_trgm_ops);"
```
gagantrivedi: Is there a reason why we are not using https://docs.djangoproject.com/en/4.2/ref/contrib/postgres/operations/#django.contrib.postgres.operations.AddIndexConcurrently?

khvn26: Will it properly get skipped for non-Postgres? I'm using pure SQL mainly to utilise `PostgresOnlyRunSQL`.

api/environments/identities/models.py
khvn26: Renamed the index on staging to `identity_identifier_idx` and fixed in 1c5aa228cba4ed306ee423f3cb93c0f2bc4bba16 + 0a0bd885fe9bfd996b6ab174f7af596868e874bd 👍

api/environments/identities/migrations/0003_add_idenity_identifier_gin_index.py
gagantrivedi: Yeah, you're right — it will break for other databases. Perhaps I expected too much from Django

gagantrivedi (approved):

PR #5372, opened 2025-04-22
fix: metadata inputs not updating values

Zaimwa9:

## Changes
Metadata fields inputs could stay stuck on non up to date values

- Added `envData` dependency in AddMetadataToEntity useEffect
- Removed useState management of metadata value (that was only triggered on first rendering)

- Manually
1. Create an environment (or feature) metadata field
2. Set it 
3. Switch between environments

https://www.loom.com/share/f980f034f6144c548a2f30efd809f828


frontend/web/components/metadata/AddMetadataToEntity.tsx
```diff
@@ -333,22 +337,20 @@ const MetadataRow: FC<MetadataRowType> = ({
       <Flex className='table-column'>{`${metadata?.name} ${
         metadata?.isRequiredFor ? '*' : ''
       }`}</Flex>
-      {metadata?.type === 'bool' ? (
+     {metadata?.type === 'bool' ? (
         <Flex className='flex-row'>
           <Switch
-            checked={!!metadataValue}
+            checked={metadataValue === true || metadataValue === 'true'}
```
tiagoapolo: Just a tip to make the code less verbose in this case, you could use 
```suggestion
            checked={[true, 'true'].includes(metadataValue}
```

frontend/web/components/metadata/AddMetadataToEntity.tsx
```diff
@@ -100,7 +100,7 @@ const AddMetadataToEntity: FC<AddMetadataToEntityType> = ({
       onChange?.(metadataParsed as CustomMetadataField[])
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [metadataFieldsAssociatedtoEntity])
+  }, [metadataFieldsAssociatedtoEntity, entityId])
```
tiagoapolo: Do we need to listen for `entityId` changes? We're not using it inside of the useEffect
```suggestion
  }, [metadataFieldsAssociatedtoEntity])
```

api/environments/serializers.py
```diff
@@ -97,6 +98,16 @@ def get_project(
...
         )
 
+    def update(
+        self, instance: models.Model, validated_data: dict[str, typing.Any]
+    ) -> Environment:
+        metadata_items = validated_data.pop("metadata", [])
+        environment = super().update(instance, validated_data)  # type: ignore[no-untyped-call]
+        self.update_metadata(environment, metadata_items)
+        environment.refresh_from_db()
+        assert isinstance(environment, Environment)
```
emyller: Is this **needed**? Not a blocker.

emyller (approved):

api/environments/serializers.py
Zaimwa9: Not needed per say but mypy is a tough opponent and it would require unrelated typing that I think is overkill in the context of a bug fix

emyller: Fairy nuff. Let me know if you need help swinging a few punches.

khvn26: FWIW, the issue is tracked https://github.com/Flagsmith/flagsmith-common/issues/71.

khvn26 (approved):

PR #5377, opened 2025-04-23
feat: TCP health check integration + ECS health checks on staging

khvn26:

## Changes

This picks the `flagsmith healthcheck` command implemented in https://github.com/Flagsmith/flagsmith-common/pull/60 and adds it to ECS task definitions on staging. This is required for service discovery (and metric collection) to work.

Tasks now show up as healthy in ECS console:

image

### DIFF
--- a/api/pyproject.toml
+++ b/api/pyproject.toml
@@ -66,13 +66,7 @@ exclude_also = [
 omit = ["scripts/*", "manage.py"]
 
 [tool.pytest.ini_options]
-addopts = [
-  '--ds=app.settings.test',
-  '-vvvv',
-  '-p',
-  'no:warnings',
-  '--dist=worksteal',
-]
+addopts = ['--ds=app.settings.test', '-vvvv', '-p', 'no:warnings', '--dist=worksteal']
 console_output_style = 'count'
 
 [tool.mypy]
--- a/api/pyproject.toml
+++ b/api/pyproject.toml
@@ -160,7 +160,7 @@ pygithub = "2.1.1"
 hubspot-api-client = "^8.2.1"
 djangorestframework-dataclasses = "^1.3.1"
 pyotp = "^2.9.0"
-flagsmith-common = { git = "https://github.com/flagsmith/flagsmith-common", branch = "feat/flagsmith-healthcheck" }
+flagsmith-common = "^1.11.0"
 django-stubs = "^5.1.3"
 tzdata = "^2024.1"
 djangorestframework-simplejwt = "^5.3.1"
--- a/api/pyproject.toml
+++ b/api/pyproject.toml
@@ -66,7 +66,13 @@ exclude_also = [
 omit = ["scripts/*", "manage.py"]
 
 [tool.pytest.ini_options]
-addopts = ['--ds=app.settings.test', '-vvvv', '-p', 'no:warnings', '--dist=worksteal']
+addopts = [
+  '--ds=app.settings.test',
+  '-vvvv',
+  '-p',
+  'no:warnings',
+  '--dist=worksteal',
+]
 console_output_style = 'count'
 
 [tool.mypy]
@@ -154,7 +160,7 @@ pygithub = "2.1.1"
 hubspot-api-client = "^8.2.1"
 djangorestframework-dataclasses = "^1.3.1"
 pyotp = "^2.9.0"
-flagsmith-common = "^1.10.0"
+flagsmith-common = { git = "https://github.com/flagsmith/flagsmith-common", branch = "feat/flagsmith-healthcheck" }
 django-stubs = "^5.1.3"
 tzdata = "^2024.1"
 djangorestframework-simplejwt = "^5.3.1"
--- a/api/scripts/healthcheck.py
+++ b/api/scripts/healthcheck.py
@@ -1,32 +1,11 @@
-import argparse
-import socket
+import logging
+import sys
 
-parser = argparse.ArgumentParser(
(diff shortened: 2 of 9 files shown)


api/scripts/healthcheck.py
```diff
@@ -1,25 +1,31 @@
...
-        f"This healthcheck, invoked by {' '.join(sys.argv)}, is deprecated. "
-        f"Use the `{HEALTH_LIVENESS_URL}` endpoint instead."
-    )
-    status_code = requests.get(HEALTH_LIVENESS_URL).status_code
-
-    if status_code != 200:
-        logger.error(f"Health check failed with status {status_code}")
-
-    sys.exit(0 if 200 >= status_code < 300 else 1)
+    args = parser.parse_args()
```
gagantrivedi: Are these TCP health checks only used for service discovery, and only in SaaS environments?

khvn26: So far, yes — but we'll use them in Helm as well, see https://github.com/Flagsmith/self-hosted/blob/d1d19e77b5f22313024d49faa6e02728652fb721/flagsmith-values.yaml#L15-L22 and https://flagsmith.slack.com/archives/C07G7UAKPNH/p1745340333442389

gagantrivedi: Re https://flagsmith.slack.com/archives/C07G7UAKPNH/p1745340333442389

Does AWS alb support exec health checks?  Think it only supports http?

khvn26: The health checks added here are for Cloud Map only, and are expected to be run by ECS, which does not support HTTP.

infrastructure/aws/staging/ecs-task-definition-task-processor.json
```diff
@@ -6,7 +6,26 @@
     "containerDefinitions": [
         {
             "name": "flagsmith-task-processor",
-            "command": ["run-task-processor"],
+            "command": [
+                "run-task-processor"
+            ],
+            "portMappings": [
```
rolodato, line 12: Is this new port mapping necessary if the health check is being done by a shell command that connects to `localhost`, and not an external TCP/HTTP request?

infrastructure/aws/staging/ecs-task-definition-web.json
```diff
@@ -17,6 +17,16 @@
                     "protocol": "tcp"
                 }
             ],
+            "healthCheck": {
+                "command": [
+                    "CMD-SHELL",
+                    "python scripts/healthcheck.py"
```
rolodato: I would suggest adding an entrypoint command to `run-docker.sh` such as `healthcheck`, to abstract ECS/k8s/etc away from needing to know about Python or the path to the `healthcheck.py` file. That way we can change the healthcheck implementation easily without needing to modify how we're calling it.

api/scripts/healthcheck.py
```diff
@@ -1,25 +1,31 @@
...
-
-HEALTH_LIVENESS_URL = "http://localhost:8000/health/liveness"
-
-
-logger = logging.getLogger(__name__)
+import argparse
+import socket
+
+parser = argparse.ArgumentParser(
+    description="Check if the API is running and accessible.",
```
rolodato:
```suggestion
    description="Check if the API is able to accept local TCP connections.",
```

"Accessible" is not necessarily accurate if the request queue is full or something else (like DB being down, network etc) prevents it from working.

infrastructure/aws/staging/ecs-task-definition-task-processor.json
khvn26: This is actually needed for Prometheus metric collection in ECS.

infrastructure/aws/staging/ecs-task-definition-web.json
khvn26: This is done in 711218a9765674fac259641c8033daf687d9b5ea .

As discussed offline, we are now using `flagsmith healthcheck` commands.

api/pyproject.toml
```diff
@@ -154,7 +160,7 @@ pygithub = "2.1.1"
 hubspot-api-client = "^8.2.1"
 djangorestframework-dataclasses = "^1.3.1"
 pyotp = "^2.9.0"
-flagsmith-common = "^1.10.0"
+flagsmith-common = { git = "https://github.com/flagsmith/flagsmith-common", branch = "feat/flagsmith-healthcheck" }
```
khvn26: Change to released `flagsmith-common` before merging.

infrastructure/aws/staging/ecs-task-definition-task-processor.json
```diff
@@ -6,7 +6,26 @@
...
+            "healthCheck": {
+                "command": [
+                    "CMD-SHELL",
+                    "flagsmith healthcheck tcp"
+                ],
+                "interval": 10,
+                "timeout": 2,
+                "retries": 5,
+                "startPeriod": 5
+            },
```
emyller, line 28: @khvn26 Were you able to test this health check in ECS before this reaches production? Sorry I'm late to the party.

emyller (approved):
LGTM if the ECS health check works.

infrastructure/aws/staging/ecs-task-definition-task-processor.json
khvn26: Yup, see screenshot evidence in the PR description + Grafana dashboard wouldn't work without these health checks.

Issue #5370, opened 2025-04-22
Frontend is too eager when searching for identities


khvn26:
When searching for identities, frontend requests to `/api/v1/environments/{environment_api_key}/identities/?q={search_term}` are too aggressive.

We should consider:
1. Lowering the request rate in the frontend somehow.
1. Rate limiting the endpoint in the backend to avoid excess queries blowing up the database.

PR remediating the problem in the backend by adding an index: #5369

Aisha-gif2003:
Hi @khvn26 , I’m interested in working on this issue. Can you assign it to me?

srdasdev:
I want to work on this

matthewelwell:
@1-23-smy sure, please go ahead - I've assigned to you. Let us know if you need any help.

Zaimwa9:
Closed by [#5379](https://github.com/Flagsmith/flagsmith/pull/5379)

PR #5379, opened 2025-04-24
fix: Reduce frequency of identity search requests in frontend

Keerthi421:

Changes:
This PR addresses issue #5370 where the frontend was making too many aggressive API requests to the identities endpoint when searching. I've implemented the following changes:

-->Increased the throttle delay in useSearchThrottle.ts from 100ms to 500ms
-->Increased the default throttle value in useInfiniteScroll.ts from 100ms to 500ms
-->These changes significantly reduce the request rate from the frontend while maintaining reasonable responsiveness. The throttling mechanism now waits longer between keystrokes before sending requests, preventing unnecessary database load.
#5370


Zaimwa9 (approved):
Thanks for the contribution! It's working really well. 

While I was comparing with using debounce. I realized the custom implementation we have of `useThrottle` in `useSearchThrottle` is actually more of a `debounce` one.

I'm approving as of because it's great but if ever in bonus you feel like renaming `useThrottle` into `useDebounce` and `useSearchThrottle` into `useDebouncedSearch` that'd be awesome

Keerthi421:
@Zaimwa9 Sure , I will change

Keerthi421:
@Zaimwa9 , I have made changes please review it

Zaimwa9 (changes requested):
Thanks for the change.

One last thing to do and we are good to go:
- ~10 components are using the `useSearchThrottle`, replace it with your new useDebouncedSearch
- Remove the useSearchThrottle.ts file

And that should be good
image

Zaimwa9 (changes requested):
Thanks again. One last back and forth as it doesn't build as of.
```
 > [linux/amd64 build-node-django 1/1] RUN mkdir /build/api && cd frontend && npm run bundledjango:
#36 46.35       .js
#36 46.35         Field 'browser' doesn't contain a valid alias configuration
#36 46.35         /build/frontend/common/useThrottle.js doesn't exist
#36 46.35       as directory
#36 46.35         /build/frontend/common/useThrottle doesn't exist
#36 46.35  @ ./web/components/pages/UserPage.tsx 41:48-86
#36 46.35  @ ./web/routes.js 15:39-77
#36 46.35  @ ./web/main.js 11:37-56
#36 46.35 
#36 46.35 webpack 5.94.0 compiled with 1 error and 7 warnings in 44940 ms
```

There is a `import useThrottle from 'common/useThrottle'` in `TableSearchFilter` that needs to be directly replaced with `import useDebounce from 'common/useDebounce'
`
line 14 `const searchItems = useDebounce(`

Then we can approve

kyle-ssg:
Having tried this it feels a little sluggish to me due to a couple reasons:

- It doesn't take the leading input, so if I type 1 character it feels like I'm waiting for any form of feedback. If we took Twitter search or GitHub for example would always immediately react and then throttle.
- A 500ms delay seems too long, I believe that 200-300ms would replicate something a bit more natural.

tiagoapolo (approved):

PR #5381, opened 2025-04-24
fix: fix test in PR #5283

matthewelwell:
This is a PR on top of #5283 to resolve the tests. 

There were 2 issues which I've resolved in this PR: 

 1. The `updated_at` value wasn't actually by updated by just renaming the environment, as explained [here](https://github.com/Flagsmith/flagsmith/pull/5283))
 2. If-Modified-Since is only accurate to the nearest second, so I've added a short sleep in the test to ensure that the test always passes. We could equally have artificially set the 'If-Modified-Since' to an earlier time, but I like that we're using the time from the previous request. 

On top of this, I have also removed the `AuthenticatedRequest` class as it wasn't really being used correctly. Although we were annotating certain instances of the `request` instance with the class, it wasn't _actually_ an instance of the `AuthenticatedRequest` class. If we want to modify the behaviour, I think this would be a wider change outside of the scope of this PR.

### DIFF
--- a/api/environments/models.py
+++ b/api/environments/models.py
@@ -165,7 +165,9 @@ class Environment(
     @hook(AFTER_UPDATE)  # type: ignore[misc]
     def clear_environment_cache(self) -> None:
         # TODO: this could rebuild the cache itself (using an async task)
-        environment_cache.delete(self.initial_value("api_key"))
+        environment_cache.delete_many(
+            [self.initial_value("api_key"), *[eak.key for eak in self.api_keys.all()]]
+        )
 
     @hook(AFTER_UPDATE, when="api_key", has_changed=True)  # type: ignore[misc]
     def update_environment_document_cache(self) -> None:
--- a/api/tests/unit/environments/test_unit_environments_views_sdk_environment.py
+++ b/api/tests/unit/environments/test_unit_environments_views_sdk_environment.py
@@ -198,25 +198,28 @@ def test_environment_document_caching(
 
     # Then - first request should return 200 and include Last-Modified header
     assert response1.status_code == status.HTTP_200_OK
-    assert response1.headers["Last-Modified"] == http_date(
-        environment.updated_at.timestamp()
-    )
+    last_modified = response1.headers["Last-Modified"]
+    assert last_modified == http_date(environment.updated_at.timestamp())
+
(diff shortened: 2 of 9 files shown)


api/environments/models.py
```diff
@@ -165,7 +165,9 @@ def create_feature_states(self) -> None:
     @hook(AFTER_UPDATE)  # type: ignore[misc]
     def clear_environment_cache(self) -> None:
         # TODO: this could rebuild the cache itself (using an async task)
-        environment_cache.delete(self.initial_value("api_key"))
+        environment_cache.delete_many(
+            [self.initial_value("api_key"), *[eak.key for eak in self.api_keys.all()]]
+        )
```
matthewelwell, line 170: This is a temporary fix for now and I think we should perhaps asynchronise this. See further context in the conversation [here](https://flagsmith.slack.com/archives/C062HRM2JTD/p1743456515206609?thread_ts=1743174154.266529&cid=C062HRM2JTD).

api/tests/unit/environments/test_unit_environments_views_sdk_environment.py
```diff
@@ -196,23 +198,25 @@ def test_environment_document_caching(
...
 
     # Then - second request should return 304 Not Modified
     assert response2.status_code == status.HTTP_304_NOT_MODIFIED
 
+    # sleep for 1s since If-Modified-Since is only accurate to the nearest second
+    # https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/If-Modified-Since
+    time.sleep(1)
+
     # When - environment is updated
-    environment.clear_environment_cache()
```
matthewelwell, line 214: This isn't needed since it should be called on `environment.save()`

api/environments/models.py
rolodato: IMO the correct fix would be to only ever use a single cache key for environments, and not have one cache key per server-side SDK key.

To achieve this, the endpoints that accept server-side SDK keys must also receive the client-side key, which is confusing with the current terminology. For example, something like `GET /environment-document/abc123 -H "Authorization: ser..."` instead of deriving the environment ID from the server-side key.

rolodato (approved):

PR #5382, opened 2025-04-24
feat: add asynchronous task for rebuilding environment caches

matthewelwell:

## Changes

Update the logic to rebuild caches used by `Environment.get_from_cache()` instead of just clearing them when an update is made to an environment. 

Updated existing tests.

### DIFF
--- a/api/environments/managers.py
+++ b/api/environments/managers.py
@@ -1,9 +1,15 @@
+import typing
+
 from django.db.models import Prefetch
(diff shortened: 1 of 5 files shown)


matthewelwell:
I'm going to close this for now. I'm not sure why it was left in draft, but I can see that we have since implemented the logic to clear the cache for server api keys [here](https://github.com/Flagsmith/flagsmith/blob/main/api/environments/models.py#L168-L170).

Issue #5388, opened 2025-04-24
[Amplitude integration] Enabled feature with boolean value `false` is sent with value `true` to user properties


anton-marin:
### How are you running Flagsmith

- [x] Self Hosted with Docker
- [ ] Self Hosted with Kubernetes
- [ ] SaaS at flagsmith.com
- [x] Some other way (add details in description below)

### Describe the bug

When the feature is:
1. enabled
2. of boolean type
3. has value `false`

it will be incorrectly sent to the User Properties in Amplitude with the value `true`.

The error happens in [this line](https://github.com/Flagsmith/flagsmith/blob/9e3a273cf38ae71405a2ab3c7fb32dfae4c480ed/api/integrations/amplitude/amplitude.py#L43), if the value is `False`,  the property will be set to`feature_state.enabled`

### Steps To Reproduce

1. Set up Amplitude integration
2. Create a feature, make it enabled, and set the value to `false`
3. Get identity flags for any user (through SDK)
4. Go to Amplitude, check user properties for this user

Observed behavior
The value of the feature in Amplitude's user properties is `true`

### Expected behavior

The value of the feature in Amplitude's user properties is `false`

### Screenshots

This can be reproduced in the [Amplitude integration unit tests](https://github.com/Flagsmith/flagsmith/blob/9e3a273cf38ae71405a2ab3c7fb32dfae4c480ed/api/tests/unit/integrations/amplitude/test_unit_amplitude.py#L36) by changing the `init_value` of the feature in the fixture to `false` and expecting `False` in the value of the user property
![Image](https://github.com/user-attachments/assets/52bd3b17-ca4b-4ecd-8ed3-8d2bc146d880)

anton-marin:
Fixed in https://github.com/Flagsmith/flagsmith/pull/5389

PR #5389, opened 2025-04-24
fix: incorrect feature value passed in integrations

anton-marin:

## Changes

Reference: #5388 

In integrations (Segment, Heap, Amplitude), the features that are both enabled and have values that can be cast to `False` (bool `False` or int `0`), `True` was always passed in the feature properties instead of the actual value.

Fix: explicitly check for `None` and allow `0` and `False` to be correctly propagated to the feature properties

Added parametrization in the unit tests and included values on which the tests previously failed


api/conftest.py
```diff
@@ -590,10 +590,12 @@ def feature_state(feature: Feature, environment: Environment) -> FeatureState:
 
 
 @pytest.fixture()
-def feature_state_with_value(environment: Environment) -> FeatureState:
+def feature_state_with_value(
+    environment: Environment, initial_value: typing.Any
```
anton-marin: This fixture now requires an explicit parameterization argument or another fixture ([ref](https://stackoverflow.com/a/60148972)). Let me know if it's better instead of changing this fixture to define and call a function with `initial_value` as an argument, defined somewhere closer to the unit tests of integrations. This fixture is, however, currently used only in the unit tests of integrations

Zaimwa9: Thanks for the link and good addition.
Just to stay consistent with previous usage I would have a default "foo" value there `initial_value: typing.Any = "foo"` and keep it simple for now

api/integrations/amplitude/amplitude.py
```diff
@@ -40,7 +40,9 @@ def generate_user_data(
         for feature_state in feature_states:
             value = feature_state.get_feature_state_value(identity=identity)
             feature_properties[feature_state.feature.name] = (
-                value if (feature_state.enabled and value) else feature_state.enabled
+                value
+                if (feature_state.enabled and value is not None)
```
Zaimwa9, line 44: 👍

api/conftest.py
Zaimwa9:
```suggestion
    environment: Environment, initial_value: typing.Any = "foo"
```

Zaimwa9 (changes requested):
Great contribution thanks. I suggested just a small change to have a default value in case it's not provided. Then looks good to go

api/conftest.py
khvn26: An alternative approach I'd prefer would be to define the `initial_value` value fixture with `params` keyword. For an example, see the `admin_client_auth_type` implementation.

anton-marin: Hi @Zaimwa9,
Thanks for the review and your comment. Sadly, if we add a default value in `initial_value: typing.Any = "foo"`, pytest no longer recognizes it as an argument that needs parameterization injection.

I found another way using indirect parametrization ([ref1](https://stackoverflow.com/a/36087408), [ref2](https://docs.pytest.org/en/7.1.x/example/parametrize.html#deferring-the-setup-of-parametrized-resources)), which should preserve the default value if none was given. LMK what you think

anton-marin: hi @khvn26,
I refactored the fixture to use indirect parametrization with the `param` attribute of the `FeatureRequest`, a somewhat similar approach used in `admin_client_auth_type`, but slightly more lightweight. LMK if you would prefer `params` instead of `param` for consistency

khvn26: Looks nice 👍

khvn26 (approved):

anton-marin:
@khvn26 @Zaimwa9 I guess you need to authorize Vercel to run the CI, otherwise let me know the next steps. Feel free to merge it

Zaimwa9 (approved):

anton-marin:
@khvn26 @Zaimwa9 could you merge it? I don't have the "Merge" button, and I do need this fix

PR #5390, opened 2025-04-25
chore: show deleted organisations in sales dashboard

matthewelwell:

## Changes

The main purpose of this PR was to add the ability to show deleted organisations in the sales dashboard. Since I was already working in the area, I also tried to tackle something that has been bothering me for a while, which is the performance of the sales dashboard. As such, I tried to optimise the annotations to only annotate the queryset with any annotation that was needed for sorting (as you can see [here](https://github.com/Flagsmith/flagsmith/commit/116f5c975d807094f9fa079cdc100e98b795b5c0)) but this didn't actually change anything due to the behaviour of django's lazily evaluated querysets. Without diving deeper into the pagination logic, I'm not sure there's an option here to optimise the queries that are needed. 

Validated the 'include deleted' logic locally when running the API.

### DIFF
--- a/api/sales_dashboard/templates/sales_dashboard/home.html
+++ b/api/sales_dashboard/templates/sales_dashboard/home.html
@@ -29,6 +29,8 @@
 
       <div class="sort-form">
         <form action="{% url 'sales_dashboard:index' %}" method="get" class="form-inline float-right">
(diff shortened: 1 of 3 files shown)


Zaimwa9 (approved):
👍 

image
image
image

PR #5391, opened 2025-04-25
fix: replaced-undefined-with-null-when-disabling-change-requests

Zaimwa9:

## Changes
Unchecking change requests was not correctly updating the value

- replaced undefined value when unchecked with null

https://www.loom.com/share/9f7a9b26b77d42f4803b279c57968f1e

### DIFF
--- a/frontend/web/components/pages/EnvironmentSettingsPage.tsx
+++ b/frontend/web/components/pages/EnvironmentSettingsPage.tsx
@@ -187,9 +187,12 @@ const EnvironmentSettingsPage: React.FC<EnvironmentSettingsPageProps> = ({
       return
     }
 
-    if (
-      environmentData?.use_v2_feature_versioning && !!currentEnv &&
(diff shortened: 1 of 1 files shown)


frontend/web/components/pages/EnvironmentSettingsPage.tsx
```diff
@@ -188,7 +188,7 @@ const EnvironmentSettingsPage: React.FC<EnvironmentSettingsPageProps> = ({
     }
 
     if (
-      environmentData?.use_v2_feature_versioning &&
+      environmentData?.use_v2_feature_versioning && !!currentEnv &&
       !currentEnv?.use_v2_feature_versioning
     ) {
```
tiagoapolo: can we instead use something like this to make the logic more explicit
```
const isEnvDataVersioningEnabled = environmentData?.use_v2_feature_versioning
const isCurrentEnvVersioningDisabled = !!currentEnv &&
      !currentEnv?.use_v2_feature_versioning
```

tiagoapolo (approved):

Issue #5030, opened 2025-01-21
Editing a scheduled change's go-live date still causes tasks to execute at the original date


rolodato:
### How are you running Flagsmith

- [x] Self Hosted with Docker
- [x] Self Hosted with Kubernetes
- [x] SaaS at flagsmith.com
- [ ] Some other way (add details in description below)

### Describe the bug

Editing a scheduled change to happen in the future will still cause scheduled tasks such as webhooks, Slack notifications and audit logs to execute at the original date. This can cause false alarms and incorrect audit log records, which decreases trust in the platform.

### Steps To Reproduce

1. In an environment with versioning and change requests disabled, schedule any flag update for ~2 mins in the future
2. Edit the created scheduled change to be later, e.g. in 1 hour
3. After 2 minutes, an audit log entry is created for the supposed scheduled change that went live. Integrations such as webhooks and Slack notifications will also fire, but the flag was never changed.

### Expected behavior

Run scheduled tasks at the expected times.

### Screenshots

These audit logs are from these events:

* The scheduled change was created, originally as `Enable my_feat in 2 min`
* Renamed the scheduled change to `Enable my_feat in 1 hour` and edited the deadline - no audit log entry is shown
* The scheduled tasks for the original go-live date ran, but the flag was not updated.

![Image](https://github.com/user-attachments/assets/83199590-8102-4ec5-9e08-5571e2e47b5a)

There are also some UI/timezone issues here:

* The times are displayed in 24-hour format but with am/pm, e.g. `18:28pm`. This is redundant and not used anywhere to my knowledge.
* The latest audit log entry's time is incorrect. It is currently 17:38, so this is showing a time in the future (18:28pm)
* The "Flag state / Remote Config" entry's date is shown in the server's timezone (running in Kubernetes, so likely UTC) but the Date column is shown in my local timezone (UTC-3).

emyller:
I was able to reproduce the issue and collected some data. Some tests:

### CASE 1: SCHEDULE UPDATE AND WAIT ✅
- 11:57: Disabled feature
    ```
    Flag state updated for feature: my_feature
    ```
- 11:57: Scheduled for 12:00
    ```
    Change Request: go3 created
    Change Request: go3 committed
    ```
- 12:00: Feature updated
    ```
    Scheduled change to Flag state / Remote config value went live for feature: my_feature by Change Request: go3
    ```

### CASE 2: SCHEDULE UPDATE, THEN POSTPONE IT 💥
[...]

PR #5393, opened 2025-04-25
feat: backend-clone-segments

Zaimwa9:

## Changes
This PR implements the backend cloning/duplication of segments

- New `CloneSegmentSerializer`, requiring a name and performing the duplication
- New endpoint `/api/v1/projects/{projectId}/segments/{segmentId}/clone/` accepting a (required) name as body payload
- Metadata new method `deep_clone_for_new_entity`, taking an entity (segment/env/feature) to deep clone and create the related new metadata

- New tests
- Postman requests
- Fullstack branch `feat/clone-segments-and-feature-segments`

**Project Segment creation**
https://www.loom.com/share/27ec8d4a2d8d4c24b443ac5d2b12231e?focus_title=1&muted=1&from_recorder=1

**Feature specific segment creation**
https://www.loom.com/share/8cc4d8e92ce84c459dfb6598ebb42acb


api/segments/models.py
```diff
@@ -147,6 +147,49 @@ def set_version_of_to_self_if_none(self):  # type: ignore[no-untyped-def]
         self.version_of = self
         self.save_without_historical_record()
 
+    def _clone_segment_rules(self, cloned_segment: "Segment") -> list["SegmentRule"]:
```
Zaimwa9, line 147: @gagantrivedi , we were discussing with @emyller about moving those types of methods into a Segment dedicated service. It seems it has been done for some apps.
Do you have any opinion on this 🙏 ?

emyller: I was particularly interested in separating this new functionality into a new service isolating all the code that's responsible for cloning the segment and its components, such as metadata and rules.

```py
class SegmentCloneService:
    def __init__(self, segment: Segment):
        self._original_segment = segment

    def make_clone(self) -> Segment: ...
        cloned_segment = Segment(...)
        self._clone_metadata_into(cloned_segment)
        self._clone_rules_into(cloned_segment)
        return cloned_segment

    def _clone_metadata_into(self, cloned_segment: Segment) -> None: ...
    def _clone_rules_into(self, cloned_segment: Segment) -> None: ...
```

This could help consolidating a place where all the cloning code should live, rather than having it scattered around models, views, serializers — or wherever else a debate might lead to. It would likely make testing easier too.

gagantrivedi: Yeah, we can move this to the service layer

Zaimwa9: @khvn26 @emyller. Version with service ready (all tests pass fyi). 
Relevant files are:
- `api/core/workflows_services` (I had to create a Committer service as `_publish_segments()` was relying on cloning methods
- `api/segments/services`
- `api/tests/unit/segments/test_unit_segments_services.py`

I opted for **class based services** because the cloning operations all operate on a shared Segment instance and involve coordinated steps. So the benefits I see using a class:
- We encapsulate the segment context once instead of passing it to every function.
- The related operations (`_clone_rules`, `_clone_metadata`) are grouped within the cloning context and scoped clearly
- We avoid duplication of validation and state across methods
- I don't like this argument of "future extension" but comparing with top level functions, it does actually provide a clean base for composability without needing to refactor later if the question arises
- As pointed out by Evandro it improves testability

I just want to align with you on the approach. If we take this path, there are just 2 items left to address:
- typing
- `flagsmith-common` and `flagsmith-workflows` have each one `instance.deep_clone()` to clean up (hence why it's duplicated in the model)

api/tests/unit/segments/test_unit_segments_services.py
```diff
@@ -0,0 +1,265 @@
+from unittest.mock import MagicMock, patch
```
Zaimwa9: FYI, I gave Agent another try following our meeting. 
The tests has been generated by Claude3.7.
It needed a couple of adjustments (class-test, errors in mocking, couple of wrong imports, missing some model hooks behavior..).
15-20minutes to clean-up so overall beneficial here

api/tests/unit/segments/test_unit_segments_services.py
```diff
@@ -0,0 +1,265 @@
...
+    # Create a real ChangeRequest object
+    change_request = ChangeRequest.objects.create(
+        project=project, title="Test change request", user=None
+    )
+
+    # Initialize the service
+    cloner = SegmentCloner(segment)
+
+    # When - properly mock the history attribute and its update method
+    with patch.object(Segment, "history", create=True) as mock_history:
```
khvn26: We don't have to use the context manager interface if we use the pytest_mock `mocker` fixture

api/tests/unit/segments/test_unit_segments_services.py
```diff
@@ -0,0 +1,265 @@
...
+
+        # Call the method under test
+        cloned_segment = cloner.shallow_clone(
+            name="Shallow Cloned Segment",
+            description="New description",
+            change_request=change_request,
+        )
+
+        # Verify history update was called
+        mock_update.assert_called_once()
```
khvn26: Yeah, these redundant comments are a very strong hint that the test was AI generated :)

Zaimwa9: 😆 i'll go over them again in any case. I wanted to align on the pattern before spending too much time on reviewing the tests

api/features/workflows/core/models.py
```diff
@@ -117,107 +112,93 @@ def approve(self, user: "FFAdminUser"):  # type: ignore[no-untyped-def]
...
-            # of the target (i.e., canonical) segment.
-            target_segment.name = segment.name  # type: ignore[union-attr]
-            target_segment.description = segment.description  # type: ignore[union-attr]
-            target_segment.feature = segment.feature  # type: ignore[union-attr]
-            target_segment.save()  # type: ignore[union-attr]
-
-            # Delete the rules in order to replace them with copies of the segment.
-            target_segment.rules.all().delete()  # type: ignore[union-attr]
-            for rule in segment.rules.all():
-                rule.deep_clone(target_segment)  # type: ignore[arg-type]
```
emyller, line 220: I like the refactor to `api/core/workflows_services.py` but it feels very out of scope to this pull request, and possibly prone to conflicts [soon](https://github.com/Flagsmith/flagsmith/issues/5030). Can we somehow integrate the work we're doing to `SegmentCloner` in here, and maybe coordinate this refactor in the near future?

api/metadata/models.py
```diff
@@ -30,41 +31,46 @@ class FieldType(models.TextChoices):
...
         max_length=255, choices=FieldType.choices, default=FieldType.STRING
     )
-    description = models.TextField(blank=True, null=True)
-    organisation = models.ForeignKey(Organisation, on_delete=models.CASCADE)
+    description: models.TextField[str | None, str | None] = models.TextField(
+        blank=True, null=True
+    )
+    organisation: models.ForeignKey[Organisation, Organisation] = models.ForeignKey(
+        Organisation, on_delete=models.CASCADE
+    )
```
emyller: All this typing should really be taken care of either by Django itself, or via the [mypy integration](https://github.com/typeddjango/django-stubs) we have installed. I'd bet you tried without this and had issues — ping me anytime and let's pair through this if needed. 😉

api/segments/models.py
```diff
@@ -147,26 +147,21 @@ def set_version_of_to_self_if_none(self):  # type: ignore[no-untyped-def]
...
+        cloned_rules = []
+        for rule in self.rules.all():
+            cloned_rule = rule.deep_clone(cloned_segment)
+            cloned_rules.append(cloned_rule)
+        cloned_segment.refresh_from_db()
+        assert (
+            len(self.rules.all())
+            == len(cloned_rules)
+            == len(cloned_segment.rules.all())
+        ), "Mismatch during rules creation"
```
emyller, line 157: I appreciate the guardrail, but I'd rather ensure this always works with tests, than allowing the possibility of a bad UX.

emyller: I see now that this was already present in the code. Feel free to leave as-is, or consider improving our confidence with tests — I believe they'd fit the scope of this work.

api/segments/services.py
```diff
@@ -0,0 +1,88 @@
...
+        cloned_rules = []
+        for rule in self.segment.rules.all():
+            cloned_rule = rule.deep_clone(cloned_segment)
+            cloned_rules.append(cloned_rule)
+        cloned_segment.refresh_from_db()
+        assert (
+            len(self.segment.rules.all())
+            == len(cloned_rules)
+            == len(cloned_segment.rules.all())
+        ), "Mismatch during rule cloning"
```
emyller, line 91: [Related thread](https://github.com/Flagsmith/flagsmith/pull/5393/files#r2101190258).

api/segments/services.py
```diff
@@ -0,0 +1,88 @@
...
+    def _clone_segment_metadata(self, cloned_segment: Segment) -> None:
+        cloned_metadata = []
+        for metadata in self.segment.metadata.all():
+            cloned_metadata.append(metadata.deep_clone_for_new_entity(cloned_segment))
+        cloned_segment.refresh_from_db()
+        assert (
+            len(self.segment.metadata.all())
+            == len(cloned_metadata)
+            == len(cloned_segment.metadata.all())
+        ), "Mismatch during metadata cloning"
```
emyller, line 102: [Related thread](https://github.com/Flagsmith/flagsmith/pull/5393/files#r2101190258).

api/segments/services.py
```diff
@@ -0,0 +1,88 @@
...
+from copy import deepcopy
+from typing import TYPE_CHECKING, Optional
+
+from segments.models import Segment
+
+if TYPE_CHECKING:
+    from features.workflows.core.models import ChangeRequest
+
+
+class SegmentCloner:
```
emyller: IMO `SegmentCloneService` would better fit the naming pattern we have.

api/tests/unit/segments/test_unit_segments_services.py
emyller: A few notes:
- The tests seem a bit redundant without reason, e.g. I think `test_clone_complex_segment_rules` supersedes `test_clone`.
- In the view tests (`api/tests/unit/segments/test_unit_segments_views.py`), we're currently taking a _functional testing_ approach to validate the final product through the view. Now that we're planning service-specific tests, and also since these are unit tests, I think the view tests should concern its own context and only ensure the service is called correctly rather than repeating the same verification from service tests.
- nit: test names can be more meaningful, e.g. `test_SegmentCloneService_clones_segment_components_recursively` for checking deep cloning.

api/features/workflows/core/models.py
Zaimwa9: I get what you mean. it's technically a bit out of scope. But the main reason I pulled `commit()` out into a service is to improve the structure overall. It didn't feel right to me to inject SegmentCloner back into the model. It brings back coupling between apps, which we're trying to move away from.

Having the ChangeRequestCommitter encapsulate that logic feels cleaner to me (especially as the code is strictly the same). I also did it with [this PR](https://github.com/Flagsmith/flagsmith/pull/5464) in mind, where it will allow to remove `model.commit` entirely and call the service directly from `flagsmith-workflows`. So this helps prep for that.

Thanks for sharing the ticket, I was unaware of it and it would make sense to coordinate but as most of the methods are held in the model atm I think the conflicts should be limited (the usage remains ISO actually being `change_request.commit(user)`

api/segments/services.py
Zaimwa9: If it's ok i will leave it for another time as this starts to get quite big

api/tests/unit/segments/test_unit_segments_services.py
Zaimwa9: Agreed, thanks for the comment. Simplified the view tests and reworked the service tests (disappointed with claude actually)

api/segments/models.py
```diff
@@ -147,26 +147,21 @@ def set_version_of_to_self_if_none(self):  # type: ignore[no-untyped-def]
...
 
+        return cloned_rules
+
+    # TODO: To be depreacted in flagsmith-common and flagsmith-workflows
     def deep_clone(self) -> "Segment":
```
Zaimwa9, line 162: @khvn26 @emyller 
What I said in a previous comment about having `flagsmith-common` and `flagsmith-workflows` use the services to replace `instance.deep_clone` and `instance.commit` is actually wrong as api depends on them, not the other way around.

To DRY, as of, I think about:
- either services live in common if needs be
- either as a short term solution, imitate what I temporarily did above with [commit](https://github.com/Flagsmith/flagsmith/pull/5393/files#:~:text=(self).-,commit,-(committed_by)) and leave the model method as an entrypoint for external repository.

N.B the workflows services should be flagsmith-workflows.
Wdyt ?

api/features/workflows/core/models.py
emyller: IMO the benefit from this improvement is already wider than discussing whether we should do it now or later. 😉  If the team agrees with keeping this work on this PR, I'm onboard.

api/segments/models.py
emyller: TBH I am not knowledgeable enough yet about the architecture to provide any helpful directions here. That said, the second option sounds better to me, but I'd ultimately defer to the team.

khvn26: I think migrating the services to common is fine; eventually, we should think on migrating all the models mentioned in common's `types` modules as well.

api/features/workflows/core/models.py
khvn26: > Thanks for sharing the ticket, I was unaware of it and it would make sense to coordinate but as most of the methods are held in the model atm I think the conflicts should be limited (the usage remains ISO actually being change_request.commit(user)

I agree with this assessment. @emyller has confirmed no code has been written on #5030 as of this moment, so I'm in favour of keeping this change.

api/segments/models.py
Zaimwa9: Is it fine if I do it in a separate PR ?

khvn26: Sounds good, as long as we have an issue for it.

api/core/workflows_services.py
```diff
@@ -0,0 +1,120 @@
...
+from features.workflows.core.exceptions import ChangeRequestNotApprovedError
+from segments.models import Segment
+from segments.services import SegmentCloneService
+
+if TYPE_CHECKING:
+    from features.workflows.core.models import ChangeRequest
+    from users.models import FFAdminUser
+
+
+class ChangeRequestCommitter:
```
emyller:
```suggestion
class ChangeRequestCommitService:
```

api/segments/models.py
Zaimwa9: [Here](https://github.com/Flagsmith/flagsmith/issues/5485#issue-3085794855)

api/metadata/models.py
emyller: Fixed in 79b9d0d4684a661eb545f384382d53c5d1732610.

api/tests/unit/segments/test_unit_segments_services.py
```diff
@@ -0,0 +1,315 @@
...
+    with pytest.raises(AssertionError, match="Mismatch during rule cloning"):
+        service.clone("Error Segment")
+
+
+def test_SegmentCloneService_clone__raises_when_cloned_metadata_count_mismatches(
+    mocker: MockerFixture,
+    segment: Segment,
+    required_a_segment_metadata_field: MetadataModelField,
+) -> None:
+    """Test that assertions are raised when metadata cloning fails"""
```
emyller: Unnecessary docstring.

api/tests/unit/segments/test_unit_segments_services.py
```diff
@@ -0,0 +1,315 @@
...
+    assert cloned_conditions2[1].operator == original_condition3.operator
+    assert cloned_conditions2[1].value == original_condition3.value
+    assert cloned_conditions2[1].created_with_segment is False
+
+
+def test_SegmentCloneService_clone__raises_when_cloned_rules_count_mismatches(
+    mocker: MockerFixture,
+    segment: Segment,
+) -> None:
+    """Test that assertions are raised when rule cloning fails"""
```
emyller: Unnecessary docstring.

api/tests/unit/segments/test_unit_segments_views.py
```diff
@@ -1556,3 +1557,51 @@ def test_include_feature_specific_query_filter__false(
     # Then
     assert response.json()["count"] == 1
     assert [res["id"] for res in response.json()["results"]] == [segment.id]
+
+
+def test_clone_segment(
```
emyller:
```suggestion
def test_clone_endpoint__uses_segment_clone_service(
```

Adds more meaning. 😉

api/tests/unit/segments/test_unit_segments_views.py
```diff
@@ -1556,3 +1557,51 @@ def test_include_feature_specific_query_filter__false(
...
+    project: Project,
+    admin_client: APIClient,
```
emyller, line 1564: ~I think `admin_client` comes from pytest-django rather than DRF's `APIClient`.~

api/tests/unit/segments/test_unit_segments_views.py
```diff
@@ -1556,3 +1557,51 @@ def test_include_feature_specific_query_filter__false(
...
+        "api-v1:projects:project-segments-clone", args=[project.id, segment.id]
+    )
+    new_segment_name = "cloned_segment"
+    data = {
+        "name": new_segment_name,
+    }
+    # When
+    response = admin_client.post(
+        url, data=json.dumps(data), content_type="application/json"
+    )
```
emyller, line 1586: ~If you use an actual instance of `APIClient` rather than `admin_client`, the JSON bits here should be built-in.~

api/tests/unit/segments/test_unit_segments_views.py
```diff
@@ -1556,3 +1557,51 @@ def test_include_feature_specific_query_filter__false(
...
+
+    # Then
+    assert response.status_code == status.HTTP_201_CREATED
+
+    response_data = response.json()
+    assert response_data["name"] == new_segment_name
+    assert response_data["project"] == project.id
+    assert response_data["id"] != segment.id
```
emyller, line 1594: This feels like a functional test, in which we're testing the service _again_. Not a blocker though the test would be slightly better with:

```py
# Given
service = mocker.patch.object(views_module, 'SegmentCloneService')
service().clone.return_value = Segment(...)
```

This would allow the test to ensure the view is doing what it should, but without poking into external territory.

api/tests/unit/segments/test_unit_segments_views.py
```diff
@@ -1556,3 +1557,51 @@ def test_include_feature_specific_query_filter__false(
...
+    url = reverse(
+        "api-v1:projects:project-segments-clone", args=[project.id, segment.id]
+    )
+    data = {
+        "no-name": "",
+    }
+    # When
+    response = admin_client.post(
+        url, data=json.dumps(data), content_type="application/json"
+    )
```
emyller, line 1612: ~Similar to the above.~

emyller (changes requested):
Looking great. A few minor adjustments needed — sorry we didn't have time while pairing today, but we're close!

api/tests/unit/segments/test_unit_segments_views.py
emyller: I am sorry, the above was me assuming `admin_client` was [pytest-django's](https://pytest-django.readthedocs.io/en/latest/helpers.html#admin-client-django-test-client-logged-in-as-admin) `admin_client`. Rather, it is an instance of DRF's `APIClient`, as per this fixture:

https://github.com/Flagsmith/flagsmith/blob/fd1e63cab0e2ed50bdf951511c7ce3248297820f/api/conftest.py#L234

Please ignore the above. :)

emyller (approved):
LGTM

PR #5394, opened 2025-04-25
feat: frontend-clone-segments 🔴  Blocked by #5393

Zaimwa9:

## Changes
🔴 Blocked by #5393 (`feat/backend-clone-segments`)

This PR implements the frontend part to consume the new clone segment API endpoint and allow the duplication of existing segments

- ⚠️  Introduces new flag `clone_segment`

- Reviewed UI to use a dropdown instead of trash button in `SegmentRow`
- `cloneSegment` mutation to consume `/api/v1/projects/{projectId}/segments/{segmentId}/clone/` accepting a (required) name as body payload
- Refactored `SegmentsPage` to extract `SegmentRow`
- Extracted `ActionItem` as single component, `calculateListPosition` as utils

- New E2E tests
- Postman requests
- Fullstack branch `feat/clone-segments-and-feature-segments`

image

**Project Segment creation**
https://www.loom.com/share/27ec8d4a2d8d4c24b443ac5d2b12231e?focus_title=1&muted=1&from_recorder=1

**Feature specific segment creation**
https://www.loom.com/share/8cc4d8e92ce84c459dfb6598ebb42acb


frontend/e2e/tests/segment-test.ts
```diff
@@ -125,9 +125,17 @@ export const testSegment1 = async () => {
   await waitAndRefresh()
   await assertTextContent(byId('user-feature-value-0'), '"medium"')
 
+  const isCloneSegmentEnabled = await isElementExists('segment-action-0')
+  if (isCloneSegmentEnabled) {
```
Zaimwa9: @tiagoapolo , so here my challenge was that I am using a feature flag ... I'd love your opinion. Because I identified 2 options - I tried both and they "could" work:
1. Integrates `flagsmith-node`:
      Use the flagsmith API key provided in the tests to get the flag value.
      **Pros**
      Properly reacts according to the flag
      **Cons**
      It adds a point of failure (external request) + potentially more maintenance of the E2E env and flags

2. Use  some `isElementExist`:
    Relies on proper element IDs and selector to ensure a feature is here or not. It looks a bit more hacky but in a way it really adapts to the UI on the spot and force to properly test the 2 paths.

In terms of process, I believe we should make sure the feature flag is ON on staging when releasing, so staging is tested against the new feature while prod ensure the backward compatibility.

In all cases, as soon as the feature is validated it would need a cleaning of the tests.

I'll keep looking at best flagged E2E best practices and let me know what you think

Zaimwa9: @kyle-ssg maybe you already dealt with this case ?

frontend/common/utils/calculateListPosition.ts
```diff
@@ -0,0 +1,12 @@
+export function calculateListPosition(
```
kyle-ssg, line 2: Maybe a comment here, prior to reading the rest of the PR it's unclear what this does

frontend/web/components/segments/SegmentRow/SegmentRow.tsx
```diff
@@ -0,0 +1,160 @@
...
+import SegmentAction from './components/SegmentAction'
+import ConfirmCloneSegment from 'components/modals/ConfirmCloneSegment'
+import { useCloneSegmentMutation } from 'common/services/useSegment'
+
+interface SegmentRowProps {
+  segment: Segment
+  index: number
+  projectId: string
+  router: RouterChildContext['router']
+  removeSegment: MutationTrigger<
```
kyle-ssg, line 24: Awesome! I never knew about this.

Would 

```
{
        projectId: number | string
        id: number
      }
```

Simply be Req['removeSegment']

frontend/e2e/tests/segment-test.ts
kyle-ssg: flagsmith/isomorphic can do this! See versioning-test. 

```
    await flagsmith.init({fetch,environmentID:Project.flagsmith,api:Project.flagsmithClientAPI})
    const hasFeature = flagsmith.hasFeature("feature_versioning")
```

If we wanted to start doing this more, we can make an e2e identity/segment or put them into the flagsmith_team segment.

Zaimwa9: Works like a charm, thanks

tiagoapolo (approved):
Can you reference the new components that are using react router in https://github.com/Flagsmith/flagsmith/pull/5442?

PR #5407, opened 2025-04-30
feat: Render usage report in sales dashboard usage directly, without sending email

rolodato:

## Changes

Previously, we were requiring users to enter an email address so that they could receive these reports:

image

This is not ideal for several reasons:

* Not all self-hosted customers have an email provider set up. This change lets these customers access reports without needing to set up email.
* Even if all customers did have email providers, email is a generally awful medium to work with. HTML, PDF, etc are much easier to use and test, for both us and customers.
* The email report was synchronously generated and sent as part of the same request; we are not sacrificing performance by returning the output via email or HTML.

This change also intentionally breaks the `/sales-dashboard/email-usage` route by renaming it to `/sales-dashboard/usage`. It is extremely unlikely that this will actually break anyone's links, considering it's not a documented route and this route would be used ~2x a year at most.

## Alternatives considered

If we really want to keep emails for some reason, we could have a plain-text report be sent directly from the client, by using a `mailto` link from the same HTML-rendered report. I decided to keep HTML only for simplicity.

Manually by browsing to `/sales-dashboard/usage` on any API instance.


matthewelwell (approved):
Happy for you to add `# pragma: no cover` where needed if it's easier than writing tests to cover the view.

Issue #5308, opened 2025-04-07
GET user permissions per project and environment


tiagoapolo:
### Is your feature request related to a problem?

Users can now have permissions directly, through groups, through roles, or through roles attached to groups.  Most of times it's difficult and time consuming to debug which permissions a user has. Also, there's no dedicated endpoint that returns all users permissions on a project and environment level. 

### Describe the solution you'd like.

We need an endpoint for providing a users permissions and where the permission is coming from, derived from group, roles, roles attached to groups and direct permissions. Also, display permissions per project and environment level (e.g. User A is in Group B ( Permissions come from Role C, which are "View Project", ...)

### Describe alternatives you've considered

`/api/v1/environments/{environment_api_key}/user-permissions/{id}/all`
`/api/v1/projects/{project_id}/user-permissions/{id}/all`
`/api/v1/organisations/{org_id}/user-permissions/{id}/all`

```
{
    "admin: boolean,
    "permissions":  
    [
       {
          "permission_key": string,
          "tags": number[],
          "is_directly_granted": boolean, // if true means that is also a user granted permission
          "derived_from": {
              "groups": Group[] // or for simplicity sake just {id: number; name: string }
              "roles": Role[] // or for simplicity sake just {id: number; name: string }
          }
      },
      ...
   ]
}
```

gagantrivedi:
I have created the following api:
![Image](https://github.com/user-attachments/assets/93e7a473-099f-4643-8755-6770658d0ce7)

The only difference in the response structure is that tags is part of the role instead of permissions, because a user can have the same permission through different roles with different tags.

tiagoapolo:
That works for me 😊

gagantrivedi:
Implemented here: https://github.com/Flagsmith/flagsmith/pull/5408

PR #5375, opened 2025-04-23
feat: Inspect Permissions

tiagoapolo:
Thanks for submitting a PR! Please check the boxes below:

- [ ] I have added information to `docs/` if required so people know about the feature!
- [x] I have filled in the "Changes" section below?
- [x] I have filled in the "How did you test this code" section below?
- [x] I have used a [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) title for this Pull Request

## Changes

Ref: [#5064](https://github.com/Flagsmith/flagsmith/issues/5064)

## How did you test this code?

Screenshot 2025-04-23 at 09 59 39
Screenshot 2025-04-23 at 09 59 33
Screenshot 2025-04-23 at 09 59 26

PR #5408, opened 2025-05-01
feat(5308): Add endpoints to support permission debugging

gagantrivedi:

## Changes

Implement Endpoints defined [here](https://github.com/Flagsmith/flagsmith/issues/5308)

- tested manually with rbac installed to make sure tags/role stuff is working correctly
- Adds unit tests

### DIFF
--- a/api/permissions/permissions_calculator.py
+++ b/api/permissions/permissions_calculator.py
@@ -19,7 +19,6 @@ from projects.models import (
 
 from .permission_service import is_user_project_admin
 from .rbac_wrapper import (  # type: ignore[attr-defined]
-    RolePermissionData,
     get_roles_permission_data_for_environment,
     get_roles_permission_data_for_organisation,
     get_roles_permission_data_for_project,
@@ -43,7 +42,7 @@ GroupPermissionQs = QuerySet[
 ]
 
 
-@dataclass
+@dataclass(kw_only=True)
 class _PermissionDataBase:
     admin: bool = False
     permissions: typing.Set[str] = field(default_factory=set)
@@ -59,7 +58,14 @@ class GroupData:
 class RoleData:
     id: int
     name: str
-    tags: typing.List[int]
+    tags: typing.Set[int]
+
+
+@dataclass
+class RolePermissionData(
+    _PermissionDataBase,
+):
+    role: RoleData
 
 
 @dataclass
@@ -104,9 +110,7 @@ class PermissionData:
 
     user: UserPermissionData
     groups: typing.List[GroupPermissionData]
-    roles: typing.List[  # type: ignore[valid-type]
-        "RolePermissionData" if "RolePermissionData" in globals() else typing.Any
-    ]
+    roles: typing.List[RolePermissionData]
     is_organisation_admin: bool = False
     admin_override: bool = False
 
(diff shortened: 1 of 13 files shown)


api/permissions/permissions_calculator.py
```diff
@@ -29,7 +28,7 @@
     from environments.models import Environment
     from users.models import FFAdminUser
 
-
+# Some random comment
```
Zaimwa9:
```suggestion
```

api/permissions/permissions_calculator.py
```diff
@@ -127,14 +164,67 @@ def tag_based_permissions(self) -> list[dict]:  # type: ignore[type-arg]
...
+            if permission_key not in permission_map:
+                permission_map[permission_key] = DetailedPermissionsData(
+                    permission_key=permission_key,
+                    is_directly_granted=bool(group is None and role is None),
+                    derived_from=PermissionDerivedFromData(),
+                )
+            if group:
+                permission_map[permission_key].derived_from.groups.append(group)
+            if role:
+                permission_map[permission_key].derived_from.roles.append(role)
```
Zaimwa9: Duplication of derived roles actually come from here (cf [thread](https://github.com/Flagsmith/flagsmith/pull/5375#discussion_r2079967616)). 

The root cause is in rbac that is returning a duplicated list. 

But here we should ensure deduplication of the payload until it's completely fix there (or maybe that was intended)

api/permissions/permissions_calculator.py
```diff
@@ -127,14 +164,67 @@ def tag_based_permissions(self) -> list[dict]:  # type: ignore[type-arg]
...
+
+        # Add role permissions
+        for role_permission in self.roles:
+            if role_permission.admin:
+                admin_permission_derived_from.roles.append(role_permission.role)
+
+            for permission_key in role_permission.permissions:
+                add_permission(permission_key, None, role_permission.role)
+
+        if self.is_organisation_admin or self.user.admin or self.admin_override:
```
Zaimwa9: This function is working well but it's also quite dense and doing a lot of things (admin resolution, formatting, permission origin, deduplication soon).

It might be worth extracting that into a dedicated PermissionAggregator-like class that could:
- abstract the different pathes
- encapsulate more the user/role/group logic
- `add_permission` being a private method 

Its goal would be to construct the graph and have a final `build` method returning the `UserDetailedPermissionsData`

api/permissions/permissions_calculator.py
```diff
@@ -70,6 +82,27 @@ class GroupPermissionData(_PermissionDataBase, _GroupPermissionBase):
     pass
 
 
+@dataclass
+class PermissionDerivedFromData:
```
Zaimwa9, line 86: One way to handle deduplication without doing a linear scan on the arrays before appending could be to internally use dicts here. Have setter methods to add a group/user and build the list with the deduped values once the permission_map is complete wdyt ?

```
    _group_map: dict[int, GroupData] # Int being the id
    _role_map: dict[int, RoleData]
    
    def build_derived_from(self) -> None:
         self.groups = list(self._group_map.values())
         self.roles = list(self._role_map.values())
```

api/permissions/permissions_calculator.py
Zaimwa9: To give some precisions on what I have in mind, it could be used as follow:

```
     aggregator= PermissionAggregator()

    for group_permission in self.groups:
        for permission_key in group_permissions.permissions:
            aggregator.add_group_permission(permission_key, group_permission.group, is_admin=group_data.admin)
            
            
        ... similar for roles and users
        
        
        return agg.build(is_admin=self.admin)
```

Zaimwa9 (changes requested):
Strong feature and improvement overall (including frontend).

Although there is a deduplication issue to solve and some suggestions on simplifying/encapsulating more the logic in `to_detailed_permissions_data` where most of the complexity lives now

api/permissions/permissions_calculator.py
gagantrivedi: Yeah, good shout! this should be done now

api/permissions/permissions_calculator.py
gagantrivedi: Yeah, that is fixed in the rbac repo

Zaimwa9 (approved):
👍 Thanks for changes on to_detailed_permissions_data (and the rbac fix), imo it became really easy to understand the logic behind.

PR #5416, opened 2025-05-02
fix: marketing consent should default to True

matthewelwell:
## Changes

As per similar functionality [here](https://github.com/Flagsmith/flagsmith/blob/d5b5e9751046a4e549a9668c6cafbcef05544f10/frontend/web/components/pages/HomePage.tsx#L55) which seemingly isn't working consistently, we now default the value of `marketing_consent_given` to `True` in the API. 

Fixes part of https://github.com/Flagsmith/flagsmith-private/issues/23

Added a new integration test.


api/tests/integration/custom_auth/end_to_end/test_custom_auth_integration.py
```diff
@@ -701,3 +702,31 @@ def test_cannot_create_superuser_if_any_user_exists(
         "A superuser can only be created through this  endpoint if no other users exist."
     ]
     assert FFAdminUser.objects.filter(email=email).exists() is False
+
+
+@pytest.mark.parametrize("marketing_consent_given", (None, True, False))
```
emyller:
```suggestion
@pytest.mark.parametrize("marketing_consent_given", [None, True, False])
```

matthewelwell: 🤔 why?

api/tests/integration/custom_auth/end_to_end/test_custom_auth_integration.py
```diff
@@ -701,3 +702,31 @@ def test_cannot_create_superuser_if_any_user_exists(
...
+def test_marketing_consent_given_defaults_to_false(
```
emyller: 🤔

api/tests/integration/custom_auth/end_to_end/test_custom_auth_integration.py
emyller: Because that is a collection of things (list) rather than a heterogenous thing (tuple).

matthewelwell: Hmm... I'm not sure how much I agree with that distinction. I'd argue that it makes sense to use a tuple here because of its immutability.

... but I don't care enough to fight it at 5pm on a Friday before going on holiday. Change inbound.

matthewelwell: Fixed in https://github.com/Flagsmith/flagsmith/pull/5416/commits/f647e1751cb7c71c44074537e56b824ff62440fc

api/tests/integration/custom_auth/end_to_end/test_custom_auth_integration.py
matthewelwell: Fixed in https://github.com/Flagsmith/flagsmith/pull/5416/commits/9da5574633b7c1f887532c9f6c64cb11a5cf70ef

khvn26 (approved):

emyller (approved):
LGTM

api/tests/integration/custom_auth/end_to_end/test_custom_auth_integration.py
emyller: Such a small thing indeed, thanks for the change.

FWIW past the low-memory-footprint argument, _immutability_ is often the reason why people choose tuples — e.g. "this will never change". Though the fact that a collection of items will [likely] not be changed in the code in the foreseeable future doesn't change the fact that it's still a _collection of **individual** items_.

To me, the difference between lists and tuples is primarily semantic. In that sense, **both the components and order define meaning**, e.g. `(True, False, None)` means _one thing_, whereas `(True, None, False)` means _another thing_. To pytest's `parametrize`, the order of each case should be irrelevant, which hints at the semantic nature of the input as a collection of individual test cases (list), rather than an individual thing (tuple).

The reason why tuples are immutable is not because we can't change them in the code; we can as needed, very similarly to how we can alter the structure of a database table. They're immutable because a tuple can only exist and make sense if it follows the defined structure for that type of object, e.g. a row in a database table. BTW you can take `namedtuple`s as an attempt to make that clearer. See, one-off tuples rarely make sense. :wink:

Another nice way of representing the semantic of tuples is their **ability to unpack**. For example, the following are most definitely tuples, as they're unpackable:

```
proto, hostname, path = ('https', 'localhost', '/foo')
first_name, last_name, birth_year = ('John', 'Doe', 2020)
```

With this, I don't mean to teach you things you already know, but I hope it's useful to any extent. Despite this being a classic interview question with a classic answer, in my experience it's also a very deep semantic issue nurtured by the community for many decades. Thankfully, harmless, in most cases.

khvn26: Aren't lists technically unpackable as well?

Not that I'd recommend unpacking collections of arbitrary length (tuples can be those too btw, consider a `tuple[Any, ...]` type), but the interpreter does support it?

FWIW, I use lists for pytest parameters for aesthetic reasons — nested square brackets are easier to distinguish inside a `pytest.mark.parametrize` function call.

emyller: True! You can most definitely unpack a list, though it will likely not be the best idea. I like using the idea of unpacking because it illustrates well the foundation of a tuple semantically speaking: its structure.

The annotation `tuple[Any, ...]` means you can have tuples of any structure _throughout_ your code. However, and for example, a given function argument typed as a tuple should never have arbitrary length or structure. That would make it a collection.

I find it easier to illustrate tuples with databases: every tuple in a table must have the same structure, declared with DDL. Every table defines the structure for their tuples. Every tuple is one individual thing in that table. That is, to my comprehension, a database table is analogous to a `list` of `tuple`s. :)

PR #5419, opened 2025-05-05
fix(e2e): Missing RBAC tests

khvn26:

## Changes

- Add missing RBAC tests
- Fix incorrect navigation to organisation settings in RBAC tests
- Allow test selection by category
- Change CI default concurrency to 16

This is a CI change.


frontend/e2e/index.cafe.js
```diff
@@ -16,19 +17,27 @@ if (fs.existsSync(dir)) {
...
-                resolve()
-            } else {
+            if (process.env.E2E_LOCAL) {
                 server = fork('./api/index');
                 server.on('message', () => {
                     resolve();
                 });
+            } else {
+                process.env.PORT = 3000;
+                resolve()
```
khvn26: Without this change, the [frontend URL](https://github.com/Flagsmith/flagsmith/blob/ad4219247153c39ed82533daf279f170bdbac145/frontend/e2e/init.cafe.js#L21-L22) was always pointing at the local testcafe-hosted frontend — which is not what we're looking for when testing Docker images built in CI.

@kyle-ssg let me know if I didn't understand it properly.

khvn26: Yeah, re-reading the Compose file, I did misunderstand it. The integration is done in build time. Reverting the change.

khvn26: Reverted the change and made the E2E Compose file use the (essentially hard-coded) 3000 port instead. It's not necessarily required to export the port anyway, but I'm doing this for consistency.

frontend/e2e/index.cafe.js
```diff
@@ -40,14 +50,14 @@ createTestCafe()
...
-                }
-            })
             .concurrency(parseInt(concurrentInstances))
+            .filter((_testName, _fixtureName, _fixturePath, testMeta, fixtureMeta) => {
+                return metaConditions.some(({ key, value }) => {
+                    return (
+                        testMeta[key] === value || fixtureMeta[key] === value
+                    );
+                });
+            })
```
Zaimwa9:
```suggestion
            .filter((_, __, ___, testMeta, fixtureMeta) =>
                 metaConditions.some(({ key, value }) =>
                      testMeta[key] === value || fixtureMeta[key] === value
                 )
           )
```

I'd just remove the returns and unused values to reduce visual noise

frontend/e2e/index.cafe.js
```diff
@@ -2,8 +2,9 @@ const createTestCafe = require('testcafe');
 const fs = require('fs');
 const path = require('path');
 const { fork } = require('child_process');
-const _options = require("../.testcaferc")
+const _options = require("../.testcaferc.js")
```
kyle-ssg: Not sure if a linter did this but .js shouldnt be needed

kyle-ssg (approved):

frontend/e2e/index.cafe.js
khvn26: Pretty sure it's me migrating to .cjs and back at some point — fixed in c4174342ece7f11b1ddeac37ba6e2011aa050bf4.

Zaimwa9 (approved):
Really good improvement, nice catch also for the roles tests. Just using the args makes it so much more understandable in terms of what belongs to oss/enterprise.

Looking around, I realized all the files in `e2e/cafe` are deprecated so I took the freedom to delete them now

PR #5420, opened 2025-05-05
chore: bump SDK major versions

rolodato:
Docs are already updated to account for the new major versions.


docs/src/components/SdkVersions.js
```diff
@@ -23,9 +23,9 @@ export const JavaVersion = ({ spec = '~7' }) => Version({ sdk: 'java', spec });
 export const AndroidVersion = ({ spec = '~1' }) => Version({ sdk: 'android', spec });
 export const CocoapodsVersion = ({ spec = '~3' }) => Version({ sdk: 'cocoapods', spec });
 export const SwiftPMVersion = ({ spec = '~3' }) => Version({ sdk: 'swiftpm', spec });
-export const DotnetVersion = ({ spec = '~7' }) => Version({ sdk: 'dotnet', spec });
+export const DotnetVersion = ({ spec = '~8' }) => Version({ sdk: 'dotnet', spec });
 export const ElixirVersion = ({ spec = '~2' }) => Version({ sdk: 'elixir', spec });
 export const RustVersion = ({ spec = '~2' }) => Version({ sdk: 'rust', spec });
 export const JsVersion = ({ spec = '~8' }) => Version({ sdk: 'js', spec });
```
Zaimwa9: We also released 9.x.x in the JS client last week. Anything to update prior to bump the version here?

Zaimwa9: if not
```suggestion
export const JsVersion = ({ spec = '~9' }) => Version({ sdk: 'js', spec });
```

Zaimwa9 (approved):
One question about JS client otherwise good to go

docs/src/components/SdkVersions.js
rolodato: Good catch - added https://github.com/Flagsmith/flagsmith/pull/5420/commits/2a2bcf14feb36e518890d0162c040880900a0fbb to remove references to `flagsmith-es`

PR #5430, opened 2025-05-07
feat: report environment metrics api

### LINKED ISSUE #3827: Epic (reporting): v1 release

There is no insight into overall analytics behind the use of feature flags within an organization or project

Set up basic metrics in the product to look at the number of features active, avg lifespan of a feature, or the amount of team members using featuring flagging. This can be expanded to include additional metrics like number of split tests, a/b tests, identity overrides active, segment overrides active, etc. 

Reporting API to give access to this outside of the UI

### Additional context

_No response_

Zaimwa9:

## Changes
⚠️  Branch with front/back: `feat/report-environment-metrics-frontend`

Dark mode with workflows
image

Light mode without workflows
image

Aim of this PR is to provide an endpoint returning a set of environment-level metrics:
- Total features
- Features enabled
- Segment override
- Open change requests
- Scheduled change to come

- Created a base metric view nested within environment view
- Added specific query-sets for each metrics
- New metric service

```
/api/v1/environments/<environment_api_key>/metrics/	
environments.views.EnvironmentMetricsViewSet	
api-v1:environments:environment-metrics-list
```
- Added environment model methods to aggregate the metrics
- Added serializers and service methods

Payload example:
```
{
    "metrics": [
        {
            "value": 3,
            "description": "Total features",
            "name": "total_features",
            "disabled": false,
            "entity": "features",
            "rank": 1
        },
        {
            "value": 2,
            "description": "Features enabled",
            "name": "enabled_features",
            "disabled": false,
            "entity": "features",
            "rank": 2
        },
        {
            "value": 0,
            "description": "Segment Overrides",
            "name": "segment_overrides",
            "disabled": false,
            "entity": "segments",
            "rank": 3
        },
        {
            "value": 1,
            "description": "Open change requests",
            "name": "open_change_requests",
            "disabled": false,
            "entity": "change_requests",
            "rank": 4
        },
        {
            "value": 1,
            "description": "Total scheduled changes",
            "name": "total_scheduled_changes",
            "disabled": false,
            "entity": "scheduled_changes",
            "rank": 5
        }
    ]
}
```

- New tests
- Via requests
```
http://localhost:8000/api/v1/environments/<EnvID>/metrics/
```

## How to
- Create segments, add segments to features
- Enable / Disable Features
- Add identity overrides to features
- Create change requests (activating 4 eyes)
- Schedule some changes via change requests

Make sure that the displayed metrics reflects reality state


api/environments/models.py
```diff
@@ -358,6 +375,187 @@ def get_feature_state(
...
+                        pk=OuterRef("change_request_id"),
+                        committed_at__isnull=True,
+                    )
+                )
+            ).filter(has_uncommitted_cr=False)
+        group_fields = ["feature_id"]
+        if extra_group_by_fields is not None:
+            group_fields.append(extra_group_by_fields)
+
+        return list(
```
gagantrivedi, line 392: Can you share the sql query and query plan generated by this

api/environments/models.py
```diff
@@ -358,6 +375,187 @@ def get_feature_state(
...
+        result: QuerySet[FeatureState] = FeatureState.objects.filter(id__in=ids)
+        return result
+
+    def _get_active_feature_states_ids(
+        self,
+        with_workflows: bool = False,
+        extra_group_by_fields: Literal["identity_id"] | None = None,
+        filter_kwargs: dict[str, typing.Any] | None = None,
+    ) -> list[int]:
+        base_qs = FeatureState.objects.filter(
```
gagantrivedi: You probably want to use `get_live_feature_states` method defined on `FeatureStateManager`

api/environments/dynamodb/wrappers/identity_wrapper.py
```diff
@@ -193,3 +193,12 @@ def get_segment_ids(
...
         return []
+
+    def get_identity_overrides_count_dynamo(self, environment_api_key: str) -> int:
+        return sum(
+            len(identity["identity_features"])
+            for identity in self.iter_all_items_paginated(
+                environment_api_key=environment_api_key,
+                overrides_only=True,
+            )
+        )
```
emyller: Is this as big and fearsome as it looks? 👀

api/environments/models.py
```diff
@@ -196,6 +206,13 @@ def __str__(self):  # type: ignore[no-untyped-def]
     def natural_key(self):  # type: ignore[no-untyped-def]
         return (self.api_key,)
 
+    @property
+    def change_requests_enabled(self) -> bool:
```
emyller: nit:

```suggestion
    def is_change_requests_enabled(self) -> bool:
```

api/environments/models.py
```diff
@@ -358,6 +375,187 @@ def get_feature_state(
             )
         )
 
+    def get_metrics_payload(self, with_workflows: bool = False) -> EnvMetricsPayload:
+        """
+        Returns total feature count and enabled-by-default feature count
+        scoped to this environment's project.
+        """
+        from edge_api.identities.models import EdgeIdentity
```
emyller: Can we import from top level?

api/environments/models.py
```diff
@@ -358,6 +375,187 @@ def get_feature_state(
...
+
+        # Closure to avoid recalculating twice the features metrics
+        def get_feature_agg(key: str) -> int:
+            nonlocal _features_aggregation_result
+            if not _features_aggregation_result:
+                _features_aggregation_result = features_qs.aggregate(
+                    total=models.Count("id"),
+                    enabled=models.Count("id", filter=Q(enabled=True)),
+                )
+            return _features_aggregation_result[key]
```
emyller: Is this a caching strategy? Can you expand on why we need it? I'd love to help flattening this into an easier to read version, if it's at all possible.

api/environments/models.py
```diff
@@ -358,6 +375,187 @@ def get_feature_state(
...
+        filter_kwargs: dict[str, typing.Any] | None = None,
+    ) -> list[int]:
+        base_qs = FeatureState.objects.filter(
+            Q(live_from__isnull=True) | Q(live_from__lte=timezone.now()),
+            environment=self,
+            **(filter_kwargs or {}),
+        )
+
+        if with_workflows:
+            from features.workflows.core.models import ChangeRequest
```
emyller: Maybe another candidate for a top-level import.

api/environments/models.py
Zaimwa9: I looked into moving it up, but there’s a circular import between the modules, so keeping it local seemed the best option

api/environments/models.py
Zaimwa9: idem 🫤

api/environments/models.py
Zaimwa9:
```
HashAggregate  (cost=10.49..10.76 rows=27 width=12)
  Output: max(features_featurestate.id), features_featurestate.feature_id, features_featurestate.identity_id
  Group Key: features_featurestate.feature_id, features_featurestate.identity_id
  ->  Hash Anti Join  (cost=1.40..10.11 rows=51 width=12)
        Output: features_featurestate.feature_id, features_featurestate.identity_id, features_featurestate.id
        Hash Cond: (features_featurestate.change_request_id = u0.id)
        ->  Seq Scan on public.features_featurestate  (cost=0.00..8.02 rows=54 width=16)
              Output: features_featurestate.id, features_featurestate.feature_id, features_featurestate.identity_id, features_featurestate.change_request_id
              Filter: ((features_featurestate.deleted_at IS NULL) AND ((features_featurestate.live_from IS NULL) OR (features_featurestate.live_from <= '2025-05-15 20:19:24.85323+00'::timestamp with time zone)) AND (features_featurestate.environment_id = 9))
        ->  Hash  (cost=1.38..1.38 rows=2 width=4)
              Output: u0.id
              ->  Seq Scan on public.workflows_core_changerequest u0  (cost=0.00..1.38 rows=2 width=4)
                    Output: u0.id
                    Filter: (u0.committed_at IS NULL)
```

My sample is relatively small but:
- No N+1 pattern identified
- grouping feature_state.id, identity_id looks correct
- we are doing a seq_scan on features_featurestate, which locally is ok, on prod it can be too heavy

It might be worth to add an index on `(environment_id, live_from, deleted_at)` wdyt ?

Zaimwa9: To go a bit further. Once completely validated, I think the metrics could be cached and invalidated on flags update etc, a bit overkill right now imo (it is controlled by a flag in the frontend for now).
Besides some pages where the code hasn't been migrated yet, it's cached with RTK and invalidated on actions related to those (not identity for now)

api/environments/dynamodb/wrappers/identity_wrapper.py
Zaimwa9: It does look a bit big and fearsome, but in practice it should be acceptable.
As DynamoDB supports "COUNT", but not nested aggregations like summing the length of a list field which we need, it didn't seem possible to have a focused aggregating query.
We're using a Query with a FilterExpression — specifically Attr("identity_features").ne([]) — which helps narrow the results. Most likely we should be charged over the number of identities within the environment.
I think the best would be to monitor and I have in mind to add a cache layer (for all the metrics actually) once validated.
If you have in mind a more optimized way to get the total, i'm happy to explore!

api/environments/models.py
Zaimwa9: Yes exactly, it was a lightweight caching strategy, using a closure to avoid evaluating the aggregate() twice.
You were right to point it, features metrics are always returned so it's a good call to favor readability here

api/environments/models.py
gagantrivedi: Can you also share the SQL query?

Zaimwa9: yes, here

```
EXPLAIN (
	ANALYZE,
	BUFFERS
)
SELECT
	MAX("features_featurestate"."id") AS "id"
FROM
	"features_featurestate"
WHERE
	(
		"features_featurestate"."deleted_at" IS NULL
		AND "features_featurestate"."deleted_at" IS NULL
		AND "features_featurestate"."environment_id" = 8
		AND (
			"features_featurestate"."environment_feature_version_id" IN (
				'36e550eb-dd3e-4923-ae00-48d910278f1a',
				'9eb662ea-0db0-46cd-871c-29356fe4b833',
				'3a83a033-afc3-4bef-9b49-343ecaf857b7'
			)
			OR "features_featurestate"."identity_id" IS NOT NULL
		)
		AND "features_featurestate"."feature_segment_id" IS NULL
		AND "features_featurestate"."identity_id" IS NULL
		AND NOT EXISTS (
			SELECT
				1 AS "a"
			FROM
				"workflows_core_changerequest" U0
			WHERE
				(
					U0."committed_at" IS NULL
					AND U0."id" = ("features_featurestate"."change_request_id")
				)
			LIMIT
				1
		)
	)
GROUP BY
	"features_featurestate"."feature_id"

```

api/environments/views.py
```diff
@@ -322,3 +326,20 @@ class EnvironmentAPIKeyViewSet(
     pagination_class = None
     permission_classes = [IsAuthenticated, EnvironmentAdminPermission]
     model_class = EnvironmentAPIKey  # type: ignore[assignment]
+
+
+class EnvironmentMetricsViewSet(NestedEnvironmentViewSet, BaseMetricsViewSet):
```
gagantrivedi: We should limit access to users with the `VIEW_ENVIRONMENT` permission.

gagantrivedi: Are we using anything from `NestedEnvironmentViewSet`?

api/metrics/views.py
```diff
@@ -0,0 +1,47 @@
...
+        """
+        Return metrics for the given object.
+        Must be implemented by subclasses.
+        """
+        pass
+
+    def get_object(self) -> Model:
+        obj_pk = self.kwargs.get(self.lookup_url_kwarg)
+        assert self.model_class is not None
+        obj = get_object_or_404(self.model_class, api_key=obj_pk)
```
gagantrivedi: Lookup should be done using pk if we want this to generic,i.e:
` obj = get_object_or_404(self.model_class, pk=obj_pk)`

api/metrics/views.py
```diff
@@ -0,0 +1,47 @@
...
+
+        self.check_object_permissions(self.request, obj)
+        return obj
+
+    def list(self, request: Request, *args: Any, **kwargs: Any) -> Response:
+        obj = self.get_object()
+        assert isinstance(obj, Environment)
```
gagantrivedi: Is `get_object` always going to return `Environment` ?

api/metrics/views.py
```diff
@@ -0,0 +1,47 @@
...
+from rest_framework.request import Request
+from rest_framework.response import Response
+from rest_framework.serializers import Serializer
+from rest_framework.viewsets import ViewSet
+
+from environments.models import Environment
+from metrics.types import EnvMetricsPayload
+
+
+class BaseMetricsViewSet(ViewSet):
```
gagantrivedi: I think this class adds more confusion. We can do without it. If we remove it, the Environment class will look something like this:
```

from rest_framework.viewsets import GenericViewSet


class EnvironmentMetricsViewSet(GenericViewSet):
    lookup_field = "api_key"
    lookup_url_kwarg = "environment_api_key"
    serializer_class = EnvironmentMetricsSerializer
    queryset = Environment.objects.all()  # type: ignore[assignment]

    def get_metrics(self, environment: Environment) -> EnvMetricsPayload:
        is_workflows_enabled = environment.is_change_requests_enabled
        return environment.get_metrics_payload(with_workflows=is_workflows_enabled)

    @swagger_auto_schema(  # type: ignore[misc]
        operation_description="Get metrics for this environment.",
        responses={200: openapi.Response("Metrics", EnvironmentMetricsSerializer)},
    )
    def list(self, request: Request, *args: Any, **kwargs: Any) -> Response:
        environment = self.get_object()
        serializer_class = self.get_serializer_class()
        serializer = serializer_class(data={"metrics": self.get_metrics(environment)})

        return Response(serializer.data, status=status.HTTP_200_OK)

```

gagantrivedi (changes requested):
I have been through the views; I still need to go through the rest of the code

api/metrics/views.py
Zaimwa9: Agreed, a second step of this feature will be project metrics, hence why I initially had a separate MetricsViewSet but no need to over-engineer until it's needed. I'll simplify with your suggestion

api/environments/views.py
```diff
@@ -358,3 +361,30 @@ class EnvironmentAPIKeyViewSet(
     pagination_class = None
     permission_classes = [IsAuthenticated, EnvironmentAdminPermission]
     model_class: Type[EnvironmentAPIKey] = EnvironmentAPIKey
+
+
+class EnvironmentMetricsViewSet(GenericViewSet[Environment]):
+    permission_classes = [IsAuthenticated]
```
gagantrivedi: We should limit this to only `VIEW_ENVIRONMENT`

api/environments/models.py
```diff
@@ -196,6 +206,13 @@ def __str__(self):  # type: ignore[no-untyped-def]
     def natural_key(self):  # type: ignore[no-untyped-def]
         return (self.api_key,)
 
+    @property
+    def is_change_requests_enabled(self) -> bool:
```
gagantrivedi: I think this is wrong... change_request can work with `minimum_change_request_approvals=0` as well.

api/environments/dynamodb/wrappers/identity_wrapper.py
```diff
@@ -193,3 +193,12 @@ def get_segment_ids(
             return [segment.id for segment in segments]
 
         return []
+
+    def get_identity_overrides_count_dynamo(self, environment_api_key: str) -> int:
```
gagantrivedi: nit: this can just be `get_identity_overrides_count`, `dyamo` part is redundant here

api/environments/models.py
gagantrivedi: All these local imports suggest this should be moved to the service layer?

api/environments/models.py
```diff
@@ -358,6 +375,179 @@ def get_feature_state(
...
+    ) -> list[int]:
+        base_qs = FeatureState.objects.get_live_feature_states(
+            environment=self,
+            **(filter_kwargs or {}),
+        )
+
+        if with_workflows:
+            from features.workflows.core.models import ChangeRequest
+
+            base_qs = base_qs.annotate(
```
gagantrivedi: Is there a reason why you have to do this? I assumed `get_live_feature_states` would take care of it?

api/environments/models.py
```diff
@@ -358,6 +375,179 @@ def get_feature_state(
...
+            {"identity__isnull": True, "feature_segment__isnull": True},
+        )
+        result: QuerySet[FeatureState] = FeatureState.objects.filter(id__in=ids)
+        return result
+
+    def _get_latest_segment_state_ids_subquery(
+        self, with_workflows: bool = False
+    ) -> list[int]:
+        segment_ids = (
+            FeatureSegment.objects.filter(environment=self)
```
gagantrivedi: Can we use `get_live_feature_states` here as well?

gagantrivedi (changes requested):

api/environments/models.py
Zaimwa9: Reviewed so it uses `get_live_feature_states` thanks. Additionally moved the logic into a service which allows avoiding the local imports (caused by circular imports)

api/environments/models.py
```diff
@@ -358,6 +367,95 @@ def get_feature_state(
...
+            feature_states_qs.values("feature_id", "feature_segment_id")
+            .annotate(latest_id=Max("id"))
+            .values_list("latest_id", flat=True)
+        )
+
+        if self.is_workflow_enabled:
+            from features.workflows.core.models import ChangeRequest
+
+            fs_base_query = fs_base_query.annotate(
+                has_uncommitted_cr=Exists(
```
gagantrivedi: I think `get_live_feature_states` takes care of this. We don't need this check

api/environments/models.py
```diff
@@ -358,6 +367,95 @@ def get_feature_state(
...
+            environment=self,
+            additional_filters=Q(
+                identity_id__isnull=True,
+                feature_segment_id__isnull=False,
+            ),
+        )
+
+        fs_base_query = (
+            feature_states_qs.values("feature_id", "feature_segment_id")
+            .annotate(latest_id=Max("id"))
```
gagantrivedi: Why are we doing Max?

gagantrivedi (changes requested):

api/environments/models.py
Zaimwa9: My bad, thanks for catching up, it's a (useless) artifact from before using get_live_feature_states

api/tests/unit/environments/test_unit_environments_models.py
```diff
@@ -1092,3 +1096,98 @@ def test_get_environment_document_from_cache_triggers_correct_metrics__cache_mis
...
+        FeatureState.objects.update_or_create(
+            feature=features[i],
+            environment=env,
+            identity=None,
+            enabled=True,
+            version=version,
+        )
+        version += 1
+
+    for i in range(segment_overrides_count):
```
Zaimwa9, line 1156: Add versions to segments to check version selection

gagantrivedi (approved):

PR #5449, opened 2025-05-14
fix: identities endpoint does not respect deleted segment overrides in versioned environments

### LINKED ISSUE #5448: Segment overrides removed in versioned environments are still returned on core `/identities` endpoint
Steps to reproduce: 

 1. Given a feature enabled via a segment override
 2. Update the feature to remove the segment override

Expected behaviour: 

The value returned via the `/identities` endpoint should be the environment default

Actual behaviour: 

The value returned via the `/identities` endpoint is the now removed segment override

matthewelwell:

## Changes

This PR resolves #5448. It does so by using a consistent approach for retrieving the latest environment flags which handles the versioning behaviour. 

Added an integration test to cover the specific use case as per #5448 .


api/tests/integration/features/versioning/test_integration_v2_versioning.py
```diff
@@ -612,3 +612,55 @@ def test_v2_versioning_carries_existing_segment_overrides_across(
...
+        url,
+        data=json.dumps(data),
+        content_type="application/json",
+    )
+
+    # Then
+    assert response.status_code == status.HTTP_201_CREATED
+
+    # now let's retrieve the flags for an identity that matches the segment
+    # and verify that they are not receiving the
```
Zaimwa9:
```suggestion
```
I think we can remove the comments

api/tests/integration/features/versioning/test_integration_v2_versioning.py
```diff
@@ -612,3 +612,55 @@ def test_v2_versioning_carries_existing_segment_overrides_across(
...
+    identity_flags_response_json = get_identity_flags_response_json(
+        num_expected_flags=1
+    )
+    assert (
+        identity_flags_response_json["flags"][0]["feature_state_value"]
+        == default_feature_value
+    )
+
+    # let's also check the environment document response to make sure that
+    # the segment override doesn't exist there
```
Zaimwa9:
```suggestion
```

api/tests/integration/features/versioning/test_integration_v2_versioning.py
```diff
@@ -612,3 +612,55 @@ def test_v2_versioning_carries_existing_segment_overrides_across(
         feature_segment_list_post_migrate_response_json["results"][0]["id"]
         == feature_segment
     )
+
+
+def test_deleting_segment_overrides(
```
Zaimwa9:
```suggestion
def test_identities_should_return_default_environment_values_after_deleting_segment_override(
```

Zaimwa9 (approved):
Tested against main, confirming the initial bug and its resolution 👍 .
Minor renaming and optional removing comments if you agree the test is self-explanatory

Zaimwa9 (approved):
All good!

PR #5460, opened 2025-05-16
docs: imported cards component, revised changes

lucaspr2:
Updated page with reviewed changes:
table, cards and CTA's to improve visual structure.

### DIFF
--- a/docs/docs/version-comparison.md
+++ b/docs/docs/version-comparison.md
@@ -2,59 +2,126 @@
 sidebar_position: 3
 ---
 
+import Card from '@site/src/components/Card';
+import CardHeader from '@site/src/components/Card/CardHeader';
+import CardBody from '@site/src/components/Card/CardBody';
+import CardFooter from '@site/src/components/Card/CardFooter';
+
 # Version Comparison
 
-You are free to run the Open Source version of Flagsmith however you see fit! There are some differences between the
-Open Source, SaaS hosted and Enterprise versions:
+Choose the right Flagsmith version for your needs. Compare our Open Source, SaaS, and Enterprise offerings:
 
-- The Open Source version has **no** API request or Identity limits - you can run as many API instances in a cluster as
-  you wish.
-- The Open Source version has **no** Dashboard User limits.
-- The Open Source version has **no** Environment limits.
-- The Open Source version is limited to a single Project.
-- The SaaS and Enterprise versions have [Change Requests and Flag Scheduling](advanced-use/change-requests.md).
-- The SaaS and Enterprise versions have [Role-Based Access Control](/system-administration/rbac).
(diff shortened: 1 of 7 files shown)


docs/docs/version-comparison.md
```diff
@@ -2,59 +2,126 @@
...
+import CardBody from '@site/src/components/Card/CardBody';
+import CardFooter from '@site/src/components/Card/CardFooter';
 
-:::
-
-## SaaS Benefits
+# Version Comparison
 
-Our SaaS platform has a number of benefits:
+Choose the right Flagsmith version for your needs. Compare our Open Source, SaaS, and Enterprise offerings:
```
rolodato, line 12: The original document was all over the place and difficult to follow, but I still consider this new version to be a downgrade overall.

There are 2 dimensions that users need to choose when deciding to use Flagsmith:

1. Self-hosted or SaaS
2. Free or paid plan

These are separate from each other - you can use SaaS or self-host for free (open source version). You can use pay for self-hosting on the Enterprise plan, or pay Flagsmith to host it for you. I feel this document is now a copy of https://www.flagsmith.com/pricing and we don't gain anything by having it.

docs/docs/version-comparison.md
```diff
@@ -2,59 +2,126 @@
...
+  </div>
+</div>
 
-- Completely Free!
-- The Open Source version has **no** API request or Identity limits - you can run as many API instances in a cluster as
-  you wish.
-- The Open Source version has **no** Dashboard User limits - you can have as many team members as you wish.
-- Deploy with one click to a number of different [IaaS and PaaS providers](/deployment#one-click-installers).
+:::tip Migration
+You can easily switch between SaaS and Self Hosted Flagsmith using our [Import and Export tools](system-administration/importing-and-exporting/organisations).
```
rolodato, line 126: I would leave out "easily" here. There is a decent amount of manual work involved in a migration, and we don't want to encourage customers to depend on this. Migration is more like an escape hatch that's available if a customer ever decides they need to switch.

docs/docs/version-comparison.md
```diff
@@ -2,59 +2,126 @@
...
+| **Infrastructure** |
+| Self-hosted | ✅ | ❌ | Optional |
+| Cloud Hosted | ❌ | ✅ | Optional |
+| Private Cloud | ❌ | ❌ | ✅ |
+| Edge API | ❌ | ✅ | Optional |
+| Database Options | PostgreSQL | PostgreSQL | PostgreSQL, Oracle, MySQL |
 
-You can run our Enterprise version either on-premise, or we can provide private cloud instance dedicated to your
-organisation.
+## Deployment Options
```
rolodato, line 53: What's being described here are SaaS plans, not deployment options (same as https://www.flagsmith.com/pricing). I would also prefer to avoid mentioning any specific pricing plans and limits in the documentation.

PR #5461, opened 2025-05-16
docs: Basic features overview

lucaspr2:
Updated page with reviewed changes.

### DIFF
--- a/docs/docs/basic-features/index.md
+++ b/docs/docs/basic-features/index.md
@@ -3,82 +3,125 @@ title: Feature Flags - An Overview
 sidebar_position: 1
 ---
 
-Feature Flags are a development methodology that allow you to ship code and features before they are finished. This
-greatly benefits Continuous Integration and Continuous Deployment (CI/CD). The typical workflow for this is as follows.
+import Card from '@site/src/components/Card';
+import CardHeader from '@site/src/components/Card/CardHeader';
+import CardBody from '@site/src/components/Card/CardBody';
 
-1. You are about to start work on a new feature. Lets imaging you are going to implement a sharing button with your
-   application.
-2. Create a new Feature Flag in Flagsmith, calling it "sharing_button". Set it to enabled on your development
-   environment, and disabled on your production environment.
-3. Start working on the feature. Whenever you write code that shows the button within the UI, wrap it in a conditional
-   statement, testing against the value of the flag "sharing button". Only show the button if the flag is set to True.
-4. Because your button only shows when the "sharing_button" flag is set to True, you are safe to commit your code as you
-   work on the feature. Your code will be live within the production platform, but the functionality is hidden behind
-   the flag.
-5. Once you are happy with your Feature, you can enable the "sharing_button" for other members of your team and with
-   Beta testers.
-6. If everything is working as intended, flip the "sharing_button" flag to True for everyone in your production
-   environment, and your feature is rolled out.
+# Feature Flags - An Overview
 
-If you want to learn more about Feature Flags,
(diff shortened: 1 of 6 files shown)


docs/docs/basic-features/index.md
```diff
@@ -3,82 +3,125 @@ title: Feature Flags - An Overview
...
-
-Similarly to individual users, you will be able to override environment defaults for features. For example showing
-certain features for a "power user" segment.
-
-For more info see [Segments](/basic-features/segments.md).
+import Card from '@site/src/components/Card';
+import CardHeader from '@site/src/components/Card/CardHeader';
+import CardBody from '@site/src/components/Card/CardBody';
+
+# Feature Flags - An Overview
```
rolodato, line 10: I'd like to define what the correct scope for this document should be. I think we are mixing two things that maybe should not belong in the same document:

* Conceptual introduction to feature flags - this is not specific to Flagsmith, just the concept of feature flags. The idea would be to introduce feature flags and explain their pros/cons, and maybe point users to how they can achieve different use cases using Flagsmith.
* Flagsmith data model overview - introduce concepts like organisations/projects/environments/etc and explain how they fit together. It's confusing that we've originally called this document "Basic features" when "Feature" is already a Flagsmith-specific term.

Right now it seems as if this document doesn't have a clear focus on either of these, and they likely shouldn't belong together.

docs/docs/basic-features/index.md
```diff
@@ -3,82 +3,125 @@ title: Feature Flags - An Overview
...
+      </CardBody>
+    </Card>
+  </div>
+</div>
+
+## Core Concepts
+
+```mermaid
+graph TD
+    A[Organization] --> B[Projects]
```
rolodato, line 54: We want to use UK English for all documentation, for example `Organisation` instead of `Organization`.

docs/docs/basic-features/index.md
```diff
@@ -3,82 +3,125 @@ title: Feature Flags - An Overview
...
+# Feature Flags - An Overview
+
+
+Feature flags enable fine-grained control over your application's functionality. They allow you to:
+
+<div className="row">
+  <div className="col col--4">
+    <Card>
+      <CardHeader>
+        <h3>🚀 Deploy Safely</h3>
```
rolodato, line 19: This should be "Release Safely" instead. The idea is that feature flags decouple releases from deployment, so the deployment process is less important/risky when feature flags are being used.

docs/docs/basic-features/index.md
```diff
@@ -3,82 +3,125 @@ title: Feature Flags - An Overview
...
+      <CardBody>
+        Enable features for specific users, segments, or environments to gather real-world feedback.
+      </CardBody>
+    </Card>
+  </div>
+  
+  <div className="col col--4">
+    <Card>
+      <CardHeader>
+        <h3>⚡ Move Fast</h3>
```
rolodato, line 41: Is this meaningfully different from the first card?

docs/docs/basic-features/index.md
```diff
@@ -3,82 +3,125 @@ title: Feature Flags - An Overview
...
+    G --> H[Traits]
+    G --> I[Segments]
+```
+
+### Organizations and Projects
+Organizations are containers for your projects and team members. Projects help you organize related features and environments.
+
+```mermaid
+graph LR
+    A[Organization] --> B[Team Management]
```
rolodato, line 69: I don't think we've ever used the concept "Team Management" in the documentation. Does this add anything, and if not could we remove it?

Issue #2064, opened 2023-03-22
Webhooks are not triggered when a Change Request is committed


matthewelwell:
Currently webhooks are not triggered when a change request is committed. Change requests that are not scheduled for the future should trigger the feature webhooks. 

To reproduce:

1. Configure an environment with webhooks
2. Create a change request
3. Approve and commit the change request
4. No webhook is triggered when the change request is committed

Expected behaviour:

A webhook event is triggered when the change request is committed. 

## Acceptance Criteria

 - [ ] When a change request is committed, a webhook event is sent to any 'Feature webhooks' configured in the given environment
 - [ ] The webhook contains all the same information as if the feature was changed without a change request

Braedon-Wooding-Displayr:
We are also facing this issue.  Is there an ETA for a fix or a workaround?

matthewelwell:
Hi @Braedon-Wooding-Displayr , we don't have an ETA on this one at the moment, but the work around here would likely be to use the organisation (audit log) webhooks instead.

Braedon-Wooding-Displayr:
Thanks!  That's what we ended up using.

Zaimwa9:
Follow-up discussion about "refactor feature change audit log into a proper audit log based integration" feasibility in #5790

PR #5464, opened 2025-05-19
feat: trigger-webhook-events-on-commiting-change-requests

Zaimwa9:

## Changes
Solves #2064 

Webhook events for FLAG_UPDATED were not triggered when committing a Change Request. This PR resolves that issue.

There are two relevant cases:
**Feature versioning v2**:
Webhooks were intentionally skipped when environment_feature_version_id is present (cf in `test_webhooks_are_not_called_for_feature_state_with_environment_feature_version`)

**Uncommitted Change Requests** (this PR's focus):
When creating a Change Request, associated FeatureState objects are created immediately. At this point, `post_save` triggers (now suppressed) were sending a half empty event while the state was not live. However, when the Change Request is committed, the associated feature states are bulk updated—bypassing signals—and no webhook is triggered.

I intentionally avoided relying on the post_save signal during commit because Change Request FeatureStates are new records without version when initially created.
Their history reflect the draft state, meaning enabled matches the future committed values. This leads to new_state == old_state in the webhook payload. So resolving FeatureState based on the previous live version ensure that we pass in the correct diff

Fix:
On commit, we now fetch the latest live FeatureState (from DB, not history) and explicitly call trigger_feature_state_change_webhooks() with both the new and previous live state

- Added tests
- Built a local webhook server


api/features/signals.py
```diff
@@ -12,6 +12,14 @@
 
 @receiver(post_save, sender=FeatureState)
 def trigger_feature_state_change_webhooks_signal(instance, **kwargs):  # type: ignore[no-untyped-def]
-    if instance.environment_feature_version_id or instance.deleted_at:
+    is_skip_change_request_created = (
```
Zaimwa9, line 17: When we create a ChangeRequest, we create a new FeatureState and it triggers a webhook with a payload as below. Except if there is a reason for this, I believe we should either create a new webhook event (e.g `CHANGE_REQUEST_CREATED`) or skip it
```
  data: {
    changed_by: 'xxxxxx@gmail.com',
    new_state: {
      enabled: true,
      environment: [Object],
      feature: [Object],
      feature_segment: null,
      feature_state_value: null,
      identity: null,
      identity_identifier: null
    },
    timestamp: '2025-05-19T07:41:38.471713Z'
  },
  event_type: 'FLAG_UPDATED'
}
```

frontend/web/components/PermissionControl.tsx
```diff
@@ -129,7 +126,10 @@ const PermissionControl: React.FC<PermissionControlProps> = ({
 
   const isViewRequiredAndNotAllowed =
     isViewPermissionRequired && !isViewPermissionAllowed
-  const isChecked = !disabled && isPermissionEnabled && (!isViewPermissionRequired || isViewPermissionAllowed)
+  const isChecked =
```
Zaimwa9: pre-commit artifact

api/core/workflows_services.py
```diff
@@ -25,15 +35,60 @@ def commit(self, committed_by: "FFAdminUser") -> None:
...
             )
 
+        logger.debug("Committing change request #%d", self.change_request.id)
+
+        feature_states_to_notify = self._get_committed_feature_states_diff()
         self._publish_feature_states()
         self._publish_environment_feature_versions(committed_by)
         self._publish_change_sets(committed_by)
         self._publish_segments()
+        self._notify_committed_changes_webhooks(feature_states_to_notify)
```
emyller, line 45: Similarly to what I'm working on right now in the Sentry integration, I'm concerned about what happens when the change request commits a flag change that will only go live in the future.

The approach I'm taking involves binding events to the "went live" audit log, which was [recently fixed](https://github.com/Flagsmith/flagsmith/issues/5030) to match the actual flag change effect.

Can you please investigate if that's going to be necessary in this context as well?

emyller: The signal `feature_state_change_went_live` introduced in [this commit](https://github.com/Flagsmith/flagsmith/pull/5531/commits/f99d25300b4276d2894785bcc7eece0b14582960) might help if the above proves necessary.

Zaimwa9: Good question.
First: It's triggering a webhook for creating a scheduled change requests => not good, will be changed.

It could make sense, but I could use a bit more of context. As far as I can see, the `feature_state_change_went_live` is tightly coupled with AuditLog, that is a different type of webhook event and only sends organisation webhooks.

For example `trigger_feature_state_change_webhooks` is in charge of formatting payloads and sending to both env/org webhooks with `WebhookEventType.FLAG_UPDATED`. My fear is that by coupling AuditLog and CR Publishing here we re-execute some paths.

I might be overlooking what you had in mind so i'd love some details

emyller: I think it depends on what is the primary event we're dealing with, i.e. to what sending this webhook should trigger to:

1. The commit of the change request, or
2. The change in the associated feature flag.

If the answer points to (2), then the new `feature_state_change_went_live ` signal I mentioned above might be useful. Yes it creates coupling with audit logs, which I also tried avoiding at first, before I learned that auditing is actually entrypoint to many integrations. cc @khvn26

Zaimwa9:
Closing to be replaced with #5790

Issue #5472, opened 2025-05-20
Usage billing periods default to incorrect range


Zaimwa9:
### How are you running Flagsmith

- [x] Self Hosted with Docker
- [x] Self Hosted with Kubernetes
- [x] SaaS at flagsmith.com
- [ ] Some other way (add details in description below)

### Describe the bug

Without a valid subscription start/end date, the usage for:
- the current period is empty (defaults to today)
- the previous period returns the last 30 days

### Steps To Reproduce

1. Have an organisation without subscription
2. Go to usage panel
3. Check current/previous period

### Expected behavior

Current period should be returning usage for period start date -> today
Previous period should be returning last period start date -> end last period start date

### Screenshots

Taken on the 20th May

<img width="974" alt="Image" src="https://github.com/user-attachments/assets/0837e0e0-5ab9-48e7-8e2b-68cb259fad2e" />

<img width="867" alt="Image" src="https://github.com/user-attachments/assets/23a01860-731c-408c-94c6-62ab812d21ee" />

rolodato:
Several ways to consider improving this:

* If there is no subscription period set (most common case for self-hosted setups), do not show any options to display usage periods. I had also considered using the 1st day of the month as a default, but I'd prefer to be safe and not mention anything about subscriptions unless we have the correct data in place.
* Default to showing usage in the past 30 days.

Zaimwa9:
@rolodato, I was working on a fix before your commented. In it, I proposed as defaults (if no subscription):
- current_period: 1st of the month -> today 
- previous_period 1st of last month -> end of last month

To align with monthly usage. If you prefer one of your proposition, let me know i'll adapt

rolodato:
I would still prefer to never assume anything about the billing period. We should either:

* Hide all options about the billing period if we don't know what the billing period is
* Show that there is no data because there is no billing period set, rather than just showing no data (less ideal, because why show that option at all to begin with?)

Billing is a sensitive topic and we can't assume fallbacks for this type of data - we have to explicitly hide it or show there's nothing and explain why.

Zaimwa9:
I get your point. I'll rework it so that we surface wether we have a billing period. If not, we can only filter by 30/90 last days

Zaimwa9:
Solved by #5479

PR #5474, opened 2025-05-21
fix: incorrect default usage period fallback

Zaimwa9:

## Changes
Related #5472 
Pre-requisite to #5479

🔴  Problem
Usage period calculation fallback did not account for the case where a subscription_cache exists but billing_periods dates are not set, defaulting to current_period range filter being similar to `where date < today and date > today ` and previous_period to `where date < today and date > today -30days`.
Tests were passing because assuming the cache was empty

🟢 Proposed solution
> I would still prefer to never assume anything about the billing period. We should either:
> 
> Hide all options about the billing period if we don't know what the billing period is
> Show that there is no data because there is no billing period set, rather than just showing no data (less ideal, because why show that option at all to begin with?)
> Billing is a sensitive topic and we can't assume fallbacks for this type of data - we have to explicitly hide it or show there's nothing and explain why.

- Added new organisation.Subscription property `has_billing_periods`
- Added new `OrganisationSubscriptionInformationCache` method `has_billing_periods` returning true if now between `current_billing_term_starts_at` and `current_billing_terms_ends_at`

- Added test
Manually:
Verify `GET organisation` payload 
```
 "subscription": {
                "id": 2,
                "has_billing_periods": true,
                "deleted_at": null,
                "uuid": "xx",
                "subscription_id": null,
                "subscription_date": null,
                "plan": "enterprise",
                "max_seats": 4,
                "max_api_calls": 50000,
                "cancellation_date": null,
                "customer_id": null,
                "billing_status": null,
                "payment_method": null,
                "notes": null
            },
```

### DIFF
--- a/api/organisations/models.py
+++ b/api/organisations/models.py
@@ -265,10 +265,10 @@ class Subscription(LifecycleModelMixin, SoftDeleteExportableModel):  # type: ign
         )
 
     @property
-    def has_billing_periods(self) -> bool:
+    def has_active_billing_periods(self) -> bool:
         return (
             self.organisation.has_subscription_information_cache()
-            and self.organisation.subscription_information_cache.has_billing_periods()
+            and self.organisation.subscription_information_cache.has_active_billing_periods()
         )
 
     @property
@@ -553,19 +553,19 @@ class OrganisationSubscriptionInformationCache(LifecycleModelMixin, models.Model
             "feature_history_visibility_days": self.feature_history_visibility_days,
         }
 
-    def has_billing_periods(self) -> bool:
+    def is_billing_terms_dates_set(self) -> bool:
+        return (
+            self.current_billing_term_starts_at is not None
+            and self.current_billing_term_ends_at is not None
+        )
+
+    def has_active_billing_periods(self) -> bool:
         """
         Returns True if current date is within the billing term.
         If either start or end date is None, returns False.
         """
-        if (
-            self.current_billing_term_starts_at is None
-            or self.current_billing_term_ends_at is None
-        ):
-            return False
-
         now = timezone.now()
-        return (
+        return self.is_billing_terms_dates_set() and (
             self.current_billing_term_starts_at
             <= now
             <= self.current_billing_term_ends_at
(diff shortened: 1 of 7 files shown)


api/tests/unit/organisations/test_unit_organisations_views.py
```diff
@@ -219,6 +219,38 @@ def test_should_update_organisation_data(
     assert organisation.restrict_project_create_to_admin is True
 
 
+@pytest.mark.parametrize(
+    "current_billing_term_starts_at, current_billing_term_ends_at, has_billing_periods",
+    [
+        (timezone.now() - timedelta(days=5), timezone.now() + timedelta(days=5), True),
+        (timezone.now() - timedelta(days=5), timezone.now() - timedelta(days=1), False),
```
khvn26, line 227: Can we add a `(None, None, False)` case here?

api/organisations/models.py
```diff
@@ -545,6 +553,24 @@ def _get_default_subscription_metadata_kwargs(self) -> dict[str, Any]:
...
+            or self.current_billing_term_ends_at is None
+        ):
+            return False
+
+        now = timezone.now()
+        return (
+            self.current_billing_term_starts_at
+            <= now
+            <= self.current_billing_term_ends_at
+        )
```
khvn26: This method does too much right now, IMO? Should it be called `has_active_billing_periods` instead?

khvn26 (changes requested):

api/organisations/models.py
```diff
@@ -545,6 +553,28 @@ def _get_default_subscription_metadata_kwargs(self) -> dict[str, Any]:
...
+
+    def has_active_billing_periods(self) -> bool:
+        """
+        Returns True if current date is within the billing term.
+        If either start or end date is None, returns False.
+        """
+        if not self.is_billing_terms_dates_set():
+            return False
+        assert self.current_billing_term_starts_at is not None
+        assert self.current_billing_term_ends_at is not None
```
khvn26: I understand this is needed for mypy to pass, so let's make it self-evident:

```suggestion
    if TYPE_CHECKING:
        assert self.current_billing_term_starts_at is not None
        assert self.current_billing_term_ends_at is not None
```

khvn26 (approved):
LGTM with 1 minor comment.

PR #5475, opened 2025-05-21
fix: Exclude identities when viewing change request

kyle-ssg:

## Changes

Change requests were incorrectly including identity feature states when diffing the states. The following (live) is incorrect. The DiffChangeRequest component now takes an optional identity to diff against.

image

vs

image

As above


Zaimwa9 (approved):
👍

PR #5476, opened 2025-05-21
fix: handle TypeError exposed by simple history logic

matthewelwell:

## Changes

Fixes [this Sentry issue](https://flagsmith.sentry.io/issues/6434075641). 

Existing tests cover the lack of regression, this change is purely to handle a scenario that is not possible to rerproduce but has been seen in production as per linked sentry issue.

### DIFF
--- a/api/core/models.py
+++ b/api/core/models.py
@@ -83,8 +83,11 @@ class _BaseHistoricalModel(models.Model):
     class Meta:
         abstract = True
 
-    def get_change_details(self) -> typing.Optional[typing.List[ModelChange]]:  # type: ignore[return]
-        if self.history_type == "~":  # type: ignore[attr-defined]
+    def get_change_details(self) -> typing.List[ModelChange]:
+        # Note that self.prev_record should never be None when self.history_type == "~" but we have
+        # seen rare cases in Sentry which cause an unhandled exception, so we instead handle the case
+        # and fall back to just return an empty list.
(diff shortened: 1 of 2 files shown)


api/core/models.py
```diff
@@ -83,8 +83,11 @@ class _BaseHistoricalModel(models.Model):
     class Meta:
         abstract = True
 
-    def get_change_details(self) -> typing.Optional[typing.List[ModelChange]]:  # type: ignore[return]
-        if self.history_type == "~":  # type: ignore[attr-defined]
+    def get_change_details(self) -> typing.List[ModelChange]:
```
Zaimwa9, line 86: Just in case we could add a test to ensure it's covered
```
def test_get_change_details_handles_exception_prev_record_is_null():
    instance = HistoricalFeatureState(history_type="~")
    instance.prev_record = None 

    assert instance.get_change_details() == []
```
Ideally we could log when it happens but I let you decide if we will really track it or covering the case is enough

Zaimwa9 (approved):
Approved with minor comment

api/core/models.py
matthewelwell: Unfortunately it's not that easy. `prev_record` is a property, not an attribute, so I think any test that we add would involve more mocking than is really worthwhile. 

Regarding the logging, since we've only seen this on a handful of cases, and the use case for this data is pretty slim, I don't think it's worth investigating further, hence negating the need for any additional logging. 

I'm going to go ahead and merge this, but let me know if you disagree with either of my points above.

PR #5479, opened 2025-05-22
fix: hide-billing-periods-if-subscription-has-no-periods

Zaimwa9:
Thanks for submitting a PR! Please check the boxes below:

- [ ] I have added information to `docs/` if required so people know about the feature!
- [X] I have filled in the "Changes" section below?
- [X] I have filled in the "How did you test this code" section below?
- [X] I have used a [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) title for this Pull Request

## Changes
Related #5472 and dependent on #5474 

- Uses `has_billing_periods` to hide `Current Billing Period` and `Previous Billing Period` in Usage page
- Fallback showing a message when no period sets (but shouldn't happen)
- Defaults to free plan views

## How did you test this code?
image

image

PR #5480, opened 2025-05-22
fix: raise-404-if-org-does-not-have-billing-periods

Zaimwa9:

## Changes
Optional depending on #5479 

Once frontend has been released, cleaner to return 404 if organisation does not have billing periods as we do not want to expose it.

- added tests


api/app_analytics/analytics_db_service.py
```diff
@@ -36,35 +37,35 @@ def get_usage_data(
...
+    sub_cache = getattr(organisation, "subscription_information_cache", None)
+    is_subscription_valid = (
+        sub_cache is not None and sub_cache.current_billing_term_starts_at is not None
+    )
 
     match period:
         case constants.CURRENT_BILLING_PERIOD:
-            if not getattr(organisation, "subscription_information_cache", None):
-                starts_at = now - timedelta(days=30)
+            if not is_subscription_valid or sub_cache is None:
```
emyller: Given the above edit, it looks like the second proposition is [redundant](https://en.wikipedia.org/wiki/De_Morgan%27s_laws).

```suggestion
            if not is_subscription_valid:
```

api/app_analytics/analytics_db_service.py
```diff
@@ -36,35 +37,35 @@ def get_usage_data(
...
-                sub_cache = organisation.subscription_information_cache
                 starts_at = sub_cache.current_billing_term_starts_at or now - timedelta(
                     days=30
                 )
-            month_delta = relativedelta(now, starts_at).months
-            date_start = relativedelta(months=month_delta) + starts_at
-            date_stop = now
+                month_delta = relativedelta(now, starts_at).months
+                date_start = relativedelta(months=month_delta) + starts_at
+                date_stop = now
```
emyller: Given the positive `if` outcome will result in a raise, I think we'd be better off without the `else` clause entirely.

api/app_analytics/analytics_db_service.py
```diff
@@ -36,35 +37,35 @@ def get_usage_data(
...
                 )
-            month_delta = relativedelta(now, starts_at).months - 1
-            month_delta += relativedelta(now, starts_at).years * 12
-            date_start = relativedelta(months=month_delta) + starts_at
-            date_stop = relativedelta(months=month_delta + 1) + starts_at
-
+                month_delta = relativedelta(now, starts_at).months - 1
+                month_delta += relativedelta(now, starts_at).years * 12
+                date_start = relativedelta(months=month_delta) + starts_at
+                date_stop = relativedelta(months=month_delta + 1) + starts_at
```
emyller: Same as above.

api/tests/unit/app_analytics/test_analytics_db_service.py
```diff
@@ -1,10 +1,11 @@
-from datetime import date, datetime, timedelta
+from datetime import date, datetime, timedelta, timezone
 
 import pytest
 from django.conf import settings
-from django.utils import timezone
+from django.utils import timezone as d_timezone
```
emyller:
```suggestion
from django.utils import timezone
```

In order to keep the diff smaller, and the code consistent, I'd keep this as it was. Above, maybe a better approach is:

```py
from datetime import UTC
```

emyller (changes requested):

api/app_analytics/analytics_db_service.py
Zaimwa9: mypy not inferring from `is_subscription_valid` that sub_cache is none :(

api/tests/unit/app_analytics/test_analytics_db_service.py
```diff
@@ -386,30 +387,27 @@ def test_get_feature_evaluation_data_calls_get_feature_evaluation_data_from_loca
...
     # Then
-    mocked_get_usage_data_from_local_db.assert_called_once_with(
-        organisation=organisation,
-        environment_id=None,
-        project_id=None,
-        date_start=datetime(2022, 12, 20, 9, 9, 47, 325132, tzinfo=timezone.utc),  # type: ignore[attr-defined]
-        date_stop=datetime(2023, 1, 19, 9, 9, 47, 325132, tzinfo=timezone.utc),  # type: ignore[attr-defined]
-    )
+    except NotFound as e:
+        assert "No billing periods found for this organisation." in str(e)
```
emyller: This does not verify that an exception was raised. Please use `pytest.raises`.

api/app_analytics/analytics_db_service.py
emyller: Painful. :(
I wonder if we should use a comment to ignore mypy's complaint, though I fear we haven't decided yet which path is best.

In hope that someday mypy/Python will be smarter at typing, I'd use comments because those are easier to find and delete.

matthewelwell: I also lean towards the `type: ignore` comment.

emyller (changes requested):
LaGTM

api/app_analytics/analytics_db_service.py
khvn26: Slightly more verbose but IMHO better for type inference:

```suggestion
            if not is_subscription_valid:
                if typing.TYPE_CHECKING:
                    assert sub_cache is None
```

Zaimwa9: Deep down i'd side with Kim.
In-between solution with an early return to also avoid repetition?
Otherwise in that specific case i'd go with `type-ignore` too as it's complaining on the 2 `starts_at`

```
    is_subscription_valid = (
        sub_cache is not None and sub_cache.is_billing_terms_dates_set()
    )

    if period in (constants.CURRENT_BILLING_PERIOD, constants.PREVIOUS_BILLING_PERIOD):
        if not is_subscription_valid:
            raise NotFound("No billing periods found for this organisation.")

    if TYPE_CHECKING:
        assert sub_cache is not None

    match period:
        case constants.CURRENT_BILLING_PERIOD:
            starts_at = sub_cache.current_billing_term_starts_at or now - timedelta(
                days=30
            )
            ...
 ```

Comment (emyller) [reply api/app_analytics/analytics_db_service.py]: > In hope that someday mypy/Python will be smarter at typing, I'd use comments because those are easier to find and delete.

Given my own argument above, I agree `if TYPE_CHECKING` also achieves making it easier to flag code we can hopefully delete eventually, while maybe being useful in preventing type errors until then. Thanks!

File: api/app_analytics/analytics_db_service.py [hunk 6]
```diff
@@ -36,35 +37,37 @@ def get_usage_data(
...
     now = timezone.now()
-
     date_start = date_stop = None
+    sub_cache = getattr(organisation, "subscription_information_cache", None)
+
+    is_subscription_valid = (
+        sub_cache is not None and sub_cache.is_billing_terms_dates_set()
+    )
+
+    if period in (constants.CURRENT_BILLING_PERIOD, constants.PREVIOUS_BILLING_PERIOD):
```
emyller, line 47: nit:

```suggestion
    if period in {constants.CURRENT_BILLING_PERIOD, constants.PREVIOUS_BILLING_PERIOD}:
```

- Semantially this isn't a tuple, and
- In sets, `in` takes `O(1)` time — not relevant in this line but might be worth noting.

emyller (approved):
LGTM

PR #5483, opened 2025-05-22
feat: Added homepage

lucaspr2:

## Changes

Added homepage based on figma design

Tested in light/dark mode, mobile and desktop res

_Please describe._


docs/src/components/Card/CardBody/index.js
```diff
@@ -0,0 +1,44 @@
+import React,  { CSSProperties } from 'react';
+import clsx from 'clsx';
+const CardBody = ({
+  className, // classNamees for the container card
+  style, // Custom styles for the container card
+  children, // Content to be included within the card
```
tiagoapolo, line 6: A suggestions is to use JSDoc instead of comments so we can provide doc information to the  IDE 

```suggestion
/**
 * CardBody
 *
 * @param {object} props
 * @param {string} props.shopName The name of the shop
 * @returns {JSX.Element}
 */

const CardBody = ({
  className, 
  style, 
  children,
 
```

docs/src/components/Card/CardHeader/index.js
```diff
@@ -0,0 +1,16 @@
+import React, { CSSProperties } from 'react'; // CSSProperties allows inline styling with better type checking.
+import clsx from 'clsx'; // clsx helps manage conditional className names in a clean and concise manner.
+const Card = ({
+  className, // Custom classes for the container card
+  style, // Custom styles for the container card
+  children, // Content to be included within the card
+  shadow, // Used to add shadow under your card. Expected values are: low (lw), medium (md), tall (tl)
+}) => {
```
tiagoapolo, line 8: same here, JSDoc would be a nice addition

PR #5484, opened 2025-05-23
fix: Improve audit logs when postponing a scheduled Change Request

### LINKED ISSUE #5030: Editing a scheduled change's go-live date still causes tasks to execute at the original date

emyller:
Closes #5030.

---

## Changes

- [x] Fix "went live" audit log when a scheduled change request is **rescheduled**.
- [x] Fix "flag state updated" audit log when the update isn't live yet.

Unit, integration, and manual tests.

## Screenshots

"Flag state updated" audit log (when scheduled):
image


api/audit/constants.py
```diff
@@ -12,6 +12,7 @@
 FEATURE_STATE_SCHEDULED_MESSAGE = (
     "Flag state / Remote Config value update scheduled for %s for feature: %s"
 )
+FEATURE_STATE_WILL_BE_UPDATED_MESSAGE = "Flag state for feature '%s' will be updated at %s UTC unless the Change Request '%s' is rescheduled or deleted."
```
khvn26: 1. IMO, the "unless rescheduled or deleted" clarification is redundant here. The user should see the full picture by looking at all the available audit logs. If we want a more helpful message, consider adding heuristics to determine whether the change was initially scheduled by a new change request, or rescheduled for an existing one.   

2. I'm not a fan of future tense in log messages. Consider improving the wording to something among the lines of

```suggestion
FEATURE_STATE_WILL_BE_UPDATED_MESSAGE = "Flag state for feature '%s' scheduled to update at %s UTC."
```

api/audit/tasks.py
```diff
@@ -48,6 +48,16 @@ def _create_feature_state_audit_log_for_change_request(  # type: ignore[no-untyp
...
 
+    if feature_state.is_scheduled:
+        logger.info(
+            "FeatureState is not due to go live yet. Likely means the change request was rescheduled.",
+        )
+        create_feature_state_went_live_audit_log.delay(
+            delay_until=feature_state.live_from,
+            args=(feature_state.id,),
+        )
+        return
```
khvn26: This code branch makes we wonder how we handle cases when CRs are rescheduled to an date earlier than the original one.

api/features/models.py
```diff
@@ -958,6 +959,12 @@ def get_create_log_message(self, history_instance) -> typing.Optional[str]:  # t
         return audit_helpers.get_environment_feature_state_created_audit_message(self)
 
     def get_update_log_message(self, history_instance) -> typing.Optional[str]:  # type: ignore[no-untyped-def]
+        if self.change_request and self.is_scheduled:
+            return FEATURE_STATE_WILL_BE_UPDATED_MESSAGE % (
+                self.feature.name,
+                self.live_from.strftime("%Y-%m-%d %H:%M"),  # type: ignore[union-attr]
```
khvn26: Why this time format specifically? Do you think it makes sense to use a predetermined one (likely ISO8601)?

api/audit/constants.py
khvn26: > UTC

It's not currently user-configurable, but there's a [source of truth](https://github.com/Flagsmith/flagsmith/blob/365ad46717cc4177936c5839eba272150e5a650c/api/app/settings/common.py#L461-L462) for which timezone is used throughout the application, and I'd like to avoid hard-coding it anywhere else.

Ideally, the timestamp used with the template should contain the tz info.

emyller: If the associated Change Request is relevant:
```suggestion
FEATURE_STATE_WILL_BE_UPDATED_MESSAGE = "Flag state for feature '%s' scheduled for update by Change Request '%s' at '%' UTC."
```

For context: https://github.com/Flagsmith/flagsmith/issues/5030#issuecomment-2902973135

Otherwise, I agree your suggestion is the most sound. Personally, as a user, I find the intertwining of scheduled updates to change requests quite confusing — [optionally] see observation (4) in the above.

api/audit/tasks.py
emyller: I did a _quick_ manual test through this case specifically and it seemed to behave correctly at a glance, so I moved on. I did another test just now and observed by the second: I noticed that the "went live" log is actually delayed to the previous schedule (future) — it got me the first time because the timestamp points to the actual _go live_ date rescheduled to (past). 🤦

Thanks for catching this. Allow me some time to work on proper tests.

api/features/models.py
emyller: Great question. ISO8601 was my first thought too, though I appreciate that audit log messages are already end-user facing, so I chose the format that most matches the UX so far (having the date and time selector from the UI in mind).

I'm not a fan of this either so I created an issue earlier to raise the question: https://github.com/Flagsmith/flagsmith/issues/5493

If the frontend will localize and format timestamps then ISO8601 is certainly the wisest. Otherwise I guess that a good portion of the user base would feel more comfortable with a slightly easier to read version.

Then again, I don't have any certainty on the train of thought above, so none of this is strongly held.

emyller: Attempted fix in 0dd1b0d5984a967dc677623cfba83a62eb9aedb5. I think resorting to Django's [DATETIME_FORMAT](https://docs.djangoproject.com/en/5.2/ref/settings/#datetime-format) is the best-effort until we're able to deliver audit logs to the frontend as JSON.

api/audit/constants.py
emyller: Should be improved with 0dd1b0d5984a967dc677623cfba83a62eb9aedb5 (ref [previous comment](https://github.com/Flagsmith/flagsmith/pull/5484#discussion_r2107906244)).

api/audit/tasks.py
emyller: Fixed in 447a9e8f62a959cdb0ac82f0ee23b5a39d58f9c5.

api/audit/constants.py
emyller: Addressed with 0dd1b0d5984a967dc677623cfba83a62eb9aedb5.

api/audit/tasks.py
```diff
@@ -48,19 +48,41 @@ def _create_feature_state_audit_log_for_change_request(  # type: ignore[no-untyp
...
-    AuditLog.objects.create(
-        related_object_id=feature_state.id,
-        related_object_type=RelatedObjectType.FEATURE_STATE.name,
-        environment=feature_state.environment,
-        project=feature_state.environment.project,
-        log=log,
-        is_system_event=True,
-        created_date=feature_state.live_from,
-    )
+    log = {
```
khvn26: nit

```suggestion
    log_data = {
```

api/audit/tasks.py
```diff
@@ -48,19 +48,41 @@ def _create_feature_state_audit_log_for_change_request(  # type: ignore[no-untyp
...
+    # NOTE: This NEEDS to leverage btree indexes on AuditLog
+    audit_logged = AuditLog.objects.filter(**log).exists()
+    if audit_logged:
+        logger.info(
+            "FeatureState update audit log already exists. "
+            "Likely the change request was rescheduled to an earlier date.",
+        )
+        return
+
+    AuditLog.objects.create(**log)
```
khvn26: Can be condensed to:

```suggestion
    _, created = AuditLog.objects.create_or_update(**log)
    if not created:
        logger.info(
            "FeatureState update audit log already exists. "
            "Likely the change request was rescheduled to an earlier date.",
        )
```

api/tests/integration/audit/test_audit_logs.py
```diff
@@ -123,6 +132,90 @@ def test_retrieve_audit_log_for_feature_state_enabled_change(
...
+        history_instance=history_instance,
+    )
+
+    # Then
+    audit_log = AuditLog.objects.first()
+    feature_name = feature_state_obj.feature.name
+    assert audit_log.log == f"Flag state updated for feature: {feature_name}"
+
+
+# Future people please bump this up when it's due
```
khvn26, line 163: 🤣

khvn26 (approved):
LGTM in general 👍

api/audit/tasks.py
emyller: This _looks_ better indeed, but I suggest we avoid the occasional third db operation (an [unnecessary write](https://github.com/django/django/blob/5.2.1/django/db/models/query.py#L1012-L1014)) if you agree.

api/audit/tasks.py
emyller: Fixed in 61601aa566a734106a7de3537e761dff4ebf5c5a.

api/audit/tasks.py
khvn26: Ah, sorry, I made a mistake — I actually meant to propose `get_or_create`.

emyller: I failed to realize you meant that despite having just looked at `get_or_create` docs myself. lol
Improved with f5644a96ea7d8db7381e65260d03fcf62b80724e.

PR #5487, opened 2025-05-23
ci: add split-testing to private cloud image

matthewelwell:

## Changes

Creates a new docker target to build a new flagsmith-split-testing image. This image includes all functionality in the flagsmith-private-cloud image, as well as the functionality added in [flagsmith-split-testing](https://github.com/flagsmith/flagsmith-split-testing). 

I've tested this using the image built as part of the PR to confirm that the split testing module is correctly included in the build, along with the rest of the private cloud modules.


khvn26:
Why have a separate shippable target for this? Is there a reason we can't add it to flagsmith-private-cloud and have it behind a feature flag?

khvn26:
> Why have a separate shippable target for this? Is there a reason we can't add it to flagsmith-private-cloud and have it behind a feature flag?

This has been clarified off-line; the split-testing module brings heavy dependencies we don't want to maintain in the main image, and we need it continuously published for our partner to be able to demo latest versions with split testing functionality enabled.

khvn26 (approved):

PR #5497, opened 2025-05-27
feat: integrations onboarding

### LINKED ISSUE #5403: Capture relevant integrations on organisation sign up
When a new organisation is created we should ask users to select from a list of integrations, or provide others that they'd like to see an integration with. Something like the following...

Image

We should research other services to see how they're handling this. It'd also be interesting to see if we should subsequently have them configure the integrations as part of the onboarding, or if this is something that we should still just delegate to later. 

This information would be valuable to the product in terms of validating which integrations are important to our users. 

We still need to work out how / where we capture this information.

### LINKED ISSUE #5495: Get / Set favourite tools

As part of the new onboarding, users will be able to select their favourite tools which will affect the welcome page.

Image

Get / set a list of favourite tools with the ability to enter a custom one. Due to this I think we keep the BE unopinionated about the values.

Maybe there's potential for other unstructured data and we could keep the endpoint generic. Something like

```
POST /preferences
{
  tools: {
   completed: true, 
   selection: ['slack', 'jira'] 
  }
}
```

```
GET /auth/users/me
{
  ...
  tools: {
   completed: true, 
   selection: ['slack', 'jira'] 
  }
}
```

Storing the data as traits

### Additional context

_No response_

kyle-ssg:

## Changes

image

_Please describe._


frontend/web/components/IntegrationSelect.tsx
```diff
@@ -0,0 +1,380 @@
...
+          }}
+          value={customTool}
+          placeholder='Enter a name...'
+        />
+        <div className={'text-end mt-4'}>
+          <Button
+            onClick={() => {
+              setShowCustomTool(false)
+              setSelected(selected.concat([customTool]))
+            }}
```
Zaimwa9, line 250: I'd just reset the state setCustomTool here so if you open it again it's empty

frontend/web/components/IntegrationSelect.tsx
```diff
@@ -0,0 +1,380 @@
...
+
+              <div className='my-2 text-center'>
+                Not listed ?{' '}
+                <Button theme='text' onClick={() => setShowCustomTool(true)}>
+                  Click here
+                </Button>
+              </div>
+            </div>
+            <hr className='mb-5 mt-0' />
+            <div className='sticky d-flex justify-content-end gap-4'>
```
Zaimwa9, line 204: I don't know why the stickiness is not working when on a sub category. So we don't see the buttons it can be a bit confusing if you have some time to get why

image

frontend/web/components/IntegrationSelect.tsx
```diff
@@ -0,0 +1,380 @@
...
+        integrations: [],
+      },
+    }).then(onComplete)
+  const submit = () =>
+    updateTools({
+      tools: {
+        completed: true,
+        integrations: selected,
+      },
+    }).then(onComplete)
```
Zaimwa9: Are you fine with using async await syntax here ?

frontend/web/components/IntegrationSelect.tsx
```diff
@@ -0,0 +1,380 @@
...
+import { IonIcon } from '@ionic/react'
+import { close } from 'ionicons/icons'
+import { useUpdateOnboardingMutation } from 'common/services/useOnboarding'
+import InfoMessage from './InfoMessage'
+
+type IntegrationSelectType = {
+  onComplete: () => void
+}
+
+const integrationSummaries: IntegrationSummary[] = [
```
Zaimwa9: Can we put them in a separate data/constant file to light up the file ?

frontend/web/components/IntegrationSelect.tsx
```diff
@@ -0,0 +1,380 @@
...
+    image: '/static/images/integrations/prometheus.svg',
+    title: 'Prometheus',
+  },
+  {
+    categories: ['Monitoring'],
+    image: '/static/images/integrations/sumologic.svg',
+    title: 'SumoLogic',
+  },
+]
+const categoryDescriptions: Record<
```
Zaimwa9: same for this one

frontend/web/components/IntegrationSelect.tsx
```diff
@@ -0,0 +1,380 @@
...
+  const [search, setSearch] = useState<string>('')
+  const integrationsData = Utils.getFlagsmithValue('integration_data')
+  const [category, setCategory] = useState<string>(ALL_CATEGORY)
+
+  // Filter integrations by category and search term
+  const allIntegrations = useMemo(() => {
+    if (!integrationsData) {
+      return [] as typeof integrationSummaries
+    }
+    const parsed = Object.values(JSON.parse(integrationsData)).concat(
```
Zaimwa9: Can we find a more self-explanatory name like `parsedCategories`

frontend/web/components/IntegrationSelect.tsx
```diff
@@ -0,0 +1,380 @@
...
+    'Receive webhooks when your features change or when audit logs occur.',
+}
+
+const ALL_CATEGORY = 'All'
+
+const IntegrationSelect: FC<IntegrationSelectType> = ({ onComplete }) => {
+  // Initialize search as an empty string
+  const [search, setSearch] = useState<string>('')
+  const integrationsData = Utils.getFlagsmithValue('integration_data')
+  const [category, setCategory] = useState<string>(ALL_CATEGORY)
```
Zaimwa9: We could rename it to selectedCategory, setSelectedCategory do not mix in case of using a map

frontend/web/components/IntegrationSelect.tsx
```diff
@@ -0,0 +1,380 @@
...
+      const matchesSearch =
+        !search || v.title.toLowerCase().includes(search.toLowerCase())
+
+      return matchesCategory && matchesSearch
+    })
+
+    return sortBy(filtered, 'title') as typeof integrationSummaries
+  }, [integrationsData, category, search])
+
+  const [selected, setSelected] = useState<string[]>([])
```
Zaimwa9: Same if you don't mind to rename to something like selectedIntegrations

frontend/web/components/IntegrationSelect.tsx
```diff
@@ -0,0 +1,380 @@
...
+                      onClick={() => {
+                        if (isSelected) {
+                          setSelected(
+                            selected.filter(
+                              (selectedItem) => selectedItem !== v.title,
+                            ),
+                          )
+                        } else {
+                          setSelected(selected.concat(v.title))
+                        }
```
Zaimwa9: We might put it in its own handler above to have the logic isolated

Zaimwa9 (changes requested):
A couple of comments, not critical but if maybe you have time to split up a bit the component. Then it should be quickly good to go

I just looked at IntegrationSelect, it needs rebased to remove the welcome page diff

frontend/web/components/IntegrationSelect.tsx
kyle-ssg: Good catch, on modal dismiss too

frontend/web/components/IntegrationSelect.tsx
kyle-ssg: I think its just down to the container having a min height, should be better now

Zaimwa9 (approved):
LGTM

PR #5500, opened 2025-05-28
feat: track welcome page tasks and integrations

Zaimwa9:

## Changes

_Please describe._

_Please describe._


api/users/models.py
```diff
@@ -111,7 +111,7 @@ class FFAdminUser(LifecycleModel, AbstractUser):  # type: ignore[django-manager-
     last_name = models.CharField("last name", max_length=150)
     google_user_id = models.CharField(max_length=50, null=True, blank=True)
     github_user_id = models.CharField(max_length=50, null=True, blank=True)
-
+    onboarding = models.JSONField(blank=True, null=True)
```
gagantrivedi: Do we gain anything from this being a JSON field? If not, can we change it to a text field? This could be a bit painful with Oracle once we merge this into EE.

Zaimwa9: Good point, I wasn't aware Oracle could be part of the stack. It was not obligatory but better to preserve automatic serialization/deserialization and native querying but indeed, that's a blocker. I'll move it to a text field

api/custom_auth/views.py
```diff
@@ -134,6 +136,24 @@ def perform_destroy(self, instance):  # type: ignore[no-untyped-def]
             )
         )
 
+    @action(
+        detail=False,
+        methods=["patch"],
+        url_path="me/onboarding",
+        permission_classes=[IsAuthenticated],
+    )
+    def patch_onboarding(self, request, *args, **kwargs):  # type: ignore[no-untyped-def]
```
gagantrivedi: We can add some typing here

api/users/serializers.py
```diff
@@ -148,17 +152,74 @@ class UserPermissionGroupSerializerDetail(UserPermissionGroupSerializer):
...
+
+class OnboardingTaskSerializer(serializers.Serializer[None]):
+    name = serializers.CharField()
+    completed_at = serializers.DateTimeField(
+        allow_null=True,
+        required=False,
+        default=lambda: datetime.now(),
+    )
+
+    def validate_completed_at(self, completed_at: datetime) -> datetime:
```
gagantrivedi: We need this after adding the default as well?

api/users/models.py
```diff
@@ -111,7 +111,7 @@ class FFAdminUser(LifecycleModel, AbstractUser):  # type: ignore[django-manager-
     last_name = models.CharField("last name", max_length=150)
     google_user_id = models.CharField(max_length=50, null=True, blank=True)
     github_user_id = models.CharField(max_length=50, null=True, blank=True)
-
+    onboarding = models.TextField(blank=True, null=True)
```
gagantrivedi: nit: can we call it onboarding data or something? Just onboarding feels a bit too vague?

api/tests/unit/users/test_unit_users_serializers.py
```diff
@@ -1,13 +1,97 @@
...
+            "completed_at": datetime.fromisoformat("2024-01-02T15:00:00Z"),
+        },
+    ]
+    assert data["tools"] == {
+        "completed": True if tools_completed is None else tools_completed,
+        "integrations": ["integration-1", "integration-2"],
+    }
+
+
+# def test_onboarding_task_to_representation_converts_datetime_to_json_compatible_string() -> (
```
gagantrivedi: What happend with this test?

gagantrivedi (changes requested):

api/tests/unit/users/test_unit_users_serializers.py
Zaimwa9: Thanks, to be removed. Now that it's only text, there was no need for a custom `to_representation` (as datetime are not json-serializable by default)

api/users/serializers.py
Zaimwa9: So my understanding is that default applies to field missing `{name: task}` but not to None.
`validate_<field>` would deal with defaulting None value to now.
I didn't use `allow_null = False` would throw an error if it's none and we didn't want to be too strict and return an error in that case

Maybe there is a cleaner way to do so I was unaware of ?

gagantrivedi: No, you are right. But can we update the type hint to indicate that the validate is supposed to deal with None as well? Maybe something like: 
`    def validate_completed_at(self, completed_at: datetime | None) -> datetime:`

gagantrivedi (approved):
Looks good apart from minor type hint nit

PR #5502, opened 2025-05-28
fix: Handle bad input before checking permission

emyller:
Fixes this exception:

```
int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```


api/features/feature_segments/permissions.py
```diff
@@ -22,13 +20,17 @@ def has_permission(self, request, view):  # type: ignore[no-untyped-def]
...
 
         if view.action == "create":
-            with suppress(Environment.DoesNotExist, ValueError):
-                environment = request.data.get("environment")
-                environment = Environment.objects.get(id=int(environment))
+            if not (environment_pk := request.data.get("environment")):
+                return False
+
+            try:
+                environment = Environment.objects.get(pk=environment_pk)
```
khvn26, line 30: What happens now with a query string like `?environment=foo`?

emyller: An `Environment.DoesNotExist` happens. It could be avoided by checking whether the input is a number, though IMO 1. it seems picky given it's a very cheap query that would run in 99.9% cases anyway, and 2. it creates bugs if we ever decide to make PKs UUIDs.

emyller: It crossed my mind consulting with the view serializer for proper validation, but it seemed like too much effort for little gain. 🤔

emyller:
```
ValueError: Field 'id' expected a number but got 'invalid'.
```

Should be fixed now with 9d118a21e4f3e7b4efa19bebecf58a8ee8f1a51e.

matthewelwell (approved):

PR #5507, opened 2025-05-29
fix: prevent signup blocking invitation signups

### LINKED ISSUE #5506: Cannot invite users using invitation link

- [x] Self Hosted with Docker
- [ ] SaaS at flagsmith.com

### Describe the bug

Running Flagsmith locally, with the following env var set: "PREVENT_SIGNUP: 1", users cannot sign-up using valid invitation link. Also the error message is unclear and we can see (Unauthorized 401) error code in the browser's console.

From Flagsmith docs: https://docs.flagsmith.com/deployment/hosting/locally-frontend

"
PREVENT_SIGNUP: Determines whether to prevent manual signups **_without invites_**. Set it to any value to prevent signups.
"

### Steps To Reproduce

Start the Flagsmith container with `PREVENT_SIGNUP: 1`, set it up as usual.
As an Admin user, send invitation link to create a new user.
As a new user, open the link and fill-in: First Name, Last Name, Email, Password, make sure all valid and click "Create Account".

### Expected behavior

The user should be created and provided access to the Organization.
Instead, got Unauthorized 401 error code in the Browser's console.
Also the error message to the user is unclear: "".

### Workaround

Set both:
```
PREVENT_SIGNUP: 0
ALLOW_REGISTRATION_WITHOUT_INVITE: 0
```

This way users won't be able to sign-up without invitations - yet - the Sign-up button appears but will return error if trying to click it.

### Screenshots

Image

Zaimwa9:

closes #5506 

## Changes
- added email and link invitation checks in `IsSignupAllowed` permissions
- Use the request error in the frontend

- Added tests

To test:
1. Run backend with `"PREVENT_SIGNUP": "1",`
2. Share invitation link / invite with email a user
3. Complete signup

https://www.loom.com/share/02c75a8dc28c4378b9856867c2ca7c85

### DIFF
--- a/api/custom_auth/permissions.py
+++ b/api/custom_auth/permissions.py
@@ -2,6 +2,8 @@ from django.conf import settings
 from django.views import View
 from rest_framework.permissions import AllowAny, IsAuthenticated
 from rest_framework.request import Request
+from organisations.invites.models import InviteLink, Invite
+from rest_framework.exceptions import PermissionDenied
 
 
 class CurrentUser(IsAuthenticated):
@@ -17,8 +19,26 @@ class CurrentUser(IsAuthenticated):
 
 
 class IsSignupAllowed(AllowAny):
+    message = "Signing-up without a valid invitation is disabled. Please contact your administrator."
+
     def has_permission(self, request: Request, view: View) -> bool:
-        return not settings.PREVENT_SIGNUP
+        if not settings.PREVENT_SIGNUP:
+            return True
+
+        email = request.data.get("email")
+        if email and Invite.objects.filter(email__iexact=email).exists():
+            return True
+
+        invite_hash = request.data.get("invite_hash")
+        if invite_hash:
+            try:
+                invite_link = InviteLink.objects.get(hash=invite_hash)
+                if not invite_link.is_expired:
+                    return True
+            except InviteLink.DoesNotExist:
+                pass
+
+        raise PermissionDenied(self.message)
 
 
 class IsPasswordLoginAllowed(AllowAny):
--- a/api/custom_auth/permissions.py
+++ b/api/custom_auth/permissions.py
@@ -20,7 +20,7 @@ class CurrentUser(IsAuthenticated):
 
 
 class IsSignupAllowed(AllowAny):
-    message = "Signing-up without a valid invitation is disabled. Please contact your administrator."
+    message = "Signing up without an invitation is disabled. Please contact your administrator."
 
     def has_permission(self, request: Request, view: View) -> bool:
         if not settings.PREVENT_SIGNUP:
@@ -34,10 +34,11 @@ class IsSignupAllowed(AllowAny):
         if invite_hash:
             try:
                 invite_link = InviteLink.objects.get(hash=invite_hash)
-                if not invite_link.is_expired:
-                    return True
             except InviteLink.DoesNotExist:
                 pass
+
+        if not invite_link.is_expired:
+            return True
 
         raise PermissionDenied(self.message)
 
--- a/api/custom_auth/permissions.py
+++ b/api/custom_auth/permissions.py
@@ -1,9 +1,10 @@
 from django.conf import settings
 from django.views import View
+from rest_framework.exceptions import PermissionDenied
 from rest_framework.permissions import AllowAny, IsAuthenticated
 from rest_framework.request import Request
-from organisations.invites.models import InviteLink, Invite
-from rest_framework.exceptions import PermissionDenied
+
+from organisations.invites.models import Invite, InviteLink
 
 
 class CurrentUser(IsAuthenticated):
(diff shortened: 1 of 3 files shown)


api/custom_auth/permissions.py
```diff
@@ -17,8 +20,26 @@ def has_object_permission(self, request, view, obj):  # type: ignore[no-untyped-
...
 class IsSignupAllowed(AllowAny):
+    message = "Signing-up without a valid invitation is disabled. Please contact your administrator."
+
     def has_permission(self, request: Request, view: View) -> bool:
-        return not settings.PREVENT_SIGNUP
+        if not settings.PREVENT_SIGNUP:
+            return True
+
+        email = request.data.get("email")
+        if email and Invite.objects.filter(email__iexact=email).exists():
```
emyller, line 30: Just checking. Do we have a db index that could favor this query? I'm not entirely sure but the `Invite` model sounds like it can hold many records.

api/custom_auth/permissions.py
```diff
@@ -17,8 +20,26 @@ def has_object_permission(self, request, view, obj):  # type: ignore[no-untyped-
...
+            return True
+
+        invite_hash = request.data.get("invite_hash")
+        if invite_hash:
+            try:
+                invite_link = InviteLink.objects.get(hash=invite_hash)
+                if not invite_link.is_expired:
+                    return True
+            except InviteLink.DoesNotExist:
+                pass
```
emyller:
```suggestion
            try:
                invite_link = InviteLink.objects.get(hash=invite_hash)
            except InviteLink.DoesNotExist:
                return False
            return not invite_link.is_expired
```

This keeps only the code that is expected to fail within the `try` clause, and also returns `False` rather than `None` — if that's ideal.

api/custom_auth/permissions.py
```diff
@@ -17,8 +20,26 @@ def has_object_permission(self, request, view, obj):  # type: ignore[no-untyped-
 
 
 class IsSignupAllowed(AllowAny):
+    message = "Signing-up without a valid invitation is disabled. Please contact your administrator."
```
emyller: This message sounds too specific for a permission with a generic name. 🤔 
Perhaps we should use it in the code branch within `has_permission`?

api/custom_auth/permissions.py
Zaimwa9: Not that much because they are deleted after usage. I also wanted to be sure there were not a breach on the flow but in `models/FFAdminUser` we have this so except abusive usage the balance should be respected:
```
    def join_organisation_from_invite_email(self, invite_email: "Invite"):  # type: ignore[no-untyped-def]
        if invite_email.email.lower() != self.email.lower():
            raise InvalidInviteError("Registered email does not match invited email")
        self.join_organisation_from_invite(invite_email)
        self.permission_groups.add(*invite_email.permission_groups.all())
        invite_email.delete()
```

api/custom_auth/permissions.py
Zaimwa9: Yes I thought the same but that's actually the only purpose of this permission afaik:
- `PREVENT_SIGNUP = false` => Always allowed
- `PREVENT_SIGNUP = true` => Needs invitation or error

emyller: I see. In that case, just a minor tweak:
```suggestion
    message = "Signing up without an invitation is disabled. Please contact your administrator."
```

api/custom_auth/permissions.py
emyller: Nice!

api/custom_auth/permissions.py
```diff
@@ -17,8 +20,26 @@ def has_object_permission(self, request, view, obj):  # type: ignore[no-untyped-
...
+        invite_hash = request.data.get("invite_hash")
+        if invite_hash:
+            try:
+                invite_link = InviteLink.objects.get(hash=invite_hash)
+                if not invite_link.is_expired:
+                    return True
+            except InviteLink.DoesNotExist:
+                pass
+
+        raise PermissionDenied(self.message)
```
emyller, line 43: Following returning `False` above, DRF should already raise using the `message` attribute from the class [on its own](https://www.django-rest-framework.org/api-guide/permissions/#custom-permissions). 😉

api/custom_auth/permissions.py
Zaimwa9: Agreed to extract the `is_expired` check from the try/catch block - thanks.

Allow me to disagree on the False/None part. It should not be returning `None` as it would:
- return true (allow-all, valid email, valid link)
- surface a server/unknown error
- if doesNotExist or is_expired => passes through the PermissionDenied.
Which to me is the correct error to return.

Zaimwa9: Also for context, DRF will short-circuit to a 401 if authentication fails (via get_authenticators()) without explicitely raising an error

emyller: Thanks for the context.

Sorry I addressed the `raise` statement in [another comment](https://github.com/Flagsmith/flagsmith/pull/5507#discussion_r2114074125).

As per DRF [documentation](https://www.django-rest-framework.org/api-guide/permissions/#custom-permissions), it will automatically raise `PermissionDenied` when `has_permission` returns `False`. Having it manually raise feels like an anti-pattern. If you need to customize how DRF will present that exception to the HTTP response, you may set the attributes `code` and `message` to the permission class.

There's also this [SO thread](https://stackoverflow.com/q/29936323/307511) with examples.

Let me know if this helps.

Zaimwa9:
Closed in favor of #5508

Issue #5272, opened 2025-03-31
As a dashboard user, I can define an action to enable / disable a flag for a given segment in a release pipeline stage


matthewelwell:
TODO

Considerations: 
 * What happens if a flag already has existing segment overrides? 
 * What happens if a flag already has an override for the segment defined by the given stage?

Answer to the above: 

This action should just ignore any existing segment overrides. The release pipeline should add any new segments, and leave existing ones, assuming there isn't any conflict. If there is already a segment override for the segment defined in the given release pipeline stage, the release pipeline should clobber the existing override. 

### Discussion 04/11/2025

We talked about the locking of a flag with regards to segment overrides. The current plan is to lock the flag to any segment override and environment level modifications, but we could perhaps in the future only lock segment override modifications for the segments that exist in the release pipeline. This would benefit customers that want to slowly migrate over to release pipelines so they may only want to use the pipelines for certain segments, and not others. 

We should also prevent the segment from being deleted when it used as part of a release pipeline.

Issue #5269, opened 2025-03-31
As a dashboard user, I can define the trigger on a release pipeline stage to be 'When a flag enters this stage'


matthewelwell:
A stage with this trigger will execute as soon as any flag in the pipeline has completed the previous stage in the pipeline. 

<img width="373" alt="Image" src="https://github.com/user-attachments/assets/3221aaf5-8e4e-44a1-af3e-351309f0ff65" />

Issue #5268, opened 2025-03-31
As a dashboard user, I can add a stage to a release pipeline


matthewelwell:
Users can add stages to a given release pipeline. When creating a stage, the user is required to provide the following information: 

1. Name
2. Environment (from within the current project)

Note that the environment can be different between different stages in a given pipelines. 

We should impose a limit on the number of stages in a given pipeline. This limit is not really a technical challenge, but more to prevent users from shooting themselves in the foot. Suggest that, to start with, we set this limit at 30 stages.

Issue #5267, opened 2025-03-31
As a dashboard user, I am able to create a new release pipeline


matthewelwell:
Release pipelines will live at the project level. When you create a release pipeline, you will need to provide the following fields:

 1. Name

Release pipelines should be initially created in a 'draft' status. When the user is ready to use the release pipeline, they should explicitly have to 'publish' the release pipeline. 

The FE should manage the data until the point at which the user clicks 'save release pipeline', at which point, it will make the necessary requests to the API. 

Only project administrators should be able to create release pipelines.

PR #5509, opened 2025-05-30
feat: Release Pipelines UI

tiagoapolo:

## Changes

This pull request implements pipeline creation flow for release pipelines, including pipeline stages and `on_enter` action. Also it's split into 3 commits:
1. Pipeline Creation and Pipeline List
2. Pipeline Stages Creation + `ON_ENTER` trigger + Stage Action
3. View Pipeline Details

Tasks:
-  [#5267](https://github.com/Flagsmith/flagsmith/issues/5267) and [5277](https://github.com/Flagsmith/flagsmith/issues/5277) => Implemented in this [commit](https://github.com/Flagsmith/flagsmith/pull/5509/commits/84ade49b7459be18c0e9a949aa295df4e612cd23)
- [#5268](https://github.com/Flagsmith/flagsmith/issues/5268), [#5269](https://github.com/Flagsmith/flagsmith/issues/5269), [#5273](https://github.com/Flagsmith/flagsmith/issues/5273) and [#5272](https://github.com/Flagsmith/flagsmith/issues/5272) => Implemented in this [commit](https://github.com/Flagsmith/flagsmith/pull/5509/commits/fa5a755522544d1d4681ade6d9c36d9c9b36119b) 
- This [commit](https://github.com/Flagsmith/flagsmith/pull/5509/commits/c0013fa9bb44ad01faa691b7b40825c301f18a88) implements the view pipeline details.

https://www.loom.com/share/ef566337763e4a5e86763d6cf9babc57?sid=eb50c14a-27d7-49d5-8838-db6d0704fc41

⚠️ Tested using branch with backend and frontend [`release-pipe-front-back`](https://github.com/Flagsmith/flagsmith/tree/release-pipe-front-back)


frontend/common/services/useReleasePipelines.ts
```diff
@@ -0,0 +1,166 @@
...
+        query: ({
+          pipelineId,
+          projectId,
+          ...rest
+        }: Req['getPipelineStages']) => ({
+          url: `projects/${projectId}/release-pipelines/${pipelineId}/stages/?${Utils.toParam(
+            rest,
+          )}`,
+        }),
+      }),
```
Zaimwa9, line 68: The stages are returned within `api/v1/projects/3/release-pipelines/2/`. If it's not in use anymore, let's remove it and the associated call

image

frontend/common/services/useReleasePipelines.ts
```diff
@@ -0,0 +1,166 @@
...
+      deleteReleasePipeline: builder.mutation<{}, Req['deleteReleasePipeline']>(
+        {
+          invalidatesTags: [{ id: 'LIST', type: 'ReleasePipelines' }],
+          query: (query: Req['deleteReleasePipeline']) => ({
+            method: 'DELETE',
+            url: `projects/${query.projectId}/release-pipelines/${query.pipelineId}/`,
+          }),
+        },
+      ),
+      getPipelineStage: builder.query<
```
Zaimwa9, line 47: Same as below, it doesn't seem to be used so maybe we can remove it if not necessary in the future

frontend/common/types/requests.ts
```diff
@@ -68,6 +71,22 @@ export type RegisterRequest = {
   organisation_name?: string
   marketing_consent_given?: boolean
 }
+
+export interface StageActionRequest {
+  action_type: StageActionType
+  action_body: { enabled: boolean; segment_id?: number }
+}
+
+export type PipelineStageRequest = {
```
Zaimwa9, line 81: I think we could favor an interface here

Zaimwa9: And also opening the discussion here but the file starts to be massive and some types are reused in components. So maybe we could start using dedicated types files wdyt ?

frontend/common/types/responses.ts
```diff
@@ -798,6 +800,52 @@ export type IdentityTrait = {
...
+  TOGGLE_FEATURE = 'TOGGLE_FEATURE',
+  TOGGLE_FEATURE_FOR_SEGMENT = 'TOGGLE_FEATURE_FOR_SEGMENT',
+}
+// TODO: Check if this is correct
+export interface StageActionResponse
+  extends Omit<StageActionRequest, 'action_body'> {
+  action_body: string
+}
+
+export type PipelineStage = {
```
Zaimwa9: So, tbh, i've always been a little confused about when using interface versus types. My rule of thumbs was:
- Type for union, primitives, functions
- Interface to represent object shapes, use in OOP context, will be extended/omitted

So i'd be in favor of an interface here and StageTrigger but i'm not that opinionated on the topic, just to have your position on this for future consistency

frontend/web/components/base/DropdownMenu.tsx
```diff
@@ -0,0 +1,110 @@
...
+  iconFill = '#9DA4AE',
+  iconWidth = 18,
+  items,
+}) => {
+  const [isOpen, setIsOpen] = useState(false)
+  const btnRef = useRef<HTMLButtonElement>(null)
+  const dropDownRef = useRef<HTMLDivElement>(null)
+  useOutsideClick(dropDownRef, () => setIsOpen(false))
+
+  function calculateListPosition(
```
Zaimwa9: It's been moved in `utils/calculateListPosition` that can be re-used

Zaimwa9: My bad actually, I didn't clean-up other places it's been used (4-5 components), not critical but if you want to do it in this PR, great

frontend/web/components/release-pipelines/CreateReleasePipeline.tsx
```diff
@@ -0,0 +1,219 @@
...
+      isSuccess: isCreatingPipelineSuccess,
+    },
+  ] = useCreateReleasePipelineMutation()
+
+  const [
+    createStages,
+    {
+      error: createStageError,
+      isLoading: isCreatingStage,
+      isSuccess: isCreatingStageSuccess,
```
Zaimwa9: Not in use atm

frontend/web/components/release-pipelines/CreateReleasePipeline.tsx
```diff
@@ -0,0 +1,219 @@
...
+  }, [isCreatingPipelineSuccess, isStagesCreationSuccess, history, projectId])
+
+  const handleOnChange = (newStageData: DraftStageType, index: number) => {
+    const updatedStages = stages.map((stage, i) =>
+      i === index ? newStageData : stage,
+    )
+    setStages(updatedStages)
+  }
+
+  const validatePipelineData = () => {
```
Zaimwa9, line 121: Here I would also ensure that all the existing stages are complete: name + action/target_action.

It's possible at the moment to create the pipeline with incomplete stages. It creates it but skips the stages

image

frontend/web/components/release-pipelines/CreateReleasePipeline.tsx
```diff
@@ -0,0 +1,219 @@
...
+
+      if ('data' in response && response.data.id) {
+        const pipelineId = response.data.id
+        const stagesCreationPromises = stages.map((stage, index) => {
+          return createStages({
+            ...stage,
+            order: index,
+            pipeline: pipelineId,
+            project: Number(projectId),
+          })
```
Zaimwa9: I know it's the API implementation but it feels a bit weird to:
1. not create a pipeline with stages if there
2. create the stages separately
3. have to create each stage individually

frontend/web/components/release-pipelines/CreateReleasePipeline.tsx
```diff
@@ -0,0 +1,219 @@
...
+        })
+
+        await Promise.all(stagesCreationPromises)
+      }
+      setStagesCreationSuccess(true)
+    } catch (error) {
+      console.error('erroroooo')
```
Zaimwa9:
```suggestion
```
😃

frontend/web/components/release-pipelines/CreateReleasePipeline.tsx
```diff
@@ -0,0 +1,219 @@
...
+    try {
+      const response = await createReleasePipeline({
+        name: pipelineData.name,
+        projectId: Number(projectId),
+        status: PipelineStatus.DRAFT,
+      })
+
+      if ('data' in response && response.data.id) {
+        const pipelineId = response.data.id
+        const stagesCreationPromises = stages.map((stage, index) => {
```
Zaimwa9: Can't we try catch specifically the create pipeline first to separate the errors ?

frontend/web/components/release-pipelines/ReleasePipelineDetail.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const HeaderWrapper = ({ children }: { children: React.ReactNode }) => (
+    <div className='app-container container'>
+      <Breadcrumb
+        items={[
+          {
+            title: 'Release Pipelines',
+            url: `/project/${projectId}/release-pipelines`,
+          },
+        ]}
+        currentPage={id}
```
Zaimwa9: Is it possible to override the displayed value with the name of the release ?

frontend/web/components/release-pipelines/CreateReleasePipeline.tsx
```diff
@@ -0,0 +1,219 @@
...
+                  showRemoveButton={index > 0}
+                  onRemove={() => handleRemoveStage(index)}
+                />
+                <div className='flex-1'>
+                  <StageArrow
+                    showAddStageButton={index === stages.length - 1}
+                    onAddStage={() =>
+                      setStages((prev) => prev.concat(blankStage))
+                    }
+                  />
```
Zaimwa9, line 240: I think we should disable this button if the previous stages are not complete

frontend/web/components/release-pipelines/StageInfo.tsx
```diff
@@ -0,0 +1,78 @@
...
+}
+
+const StageInfo = ({
+  environmentsData,
+  projectId,
+  stageData,
+}: StageInfoProps) => {
+  const environmentData = environmentsData?.find(
+    (environment) => environment.id === stageData?.environment,
+  )
```
Zaimwa9: We could pass the environment name from the parent that would calculate it for all the stages

frontend/web/styles/components/_release-pipelines.scss
```diff
@@ -0,0 +1,24 @@
...
+        border-bottom-width: 3px;
+        background-color: $hr-border-color;
+    }
+
+    .arrow-container {
+        overflow: hidden;
+        margin-right: -12px;
+        &__wrapper {
+            transform: translateX(-8px);
+        }
```
Zaimwa9, line 22: I'll put the comment here but I don't know where it belongs exactly. There is a kind of overflow conflict
image

But the problem comes from the select actually that should be absolute.

Or the overflow could be in the card and not on the big container. 

Just putting on your radar

frontend/web/components/release-pipelines/CreateReleasePipeline.tsx
```diff
@@ -0,0 +1,219 @@
...
+                placeholder='Release Pipeline Name'
+                onChange={(event: InputEvent) => {
+                  setPipelineData({
+                    ...pipelineData,
+                    name: Utils.safeParseEventValue(event),
+                  })
+                }}
+              />
+            )}
+            <Button
```
Zaimwa9, line 197: Not critical at all as the platform is not responsive but saw this. if low effort as I know this feature has still a lot to do otherwise no problem
image

frontend/web/components/release-pipelines/ReleasePipelineDetail.tsx
```diff
@@ -0,0 +1,124 @@
...
+      {
+        page: 1,
+        page_size: 100,
+        pipelineId: Number(id),
+        projectId: Number(projectId),
+      },
+      {
+        skip: !id || !projectId,
+      },
+    )
```
Zaimwa9, line 64: If the stages are in the release payload we can remove this request

Zaimwa9 (changes requested):
Honestly good work. it's overall working super well.

I'll just give another look to `CreatePipelineStage` and might add some comments when reviewing the api but really solid base.

There's just an API interface i'd mention in the other request that could have breaking change. Which is the way the stages are saved that could be simplified maybe

frontend/common/services/useReleasePipelines.ts
tiagoapolo: Yeah, this was updated recently before the UI was implemented, so it's worth remove the `/stages/` call

frontend/web/components/release-pipelines/CreateReleasePipeline.tsx
tiagoapolo: We don't need to validate it in this PR, it will be validated in [here](https://github.com/Flagsmith/flagsmith/pull/5523/files)

frontend/web/components/release-pipelines/CreateReleasePipeline.tsx
tiagoapolo: Yes, this has been raised to Gagan. @gagantrivedi I don't remember but have you implemented creating stages and pipelines in a single push?

frontend/web/components/release-pipelines/ReleasePipelineDetail.tsx
tiagoapolo: I don't have a strong opinion about this one, we could use the name.

frontend/web/components/release-pipelines/CreateReleasePipeline.tsx
tiagoapolo: Hmmm not sure if this is a good UX. It's best to let the user know what's missing than blocking him to create stages

Zaimwa9: I see what you mean and a benefit being that you can prepare your stages and then fill them up. But in that case I would either (or both):
- show a warning on incomplete stages
- be stricted on submission and block it unless stages are complete
Root problem to avoid imo is submitting a release sucesfully and realizing a stage is missing

frontend/web/styles/components/_release-pipelines.scss
Zaimwa9: Adding some details given the quality of the screenshot. When you open a select, the scroll happen on the container, not the card. It gets worse if you have many environments/segments

frontend/web/components/release-pipelines/CreateReleasePipeline.tsx
tiagoapolo: That is done in the next PR. I'm sorry for the confusion

frontend/web/components/release-pipelines/CreateReleasePipeline.tsx
```diff
@@ -75,11 +70,31 @@ function CreateReleasePipeline({
...
+    history.push(`/project/${projectId}/release-pipelines`)
+    toast('Release pipeline created successfully')
+  }
+
+  useEffect(() => {
+    if (isCreatingPipelineSuccess && !stages.length) {
+      return handleSuccess()
+    }
+
+    if (isCreatingPipelineSuccess && stages.length && isStagesCreationSuccess) {
```
Zaimwa9: Can we put them in 2 variables and have one or condition ?

frontend/web/components/release-pipelines/ReleasePipelineDetail.tsx
tiagoapolo: We're updating to use the stages coming release payload in here https://github.com/Flagsmith/flagsmith/pull/5523/files#diff-796a09b0d325c7398644d64a565089249d79375edf256c4e37fa9f334d2cd76bR96

frontend/common/services/useReleasePipelines.ts
tiagoapolo: This will be resolved in https://github.com/Flagsmith/flagsmith/pull/5523/files#diff-796a09b0d325c7398644d64a565089249d79375edf256c4e37fa9f334d2cd76bR96

frontend/common/types/requests.ts
tiagoapolo: We can use interfaces for sure specially for inheriting from other types, this type will be remove in a next PR

frontend/web/components/release-pipelines/CreateReleasePipeline.tsx
tiagoapolo: done ✅

Zaimwa9 (changes requested):

frontend/web/styles/components/_release-pipelines.scss
tiagoapolo: fixed

frontend/web/components/release-pipelines/CreatePipelineStage.tsx
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [environmentOptions, stageData])
+
+  useEffect(() => {
+    if (selectedAction.value === '') {
+      return handleOnChange('actions', [])
+    }
+
```
Zaimwa9: nit but to solve the above selectedSegment discrepancy we could reset here 
```
  if (!selectedAction.value.includes('FOR_SEGMENT')) {
    setSelectedSegment(undefined)
  }
```

frontend/web/components/release-pipelines/CreatePipelineStage.tsx
```diff
@@ -0,0 +1,237 @@
...
+  useEffect(() => {
+    if (environmentOptions?.length && stageData.environment === -1) {
+      handleOnChange('environment', environmentOptions?.[0]?.value)
+    }
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [environmentOptions, stageData])
+
+  useEffect(() => {
```
Zaimwa9: As those useEffects are side effects triggered by changes in single values.
Couldn't we simplify and directly have a handler for each of the input ? `onSelectSegment`, `onSelectAction` ?
I believe it would have the same results without the useEffects risks ?

frontend/web/components/release-pipelines/CreatePipelineStage.tsx
```diff
@@ -0,0 +1,237 @@
...
+}) => {
+  const [searchInput, setSearchInput] = useState('')
+  const [selectedAction, setSelectedAction] = useState<{
+    label: string
+    value: string
+  }>({ label: 'Select an action', value: '' })
+  const [selectedSegment, setSelectedSegment] = useState<{
+    label: string
+    value: number | null
+  }>()
```
Zaimwa9:
```suggestion
  const [selectedSegment, setSelectedSegment] = useState<{
    label: string
    value: number | null
  } | undefined>()
```

Zaimwa9: There is a state desynchronization here, we can have a stale segment. Not that critical because it's cleaned up before sending the request but I'd be keen to clean it now as those components might get more complex over time

To reproduce:
1. Select trigger action segment
2. Select a segment
3. Change the trigger
4. The selected value hangs around and is not matching the parent data anymore.

Check the selectedSegment, the value is still from previous selection

Zaimwa9 (changes requested):
Couple of comments on CreatePipielineStage and should be good to go

Zaimwa9 (approved):
Let's go to the next stage 👍

Issue #2771, opened 2023-09-14
Epic: Allow for large Segment rule values


dabeeeenster:
**Is your feature request related to a problem? Please describe.**

Sometimes people want to provide a large (> 100) list of UUIDs or other identifiable datatype to target Identities. For various reasons (e.g. local evaluation mode with limited user context data in memory on the server), they might not be able to use traits. 

**Describe the solution you'd like**

Allow the uploading of large lists of data that back segment rules like `IN`. This would be stored in a new type of data store such as Amazon S3 (or a new table in Postgres that accepts larger data values) and merged into memory of the flag engine at runtime. 

We would need to design how the data is provided/ingested (via API, CSV upload etc) and what the performance considerations/limits of this approach might be. 

**Describe alternatives you've considered**

Traits are the only way to achieve this sort of thing currently but sometimes disparate data makes them unworkable.

rolodato:
If we plan on having the Edge Proxy sync these segments, it might be worth considering if/how we could design this to support one/two-way sync of user traits. Currently, when using local flag evaluation, traits are limited in two ways:

1. Traits stored by Flagsmith are not accessible in local flag evaluation.
2. Applications using local flag evaluation do not propagate trait updates to Flagsmith.

The second limitation can be somewhat mitigated by making an explicit remote flag evaluation request to Flagsmith, but this is not very useful if server-side applications cannot normally access this data.

I mention this now in case we want applications to receive traits in addition to the list of identities that make up a cohort. And if we're doing that, it might be worth designing it in a way that could support two-way trait sync in the future.

It's not yet clear to me whether two-way trait sync is desirable because of all the complications that brings - at the very least it does not belong as part of the MVP. I can think of some reasons to implement it later:

* The Flagsmith frontend will show the actual trait and segment values for all identities. This makes using the frontend more desirable and adds more value to each user seat.
* Allows server-side SDKs to exclusively use local flag evaluation, which simplifies implementation and mitigates concerns about API call limits in SaaS. This eliminates an otherwise critical design decision that customers need to make early on about their Flagsmith implementation.

matthewelwell:
Discussing with @khvn26, we think that one way that we could achieve this by accepting a list of identifiers, upserting them in the background and adding a 'hidden trait' and segment to be able to segment those users. 

[...]

Issue #5360, opened 2025-04-21
Handle large segments in LaunchDarkly importer


khvn26:
We should detect imported segment condition values for the IN operator over 2000 characters and break them down to smaller OR-concatenated conditions. In SaaS, even after performing the remediation, be conscious of the 400kb limit and surface a warning when it’s hit.

khvn26:
This is now done in #5519 (except the SaaS part)

PR #5519, opened 2025-06-02
feat(launch-darkly-importer): Support large segments import, other improvements

khvn26:

## Changes

Contributes to #5360.

This PR adds the following improvements to LD imports:

### Create multiple conditions instead of one if values exceed character limit

For large user lists that get imported as `IN` segment conditions, and for otherwise large condition lists, we now make sure that they don't exceed the application's segment condition character limit. If they do, we break them down into smaller `OR`-concatenated conditions. If some of them still exceed the limit, e.g. due to a single list element being longer than the limit, we skip those altogether and soft-fail the import request (see `incomplete` status below).

### Add `incomplete` import request status for rate limits and soft failures. Don't retry the import task on hard failures

Currently, the `failed` status is set on all failures, wiping the user-provided LD token, thus breaking any subsequent tries. We now have an `incomplete` status that's used when:
1. The import request processing was rate limited by LD, and signals that the import request can be retried.
2. Some of the imported entities were knowingly skipped due to Flagsmith platform limitations.

The hard-failed import requests with `failed` status will not get retried by the Task Processor anymore, and the soft-failed rate limited ones will get properly backed off.

Added unit tests for the new behaviour.


api/integrations/launch_darkly/services.py
```diff
@@ -313,7 +328,17 @@ def _create_segment_rule_for_segment(
...
+                if len(value) > settings.SEGMENT_CONDITION_VALUE_LIMIT:
+                    _log_error(
+                        import_request=import_request,
+                        error_message=(
+                            f"Segment condition value '{truncate(value)}' for property '{_property}' exceeds the limit of"
+                            f" {settings.SEGMENT_CONDITION_VALUE_LIMIT} characters,"
+                            f" skipping for segment '{segment.name}'"
+                        ),
+                    )
+                    continue
```
emyller, line 340: Do you think this might impact trust in the import process? I am not sure how important this is nor have any better idea right now.

khvn26: There's no way around this until we close #2771, unfortunately, and we want to unblock the customer ASAP.

api/tests/unit/integrations/launch_darkly/test_services.py
```diff
@@ -542,3 +548,65 @@ def test_process_import_request__rules_imported(  # type: ignore[no-untyped-def]
...
+        "Large User List (Override for production)",
+    ]
+    ld_client_class_mock.return_value.get_segments.return_value = json.load(
+        open(
+            join(
+                dirname(abspath(__file__)),
+                "client_responses/get_segments__large_segments.json",
+            )
+        )
+    )
```
emyller: nit: I think dangling file pointers might be irrelevant in tests, though I think `pathlib` fits well within the code style:

```suggestion
    json_path = Path(__file__) / "../client_responses/get_segments__large_segments.json"
    ld_client_class_mock().get_segments.return_value = json.loads(json_path.read_text())
```

api/tests/unit/integrations/launch_darkly/test_services.py
```diff
@@ -542,3 +548,65 @@ def test_process_import_request__rules_imported(  # type: ignore[no-untyped-def]
...
+
+    # When
+    process_import_request(import_request)
+
+    # Then
+    assert (
+        json.dumps(import_request.status, indent=2)
```
emyller: I think `sort_keys=True` might prevent occasional flakiness here.

emyller (approved):
LGTM with non-blocking comments.

api/tests/unit/integrations/launch_darkly/test_services.py
khvn26: Improved in 6e10f360ea63e390dfb83725d97b47596ae49267.

api/tests/unit/integrations/launch_darkly/test_services.py
khvn26: Fixed in f275d1e34b06f83f6fa9f8b74d73e7e74c5f20df.

emyller (approved):
✨

Issue #5270, opened 2025-03-31
As a dashboard user, I can define the trigger on a release pipeline stage to be 'Wait a period of time'


matthewelwell:
Current concerns / questions: 

 1. We want to be able to automate a 'circuit breaker' functionality (in the future) which could revert the previous change that occurred. The concern is that, if this is a trigger that happens _before_ the action, then it's unclear what we should revert? The previous stage? 

Based on this, should we have a concept of pre-post triggers? Should 'Wait a period of time' actually be an action (and stages can have multiple actions?). If so, what would be the point of the 'initial actions' in that case. 

example of what a pre/post trigger set up might look like

<img width="626" alt="Image" src="https://github.com/user-attachments/assets/1dab4814-c32c-4bec-a13a-3ac45c27b5dd" />

gagantrivedi:
We have decided to move pre- and post-triggers to the pipeline level (to be implemented in the next phase). This will then become a simple trigger: when a flag is added to the timeline, wait for an X amount of time.

PR #5523, opened 2025-06-03
feat: Add wait for trigger to release pipelines

tiagoapolo:

## Changes

Implements WAIT_FOR trigger
Ref: https://github.com/Flagsmith/flagsmith/issues/5270

Screenshot 2025-06-03 at 01 42 57

⚠️ Tested with branch containing backend and frontend `release-wait-trigger-backend-ui`

### DIFF
--- a/frontend/common/types/requests.ts
+++ b/frontend/common/types/requests.ts
@@ -78,11 +78,16 @@ export interface StageActionRequest {
   action_body: { enabled: boolean; segment_id?: number }
 }
 
+export type ReleasePipelineRequest = {
+  project: number
+  name: string
+  description?: string
+  stages: PipelineStageRequest[]
+}
+
 export type PipelineStageRequest = {
   name: string
-  project: number
   environment: number
-  pipeline: number
   order: number
   trigger: StageTrigger
   actions: StageActionRequest[]
@@ -698,11 +703,7 @@ export type Req = {
   updateOnboarding: Partial<Onboarding>
   getReleasePipelines: PagedRequest<{ projectId: number }>
   getReleasePipeline: { projectId: number; pipelineId: number }
-  createReleasePipeline: {
-    projectId: number
-    name: string
-    status: PipelineStatus
-  }
+  createReleasePipeline: ReleasePipelineRequest
   getPipelineStages: PagedRequest<{
     projectId: number
     pipelineId: number
@@ -712,7 +713,6 @@ export type Req = {
     pipelineId: number
     stageId: number
   }
-  createPipelineStage: PipelineStageRequest
   deleteReleasePipeline: {
     projectId: number
     pipelineId: number
--- a/frontend/web/components/release-pipelines/CreatePipelineStage.tsx
+++ b/frontend/web/components/release-pipelines/CreatePipelineStage.tsx
@@ -3,20 +3,26 @@ import { useGetEnvironmentsQuery } from 'common/services/useEnvironment'
 import InputGroup from 'components/base/forms/InputGroup'
 import Utils from 'common/utils/utils'
 import { useGetSegmentsQuery } from 'common/services/useSegment'
-import { StageActionType, StageTrigger } from 'common/types/responses'
+import {
+  StageActionType,
+  StageTrigger,
+  StageTriggerType,
+} from 'common/types/responses'
 import Button from 'components/base/forms/Button'
 import Icon from 'components/Icon'
-import { FLAG_ACTIONS, TRIGGER_OPTIONS } from './constants'
-import { StageActionRequest, PipelineStageRequest } from 'common/types/requests'
+import {
+  FLAG_ACTION_OPTIONS,
+  TIME_UNIT_OPTIONS,
+  TRIGGER_OPTIONS,
+} from './constants'
+import { StageAction } from 'common/types/responses'
+import moment from 'moment'
+import Input from 'components/base/forms/Input'
+import { PipelineStageRequest } from 'common/types/requests'
 
-type DraftStageType = Omit<PipelineStageRequest, 'id' | 'pipeline' | 'project'>
+type DraftStageType = PipelineStageRequest
 
-const getFlagActions = (trigger: string | undefined) => {
-  if (trigger === 'ON_ENTER') {
-    return FLAG_ACTIONS.ON_ENTER
-  }
-  return []
-}
+type TimeUnit = 'days' | 'hours' | 'minutes'
 
 const CreatePipelineStage = ({
   onChange,
@@ -32,6 +38,8 @@ const CreatePipelineStage = ({
   onRemove?: () => void
 }) => {
   const [searchInput, setSearchInput] = useState('')
+  const [amountOfTime, setAmountOfTime] = useState(1)
+  const [selectedTimeUnit, setSelectedTimeUnit] = useState<TimeUnit>('days')
   const [selectedAction, setSelectedAction] = useState<{
     label: string
     value: string
@@ -83,9 +91,21 @@ const CreatePipelineStage = ({
 
   const handleOnChange = (
     fieldName: keyof DraftStageType,
-    value: string | number | StageTrigger | StageActionRequest[],
+    value: string | number | StageTrigger | Omit<StageAction, 'id'>[],
   ) => {
     onChange({ ...stageData, [fieldName]: value })
+  }
+
+  function formatDurationToHHMMSS(duration: moment.Duration) {
+    const totalSeconds = duration.asSeconds()
+
+    const hours = Math.floor(totalSeconds / 3600)
+    const minutes = Math.floor((totalSeconds % 3600) / 60)
+    const seconds = Math.floor(totalSeconds % 60)
+
+    return `${hours}:${String(minutes).padStart(2, '0')}:${String(
+      seconds,
+    ).padStart(2, '0')}`
   }
 
   useEffect(() => {
@@ -110,7 +130,7 @@ const CreatePipelineStage = ({
         return action
       })
 
-      handleOnChange('actions', actions as StageActionRequest[])
+      handleOnChange('actions', actions)
     }
   }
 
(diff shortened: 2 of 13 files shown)


frontend/web/components/release-pipelines/CreatePipelineStage.tsx
```diff
@@ -192,6 +212,34 @@ const CreatePipelineStage = ({
...
+                const value = Utils.safeParseEventValue(e)
+                setAmountOfTime(Number(value) || 1)
+              }}
+              type='number'
+              placeholder='Amount of time'
+            />
+          </div>
+          <div className='w-50'>
+            <Select
+              value={Utils.toSelectedValue(selectedTimeUnit, TIME_UNIT_OPTIONS)}
```
kyle-ssg, line 279: Ah cool first time seeing toSelectedValue, great idea

frontend/common/types/requests.ts
```diff
@@ -78,11 +78,16 @@ export interface StageActionRequest {
   action_body: { enabled: boolean; segment_id?: number }
 }
 
+export type ReleasePipelineRequest = {
```
Zaimwa9, line 81:
```suggestion
export interface ReleasePipelineRequest = {
```

frontend/web/components/release-pipelines/CreatePipelineStage.tsx
```diff
@@ -134,6 +154,40 @@ const CreatePipelineStage = ({
...
+          }
+        }
+        return action
+      })
+
+      handleOnChange('actions', actions as StageAction[])
+    }
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [selectedSegment])
```
Zaimwa9: I think this is safe to remove as done in the `handleSegmentChange`

frontend/web/components/release-pipelines/CreatePipelineStage.tsx
```diff
@@ -134,6 +154,40 @@ const CreatePipelineStage = ({
...
+
+  useEffect(() => {
```
Zaimwa9: This could be moved in its handler too for better control

frontend/web/components/release-pipelines/CreatePipelineStage.tsx
```diff
@@ -196,6 +251,34 @@ const CreatePipelineStage = ({
...
+              onChange={(e: React.ChangeEvent<HTMLInputElement>) => {
+                const value = Utils.safeParseEventValue(e)
+                setAmountOfTime(Number(value) || 1)
+              }}
+              type='number'
+              placeholder='Amount of time'
+            />
+          </div>
+          <div className='w-50'>
+            <Select
```
Zaimwa9, line 278: Thanks for having fixed the dropdown, it's way smoother

frontend/web/components/release-pipelines/CreateReleasePipeline.tsx
```diff
@@ -123,45 +96,23 @@ function CreateReleasePipeline({
...
-      })
-
-      setStagesCreationSuccess(false)
-      try {
-        await Promise.all(stagesCreationPromises)
-        setStagesCreationSuccess(true)
-      } catch (error) {
-        toast('Error creating release stages', 'error')
-      }
-    }
```
Zaimwa9, line 159: 🚀

frontend/web/components/release-pipelines/CreateReleasePipeline.tsx
```diff
@@ -53,38 +37,27 @@ function CreateReleasePipeline({
...
-  const isPipelineSuccess = isCreatingPipelineSuccess && !stages.length
-  const isStagesSuccess =
-    isCreatingPipelineSuccess && stages.length && isStagesCreationSuccess
-
   useEffect(() => {
-    if (isPipelineSuccess || isStagesSuccess) {
+    if (isCreatingPipelineSuccess) {
       return handleSuccess()
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
```
Zaimwa9: I checked a bit how to cleanly deal with this. It seems wrapping handleSuccess in a callback would stabilize the reference and allow to use it as a dep to calm down react while avoiding inifinite loop. Worth a try you think ?

frontend/web/components/release-pipelines/CreateReleasePipeline.tsx
```diff
@@ -96,10 +69,10 @@ function CreateReleasePipeline({
   }, [isCreatingPipelineError, createPipelineError])
 
   const handleOnChange = (newStageData: DraftStageType, index: number) => {
-    const updatedStages = stages.map((stage, i) =>
+    const updatedStages = pipelineData.stages.map((stage, i) =>
```
Zaimwa9, line 71:
```suggestion
    const updatedStages = pipelineData.stages?.map((stage, i) =>
```
Is there some cases when stages can be empty ? Or will it be `blankStages` ?

frontend/web/components/release-pipelines/FlagActionDetail.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  if (actionType === StageActionType.TOGGLE_FEATURE) {
+    return (
+      <span>
+        {actionBody.enabled ? 'Enable' : 'Disable'} flag for <b>everyone</b>
+      </span>
+    )
+  }
+
+  return null
```
Zaimwa9: Optional: As the actions might be extended, I wonder if a function switch renderContent: `switch(actionType) , default null` could be clearer in terms of render paths

frontend/web/components/release-pipelines/ReleasePipelineDetail.tsx
```diff
@@ -2,23 +2,15 @@ import Breadcrumb from 'components/Breadcrumb'
 
 import PageTitle from 'components/PageTitle'
 
-import {
-  useGetPipelineStagesQuery,
-  useGetReleasePipelineQuery,
-} from 'common/services/useReleasePipelines'
-import { withRouter, RouteComponentProps } from 'react-router'
+import { useGetReleasePipelineQuery } from 'common/services/useReleasePipelines'
+import { useParams } from 'react-router'
```
Zaimwa9:
```suggestion
import { useParams } from 'react-router-dom'
```
We should import from react-router-dom. It wraps react-router specifically for DOM usage. Using react-router can cause some sneaky issues as it's supposed to be platform agnostic. Pretty rare I guess but worth avoiding

frontend/web/components/release-pipelines/CreateReleasePipeline.tsx
```diff
@@ -5,28 +5,12 @@ import { Button } from 'components/base/forms/Button'
...
-} from 'common/types/responses'
+import { StageActionType, StageTriggerType } from 'common/types/responses'
 import Icon from 'components/Icon'
-import {
-  useCreatePipelineStagesMutation,
-  useCreateReleasePipelineMutation,
-} from 'common/services/useReleasePipelines'
-import { withRouter, RouteComponentProps } from 'react-router'
+import { useCreateReleasePipelineMutation } from 'common/services/useReleasePipelines'
+import { useHistory, useParams } from 'react-router'
```
Zaimwa9:
```suggestion
import { useHistory, useParams } from 'react-router-dom'
```

frontend/web/components/release-pipelines/StageFeatureDetail.tsx
```diff
@@ -0,0 +1,55 @@
...
+  useEffect(() => {
+    getProjectFlags()
+  }, [getProjectFlags])
+
+  return (
+    <>
+      <h6>Features ({projectFlags.length ?? 0})</h6>
+      {!features.length && (
+        <p className='text-muted'>No features added to this stage yet.</p>
+      )}
```
Zaimwa9: This might be conflicting with the above line where we see the both the 0 and the text no ? Can we move this in an early return?

frontend/web/components/release-pipelines/constants.ts
```diff
@@ -1,19 +1,27 @@
...
-    { label: 'Enable flag for everyone', value: 'TOGGLE_FEATURE' },
-    { label: 'Disable flag for everyone', value: 'TOGGLE_FEATURE_DISABLE' },
-    { label: 'Enable flag for segment', value: 'TOGGLE_FEATURE_FOR_SEGMENT' },
-    {
-      label: 'Disable flag for segment',
-      value: 'TOGGLE_FEATURE_DISABLE_FOR_SEGMENT',
-    },
-  ],
-}
+const FLAG_ACTION_OPTIONS = [
```
Zaimwa9, line 11: If this is `{ label: string; value: StageActionType }[] `

Shouldn't we add the 2 disable in the enum ?

```
export enum StageActionType {
  TOGGLE_FEATURE = 'TOGGLE_FEATURE',
  TOGGLE_FEATURE_FOR_SEGMENT = 'TOGGLE_FEATURE_FOR_SEGMENT',
}
```

frontend/web/components/release-pipelines/StageFeatureDetail.tsx
```diff
@@ -0,0 +1,55 @@
...
+            project: projectId,
+          }),
+        ),
+      )
+      const projectFlags = res.map((r) => r.data)
+      setProjectFlags(projectFlags)
+    } catch (error) {
+      console.error(error)
+    }
+  }, [features, projectId])
```
Zaimwa9: Not 100% but since features is an array, its reference might change even if values don't could cause the callback to re-run. Maybe we could use features.join(',') to avoid it

Zaimwa9 (changes requested):
looks good overall, couple of non-critical comments

frontend/web/components/release-pipelines/ReleasePipelineDetail.tsx
tiagoapolo: yes, the IDE did not suggest the right import

frontend/web/components/release-pipelines/CreateReleasePipeline.tsx
tiagoapolo: it won't be empty

frontend/web/components/release-pipelines/StageFeatureDetail.tsx
tiagoapolo: hmmm you're right, sometimes we forget about shallow rendering

Zaimwa9 (approved):
👍

PR #5531, opened 2025-06-03
feat: Integrate with Sentry feature flag Change Tracking

emyller:
Read more: https://docs.sentry.io/product/issues/issue-details/feature-flags/

---

## Changes

- Introduce a new endpoint to set up a Sentry feature flags webhook.
- Send feature flag updates to Sentry (Change Tracking).

- Unit tests.
- Integration tests.
- Manual tests with a local web app, a Flagsmith account, and a Sentry account.


api/integrations/sentry/change_tracking.py
emyller, line 1: Reviewing note: the `SentryChangeTracking` class doesn't quite follow examples of other integrations strictly because of architecture. The existing pattern supports API integrations (base URL + key), whereas Sentry does webhooks. There's some similarity, and that was leveraged I believe. I'm also aware we have our own webhooks but I think this is best separated.

api/integrations/sentry/models.py
```diff
@@ -0,0 +1,30 @@
+from django.core import validators
+from django.db import models
+
+from integrations.common.models import EnvironmentIntegrationModel
+
+
+class SentryChangeTrackingConfiguration(EnvironmentIntegrationModel):
```
khvn26, line 7: This is not a comment to address, rather a note to myself.

From [Sentry's webhook docs](https://github.com/getsentry/sentry/blob/ed324708947ca26392d6579ed76b93fc94a61ab6/src/sentry/flags/docs/api.md#create-generic-flag-log-post):

> Sentry does not currently have a concept of disambiguating flag changes by project or environment. Everything is done at the organization level.

Because of the above, I think the decision to implement the integration at environment level makes sense.

api/audit/signals.py
```diff
@@ -71,19 +73,31 @@ def call_webhooks(sender, instance, **kwargs):  # type: ignore[no-untyped-def]
...
+
+
+def track_only_feature_related_events(signal_function):  # type: ignore[no-untyped-def]
+    allowed_types = [
+        RelatedObjectType.FEATURE,
+        RelatedObjectType.FEATURE_STATE,
+        RelatedObjectType.SEGMENT,
+        RelatedObjectType.EF_VERSION,
+    ]
+    return track_only(allowed_types)(signal_function)
```
khvn26: Both decorators can be relatively painlessly type annotated. I appreciate a lot of time has been spent on this, so leaving this comment as a nit.

api/audit/signals.py
```diff
@@ -166,3 +180,22 @@ def send_audit_log_event_to_slack(sender, instance, **kwargs):  # type: ignore[n
         api_token=slack_project_config.api_token, channel_id=env_config.channel_id
     )
     _track_event_async(instance, slack)  # type: ignore[no-untyped-call]
+
+
+@receiver(post_save, sender=AuditLog)
+@track_only([RelatedObjectType.FEATURE_STATE])
```
khvn26, line 206: This means we only track vanilla environment feature states, excluding:

- Versions published in versioning environments – not sure if such changes are covered by `RelatedObjectType.FEATURE_STATE` audit logs, @matthewelwell?
- Segments and segment overrides — we might be not interested in those as Sentry change tracking is only intended for global flags at the moment.
- Feature creation and deletion — these _might_ be covered by feature state logs as we probably generate them for cascading deletions.

I guess points 2 and 3 are addressed by the docstring for `SentryChangeTracking`, so this comment boils down to asking whether we should track feature versions or not.

api/integrations/sentry/change_tracking.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def __init__(self, webhook_url: str, secret: str) -> None:
+        self.webhook_url = webhook_url
+        self.secret = secret
+
+    @staticmethod
+    def generate_event_data(audit_log_record: AuditLog) -> dict[str, Any]:
+        feature_state: FeatureState = get_audited_instance_from_audit_log_record(  # type: ignore[assignment]
+            audit_log_record,
+        )
```
khvn26: I'd rather prefer a `if instance and isinstance(instance, FeatureState): # pragma: no cover` branch than a forced type annotation here. If we add some other tracked entities to the `track_only` decorator forgetting to update code in here, this `type: ignore` will mask a bug.

khvn26 (commented):
Looks good overall with only a couple comments/questions to address 👍

api/integrations/sentry/models.py
emyller: That's correct. Sentry support for feature flags is currently in open beta, so this might change to a more granular setup, hopefully.

api/audit/signals.py
emyller: > Versions published in versioning environments – not sure if such changes are covered by RelatedObjectType.FEATURE_STATE audit logs (...)

I did not cover this because I was unaware of the use case. 🥲 Thanks for shedding light on the topic. If feature versions affect an environment's flag state globally, then yes they should be pushed as _updates_ to Sentry.

> I guess points 2 and 3 are addressed by the docstring for SentryChangeTracking, (...)

Positive. TBH normally the architecture would bother me a bit because this relies on signals, which relies on hooks, which relies on assumptions sometimes. However, because this design is well-established as a pattern throughout the code, and also very well covered with both unit, integration, and manual tests, I'm fairly comfortable with the patch. I did add some extra docstring and pointers around — _and they already proved useful._ 🙂

You're right: segments and other overrides are not accounted for in Change Tracking. The other integration, Evaluation Tracking, should give a clearer view of the actual boolean flags present in the context of an exception.

api/integrations/sentry/change_tracking.py
emyller: Yeah the [test failure](https://github.com/Flagsmith/flagsmith/actions/runs/15598670572/job/43934472200?pr=5531) might prove this branching necessary. Thanks for the pointer!

emyller: Fixed with e5264e8c990f7b7efeba2d35a84dc70a3e3dc125.

api/audit/signals.py
emyller: Fixed with b9e6991cc852ffcfe1a6393c2234e89e6656a089.
P.S. _relatively_ is indeed relative.

api/audit/signals.py
emyller: FTR @matthewelwell and I discussed about feature versioning. We decided to handle it separately with a new issue because of the added complexity. For now this patch should already cover most use cases I believe.

api/integrations/sentry/change_tracking.py
khvn26: It is fixed in the `send_feature_flag_went_live_signal` body, but I'd like it fixed here in `generate_event_data` as well.

emyller: It took me a while to realize I wrote this code twice. 👴
3862258622e7ee4c6cc363ac5a67ac6105d8039d

api/audit/signals.py
```diff
@@ -121,7 +122,11 @@ def track_only_feature_related_events(signal_function):  # type: ignore[no-untyp
 
 
 def _track_event_async(instance, integration_client):  # type: ignore[no-untyped-def]
-    event_data = integration_client.generate_event_data(audit_log_record=instance)
+    try:
+        event_data = integration_client.generate_event_data(audit_log_record=instance)
+    except InvalidIntegrationParameters as error:  # pragma: no cover
+        logger.error("Cannot prepare integration: %s", repr(error))
```
khvn26: We're going to have a bunch of error logs during normal integration operation. Consider changing this to `info`/`warning`.

api/integrations/sentry/change_tracking.py
```diff
@@ -36,7 +37,9 @@ def generate_event_data(audit_log_record: AuditLog) -> dict[str, Any]:
         if not (
             feature_state and isinstance(feature_state, FeatureState)
         ):  # pragma: no cover
-            raise ValueError(f"{repr(feature_state)} is not a FeatureState object.")
+            raise InvalidIntegrationParameters(
+                f"{repr(feature_state)} is not a FeatureState object."
+            )
```
khvn26: 1. The error message is not helpful enough. I'd prefer something among the lines of `f"{type(audit_log)} is not yet supported by Sentry Change Tracking integration`.
2. Reiterating on [my other comment](https://github.com/Flagsmith/flagsmith/pull/5531/commits/c2af35844be315c6b56fb2fd783e1ee1818c3e91#r2150035581), I don't think this branch necessarily means an error, so not sure if we should raise an exception here? No strong opinion held, though.

docs/docs/integrations/error-tracking/_category_.json
```diff
@@ -0,0 +1,5 @@
+{
+    "label": "Error Tracking",
```
rolodato: This category is Error Tracking but the linked documentation is about Change Tracking and Evaluation Tracking. Is that intended or am I misunderstanding?

emyller: @rolodato Valid point. The label "Error Tracking" is used to categorize tools such as Sentry, given _error tracking_ is their thing. Within Sentry, Change Tracking and Evaluation Tracking are tools to help identifying flag changes as culprits for errors. Do you think of a better label?

api/audit/signals.py
emyller: > We're going to have a bunch of error logs during normal integration operation.

We shouldn't, as this only happens if there's a `SentryChangeTrackingConfiguration` object is found, but the integration doesn't work out.

This raises an interesting question: for a self-hosted Flagsmith install, I'd really want this to be louder than _warning_. Now for SaaS, I think a `warning` log would sound better. Which should we favor?

docs/docs/integrations/error-tracking/_category_.json
rolodato: Would the existing Observability category be a better fit? Sentry is for sure an observability tool, and you would integrate Change/Evaluation Tracking to improve the observability of your feature flags.

emyller: Fixed with 8ca6d092d695b3f188b1c6c65fb2e9d1d86f7959.

api/integrations/sentry/change_tracking.py
emyller: Fixed with 3439821b7114909a14179f856d1ccb798743056c.

khvn26 (approved):
LGTM 👍

PR #5536, opened 2025-06-04
fix(launch-darkly-importer): Seconds `X-Ratelimit-Reset` timestamps are expected for rate limited requests

khvn26:

## Changes

Currently, we interpret `X-Ratelimit-Reset` header values as seconds, which is not correct — they are in fact being returned as milliseconds, see [LD docs](https://launchdarkly.com/docs/api#global-rate-limits).

Verified that millisecond timestamps are returned by manually triggering the rate limit.

Updated unit tests.


emyller (approved):
That was a lot of waiting. 😂

Issue #5301, opened 2025-04-04
As a dashboard user, I am able to publish a release pipeline


matthewelwell:
Publishing a workflow sets the status of the workflow to 'published' which will mean it shows in the list of available release pipelines to add a flag to.

Need an endpoint like the following:

```
POST /.../pipelines/:id/publish/

 + response 200
{
  "id": 1,
  ...
  "status": "published"
}
```

Issue #5275, opened 2025-03-31
As a dashboard user, when viewing a flag in a given environment, I can add it to a release pipeline


matthewelwell:
TODO

Considerations:  
 * Adding a flag to a release pipeline should lock the flag for editing. Users should have to remove it from the release pipeline before making manual changes to the flag. 
 * Should we only allow flags to be added to release pipelines for which the first stage is for the current environment the user is viewing the flag in? 
     * Alternatives: 
         * Have a view of a feature at the project level
         * Just allow them to add flags to any release pipeline, regardless of the environment (but perhaps show a warning when adding to a release pipeline for which the first stage is not in the current environment)
             * Agreed that this is the best option.

gagantrivedi:
> Adding a flag to a release pipeline should lock the flag for editing

In all environments?

matthewelwell:
> > Adding a flag to a release pipeline should lock the flag for editing
> 
> In all environments?

Good question. This depends on if we think we can allow pipelines to work across multiple environments or not?

gagantrivedi:
> This depends on if we think we can allow pipelines to work across multiple environments or not?

The Figma assumes that we do... it will be very limited otherwise?

matthewelwell:
Yes, but it was one of the things we though about restricting in the first version. I'm happy to include it in the first version if we think that we can technically achieve it. In which case, it should lock it in all environments, yes.

matthewelwell:
As discussed, since we're allowing release pipelines to span multiple environments, we should lock the flag in all environments.

PR #5542, opened 2025-06-05
feat: Release Pipeline Publish and Add Flag

tiagoapolo:

## Changes

Ref: https://github.com/Flagsmith/flagsmith/issues/5301
[#5275](https://github.com/Flagsmith/flagsmith/issues/5275)

Implements:
- Publishing release pipeline, changes release pipeline from draft to published
- Adding a feature flag to published pipeline

Screenshot 2025-06-05 at 10 13 35
Screenshot 2025-06-05 at 10 13 54
Screenshot 2025-06-05 at 10 14 17
Screenshot 2025-06-05 at 10 14 43

⚠️ Tested using branch containing backend and frontend `release-pipe-publish-back-front`


frontend/web/components/modals/CreateFlag.js
```diff
@@ -1071,6 +1083,13 @@ const CreateFlag = class extends Component {
                 )
               const { featureError, featureWarning } = this.parseError(error)
 
+              const isReleasePipelineEnabled =
+                Utils.getFlagsmithHasFeature('release_pipelines')
+              // TODO: Check correct permission
```
tiagoapolo: @matthewelwell which would be the best plans permission to use? Or even use one?

frontend/web/components/release-pipelines/CreatePipelineStage.tsx
```diff
@@ -134,6 +154,35 @@ const CreatePipelineStage = ({
...
+    } as StageTrigger)
+  }
+
+  const handleTriggerChange = (option: { value: string; label: string }) => {
+    if (option.value === StageTriggerType.WAIT_FOR) {
+      const time = 1
+      const unit = 'days'
+
+      setAmountOfTime(time)
+      setSelectedTimeUnit(unit)
```
Zaimwa9: The set amount of time is broken in this version (maybe due to local conflicts).

This should be removed here and added in setWaitForTrigger

frontend/web/components/release-pipelines/CreatePipelineStage.tsx
```diff
@@ -134,6 +154,35 @@ const CreatePipelineStage = ({
     handleOnChange('actions', [{ action_body, action_type }])
   }
 
+  const setWaitForTrigger = (time: number, unit: TimeUnit) => {
+    const duration = moment.duration(time, unit)
+    const formatted = formatDurationToHHMMSS(duration)
+
```
Zaimwa9, line 163: Should update local states:
setAmountOfTime and setSelectedTimeUnit

frontend/web/components/release-pipelines/CreateReleasePipeline.tsx
```diff
@@ -53,38 +37,26 @@ function CreateReleasePipeline({
...
 
-  const [createStages] = useCreatePipelineStagesMutation()
-
-  const [pipelineData, setPipelineData] = useState<DraftPipelineType>({
+  const [pipelineData, setPipelineData] = useState<ReleasePipelineRequest>({
     name: '',
     project: Number(projectId),
-    status: PipelineStatus.DRAFT,
+    stages: [blankStage],
   })
```
Zaimwa9: There's a little problem with having stages and the pipeline high-level data in the same state here is that it re-renders all the stages states/related components when we update one or the other.

It's not dramatic given that the data is not super big but if we want to have it clean from here we could:
- separate local states stages from name and project
- and/or memoize `react.memo(CreatePipelineStage)` but it would need a specific isEqual callback (memoizing can be sufficient)

frontend/web/components/modals/CreateFlag.js
```diff
@@ -1163,28 +1176,49 @@ const CreateFlag = class extends Component {
...
-                                              identity,
+                                    <div className='text-right'>
+                                      <Permission
+                                        level='environment'
+                                        permission={'UPDATE_FEATURE_STATE'}
+                                        id={this.props.environmentId}
+                                      >
+                                        {({
+                                          permission: updateFeatureState,
+                                        }) => {
```
Zaimwa9: Need to inline return, as of the button doesn't appear

frontend/web/components/release-pipelines/CreateReleasePipeline.tsx
```diff
@@ -123,45 +95,23 @@ function CreateReleasePipeline({
...
-      } catch (error) {
-        toast('Error creating release stages', 'error')
-      }
-    }
   }
 
   const handleRemoveStage = (index: number) => {
-    const newStages = stages.filter((_, i) => i !== index)
-    setStages(newStages)
+    const newStages = pipelineData.stages.filter((_, i) => i !== index)
```
Zaimwa9: The problem with a full filtere here is that if we update a previous stage it removes all the following

frontend/web/components/release-pipelines/AddToReleasePipelineModal.tsx
```diff
@@ -0,0 +1,102 @@
...
+      return
+    }
+  }, [isSuccess, isError, error])
+
+  return (
+    <div className='p-4'>
+      <InputGroup
+        title='Release Pipeline'
+        component={
+          <Select
```
Zaimwa9, line 79: The backend correctly returns an error but I think we should disable the draft pipelines or have a tag to show they are not available
image

frontend/web/components/release-pipelines/AddToReleasePipelineModal.tsx
```diff
@@ -0,0 +1,102 @@
...
+  useAddFeatureToReleasePipelineMutation,
+  useGetReleasePipelinesQuery,
+} from 'common/services/useReleasePipelines'
+import { ReleasePipeline } from 'common/types/responses'
+import Utils from 'common/utils/utils'
+import Button from 'components/base/forms/Button'
+import InputGroup from 'components/base/forms/InputGroup'
+
+type AddToReleasePipelineModalProps = {
+  projectId: string
```
Zaimwa9: Should be a number here

frontend/web/components/release-pipelines/AddToReleasePipelineModal.tsx
```diff
@@ -0,0 +1,102 @@
...
+const AddToReleasePipelineModal = ({
+  featureId,
+  projectId,
+}: AddToReleasePipelineModalProps) => {
+  const [selectedReleasePipeline, setSelectedReleasePipeline] =
+    useState<string>('')
+  const { data: releasePipelines } = useGetReleasePipelinesQuery({
+    page: 1,
+    page_size: 100,
+    projectId: Number(projectId),
```
Zaimwa9: Same, I think we need to start typing as a number so it is surfaced

frontend/web/components/release-pipelines/AddToReleasePipelineModal.tsx
```diff
@@ -0,0 +1,102 @@
...
+  projectId: string
+  featureId: number
+}
+
+const AddToReleasePipelineModal = ({
+  featureId,
+  projectId,
+}: AddToReleasePipelineModalProps) => {
+  const [selectedReleasePipeline, setSelectedReleasePipeline] =
+    useState<string>('')
```
Zaimwa9: Same here, as it's supposed to be a number we should type it as of

frontend/web/components/release-pipelines/ReleasePipelineDetail.tsx
```diff
@@ -2,25 +2,24 @@ import Breadcrumb from 'components/Breadcrumb'
...
+import { useGetReleasePipelineQuery } from 'common/services/useReleasePipelines'
+import { useParams } from 'react-router-dom'
 import { useGetEnvironmentsQuery } from 'common/services/useEnvironment'
 import Icon from 'components/Icon'
 import StageCard from './StageCard'
 import StageInfo from './StageInfo'
 import { PipelineStage } from 'common/types/responses'
 import { Environment } from 'common/types/responses'
+import StageFeatureDetail from './StageFeatureDetail'
+import Tag from 'components/tags/Tag'
```
Zaimwa9, line 14:
```suggestion
```

frontend/common/types/requests.ts
```diff
@@ -694,15 +698,14 @@ export type Req = {
     userId: number | undefined
     level: PermissionLevel
   }
-  getProfile: {}
+  getProfile: {
+    id?: number
+  }
+  getUser: { id: number }
```
Zaimwa9, line 704: I believe it's not used in final version

frontend/web/components/modals/CreateFlag.js
```diff
@@ -1163,28 +1176,49 @@ const CreateFlag = class extends Component {
...
+                                          isReleasePipelineEnabled &&
+                                            updateFeatureState && (
+                                              <Button
```
Zaimwa9: We can disable if there is no pipeline (but do we have the data here ?)

frontend/web/components/release-pipelines/StageFeatureDetail.tsx
```diff
@@ -47,13 +58,21 @@ const StageFeatureDetail = ({
     )
   }
 
+
   return (
```
Zaimwa9: For the above text wdyt about:
Wdyt about "No features at this stage" ? As it goes back to this state after features completed the stage

Zaimwa9 (changes requested):
Overall well working along with the backend (tested the pipeline entirely, with timers and different actions).
A couple of regression though and remarks

frontend/web/components/release-pipelines/CreatePipelineStage.tsx
```diff
@@ -154,22 +154,6 @@ const CreatePipelineStage = ({
     handleOnChange('actions', [{ action_body, action_type }])
   }
```
Zaimwa9, line 156: ⚠️ Important one there is a big bug here:
When selecting a segment, it sets enabled to `true` so disable segment for flag action is not working as of
```

  const handleSegmentChange = (option: { value: number; label: string }) => {
    setSelectedSegment(option)

    if (option?.value) {
      const actions = stageData.actions.map((action) => {
        if (action.action_type === StageActionType.TOGGLE_FEATURE_FOR_SEGMENT) {
          return {
            ...action,
            action_body: { enabled: true, segment_id: option?.value },
          }
        }
        return action
      })

      handleOnChange('actions', actions)
    }
  }
```

Zaimwa9 (commented):
Found a bug actually while testing out

frontend/web/components/release-pipelines/AddToReleasePipelineModal.tsx
tiagoapolo: I forgot to add a check here.

frontend/web/components/modals/CreateFlag.js
```diff
@@ -229,6 +232,18 @@ const CreateFlag = class extends Component {
...
+    getReleasePipelines(getStore(), {
+      projectId: this.props.projectId,
+    }).then((res) => {
+      const pipelines = res.data.results
+      const hasPublishedReleasePipelines =
+        pipelines?.some((pipeline) => pipeline?.published_by) ?? false
+
+      this.setState({ hasPublishedReleasePipelines })
+    })
+  }
```
Zaimwa9, line 249: Needs a catch to setHasPublishedReleasePipelines to false (otherwise error for non enterprise org)

frontend/web/components/release-pipelines/CreatePipelineStage.tsx
```diff
@@ -124,7 +124,7 @@ const CreatePipelineStage = ({
         if (action.action_type === StageActionType.TOGGLE_FEATURE_FOR_SEGMENT) {
```
Zaimwa9, line 124: Sorry, I'm adding here some comments about something I didn't notice => Non-enteprise orgs.
- The release pipelines nav item is available (I would personally disable it and adding a tooltip to warn it's for enterprise but tbd with product)
- You have access to the whole flow and only get the error at the end
- The failure message is success and not error

image

Zaimwa9: It might be out of this PR scope so maybe we can create a ticket specifically to address the non available state of this feature

Zaimwa9: So we have this in the organisation payload:
```
{
  "id": 2,
  "has_active_billing_periods": true,
  "deleted_at": null,
  "uuid": "xxxx",
  "subscription_id": "123",
  "subscription_date": null,
  "plan": "enterprise",
  "max_seats": 400,
  "max_api_calls": 50000,
  "cancellation_date": null,
  "customer_id": null,
  "billing_status": null,
  "payment_method": null,
  "notes": null
}
```

In the context of release, the backend definition of enterprise is:
`!!subscription_id && !is_cancellation_date && plan === 'enterprise`, to see if it exactly matches the frontend one

frontend/web/components/release-pipelines/CreateReleasePipeline.tsx
Zaimwa9: Re-opening as this one hasn't been resolved

https://www.loom.com/share/81069cfa500747b19e88aef5c5dbf753

Zaimwa9 (changes requested):
Almost good to go, 2 things to solve before ✅ :
- The update of a stage removing the following stages
- The catch on getReleasePipelines

We can create a ticket to properly deal with enterprise

frontend/web/components/release-pipelines/CreatePipelineStage.tsx
tiagoapolo: This can be handled in a separate ticket

Zaimwa9 (approved):
👍

PR #5547, opened 2025-06-05
feat: use-project-id-from-context

Zaimwa9:

## Changes
The objective of this PR is to provide a single source for current projectId/environmentId/organisationId (based on url)

_Please describe._


tiagoapolo:
I tested and it works without any issues. Is the idea using useRouteContext instead of useParams and useRouteMatch ?

Zaimwa9:
Yes that would be the idea. Especially for projectId (less important for environmentID being a string). So it's accessible from only one place instead of having an util or dealing with the type every time

tiagoapolo:
> Yes that would be the idea. Especially for projectId (less important for environmentID being a string). So it's accessible from only one place instead of having an util or dealing with the type every time

Ok, can you create a ticket for replacing every place that's using `useParams` and `useRouteMatch` for pulling projectId, environmentId and organisationId?

tiagoapolo (approved):

Zaimwa9:
Did it directly in this PR but I'll create a ticket to re-type the component using it as a string

PR #5560, opened 2025-06-10
fix: hubspot attribution issues

Zaimwa9:

## Changes
⚠️ Hubspot is now in charge of initially creating (if not exists) / associating organisation with users. Reconciliation is based on domain.

- Pass hubspot cookie from frontend on signup action
- Reworked the hubspot tracking flow so that:
1. Every new sign-up creates a hubspot contact (using `create_lead_form` with or without cookie) and saves `HubspotLeads`
2. On UsersOrganisation creation
    a) If organisation was just created and exists in hubspot => get company data from hubspot, saves the `HubspotOrganisation` object with correct id
    b) if doesn't exist in hubspot for X reason => creates it and saves hubspot ID
3. On UsersOrganisation creation, associates the contact with the 

- Removed the 30min background checker as hubspot creates/associates automatically the company

_Please describe._
- Added battery of tests on client and lead_tracker
Functionally:
Set `"ENABLE_HUBSPOT_LEAD_TRACKING": "True"` and `HUBSPOT_ACCESS_TOKEN`
Create user, organisation, invite users


api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -44,88 +45,134 @@ def should_track(user: FFAdminUser) -> bool:
...
+        self.client.create_lead_form(user=user, hubspot_cookie=tracker_cookie)
+        contact = self.client.get_contact(user)
+        hubspot_contact_id: str | None = contact.get("id")
+
+        # Hubspot creates contact asynchronously
+        # If not available on the spot, following steps will sync database with Hubspot
+        if hubspot_contact_id is not None:
+            HubspotLead.objects.update_or_create(
+                user=user, defaults={"hubspot_id": hubspot_contact_id}
+            )
```
Zaimwa9: Discussed with Matt on this case.
Basically, it's a matter of synchronization. Contact is created, but our `get_contact` request *might* be sent before processed by hubspot. When associating a user with an organisation, there is a fallback to still check his existence in hubspot so that will resynchronize us with hubspot anyways. Only user creating an account and no organization will be "zombies"
As we use the hubspot_id only for association and marketing does not really need it, it's an accepted trade-off

Alternatives were:
- Top notch solution would be webhook but overkill
- Retry mechanism

api/custom_auth/views.py
```diff
@@ -125,6 +128,8 @@ def get_throttles(self):  # type: ignore[no-untyped-def]
 
     def create(self, request: Request, *args: Any, **kwargs: Any) -> Response:
         response = super().create(request, *args, **kwargs)
+        request.user = self.user
+        register_hubspot_tracker(request)
```
matthewelwell: Rather than interfering with the django behaviour here, couldn't we just have `register_hubspot_tracker` accept the user model too?

api/integrations/lead_tracking/hubspot/client.py
```diff
@@ -103,38 +108,48 @@ def create_lead_form(
...
+                            "associationCategory": "HUBSPOT_DEFINED",
+                            "associationTypeId": 1,
+                        }
+                    ],
+                    "to": {"id": hubspot_company_id},
+                }
+            ]
+
+        response = self.client.crm.contacts.basic_api.create(**create_params)
         return response.to_dict()  # type: ignore[no-any-return]
```
matthewelwell: Since this is now seemingly only used by `create_self_hosted_contact`, would it make sense to move all this code into that method itself to remove any confusion? 

If not, I do like to try and maintain a convention that all public methods are listed above the private methods (i.e. please move this private method below the rest of the public methods)

api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -44,88 +45,134 @@ def should_track(user: FFAdminUser) -> bool:
...
-
-        if contact_data:
-            # The user is already present in the system as a lead
-            # for an existing organisation, so return early.
-            return
+    def create_user_hubspot_contact(self, user: FFAdminUser) -> str | None:
+        tracker = HubspotTracker.objects.filter(user=user).first()
+        tracker_cookie = tracker.hubspot_cookie if tracker else None
+        self.client.create_lead_form(user=user, hubspot_cookie=tracker_cookie)
+        contact = self.client.get_contact(user)
```
matthewelwell: We probably need to add a comment here to explain why we use the form and then retrieve the contact afterwards. I believe it's just because the form doesn't respond with the contact's ID?

api/integrations/lead_tracking/hubspot/lead_tracker.py
matthewelwell: Agreed, but let's write this out in a bit more detail as a comment, including the request in my above comment.

matthewelwell: Thinking about it more, since this is a background task already running this, I wonder if we should add a retry mechanism with back-off and then fire an exception after say 3 retries so that we can see in Sentry how often we're missing the contact ID? 

(perhaps this can be done as a separate PR)

api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -44,88 +45,134 @@ def should_track(user: FFAdminUser) -> bool:
...
+        self.client.associate_contact_to_company(
+            contact_id=hubspot_contact_id,
+            company_id=hubspot_org_id,
         )
 
-        if tracker := HubspotTracker.objects.filter(user=user).first():
-            self.client.create_lead_form(
-                user=user, hubspot_cookie=tracker.hubspot_cookie
-            )
+    def get_or_create_user_hubspot_id(self, user: FFAdminUser) -> str | None:
```
matthewelwell: I think this method can be made 'private' ?

api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -44,88 +45,134 @@ def should_track(user: FFAdminUser) -> bool:
...
+        if hubspot_contact_id is not None:
+            HubspotLead.objects.update_or_create(
+                user=user, defaults={"hubspot_id": hubspot_contact_id}
+            )
 
-        hubspot_id = self.get_or_create_organisation_hubspot_id(user, organisation)
+        return hubspot_contact_id
 
-        response = self.client.create_contact(user, hubspot_id)
+    def create_organisation_lead(
```
matthewelwell: I think this method is more like `create_user_organisation_association` ?

api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -44,88 +45,134 @@ def should_track(user: FFAdminUser) -> bool:
...
+        self, user: FFAdminUser, organisation: Organisation | None = None
+    ) -> None:
+        hubspot_contact_id = self.get_or_create_user_hubspot_id(user)
+        hubspot_org_id = self.get_or_create_organisation_hubspot_id(user, organisation)
+        if not hubspot_contact_id or not hubspot_org_id:
```
matthewelwell:
```suggestion
        if not (hubspot_contact_id and hubspot_org_id):
```

matthewelwell: If `get_or_create_user_hubspot_id` returns `None`, is there a reason to wait until after `get_or_create_organisation_hubspot_id` to bail out? Shouldn't we be economical and not call the second method if the first 'fails'?

api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -44,88 +45,134 @@ def should_track(user: FFAdminUser) -> bool:
...
         )
 
-        if tracker := HubspotTracker.objects.filter(user=user).first():
-            self.client.create_lead_form(
-                user=user, hubspot_cookie=tracker.hubspot_cookie
-            )
+    def get_or_create_user_hubspot_id(self, user: FFAdminUser) -> str | None:
+        hubspot_lead = HubspotLead.objects.filter(user=user).first()
+        if hubspot_lead:
+            hubspot_contact_id: str | None = hubspot_lead.hubspot_id
```
matthewelwell, line 115: I'm not too sure why `hubspot_id` is nullable - I've checked the production database and confirmed that there are no records where the `hubspot_id` is null. I'd say that we make it non-nullable and (to be safe) just write a custom migration function that deletes any hubspot lead records that don't have a `hubspot_id` ?

api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -44,88 +45,134 @@ def should_track(user: FFAdminUser) -> bool:
...
+                hubspot_contact_id = self.create_user_hubspot_contact(user)
+
+        return hubspot_contact_id or None
 
     def get_or_create_organisation_hubspot_id(
         self,
         user: FFAdminUser,
-        organisation: Organisation = None,  # type: ignore[assignment]
-    ) -> str:
+        organisation: Organisation | None = None,
```
matthewelwell: It feels like `organisation` should still be a required argument here, and we can just skip calling this method if the organisation is null in `create_organisation_lead` ?

api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -44,88 +45,134 @@ def should_track(user: FFAdminUser) -> bool:
...
+        domain = user.email_domain
+        company_kwargs = {"domain": domain}
+        company_kwargs["name"] = organisation.name
+        company_kwargs["organisation_id"] = organisation.id
+        company_kwargs["active_subscription"] = organisation.subscription.plan
+
+        # As Hubspot creates/associates companies based on contact domain
+        # we need to get the hubspot id when this user creates the company for the first time
+        # and update the company name
+        company = self._get_or_create_company_by_domain(**company_kwargs)
```
matthewelwell: According to the behaviour listed [here](https://www.notion.so/flagsmith/Proposed-New-Behaviour-Logic-Map-207a2b634bf98057b674db968df8a22c?source=copy_link#207a2b634bf98028b5c4c15316ee3974) we should only try and create the company by domain if they haven't joined an organisation after 30 minutes. 

Saying that aloud, I'm not sure that really makes sense, because hubspot will definitely (?) already have created the company by that point, right?

api/integrations/lead_tracking/hubspot/tasks.py
```diff
@@ -5,7 +5,7 @@
 
 
 @register_task_handler()
-def track_hubspot_lead(user_id: int, organisation_id: int = None) -> None:  # type: ignore[assignment]
+def track_hubspot_organisation_lead(user_id: int, organisation_id: int = None) -> None:  # type: ignore[assignment]
```
matthewelwell: I don't love the `organisation_lead` terminology here, I don't really know what it means - are we creating an organisation and the organisation is a lead? Are we creating an organisation and a lead? Are we creating a lead within an organisation? It's a little ambiguous I think. Is there a reason to not keep the previous name here? We're still creating the lead, right? We haven't really changed the functionality of this task I don't think, just the timing?

api/integrations/lead_tracking/hubspot/tasks.py
```diff
@@ -27,41 +26,37 @@ def track_hubspot_lead(user_id: int, organisation_id: int = None) -> None:  # ty
             id=organisation_id
         )
 
-    hubspot_lead_tracker.create_lead(**create_lead_kwargs)
+    hubspot_lead_tracker.create_organisation_lead(**create_lead_kwargs)
 
 
 @register_task_handler()
-def update_hubspot_active_subscription(subscription_id: int) -> None:
+def track_hubspot_user_contact(user_id: int) -> None:
```
matthewelwell: Again, this name isn't very clear to me. Perhaps it stems from the naming in the lead tracker? 

For example, I see that we're calling `hubspot_lead_tracker.create_user_hubspot_contact` - this would be better named (I think) as `hubspot_lead_tracker.create_hubspot_contact_for_user` ?

api/integrations/lead_tracking/hubspot/tasks.py
```diff
@@ -27,41 +26,37 @@ def track_hubspot_lead(user_id: int, organisation_id: int = None) -> None:  # ty
...
-    from users.models import FFAdminUser
+    from .lead_tracker import HubspotLeadTracker
 
-    user = FFAdminUser.objects.get(id=user_id)
-    if hasattr(user, "hubspot_lead"):
-        # Since this task is designed to be delayed, there's a chance
-        # that the user will have joined an organisation and thus been
-        # tracked in hubspot already. If so, do nothing.
-        return
+    assert settings.ENABLE_HUBSPOT_LEAD_TRACKING
```
matthewelwell: The git diff is a bit odd here, I think just because we've moved the `assert` statement below the imports. Was there a reason for doing this?

api/tests/unit/integrations/lead_tracking/hubspot/test_unit_hubspot_client.py
```diff
@@ -44,10 +50,34 @@ def test_create_lead_form(
...
+        "objectTypeId": "0-1",
+        "name": "lastname",
+        "value": staff_user.last_name,
+    } in fields
+
+    context = request_body.get("context", {})
+    if hubspot_cookie_body is not None:
+        assert context["hutk"] == hubspot_cookie_body
+    else:
+        assert context == {}
```
matthewelwell: I'm not a big fan of having conditional assert statements in tests - I'd rather create these as separate tests if possible, but I appreciate that might result in a fair amount of code duplication - is there a way we can share the code without having these conditional assert statements?

api/tests/unit/integrations/lead_tracking/hubspot/test_unit_hubspot_lead_tracking.py
```diff
@@ -1,469 +1,348 @@
...
-                "hs_object_source_id": "2902325",
-                "hs_object_source_label": "INTEGRATION",
-                "hs_pipeline": "contacts-lifecycle-pipeline",
-                "lastmodifieddate": "2024-02-26T20:02:50.069Z",
-                "lastname": user.last_name,
-            },
-            "properties_with_history": None,
-            "updated_at": datetime.datetime(2024, 2, 26, 20, 2, 50, 69000),
-        },
+        return_value={"id": HUBSPOT_USER_ID},
```
matthewelwell: We should perhaps add a comment here explaining that we're mocking the response to get_contact to simulate the situation where hubspot creates the contact immediately after we submit the form?

api/tests/unit/integrations/lead_tracking/hubspot/test_unit_hubspot_lead_tracking.py
```diff
@@ -1,469 +1,348 @@
...
+    tracker = HubspotLeadTracker()
+    tracker.create_organisation_lead(user=user, organisation=organisation)
 
+    mock_client_existing_contact.get_contact.assert_not_called()
+    mock_client_existing_contact.create_lead_form.assert_not_called()
+    mock_client_existing_contact.create_company.assert_not_called()
 
-def test_hubspot_with_filtered_email_domain_contact_and_new_organisation(
+
+def test_hubspot_find_company_by_domain_and_creates_organisation_lead_and_associates_existing_contact_from_hook(
```
matthewelwell: hmm... this test name is quite wordy and not really very descriptive - can we improve it at all? (perhaps some of the naming suggestions above factor in here too)

api/tests/unit/integrations/lead_tracking/hubspot/test_unit_hubspot_lead_tracking.py
```diff
@@ -1,469 +1,348 @@
...
-        },
+    mock_client = mocker.MagicMock()
+    mock_client.get_contact.side_effect = [None, {"id": HUBSPOT_USER_ID}]
+    mock_client.create_lead_form.return_value = {}
+    mock_client.create_company.return_value = {"id": HUBSPOT_COMPANY_ID}
+    mock_client.get_company_by_domain.return_value = None
+    mocker.patch(
+        "integrations.lead_tracking.hubspot.lead_tracker.HubspotLeadTracker._get_client",
+        return_value=mock_client,
     )
```
matthewelwell, line 180:
```suggestion
    mocker.patch(
        "integrations.lead_tracking.hubspot.lead_tracker.HubspotLeadTracker.HubspotClient",
        return_value=mock_client,
    )
```

I think this should give us the same behaviour, but means we don't have to mock a private method.

api/tests/unit/organisations/test_unit_organisations_models.py
```diff
@@ -588,3 +590,23 @@ def test_organisation_has_billing_periods(
...
+    mocker.patch("users.models.track_hubspot_user_contact")
+    user = FFAdminUser.objects.create(
+        email="test@example.com", first_name="John", last_name="Doe"
+    )
+
+    # When
+    user.add_organisation(organisation)
+    track_hubspot_organisation_lead.delay.assert_called_once_with(
+        args=(user.id, organisation.id)
+    )
```
matthewelwell:
```suggestion
    # When
    user.add_organisation(organisation)
    
    # Then
    track_hubspot_organisation_lead.delay.assert_called_once_with(
        args=(user.id, organisation.id)
    )
```

api/users/models.py
```diff
@@ -21,7 +20,8 @@
 )
 
 from integrations.lead_tracking.hubspot.tasks import (
-    track_hubspot_lead_without_organisation,
+    track_hubspot_user_contact,
+    # track_hubspot_lead_without_organisation,
```
matthewelwell:
```suggestion
```

api/tests/unit/users/test_unit_users_models.py
```diff
@@ -263,3 +266,17 @@ def test_user_create_does_not_call_pipedrive_tracking_if_ignored_domain(  # type
...
+) -> None:
+    # Given
+    settings.ENABLE_HUBSPOT_LEAD_TRACKING = True
+    track_hubspot_user_contact = mocker.patch("users.models.track_hubspot_user_contact")
+
+    # When
+    user = FFAdminUser.objects.create(
+        email="test@example.com", first_name="John", last_name="Doe"
+    )
+    track_hubspot_user_contact.delay.assert_called_once_with(args=(user.id,))
```
matthewelwell:
```suggestion
    # When
    user = FFAdminUser.objects.create(
        email="test@example.com", first_name="John", last_name="Doe"
    )
    
    # Then
    track_hubspot_user_contact.delay.assert_called_once_with(args=(user.id,))
```

frontend/env/project_local.js
```diff
@@ -14,6 +14,7 @@ module.exports = global.Project = {
   flagsmithClientAPI: 'https://edge.api.flagsmith.com/api/v1/',
 
   flagsmithClientEdgeAPI: 'https://edge.api.flagsmith.com/api/v1/',
+  hubspot: '//js-eu1.hs-scripts.com/143451822.js',
```
matthewelwell: Do we need this in `project_local` or was this just left from your testing?

matthewelwell (commented):
I've added quite a lot of comments here but I think my main observations really are: 

 1. The naming could definitely be improved in a few places
 2. It's not immediately clear to me whether we are correctly implementing the behaviour as designed [here](https://www.notion.so/flagsmith/Proposed-New-Behaviour-Logic-Map-207a2b634bf98057b674db968df8a22c?source=copy_link#207a2b634bf98028b5c4c15316ee3974). I'm not sure how to make this so - maybe a few extra comments? Also I definitely think better naming would help this point as well. As far as I'm aware, this work should ideally have resulted in 3 tasks: 

`create_hubspot_lead` (triggered on user registration, using the create lead form)
`upsert_hubspot_contact_company_association` (triggered on userorganisation creation - this task can either get or create the company in hubspot )
`update_hubspot_company` (this task can accept some payload of data, like the name, subscription information, etc. but it should be a single task I think?)

api/tests/unit/integrations/lead_tracking/hubspot/test_unit_hubspot_client.py
Zaimwa9: True, better with parametrization thanks

api/integrations/lead_tracking/hubspot/tasks.py
Zaimwa9: It's a mix between removing `track_hubspot_lead_without_organisation` and switching the import, but true, I reverted the latter as no need to import if the assert is falsy

api/integrations/lead_tracking/hubspot/tasks.py
Zaimwa9: The timing didn't change nor the output actually, it's more about the action.
My rationale was that `track_hubspot_lead` was a bit too generic.

Although I agree `organisation_lead` isn't so much better, i'd propose to still change the naming to reflect what we are doing here and so next time we know exactly what we achieve.
Before it created (if new) the organisation, and the contact. Now both of them already exists and it brings them together. So as you suggested above, i'd stick to an explicit `track_hubspot_user_organisation_association` which matches the model' hook.

api/integrations/lead_tracking/hubspot/lead_tracker.py
Zaimwa9: Yes naming again, my bad, as domain is only used to search the company. With your comments and making organisation mandatory, if we have to create it at this stage, it will be using the domain.

Although, as hubpost will create it, the creation part is a safeguard

api/integrations/lead_tracking/hubspot/lead_tracker.py
Zaimwa9: It's not nullable in the database as of. This type is shared with the [lines](https://github.com/Flagsmith/flagsmith/pull/5560/files) below when we create the contact and where because of the async it can be null there so it has to be declared in the first assignment.

NB: I'm adding the retry

frontend/common/stores/account-store.js
```diff
@@ -252,9 +252,11 @@ const controller = {
   },
   register: ({ contact_consent_given, organisation_name, ...user }) => {
     store.saving()
+
     return data
       .post(`${Project.api}auth/users/`, {
         ...user,
+        hubspotutk: API.getCookie('hubspotutk'),
```
tiagoapolo, line 259: 👍

api/tests/unit/integrations/lead_tracking/hubspot/test_unit_hubspot_lead_tracking.py
Zaimwa9: Better indeed, thanks

Zaimwa9:
Thanks for the detailed feedback.

> The naming could definitely be improved in a few places

Definitely, let me know if this feels clearer

> As far as I'm aware, this work should ideally have resulted in 3 tasks

That should be the case, more precisely there are 2 entrypoints:
1. User registration triggers `create_hubspot_contact_for_user` and ultimately the lead_form
2. New UserOrganisation entity triggers `track_hubspot_user_organisation_association` that wraps `update_company` if name needs to be updated (merged with update subscription)

Additionally, I added a basic retry with backoff in create_lead_form, without raising for now

api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -44,88 +46,147 @@ def should_track(user: FFAdminUser) -> bool:
...
+    def create_user_hubspot_contact(self, user: FFAdminUser) -> str | None:
+        tracker = HubspotTracker.objects.filter(user=user).first()
+        tracker_cookie = tracker.hubspot_cookie if tracker else None
+        self.client.create_lead_form(user=user, hubspot_cookie=tracker_cookie)
+
+        # Create lead form creates a contact asynchronously in hubspot but does not return the contact id
+        # We need to get the contact id separately and retry 3 times
+        contact = self._get_new_contact_with_retry(user, max_retries=3)
+        if not contact:
+            return None
```
matthewelwell: I'd love to know here how often this fails so we can perhaps change it if needed. We could add a `logger.error` and see how often it turns up in sentry.

api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -44,88 +46,147 @@ def should_track(user: FFAdminUser) -> bool:
...
+
+        hubspot_contact_id = contact.get("id")
+
+        # Hubspot creates contact asynchronously
+        # If not available on the spot, following steps will sync database with Hubspot
```
matthewelwell:
```suggestion
        # If not available on the spot, later steps will sync database with Hubspot
```

api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -44,88 +46,147 @@ def should_track(user: FFAdminUser) -> bool:
...
+        if hubspot_contact_id is not None:
+            HubspotLead.objects.update_or_create(
+                user=user, defaults={"hubspot_id": hubspot_contact_id}
             )
 
-    def get_or_create_organisation_hubspot_id(
+        return hubspot_contact_id
```
matthewelwell: I think we can simplify some of this since we're essentially verifying earlier that the contact is returned earlier. Maybe even, to be safe / appease mypy, the retry function should verify that the contact object we get back from hubspot has an id before returning from the method?

api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -44,88 +46,147 @@ def should_track(user: FFAdminUser) -> bool:
...
+        else:
+            # Fallback to sync database with Hubspot if contact hubspot_id was not saved
+            contact_data = self.client.get_contact(user)
+            if contact_data:
+                hubspot_contact_id = contact_data["id"]
+                HubspotLead.objects.update_or_create(
+                    user=user, defaults={"hubspot_id": hubspot_contact_id}
+                )
+            else:
+                hubspot_contact_id = self.create_user_hubspot_contact(user)
```
matthewelwell, line 128: I think we expect that this code will _never_ be called, right?

api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -44,88 +46,147 @@ def should_track(user: FFAdminUser) -> bool:
...
+
+        return hubspot_contact_id or None
```
matthewelwell: I don't think the `or None` is adding anything here is it? 

```suggestion
        return hubspot_contact_id
```

api/integrations/lead_tracking/lead_tracking.py
```diff
@@ -10,11 +12,13 @@ def __init__(self, client: typing.Any = None):
 
     @staticmethod
     @abc.abstractmethod
-    def should_track(user: FFAdminUser) -> bool:
+    def should_track(user: "FFAdminUser") -> bool:
         raise NotImplementedError()
 
     @abc.abstractmethod
-    def create_lead(self, user: FFAdminUser):  # type: ignore[no-untyped-def]
+    def create_user_organisation_association(
```
matthewelwell: This is where I'm still a little uneasy. I guess the debate here is that we've tried to make a generic abstract interface for 'creating a lead' - this interface shouldn't really care about what that means. It's a hubspot implementation details that means that a 'lead' _has_ to be associated with a company. 

I think we should just leave this as `create_lead` and ensure that the rest of the methods / functions are appropriately names so that we can easily see what 'creating a lead' actually means in hubspot. 

Otherwise, we might as well just remove this interface entirely.

matthewelwell (commented):
The code look a LOT cleaner now - so much easier to follow. I've added a few more comments, but I think that we're pretty close. 

I've resolved all of the previous comments as well.

api/integrations/lead_tracking/hubspot/lead_tracker.py
Zaimwa9: yes, I can remove but I don't think it costs a lot to have this fallback

api/integrations/lead_tracking/lead_tracking.py
Zaimwa9: alright, in that case it makes sense, and this will wrap the association between user and the organisation

api/integrations/lead_tracking/hubspot/lead_tracker.py
matthewelwell: Yeah, I agree that from an execution point of view, it doesn't cost a lot - my main concern is that I want to know when our expected path fails. I think having this fallback will mean that we have no idea when people are falling through the cracks, and I'd rather understand the reason for that and solve the root cause.

Zaimwa9: The reasons I see that **could** theoritically lead to this path would be:
- Form submission (and creating contact) failed at signup/invite => right now non blocking error
```
logger.error(
                f"Problem posting data to Hubspot's form API due to {response.status_code} "
                f"status code and following response: {response.text}"
            )
```
- No HubspotLead (we couldn't get the contact on the retries) AND get_contact failed

So we could either rely on the sentry error in that case if we want to manually deal with the form failure, let this path silently retry and maybe we could add a log here too to catch a permanent issue with this user <> hubspot ?

matthewelwell: Sure, yeah, I'd rather just add the log message in here and let Sentry notify us of the issue for us to investigate. Maybe I'll regret this after we release, and we end up spamming Sentry, but let's see!

api/integrations/lead_tracking/hubspot/tasks.py
```diff
@@ -5,7 +5,7 @@
 
 
 @register_task_handler()
-def track_hubspot_lead(user_id: int, organisation_id: int = None) -> None:  # type: ignore[assignment]
+def track_create_lead(user_id: int, organisation_id: int) -> None:
```
matthewelwell: The name of this function should still be `track_hubspot_lead` I think? BUT, I'm not sure if we'd have issues during the release by doing this, because we're changing the signature. 

For example, if we write a task from the API which doesn't include the `organisation_id`, which the new task processor instance picks up, I think it would hard fail. 

So, should we create a new task function: `track_hubspot_lead_v2` which has this signature, but just calls `track_hubspot_lead` with the `organisation_id` always provided? 

Something like: 

```python
@register_task_handler()
def track_hubspot_lead_v2(user_id: int, organisation_id: int) -> None:
    track_hubspot_lead(user_id=user_id, organisation_id=organisation_id)

@register_task_handler()
def track_hubspot_lead(user_id: int, organisation_id: int = None) -> None:
    ...
```

api/integrations/lead_tracking/pipedrive/lead_tracker.py
```diff
@@ -49,7 +50,7 @@ def should_track(user: FFAdminUser):  # type: ignore[no-untyped-def]
 
         return True
 
-    def create_lead(self, user: FFAdminUser) -> PipedriveLead:
+    def create_organisation_lead(self, user: FFAdminUser) -> PipedriveLead:
```
matthewelwell, line 53: My expectation from this latest set of changes was that this method would be removed / reverted to just be `create_lead` again?

api/integrations/lead_tracking/hubspot/lead_tracker.py
Zaimwa9: Conclusion: add logging

api/integrations/lead_tracking/pipedrive/lead_tracker.py
Zaimwa9: resolved in #5626

matthewelwell (approved):

api/integrations/lead_tracking/hubspot/services.py
```diff
@@ -7,37 +7,37 @@
...
-        .exclude(user=request.user)
+        .exclude(user=track_user)
         .exists()
     ):
         logger.info(
-            f"HubspotTracker could not be created for user {request.user.id}"
+            f"HubspotTracker could not be created for user {track_user.id}"
             f" due to cookie conflict with cookie {hubspot_cookie}"
         )
         return
```
Zaimwa9, line 31: @matthewelwell , I think we can remove this guard and allow for multiple identical cookies (cf the user who created 3 accounts. We lost the info on the account we cared about)

PR #5564, opened 2025-06-11
fix: handle (ignore) unrecognised analytics data

### LINKED ISSUE #3560: Analytics endpoint should store valid analytics data even if some is invalid
At the moment, the analytics endpoint hard fails (although still returns a 200) if one or more of the feature names are invalid. We should just filter those invalid feature names out instead and still store the data for the valid features.

matthewelwell:

## Changes

Update API logic to track analytics for any valid features in a payload, even if there are other features which are invalid. 

Updated existing tests.


api/app_analytics/serializers.py
```diff
@@ -37,3 +45,38 @@ class SDKAnalyticsFlagsSerializerDetail(serializers.Serializer):  # type: ignore
...
+        )
+        return {
+            "evaluations": [
+                evaluation
+                for evaluation in attrs["evaluations"]
+                if evaluation["feature_name"] in environment_feature_names
+            ]
+        }
+
+    def save(self, **kwargs: typing.Any) -> typing.Any:
```
emyller: `Any` 🤔

emyller (approved):

api/app_analytics/serializers.py
emyller: I think this could make more sense, though I don't have the full context of it.

```suggestion
    def save(self, **kwargs: dict[str, typing.Any]) -> None:
```

matthewelwell: As I understand it, my approach for the `kwargs` arguments is correct (see [here](https://stackoverflow.com/a/37032111), and when is SO ever wrong?). 

I have updated the return type though as you are correct - it makes no sense to use Any there.

emyller: > my approach for the `kwargs` arguments is correct

Yes, I've learned that recently. :) Thanks for updating the type.

PR #5568, opened 2025-06-12
fix: use-query-from-workflow-list-change-request

Zaimwa9:

## Changes
Scheduled changes calculation was solely relying on feature states while real use case is based on change requests (and depending on environment versioning).

Reworked the calculation by using the query used in flagsmith-workflows `GET list-change-requests` as to count exactly the  records that would be shown in scheduled changes page

- Created scheduled changes and verified the count
- Reworked the tests

### DIFF
--- a/api/environments/models.py
+++ b/api/environments/models.py
@@ -432,27 +432,28 @@ class Environment(
     def get_scheduled_metrics_queryset(self) -> QuerySet[FeatureState]:
         from features.workflows.core.models import ChangeRequest
 
-        change_requests: QuerySet["ChangeRequest"] = (
-            ChangeRequest.objects.filter(
-                environment=self,
-                committed_at__isnull=False,
-                deleted_at__isnull=True,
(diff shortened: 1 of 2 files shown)


api/environments/models.py
```diff
@@ -430,13 +430,31 @@ def get_change_requests_metrics_queryset(self) -> QuerySet["ChangeRequest"]:
...
+                | Q(
+                    Q(environment_feature_versions__deleted_at__isnull=True)
+                    & Q(environment_feature_versions__live_from__gt=timezone.now())
+                )
+                | Q(
+                    Q(change_sets__deleted_at__isnull=True)
+                    & Q(change_sets__live_from__gt=timezone.now())
+                )
+            )
+            .distinct()
```
gagantrivedi: Ah, distinct is very painful to deal with. Can we do without it? Maybe filter the queryset ourselves as a last resort?

gagantrivedi (approved):

Issue #4955, opened 2025-01-06
Implement change tracking functionality for Sentry integration


matthewelwell:
See details [here](https://develop.sentry.dev/sdk/data-model/event-payloads/contexts/#feature-flag-context) and [here](https://docs.sentry.io/product/explore/feature-flags/#change-tracking). 

Note: this issue was originally created to handle the implementation of the contexts interface / evaluation tracking functionality. As per the comments on this issue, we have implemented the change tracking functionality and raised a separate item for the backlog for evaluation tracking here: #5579.

emyller:
One trait of Sentry's Context is that it comes with some [known contexts](https://develop.sentry.dev/sdk/data-model/event-payloads/contexts/), and also allow you to add [custom context](https://docs.sentry.io/platforms/python/enriching-events/context/). From my understanding, we want to populate the "Feature flag" known context.

Sentry offers two integrations regarding feature flags:
1. **Evaluation tracking.** This is useful to know the specific flags evaluated individually (per-identity) in one transaction.
2. **Change tracking.** This marks feature flag changes as _suspect_ when errors occur, and also brings a visualization of changes over time in the UI.

Both integrations contribute to the Feature flags context within a Sentry issue, e.g.
![Image](https://github.com/user-attachments/assets/0f353658-418b-4932-b3c7-943b169434e4)

Note that Evaluation tracking is the source of truth in a transaction. Change tracking has little value for segment or identity overrides.

Sentry mentions a few feature flag providers in their documentation:
- LaunchDarkly
- Statsig
- Unleash
- OpenFeature (evaluation)

It also provides a generic integration alternative path that hints at the work we'll need done in order to have Flagsmith listed in Sentry docs. Each provider seems to implement both integrations above, e.g.
- Unleash ([evaluation tracking](https://docs.sentry.io/platforms/python/integrations/unleash/))
- Unleash ([change tracking](https://docs.sentry.io/organization/integrations/feature-flag/unleash/#change-tracking))

## Scope

### Evaluation tracking

[...]

PR #5573, opened 2025-06-17
fix: Fix Sentry Change Tracking response handling

emyller:
Related to #4955.

## Changes

- Sentry requires "change_id" for _created_ and _deleted_ events as well.
- structlog 💅 
- I was naive to assume this: https://github.com/Flagsmith/flagsmith/blob/4d415a10cf3e1b171dca9f9b982dd1c3c09f3e01/api/integrations/sentry/change_tracking.py#L92

Sentry webhook [documentation](https://github.com/getsentry/sentry/blob/ed324708947ca26392d6579ed76b93fc94a61ab6/src/sentry/flags/docs/api.md#create-generic-flag-log-post) is incorrect about the expected response. I made the mistake to assume accuracy without manual tests. Details in integration tests.

- Better tests.
- Manual tests through the lifecycle of a feature flag.

image

P.S. At this time tonight I'll assume scheduled changes work since audit logs were recently fixed. 🛌


api/tests/integration/sentry/test_change_tracking_webhook_integration.py
```diff
@@ -251,25 +281,54 @@ def test_sentry_change_tracking__flag_deleted__sends_update_to_sentry(
...
     requests_post: Mock,
-    sentry_configuration: SentryChangeTrackingConfiguration,
+    sentry_fail: Callable[[Mock], None],
 ) -> None:
     # Given
-    caplog.set_level(logging.DEBUG)
-    sentry_error = RequestException("Bad Sentry")
-    requests_post.return_value.raise_for_status.side_effect = sentry_error
+    requests_post.return_value.status_code = 200  # Sentry always responds 200 OK :(
+    sentry_fail(requests_post)
```
khvn26: nit: instead of mocking Requests, we could use `responses`/`requests-mock`.

We probably should standardise on one of the above, but certainly not introduce a _third_ way to mock HTTP responses.

api/tests/integration/sentry/test_change_tracking_webhook_integration.py
```diff
@@ -251,25 +281,54 @@ def test_sentry_change_tracking__flag_deleted__sends_update_to_sentry(
     )
     signature = requests_post.call_args[1]["headers"]["X-Sentry-Signature"]
     assert re.match(r"^[0-9a-f]{64}$", signature)
-    logs = ((record.levelname, record.message) for record in caplog.records)
-    assert ("DEBUG", f"Sending '{feature_name}' (deleted) to Sentry...") in logs
-    assert ("DEBUG", f"Sent '{feature_name}' (deleted) to Sentry") in logs
+    assert_logs(
+        ("debug", "sending", "deleted", feature_name),
+        ("info", "success", "deleted", feature_name),
+    )
```
khvn26: Can we use the `StructuredLogCapture` fixture instead? Example [here](https://github.com/Flagsmith/flagsmith/blob/50cac6f29fb740a3ba2b6a354021cac884544347/api/tests/unit/features/feature_health/test_services.py#L26-L36)

khvn26 (approved):
I approve of the fix, with a couple minor suggestions for tests.

api/tests/integration/sentry/test_change_tracking_webhook_integration.py
emyller: This is a wrapper around the `StructuredLogCapture` fixture [already](https://github.com/Flagsmith/flagsmith/blob/861cf045404cffdffe344dd121172a413f91e36e/api/tests/integration/sentry/test_change_tracking_webhook_integration.py#L42). Asserting `log.events == [` directly wouldn't favor readability because of log entries are big.

khvn26: > This is a wrapper around the StructuredLogCapture fixture [already](https://github.com/Flagsmith/flagsmith/blob/861cf045404cffdffe344dd121172a413f91e36e/api/tests/integration/sentry/test_change_tracking_webhook_integration.py#L42).

Sorry, seems like my brain got too impatient to unpack the fixture code.

> wouldn't favor readability

I appreciate the intent to make test code look cleaner but I'd argue that this is less readable, not more — I have very little idea about what the tuple elements are meant to represent and still have to check the fixture code to understand it.

> because of log entries are big.

They are not _that_ big to obsure things tbh. If we don't care about certain values, we can use `mocker.ANY` to omit them from comparison (see [example](https://github.com/Flagsmith/flagsmith/blob/b5a2af7a6aa2e5a165067c1ddf904d4f59c1dcf5/api/tests/unit/features/feature_health/test_services.py#L32-L33)).

If you're adamant on keeping this fixture, consider renaming it to denote that it's used for certain events only (e.g., `assert_sentry_change_tracking_logs`).

api/tests/integration/sentry/test_change_tracking_webhook_integration.py
khvn26: As discussed elsewhere, we're proceeding with a mock.

emyller: Attempted fix: 845b6707ca33aa8c42a729eff32c35a931dc5581

api/tests/integration/sentry/test_change_tracking_webhook_integration.py
emyller: Fixed with 2fcfdc3b91849eb5ee39affccd8d37599277b3b1.

khvn26: Awesome! 👍

api/tests/integration/sentry/test_change_tracking_webhook_integration.py
khvn26: Looks ok.

Issue #5572, opened 2025-06-16
Login page broken with `COOKIE_AUTH_ENABLED`


rolodato:
### How are you running Flagsmith

- [x] Self Hosted with Docker
- [x] Self Hosted with Kubernetes
- [ ] SaaS at flagsmith.com
- [ ] Some other way (add details in description below)

### Describe the bug

Since commit 52cf21bec5ca9919dea3b1e84e4e210523a9a30f (#5331), opening the login page results in an infinite spinner when using `COOKIE_AUTH_ENABLED`:

<img width="1512" alt="Image" src="https://github.com/user-attachments/assets/44161489-9077-42ff-a355-61324ea9abcd" />

On the previous commit 65d835b6fc36a402f193212d866b9efbdf448b90, the API responses are the same, but the login page works. This points to a frontend issue, which I haven't been able to identify:

<img width="1512" alt="Image" src="https://github.com/user-attachments/assets/9b05dffb-1047-4db4-a5b6-ae080bbf7855" />

### Steps To Reproduce

1. Check out revision 52cf21bec5ca9919dea3b1e84e4e210523a9a30f
2. Run the API locally using `COOKIE_AUTH_ENABLED=true python manage.py runserver`
3. Run the frontend locally
4. Browse to http://localhost:8080

### Expected behavior

Don't break the login page.

rolodato:
Fixed by https://github.com/Flagsmith/flagsmith/pull/5621

PR #5621, opened 2025-06-18
fix: fix cookie validation in authentication 

tiagoapolo:

## Changes

Ref: https://github.com/Flagsmith/flagsmith/issues/5572

The error was introduced in https://github.com/Flagsmith/flagsmith/commit/52cf21bec5ca9919dea3b1e84e4e210523a9a30f (https://github.com/Flagsmith/flagsmith/pull/5331) possibly by the change below and happened only if the user was not authenticated in the server, meaning it did not exchange cookies with the server:
Screenshot 2025-06-18 at 16 11 25

Failure
Untitled-2025-04-08-1703 (1)

Tested manually using api running locally

Frontend
```
  cookieAuthEnabled: true,
```

Backend
```
COOKIE_AUTH_ENABLED=true \
CORS_ORIGIN_ALLOW_ALL=true \
COOKIE_SAME_SITE=Lax \
USE_SECURE_COOKIES=false
```

### DIFF
--- a/frontend/common/services/useBuildVersion.ts
+++ b/frontend/common/services/useBuildVersion.ts
@@ -12,28 +12,37 @@ export const buildVersionService = service
       getBuildVersion: builder.query<Version, Req['getBuildVersion']>({
         providesTags: () => [{ id: 'BuildVersion', type: 'BuildVersion' }],
         queryFn: async (args, _, _2, baseQuery) => {
-          const [frontendRes, backendRes] = await Promise.all([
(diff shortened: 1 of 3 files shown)


Zaimwa9 (approved):
LGTM. Good catch.
I struggled a bit because of some env that I had, slightly similar to yours but not.

Works well with yours so maybe for later it would be nice to have the cookie_auth setup documented somewhere so it's easy to be customer-like

rolodato:
Also LGTM - could we please get this merged and released?

Issue #2479, opened 2023-07-20
Epic: Send client information from SDKs


novakzaballa:
**Is your feature request related to a problem? Please describe.**
We need to collect data about the usage of our different SDKs in order to prioritize, assign resources, and better support them.

Currently, we don´t have usage data about our different SDKs, which makes it difficult to make data-driven decision-making when it comes to prioritizing, assigning resources for maintenance, and better supporting each of them.

As a technical decision-maker at Flagsmith, I would like to have a usage dashboard/report of all (or various) of our SDKs. Metrics should include total API calls in a period of time, active clients in a period of time, and any other relevant metrics that indicate the overall adoption of SDKs. The data should be displayed in numeric and graphical formats and should include:

- Number of calls from every SDK to the API in a period of time (at least monthly).
- The percentage of every SDK calls regarding the total API calls in a period of time (at least monthly).
- Active customer accounts in a period of time per SDK.
- The history of the number (and percentage) of monthly calls during a period of time of various SDKs (or at least a particular SDK), in order to appreciate its evolution over time.

**Describe the solution you'd like**
Collect data about the usage of our different SDKs by adding the SDK Id and version to the `user-agent` header of every SDK request to the API.

As part of this work, it would be great to expose the information about any outdated / vulnerable Flagsmith clients that a customer might have in their tech stack. 

**Additional context**
Suggested headers are:
```
UserAgent=FlagsmithPythonClient/v2.5.6
X-Customer-Application-Name: CustomerAppName
X-Customer-Application-Version=<customer app version>
```

dabeeeenster:
Related - https://github.com/Flagsmith/flagsmith/issues/2354 and https://github.com/Flagsmith/flagsmith/issues/1898 - @novakzaballa can you consolidate these issues down to one thanks!

matthewelwell:
Suggested headers to add:

```
UserAgent=FlagsmithPythonClient/v2.5.6
X-Customer-Application-Name: CustomerAppName
X-Customer-Application-Version=<customer app version>
```

This will require the addition of additional arguments to the client constructor to set the customer's application name and version.

khvn26:
As discussed offline with @matthewelwell and @Zaimwa9, this epic now only covers storage of a pre-approved variety of `User-Agent` header values. `Application-Name` and `Application-Version` are not being stored anymore.

PR #5623, opened 2025-06-19
feat: Backend for SDK metrics

### LINKED ISSUE #5581: Implement API for tracking SDK metrics

khvn26:

## Changes

Closes #5581.

Contributes to #2479.

This PR adds support for optional header tracking in the backend. It adds extensible, fully typed mechanism to support other optional labels for analytics in the future. The usage data and feature evaluation data routes are updated to return labeled data buckets. Additionally, numerous typing and code structure improvements are introduced to the analytics-related code base.

### SDK API

All tracked SDK APIs now expect the following optional headers, translated to labels written to PostgreSQL or InfluxDB usage and evaluation data buckets:

- `Flagsmith-Application-Name` header: `client_application_name` label
- `Flagsmith-Application-Version` header: `client_application_version` label
- `User-Agent` header: `user_agent` label

Relevant caches and tasks are changed to properly compact the raw data, producing buckets with unique label sets.

### Management API

The following API routes are updated to support labels:

- `/api/v1/organisations/{organisation_pk}/usage-data/`
- `/api/v1/projects/{project_pk}/features/{id}/evaluation-data/`

Previously, these routes were returning one bucket per day. Now, multiple buckets can be attributed to the same day, as long as they have different unique label sets. The label sets are returned under the `label` key for each bucket:

```jsonc
// GET /api/v1/projects/1/features/19/evaluation-data/

[
    {
        "day": "2025-07-09",
        "count": 674,
        "labels": {
            "client_application_name": null,
            "client_application_version": null,
            "user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/137.0.0.0 Safari/537.36"
        }
    },
    {
        "day": "2025-07-09",
        "count": 62,
        "labels": {
            "client_application_name": null,
            "client_application_version": null,
            "user_agent": "python-requests/2.32.3"
        }
    }
]
```

Both of them now accept new query parameters:

- `client_application_name`
- `client_application_version`
- `user_agent`

to allow bucket filtering by label.

The feature is hidden behind the `"sdk_metrics_labels"` flag.

### Task processing

Due to incompatible signature change, the `track_feature_evaluation` is decommissioned in favour of the new `track_feature_evaluations_by_environment` task, which now supports labels included as part of feature evaluation data.

Unit tests have been added and updated to reflect new behaviour. Manually with local PostgreSQL storage. Additional tests will be conducted in staging against InfluxDB Cloud to measure the performance.


api/app_analytics/analytics_db_service.py
```diff
@@ -200,15 +186,61 @@ def get_feature_evaluation_data_from_local_db(
...
+    """
+    Populate start and stop date for the given period
+    from the organisation's subscription information.
+    """
+    if period in {constants.CURRENT_BILLING_PERIOD, constants.PREVIOUS_BILLING_PERIOD}:
+        if not (sub_cache and sub_cache.is_billing_terms_dates_set()):
+            raise NotFound("No billing periods found for this organisation.")
+
+    if TYPE_CHECKING:
+        assert sub_cache
```
emyller: Context aside, this feels a bit fragile. Can we make sure `sub_cache` is not None before this `assert` statement?

emyller (dismissed):
LGTM

api/app_analytics/analytics_db_service.py
khvn26: We actually are doing it by raising `NotFound`, but mypy is too dumb to understand that.

emyller: It raises only if `period` is found in that set, + if `sub_cache.is_billing_terms_dates_set()` evaluates to true, which from here feels like `sub_cache` could still be `None` at L216. Perhaps tweaking the logical expression could help mypy? This is a nit unless there's actual risk of an type/attribute error.

khvn26: Nope, it raises if `sub_cache` is `None`. `(sub_cache and sub_cache.is_billing_terms_dates_set())` evaluates to `None`, and `not (sub_cache and sub_cache.is_billing_terms_dates_set())` evaluates to `False`.

khvn26: Sorry, I'm slow today — will fix

api/app_analytics/analytics_db_service.py
```diff
@@ -17,86 +19,57 @@
...
 )
+from app_analytics.types import Labels, PeriodType
 from environments.models import Environment
 from features.models import Feature
-from organisations.models import Organisation
+from organisations.models import Organisation, OrganisationSubscriptionInformationCache
 
-from . import constants
-from .types import PERIOD_TYPE
+logger = getLogger(__name__)
```
matthewelwell: What's the reasoning for not using structlog here?

api/app_analytics/analytics_db_service.py
```diff
@@ -168,30 +134,50 @@ def get_total_events_count(organisation) -> int:  # type: ignore[no-untyped-def]
 
 
 def get_feature_evaluation_data(
-    feature: Feature, environment_id: int, period: int = 30
+    feature: Feature,
+    environment_id: int,
+    period: int = 30,
```
matthewelwell: Looks like this was an issue previously too, but it's odd to me that we are using the `period` argument in 2 different ways in this module - one is an integer and one is the `PeriodType` string literal type. If we're not able to standardise to one of these, I think we should at least update the argument names.

api/app_analytics/analytics_db_service.py
```diff
@@ -200,15 +186,61 @@ def get_feature_evaluation_data_from_local_db(
...
 
 
-def _get_environment_ids_for_org(organisation) -> List[int]:  # type: ignore[no-untyped-def]
+def _get_environment_ids_for_org(organisation: Organisation) -> list[int]:
     # We need to do this to prevent Django from generating a query that
     # references the environments and projects tables,
     # as they do not exist in the analytics database.
     return [
         e.id for e in Environment.objects.filter(project__organisation=organisation)
     ]
```
matthewelwell: Unrelated to this PR, but isn't this better written as the following? 

```python
return list(Environment.objects.filter(project__organisation=organisation).values("id", flat=True))
```

api/app_analytics/analytics_db_service.py
```diff
@@ -200,15 +186,61 @@ def get_feature_evaluation_data_from_local_db(
...
+    sub_cache: OrganisationSubscriptionInformationCache | None,
+    period: PeriodType | None = None,
+) -> tuple[datetime | None, datetime | None]:
+    """
+    Populate start and stop date for the given period
+    from the organisation's subscription information.
+    """
+    if period in {constants.CURRENT_BILLING_PERIOD, constants.PREVIOUS_BILLING_PERIOD}:
+        if not (sub_cache and sub_cache.is_billing_terms_dates_set()):
+            raise NotFound("No billing periods found for this organisation.")
```
matthewelwell: I don't have all the context yet, but just flagging this now - I'm a little concerned about how this might work for organisations on annual billing plans. In other places we end up having to do some modulo arithmetic to work out the right billing period. Maybe it's fine, because maybe the billing term dates are still set, but I'm flagging for visibility.

api/app_analytics/analytics_db_service.py
matthewelwell: @khvn26 just checking - I think this still needs fixing, right?

api/app_analytics/cache.py
```diff
@@ -51,40 +63,47 @@ def track_request(
...
-                track_feature_evaluation_influxdb.delay(
-                    kwargs={
-                        "environment_id": environment_id,
-                        "feature_evaluations": feature_evaluations,
-                    }
-                )
+            track_feature_evaluations_by_environment.delay(
+                kwargs={
+                    "environment_id": environment_id,
+                    "feature_evaluations": list(feature_evaluations.items()),
```
matthewelwell: Isn't `list(feature_evaluations.items())` a list of tuples? In which case, I don't think it's json serializable and hence the task processor won't be able to write it / read it from the task processor table? (although I'm surprised that the tests aren't failing based on this, so maybe I'm missing something)

api/app_analytics/constants.py
```diff
@@ -1,6 +1,27 @@
...
+CURRENT_BILLING_PERIOD: PeriodType
+PREVIOUS_BILLING_PERIOD: PeriodType
+NINETY_DAY_PERIOD: PeriodType
+(
+    CURRENT_BILLING_PERIOD,
+    PREVIOUS_BILLING_PERIOD,
+    NINETY_DAY_PERIOD,
+) = get_args(PeriodType)
+
+# Optional headers sent from client SDK mapped to their respetive labels.
```
matthewelwell:
```suggestion
# Optional headers sent from client SDK mapped to their respective labels.
```

api/app_analytics/constants.py
```diff
@@ -1,6 +1,27 @@
...
+TRACK_HEADERS: dict[str, Label] = {
+    "Flagsmith-Application-Name": "client_application_name",
+    "Flagsmith-Application-Version": "client_application_version",
+}
```
matthewelwell: What about other default headers that we might want to track? e.g. `User-Agent` ?

api/app_analytics/mappers.py
```diff
@@ -0,0 +1,87 @@
...
+        labels = row["labels"]
+        key = AnnotatedAPIUsageKey(
+            date=date,
+            labels=tuple(labels.items()),
+        )
+        if key not in data_by_key:
+            data_by_key[key] = UsageData(
+                day=date,
+                labels=labels,
+            )
```
matthewelwell, line 53: i'm a bit lost by this function in general, but specifically here it's odd to me that we're instantiating the key and the value with the same attributes here (date and labels). Seems like there's redundant data here.

api/app_analytics/types.py
```diff
@@ -1,9 +1,50 @@
...
+class AnnotatedAPIUsageKey(NamedTuple):
+    date: date
+    labels: tuple[tuple["Label", str], ...]
+
+
+# Optional labels Flagsmith stores for API usage and feature evaluation data.
+Label = Literal[
+    "client_application_name",
+    "client_application_version",
+]
```
matthewelwell: Can't we do something like this: `Label = Literal[*TRACK_HEADERS.values()]` so that we're not defining the labels in 2 separate places?

api/app_analytics/serializers.py
```diff
@@ -1,16 +1,45 @@
...
+
+
+class LabelsQuerySerializerMixin(_SerializerType):
+    def get_fields(self) -> dict[str, serializers.Field[Any, Any, Any, Any]]:
+        return {**super().get_fields(), **_get_label_fields()}
+
+    def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
+        """
+        Yank label filters from the top level and place them under a
+        separate `labels_filter` key
```
matthewelwell: This explains what the code does, which is fairly easy to understand from reading it. What I'd be more keen on documenting here, is _why_?

api/app_analytics/analytics_db_service.py
khvn26: fixed in 98627b502b0258df29f0cbd71af9d83d54b02b84.

api/app_analytics/tasks.py
```diff
@@ -91,20 +94,27 @@ def track_feature_evaluation_v2(
 
 
 @register_task_handler()
-def track_feature_evaluation(
+def track_feature_evaluations_by_environment(
     environment_id: int,
-    feature_evaluations: dict[str, int],
+    feature_evaluations: list[tuple[FeatureEvaluationKey, int]],
```
matthewelwell: Ok, so I've checked and I think I can understand why this works (in reference to my previous comment about tuple serialization), but I think the typing here is wrong. See the following example for why. 

```
>>> t = (1, 2, 3)
>>> type(t)
<class 'tuple'>
>>> type(json.loads(json.dumps(t)))
<class 'list'>
```

I guess the issue then (at least from a typing perspective), is that I don't think we can typehint a list with multiple types like we have here for the tuple can we?

api/app_analytics/tasks.py
```diff
@@ -91,20 +94,27 @@ def track_feature_evaluation_v2(
 
 
 @register_task_handler()
-def track_feature_evaluation(
+def track_feature_evaluations_by_environment(
```
matthewelwell, line 98: I assume that the reason for the name change mainly to get around the fact that we don't have versioning for our tasks, but as a side effect, we're choosing to be more explicit with the naming anyway?

api/conftest.py
```diff
@@ -167,15 +174,28 @@ def django_db_setup(request: pytest.FixtureRequest) -> None:
...
         # Pytest's worker indices are 0-based
         test_db_suffix = str(int(xdist_worker_id_suffix) + 1)
+
+        from django.conf import settings
+
+        for db_settings in settings.DATABASES.values():
+            test_db_name = (
+                f"{TEST_DATABASE_PREFIX}{db_settings['NAME']}_{test_db_suffix}"
+            )
+            db_settings["NAME"] = test_db_name
```
matthewelwell: I'm a little unsure what this is doing / why it's needed - it could definitely use some documentation. Also, is there a reason that we can't just use the `settings` fixture instead of importing from `django.conf` ?

api/conftest.py
```diff
@@ -167,15 +174,28 @@ def django_db_setup(request: pytest.FixtureRequest) -> None:
...
 
-    for db_settings in settings.DATABASES.values():
-        test_db_name = f"{TEST_DATABASE_PREFIX}{db_settings['NAME']}_{test_db_suffix}"
-        db_settings["NAME"] = test_db_name
+        # Ensure psycopg type handlers are registered.
+        # This is necessary for tests that involve `HStoreField`.
+        for connection in connections.all():
+            if connection.vendor == "postgresql":
+                with django_db_blocker.unblock():
+                    register_type_handlers(connection)
```
matthewelwell: Pretty yucky! Do we have something we can reference here as to why this is necessary?

matthewelwell (commented):
I haven't made it through the test code yet, but submitting my thoughts on the functional code for now.

api/app_analytics/constants.py
khvn26: Added `User-Agent` in c214840c640a2f07a7e8edb22ec3ab09950a3667.

api/app_analytics/types.py
khvn26: No. Literal types deliberately can only be defined with literal values.

In terms of dependencies in code, types come first, then the constants. The `TRACK_HEADERS` constant does not define the label, since they are defined here, rather, it establishes relations between header names and labels, which are not necessarily a 1:1 mapping. In the future, we might want to stop tracking some header while preserving the label, or track different headers under the same label. So I don't consider this an issue.

api/app_analytics/serializers.py
khvn26: Improved in c1c5bb46a49b8a4ef0417712a1ed63368c69385c.

TL/DR: I wanted neither to invent a filtering syntax like `?labels_filter=client_application_name:app`, and lose auto-generated API documentation, nor use `**kwargs` in the service functions.

api/app_analytics/tasks.py
khvn26: Yes, the current typing is a cheat/shortcut around a bigger problem I didn't want to solve as part of this PR. Even if I came up with something like

```python
class FeatureEvaluationData(TypedDict):
   feature_name: str
   labels: dict[Label, str]
```

...which is closer to what `json.loads` can produce, in typing terms it still would be cheating since `json.loads` has no awareness of neither `FeatureEvaluationData` typed dict shape nor `Label` literal constraints.

The real solution is to add support for complex type serialisation/deserialisation to Task Processor, but it's something we need to discuss ahead of implementing as there are some unknowns and implications to sort out, and conventions to come up with.

api/app_analytics/cache.py
khvn26: See https://github.com/Flagsmith/flagsmith/pull/5623#discussion_r2169468699.

api/app_analytics/tasks.py
khvn26: The assumption is correct.

api/app_analytics/mappers.py
khvn26: Added a docstring in 602c7258c8d7501d532806d6a40618d1156649fe that hopefully clears it up a bit.

> Seems like there's redundant data here.

Omitting it and then backfilling from the keys would be an over-optimisation, and arguably less readable.

api/app_analytics/analytics_db_service.py
khvn26: Fixed in 227b7b2d35ff7012c8a43a083813faaa7aa26686.

api/app_analytics/analytics_db_service.py
khvn26: Changed to `period_days` in 227b7b2d35ff7012c8a43a083813faaa7aa26686.

api/app_analytics/analytics_db_service.py
khvn26: My idea of using structlog is to emit events (e.g. `task-queued`), however, here I'm using a logger only to emit warnings which make less sense as structured events. In fact, I realised we have no convention re: user warnings — should we use `warnings.warn`, `logging.warning` or go structlog all the way?

api/app_analytics/analytics_db_service.py
khvn26: Acknowledged. This code only concerns billing/notifications, so I'm not taking any action as part of this PR.

api/conftest.py
khvn26: Introduced a much nicer (and correct) solution in 6f902defddd72faecded1394469030b11df11bfa 👍

api/conftest.py
khvn26: See https://github.com/Flagsmith/flagsmith/pull/5623#discussion_r2171973658

api/app_analytics/types.py
```diff
@@ -1,9 +1,51 @@
...
+    from app_analytics.models import Resource
 
-PERIOD_TYPE = Literal[  # type: ignore[valid-type]
-    constants.CURRENT_BILLING_PERIOD,
-    constants.PREVIOUS_BILLING_PERIOD,
-    constants.NINETY_DAY_PERIOD,
+PeriodType = Literal[
+    "current_billing_period",
+    "previous_billing_period",
+    "ninety_day_period",
```
Zaimwa9: @khvn26 careful, that's breaking the frontend contract which uses `90_day_period`.
I also prefer full letters but we need to prepare a separate PR (to avoid overcomplicating a release)

khvn26: I'm fine with leaving it as it was — will fix soon. Thanks for pointing out!

khvn26: Fixed in 03d8914e000473b931a57a188d10c6f26deceb73.

api/app_analytics/tasks.py
khvn26: Ended up using the `TypedDict` after all in 0d8d43f348c54fd8bde4f16130c6d5869b71f372.

khvn26:
Added `"sdk_metrics_labels"` flag, so this can be merged after approval.

CC @matthewelwell @emyller

api/app/settings/common.py
```diff
@@ -79,6 +79,7 @@
     "django.contrib.contenttypes",
     "django.contrib.sessions",
     "django.contrib.messages",
+    "django.contrib.postgres",
```
matthewelwell, line 82: This _might_ cause an issue in EE repo for oracle?

api/app_analytics/analytics_db_service.py
matthewelwell: It seems to me like we should standardise on one approach, and we can always filter events out that we don't care about on the other end, right?

api/integrations/flagsmith/data/environment.json
matthewelwell, line 1: Sorry, I perhaps shouldn't have merged the auto generated PR. This will now conflict with main branch - rebasing and just blindly accepting the changes from one of the 2 branches should solve.

matthewelwell (dismissed):

api/integrations/flagsmith/data/environment.json
khvn26: Seems like it's okay, Github doesn't see any conflicts here.

api/app/settings/common.py
khvn26: Do we have to have `settings/common.py` identical to the open source one?

khvn26 (commented):
> It seems to me like we should standardise on one approach, and we can always filter events out that we don't care about on the other end, right?  

Switched to structlog in 34de05e07c51d4f8f89bbc25a54f599119c9d2b5.

matthewelwell (approved):

PR #5630, opened 2025-06-19
feat: collect and track utms at signup

Zaimwa9:

## Changes
The point of this PR is to collect (from the webflow redirections ✅ ) the marketing UTMs. They are passed at signup, saved along Hubspot tracker and submitted within the `create_lead_form`

- Made hubspot_cookie optional in `Hubspot_tracker` (could be normalized and a new schema for UTMs)
- Parse, store and provide UTMs when creating lead form

**Webflow**
Classic parse and store UTMs params code (`settings -> Custom Code -> head code`
```

<script>
  // UTM capture
  // --- CONFIG ---
const UTM_KEYS = ['utm_source', 'utm_medium', 'utm_campaign', 'utm_term', 'utm_content'];
const STORAGE_KEY = 'flagsmith_utm_data';

// --- CAPTURE UTM PARAMETERS FROM URL ---
function getUTMParamsFromURL() {
  const params = new URLSearchParams(window.location.search);
  console.log(params, window.location)
  const utms = {};
  UTM_KEYS.forEach(key => {
    const val = params.get(key);
    if (val) utms[key] = val;
  });
  return utms;
}

// --- SAVE UTM PARAMS TO LOCALSTORAGE ---
function saveUTMToLocalStorage(utms) {
  if (Object.keys(utms).length > 0) {
    localStorage.setItem(STORAGE_KEY, JSON.stringify(utms));
  }
}

// --- GET UTM DATA FROM LOCALSTORAGE ---
function getUTMFromLocalStorage() {
  try {
    return JSON.parse(localStorage.getItem(STORAGE_KEY)) || {};
  } catch {
    return {};
  }
}

// --- RUN ---
const utmParams = getUTMParamsFromURL();
if (Object.keys(utmParams).length > 0) {
  saveUTMToLocalStorage(utmParams);
}
  
// --- Update app redirect links with UTMs ---
document.addEventListener('DOMContentLoaded', () => {
  const utms = JSON.parse(localStorage.getItem('flagsmith_utm_data') || '{}');
  if (Object.keys(utms).length === 0) return;

  const links = document.querySelectorAll('a[href]');
  const matchUrl = /^https:\/\/app\.flagsmith\.com(\/.*)?$/;

  links.forEach((link) => {
    const href = link.getAttribute('href');
    if (!matchUrl.test(href)) return;

    const url = new URL(href);
    Object.entries(utms).forEach(([k, v]) => url.searchParams.set(k, v));
    link.setAttribute('href', url.toString());
  });
});
</script>
```

- Tests

**Steps to reproduce**
- Activate hubspot lead tracking
- create a user by coming to `http://localhost:8080/signup?utm_campaign=xxx&utm_source=yyy[...]`
- verify in hubspot they have been tracked 

image


api/integrations/lead_tracking/hubspot/services.py
```diff
@@ -12,15 +12,20 @@
 logger = logging.getLogger(__name__)
 
 
-def register_hubspot_tracker(request: Request, user: FFAdminUser | None = None) -> None:
+def register_hubspot_tracker(
+    request: Request,
+    user: FFAdminUser | None = None,
```
khvn26, line 18: nit: consider removing since it's not used anywhere
```suggestion
```

frontend/web/components/pages/HomePage.tsx
```diff
@@ -29,6 +29,15 @@ import AccountProvider from 'common/providers/AccountProvider'
...
 import { useGetBuildVersionQuery } from 'common/services/useBuildVersion'
+import { UtmsType } from 'common/types/utms'
+
+const TRACKED_UTMS = [
+  'utm_source',
+  'utm_medium',
+  'utm_campaign',
+  'utm_content',
+  'utm_term',
+]
```
khvn26: Duplicate of `frontend/common/types/utms.ts`?

api/integrations/lead_tracking/hubspot/client.py
```diff
@@ -72,6 +76,13 @@ def create_lead_form(
...
+        if utm_data:
+            for key, utm_value in utm_data.items():
+                if key in ALLOWED_UTM_KEYS and utm_value:
+                    fields.append(
+                        {"objectTypeId": "0-1", "name": key, "value": utm_value}
+                    )
+
         context = {}
         if hubspot_cookie:
             context = {
```
khvn26, line 85: Question: Are we ok with the hardcoded `pageUri`/`pageName`?

kyle-ssg (approved):
FE looks good to me!

api/users/constants.py
```diff
@@ -1 +1,9 @@
 DEFAULT_DELETE_ORPHAN_ORGANISATIONS_VALUE = False
+
+ALLOWED_UTM_KEYS = {
+    "utm_source",
```
kyle-ssg: My gut feeling with this was whether to just blindly send up utm_x params, assuming the BE just ignores any that aren't recognised?

frontend/web/components/pages/HomePage.tsx
```diff
@@ -29,6 +29,15 @@ import AccountProvider from 'common/providers/AccountProvider'
 import AccountStore from 'common/stores/account-store'
 import { LoginRequest, RegisterRequest } from 'common/types/requests'
 import { useGetBuildVersionQuery } from 'common/services/useBuildVersion'
+import { UtmsType } from 'common/types/utms'
+
+const TRACKED_UTMS = [
```
kyle-ssg: Duplicate in types

frontend/web/components/pages/HomePage.tsx
```diff
@@ -57,6 +67,13 @@ const HomePage: React.FC = () => {
     API.setCookie('marketing_consent_given', 'true')
   }, [])
 
+  useEffect(() => {
+    const utmsParams = Utils.getUtmsFromUrl(document.location.href)
```
kyle-ssg: Minor but this is a great usecase for a hook

api/integrations/lead_tracking/hubspot/client.py
```diff
@@ -72,6 +76,13 @@ def create_lead_form(
             {"objectTypeId": "0-1", "name": "lastname", "value": user.last_name},
         ]
 
+        if utm_data:
+            for key, utm_value in utm_data.items():
+                if key in ALLOWED_UTM_KEYS and utm_value:
+                    fields.append(
+                        {"objectTypeId": "0-1", "name": key, "value": utm_value}
+                    )
```
matthewelwell: Just a stylistic comments here really, but I think there are a couple of options we can take to remove the 3 layer deep nesting here (which I have a slight aversion to). 

First up, we could remove the initial if statement if we wanted to. Which we'd probably want to do at the beginning of the method like so:

```python
for create_lead_form(..., utm_data: dict[str, str] | None = None) -> dict[str, Any]:
    utm_data = utm_data or {}
    
    ...

    for key, utm_value in utm_data.items():
        ...
```

Secondly, I think the next 2 levels (`for` + `if`) can be 'simplified' as follows:

```suggestion
        if utm_data:
            fields.extend(
                [
                    {"objectTypeId": "0-1", "name": key, "value": utm_value} 
                    for key, utm_value in utm_data.items() 
                    if key in ALLOWED_UTM_KEYS and utm_value
                ]
            )
```


If it were me, I think I'd probably opt for at least adding the second improvement here, but really, I'd argue that what you have already is the most explicit so I'll leave it to you to decide the best decision here. I'm not averse to just making no change here!

api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -68,7 +68,10 @@ def update_company_active_subscription(
     def create_user_hubspot_contact(self, user: FFAdminUser) -> str | None:
         tracker = HubspotTracker.objects.filter(user=user).first()
         tracker_cookie = tracker.hubspot_cookie if tracker else None
-        self.client.create_lead_form(user=user, hubspot_cookie=tracker_cookie)
+        utm_data = tracker.utm_data if tracker else None
+        self.client.create_lead_form(
+            user=user, hubspot_cookie=tracker_cookie, utm_data=utm_data
+        )
```
matthewelwell: Again, this is quite explicit and works just fine, so I'm not mandating that we change anything here. If we want to though, we can remove one of the independent null checks on the `tracker` by 'simplifying' it to something like: 

```python
        create_lead_form_kwargs = {"user": user}
        if tracker := HubspotTracker.objects.filter(user=user).first():
            create_lead_form_kwargs.update(
                {"utm_data": tracker.utm_data, "hubspot_cookie": tracker.hubspot_cookie}
            )
        self.client.create_lead_form(**create_lead_form_kwargs)
```

api/users/models.py
```diff
@@ -469,6 +469,33 @@ class HubspotTracker(models.Model):
...
+
+    @staticmethod
+    def build_utm_data(utm_dict: dict[str, str] | None) -> dict[str, str]:
+        if not utm_dict:
+            return {}
+        return {
+            key: utm_value
+            for key, utm_value in utm_dict.items()
+            if key in ALLOWED_UTM_KEYS and utm_value
+        }
```
matthewelwell: We're also doing this filter in the `create_lead_form` method (as per my other lengthy comment). Can we deduplicate this logic to a single place?

api/users/models.py
```diff
@@ -469,6 +469,33 @@ class HubspotTracker(models.Model):
...
+
+    @property
+    def utm_data(self) -> dict[str, str]:
+        return {
+            key: getattr(self, key) for key in ALLOWED_UTM_KEYS if getattr(self, key)
```
matthewelwell: Uh oh, we're also doing some of that same logic here!

api/integrations/lead_tracking/hubspot/services.py
```diff
@@ -12,15 +12,20 @@
...
+) -> None:
     hubspot_cookie = request.data.get(HUBSPOT_COOKIE_NAME)
+    utm_data = HubspotTracker.build_utm_data(request.data)
```
matthewelwell: Looking at this here, it feels like we should be utilising a serializer somewhere (which would likely solve a lot of the duplication of the filtering logic on `ALLOWED_UTM_KEYS`. Which endpoint does this get called from? Can we update its serializer and just pass in the utm_data to this function instead?

api/integrations/lead_tracking/hubspot/services.py
```diff
@@ -12,15 +12,20 @@
...
     track_user = user if user else request.user
-    if not hubspot_cookie:
+    if not hubspot_cookie and not utm_data:
```
matthewelwell:
```suggestion
    if not (hubspot_cookie or utm_data):
```

api/organisations/invites/views.py
```diff
@@ -42,13 +42,12 @@ def join_organisation_from_email(request, hash):  # type: ignore[no-untyped-def]
     invite = get_object_or_404(Invite, hash=hash)
     try:
         request.user.join_organisation_from_invite_email(invite)
+        register_hubspot_tracker(request)
```
matthewelwell: I'm not sure why this needed to be moved into the `try` / `except` here. I can't see any reason that `register_hubspot_tracker` would throw an `InvalidInviteError`?

matthewelwell (changes requested):
I've reviewed the python parts of this PR only. 

Overall, it looks pretty good and most of my suggestions are just stylistic. There is one main thing that I'd like to solve though, and that's the fact that we are implementing the same filter logic on the `ALLOWED_UTM_KEYS` in a number of different places. Can we simplify that at all? Perhaps by using a serializer as per one of my suggestions?

api/organisations/invites/views.py
Zaimwa9: ah yes my bad, I had a version removing it and put it back in the wrong place 🙏

api/users/models.py
Zaimwa9: Ok, I removed the duplication. I'll answer here concerning the serializer. I didn't go with one as the utms are sent as individual values, and in the DB they are stored in the same way (to avoid JSON manipulation).
So it looked like doing so in the serializer would (from what I see) add an extra formatting step?

`ALLOWED_UTM_KEYS` has 2 purposes, filtering and providing the keys to get the correct attributes.

The way I see it, without duplicated logic, is that:
- the `@property` acts as the getter from the model
- the `@static` is more of a helper to build a dict from unstructured data (maybe this one could be moved)

Let me know what you think in the new version or if it's worth using a serializer

api/users/models.py
```diff
@@ -469,6 +469,41 @@ class HubspotTracker(models.Model):
...
+        unique=True,
+        max_length=100,
+        null=True,
+        blank=True,
+    )
+    utm_source = models.CharField(max_length=256, null=True, blank=True)
+    utm_medium = models.CharField(max_length=256, null=True, blank=True)
+    utm_campaign = models.CharField(max_length=256, null=True, blank=True)
+    utm_term = models.CharField(max_length=256, null=True, blank=True)
+    utm_content = models.CharField(max_length=256, null=True, blank=True)
```
matthewelwell: This is something that we're only ever going to be worried about in SaaS, and hence we can build assuming that we only need to support postgres. Based on this, should we instead be using a JSONField here for `utm_data` instead of creating a field for each `utm` property. 

Sorry, I should probably have raised this on my first review... 

Keen to get @khvn26's opinion here too based on his recent experience using HStore field.

matthewelwell, line 482: This is something that we're only ever going to be worried about in SaaS, and hence we can build assuming that we only need to support postgres. Based on this, should we instead be using a JSONField here for `utm_data` instead of creating a field for each `utm` property. 

Sorry, I should probably have raised this on my first review... 

Keen to get @khvn26's opinion here too based on his recent experience using HStore field.

api/users/models.py
```diff
@@ -469,6 +469,41 @@ class HubspotTracker(models.Model):
...
     created_at = models.DateTimeField(auto_now_add=True)
     updated_at = models.DateTimeField(auto_now=True)
+
+    # @staticmethod
+    # def filter_utm_data(utms: dict[str, str]) -> dict[str, str]:
+    #     return {
+    #         key: value
+    #         for key in ALLOWED_UTM_KEYS
+    #         if (value := utms.get(key)) is not None
+    #     }
```
matthewelwell: This needs removing still.

api/users/models.py
matthewelwell: I think this discussion is somewhat linked to my new comment [here](https://github.com/Flagsmith/flagsmith/pull/5630#discussion_r2175284101). 

I certainly see the argument for not having to do the JSON manipulation (although this isn't an endpoint where we need to worry _too_ much about performance).

I am much happier about the reduced duplication in the new version, but I am somewhat concerned about the lack of extensibility, as per the linked comment.

api/users/models.py
khvn26: I'm fine with either

a) using a `JSONField`/`HStoreField` for the kv storage, desribed by a `TypedDict` or a Pydantic model
b) storing the UTM properties in a separate one-to-one model

khvn26 (commented):
Just a couple comments (one is a reply to @matthewelwell's thread)

api/users/models.py
Zaimwa9: So I opted for the JSONField as Pydantic model, especially to answer the (good) comment on extensibility while reducing the repetition and ensuring control over the object

api/users/models.py
```diff
@@ -469,6 +479,31 @@ class HubspotTracker(models.Model):
...
+                self.utm_data = None
+
+        super().save(*args, **kwargs)
+
+    @staticmethod
+    def build_utm_data(data: dict[str, str]) -> dict[str, str]:
+        try:
+            return UTMDataModel(**data).model_dump(exclude_none=True)
+        except ValidationError:
+            return {}
```
khvn26: This is not being used as far as I can understand?

api/users/models.py
```diff
@@ -469,6 +479,31 @@ class HubspotTracker(models.Model):
...
     updated_at = models.DateTimeField(auto_now=True)
+
+    def save(self, *args: typing.Any, **kwargs: typing.Any) -> None:
+        if self.utm_data:
+            try:
+                self.utm_data = UTMDataModel(**self.utm_data).model_dump(
+                    exclude_none=True
+                )
+            except ValidationError:
+                self.utm_data = None
```
khvn26: This does ensure the right data shape at runtime but doesn't give enough visibility for code maintenance, IMO.

Check the `LaunchDarklyImportRequest.status` JSONField implementation. It uses a `TypedDict` annotation so the IDE and mypy infer the shape of the dict.

I'm reluctant when it comes to customising Django model behaviour. Instead of overriding `save` I'd suggest explicitly mapping the values inside the `register_hubspot_tracker` service. Being explicit about it removes ambiguity/magic. You can use a Pydantic `TypeAdapter` to map to a `TypedDict`.

khvn26 (changes requested):

khvn26 (approved):
LGTM 👍

matthewelwell (approved):

Issue #5635, opened 2025-06-20
Add the ability to create environment-specific segments


matthewelwell:
Currently, it is only possible to create segments at the project level. This means that changes to segments cannot be adequately tested without affecting other environments, or nasty work arounds like duplicating segments for each environment.

matthewelwell:
Blocked by #3763

matthewelwell:
Closed as duplicate of https://github.com/Flagsmith/flagsmith/issues/5673 and https://github.com/Flagsmith/flagsmith/issues/5676 (in that this story will be complete ionce those 2 epics are completed)

Issue #5557, opened 2025-06-10
As a Flagsmith user, I want to be able to enable a feature for X% of the users of my application, while maintaining a consistent experience for users that belong to a given, well-defined, group


matthewelwell:
Example: I am a developer at Acme Inc, a B2B platform where each user on the platform is associated with at least 1 organisation. I want to release a new feature by rolling it out to 10% of the organisations in the Acme Inc. platform, such that all users who are part of those 10% of organisations see the feature.

Currently, in order to achieve this in Flagsmith, I would need to identify each organisation in the platform as an identity and then request the flags for the organisation rather than the user. This often results in multiple flag evaluations / confusion over project splits / an increase in API requests.

**Proposed solution**

In the definition of a percentage split condition on a given segment, we allow the user to choose the trait key which is used to generate the hash. This would mean that users with the same value for that trait would get the same experience. 

**Additional questions**

 - How does this apply to MV features? 
   - I think we could easily extend this same functionality to MV features by choosing the relevant trait key when you define the feature (or maybe it should be set at the project / environment level)

matthewelwell:
Closed as duplicate of #5673 and #5676 (in that this story will be complete ionce those 2 epics are completed)

Issue #3971, opened 2024-05-17
Epic: Allow date comparisons in segment rules


orionxiao:
### Is your feature request related to a problem? Please describe.

I would like to be able to do date comparisons on a trait in a way that is easy for non-technical users to configure.

### Describe the solution you'd like.

New rule operators that allow for date comparison, e.g. `some_trait [before | on | after] YYYY-MM-DD`, where `some_trait` could be an ISO-formatted date string, epoch time, or some other format (doesn't really matter). In terms of impl, would probably just be a wrapper around numeric comparison with epoch time.

### Describe alternatives you've considered

This is currently possible with numeric comparison if we use epoch time on both sides. However, it would be a nice QOL feature (especially for non-technical users) if we didn't have to do the conversion or think about epoch time at all.

utkarsh-1905:
Can i take this up?

matthewelwell:
Hi @utkarsh-1905 , this is quite a large amount of work as it will need to be implemented in the engine code in all of our server side SDKs. If you're willing to take all of that on, then great, yes! 

In terms of the design for this, I think we will need some new operators to ensure that the engine knows when to correctly format the value as a date. When these operators are chosen in the UI, it should provide a date picker to select from.

utkarsh-1905:
Yes, I am willing to take this up, in turn this will help me to understand the codebase better, so I can contribute more in the future, though I will ask a lot of doubts here 😅

dabeeeenster:
We need to give consideration here to timezones too?

matthewelwell:
> We need to give consideration here to timezones too?

Hmm... yes, I guess we either need the dates to always be TZ aware or always NOT TZ aware (and hence we assume UTC). We perhaps needs to do some research into what other feature flagging / analytics tools are doing in this regard to get some inspiration. @utkarsh-1905 could you take a look?

utkarsh-1905:
Sure I will

tomerplus500:
@utkarsh-1905, hi,
any estimation of when it will be implemented?

matthewelwell:
@tomerplus500 at the moment, we don't have a timeline for implementing this one. In the meantime, you can use integer timestamps. e.g. set your trait as follows: 

```python
flagsmith.get_identity_flags("my-identity", traits={"registered_at": int(datetime.now().timestamp())}
```

Then you can use the standard inequality operators in your Flagsmith segments as if the value is numeric.

tomerplus500:
[...]

Issue #136, opened 2021-06-18
Segment rules should be able to operate on Identity IDs


dabeeeenster:
Currently segment rule definitions can only be run against Identity traits. Users often expect the rules to operate on Identity `ids` as well.

We should introduce a reserved trait name, suggest `id`, which evaluates to the Identity unique identifier when the rules are being evaluated. 

## Technical Implementation

We can add the `id `transiently to the segment list with a special name e.g. `id` or similar before processing the rules.

## Issues

It would mean that SDKs could not set a trait with `id` which would be a breaking change. What would happen to existing traits that have an ID set? Suggest that if they do, this overrides the coercion of the id.

kyle-ssg:
We should have a special prefix for internal traits like this, one thing that was originally planned was automated traits such as device, browser. Doing so would also allow for this.

e.g $FS_id, $FS_browser

matthewelwell:
We should also consider adding date / time based values like: $FS_currentday, $FS_currenthour, $FS_currentminute. This would allow people to use segment overrides to create recurring schedules for their flags.

khvn26:
This has been implemented in https://github.com/Flagsmith/flagsmith-engine/pull/198 with the '$_' prefix for external values.

Now that I think of it, I'm in favour of adopting OpenAPI's reference syntax for local references, e.g.:
 - `#/identity/identifier`
 - `#/event/datetime`
 - `#/contexts/company/address/postalCode`

matthewelwell:
Closed as duplicate of https://github.com/Flagsmith/flagsmith/issues/5673 and https://github.com/Flagsmith/flagsmith/issues/5676 (in that this story will be complete ionce those 2 epics are completed)

PR #5654, opened 2025-06-25
docs: Context values for Segments

### LINKED ISSUE #5615: Spike: Generate user documentation for functionality #5557
Output of this spike should be example user documentation highlighting the functionality as scoped in #5557 in order to share with the team, to make a decision on whether we should implement it or not.

khvn26:

## Changes

Closes #5615.

This PR outlines the Context values for Segments functionality primarily aimed to solve the use case outlined in #5557, open the door for solving #136, #5635, #3971, and more.

Context values are used as Segment rule properties to enable more flexible segmentation. Currently, Context values include the Environment Name, Identity Identifier, and Identity Traits. The `% Split` operator, previously locked to Identity Identifier, can now be used with any Context value just as any other operator.

Recording 2025-06-25 at 18 35 21

In the future, we'll be able to add more useful predefined Context values like:
- User-defined contexts, for proper multi-context support.
- Feature evaluation results, should allow us to implement dependent flags.
- User application names/versions and user agents from the SDK metrics tracking (being added in #5623).
- Current timestamp, to segment by datetime (do we need such a convenient shortcut for scheduled flags? 😄)

The PR contains the following changes:
- Documentation for Context values. I'll be gathering feedback on that to more closely understand how to scope out the feature further.
- Mocked-up (not actually working) changes to Segment creation UI to exhibit an idea how the feature should look for end users. 
- The Evaluation context schema now includes the Environment Name Context value. Rather than using it for the SDKs as initially planned, I'm looking to repurpose the schema for the next major version of the Flagsmith evaluation engine, as it clearly defines dependencies for flag evaluation. 

N/A

### DIFF
--- a/docs/docs/basic-features/segments.md
+++ b/docs/docs/basic-features/segments.md
@@ -99,10 +99,11 @@ flag is:
 In addition to identity [traits](managing-identities.md#identity-traits), you can use the following context values as
 Segment rule properties:
 
-- **Identifier**: a unique string value used to identify your identity. Useful for bucketing via the
+- **Identifier**: a unique identifier for an identity, used for segment and multivariate feature flag targeting, and displayed in the Flagsmith UI. Useful for bucketing via the
   [`% Split`](?operators=percent#operator-details) operator, or managing a list of users via the
   [`In`](?operators=in#operator-details) operator.
-- **Environment Name**: the name of your Flagsmith environment. Useful for restricting Segments to certain environments.
+- **Identity Key**: a key used when selecting a value for a multivariate feature, or for % split segmentation. Set to an internal identifier or a composite value based on the environment key and identifier, depending on Flagsmith implementation.
+- **Environment Name**: an environment's human-readable name. Useful for restricting Segments to certain environments.
 
 Context values can be used to control your targeting more precisely:
 
--- a/docs/docs/basic-features/segments.md
+++ b/docs/docs/basic-features/segments.md
@@ -109,7 +109,7 @@ Context values can be used to control your targeting more precisely:
 
 ### Gradual rollout across tenants
 
-Suppose your application supports users who can belong to multiple organisations. Now imagine you want to roll out a new
+Suppose your application supports users who can belong to multiple organisations. Now, you want to roll out a new
 feature to 20% of your organisations, ensuring that users only have access to the feature when they are part of an organisation included in that 20%. If they switch to an organisation outside this group, they will no longer have access to the feature.
 
 Here's how you might define a segment to achieve this:
--- a/docs/docs/basic-features/segments.md
+++ b/docs/docs/basic-features/segments.md
@@ -96,6 +96,12 @@ flag is:
 
 ## Context values
 
+:::warning
+
+Currently, Context values are only available for remote evaluation. In Local evaluation, rules using Context values will evaluate to `false`.
+
+:::
+
 In addition to identity [traits](managing-identities.md#identity-traits), you can use the following context values as
 Segment rule properties:
 
--- a/docs/docs/basic-features/segments.md
+++ b/docs/docs/basic-features/segments.md
@@ -245,15 +245,11 @@ You can use Percentage Split to drive [A/B tests](/advanced-use/ab-testing) and
 [staged feature rollouts](/guides-and-examples/staged-feature-rollouts#creating-staged-rollouts).
 
 Percentage Split deterministically assigns an Identity to a bucket based on a provided [context value](#context-values)
-or a trait. This means that Segment overrides that use Percentage Split will always result in the same feature value for
-a given identity in the chosen context.
+or a trait. This means that Segment overrides that use Percentage Split will always result in the same feature state and value for
+a given identity.
 
 If you create a Segment with a single Percentage Split rule of 10% over identifier, Identities who are members of that  
 split will be guaranteed to also be in that split if it is changed to a value higher than 10%.
-
-Percentage Split deterministically assigns a "bucket" to each identity based on a provided identifier or a
-[context value](#context-values). This means that Segment overrides that use Percentage Split will always result in the
-same feature value for a given identity in the chosen context.
 
 If the Percentage Split is reduced in value, some Identities will be removed from that Percentage Split to maintain the
 balance. The algorithm is fairly simple and good to understand - it is
--- a/docs/docs/basic-features/segments.md
+++ b/docs/docs/basic-features/segments.md
@@ -98,7 +98,7 @@ flag is:
 
 :::warning
 
-Currently, Context values are only available for remote evaluation. In Local evaluation, rules using Context values will evaluate to `false`.
+Currently, context values are only available for remote evaluation. In local evaluation, rules using context values will evaluate to `false`.
 
 :::
 
--- a/docs/docs/basic-features/segments.md
+++ b/docs/docs/basic-features/segments.md
@@ -7,8 +7,8 @@ import Tabs from '@theme/Tabs'; import TabItem from '@theme/TabItem';
 # Segments
 
 A segment is a subset of [identities](/basic-features/managing-identities.md), defined by a set of rules that match
-identity [traits](managing-identities.md#identity-traits). An identity always belongs to a single environment and can
-belong to any number of segments.
+identity [traits](managing-identities.md#identity-traits) or other [context values](#context-values). An identity always
+belongs to a single environment and can belong to any number of segments.
 
 Once you have defined a segment, you can create **segment overrides** for features within an environment. A segment
 override allows you to control the state of a feature only for identities that belong to a specific segment. This is
@@ -69,14 +69,14 @@ Segment rules are evaluated in order, i.e. from top to bottom when viewed in the
 
 For example, consider the following segment:
 
-1. 10% percentage split
+1. 10% percentage split over identifier
 2. `is_subscriber = true`
 
 This segment would first select 10% of _all_ identities, and then choose subscribers from that cohort. Instead, if we
 used the opposite order:
 
 1. `is_subscriber = true`
-2. 10% percentage split
+2. 10% percentage split over identifier
 
 This would first select all subscriber identities, and then randomly choose 10% of them.
 
@@ -93,6 +93,16 @@ flag is:
 1. Identity overrides
 2. Segment overrides
 3. Default value for the current environment
+
+## Context values
+
+In addition to identity [traits](managing-identities.md#identity-traits), you can use the following context values as
+Segment rule properties:
+
+- **Identifier**: a unique string value used to identify your identity. Useful for bucketing via the
+  [`% Split`](?operators=percent#operator-details) operator, or managing a list of users via the
+  [`In`](?operators=in#operator-details) operator.
+- **Environment Name**: the name of your Flagsmith environment. Useful for restricting Segments to certain environments.
 
 ## Trait data types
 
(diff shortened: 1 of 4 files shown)


docs/docs/basic-features/segments.md
```diff
@@ -193,16 +203,15 @@ Versions are compared as defined by the [Semantic Versioning specification](http
...
-Percentage Split deterministically assigns a "bucket" to each identity solely based on its ID and not any traits,
-meaning that Segment overrides that use Percentage Split will always result in the same feature value for a given
-identity.
+Percentage Split deterministically assigns a "bucket" to each identity based on a provided identifier or a
+[context value](#context-values). This means that Segment overrides that use Percentage Split will always result in the
+same feature value for a given identity in the chosen context.
 
-If you create a Segment with a single Percentage Split rule, Identities who are members of that split when the split
-value is set to, say, 10% will be guaranteed to also be in that split if it is changed to a value higher than 10%.
+If you create a Segment with a single Percentage Split rule of 10% over Identities, Identities who are members of that
```
matthewelwell: I _think_ here you mean the following, but I wasn't completely sure. 

```suggestion
If you create a Segment with a single Percentage Split rule of 10% over identifier, Identities who are members of that
```

matthewelwell (commented):
I've added one minor comment to a specific line I wasn't sure about. 

On the whole though, it looks good to me. I think that one thing that I'd love to see (whether they make it into the docs for release or not) is some example use cases listed out and how to achieve them. 

For example, 

"A multi-tenant platform where we want to roll-out a feature by percentage such that all users in a given tenancy get a consistent experience" 
 - I think this can be achieved using this functionality + transient traits

I'm sure there are other use cases that we've received feedback about that we're trying to achieve with this too. What about different device types, are there use cases there? 

Also, on a general functionality point of view, should we consider being able to define logical context values somehow? Like a hash of their identifier, and some other trait value for example? I'm not suggesting this is part of v1, but does it need to be considered as part of the data-model?

docs/docs/basic-features/segments.md
```diff
@@ -193,16 +203,15 @@ Versions are compared as defined by the [Semantic Versioning specification](http
...
-Percentage Split is the only operator that does not require a trait. You can use it to drive
-[A/B tests](/advanced-use/ab-testing) and
+You can use Percentage Split to drive [A/B tests](/advanced-use/ab-testing) and
 [staged feature rollouts](/guides-and-examples/staged-feature-rollouts#creating-staged-rollouts).
 
-Percentage Split deterministically assigns a "bucket" to each identity solely based on its ID and not any traits,
-meaning that Segment overrides that use Percentage Split will always result in the same feature value for a given
-identity.
+Percentage Split deterministically assigns a "bucket" to each identity based on a provided identifier or a
+[context value](#context-values). This means that Segment overrides that use Percentage Split will always result in the
```
emyller: Let's say I have the following segment setup:

- Identifier `IN` id3,id4,id5
- "some trait" `% Split` 50

Can identity "id4" fall into the 50% bucket?

emyller (commented):
Looks promising!

docs/docs/basic-features/segments.md
```diff
@@ -193,16 +203,15 @@ Versions are compared as defined by the [Semantic Versioning specification](http
...
+same feature value for a given identity in the chosen context.
 
-If you create a Segment with a single Percentage Split rule, Identities who are members of that split when the split
-value is set to, say, 10% will be guaranteed to also be in that split if it is changed to a value higher than 10%.
+If you create a Segment with a single Percentage Split rule of 10% over Identities, Identities who are members of that
+split will be guaranteed to also be in that split if it is changed to a value higher than 10%.
```
Zaimwa9, line 258: Correct me if i'm wrong or too obvious but in that case we might want to precise that lowering the split would affect the initial bucket

khvn26: I believe this is mentioned in the next paragraph.

docs/docs/basic-features/segments.md
khvn26: Based on the docs, it does. Hopefully the % Split documentation is now clearer after 584a1a171a1010e989ff976b142e716d2c764e84.

docs/docs/basic-features/segments.md
khvn26: Fixed in 584a1a171a1010e989ff976b142e716d2c764e84.

khvn26:
@matthewelwell 

> On the whole though, it looks good to me. I think that one thing that I'd love to see (whether they make it into the docs for release or not) is some example use cases listed out and how to achieve them.

I've added 2 use cases in 08b9330b0fe1ad3bf129eda0f82ed4fda8dd84cc that made sense to me, let me know your thoughts.

> Also, on a general functionality point of view, should we consider being able to define logical context values somehow?

I agree this might be a worthwhile endeavour. I can see two ways to achieve that:

1. Add more operators, and build proper UX around nested conditions.
2. Think on supporting some DSL for conditions ([jsonlogic](https://jsonlogic.com/)?)

If we go route 2, we might want to think early on data accessor format to use for Context values, so the answer to 

> does it need to be considered as part of the data-model?

should be "Yes".

khvn26:
Some additional thoughts as to opportunities unlocked by Context values:

- We'll be able to nuke MV features backend by preserving existing MV UX around environment-scope segments.
- Per-identity overrides can be condensed to `environment == x, identifier IN y[]` type segments, which should save us a lot of storage and memory, and improve the local evaluation experience, especially in conjunction with #3763.

docs/docs/basic-features/segments.md
emyller: > it does

Sorry, not clear: it _does_, or it _can_? Given the two rules are combined with an `OR` operator, I'd assume it _can_.

Also, aren't these two paragraphs very similar? To my Friday brain, the explanation around combining traits and context values feels a bit confusing.

L247 (as of 3405ef0d4adf362c6569ce236c971c39995d61f5):
```
Percentage Split deterministically assigns an Identity to a bucket based on a provided [context value](#context-values)
or a trait. This means that Segment overrides that use Percentage Split will always result in the same feature value for
a given identity in the chosen context.
```

L254 (as of 3405ef0d4adf362c6569ce236c971c39995d61f5):
```
Percentage Split deterministically assigns a "bucket" to each identity based on a provided identifier or a
[context value](#context-values). This means that Segment overrides that use Percentage Split will always result in the
same feature value for a given identity in the chosen context.
```

Here's a suggestion, based on my own understanding. It's close to the first variation above:

> Percentage Split deterministically assigns a "bucket" to each identity based on a provided [context value](#context-values) or a trait. This means that Segment overrides that use Percentage Split will always result in the same feature evaluation for a given identity.

I feel like the following sounds confusing, though it can be my own lack of familiarity:
- "feature **value**" makes me wonder if this is also valid for a feature _state_.
- "in the chosen context" sounds confusing. I know we're introducing the idea of _contexts_ though this phrase sounds like I need to take a step back and learn something.

khvn26:
For reference, these are free logical operations we receive with jsonpath-ng extensions:

https://github.com/h2non/jsonpath-ng?tab=readme-ov-file#extensions

khvn26:
Thinking we could use jq syntax as well, if we prove the [jq bindings](https://github.com/mwilliamson/jq.py) to be performant enough.

jq is [Turing-complete](https://github.com/odnoletkov/advent-of-code-jq) so this could be an overkill, or just the right solution depending on how flexible we want to allow this feature to be.

emyller:
~https://jmespath.org/ is worth mentioning — it's [adopted](https://docs.aws.amazon.com/cli/v1/userguide/cli-usage-filter.html#cli-usage-filter-client-side) by the AWS CLI.~

Seems [abandoned in GitHub](https://github.com/jmespath).

docs/docs/basic-features/segments.md
```diff
@@ -94,6 +94,62 @@ flag is:
...
+Suppose your application supports users who can belong to multiple organisations. Now imagine you want to roll out a new
+feature to 50% of your total user base — but instead of targeting users randomly, you'd like to gradually release the
+feature based on the organisations they belong to.
+
+Here's how you might define a segment to achieve this:
+
+| Rule Order | Property                  | Operator | Value |
+| ---------- | ------------------------- | -------- | ----- |
+| 1          | Identifier                | % Split  | 50    |
+| 2          | `organisation_name` Trait | % Split  | 20    |
```
matthewelwell: I'd perhaps argue that, while this use case showcases what's possible, when you think about it, it feels like it boils down to the same as what we have right now where not everyone in a given tenancy is getting a consistent experience. As I understood it, that was the key use case that we're trying to solve.

matthewelwell (commented):
As discussed on our call earlier today, in answer to the questions [here](https://github.com/Flagsmith/flagsmith/pull/5654#issuecomment-3013849683), we plan to go ahead with JSONPath for context values in segment conditions. 

We verified that there are currently no conditions in our SaaS DB that begin with `$.` so we can (fairly) safely assume that we can rely on that to determine if a condition is context value or a trait (although thinking about it, the user has to make a decision in defining the segment, so why don't we just store them separately in the data model or add a flag to the model?).

docs/docs/basic-features/segments.md
khvn26: Fixed in 24ea414d494c55d825c94e21d838fdb858e51697.

matthewelwell (dismissed):

khvn26:
One more thought: Looks like we can ditch the `FeatureSegment` model as well if we come up with a performant enough implementation for the `$.feature.name` context value.

docs/docs/basic-features/segments.md
emyller: @khvn26 Let's please:
1. Remove/adjust the repeating text introduced in 584a1a171a1010e989ff976b142e716d2c764e84.
2. Improve clarity — if you agree.

emyller: Regarding point (1), this is a screenshot of the latest preview:

image

khvn26: Hopefully fixed in d8be7e8c29cafbe8e945e34dafd2f8db9fa23506.

emyller (dismissed):
Looks great!

docs/docs/basic-features/segments.md
```diff
@@ -94,6 +94,54 @@ flag is:
...
+  [`% Split`](?operators=percent#operator-details) operator, or managing a list of users via the
+  [`In`](?operators=in#operator-details) operator.
+- **Identity Key**: a key used when selecting a value for a multivariate feature, or for % split segmentation. Set to an internal identifier or a composite value based on the environment key and identifier, depending on Flagsmith implementation.
+- **Environment Name**: an environment's human-readable name. Useful for restricting Segments to certain environments.
+
+Context values can be used to control your targeting more precisely:
+
+### Gradual rollout across tenants
+
+Suppose your application supports users who can belong to multiple organisations. Now imagine you want to roll out a new
```
Zaimwa9:
```suggestion
Suppose your application supports users who can belong to multiple organisations. Now, you want to roll out a new
```
Super details but as we are already supposing in the previous sentence :)

docs/docs/basic-features/segments.md
```diff
@@ -94,6 +94,54 @@ flag is:
...
+| Rule Order | Property                  | Operator | Value |
+| ---------- | ------------------------- | -------- | ----- |
+| 1          | `organisation_name` Trait | % Split  | 20    |
+
+This setup, along with a segment override, instructs the evaluation engine to enable the feature if the user's current organisation belongs to consistent 20% of all organisations.
+
+:::info Use transient traits for multi-tenancy
+
+To avoid persisting the `organisation_name` trait on the user identity, mark it as
+[transient](../advanced-use/transient-traits.md).
```
Zaimwa9, line 132: Out of clarity, I would add just a sentence before explaining the default behavior that would store the organisation_name and why that transient trait would be preferable in some situations

Zaimwa9 (dismissed):
Approving because it's already good as of but I have 2 minor comments that I let you decide if worth

docs/docs/basic-features/segments.md
khvn26: I'd like to avoid duplicating the trait docs tbh. I think the word "persisting" gives a strong hint here.

Zaimwa9 (dismissed):

docs/docs/basic-features/segments.md
```diff
@@ -94,6 +94,60 @@ flag is:
 2. Segment overrides
 3. Default value for the current environment
 
+## Context values
+
+:::warning
+
+Currently, Context values are only available for remote evaluation. In Local evaluation, rules using Context values will evaluate to `false`.
```
matthewelwell:
```suggestion
Currently, context values are only available for remote evaluation. In local evaluation, rules using context values will evaluate to `false`.
```

matthewelwell (approved):

emyller (approved):

PR #5658, opened 2025-06-26
fix: allow access to dashboard for FoF when sdk init error 🔴 blocked by SDK

Zaimwa9:

## Changes
⚠️ Behavior changed only for FoF

Instead of Maintenance screen or infinite loading, FoF users have access to the platform

- Added `isFlagsmithOnFlagsmith` computed env 
- Changed initialization error from boolean to literals
- Allow dashboard access with default flags in case of initialization error (Specifically empty API and ENV_ID for now)
- Sticky error message

image


frontend/web/components/Maintenance.tsx
```diff
@@ -1,34 +1,21 @@
...
-            >
-              @getflagsmith
-            </a>{' '}
-            for updates.
-          </>
-        }
+        <>
+          {' '}
+          Check{' '}
+          <a target='_blank' href='https://x.com/getflagsmith' rel='noreferrer'>
```
tiagoapolo, line 14: This is not fundamental part of this PR but we should consider replacing this with https://status.flagsmith.com/ because we're not posting status updates on twitter

frontend/web/components/base/errors/init.error.ts
```diff
@@ -0,0 +1,25 @@
+export enum FlagsmithStartupErrors {
+  FOF_INIT_ERROR = 'fof_init_error',
+  UNKNOWN_ERROR = 'unknown_error',
+  MAINTENANCE_MODE = 'maintenance_mode',
+}
+
+const errorLabels: Record<FlagsmithStartupErrors, string> = {
+  [FlagsmithStartupErrors.FOF_INIT_ERROR]:
+    'Flagsmith on Flagsmith initialisation error. Please check your configuration',
```
khvn26: 1. I think it makes sense to surface the exact error here.
2. I think a link to FoF docs will make sense here: https://docs.flagsmith.com/deployment/#running-flagsmith-on-flagsmith

Zaimwa9: 2. Done

1. Right now the errors are plain strings returned inconsistently in init (some passed down as of). In this case I think we have 2 errors we can detect and we are interested in:
- The empty FoF URL and EnvironmentID (but that is actually a development edge case or deployment mistake) -> have it
- Initial flags fetching error (could be wrong URL 404, or Cors etc, tldr: API error) -> I have it as generic. I'll try to identify and surface root causes if possible

There is a pre-requisite work to do that would lead to breaking change: Standardize the errors, at least in init. I created the ticket [here](https://github.com/Flagsmith/flagsmith-js-client/issues/324)

Zaimwa9: image

and

image

khvn26: Ok, some more thoughts after seeing the detailed error messages:

I think that, since it's only remedied with a redeploy/restart, this error should not be dismissable. IMO, instead of using unified error handling, the FoF error should be displayed permanently. I think [the butter bar](https://github.com/Flagsmith/flagsmith/blob/562cf0853bdfc4d088d67eb45ca70bfdc885593e/frontend/web/components/ButterBar.tsx) is the best place for it, especially since its contents are supposed to be set with a FoF flag.

Even with details and doc link included, the errors still don't look comprehensive enough. I think we should think a little more on how to improve the messaging:
1. "Flagsmith configuration" is very ambiguous as Flagsmith can be configured in project settings, org settings, environment variables, and Flagsmith-on-Flagsmith. We should be very clear that the error relates to Flagsmith-on-Flagsmith.
2. Users should be able to understand the scope/implications of a failing FoF configuration more clearly.

I came up with this copy and applied it to Flagsmith staging:

```
#### Could not initialise Flagsmith-on-Flagsmith

The error was: 

 `Error 500 when querying https://my-fof.com/api/v1/flags` 

Flag evaluation for is not affected, but some features might be unavailable. 

Check with [Flagsmith-on-Flagsmith documentation](https://docs.flagsmith.com/deployment/#running-flagsmith-on-flagsmith) for more info.
```

It doesn't look ideal because the markdown rendering is really weird, still, it's more informative, IMO.

Zaimwa9: image
Similar to your suggestion

frontend/web/components/Maintenance.tsx
tiagoapolo, line 14:
```suggestion
          <a target='_blank' href='https://status.flagsmith.com/' rel='noreferrer'>
```

frontend/web/components/base/errors/init.error.ts
khvn26: Awesome!

frontend/web/components/Maintenance.tsx
```diff
@@ -1,34 +1,21 @@
...
+            @getflagsmith
```
tiagoapolo, line 15:
```suggestion
            Flagsmith status
```

tiagoapolo (approved):
Approved! Only one minor optional change

frontend/web/components/Maintenance.tsx
matthewelwell: Definitely agree with this, but let's include this in a separate PR.

frontend/web/components/base/errors/init.error.ts
khvn26: "Flag evaluation for" 😬

frontend/web/components/ButterBar.tsx
```diff
@@ -59,6 +64,35 @@ const ButterBar: React.FC<ButterBarProps> = ({ billingStatus, projectId }) => {
...
+      <div
+        className='announcement-container font-weight-medium bg-danger'
+        style={{ padding: '10px' }}
+      >
+        <Row>
+          <h6>Could not initialise Flagsmith-on-Flagsmith</h6>
+        </Row>
+        <p className='mb-1'>The error was: "{fofError}"</p>
+        <p className='mb-1'>
+          Flag evaluation for is not affected, but some dashboard features might
```
khvn26, line 78: Sorry, looks like there's a typo:
```suggestion
          Flag evaluation is not affected, but some dashboard features might
```

Zaimwa9:
Closing as deprioritized and would need a fairly large change in SDK.

It would be easier to re-work on it from scratch once needed

PR #5670, opened 2025-06-30
feat: generate-a-auth-cookie-on-oauth-succesful-logins

Zaimwa9:

## Changes
Allow Google and Github OAuth with cookie login.
- Set a cookie when cookie_auth is enabled and oauth succesful

- Manually
- ⚠️ couldn't test github due to redirect_uri not set for local


api/custom_auth/oauth/views.py
```diff
@@ -35,6 +38,10 @@ def login_with_google(request):  # type: ignore[no-untyped-def]
         )
         serializer.is_valid(raise_exception=True)
         token = serializer.save()
+        if settings.COOKIE_AUTH_ENABLED:
+            return authorise_response(
+                serializer.user, Response(status=HTTP_204_NO_CONTENT)
```
emyller: How about we obtain the user from the token object?

```suggestion
                token.user, Response(status=HTTP_204_NO_CONTENT)
```

api/custom_auth/oauth/serializers.py
```diff
@@ -52,6 +52,7 @@ def create(self, validated_data):  # type: ignore[no-untyped-def]
             )
 
         user = self._get_user(user_info)
+        self.user = user
```
emyller:
```suggestion
```

We can access the user from the produced token. See [this comment](https://github.com/Flagsmith/flagsmith/pull/5670#discussion_r2175318663).

FWIW, in general it's best to avoid adding new attributes to objects mid-flight. If we need, we could update the class to be aware of this attribute and make it part of its mechanics.

api/custom_auth/oauth/views.py
```diff
@@ -58,6 +65,10 @@ def login_with_github(request):  # type: ignore[no-untyped-def]
...
```
emyller:
```suggestion
                token.user, Response(status=HTTP_204_NO_CONTENT)
```

Same as above.

emyller (changes requested):
Solution looks good with a slight re-shaping. Also please add tests to cover the `settings.COOKIE_AUTH_ENABLED` case. Thanks!

api/custom_auth/oauth/serializers.py
emyller: I think we still need to delete this line, right?

api/tests/unit/custom_auth/oauth/test_unit_oauth_views.py
```diff
@@ -301,3 +302,73 @@ def test_user_with_duplicate_accounts_authenticates_as_the_correct_oauth_user(
...
+
+    url = reverse("api-v1:custom_auth:oauth:google-oauth-login")
+
+    # When
+    response = api_client.post(url, data={"access_token": "some-token"})
+
+    # Then
+    assert response.status_code == status.HTTP_204_NO_CONTENT
+    assert (jwt_access_cookie := response.cookies.get("jwt")) is not None
+    assert jwt_access_cookie["httponly"]
```
emyller:
```suggestion
    assert response.cookies["jwt"]["httponly"] is True
```

api/tests/unit/custom_auth/oauth/test_unit_oauth_views.py
```diff
@@ -301,3 +302,73 @@ def test_user_with_duplicate_accounts_authenticates_as_the_correct_oauth_user(
...
+
+    url = reverse("api-v1:custom_auth:oauth:github-oauth-login")
+
+    # When
+    response = api_client.post(url, data={"access_token": "some-token"})
+
+    # Then
+    assert response.status_code == status.HTTP_204_NO_CONTENT
+    assert (jwt_access_cookie := response.cookies.get("jwt")) is not None
+    assert jwt_access_cookie["httponly"]
```
emyller:
```suggestion
    assert response.cookies["jwt"]["httponly"] is True
```

emyller (approved):

Issue #5669, opened 2025-06-30
Segment rule ordering is not persisted


khvn26:
We have no way to explicitly provide segment rule order to the API when creating/updating them, nor any ordering defined when selecting them.

We should:
1. Add a numeric `Rule.priority` model field
2. Expose it in the API in a backwards-compatible way:
  - Add default `ordering = ["priority"]` model setting
  - Assume condition ids as priority values if no explicit `"priority"` keys offered in the mutating API requests.

I don't yet know how to reliably assume default values for the new field in a migration. We should test with current data/queries to come up with a path forward.

khvn26:
Introduced a partial solution in #5671. We still need to think on a migration plan to make the setting on by default, and deprecate it over time.

PR #5671, opened 2025-06-30
feat: Add explicit ordering for segment rule conditions

khvn26:

## Changes

This adds a `SEGMENT_RULES_CONDITIONS_EXPLICIT_ORDERING_ENABLED` setting to force `ORDER BY id ASC` clause to all segment condition selects, which remedies the existing problem with no guaranteed ordering (#5669).

Modified the existing view test to use the setting and avoid the explicit test data result ordering, which obviously masked the problem in the first place.


api/segments/models.py
```diff
@@ -300,6 +301,21 @@ def deep_clone(self, cloned_segment: Segment) -> "SegmentRule":
         return cloned_rule
 
 
+class ConditionManager(SoftDeleteExportableManager):
+    def get_queryset(
+        self,
+    ) -> models.QuerySet["Condition"]:
+        # Effectively `Condition.Meta.ordering = ("-id",) if ... else ()`,
```
emyller:
```suggestion
        # Effectively `Condition.Meta.ordering = ("id",) if ... else ()`,
```

emyller (approved):
LGTM without a comment typo.

I think you mentioned the possibility of making `SEGMENT_RULES_CONDITIONS_EXPLICIT_ORDERING_ENABLED = True` default for new users. Is that something we'd like to pursue in this PR?

khvn26:
> the possibility of making SEGMENT_RULES_CONDITIONS_EXPLICIT_ORDERING_ENABLED = True default for new users. Is that something we'd like to pursue in this PR?

Not in this PR, as it's intended to quickly unblock a customer.

api/tests/unit/segments/test_unit_segments_models.py
```diff
@@ -2,12 +2,88 @@
...
 from flag_engine.segments.constants import EQUAL, PERCENTAGE_SPLIT
+from pytest_django.fixtures import SettingsWrapper
 from pytest_mock import MockerFixture
 
 from segments.models import Condition, Segment, SegmentRule
 from segments.services import SegmentCloneService
 
 
+@pytest.mark.xfail(reason="Intentional flaky test to explain Postgres implied ordering")
+def test__ConditionManager__get_queryset__explicit_ordering_disabled__cant_guarantee_condition_ordering(
```
emyller: I was surprised to see this test failing in CI while passing in my local, then it also failed in my local. I've added an xpass exception and it's proven to xpass/xfail at random. I find it useful for documentation, but feel free to delete it.

khvn26: I don't think we should test for unspecified behaviour. Will delete.

PR #5680, opened 2025-07-01
feat: Make dark mode local

### LINKED ISSUE #5569: Dark mode inconsistent in default self-hosted installs

- [x] Self Hosted with Docker
- [ ] SaaS at flagsmith.com

### Describe the bug

When self-hosting, the identifier used for Flagsmith-on-Flagsmith is set to the local user ID. For example:

Image

Local user IDs are autoincrementing integers, meaning that they are never unique between Flagsmith instances. This means that any flags that we target using traits, such as `dark_mode`, are shared across users of unrelated instances.

In practice, this means that if a user with an ID of `1` toggles dark mode, it will cause all users with the same ID and the default FoF API URL to have their dark mode toggled without their intervention. This explains why users sometimes see dark mode randomly enabled/disabled when using their self-hosted Flagsmith.

### Steps To Reproduce

1. Create a self-hosted Flagsmith instance using the default Docker Compose setup
2. Log in and toggle dark mode
3. Create a new self-hosted instance or reset the database
4. Dark mode will be set to whatever it was set in step 2

### Expected behavior

Dark mode should not depend on any global state such as FoF, and should be a fully local setting.

### Screenshots

_No response_

kyle-ssg:

## Changes

- Moves dark mode to local storage

Note: we will still need to keep the dark mode segment and archive the dark mode flag due to users on old self hosted instasnces.

- Toggled dark mode / refreshed page

### DIFF
--- a/frontend/web/components/App.js
+++ b/frontend/web/components/App.js
@@ -23,7 +23,7 @@ import Announcement from './Announcement'
 import { getBuildVersion } from 'common/services/useBuildVersion'
 import AccountProvider from 'common/providers/AccountProvider'
 import Nav from './navigation/Nav'
-
+import 'project/darkMode'
 const App = class extends Component {
   static propTypes = {
     children: propTypes.element.isRequired,
@@ -200,10 +200,6 @@ const App = class extends Component {
         })
       }
     }
-
-    if (Utils.getFlagsmithHasFeature('dark_mode')) {
-      document.body.classList.add('dark')
-    }
   }
 
   handleScroll = () => {
@@ -231,12 +227,6 @@ const App = class extends Component {
   }
 
   render() {
-    if (
-      Utils.getFlagsmithHasFeature('dark_mode') &&
-      !document.body.classList.contains('dark')
-    ) {
-      document.body.classList.add('dark')
-    }
     const { location } = this.props
     const pathname = location.pathname
 
--- a/frontend/web/project/darkMode.ts
+++ b/frontend/web/project/darkMode.ts
@@ -1,9 +1,18 @@
+import API from './api'
+
+export const getDarkMode = () => {
+  return API.getCookie('dark_mode') === 'true'
+}
 export const setDarkMode = (enabled: boolean) => {
   if (enabled) {
+    API.setCookie('dark_mode', true)
     document.body.classList.add('dark')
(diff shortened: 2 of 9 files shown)


emyller (approved):

frontend/web/project/darkMode.ts
```diff
@@ -1,9 +1,18 @@
+import API from './api'
+
+export const getDarkMode = () => {
+  return API.getCookie('dark_mode') === 'true'
+}
 export const setDarkMode = (enabled: boolean) => {
```
Zaimwa9, line 6: can we rename in `setDarkModeCookie` or something to separate from the above local state ?

Zaimwa9 (changes requested):
As of it resets to light mode ALL the users without cookie. I think we temporarily need to use the default flagsmith value for a couple of days to gently migrate it

kyle-ssg:
> As of it resets to light mode ALL the users without cookie. I think we temporarily need to use the default flagsmith value for a couple of days to gently migrate it

We can't since there's an issue with self hosted instances

frontend/web/project/darkMode.ts
kyle-ssg: Hmm but it doesn't just set a cookie, I'm not sure I understand what you mean

frontend/web/components/App.js
```diff
@@ -200,10 +200,6 @@ const App = class extends Component {
         })
       }
     }
-
-    if (Utils.getFlagsmithHasFeature('dark_mode')) {
-      document.body.classList.add('dark')
-    }
```
Zaimwa9, line 206: In this file we have `API` so we could use it here.

I see 2 ways, either in a `componentWillMount` or with a dedicated small FC that just have a useEffect on mounting to set the initial cookie with the flagsmith value if it doesn't exist.

In both case, a comment to clean up and I'll add a reminder in my calendar

kyle-ssg: I'm not sure what you mean here, as in how do we setup the initial class of the app?

This happens in darkMode.ts 


```
setDarkMode(getDarkMode())
```

This is better than above since there's no flicker.

Zaimwa9 (approved):

Issue #5279, opened 2025-03-31
As a dashboard user, I am able to remove a flag from a release pipeline


matthewelwell:
After removing a a flag from a release pipeline, the state of the flag should not change, i.e. any changes that had already been applied by the release pipeline should still be visible on the flag. 

The flag should no longer be visible as part of the release pipeline, but we should add an audit log record to show that it was removed from the pipeline.

gagantrivedi:
This needs a bit more scoping. Here are a few questions off the top of my head:

        Should it delete the feature state values changed by the release pipeline?

        Should it be included in the completed features of the release pipeline?

    I think I’d prefer both of the above to be No.

matthewelwell:
@gagantrivedi I've updated the description of this issue in answer to your questions. 

@gagantrivedi @tiagoapolo can you guys add the expected story points for this item?

PR #5575, opened 2025-06-17
feat: Adds release pipeline status to feature modal

tiagoapolo:
Thanks for submitting a PR! Please check the boxes below:

- [ ] I have added information to `docs/` if required so people know about the feature!
- [x] I have filled in the "Changes" section below?
- [x] I have filled in the "How did you test this code" section below?
- [x] I have used a [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) title for this Pull Request

## Changes

ref: #5278 

Screenshot 2025-06-17 at 02 25 32

## How did you test this code?

_Please describe._

PR #5688, opened 2025-07-02
feat: removes feature from release pipeline

tiagoapolo:

## Changes

Screenshot 2025-07-03 at 23 19 15

Ref: [#5279](https://github.com/Flagsmith/flagsmith/issues/5279)

_Please describe._

### DIFF
--- a/frontend/web/components/modals/CreateFlag.js
+++ b/frontend/web/components/modals/CreateFlag.js
@@ -53,7 +53,10 @@ import FeatureHealthTabContent from './FeatureHealthTabContent'
 import { IonIcon } from '@ionic/react'
 import { warning } from 'ionicons/icons'
 import AddToReleasePipelineModal from 'components/release-pipelines/AddToReleasePipelineModal'
-import { getReleasePipelines } from 'common/services/useReleasePipelines'
+import {
+  getReleasePipelines,
+  removeFeatureFromReleasePipeline,
+} from 'common/services/useReleasePipelines'
 import FeaturePipelineStatus from 'components/release-pipelines/FeaturePipelineStatus'
 
 const CreateFlag = class extends Component {
@@ -641,6 +644,20 @@ const CreateFlag = class extends Component {
         featureId={this.props.projectFlag.id}
       />,
     )
+  }
+
+  removeFromReleasePipeline = () => {
+    removeFeatureFromReleasePipeline(getStore(), {
+      featureId: this.props.projectFlag.id,
+      pipelineId: this.getReleasePipelineId(),
+      projectId: this.props.projectId,
+    })
+      .then(() => {
+        toast('Feature removed from release pipeline')
+      })
+      .catch(() => {
+        toast('Error removing feature from release pipeline', 'danger')
+      })
   }
 
   render() {
@@ -1236,17 +1253,22 @@ const CreateFlag = class extends Component {
                                           }) =>
                                             isReleasePipelineEnabled &&
                                             updateFeatureState &&
-                                            !releasePipelineId &&
                                             this.state
                                               .hasPublishedReleasePipelines && (
                                               <Button
                                                 className='mr-2'
                                                 theme='secondary'
                                                 onClick={
-                                                  this.openReleasePipelineModal
+                                                  releasePipelineId
+                                                    ? this
+                                                        .removeFromReleasePipeline
+                                                    : this
+                                                        .openReleasePipelineModal
                                                 }
                                               >
-                                                Add to Release Pipeline
+                                                {releasePipelineId
+                                                  ? 'Remove from Release Pipeline'
+                                                  : 'Add to Release Pipeline'}
                                               </Button>
                                             )
                                           }
--- a/frontend/web/components/modals/CreateFlag.js
+++ b/frontend/web/components/modals/CreateFlag.js
@@ -645,7 +645,9 @@ const CreateFlag = class extends Component {
         projectId={this.props.projectId}
         featureId={this.props.projectFlag.id}
         onSuccess={() => {
-          this.fetchReleasePipelines()
+          setTimeout(() => {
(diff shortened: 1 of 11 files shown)


kyle-ssg:
This may be me not knowing too much about release pipelines, but it's not very clear what this actually does.

Naively what comes to mind is:

- We should be showing a list of statuses of pipelines this feature applies to, we probably need to know that a pipeline is not including this feature once it's removed.
- Explain clearly what removing a pipeline would do. I'm assuming this means that no pipeline "actions" would apply to this feature where as they do by default.
- We should probably have a very similar way to re-add to a pipeline once removed. This suggests to me we should just move all of this to a pipelines tab within the feature modal. If we do that, do we need to show an info message that "Excluded for 1 or more pipelines"?


If the last point isn't true, we should maybe consider changing the 3 buttons in the screenshot into an button dropdown, defaulting to "Update Feature Value"/"Create Change Request".

tiagoapolo:
> This may be me not knowing too much about release pipelines, but it's not very clear what this actually does.
> 
> Naively what comes to mind is:
> 
> * We should be showing a list of statuses of pipelines this feature applies to, we probably need to know that a pipeline is not including this feature once it's removed.
> * Explain clearly what removing a pipeline would do. I'm assuming this means that no pipeline "actions" would apply to this feature where as they do by default.
> * We should probably have a very similar way to re-add to a pipeline once removed. This suggests to me we should just move all of this to a pipelines tab within the feature modal. If we do that, do we need to show an info message that "Excluded for 1 or more pipelines"?
> 
> If the last point isn't true, we should maybe consider changing the 3 buttons in the screenshot into an button dropdown, defaulting to "Update Feature Value"/"Create Change Request".

Those are valid points but to give you more context, we already show a pipeline status on the feature modal. 
Also, a feature/flag can only be added to a single pipeline. That said, I believe it makes sense having the actions as a dropdown because we want the pipeline actions to be highlighted or a primary type of action just like "Update Feature Value"/"Create Change Request" and not in a separate tab. To be more informative a modal can be shown when clicking on "Remove Feature From Pipeline" what are the consequences.

This is what it looks like when a feature is in a pipeline
image
https://github.com/Flagsmith/flagsmith/pull/5575

tiagoapolo:
> This may be me not knowing too much about release pipelines, but it's not very clear what this actually does.
> 
> Naively what comes to mind is:
> 
> * We should be showing a list of statuses of pipelines this feature applies to, we probably need to know that a pipeline is not including this feature once it's removed.
> * Explain clearly what removing a pipeline would do. I'm assuming this means that no pipeline "actions" would apply to this feature where as they do by default.
> * We should probably have a very similar way to re-add to a pipeline once removed. This suggests to me we should just move all of this to a pipelines tab within the feature modal. If we do that, do we need to show an info message that "Excluded for 1 or more pipelines"?
> 
> If the last point isn't true, we should maybe consider changing the 3 buttons in the screenshot into an button dropdown, defaulting to "Update Feature Value"/"Create Change Request".

I created a split dropdown, wdyt if we only add primary actions to that dropdown list and leave "schedule update" out?

Screenshot 2025-07-02 at 19 54 23

frontend/web/components/modals/CreateFlag.js
```diff
@@ -1335,11 +1341,24 @@ const CreateFlag = class extends Component {
                                                       invalid ||
                                                       !savePermission
                                                     }
+                                                    dropdownItems={[
```
Zaimwa9: We could have the is4eyes condition here and have the items declared above

frontend/web/components/modals/CreateFlag.js
```diff
@@ -1335,11 +1341,24 @@ const CreateFlag = class extends Component {
...
+                                                      {
+                                                        label: releasePipelineId
```
Zaimwa9: Do you think we could add a loader here the time to fetch the pipelines if not too much of a hustle?
My local is slow sometimes it takes up to 5sec so it emphasizes it. Leading to have Add features pipelines and as I still don't have them, my dropdown in the modal is also empty.
https://www.loom.com/share/978e0eb7ddb64146bdac10894341123b?focus_title=1&muted=1&from_recorder=1

frontend/web/components/modals/CreateFlag.js
```diff
@@ -1335,11 +1341,24 @@ const CreateFlag = class extends Component {
...
+                                                            : this
+                                                                .openReleasePipelineModal,
+                                                      },
+                                                    ]}
                                                   >
                                                     {isSaving
                                                       ? 'Updating'
                                                       : 'Update Feature Value'}
-                                                  </Button>
+                                                  </ButtonDropdown>
```
Zaimwa9: I would put a "+" icon personally. So for us the dropdown is way better I'd say because we know the flow but as for a new feature I would sacrifice the esthetic to have it really obvious.

Or a + to trigger a bit of curiosity of what can I add there? I'm not opiniated at all, just trying what would be the most obvious for a user

Zaimwa9 (approved):
Besides the UX question raised by Kyle, looks good to me with 2 minor comments

frontend/web/components/modals/CreateFlag.js
```diff
@@ -242,10 +244,13 @@ const CreateFlag = class extends Component {
...
-
-    getReleasePipelines(getStore(), {
-      projectId: this.props.projectId,
-    })
+    getReleasePipelines(
+      getStore(),
+      {
+        projectId: this.props.projectId,
+      },
+      { forceRefetch: true },
```
tiagoapolo: @Zaimwa9  I'm not sure if you experienced this before, but the value returned from the api is the most recent data but RTK is returning a stale data. For instance, I add a feature to a pipeline and a new GET /release-pipelines/ response from the api includes that feature in the response but RTK still returns the stale data that does not include that feature (`pipeline[n].features`)

PR #5711, opened 2025-07-04
fix: trigger create hubspot user post tracker

### LINKED ISSUE #5697: Create hubspot contact after HubspotTracker

We use an after_create hook on User to dispatch the hubspot contact creation while we need the HubspotTracker (cookie) to be saved in database.

Trigger the contact creation after handling the tracker/cookie/utm

TBD

### Additional context

_No response_

### LINKED ISSUE #5698: Improve handling of asynchronous contact creation in hubspot

Submitting a lead_form to hubspot creates asynchronously a contact. Which can lead to not receiving the hubspot_id and later on, submit a second time the form for the same user

Options: 
- Webhook
- Rework getting the contact
- Greater retry backoff

N/C

### Additional context

_No response_

Zaimwa9:

## Changes
We need the cookie (and utms) before creating the hubspot contact. As it was deferred to the task processor as a user post_create hook, the creation could happen before storing the relevant data.

This PR moved the dispatch after creating the Hubspot tracker

- Added tests


api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -67,8 +67,15 @@ def update_company_active_subscription(
...
-        tracker_cookie = tracker.hubspot_cookie if tracker else None
-        self.client.create_lead_form(user=user, hubspot_cookie=tracker_cookie)
+        create_lead_form_kwargs: dict[str, Any] = {"user": user}
+        if tracker:
+            create_lead_form_kwargs.update(
+                {
+                    "hubspot_cookie": tracker.hubspot_cookie,
+                    "utm_data": tracker.utm_data,
+                }
+            )
```
emyller: Slight nit unnesting.
```suggestion
            create_lead_form_kwargs.update(
                hubspot_cookie=tracker.hubspot_cookie,
                utm_data=tracker.utm_data,
            )
```

api/tests/unit/custom_auth/test_unit_custom_auth_views.py
```diff
@@ -141,3 +143,50 @@ def test_patch_user_onboarding_returns_error_if_tasks_and_tools_are_missing(
...
+    hubspot_tracker = HubspotTracker.objects.filter(user=user).first()
+
+    # Then
+    assert response.status_code == status.HTTP_201_CREATED
+    assert user is not None
+    if hubspot_cookie:
+        assert hubspot_tracker is not None
+        assert hubspot_tracker.hubspot_cookie == hubspot_cookie
+    else:
+        assert hubspot_tracker is None
```
emyller:
```suggestion
    assert not hubspot_cookie or hubspot_tracker.hubspot_cookie == hubspot_cookie
```

api/users/models.py
```diff
@@ -463,12 +464,29 @@ class HubspotLead(models.Model):
...
-    hubspot_cookie = models.CharField(unique=True, max_length=100, null=False)
+    hubspot_cookie = models.CharField(
+        unique=True,
+        max_length=100,
+        null=True,
+        blank=True,
+    )
+    utm_data: HubspotTrackerUTMData = models.JSONField(
+        default=None, blank=True, null=True
+    )  # type: ignore[assignment]
```
emyller: It looks to me like this is indeed not the correct typing, though I understand the intention. Since we already have a serializer, I'd suggest we delete `HubspotTrackerUTMData` for now unless it's very relevant somehow.

emyller (changes requested):
Looks great with minor changes.

api/tests/unit/users/test_unit_users_views.py
```diff
@@ -56,17 +50,14 @@ def test_join_organisation_via_link(
...
 
     # When
-    response = staff_client.post(url, data)
+    response = staff_client.post(url)
     staff_user.refresh_from_db()
 
     # Then
     assert response.status_code == status.HTTP_200_OK
     assert organisation in staff_user.organisations.all()
-    assert HubspotTracker.objects.filter(user=staff_user).exists()
```
Zaimwa9, line 69: Removed as this is handled during user creation

frontend/common/stores/account-store.js
```diff
@@ -18,9 +18,7 @@ const controller = {
     API.setInvite('')
     API.setInviteType('')
     return data
-      .post(`${Project.api}users/join/link/${id}/`, {
-        hubspotutk: API.getCookie('hubspotutk'),
-      })
```
Zaimwa9, line 23: Removed as this is handled during user creation

frontend/common/stores/account-store.js
```diff
@@ -95,7 +91,6 @@ const controller = {
 
     return data
       .post(`${Project.api}organisations/`, {
-        hubspotutk: API.getCookie('hubspotutk'),
```
Zaimwa9, line 98: Removed as this is handled during user creation

Zaimwa9:
Some context:

For attribution reasons, it is not possible to use the hubspot contact API. One root cause to the contact creation fallback triggered [here](https://flagsmith.sentry.io/issues/6719774706/?referrer=slack&notification_uuid=d28af610-df43-41f4-a839-d37219d10344&environment=production&alert_rule_id=14542360&alert_type=issue) might be the trigger of both the `create_hubspot_contact_for_user` and `track_hubspot_lead_v2` in the case of an invitation acceptation - contact not found while associating.

The asynchronous creation of a contact following the submission of a form might lead to some delay before it is retrievable ([doc](https://developers.hubspot.com/docs/guides/api/crm/search#limitations), [discussions](https://community.hubspot.com/t5/APIs-Integrations/Add-new-record-Search-API-why-delay/m-p/887647?utm_source=chatgpt.com)).

Latest commit changes:
- Google/Github auth handle cookies and UTMs + create hubspot contact
- Hubspot tracker and contact creation solely dealt during user creation (removed from invitation and join org which became redundant)
- Added 3min delay on `track_hubspot_lead_v2` to avoid concurrency and as real-time is not a constraint

api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -94,11 +92,14 @@ def create_user_hubspot_contact(self, user: FFAdminUser) -> str | None:
...
 
     def create_lead(self, user: FFAdminUser, organisation: Organisation) -> None:
+        print("hello")
         hubspot_contact_id = self._get_or_create_user_hubspot_id(user)
         if not hubspot_contact_id:
+            print("No hubspot contact id")
             return
         hubspot_org_id = self._get_or_create_organisation_hubspot_id(user, organisation)
         if not hubspot_org_id:
+            print("No hubspot org id")
```
emyller: 🖨️

api/organisations/invites/views.py
```diff
@@ -47,8 +44,7 @@ def join_organisation_from_email(request, hash):  # type: ignore[no-untyped-def]
         error_data = {"detail": str(e)}
         return Response(data=error_data, status=status.HTTP_400_BAD_REQUEST)
 
-    register_hubspot_tracker(request)
-
+    # register_hubspot_tracker_and_track_user(request)
```
emyller: Please either remove this, or explicitly add a comment to explain why it's commented out.

api/organisations/invites/views.py
```diff
@@ -67,7 +63,7 @@ def join_organisation_from_link(request, hash):  # type: ignore[no-untyped-def]
     if invite.is_expired:
         raise InviteExpiredError()
 
-    register_hubspot_tracker(request)
+    # register_hubspot_tracker_and_track_user(request)
```
emyller: Same

api/organisations/views.py
```diff
@@ -114,8 +111,8 @@ def create(self, request, **kwargs):  # type: ignore[no-untyped-def]
         Override create method to add new organisation to authenticated user
         """
 
-        register_hubspot_tracker(request)
         user = request.user
+        # register_hubspot_tracker_and_track_user(request, user)
```
emyller: One more

api/tests/unit/custom_auth/oauth/test_unit_oauth_serializers.py
```diff
@@ -1,32 +1,46 @@
...
-    request = rf.post("/api/v1/auth/oauth/google/")
+    data = {
+        "access_token": access_token,
+        "sign_up_type": sign_up_type,
+        "hubspotutk": "test-hubspot-utk",
+        "utm_data": {"utm_source": "test-utm-data"},
+    }
+    rf = APIRequestFactory()
+    django_request = rf.post("/api/v1/auth/oauth/google/", data=data, format="json")
+    request = Request(django_request, parsers=[JSONParser()])
```
emyller, line 51: I'd expect `APIRequestFactory` not to explicit require DRF's own `JSONParser`. At least that's what [docs](https://www.django-rest-framework.org/api-guide/testing/#using-the-format-argument) imply. Can we double check this? It can be unimportant but I'm trying to explore:
- Not forcing a test pattern that might hide bugs or create a bad habit.
- More elegant test code.

api/tests/unit/custom_auth/test_unit_custom_auth_views.py
```diff
@@ -141,3 +143,48 @@ def test_patch_user_onboarding_returns_error_if_tasks_and_tools_are_missing(
...
+
+    user = FFAdminUser.objects.filter(email="test@exemple.fr").first()
+    hubspot_tracker = HubspotTracker.objects.filter(user=user).first()
+
+    # Then
+    assert response.status_code == status.HTTP_201_CREATED
+    assert user is not None
+    assert not hubspot_cookie or (
+        hubspot_tracker is not None and hubspot_tracker.hubspot_cookie == hubspot_cookie
+    )
```
emyller:
```suggestion
    assert not hubspot_cookie or hubspot_tracker.hubspot_cookie == hubspot_cookie
```

The `hubspot_tracker is not None` is redundant. If it happens to be `None`, the test will fail with an `AttributeError`.

api/tests/unit/custom_auth/oauth/test_unit_oauth_serializers.py
Zaimwa9: I was expecting the same but after researches, my understanding is that we must set the parsers ourselves are we building the request manually outside of a viewset. Otherwise media-type is not accepted. Which is basically the returned error from django: `rest_framework.exceptions.UnsupportedMediaType: Unsupported media type "application/json" in request.`

In the context of a view:
```
class APIView(View):

    # The following policies may be set at either globally, or per-view.
    renderer_classes = api_settings.DEFAULT_RENDERER_CLASSES
    parser_classes = api_settings.DEFAULT_PARSER_CLASSES
```

With 
```
'DEFAULT_PARSER_CLASSES': [
        'rest_framework.parsers.JSONParser',
        'rest_framework.parsers.FormParser',
        'rest_framework.parsers.MultiPartParser',
    ],
```

And later on

```
    def initialize_request(self, request, *args, **kwargs):
        """
        Returns the initial request object.
        """
        parser_context = self.get_parser_context(request)

        return Request(
            request,
            parsers=self.get_parsers(),
            authenticators=self.get_authenticators(),
            negotiator=self.get_content_negotiator(),
            parser_context=parser_context
        )

```

api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -94,6 +92,7 @@ def create_user_hubspot_contact(self, user: FFAdminUser) -> str | None:
         return hubspot_contact_id
 
     def create_lead(self, user: FFAdminUser, organisation: Organisation) -> None:
+        print("hello")
```
emyller: 👋

Zaimwa9: Easter egg sympathy move. Thanks for catching

api/custom_auth/oauth/serializers.py
```diff
@@ -87,9 +95,15 @@ def _get_user(self, user_data: dict):  # type: ignore[type-arg,no-untyped-def]
                 email=email, sign_up_type=self.validated_data.get("sign_up_type")
             )
 
-            return UserModel.objects.create(
+            user = UserModel.objects.create(
                 **user_data, email=email.lower(), sign_up_type=sign_up_type
             )
+
+            # On first OAuth signup, we register the hubspot cookies and utms before creating the hubspot contact
+            register_hubspot_tracker_and_track_user(self.context.get("request"), user)  # type: ignore[arg-type]
```
emyller: Note that `request` has to be explicitly passed on to the serializer context, e.g. [Google](https://github.com/Flagsmith/flagsmith/blob/e4bc53a78ea17d162244c29a6b8abcfca7c98226/api/custom_auth/oauth/views.py#L37) and [GitHub](https://github.com/Flagsmith/flagsmith/blob/e4bc53a78ea17d162244c29a6b8abcfca7c98226/api/custom_auth/oauth/views.py#L62) views.

Let's please remove the `ignore[arg-type]` comment and discuss whether we should:
1. Not call `register_hubspot_tracker_and_track_user` if the request isn't set into the serializer context, or
2. Ensure the request is **always** present in **every** view creating an instance of any `OAuthLoginSerializer` subclass and accept `self.context["request"]` to crash otherwise — let's not use `.get()` in that case.

```suggestion
            if request := self.context.get("request"):
                register_hubspot_tracker_and_track_user(request, user)
```

Suggested code above covers option (1), though it's meant for convenience not an actual suggestion.

api/custom_auth/oauth/views.py
```diff
@@ -63,6 +66,7 @@ def login_with_github(request):  # type: ignore[no-untyped-def]
         )
         serializer.is_valid(raise_exception=True)
         token = serializer.save()
+        register_hubspot_tracker_and_track_user(request, token.user)
```
emyller: Why is this present here but not in other `login_with_*` functions? I'd expect the `register_hubspot_tracker_and_track_user` call in `OAuthLoginSerializer._get_user` to cover all cases since it lives in an upper layer of the authentication code.

api/tests/unit/custom_auth/oauth/test_unit_oauth_serializers.py
```diff
@@ -48,7 +64,15 @@ def test_create_oauth_login_serializer(
     response = serializer.save()
 
     # Then
-    assert FFAdminUser.objects.filter(email=email, sign_up_type=sign_up_type).exists()
+    user = FFAdminUser.objects.filter(email=email, sign_up_type=sign_up_type).first()
+    assert user is not None
+    hubspot_tracker = HubspotTracker.objects.filter(
+        user=user,
+    ).first()
+    assert hubspot_tracker is not None
```
emyller:
```suggestion
    hubspot_tracker = HubspotTracker.objects.get(user=user)
```

This will fail if there's no tracker object so `assert obj is not None` is not necessary.

api/tests/unit/custom_auth/oauth/test_unit_oauth_serializers.py
```diff
@@ -1,32 +1,46 @@
...
 
 @mock.patch("custom_auth.oauth.serializers.get_user_info")
 def test_create_oauth_login_serializer(
-    mock_get_user_info: mock.MagicMock, db: None
+    mock_get_user_info: mock.MagicMock,
+    db: None,
+    settings: SettingsWrapper,
+    mocker: MagicMock,
 ) -> None:
+    settings.ENABLE_HUBSPOT_LEAD_TRACKING = True
```
emyller: Can we keep this organized within the `# Given` section if it makes sense?

api/tests/unit/custom_auth/oauth/test_unit_oauth_serializers.py
```diff
@@ -103,16 +127,18 @@ def test_OAuthLoginSerializer_calls_is_authentication_method_valid_correctly_if_
...
+    request = Request(django_request, parsers=[JSONParser()])
     user_email = "test_user@test.com"
     mocked_auth_controller = mocker.MagicMock()
     mocker.patch.dict(
         "sys.modules", {"auth_controller.controller": mocked_auth_controller}
     )
 
     serializer = OAuthLoginSerializer(  # type: ignore[abstract]
-        data={"access_token": "some_token"}, context={"request": request}
+        data=data, context={"request": request}
```
emyller, line 151: I think moving the payload to a variable doesn't help because it requires more eye travel to understand the testing.

api/tests/unit/custom_auth/oauth/test_unit_oauth_serializers.py
```diff
@@ -139,16 +165,18 @@ def test_OAuthLoginSerializer_allows_registration_if_sign_up_type_is_invite_link
...
+    request = Request(django_request, parsers=[JSONParser()])
     user_email = "test_user@test.com"
 
     serializer = OAuthLoginSerializer(  # type: ignore[abstract]
-        data={
-            "access_token": "some_token",
-            "sign_up_type": SignUpType.INVITE_LINK.value,
-            "invite_hash": invite_link.hash,
-        },
+        data=data,
```
emyller, line 193: Adding to the previous comment: it also makes the PR longer than needed.

api/tests/unit/custom_auth/test_unit_custom_auth_views.py
```diff
@@ -141,3 +143,47 @@ def test_patch_user_onboarding_returns_error_if_tasks_and_tools_are_missing(
...
+        "hubspotutk": hubspot_cookie,
+    }
+
+    mock_create_hubspot_contact_for_user = mocker.patch(
+        "integrations.lead_tracking.hubspot.services.create_hubspot_contact_for_user"
+    )
+
+    url = reverse("api-v1:custom_auth:ffadminuser-list")
+    # When
+    response = staff_client.post(url, data=data, format="json")
```
emyller, line 172:
```suggestion
    # When
    url = reverse("api-v1:custom_auth:ffadminuser-list")
    response = staff_client.post(url, data=data, format="json")
```

Although I'd rather use plain URLs as to prevent unexpected schema changes and holding on to occasional typos. 🤔  I appreciate reversing URLs is a pattern in tests so I'm OK with dismissing this portion of the comment and discussing with the team.

api/tests/unit/custom_auth/test_unit_custom_auth_views.py
```diff
@@ -141,3 +143,47 @@ def test_patch_user_onboarding_returns_error_if_tasks_and_tools_are_missing(
...
+
+    user = FFAdminUser.objects.filter(email="test@exemple.fr").first()
+    hubspot_tracker = HubspotTracker.objects.filter(user=user).first()
+    hubspot_cookie_db = hubspot_tracker.hubspot_cookie if hubspot_tracker else None
+    # Then
+    assert response.status_code == status.HTTP_201_CREATED
+    assert user is not None
+    assert hubspot_cookie_db == hubspot_cookie
+
+    mock_create_hubspot_contact_for_user.delay.assert_called_once_with(args=(user.id,))
```
emyller:
```suggestion
    # Then
    assert response.status_code == status.HTTP_201_CREATED
    user = FFAdminUser.objects.get(email="test@exemple.fr")
    assert not hubspot_cookie or HubspotTracker.objects.filter(
        user=user,
        hubspot_cookie=hubspot_cookie,
    ).exists()
    mock_create_hubspot_contact_for_user.delay.assert_called_once_with(args=(user.id,))
```

IMO the `or` suggested above is the farthest we could go logic-wise within this test before dismissing `parametrize` and breaking it into two.

api/tests/unit/integrations/lead_tracking/hubspot/test_unit_hubspot_lead_tracking.py
```diff
@@ -112,22 +115,29 @@ def test_create_organisation_lead_skips_all_tracking_when_lead_exists(
     mock_client_existing_contact.create_company.assert_not_called()
 
 
+@pytest.mark.freeze_time("2023-01-19T09:09:47.325132+00:00")
```
emyller:
```suggestion
@pytest.mark.freeze_time("2023-01-19T09:09:47+00:00")
```

Microseconds will occasionally cause test flakiness. :(

api/tests/unit/integrations/lead_tracking/hubspot/test_unit_hubspot_lead_tracking.py
```diff
@@ -136,15 +146,20 @@ def test_hubspot_user_org_hook_creates_hubspot_user_and_organisation_association
...
     # Then
+    mock_track_hubspot_lead_v2.delay.assert_called_once_with(
+        args=(user.id, organisation.id),
+        delay_until=timezone.now() + timedelta(minutes=3),
+    )
+
+    # Triggering it manually to void the delay
+    track_hubspot_lead_v2(user.id, organisation.id)
     organisation.refresh_from_db()
+    assert organisation.hubspot_organisation is not None
```
emyller, line 166: Do we know what `organisation.hubspot_organisation` is expected to be more specifically than `not None`? Perhaps the `.return_value` of some mocked machinery?

api/tests/unit/organisations/test_unit_organisations_models.py
```diff
@@ -652,13 +652,14 @@ def test_organisation_has_billing_periods(
     assert result == expected_result
 
 
+@pytest.mark.freeze_time("2023-01-19T09:09:47.325132+00:00")
```
emyller: Same as above.

```suggestion
@pytest.mark.freeze_time("2023-01-19T09:09:47+00:00")
```

api/custom_auth/oauth/serializers.py
Zaimwa9: The ignore[arg-type] was actually for the user, coming from the model and not FFAdminUser there. But I solved it and made the change according to your feedback too

emyller (dismissed):
Logic-wise it looks like we're on the right path! Few questions and we should be closer to merging.

api/custom_auth/oauth/views.py
Zaimwa9: Removed this one, it's a miss on my side

api/custom_auth/oauth/serializers.py
emyller: Thanks. Given your changes from [f648981](https://github.com/Flagsmith/flagsmith/pull/5711/commits/f648981d9ae80ac3593e8ba19ec703ca7ec56809), I see you've picked option (1). In that case, please stick to `.get` — as per the suggested code in my comment above — or we might crash with a `KeyError` if the request is absent in the serializer context.

api/tests/unit/custom_auth/test_unit_custom_auth_views.py
Zaimwa9: Honestly for this one I would leave it as is:
- Without cookie (and utm), the hubspot tracker doesn't exist, but the team already mentioned that conditional assert were out of standard (`if param assert XX`), so it's basically testing the logic we want but being the workaround for the following type error `Item "None" of "HubspotTracker | None" has no attribute "hubspot_cookie"`
- breaking in 2 tests is a bit overkill and very redundant

IMO adding both a `or` and database query directly in the assert adds too much mental overhead in the context of a test.

If you feel the ternary is really a no-go I would split the test

api/tests/unit/integrations/lead_tracking/hubspot/test_unit_hubspot_lead_tracking.py
Zaimwa9: Yes, it's a relation with `HubspotOrganisation` that is basically just an id
```
class HubspotOrganisation(models.Model):
    organisation = models.OneToOneField(
        Organisation,
        related_name="hubspot_organisation",
        on_delete=models.CASCADE,
    )
    hubspot_id = models.CharField(max_length=100)
    updated_at = models.DateTimeField(auto_now=True)
```

api/tests/unit/custom_auth/test_unit_custom_auth_views.py
emyller: Having a ternary feels like hiding a code branch under the rug.

I really favor the suggested syntax, though I appreciate your feedback at the same time:

> adding both a or and database query directly in the assert adds too much mental overhead

How about breaking that into more readable steps? e.g.

```py
hubspot_cookie_matches = HubspotTracker.objects.filter(...).exists()
assert not hubspot_cookie or hubspot_cookie_matches
```

Zaimwa9: Matthew judged in between and suggest splitting the test

matthewelwell (approved):
Approving based on conversations with @Zaimwa9 and @emyller. 

Since all of the feedback from @emyller has been addressed, along with the change to split out the parameterised test, as originally suggested by @Zaimwa9 , the PR has been sufficiently reviewed and improved. Let's merge.

PR #5713, opened 2025-07-07
fix: Add missing Swagger schema for GET feature by ID

rolodato:

## Changes

Currently, there is no response schema for the "GET feature by ID" endpoint https://api.flagsmith.com/api/v1/docs/#/api/api_v1_projects_features_read

image

This PR adds the correct response schema:

image

```
DEBUG=True python manage.py runserver
```

then browse to http://127.0.0.1:8000/api/v1/docs/#/api/api_v1_projects_features_read

### DIFF
--- a/api/features/views.py
+++ b/api/features/views.py
@@ -113,7 +113,6 @@ class FeatureViewSet(viewsets.ModelViewSet):  # type: ignore[type-arg]
     def get_serializer_class(self):  # type: ignore[no-untyped-def]
         return {
             "list": ListFeatureSerializer,
-            "retrieve": ListFeatureSerializer,
             "create": ListFeatureSerializer,
             "update": UpdateFeatureSerializer,
             "partial_update": UpdateFeatureSerializer,
--- a/api/features/views.py
+++ b/api/features/views.py
@@ -205,9 +205,10 @@ class FeatureViewSet(viewsets.ModelViewSet):  # type: ignore[type-arg]
         instance.delete()
 
     def get_serializer_context(self):  # type: ignore[no-untyped-def]
-        if getattr(self, "swagger_fake_view", False):
-            return None
         context = super().get_serializer_context()
+        if getattr(self, "swagger_fake_view", False):
+            return context
+
         feature_states = getattr(self, "_feature_states", {})
         project = get_object_or_404(Project.objects.all(), pk=self.kwargs["project_pk"])
         context.update(
@@ -506,6 +507,8 @@ class BaseFeatureStateViewSet(viewsets.ModelViewSet):  # type: ignore[type-arg]
         """
         Override queryset to filter based on provided URL parameters.
         """
+        if getattr(self, "swagger_fake_view", False):
+            return FeatureState.objects.none()
 
         environment_api_key = self.kwargs["environment_api_key"]
 
--- a/api/features/views.py
+++ b/api/features/views.py
@@ -102,6 +102,10 @@ def get_feature_by_uuid(request, uuid):  # type: ignore[no-untyped-def]
     name="list",
     decorator=swagger_auto_schema(query_serializer=FeatureQuerySerializer()),
 )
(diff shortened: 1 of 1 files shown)


khvn26 (approved):

api/features/views.py
```diff
@@ -102,6 +102,10 @@ def get_feature_by_uuid(request, uuid):  # type: ignore[no-untyped-def]
     name="list",
     decorator=swagger_auto_schema(query_serializer=FeatureQuerySerializer()),
 )
+@method_decorator(
+    name="retrieve",
+    decorator=swagger_auto_schema(responses={200: ProjectFeatureSerializer()}),
+)
```
matthewelwell: This looks like a hack to work around an underlying issue somewhere in auto generating the docs? It also doesn't match the right serializer that's being returned in `get_serializer_class` below.

rolodato: I've removed the extra serializer class, which keeps the same behaviour and hopefully makes it a bit clearer. I have no idea what could be causing this to fail. I'll leave it up to you if we want to merge as-is and unblock the customer, or figure out what the underlying issue could be.

matthewelwell: > which keeps the same behaviour

As far as I can tell, this is not true. Currently, the endpoint to retrieve a feature by it's ID uses the `ListFeatureSerializer` and this PR changes it to use the `ProjectFeatureSerializer`, seemingle just because that's what was arbitrarily added in the swagger_auto_schema. If that's the purpose of the PR, I'd expect to see more information in the description about why we're changing the serializer.

rolodato: I've fixed the root cause of the issue, which was this warning shown when generating the Swagger docs:

```
drf_yasg.inspectors.base WARNING  view's FeatureViewSet raised exception during schema generation; use `getattr(self, 'swagger_fake_view', False)` to detect and short-circuit this
  File "/Users/rolodato/Library/Caches/pypoetry/virtualenvs/flagsmith-api-Avcq7jRJ-py3.12/lib/python3.12/site-packages/drf_yasg/inspectors/base.py", line 42, in call_view_method
    return view_method()
           ^^^^^^^^^^^^^
  File "/Users/rolodato/Library/Caches/pypoetry/virtualenvs/flagsmith-api-Avcq7jRJ-py3.12/lib/python3.12/site-packages/rest_framework/generics.py", line 113, in get_serializer
    kwargs.setdefault('context', self.get_serializer_context())
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/rolodato/source/flagsmith/flagsmith/api/features/views.py", line 210, in get_serializer_context
    project = get_object_or_404(Project.objects.all(), pk=self.kwargs["project_pk"])
                                                          ~~~~~~~~~~~^^^^^^^^^^^^^^
KeyError: 'project_pk'
```

This keeps the same serializer and doesn't add any overrides to the schema generation.

api/features/views.py
```diff
@@ -205,6 +205,8 @@ def perform_destroy(self, instance):  # type: ignore[no-untyped-def]
         instance.delete()
 
     def get_serializer_context(self):  # type: ignore[no-untyped-def]
+        if getattr(self, "swagger_fake_view", False):
+            return None
```
matthewelwell: This should be done after the call to super and return the result of it. See [here](https://github.com/Flagsmith/flagsmith/pull/5724)

api/features/views.py
```diff
@@ -504,8 +506,6 @@ def get_queryset(self):  # type: ignore[no-untyped-def]
         """
         Override queryset to filter based on provided URL parameters.
         """
-        if getattr(self, "swagger_fake_view", False):
-            return FeatureState.objects.none()
```
matthewelwell: Why was this removed? I think this just moves the swagger generation warning from `get_serializer_context` to here instead.

matthewelwell (approved):

PR #5719, opened 2025-07-08
fix: User display name

### LINKED ISSUE #5710: If user first and last name is missing from user profile, the assignee in change request is blank

- [ ] Self Hosted with Docker
- [x] SaaS at flagsmith.com

### Describe the bug

When you assign a user to a change request, and that user is missing their name information, the assignee on the change request is blank, which is confusing UX. It's somewhat of an edge case, however it's entirely possible to create an account missing this information via SAML. 

### Steps To Reproduce

Create an account via SAML that is missing first name and last name. 

Create that change request and assign it to that user. 

### Expected behavior

Rather than appear blank, it should perhaps default to displaying the user email instead. 

### Screenshots

Image

Image

kyle-ssg:

## Changes

Some users may not have a first name / last name set, there were various touch points where we were relying on it. Instead, this creates a generic approach where we fallback to email.

Viewed touchpoints as a user without a name 
image


frontend/common/utils/utils.tsx
```diff
@@ -278,58 +279,6 @@ const Utils = Object.assign({}, require('./base/_utils'), {
       timeUnit: TimeUnit.MINUTE,
     }
   },
-  getExistingWaitForTime: (
```
kyle-ssg, line 281: Note: not sure what happened here but there were 3 instances of this function

Zaimwa9 (approved):

PR #5733, opened 2025-07-08
fix: Integrations size

kyle-ssg:

## Changes

Before: 

image

After: 

image

_Please describe._

### DIFF
--- a/frontend/web/components/IntegrationSelect.tsx
+++ b/frontend/web/components/IntegrationSelect.tsx
@@ -99,7 +99,7 @@ const IntegrationSelect: FC<IntegrationSelectType> = ({ onComplete }) => {
               ))}
             </div>
           </div>
-          <div className='col-md-9 h-100'>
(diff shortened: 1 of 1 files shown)


tiagoapolo (dismissed):

matthewelwell:
In my opinion, this definitely helps, but I'd rather see a few other changes as well. 

Specifically:

 1. I'd like to make the header / sub header more obvious by bringing it into the main container
 2. I think that we should move the skip / confirm buttons to the top as well so it's easier for people to skip if they want (although maybe this is counter-intuitive and will discourage people from using the new functionality)

Mock up to explain my suggestions:

Screenshot 2025-07-11 at 09 18 58

kyle-ssg:
Ok yep makes sense, here's what this looks like 

image

matthewelwell:
> Ok yep makes sense, here's what this looks like
> 
> image

Awesome, this looks so much better! Thanks!

matthewelwell (approved):

PR #5739, opened 2025-07-09
fix: rely-solely-on-hubspot-to-create-organisations

Zaimwa9:

## Changes
Hubspot does warn about XX seconds delay between a record is created and it's availability in their search API. It has created duplicates because on our side, we trigger `create_lead` that:
- Get hubspot_contact_id and sync our DB if not exists
- Get hubspot_org_id and sync our DB if not exists
- ~~Used to create the company in hubspot if not found~~ This part is to be removed to trust hubspot association 

Also #5711  should reduce this case from happening by adding a delay before triggering `create_lead`

This PR aims to remove the fallback of creating the organisation in hubspot if we couldn't find it by domain

- Updated tests


api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -179,21 +181,11 @@ def _get_or_create_organisation_hubspot_id(
 
         return org_hubspot_id
 
-    def _get_or_create_hubspot_company(
+    def _get_company_from_hubspot_by_domain(
```
emyller:
```suggestion
    def _get_hubspot_company_by_domain(
```

Sounds slightly better to me, though not a strong suggestion.

api/tests/unit/integrations/lead_tracking/hubspot/test_unit_hubspot_lead_tracking.py
```diff
@@ -190,16 +190,8 @@ def test_create_organisation_lead_creates_contact_when_not_found(
...
-        name=organisation.name,
-        active_subscription="free",
-        organisation_id=organisation.id,
-        domain="example.com",
-    )
-    mock_client.associate_contact_to_company.assert_called_once_with(
-        contact_id=HUBSPOT_USER_ID,
-        company_id=HUBSPOT_COMPANY_ID,
-    )
+    mock_client.create_company.assert_not_called()
```
emyller: Can we add a comment explaining why we assert for not called?
```suggestion
    mock_client.create_company.assert_not_called()  # We rely on Hubspot creating contacts
```

...or perhaps a link to the issue.

api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -156,10 +156,12 @@ def _get_or_create_organisation_hubspot_id(
...
         company_kwargs["active_subscription"] = organisation.subscription.plan
 
-        # As Hubspot creates/associates companies based on contact domain
+        # As Hubspot creates/associates companies automatically based on contact domain
         # we need to get the hubspot id when this user creates the company for the first time
         # and update the company name
-        company = self._get_or_create_hubspot_company(**company_kwargs)
+        company = self._get_company_from_hubspot_by_domain(domain)
+        if not company:
+            return None
```
emyller, line 164: How can a company not exist in Hubspot at this point? When it happens, what should we do?

Zaimwa9: There is a small indexation delay. Duplicated companies has been observed empirically and a case happened last week (I can share offline if you are interested).
With hubspot creating/associating under the hood the contact with the company. And this logic getting executed within 10 seconds => company not available from their API.
[doc](https://developers.hubspot.com/docs/guides/api/crm/search#limitations).
`It may take a few moments for newly created or updated CRM objects to appear in search results.`

So conclusion discussed with Uriel is to let it go, not have action on our side and rely only on hubspot engine. It's acceptable that our DB doesn't reflect 100% hubspot in similar cases

**More details**
Unique/clean data in hubspot > Our DB synchronization because before any update in hubspot (especially `update_active_subscription`) we:
1. Check `HubspotOrganisation`
2. If not synchronized `get_company_by_domain` and create `HubspotOrganisation` for next time (and obviously in those later steps it should be available)

We don't make so much use of the `hubspot_organisation_id` besides those operations.

Last thing, in this [one](https://github.com/Flagsmith/flagsmith/pull/5711), there is an additional 2 (or 3) minutes buffer before triggering this logic, that should mitigate the risks of not finding the record

emyller: Gotcha, thank you for the explanation.

Couple things, if _letting go_ is indeed the business decision here:
1. Let's add a comment so this `return None` isn't prone to the reader's imagination. 😅 
2. Question: can _letting go_ possibly lead to inconsistencies in the CRM layer?

api/tests/unit/integrations/lead_tracking/hubspot/test_unit_hubspot_lead_tracking.py
```diff
@@ -190,16 +190,8 @@ def test_create_organisation_lead_creates_contact_when_not_found(
...
-        active_subscription="free",
-        organisation_id=organisation.id,
-        domain="example.com",
-    )
-    mock_client.associate_contact_to_company.assert_called_once_with(
-        contact_id=HUBSPOT_USER_ID,
-        company_id=HUBSPOT_COMPANY_ID,
-    )
+    mock_client.create_company.assert_not_called()  # We rely on Hubspot creating contacts
+    mock_client.associate_contact_to_company.assert_not_called()
```
emyller, line 194: The changes make sense. Can we also update the test name accordingly?

matthewelwell (approved):
Approving based on conversations with @emyller and @Zaimwa9 similar to the comment left on #5711

PR #5743, opened 2025-07-10
refactor: only clone latest feature states

matthewelwell:
## Changes

This PR changes the behaviour when cloning an environment. Previously, the logic would create a clone of _all_ feature states, including old versions, feature states associated with uncommitted change requests, etc. This PR updates the logic to only clone the latest feature states, and mark them as live from the moment the environment is created. 

This does change the user experience, but only really for those using v2 versioning (which is currently very few customers), in that they will no longer see a history of changes in the new environment. In my opinion, it was odd behaviour to show the history from another environment in the new environment anyway, so I think this is an improvement. 

Refactor environment clone functionality to only clone the latest feature states. This will optimise the logic to speed it up reduce the impact of cloning environments on the database. I've also included some additional DB optimisations using select and prefetch related where necessary. 

Note that this also fixes a bug in the logic for enabling versioning on new environments which existed because we were previously using the `enable_v2_versioning` task to create the versions, but this expected the input environment to not have versioning enabled. 

Existing tests should cover the refactor. Updated an existing test with better assertions to verify the bug fix, and new logic.

### DIFF
--- a/api/environments/tasks.py
+++ b/api/environments/tasks.py
@@ -84,10 +84,10 @@ def clone_environment_feature_states(
         additional_filters=Q(identity__isnull=True),
     )
 
-    # Since we only want to create a single version for each environment / feature
-    # combination in the new environment to create a 'snapshot' of the source
-    # environment, we keep a local cache of environment feature version objects
-    # to avoid having to use get_or_create and hit the db unnecessarily.
+    # Since, in versioned environments, we only want to create a single version for
+    # each feature to create a 'snapshot' of the source environment, we keep a local
+    # cache of EnvironmentFeatureVersion objects to avoid having to use get_or_create
+    # and hit the db unnecessarily.
     version_map: dict[int, EnvironmentFeatureVersion] = {}
 
     for feature_state in source_feature_states:
--- a/api/environments/tasks.py
+++ b/api/environments/tasks.py
@@ -88,17 +88,17 @@ def clone_environment_feature_states(
     # each feature to create a 'snapshot' of the source environment, we keep a local
     # cache of EnvironmentFeatureVersion objects to avoid having to use get_or_create
     # and hit the db unnecessarily.
-    version_map: dict[int, EnvironmentFeatureVersion] = {}
+    efv_by_feature_id: dict[int, EnvironmentFeatureVersion] = {}
 
     for feature_state in source_feature_states:
         kwargs = {"env": clone}
 
         if clone.use_v2_feature_versioning:
-            if not (efv := version_map.get(feature_state.feature_id)):
+            if not (efv := efv_by_feature_id.get(feature_state.feature_id)):
                 efv = EnvironmentFeatureVersion.create_initial_version(
                     environment=clone, feature=feature_state.feature
                 )
-                version_map[feature_state.feature_id] = efv
+                efv_by_feature_id[feature_state.feature_id] = efv
 
             kwargs.update(environment_feature_version=efv)
         else:
--- a/api/environments/tasks.py
+++ b/api/environments/tasks.py
@@ -1,5 +1,3 @@
-from datetime import timedelta
-
 from django.db.models import Prefetch, Q
 from django.utils import timezone
 from task_processor.decorators import (
@@ -64,7 +62,7 @@ def delete_environment(environment_id: int) -> None:
     Environment.objects.get(id=environment_id).delete()
 
 
-@register_task_handler(timeout=timedelta(hours=1))
+@register_task_handler()
 def clone_environment_feature_states(
     source_environment_id: int, clone_environment_id: int
 ) -> None:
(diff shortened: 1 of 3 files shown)


api/environments/tasks.py
```diff
@@ -64,39 +67,44 @@ def clone_environment_feature_states(
...
+        if efv := feature_state.environment_feature_version:
+            efv_clone = efv.clone_to_environment(environment=clone)
+            fs_clone_kwargs.update(environment_feature_version=efv_clone)
+        elif clone.use_v2_feature_versioning:
+            efv = EnvironmentFeatureVersion.objects.create(
+                environment=clone,
+                feature=feature_state.feature,
+                live_from=feature_state.live_from,
+                published_at=now,
+            )
```
khvn26: Why not

```suggestion
            efv = EnvironmentFeatureVersion.create_initial_version(
                environment=clone,
                feature=feature_state.feature,
            )
```

How is `EnvironmentFeatureVersion(live_from=feature_state.live_from)` different from `EnvironmentFeatureVersion(live_from=None)`?

matthewelwell: Manual testing has shown that this logic doesn't really work when there are multiple feature states for a given feature in the source environment (e.g. when there are segment overrides). It makes it look like, in the cloned environment, there are changes stretching back to when those feature states were created, when really I think what we want is just to have a version of the feature that exists at the time of the cloning, and that is v0. 

As such, I'll have to rework some of this (and likely utilise your suggestion), but I'll have to do it outside the loop over the feature states so we only create one EnvironmentFeatureVersion object for each environment / feature pair.

api/environments/tasks.py
```diff
@@ -64,39 +69,42 @@ def clone_environment_feature_states(
...
-            feature_state.clone(clone, live_from=feature_state.live_from)
+        ],
+        additional_filters=Q(identity__isnull=True),
+    )
+
+    # Since, in versioned environments, we only want to create a single version for
+    # each feature to create a 'snapshot' of the source environment, we keep a local
+    # cache of EnvironmentFeatureVersion objects to avoid having to use get_or_create
+    # and hit the db unnecessarily.
+    version_map: dict[int, EnvironmentFeatureVersion] = {}
```
khvn26: nit

```suggestion
    efv_by_feature_id: dict[int, EnvironmentFeatureVersion] = {}
```

Minor change, but helps a lot to understand the code.

api/tests/unit/environments/test_unit_environments_models.py
```diff
@@ -1252,16 +1252,26 @@ def test_environment_clone_from_non_versioned_environment_with_use_v2_feature_ve
     project: Project,
     environment: Environment,
     feature: Feature,
+    segment_featurestate: FeatureState,
     enable_v2_versioning_for_new_environments: typing.Callable[[], None],
 ) -> None:
     # Given
+    assert not environment.use_v2_feature_versioning
```
khvn26: Seems more resilient/correct:

```suggestion
    environment.use_v2_feature_versioning = False
    environment.save()
```

khvn26 (commented):
Two minor suggestions. The logic looks fine to me.

api/tests/unit/environments/test_unit_environments_models.py
matthewelwell: Hmm... I'm not sure I agree with this. The assertion is there to ensure that there's nothing going awry with the fixture and I want to know about it if there is, and implementing the change suggested here would just mask it.

api/environments/tasks.py
matthewelwell: Resolved here: https://github.com/Flagsmith/flagsmith/pull/5743/commits/1f6ff5dd0bed94459e1c7296a36c27a29d892ee4

api/tests/unit/environments/test_unit_environments_models.py
khvn26: I do not understand why having v2 versioning enabled would mean that the fixture went awry?

matthewelwell: The `environment` fixture should not have v2 versioning enabled - that's what we have the `environment_v2_versioning` fixture for. The fixture I'm talking about going awry (or acting out of turn, more specifically) is the `enable_v2_versioning_for_new_environments` one in the case that it somehow mocks the flag before I want it to. Obviously that's not the case unless something changes, but I really want to know if anything changes because this test is moot if `use_v2_feature_versioning == True`.

khvn26: > The environment fixture should not have v2 versioning enabled - that's what we have the environment_v2_versioning fixture for

Why shouldn't it? Isn't the end goal to have v2 versioning enabled by default?

matthewelwell: Yes, but that will involve a much larger change than just this test. 

Either way, I've just added your suggestion, despite my aversion to modifying fixtures like this in tests.

khvn26 (approved):

Issue #2183, opened 2023-05-04
Improve last login data


matthewelwell:
**Is your feature request related to a problem? Please describe.**

The last login data only shows data for users that have logged in rather than accessed the platform. 

**Describe the solution you'd like**

Last login should reflect when users last accessed the platform. 

**Describe alternatives you've considered**

N/a

**Additional context**

Related to [this PR](https://github.com/Flagsmith/flagsmith/pull/1958) and [this issue](https://github.com/Flagsmith/flagsmith/issues/1854).

zachaysan:
Steps to reproduce the issue:

1. Be an organisation admin.
2. Copy an invite link.
3. Paste the link in a different browser or incogneto window.
4. Complete signup (which logs the user in at the same time).
5. Go back to the organisation admin browser and go to the `members` tab of the manage page, they now all say "Never" despite being logged in.

<img width="1083" alt="image" src="https://github.com/Flagsmith/flagsmith/assets/222715/d03ecacb-45fb-4ab6-8ae2-2be5f412ca3b">

matthewelwell:
Suggestion from @Zaimwa9 is to just update the last_login field on the user on a new call to `/me` endpoint.

PR #5745, opened 2025-07-10
feat: improve last login data

Zaimwa9:

## Changes
Solves #2183 

last_login was natively updated by DRF only during authentication. However we persist session for multiple weeks leading to incorrect data.

This PR aims to update last-login on `GET /me` (once per session in a ideal world) with a 10min buffer. This could be multiple hours as we only show `Today` in the dashboard.

Additionally: 
- Refactored LastLogin component in the frontend
We used to show `Never | within 30 days | with {xx} days if > 30` => Allowed to add `Today` `Yesterday` and number of days within 30 days (`7 days ago` e.g). 
- Refactored Members Table in the frontend

- Added tests
- On the dashboard


api/custom_auth/views.py
```diff
@@ -141,6 +143,17 @@ def perform_destroy(self, instance):  # type: ignore[no-untyped-def]
             )
         )
 
+    def retrieve(self, request: Request, *args: Any, **kwargs: Any) -> Response:
+        user = request.user
+        assert isinstance(user, FFAdminUser)
+        if not user.last_login or timezone.now() - user.last_login > timedelta(
+            minutes=10
```
Zaimwa9: This could be couple of hours

frontend/web/components/users-permissions/OrganisationUsersTable/components/LastLogin.tsx
```diff
@@ -0,0 +1,47 @@
...
+      return { relativeDate: 'Never' }
+    }
+
+    const diff = moment().diff(loginMoment, 'days')
+
+    if (diff === 0) return { relativeDate: 'Today' }
+    if (diff === 1) return { relativeDate: 'Yesterday' }
+
+    return {
+      date: diff > 30 ? moment(last_login).format('Do MMM YYYY') : undefined,
```
Zaimwa9, line 29: As before, we show the full date only if > 30 days
image
image

api/custom_auth/views.py
```diff
@@ -141,6 +143,17 @@ def perform_destroy(self, instance):  # type: ignore[no-untyped-def]
             )
         )
 
+    def retrieve(self, request: Request, *args: Any, **kwargs: Any) -> Response:
```
emyller, line 146: I'm not sure this is the right context in which to update the user's `last_login`. I wonder, is there somewhere more appropriate to implement authentication side-effects?

api/custom_auth/views.py
khvn26: Suggestion: make this configurable.

khvn26 (approved):
One minor comment, otherwise LGTM.

api/custom_auth/views.py
khvn26: @emyller IMO:

1. It's clear enough from the issue description why the `/me` endpoint is being used.
2. Considering doing this from the `/me` endpoint is a requirement, this is a fairly comprehensible and straightforward implementation.

api/tests/unit/users/test_unit_users_views.py
```diff
@@ -930,3 +932,27 @@ def test_list_user_groups(
         (user1.pk, False),
         (user2.pk, True),
     }
+
+
+@pytest.mark.django_db
+@freeze_time("2024-01-01T10:00:00Z")
+def test_get_me_view_updates_last_login(
```
khvn26, line 952: Let's add a test proving that we don't update the field on every call.

khvn26 (changes requested):
Actually, having read the backend test, I have one request now.

api/custom_auth/views.py
emyller: @khvn26 I can't remember reading this earlier. 🤔  Nevertheless you're right.

> last_login was natively updated by DRF only during authentication. However we persist session for multiple weeks leading to incorrect data.

api/custom_auth/views.py
Zaimwa9: Done and increased to 30minutes

api/tests/unit/users/test_unit_users_views.py
```diff
@@ -930,3 +932,38 @@ def test_list_user_groups(
...
+@pytest.mark.django_db
+@freeze_time("2024-01-01T10:00:00Z")
+@pytest.mark.parametrize(
+    "last_login, expected_last_login",
+    [
+        (None, lambda: timezone.now()),
+        ("2023-01-01T10:00:00Z", lambda: timezone.now()),
+        ("2024-01-01T09:59:00Z", lambda: timezone.now() - timedelta(minutes=1)),
+    ],
+)
```
khvn26: I fail to see see how that proves that we use the `LAST_LOGIN_UPDATE_THRESHOLD_MINUTES`?

api/tests/unit/users/test_unit_users_views.py
```diff
@@ -930,3 +932,45 @@ def test_list_user_groups(
...
+    # When
+    response = api_client.get(url)
+
+    # Then
+    assert response.status_code == status.HTTP_200_OK
+    test_user.refresh_from_db()
+
+    expected_last_login = (
+        timezone.now() if should_update_last_login else test_user.last_login
+    )
```
khvn26: I disapprove of conditional logic in tests, particularly in the assertion ("Then") part. It's less readable and more prone to errors. In general, we should avoid it.

This particular branching can be avoided by parametrising `expected_last_login`:

```python
@freeze_time("2024-01-01T10:00:00Z")
@pytest.mark.parametrize(
    "last_login,expected_last_login",
    [
        (None, "2024-01-01T10:00:00Z"),
        ("2023-01-01T10:00:00Z", "2023-01-01T10:00:00Z"),
        ("2024-01-01T09:59:00Z", "2024-01-01T10:00:00Z"),
    ],
)
```

khvn26 (approved):

PR #5750, opened 2025-07-14
fix: e2e not able to connect to db

tiagoapolo:

## Changes

Screenshot 2025-07-14 at 12 00 49

We were using `postgresql:15` and not specifying a minor, a breaking change was introduced in 15.13-13
https://github.com/cloudnative-pg/postgres-containers/issues/248

1. Go to `frontend` folder
2. Run `docker-compose -f docker-compose-e2e-tests.yml up`
3. Check if migrations are being applied, meaning the app was able to connect to DB

### DIFF
--- a/frontend/docker-compose-e2e-tests.yml
+++ b/frontend/docker-compose-e2e-tests.yml
@@ -6,7 +6,7 @@
 version: '3'
 services:
   db:
-    image: ghcr.io/cloudnative-pg/postgresql:15
+    image: ghcr.io/cloudnative-pg/postgresql:15.13-12
     environment:
       POSTGRES_PASSWORD: password
       POSTGRES_DB: flagsmith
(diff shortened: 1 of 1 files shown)


frontend/docker-compose-e2e-tests.yml
```diff
@@ -6,7 +6,7 @@
 version: '3'
 services:
   db:
-    image: ghcr.io/cloudnative-pg/postgresql:15
+    image: ghcr.io/cloudnative-pg/postgresql:15.13-12
```
emyller: I don't like the minor version pinning because it looks like a fix to someone else's bug. 🤔

Another option is to pull the official Postgres image from Docker Hub (suggested code). I also don't love it because it's prone to some latency and throttling from docker.io though I like it better than relying on a broken Docker image.

```suggestion
    image: docker.io/library/postgres:15-alpine
```

Can we please give that a try?

tiagoapolo: I have nothing against it. We can try

khvn26 (approved):

emyller (approved):
docker.io seems like the right option for now.

PR #5763, opened 2025-07-16
feat: Release pipelines logic v0.2.1

### LINKED ISSUE #5638: As a dashboard user, I should be able to track pipeline actions through audit logs
Audit logs should be generated for the following actions:
1. Creating a release pipeline
2. publishing a release pipeline
3. Editing a release pipeline 
4. Deleting a release pipeline
5. Adding a feature to the release pipeline
6. For feature moving through the release pipeline

gagantrivedi:

## Changes

- Add unpublish release pipelines API
- Add clone release pipeline API
- Generate audit logs for update release pipeline 
Closes: https://github.com/Flagsmith/flagsmith/issues/5638

Adds unit tests

### DIFF
--- a/api/features/release_pipelines/core/models.py
+++ b/api/features/release_pipelines/core/models.py
@@ -1,4 +1,6 @@
 import typing
+import uuid
+from copy import deepcopy
 
 from django.core.validators import MaxValueValidator
 from django.db import models
@@ -7,9 +9,7 @@ from django.utils import timezone
 from audit.constants import (
     RELEASE_PIPELINE_CREATED_MESSAGE,
     RELEASE_PIPELINE_DELETED_MESSAGE,
-    RELEASE_PIPELINE_PUBLISHED_MESSAGE,
 )
-from audit.models import AuditLog
 from audit.related_object_type import RelatedObjectType
 from core.models import (
     SoftDeleteExportableModel,
@@ -17,6 +17,7 @@ from core.models import (
 )
 from features.release_pipelines.core.constants import MAX_PIPELINE_STAGES
 from features.release_pipelines.core.exceptions import InvalidPipelineStateError
+from features.versioning.models import EnvironmentFeatureVersion
 from projects.models import Project
 from users.models import FFAdminUser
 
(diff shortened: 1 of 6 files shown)


api/features/release_pipelines/core/models.py
```diff
@@ -90,18 +108,14 @@ def get_delete_log_message(
     ) -> typing.Optional[str]:
         return RELEASE_PIPELINE_DELETED_MESSAGE % self.name
 
+    def has_feature_in_flight(self) -> bool:
+        return EnvironmentFeatureVersion.objects.filter(  # type: ignore[no-any-return]
```
Zaimwa9: It pleases mypy to assign the variable with type to avoid the type ignore:
```
       has_feature_in_flight: bool = EnvironmentFeatureVersion.objects.filter(
            published_at__isnull=True, pipeline_stage__in=self.stages.all()
        ).exists()
        return has_feature_in_flight
```

api/features/release_pipelines/core/models.py
```diff
@@ -72,7 +73,24 @@ def publish(self, published_by: FFAdminUser) -> None:
...
-        self._create_pipeline_published_audit_log()
+
+    def unpublish(self, unpublished_by: FFAdminUser) -> None:
+        if self.published_at is None:
+            raise InvalidPipelineStateError("Pipeline is not published.")
+        self.published_at = None
+        self.published_by = None
+        self.save()
+
+    def clone(self) -> "ReleasePipeline":
```
Zaimwa9: As it was raised by emyller in this [one](https://github.com/Flagsmith/flagsmith/commit/365ad46717cc4177936c5839eba272150e5a650c).

It doesn't feel right to perform the cloning in the models as deep down it's logic business and it ends up with:
- bloating the models
- having the logic scattered between the pipeline, stage, trigger and action models
while it indeed makes sense to have a service triggering the pipeline cloning and cascading down. 

I'd be in favor to use a service

Zaimwa9: Do we plan to add the new name in the pipeline (as for segment?). Do we want to implement it now in the backend in that case?

Zaimwa9 (changes requested):
Minor comment and open discussion

api/features/release_pipelines/core/models.py
gagantrivedi: Hmm, interesting point. I'd argue that there's not much business logic here—it's as simple as implementing a copy method, which should live on the class itself to be idiomatic Django. But I do see your point, and it leaves us with two ways of creating a duplicate of a model in our codebase. To avoid that, I'd move this to a service layer on release pipelines.

gagantrivedi: Okay, Done here: https://github.com/Flagsmith/flagsmith-private/pull/83

Zaimwa9 (dismissed):
👍

Zaimwa9 (approved):

PR #5764, opened 2025-07-16
chore: Update and repurpose evaluation context schema for Context values

khvn26:
Thanks for submitting a PR! Please check the boxes below:

- [ ] I have added information to `docs/` if required so people know about the feature!
- [x] I have filled in the "Changes" section below?
- [x] I have filled in the "How did you test this code" section below?
- [x] I have used a [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) title for this Pull Request

## Changes

This updates the evaluation context schema to reflect Flagsmith Engine dependencies.

The schema now represents an evaluation object used by the Flagsmith Engine when parsing Context values introduced by #5673.

## How did you test this code?

N/A

Issue #5675, opened 2025-06-30
Frontend: Implement Context values for Segments


khvn26:
Blocked on https://github.com/Flagsmith/flagsmith/issues/5674.

## DoD

![Recording 2025-06-25 at 18 35 21](https://github.com/user-attachments/assets/d5df054c-aa71-4567-9088-560466842e87)

As a user,
When inputting segment condition propertie in the Segment flyout for all operators,
I'm able to choose from the following pre-set values in the Trait input text box:
- Environment Name
- Identifier

As a user,
When inputting segment conditions in the Segment flyout for all operators,
I'm able to enter an arbitrary Trait key in the Trait input text box, in addition to the pre-set Context values.

- The 'is set' and 'is not set' operators should show a warning if the user selects a pre-set context value as they will always evaluate to true / false respectively. 

## Notes

UX mockup/docs: #5654

Zaimwa9:
@matthewelwell, to keep you updated, we came up with some UI arrangements with Kim to make it more understandable.
So we can eliminate the tooltip that is quite noisy/intrusive and include it within the dropdown

<img width="213" height="291" alt="Image" src="https://github.com/user-attachments/assets/7f82a6c7-c1f8-44c4-90af-3778c5826926" />

PR #5766, opened 2025-07-16
feat: frontend context values in segment

Zaimwa9:

## Changes
Related: #5764 and #5675 

⚠️ will still need some interfacing with API once ready

- Allow to select pre-defined context values (`identifier` or `environment_name` for now) as condition properties.
- Implemented an input-dropdown component usable when creating/editing rules of a segment
- Migrated related components to functional components
- Removed the tooltip on the input component

image

image

image

- E2E tests


khvn26:
> Traits are not supported for the selected operator

Oh they certainly are going to be supported, no need to impose the restriction here.

frontend/web/components/segments/Rule/components/RuleConditionPropertySelect.tsx
```diff
@@ -0,0 +1,157 @@
...
+  ruleIndex,
+  setRuleProperty,
+}: RuleConditionPropertySelectProps) => {
+  const [localCurrentValue, setLocalCurrentValue] = useState(propertyValue)
+
+  const isContextPropertyEnabled =
+    Utils.getFlagsmithHasFeature('context_values')
+  const contextValues: OptionType[] = JSON.parse(
+    Utils.getFlagsmithValue('context_values') || '{}',
+  )
```
Zaimwa9: I'm not convinced with this approach. We loose in control / validation and i'm not sure about the upside.
Maybe use it to disable specific values if needed?

frontend/web/components/segments/Rule/components/RuleConditionRow.tsx
```diff
@@ -0,0 +1,167 @@
...
+          <Flex className='or-divider__line' />
+        </Row>
+      )}
+      <Row noWrap className='rule align-items-center justify-content-between'>
+        <RuleConditionPropertySelect
+          dataTest={`${dataTest}-property-${ruleIndex}`}
+          ruleIndex={ruleIndex}
+          setRuleProperty={setRuleProperty}
+          propertyValue={rule.property}
+          isTraitDisabled={operatorObj?.value === 'PERCENTAGE_SPLIT'}
```
khvn26: This is not needed. As part of https://github.com/Flagsmith/flagsmith-engine/issues/219 we're unlocking all operators for any value types, including traits.

frontend/web/components/segments/Rule/components/RuleConditionPropertySelect.tsx
khvn26: I wonder if we can utilise the evaluation context schema: https://github.com/Flagsmith/flagsmith/pull/5764

khvn26 (changes requested):

frontend/web/components/segments/Rule/components/RuleConditionPropertySelect.tsx
Zaimwa9: A bit complicated in its current shape because it doesn't work with the json-path we are using.

For identifier we would need something like `$.defs.IdentityEvaluationContext.properties.identifier`.title so it looks like quite a lot of gymnastic

khvn26: Jsonschema doesn't seem too hard to parse? https://stackoverflow.com/a/60297881

khvn26: As discussed offline, [added custom properties to the schema](https://github.com/Flagsmith/flagsmith/pull/5764/commits/71d107f203d4ef68f54fe6a5a7a68e7fa692fdd1) to simplify the schema parsing. It shouldn't be a problem to use titles and descriptions from the schema now.

frontend/web/components/segments/Rule/Rule.tsx
```diff
@@ -0,0 +1,169 @@
...
+      const isPropertyContext = [
+        RuleContextValues.IDENTIFIER?.toString(),
+        RuleContextValues.ENVIRONMENT_NAME?.toString(),
+      ].includes(condition.property)
+
+      if (!isPropertyContext) {
+        updates.property = ''
+        updates.value = ''
+      }
+    }
```
khvn26: We are now allowing `% Split` usage with traits as well, so this code is not needed, I believe.

We need, however, to assume the `Identity Key` context value as the default for `% Split` conditions that have no property.

frontend/web/components/segments/Rule/components/RuleConditionPropertySelect.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+  const isValueFromContext = !!contextOptions.find(
+    (option) => option.value === localCurrentValue,
+  )?.value
+
+  const displayedLabel =
+    contextOptions.find((option) => option.value === propertyValue)?.label ||
+    propertyValue
+
+  const optionsWithTrait = [
```
tiagoapolo: nit: Can we place these lists into separate variables to make it easier to read?

tiagoapolo (commented):
Code-wise it looks good, I'd just try to work on this [suggestion](https://github.com/Flagsmith/flagsmith/pull/5766/files#r2219736544) from kim

tiagoapolo (dismissed):

khvn26 (approved):

Issue #5769, opened 2025-07-16
Move segment create/update logic back to core


emyller:
Because Change Requests are restricted to only some customers, the functionality around it lives in a separate repository ([flagsmith-workflows](https://github.com/Flagsmith/flagsmith-workflows)). And because we're now associating segments to change requests, we decided to move some segment functionality into flagsmith-common (see [pull request](https://github.com/Flagsmith/flagsmith-common/pull/4/files)) so that it could be reused from both flagsmith-workflows, and also from the core API.

After realizing some performance and consistency pitfalls in the segregated approach, we’ll now 1. move segment logic back to the core API, and 2. delete code that overrides the original serializer functionality in order to rules and conditions — might need help from [drf-writable-nested](https://github.com/beda-software/drf-writable-nested).

flagsmith-workflows will only trigger functionality from the core API code. It does not need to include tests for core functionality.

flagsmith-common will not be necessary in this context.

The overall architecture of our codebase is likely to be favored with this change.

emyller:
Blocked on #5846

Issue #5771, opened 2025-07-16
Commit draft segment change requests


emyller:
User story:

> As a dashboard user, while reviewing a change request, I can approve and publish (commit) the changes proposed in segments.

We need to replace the existing logic for publishing segment changes so that it will:

- Make a copy of the live segment for historical purposes with a new `Segment.clone` method, and get rid of `SegmentCloneService` and existing `Segment.*clone` methods entirely. The new function must create a revision of the segment, as well as its nested rules and conditions.
- Overwrite **attributes of** segments, then replace all nested rules and conditions.

emyller:
The new clone code was implemented in [this effort](https://github.com/Flagsmith/flagsmith/pull/5748), later replaced with #5768 (parent of this issue).

Issue #5770, opened 2025-07-16
Point draft segment rules and conditions to their live counterparts


emyller:
User story:

> As a dashboard user, while reviewing a change request, I can see the difference between the rules and conditions of the live segment, and the proposed one.
> 

**In flagsmith:**

Need to introduce a new `version_of` attribute to rules and conditions, similarly to `Segment.version_id` ([ref](https://github.com/Flagsmith/flagsmith/blob/3e7dc44f86594613cfa1fb1d09b40397ef438c4a/api/segments/models.py#L71)).

Rules and conditions associated to draft segments and historical segments should always point to their live counterparts.

**In flagsmith-workflow:**

The endpoint `/api/v1/projects/<id>/change-requests/` must **require** in the HTTP payload, when pointing to existing objects in the database:

- `segments[].version_of` (always present)
- `segments[].rules[].version_of`
- `segments[].rules[].conditions[].version_of`
- `segments[].rules[].rules[].conditions[].version_of`

Additionally, because a **new segment** is going to be created as *draft*, as well as nested rules and conditions, the payload **must not include any ID field**, as to not replace existing objects.

Therefore, the value of `version_of` in each entity must point to the ID of its counterpart in the live segment.

The serializer will:

- Reject any “id” field in segments, rules, or conditions.
- Accept “version_of” in segments, rules, and conditions.

---

**Glossary**

*Draft segments* are segments created along with change requests.

*Historical segments* are segments created as a revision when updating a segment, either directly, or via a change request.

emyller:
The new fields were implemented in [this effort](https://github.com/Flagsmith/flagsmith/pull/5748), later replaced with #5768 (parent of this issue).

emyller:
Blocked on #5846

emyller:
We're not going to pursue associating rules and conditions to their previous version. Because the frontend will do text-based diffing, we can go ahead and wipe rules and conditions from a segment before saving new ones. In the future, we'll also modify the data model so that rules and conditions are persisted as JSON blobs.

Issue #5768, opened 2025-07-16
Make segment change requests work


emyller:
Change Requests are currently only used to propose changes to feature flags. We also want to allow users to propose changes to **segment rules**, i.e. *segment change requests* — sometimes referred to as *project change requests*, due to the segment change request belonging to a project rather than an environment, such is the case of feature change requests.

Well actually, the feature [already exists](https://github.com/Flagsmith/flagsmith/pull/4265), but it's **not yet live because the code can't make reference of previous rules and conditions for diffing purposes**. A [pull request](https://github.com/Flagsmith/flagsmith/pull/4988) was submitted to help solve that problem, though we decided to revisit the whole implementation and make fundamental changes to avoid performance and consistency pitfalls — to be documented as sub issues of this epic.

The journey *as of this writing*:

| Merged | What | Pull request |
| --- | --- | --- |
| 2024-10-22 | Segment-specific logic was moved to flagsmith-common. | https://github.com/Flagsmith/flagsmith-common/pull/4 |
| 2024-11-21 | Change Requests now imports from flagsmith-common to manage segments. | https://github.com/Flagsmith/flagsmith-workflows/pull/39 |
| 2024-11-21 | Core API implements segment change requests. | https://github.com/Flagsmith/flagsmith/pull/4265 |
| **Open** | Frontend for segment change requests. It's hidden under the FoF flag `segment_change_requests` and should be merged when the backend is working. | https://github.com/Flagsmith/flagsmith/pull/4852 |

There was an ongoing effort to enable comparing each rule and condition of a *draft* segment — i.e. the segment revision being proposed in a change request — against the **live** segment. Relevant work:

| Merged | What | Pull request |
| --- | --- | --- |
| 2025-01-27 | Prepares rules and condition serializers for the new reference field added to the models. | https://github.com/Flagsmith/flagsmith-common/pull/11 |
| **Open** | Adds the new reference field to the models, and implements an algorithm to match a segment revision's rules and conditions to the live segment in order to populate the reference field. | https://github.com/Flagsmith/flagsmith/pull/4988 |
| **Open** | Closes the loop by calling the above algorithm from the Change Request context. | https://github.com/Flagsmith/flagsmith-workflows/pull/63 |

The following diagram illustrates the functionality flow in use cases “Create a Segment Change Request” and “Commit a Segment Change Request”. It presents the architecture and some of the relevant components from the work listed above:

[...]

PR #5775, opened 2025-07-16
chore: Move SegmentSerializer and MetadataSerializer back to core API

emyller:
Contributes to #5769.

## Changes

- [x] Move `SegmentSerializer`, `RuleSerializer`, and `ConditionSerializer` back to the core API code. They were moved [in the past](https://github.com/Flagsmith/flagsmith-common/pull/4), and have since evolved.
- [x] Delete custom logic to handle nested objects in favor of `WritableNestedModelSerializer`.

Replaced multiple organs, patient walked out smiling.

- [x] Double check test coverage.


api/segments/serializers.py
emyller, line 1: > [!NOTE]
>
> This code was imported from the [original source](https://github.com/Flagsmith/flagsmith-common/blob/49da185f7f3ea5d341b7eaeeee1abd2f8f7c035e/src/common/segments/serializers.py), and refactored to meet the goals set by #5768.
> 
> The biggest changes here are:
> 
> 1. The removal of custom behavior to manage nested rules and conditions. This is now handled by `WritableNestedModelSerializer`.
> 2. Refactor of pretty much everything.

api/tests/unit/segments/test_unit_segments_views.py
emyller, line 1: Adding new tests is OK, changing existing ones is _not ideal_.

However, we've decided to accept the proposed changes because they fix validation error responses in rare cases. cc @matthewelwell

api/segments/serializers.py
```diff
@@ -1,6 +1,321 @@
...
+    def get_project(
+        self,
+        validated_data: dict[str, Any] | None = None,
+    ) -> Project:
+        project: Project
+        if validated_data and "project" in validated_data:
+            project = validated_data["project"]
+            return project
+        project = Project.objects.get(id=self.context["view"].kwargs["project_pk"])
+        return project
```
khvn26: We're still leaving `SerializerWithMetadata` in flagsmith-common, right? Can we bring it over as well?

khvn26 (dismissed):
The "new" code, meaning the test code, looks good to me.

I have a plethora of questions to the "old" code, but, taking @emyller's note into account, I'm leaving them out of the scope of this PR. Mainly, I feel that neither Django models nor DRF serializers are good vessels for business logic, but serializers are worse in that regard because their lifecycle is harder to comprehend/memorise.

Approving in hopes to see this cleaned up in the future. As usual, my advice is to establish a proper service layer.

emyller:
> I have a plethora of questions to the "old" code, (...)

@khvn26 Please go ahead and ask away. I think I was too light in my words "imported and refactored" because all the serializers reintroduced to core were also **rewritten** — perhaps except for one or two validation methods I thought we could refactor later.

As of this patch, the serializer is only meant to house _validation_ rather than handling database objects. There is code I'd like to delete but for now it needs to exist for backwards compatibility — I left TODOs for later deprecation. I'm all ears to any advice in the structure, as I'd love for [at least] most of this to look like the patterns we're intending to establish.

api/segments/serializers.py
```diff
@@ -1,6 +1,305 @@
...
+        try:
+            self.update_metadata(instance, metadata_data)
+            instance = super().update(instance, validated_data)  # type: ignore[no-untyped-call]
+        except Exception:
+            # Roll back ¯\_(ツ)_/¯
+            instance.refresh_from_db()
+            instance.version = cloned_segment.version
+            instance.save()
+            cloned_segment.hard_delete()
+            raise
```
emyller: This is original code. It is going to be rewritten as part of #5771.

api/segments/serializers.py
```diff
@@ -1,6 +1,305 @@
...
+            "value",
+            "description",
+            "delete",
+            "version_of",
+        ]
+
+    def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
+        attrs = super().validate(attrs)
+        if attrs.get("operator") != PERCENTAGE_SPLIT and not attrs.get("property"):
+            raise ValidationError({"property": ["This field may not be blank."]})
```
khvn26: Arguably, this requirement is not valid anymore. FE will default to `"$.identity.key"` for this operator instead. This could be a chance to simplify validation.

CC @Zaimwa9

api/segments/serializers.py
```diff
@@ -1,6 +1,305 @@
...
+                if rule_id := rule_data.get("id"):
+                    rules_ids.add(rule_id)
+                for condition_data in rule_data.get("conditions", []):
+                    if condition_id := condition_data.get("id"):
+                        rule_condition_ids.add((rule_id, condition_id))
+                collect_ids(rule_data.get("rules", []))
+
+        rules_ids: set[int] = set()
+        rule_condition_ids: set[tuple[int | None, int]] = set()
+        collect_ids(rules_data)
```
khvn26: I hate that this works:

```python
def scoping_shenanigans():
    unset.append(1)

unset = []
scoping_shenanigans()
print(unset)  # [1]
```

Why not have a `_collect_ids(self, rules_data) -> dict[int | None, set[int]]` method instead?

api/segments/serializers.py
```diff
@@ -1,6 +1,305 @@
...
+        """
+        for r in range(len(rules_data)):
+            if (rule := rules_data[r]).get("delete") is True:
+                del rules_data[r]
+                continue
+            for c in range(len(conditions := rule.get("conditions", []))):
+                if conditions[c].get("delete") is True:
+                    del conditions[c]
+            if subrules := rule.get("rules", []):
+                self._remove_segment_rules_conditions_marked_for_deletion(subrules)
```
khvn26: I appreciate this is soon to be deprecated code, and maybe it's just my eyes adjusted to reading list comprehensions, but it would improve in readability so much if it didn't mutate the original data but returned a filtered copy:

```python
def _get_rules_without_deleted(
    self,
    rules_data: DictList,
) -> DictList:
    return [
        {
            **rule_data,
            "conditions": [
                condition_data
                for condition_data in rule.get("conditions") or []
                if not condition_data.get("delete")
            ],
            "rules": self._get_rules_without_deleted(
                rule_data.get("rules") or [],
            )
        }
        for rule_data in rules_data
    ]
```

api/metadata/serializers.py
```diff
@@ -83,3 +87,129 @@ class ContentTypeSerializer(serializers.ModelSerializer):  # type: ignore[type-a
...
+            metadata_model_field = metadata_item["model_field"]
+            if metadata_model_field is not None and metadata_model_field.id is not None:
+                Metadata.objects.update_or_create(
+                    model_field=metadata_model_field,
+                    content_type=content_type,
+                    object_id=instance.pk,
+                    defaults={
+                        "field_value": metadata_item["field_value"],
+                    },
+                )
```
khvn26: Can we use `bulk_create` with `update_conflicts` here?

api/segments/serializers.py
```diff
@@ -1,6 +1,305 @@
...
+        rules_data: DictList,
+    ) -> int:
+        count = 0
+        for rule_data in rules_data:
+            child_rules: DictList = rule_data.get("rules", [])
+            if child_rules:
+                count += self._calculate_condition_count(child_rules)
+            conditions: DictList = rule_data.get("conditions", [])
+            for condition in conditions:
+                if condition.get("delete", False) is True:
```
khvn26: This is supposedly called called after `_remove_segment_rules_conditions_marked_for_deletion`, so I'm surprised this branch is hit by coverage.

api/segments/serializers.py
```diff
@@ -1,6 +1,305 @@
...
+        self._validate_relations_of_nested_rules_and_conditions(attrs["rules"])
+        return attrs
+
+    def create(self, validated_data: dict[str, Any]) -> Segment:
+        project = validated_data["project"]
+        self._validate_project_segment_limit(project)
+
+        metadata_data = validated_data.pop("metadata", [])
+
+        self.context["is_creating_segment"] = True
```
khvn26, line 118: This line makes me think, why have separate `SegmentRuleSerializer` and `ConditionSerializer` at all? Are they ever used outside of `SegmentSerializer`?

But I guess the real question would be, why store rules and conditions separately at all?

api/segments/serializers.py
```diff
@@ -1,6 +1,305 @@
...
+        return attrs
+
+    def to_internal_value(self, data: dict[str, Any]) -> Any:
+        # Conversion to correct value type is handled elsewhere
+        data["value"] = str(data["value"]) if "value" in data else None
+        return super().to_internal_value(data)
+
+    def create(self, validated_data: dict[str, Any]) -> Condition:
+        if self.context.get("is_creating_segment"):
+            validated_data["created_with_segment"] = True
```
khvn26, line 42: Side note: `created_with_segment` is just such a quirky little field. As far as I can tell, it's only used to hide initial rules from the audit log. I couldn't find other references, and can't think of any other potential use for it. If I had to write this, I'd stop and try to reconsider the whole data model for sure.

api/segments/serializers.py
khvn26: BTW, we should get rid of `models.Segment.id_exists_in_rules_data`, right?

api/segments/serializers.py
```diff
@@ -1,6 +1,305 @@
...
+
+        existing_rule_conditions_ids = set(
+            Condition.objects.filter(
+                pk__in=condition_ids,
+            ).values_list("rule_id", "id")
+        )
+
+        # Ensure existing conditions continue belonging to their rules
+        if not rule_condition_ids <= existing_rule_conditions_ids:
+            raise ValidationError({"segment": "Cannot move conditions between rules"})
```
khvn26: The original problem we were trying to solve was that someone had shared a rule between multiple segments. This caused confusion when the rule was modified and made it difficult to trace.

The code here looks OK. It does the job, it's well tested, and I trust that it will prevent most consistency problems in future. However, I can't help but wonder why we opted for this design instead of simply stripping the data of IDs and copying the incoming rule/condition data over to the recipient segment. I imagine that would provide a much more predictable API behaviour from the user standpoint and be inherently safe.

One might argue that we use the IDs to distinguish between rule creation and updates for auditing purposes (and possibly to generate nicer diffs for versioning, but I'm not sure), but I doubt this is valuable to the user. If we want to keep this functionality, relying on the list index would arguably be just as useful without the silly effect of rule data being rejected because one provided an ID belonging to another segment.

khvn26 (changes requested):

api/segments/serializers.py
emyller: @khvn26 Should we account for API requests made outside the frontend that might not include `property`?

khvn26: The answer might be tricky. The current contract is as follows:

image

Strictly speaking, we shouldn't break it, but we're kind of already are for the majority of operators, so perhaps sneaking in a breaking change wouldn't be too bad in this case.

I'm leaning towards "No, we shouldn't", with an additional "let's make the field officially required".

api/metadata/serializers.py
emyller: I forgot to note that this is mostly original code. That said I think it's worth checking out.

api/segments/serializers.py
emyller: I agree. For now this was the least painful way to persist this flag to memory, for [later use](https://github.com/Flagsmith/flagsmith/blob/cf1983114a2f48217a8e94ad48e2756e441b7149/api/segments/serializers.py#L48) in the Condition serializer. Perhaps we'll soon get to see this going away with [JSON rules](https://github.com/Flagsmith/flagsmith/issues/5846)? 👀

api/segments/serializers.py
emyller: > This line makes me think, why have separate `SegmentRuleSerializer` and `ConditionSerializer` at all? Are they ever used outside of `SegmentSerializer`?

They need to be separate because of the recursive/nested architecture of rules and conditions. At least for the current data model.

> But I guess the real question would be, why store rules and conditions separately at all?

Yup. They could/[will](https://github.com/Flagsmith/flagsmith/issues/5846) be a JSON blob.

api/segments/serializers.py
emyller: Thanks for the snippet! Improved in 8b04d24b064dbe5c311cfddca4b74b854ef0dbe1.

api/segments/serializers.py
emyller: I agree it's best simplified. Improved in bbcec4b44799c05b33bb3d5490f2733cf7a5b7a4.

api/segments/serializers.py
emyller: I'm onboard. @Flagsmith/flagsmith-back-end what do you think?

api/tests/unit/segments/test_unit_segments_views.py
```diff
@@ -712,148 +711,6 @@ def test_update_segment_add_new_condition(
...
-        url, data=json.dumps(data), content_type="application/json"
-    )
-
-    # Then
-    assert response.status_code == status.HTTP_400_BAD_REQUEST
-    assert response.json() == {"segment": "Mismatched segment is not allowed"}
-    existing_condition.refresh_from_db()
-    assert existing_condition._get_segment() != segment
-
-
```
emyller, line 856: These tests are deleted because we no longer care about mismatching IDs of rules or segments.

api/segments/serializers.py
emyller: > Why not have a `_collect_ids(self, rules_data) -> dict[int | None, set[int]]` method instead?

The code was deleted. 🎉 


> BTW, we should get rid of `models.Segment.id_exists_in_rules_data`, right?

Great catch, fixed in 81f7cdb9ec9e9b86f4573e5b44b5eb789397ef6f.

api/metadata/serializers.py
emyller: I actually refactored the wholeness of metadata serializers, see 88dd8a78735d975289fbca535e3b7f142961d8ab.

api/segments/serializers.py
emyller: I rewrote the code as of 88dd8a78735d975289fbca535e3b7f142961d8ab. The cloning shenanigans are still going to be dealt with in #5771.

api/metadata/serializers.py
emyller, line 1: > [!NOTE]
>
> This code was imported from the [original source](https://github.com/Flagsmith/flagsmith-common/blob/d22059fda2468df4e0f71a64513dcdd066f39b04/src/common/metadata/serializers.py), and **rewritten** to improve its architecture, and to meet the goals set by #5768.

api/segments/serializers.py
emyller: Done in 488feb1c0de5cf0c89168995ca3cfe22f988ab05.

api/segments/serializers.py
emyller: TL;DR: the code above, along with other complexity (see updated patch), is no more. 🎉 

---

> I can't help but wonder why we opted for this design instead of simply stripping the data of IDs and copying the incoming rule/condition data over to the recipient segment. I imagine that would provide a much more predictable API behaviour from the user standpoint and be inherently safe.

We actually went with that, thanks for raising the discussion @khvn26. IDs should now be ignored by the interface (both `id` and `version_id` fields of rules and conditions), while keeping compatibility with the current API.

> One might argue that we use the IDs to distinguish between rule creation and updates for auditing purposes (and possibly to generate nicer diffs for versioning, but I'm not sure), but I doubt this is valuable to the user. If we want to keep this functionality, relying on the list index would arguably be just as useful without the silly effect of rule data being rejected because one provided an ID belonging to another segment.

Yeah diffing in the UI is valuable to the user, and this along with #5770 (now closed) had the intention to make that safely possible given the data model and the `version_of` requirement — which was thankfully dropped in favor of [text-based diffing](https://github.com/Flagsmith/flagsmith/issues/5770#issuecomment-3145221419). Again, thanks @khvn26!

api/features/serializers.py
```diff
@@ -291,25 +287,25 @@ def validate(self, attrs):  # type: ignore[no-untyped-def]
...
 
-    def get_num_identity_overrides(self, instance) -> typing.Optional[int]:  # type: ignore[no-untyped-def]
+    def get_num_identity_overrides(self, instance: Feature) -> int | None:
         try:
             return self.context["overrides_data"][instance.id].num_identity_overrides  # type: ignore[no-any-return]
         except (KeyError, AttributeError):
             return None
 
-    def get_is_num_identity_overrides_complete(self, instance) -> typing.Optional[int]:  # type: ignore[no-untyped-def]  # noqa: E501
+    def get_is_num_identity_overrides_complete(self, instance: Feature) -> int | None:
```
khvn26:
```suggestion
    def get_is_num_identity_overrides_complete(self, instance: Feature) -> bool | None:
```

api/features/serializers.py
```diff
@@ -328,31 +324,23 @@ def get_last_modified_in_current_environment(
...
-        project = self.context.get("project")
-        if project:
-            return project  # type: ignore[no-any-return]
-        else:
-            raise serializers.ValidationError(
-                "Unable to retrieve project for metadata validation."
-            )
+    def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
+        attrs = super().validate(attrs)
+        project = self.instance.project if self.instance else self.context["project"]  # type: ignore[union-attr]
```
khvn26, line 335: Note: we're going to get `AttributeError`s if this serializer is incorrectly used in the future, which is, IMO, better than previous behaviour of raising a `ValidationError` 👍

api/features/serializers.py
emyller: Painful not to refactor all this right now. 🤦 5c1c308d82fbb69e79f5e17ae3527f5f3e3095ce

api/segments/serializers.py
Zaimwa9: I would lean toward having the breaking change. If not done already, our frontend would be quickly compatible.
Do we have a way to estimate the amount of rules created from API requests with an empty property on `PERCENTAGE_SPLIT` ?
It's kind of specific so except if there are some recent ones I would agree in simplifying and making it mandatory but making it obvious in the doc that `$.identify.key` is the default
cc @khvn26

khvn26: > Do we have a way to estimate the amount of rules created from API requests with an empty property on `PERCENTAGE_SPLIT` ?

3475 rows in prod SaaS. I think a data migration should be an easy one.

emyller: Done in 635dd77bce207dc6a8486f776f5327d732ccba91

matthewelwell: Making a column non-nullable in postgres requires a lock on the table. Given how often the table is accessed for critical workflows, I'm not sure I agree with the approach taken here as it will likely result in a short blip in availability of the core API. 

What's the benefit of making the field non-nullable?

khvn26: I was not proposing any changes to the db schema, though. We could just make the serializer field required.

matthewelwell: Yep, I'm happy to go along with the breaking API change, but we should revert the database level constraint added [here](https://github.com/Flagsmith/flagsmith/commit/635dd77bce207dc6a8486f776f5327d732ccba91).

emyller: I decided to drop the `NOT NULL` constraint because, in fact, there's no condition (in SaaS) with a NULL property. Instead, they're empty strings — to be filled with `"$.identity.key"`. I thought it would improve consistency but we can [avoid schema changes](https://github.com/Flagsmith/flagsmith/pull/5775/files#r2254572034) with 3bbe78417dc0c6d9700584e4b25129c2c81012cf.

api/segments/migrations/0028_condition_property_required.py
emyller, line 1: SQL produced:

```sql
BEGIN;
--
-- Alter field property on condition
--
-- (no-op)
--
-- Alter field property on historicalcondition
--
-- (no-op)
--
-- Raw SQL operation
--
UPDATE segments_condition SET property = '$.identity.key' WHERE operator = 'PERCENTAGE_SPLIT' AND (property IS NULL OR property ~ '^\s*$');
--
-- Raw SQL operation
--
UPDATE segments_historicalcondition SET property = '$.identity.key' WHERE operator = 'PERCENTAGE_SPLIT' AND (property IS NULL OR property ~ '^\s*$');
COMMIT;
```

khvn26: LGTM 👌

khvn26 (dismissed):

api/metadata/serializers.py
```diff
@@ -83,3 +86,67 @@ class ContentTypeSerializer(serializers.ModelSerializer):  # type: ignore[type-a
...
+                object_id=instance.pk,
+                field_value=metadata_item["field_value"],
+            )
+            for metadata_item in metadata_data
+            if (model_field := metadata_item["model_field"]) and model_field.pk
+        ]
+
+        with transaction.atomic():
+            existing_metadata.delete()
+            Metadata.objects.bulk_create(new_metadata)
```
matthewelwell, line 152: Sorry if this has been covered already, but I'm curious why we're going with the delete and recreate approach vs the previous update_or_create approach?

api/segments/migrations/0028_condition_property_required.py
```diff
@@ -0,0 +1,45 @@
...
+            field=models.CharField(max_length=1000, null=True),
+        ),
+        migrations.RunSQL(
+            sql=r"UPDATE segments_condition SET property = '$.identity.key' WHERE operator = 'PERCENTAGE_SPLIT' AND (property IS NULL OR property !~ '\S')",
+            reverse_sql="UPDATE segments_condition SET property = '' WHERE operator = 'PERCENTAGE_SPLIT' AND property = '$.identity.key'",
+        ),
+        migrations.RunSQL(
+            sql=r"UPDATE segments_historicalcondition SET property = '$.identity.key' WHERE operator = 'PERCENTAGE_SPLIT' AND (property IS NULL OR property !~ '\S')",
+            reverse_sql="UPDATE segments_historicalcondition SET property = '' WHERE operator = 'PERCENTAGE_SPLIT' AND property = '$.identity.key'",
+        ),
```
matthewelwell: Why do we need to revert to raw SQL here? The reason that I ask is that there's a tax to doing so since this migration will also need to run on oracle and mysql databases for our enterprise edition.

matthewelwell (commented):
Looks pretty good on the whole - definitely cleaner than it was, so nice work! I've added a couple of questions, but nothing that I'd be concerned will present any blockers.

api/segments/migrations/0028_condition_property_required.py
emyller: I think no reason other than myself being lazy — SQL is just quicker to write. Improved with ec8650a8d96441244e06df136cbcf4641a62d9b8.

Also: do we actually need to update historical objects? 🤔

api/metadata/serializers.py
emyller: I think only because the delete-recreate approach rendered to less logic to maintain, then the choice prevailed because metadata is not too many objects.

I can also go with the update-or-create (and delete-obsolete) approach if you see value in it. 👀

api/segments/migrations/0028_condition_property_required.py
```diff
@@ -0,0 +1,69 @@
...
+from django.apps.registry import Apps
+from django.db import migrations, models
+
+
+def fill_percentage_split_property_for(model_name: str) -> Callable[[Apps, Any], None]:
+    def fill_property(apps: Apps, _: Any) -> None:
+        model = apps.get_model("segments", model_name)
+        model.objects.filter(
+            models.Q(operator="PERCENTAGE_SPLIT") & (
+                models.Q(property__isnull=True) | models.Q(property__regex=r"^\s*$")
```
khvn26:
```suggestion
                models.Q(property__isnull=True) | models.Q(property="")
```

Is there something I'm missing?

emyller: e.g. `" "`, or any amount of whitespaces that could've slipped from the frontend.

khvn26: Any evidence of such values (e.g. from Metabase)?

emyller: No evidence from the SaaS standpoint, though I preferred to stay on the safer side and also cover `null`s and other empty-ish that could occur in private installs for whatever reason. TBH not sure this is worth the database CPU effort in the migration but definitely safer.

khvn26: > No evidence from the SaaS standpoint

Good to know, thanks! I think it's a good argument for applying the suggestion as well.

> not sure this is worth the database CPU effort in the migration but definitely safer.

I'm strictly on the side of performant rather than redundant data migrations.

emyller: Improved with 6083657264b3724cefd88831f7e3aac7126c2cc3

khvn26 (dismissed):

api/segments/migrations/0028_condition_property_required.py
matthewelwell: No, historical objects are fine.

api/metadata/serializers.py
matthewelwell: I guess since we're using a transaction, and not tracking any version history, it doesn't really matter so if you're happy that the code is less complex this way, then I'm happy.

matthewelwell (dismissed):

api/segments/migrations/0028_condition_property_required.py
emyller: Fixed with ad096a0d70683bc1827f993664fd687dbff5581d

matthewelwell (approved):

PR #5798, opened 2025-07-21
feat: Backend for Context values

### LINKED ISSUE #5674: Backend: Implement Context values for Segments
Blocked on https://github.com/Flagsmith/flagsmith-engine/issues/219.

## DoD

As an API user,
When performing
 - `POST /api/v1/projects/{project_pk}/segments/`
 - `PUT /api/v1/projects/{project_pk}/segments/{id}/`
 - `PATCH /api/v1/projects/{project_pk}/segments/{id}/`

When providing a value under `$.rules[].conditions[].property` which is an invalid JSONPath within the evaluation context object, it is assumed to be a trait. This can be simplified to any property that doesn't start with `$.`. 

As an API user,
When performing
 - `POST /api/v1/projects/{project_pk}/segments/`
 - `PUT /api/v1/projects/{project_pk}/segments/{id}/`
 - `PATCH /api/v1/projects/{project_pk}/segments/{id}/`

I'm able to provide the following values under `$.rules[].conditions[].property`:

- `$.identity.identifier`
- `$.identity.traits.{any trait name}`
- `$.environment.name`.

Then,

As an API user,
When performing
- `GET /api/v1/flags/`
- `GET /api/v1/identities/`

I can observe the Core API correctly fill the condition values for these properties when presenting the context for the engine, and evaluate flags using the expected segments.

khvn26:
Closes #5674.

As part of this, I realised that using flagsmith-flag-engine ^6.0.0 together with the Flagsmith SDK is impossible, so had to publish flagsmith-flag-engine 5.4.3 with a backport of `get_identity_segments`.

## Changes

This bumps Flagsmith Engine to a version supporting Context values, enabling Context values evaluation in the Core API. Some code paths were updated to support the context object now expected by some of the Engine's internal-ish APIs. 

Updated relevant tests. The evaluation itself was tested in the engine. This will be tested on staging as part of the integration testing, together with UI.


Zaimwa9 (dismissed):
LGTM code-wise

api/util/mappers/engine.py
```diff
@@ -418,6 +419,27 @@ def map_identity_to_engine(
...
 
 
+def map_engine_identity_to_context(
+    identity: IdentityModel,
+) -> "EvaluationContext":
+    """
+    A special mapper to produce a minimal EvaluationContext
+    in an environment-less form.
+    Used when an environment object is not available,
+    like when evaluating segments in UI, which happens at the project level.
```
matthewelwell: I'm slightly confused by this - the only place I can see this mapper function being used relates to webhooks. I'm not sure what 'evaluating segments in UI' means?

matthewelwell (commented):
Nice to see that no tests needed to change for this!

api/util/mappers/engine.py
khvn26: Fixed in e692a89c5227a54370127f25f468e4af4c07d60a.

matthewelwell (approved):

PR #5823, opened 2025-07-23
feat: Process `User-Agent` strings

### LINKED ISSUE #5735: Process `User-Agent` strings
We are tracking `User-Agent` strings. However, in order for the data to be actually useful, and [because browsers can't reliably override UAs for fetches](https://issues.chromium.org/issues/40450316), we need to distinguish between SDK UAs and end user UAs.

We should expect a new header that will identify JS SDKs, and cater for existing JS SDKs that do not send it to avoid storing end user browser UA strings as label values.

## DoD

- An additional `Flagsmith-SDK-User-Agent` string is expected, expected to be sent by the JS SDK, and possibly by some other client-side SDKs
- Incoming `User-Agent` are inspected for end user browser UAs. These are dropped/normalised.

khvn26:

## Changes

Closes #5735.

This adds `User-Agent` processing ahead of labelling the usage and feature evaluation data. Browser UA strings are stripped, the rest are considered server-side SDKs, and a new `Flagsmith-SDK-User-Agent` header is expected to denote a browser SDK version.

# How did you test this code?

Updated existing view unit test to reflect the new behaviour.

### DIFF
--- a/api/app_analytics/mappers.py
+++ b/api/app_analytics/mappers.py
@@ -125,7 +125,7 @@ def map_input_labels_to_labels(input_labels: InputLabels) -> Labels:
             # represent server-side SDKs.
             parsed_ua_string: str = parse_ua(value)
             is_server_side_sdk = parsed_ua_string.startswith("Other - ")
-            
+
             # Skip browser SDKs that don't send the special header
             if not is_server_side_sdk:
                 continue
--- a/api/app_analytics/mappers.py
+++ b/api/app_analytics/mappers.py
@@ -121,10 +121,13 @@ def map_input_labels_to_labels(input_labels: InputLabels) -> Labels:
             labels["user_agent"] = value
             continue
         elif label == "user_agent":
+            # fastuaparser classifies unrecognized UAs as "Other" — assume these to
+            # represent server-side SDKs.
             parsed_ua_string: str = parse_ua(value)
-            # Assume UA strings categorised as "Other" to represent server-side SDKs.
-            # Skip browser SDKs that don't send the special header.
-            if parsed_ua_string.split(" - ")[0] != "Other":
+            is_server_side_sdk = parsed_ua_string.startswith("Other - ")
+            
+            # Skip browser SDKs that don't send the special header
+            if not is_server_side_sdk:
                 continue
         labels[label] = value
     return labels
--- a/api/app_analytics/mappers.py
+++ b/api/app_analytics/mappers.py
@@ -124,9 +124,8 @@ def map_input_labels_to_labels(input_labels: InputLabels) -> Labels:
             # fastuaparser classifies unrecognized UAs as "Other" — assume these to
             # represent server-side SDKs.
             parsed_ua_string: str = parse_ua(value)
-            is_server_side_sdk = parsed_ua_string.startswith("Other - ")
-
-            # Skip browser SDKs that don't send the special header
+            is_server_side_sdk = parsed_ua_string.startswith("Other")
+            # Skip browser SDKs that don't send the special header.
             if not is_server_side_sdk:
                 continue
         labels[label] = value
--- a/api/app_analytics/mappers.py
+++ b/api/app_analytics/mappers.py
@@ -3,6 +3,7 @@ from functools import partial
 from typing import Any, Iterable
 
 from django.http import HttpRequest
+from fastuaparser import parse_ua  # type: ignore[import-untyped]
 from influxdb_client.client.flux_table import FluxTable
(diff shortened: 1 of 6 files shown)


api/pyproject.toml
```diff
@@ -167,6 +167,7 @@ djangorestframework-simplejwt = "^5.3.1"
 structlog = "^24.4.0"
 prometheus-client = "^0.21.1"
 django_cockroachdb = "~4.2"
+fastuaparser = "^0.1.4"
```
emyller, line 170: [fastuaparser](https://github.com/almarklein/fastuaparser) looks like a very interesting initiative, though it also seems like [uap-python](https://github.com/ua-parser/uap-python), _as defacto_, for now offers more community support and updates. Do you think it's a good idea we switch to it instead?

api/app_analytics/mappers.py
```diff
@@ -112,19 +114,35 @@ def map_flux_tables_to_feature_evaluation_data(
...
+    for label, value in input_labels.items():
+        if label == "sdk_user_agent":
+            labels["user_agent"] = value
+            continue
+        elif label == "user_agent":
+            parsed_ua_string: str = parse_ua(value)
+            # Assume UA strings categorised as "Other" to represent server-side SDKs.
+            # Skip browser SDKs that don't send the special header.
+            if parsed_ua_string.split(" - ")[0] != "Other":
+                continue
```
emyller: I'm not sure what's happening here, despite the comment, sorry. I assume it's domain-specific to fastuaparser?

The alternative uap-parser (mentioned in the other thread) might offer a cleaner API, e.g.

```suggestion
        # Assume unrecognized User Agents to represent server-side SDKs
        elif label == "user_agent" and not parse_user_agent(value):
            continue
```

emyller (changes requested):
Looks good in general! Let's choose a safer option to parse UAs if you agree.

api/pyproject.toml
khvn26: The reason I chose fastuaparser was performance. We don't need precision as much as we need performance here, as this code is called from a middleware.

emyller: Sounds fair. Do you believe the performance difference **here** is worth the risk of relying on unmaintained software? [I hope the project gets traction though!]

api/app_analytics/mappers.py
khvn26: > I'm not sure what's happening here, despite the comment, sorry.

Let me know how can we improve the comment.

> The alternative uap-parser (mentioned in the other thread) might offer a cleaner API, e.g.

Sure, but see considerations above.

api/pyproject.toml
khvn26: If you look at it, it's just a single Python module with a single function. Given our use case, where we simply skip most of the UAs, I think the risk is justified. Honestly, if we ever need to move away from this dependency, I'd just steal parts of its logic instead of swapping for something more feature-complete.

api/app_analytics/mappers.py
emyller: > Let me know how can we improve the comment.

The suggested snippet includes a suggestion of comment, though its readability largely depends on not having to explain how fastuaparser works. If you're adamant in keeping fastuaparser considering my [last question](https://github.com/Flagsmith/flagsmith/pull/5823#discussion_r2226626243), I think a slightly longer comment here could help. Also, sorry, my suggestion was incomplete in respect to the "skip browser SDKs" bit.

```suggestion
        elif label == "user_agent":
            # fastuaparser classifies unrecognized UAs as "Other" — assume these to
            # represent server-side SDKs.
            parsed_ua_string: str = parse_ua(value)
            is_server_side_sdk = parsed_ua_string.startswith("Other - ")
            
            # Skip browser SDKs that don't send the special header
            if not is_server_side_sdk:
                continue
```

api/pyproject.toml
emyller: Yup, I saw the code, it really is just a very copy-pastable snippet. I agree with keeping it until we can't, and will hope we'll can!

khvn26: It does pass the ua-parser's test suite though, so we can more or less rely on the results.

emyller (dismissed):
Well let's :shipit:

api/app_analytics/mappers.py
khvn26: Sorry, applying the suggestion broke the logic. Fixed in 0b7e0d88814be2c90eae85704f18c812e1f4a974, but need a re-approval now.

emyller (approved):
image

Issue #5804, opened 2025-07-22
Context Values: populate Environment names


khvn26:
## DoD

As a dashboard user,
When I create an edit a Segment,
When I select "Environment Name" in the Context value searchable dropdown,
I expect the "Value" field to turn into a searchable dropdown containing all of the project Environment names.

Zaimwa9:
To only use with Exactly matches and does not match

Issue #5805, opened 2025-07-22
Context Values: validate rules that always evaluate to one of `false` or `true`


khvn26:
As we allow more flexibility in segmenting with Context Values, it's important to notify the users of some rules that will always evaluate to `false`:

| Context Value | Operator | Rule Value |
| ------------- | -------- | ---------- |
| Any, except Trait | Is Not Set | N/A |
| Any | % Split | 0 |
| Environment Name | Any | Not one of current Project's Environment Names |

or `true`:

| Context Value | Operator | Rule Value |
| ------------- | -------- | ---------- |
| Any, except Trait | Is Set | N/A |
| Any | % Split | 100 |
| Environment Name | In | All of the Project's Environment Names |

## DoD

As a dashboard user,
When creating or updating Segments,
When selecting one of the rules above,
I expect to see a label warning me that this rule will always evaluate to `false`.

matthewelwell:
Discussed with the team that we should wait for customer feedback on context values before progressing with this item. We see that there are a few paths we could take in order to implement a solution to the problem this is trying to solve: 

 1. a 'simple' UI warning based on the defined set of rules as per the issue description
 2. validation on create / update in the API based on the same set of rules
 3. add a new 'test your segment' function which allows the user to test the rules of their segment, perhaps against a given identity, or set of identities
 4. periodically, or dynamically, tag segments with rules that match these rules

As such, we are parking this issue for now until we get more feedback.

matthewelwell:
Closing until we receive more feedback and this becomes necessary.

PR #5832, opened 2025-07-25
feat: use dropdown for values with context environment

Zaimwa9:

## Changes
When using the `Environment name` context value (and not `PERCENTAGE_SPLIT`), the input value becomes a dropdown with the project's environment values

N/A
image

image

https://www.loom.com/share/9c00c9440fed4a499ea16192a42cb932?focus_title=1&muted=1&from_recorder=1


khvn26 (changes requested):
I see the rationale behind using a dropdown. Consider, however, a case when an environment used in the segment condition was deleted. I believe we should use a searchable dropdown allowing arbitrary values — should work nicely with a warning added by #5805.

khvn26:
Closes #5804.

frontend/web/components/base/SearchableDropdown.tsx
```diff
@@ -0,0 +1,89 @@
+import Icon from 'components/Icon'
```
kyle-ssg, line 1: It feels like we need a way for this and EnvironmentSelect to be the same component, they are too similar ?

Zaimwa9: What I had in mind is that this one is just the UI that works with the options and categories directly.
While we have 2 containers using them `RuleConditionPropertySelect` and `EnvironmentSelectDropdown` just in charge of building the options and dealing with the localState.

But actually the most reusable part here is the `GroupLabel` that I could keep here and then just use the `Select` directly in the end wdyt?

Also I noticed that the screenshots are not up to date as we added a custom selection:

image

frontend/web/components/base/SearchableDropdown.tsx
```diff
@@ -0,0 +1,94 @@
...
+  maxMenuHeight,
+  noOptionsMessage,
+  onBlur,
+  onChange,
+  onInputChange,
+  options,
+  placeholder,
+  value,
+}) => {
+  console.log('components', components)
```
tiagoapolo:
```suggestion
```

tiagoapolo (dismissed):
Besides the console log, looks good on my end

tiagoapolo (dismissed):

tiagoapolo (dismissed):

khvn26 (approved):

PR #5833, opened 2025-07-25
feat(CockroachDB support): Switch to `JSONField` for SDK metrics labels

khvn26:

## Changes

This migrates all the `labels` fields added in #5623 to `jsonb` as the `hstore` support turned out to be problematic:
1. The `hstore` extension requires enabling additional parameter sets in some cloud environments (e.g. Azure).
2. Psycopg2 type adapters don't play well with CockroachDB, which should be compatible otherwise.

Note that due to problem 1, I've removed the `CREATE EXTENSION` operation from the original migration.

Migrations and tests run ok.
Will be smoke-tested against CockroachDB as well.


scarlson:
Confirming this fix allows CockroachDB to work without needing to update Psycopg to 3.1+.

api/app_analytics/migrations/0007_labels_jsonb.py
```diff
@@ -0,0 +1,33 @@
+# Generated by Django 4.2.22 on 2025-07-25 14:09
```
Zaimwa9: This looks redundant with 0006, I guess from local dev, can't we squash them together ?

Zaimwa9 (changes requested):
The migrations look redundant except if I am missing something? Can we see to merge them together maybe ?

khvn26:
@Zaimwa9 Indeed these migrations look very similar, but their purpose is actually different:
- 0006 adds correct jsonb columns to the tables. The original migration is modified to avoid customers having to add hstore capability to their databases since it's no longer a requirement. This exists primarily for new installations/fresh dbs.
- 0007 alters existing hstore columns to jsonb. This has to happen for installations where previous 0006 migration was already applied.

So, to answer your question, no, we can't merge these migrations as they cater for different scenarios, and I argue they are not in fact redundant.

Zaimwa9 (dismissed):
Alright thanks for the details, I felt that under the hood there was a migration purpose

api/app_analytics/migrations/0007_labels_jsonb.py
```diff
@@ -0,0 +1,33 @@
...
+        migrations.AlterField(
+            model_name="featureevaluationbucket",
+            name="labels",
+            field=models.JSONField(default=dict),
+        ),
+        migrations.AlterField(
+            model_name="featureevaluationraw",
+            name="labels",
+            field=models.JSONField(default=dict),
+        ),
```
emyller: To my understanding, this will lead to multiple `alter column ... type` statements. Is casting hstore to json as simple as that in, at least, Postgres?

I'm also curious as to the impact of this to existing installs.

emyller (changes requested):

emyller:
I forgot to add as notes to my last review: not actually requesting any changes unless proven necessary by [the comment](https://github.com/Flagsmith/flagsmith/pull/5833#discussion_r2236933750) I added.

api/app_analytics/migrations/0007_labels_jsonb.py
khvn26: Good catch, we actually need a `USING` clause. Added in 091006e7f930ca7a4b1360994113a7d9d43d93c2.

api/app_analytics/migrations/0007_labels_jsonb.py
```diff
@@ -0,0 +1,73 @@
...
+                                relname
+                            );
+                        END LOOP;
+                    END
+                    $$;
+                    """,
+                    # We don't want hstore in the database at all, 
+                    # so don't do anything for reverse SQL.
+                    reverse_sql="",
+                ),
```
emyller, line 72: The SQL code looks fine, though I think we should benefit from the ORM abstraction until we need to fall back to raw SQL.

Here's an untested brain dump.

Given:

```python
def upgrade_field(model_name: str, field_name: str) -> Callable[[Apps, SchemaEditor], None]:
    def upgrade(apps: Apps, schema_editor: SchemaEditor) -> None:
        model = apps.get_model("app_analytics", model_name)
        field = model._meta.get_field(field_name)
        if isinstance(field, models.JSONField):
            continue
        schema_editor.execute(f"""
            ALTER TABLE {model._meta.db_table}
            ALTER COLUMN {field.column} TYPE jsonb USING hstore_to_json({field.column}),
            ALTER COLUMN {field.column} SET DEFAULT '{{}}'::jsonb
        """)
    return upgrade
```

Suggested replacement for the SQL script:

```suggestion
                migrations.RunPython(
                    code=upgrade_field("apiusagebucket", "labels"),
                    reverse_code=migrations.RunPython.noop,
                ),
                migrations.RunPython(
                    code=upgrade_field("apiusageraw", "labels"),
                    reverse_code=migrations.RunPython.noop,
                ),
                migrations.RunPython(
                    code=upgrade_field("featureevaluationbucket", "labels"),
                    reverse_code=migrations.RunPython.noop,
                ),
                migrations.RunPython(
                    code=upgrade_field("featureevaluationraw", "labels"),
                    reverse_code=migrations.RunPython.noop,
                ),
```

krahllucas:
any updates?

khvn26:
@krahllucas Hello, sorry for the radio silence here — we've just prioritised this and are planning to release it until the end of Q1.

api/app_analytics/migrations/0007_labels_jsonb.py
khvn26: Good idea in principle, but I think `isinstance(field, models.JSONField)` would check the Django model state rather than the actual DB column type. Since this runs inside `SeparateDatabaseAndState`, the state operations have already declared the field as `JSONField`, so the check would always be true and the conversion would never execute, even when the DB actually has `hstore` columns.

The raw SQL approach intentionally introspects `pg_type` to check the real column type. We could use `RunPython` with the same `pg_type` query, but that would just be wrapping the same SQL in Python without much benefit.

emyller (approved):
LGTM

PR #5845, opened 2025-07-28
feat(feature-health): Backend for health event dismissal

### LINKED ISSUE #5842: Feature Health: Backend for Feature health event dismissal
1. For the `/api/v1/projects/{project_pk}/feature-health/events/` route, expose the `id` attribute.

2. Add `POST /api/v1/projects/{project_pk}/feature-health/events/{id}/dismiss` route that:
 - Adds a `HEALTHY` event for the provider/external_id combination of the provided event, if it's `UNHEALTHY`.
 - Triggers the task to reassign the appropriate feature health tags.

khvn26:

## Changes

Closes #5842.

- Adds the "id" field to health event objects in `GET /api/v1/projects/{project_pk}/feature-health/events/`.
- Adds a `POST /api/v1/projects/{project_pk}/feature-health/events/{id}/dismiss/` endpoint.

Adds two typecheck exclusions but removes three to maintain a net positive.

Added functional tests for the new endpoint and updated existing tests to verify that the `"id"` attribute is returned by the existing endpoint.


api/features/feature_health/services.py
```diff
@@ -107,3 +112,46 @@ def update_feature_unhealthy_tag(feature: "Feature") -> None:
...
+        )
+        return
+
+    logger.warning(
+        "feature-health-event-dismissal-not-supported",
+        feature_health_event_id=feature_health_event.id,
+        feature_health_event_external_id=feature_health_event.external_id,
+        feature_health_event_type=feature_health_event.type,
+        provider_name=feature_health_event.provider_name,
+    )
```
Zaimwa9: Maybe we could reverse the block and just early return there

api/features/feature_health/services.py
```diff
@@ -107,3 +112,46 @@ def update_feature_unhealthy_tag(feature: "Feature") -> None:
...
+        }
+
+        FeatureHealthEvent.objects.create(
+            feature=feature_health_event.feature,
+            environment=feature_health_event.environment,
+            type=FeatureHealthEventType.HEALTHY,
+            reason=json.dumps(reason),
+            provider_name=feature_health_event.provider_name,
+            external_id=feature_health_event.external_id,
+        )
```
Zaimwa9: What would be the other alternatives? I'm a bit concerned about a couple of points. 
- We are basically piling a new event on top of the unhealthy one to "cover" it.
- We are not updating at all the previous one. In a way we are completely dependent on `get_latest_by_*` or `external_id` + `created_at`
- The model is based on simple history but we create a new history chain here

It feels like we are hiding the unhealthy event under the rug while it doesn't necessarily mean it is healthy, it could also be irrelevant.

I'm wondering whether it is worth the effort (you know better - and also if there are use cases for this) to introduce a DISMISSED type or `dismissed_from` (with self FK) ?

Zaimwa9: Also, wouldn't it create a misleading healthy audit log from the newly created history instance 
(`create_audit_log_from_historical_record`)?

Does it make sense to track a dismissed audit log too ?

Zaimwa9 (changes requested):
Overall solid, I just have some comments on how we dismiss the events right now

api/features/feature_health/services.py
khvn26: Hmm... what is the benefit of this?

api/features/feature_health/services.py
khvn26: So the way the feature health events work is:

1. They're supposed to represent a specific event at a specific point in time, so they're immutable once written. It's designed this way to be able to present feature health timelines in the future.
2. The feature health status is based on whether we have a latest written UNHEALTHY event for a given feature_id and a unique provider+external_id pair. 

I'd argue that we're doing the _opposite_ of hiding the unhealthy event under the rug by preserving it in its original state. At the same time you're offering to alter it, which looks _more_ like sweeping it under the rug to me.

The generated audit log will look like the following:

```
Health status changed to Healthy for feature Test Feature.

Reason:
Manually dismissed by testuser@example.com.
```

Let me know if it's misleading or not — looks pretty clear to me.

api/features/feature_health/services.py
Zaimwa9: Looks more natural to me for the brain to discard the negative case early and focus on the happy (larger) path

api/features/feature_health/services.py
emyller: I mostly concur with the points above. Will add:

> The model is based on simple history but we create a new history chain here

@khvn26 Why is history necessary for the `FeatureHealthEvent` model particularly?

> introduce a DISMISSED type or `dismissed_from` (with self FK)

This, or at least specifying in `reason` that the feature health state was dismissed, would be a good idea IMO.

> Does it make sense to track a dismissed audit log too ?

👍

khvn26: > @khvn26 Why is history necessary for the FeatureHealthEvent model particularly?

To create audit logs.

> This, or at least specifying in reason that the feature health state was dismissed, would be a good idea IMO.

This is exactly what's happening.

api/features/feature_health/services.py
khvn26: Okay, fixed in 3280aeb5a920c9707f920fe0f87052dd5d1e5062.

emyller (approved):

Zaimwa9 (approved):
👍

PR #5848, opened 2025-07-29
fix: check-split-length-before-accessing-with-index

### LINKED ISSUE #4951: Logging in with an invalid email address (without an '@' character) causes 500
Sentry Issue: [FLAGSMITH-API-51Z](https://flagsmith.sentry.io/issues/6194522716/?referrer=github_integration)

```
IndexError: list index out of range
(11 additional frame(s) were not displayed)
...
  File "custom_auth/views.py", line 47, in post
```

Zaimwa9:

closes #4951 
## Changes
- Check split length before accessing

Added test in models
To reproduce:
Try login with invalid email 

New error (without auth-controller cf this [PR](https://github.com/Flagsmith/flagsmith-auth-controller/pull/28)):

``` {
    "non_field_errors": [
        "Unable to log in with provided credentials."
    ]
}
```

### DIFF
--- a/api/custom_auth/serializers.py
+++ b/api/custom_auth/serializers.py
@@ -1,8 +1,12 @@
+import re
 from typing import Any
 
 from common.core.utils import is_saas
 from django.conf import settings
-from djoser.serializers import UserCreateSerializer  # type: ignore[import-untyped]
+from djoser.serializers import (  # type: ignore[import-untyped]
+    TokenCreateSerializer,
+    UserCreateSerializer,
+)
 from rest_framework import serializers
 from rest_framework.authtoken.models import Token
 from rest_framework.exceptions import PermissionDenied
@@ -18,6 +22,15 @@ from .constants import (
     INVALID_PASSWORD_ERROR,
     USER_REGISTRATION_WITHOUT_INVITE_ERROR_MESSAGE,
 )
+
+EMAIL_REGEX = re.compile(r"^[^@]+@[^@]+\.[^@]+$")
+
+
+class CustomTokenCreateSerializer(TokenCreateSerializer):  # type: ignore[misc]
+    def validate_email(self, value: str) -> str:
+        if not EMAIL_REGEX.match(value):
+            raise serializers.ValidationError("Invalid email format.")
+        return value
 
 
 class CustomTokenSerializer(serializers.ModelSerializer):  # type: ignore[type-arg]
--- a/api/custom_auth/serializers.py
+++ b/api/custom_auth/serializers.py
@@ -23,7 +23,7 @@ from .constants import (
     USER_REGISTRATION_WITHOUT_INVITE_ERROR_MESSAGE,
 )
 
-EMAIL_REGEX = re.compile(r"^[^@]+@[^@]+\.[^@]+$")
+EMAIL_REGEX = re.compile(r"^[^@]+@(?:\w+\.)+\w{2,}$")
 
 
 class CustomTokenCreateSerializer(TokenCreateSerializer):  # type: ignore[misc]
--- a/api/integrations/lead_tracking/hubspot/services.py
+++ b/api/integrations/lead_tracking/hubspot/services.py
@@ -69,7 +69,9 @@ def register_hubspot_tracker(
 def create_self_hosted_onboarding_lead(
     email: str, first_name: str, last_name: str, organisation_name: str
 ) -> None:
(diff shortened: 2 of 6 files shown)


api/integrations/lead_tracking/hubspot/services.py
```diff
@@ -69,7 +69,9 @@ def register_hubspot_tracker(
 def create_self_hosted_onboarding_lead(
     email: str, first_name: str, last_name: str, organisation_name: str
 ) -> None:
-    email_domain = email.split("@")[1]
+    email_parts = email.split("@")
+    if len(email_parts) > 1:
+        email_domain = email_parts[1]
```
emyller: This is prone to cause a name error a few lines further when reading from `email_domain`, which may not exist.

Though more important than that: what should happen to creating a company in Hubspot when we can't determine the email address domain?

api/tests/unit/users/test_unit_users_models.py
```diff
@@ -234,5 +234,13 @@ def test_delete_user():  # type: ignore[no-untyped-def]
...
-    assert FFAdminUser(email="test@example.com").email_domain == "example.com"
+@pytest.mark.parametrize(
+    "email, expected_email_domain",
+    [
+        ("test@example.com", "example.com"),
+        ("test@subdomain.example.co.uk", "subdomain.example.co.uk"),
+        ("incorrect_email", None),
+    ],
+)
+def test_user_email_domain_property(email, expected_email_domain):  # type: ignore[no-untyped-def]
```
emyller:
```suggestion
def test_user_email_domain_property(email: str, expected_email_domain: str | None) -> None:
```

api/users/models.py
```diff
@@ -194,7 +194,10 @@ def full_name(self):  # type: ignore[no-untyped-def]
 
     @property
     def email_domain(self):  # type: ignore[no-untyped-def]
-        return self.email.split("@")[1]
+        email_parts = self.email.split("@")
+        if len(email_parts) > 1:
+            return email_parts[1]
+        return None
```
emyller: This looks good as functionality **only if** domain-less email addresses exist in our database or use cases. Otherwise, I'd rather prevent invalid email addresses at input validation layer, or perhaps even a database constraint if really necessary.

emyller (changes requested):

api/integrations/lead_tracking/hubspot/services.py
Zaimwa9: Right good catch thanks.
I fallbacked to the organisation name in that case here

api/users/models.py
Zaimwa9: looks better. Overrid the token serializer to validate there but i'd prefer not touch the database as this is not so critical

api/custom_auth/serializers.py
```diff
@@ -19,6 +23,15 @@
...
 
+EMAIL_REGEX = re.compile(r"^[^@]+@[^@]+\.[^@]+$")
+
+
+class CustomTokenCreateSerializer(TokenCreateSerializer):  # type: ignore[misc]
+    def validate_email(self, value: str) -> str:
+        if not EMAIL_REGEX.match(value):
+            raise serializers.ValidationError("Invalid email format.")
+        return value
+
```
emyller: Nice work, but I think we still need a bit of improvement to the validation.

1. The regular expression above does not protect inner layers of the code. It will allow invalid addresses, e.g. `"foo@..!.bar."`.
2. Potentially better than implementing email validation ourselves: perhaps we could override the `email` field in the serializer with an [EmailField](https://www.django-rest-framework.org/api-guide/fields/#emailfield) instance?

If a regex proves to be the right tool, here's an improvement:

```
^[^@]+@(?:\w+\.)+\w{2,}$
```

api/integrations/lead_tracking/hubspot/services.py
emyller: Sorry — unresolved because I don't see the fix to the potential `NameError`.

Zaimwa9: mmh didn't commit it

emyller: I see [2f0368e](https://github.com/Flagsmith/flagsmith/pull/5848/commits/2f0368ef32fd79a13b100cf511cf22c822c2756a) now, thanks!

emyller (approved):

PR #5851, opened 2025-07-29
docs: kapa ai widget

kyle-ssg:

## Changes

_Please describe._

_Please describe._


docs/src/theme/Layout/index.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import OriginalLayout from '@theme-original/Layout';
+import CrispLoader from './crisp';  // adjust path if you moved it
```
Zaimwa9: Not directly related but there is a log in the util function setting crisp (`setupCrip`) that we could clean with this one

Zaimwa9 (changes requested):
looks good code wise (following [here](https://docs.kapa.ai/integrations/website-widget/installation/general) and docusaurus). One minor clean-up so it's done 🙏 
Will there be a way to track the usage of kapa?

docs/src/theme/Layout/index.js
kyle-ssg: Right I see, so make the app use a similar hook to docs? It's unrelated to PR but yeah we could

Zaimwa9 (approved):

PR #5852, opened 2025-07-29
chore: fix codeowners

matthewelwell:
## Changes

Fix root path in CODEOWNERS

Github validation.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,5 +1,7 @@
+# Global default rule
 *                @flagsmith/flagsmith-maintainers
 
+# Specific rules based on directory
 /api/            @flagsmith/flagsmith-back-end
 /docs/           @flagsmith/flagsmith-docs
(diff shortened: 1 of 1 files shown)


khvn26 (dismissed):
For some reason, the file is not valid when checked on https://codeowners.org/ unless I remove the blank line 2.

matthewelwell:
> For some reason, the file is not valid when checked on https://codeowners.org/ unless I remove the blank line 2.

Yeah, not entirely sure why. But I have validated the format added here: https://github.com/Flagsmith/flagsmith/pull/5852/commits/7f23473affef5ce6b84030da3f59fdaf7f77dd5b on codeowners.org - I feel like the comments are somewhat superfluous, but they help I guess to visually separate the sections that we were trying to achieve with just whitespace previously.

khvn26 (approved):

Issue #5811, opened 2025-07-23
As a Flagsmith dashboard user, I should not be allowed to change the value of a feature that is currently in a release pipeline


gagantrivedi:
## Acceptance Criteria

- [ ] If a feature is 'in-flight' in a release pipeline, I should not able to modify the state, value or segment overrides for that feature in any environment
- [ ] If a feature has not been added to any release pipeline, I should be able to modify the feature as normal
- [ ] If a feature has completed a release pipeline, I should be able to modify the feature as normal
- [ ] Regardless of a feature's status in a release pipeline, I should be able to modify the metadata about it (e.g. the description, tags, etc.)

PR #5853, opened 2025-07-29
feat: disable value and segment change for in-flight feature

tiagoapolo:

## Changes

[#5811](https://github.com/Flagsmith/flagsmith/issues/5811)

Screenshot 2025-07-29 at 17 01 17

Screenshot 2025-07-29 at 17 01 13

_Please describe._


Zaimwa9 (approved):
Approving but we really need to deal with CreateFlag. There are lot of redundancy (out of scope ofc) and deep down I don't think this should be wrapping so much JSX but it's too much refacto otherwise

tiagoapolo:
> Approving but we really need to deal with CreateFlag. There are lot of redundancy (out of scope ofc) and deep down I don't think this should be wrapping so much JSX but it's too much refacto otherwise

Yes, initially I started doing some refactors but it was too much for a simple task

Issue #5730, opened 2025-07-08
Add the CRUD endpoints to manage code references against a given feature


matthewelwell:
## Task list

- [x] Define the data model to store code references in the database.
    Recommendation:
    ```
    FeatureFlagCodeReferencesScan:
        project: Project
        created_at: datetime
        repository_url: str
        revision: str
        code_references:
          - feature_name: <str>
            file_path: <str>
            line_number: <int>
          - ...
        ...
    ```
- [x] Implement view to submit code references (from CI).
    ```yaml
    # /api/v1/projects/{project_pk}/feature-code-references/
    repository_url: https://github.com/Flagsmith/flagsmith
    revision: <commit_hash>
    code_references:
      - feature_name: segment_cloning
        file_path: api/something/foo/bar.py
        line_number: 32
      - feature_name: <str>
        file_path: <str>
        line_number: <int>
      - ...
    ```
- [x] Fetch code references for **one** feature flag (for the feature detail UI).
    ```yaml
    # /api/v1/projects/{project_pk}/features/{feature_pk}/code-references/
    - repository_url: <str>
      last_successful_repository_scanned_at: <timestamp>
      last_feature_found_at: <timestamp>
      code_references:
        - file_path: <str>
          line_number: <int>
          permalink: <str>  # For now only GitHub
          vcs_revision: <str>
        - ...
    - 
    ```
- [x] Add code reference **counts** to the feature list endpoint (for the list UI).
    ```yaml
    # /api/v1/projects/{project_pk}/features/
    ...
    results:
      - id: <int>
        name: <str>
        ...
        code_references_counts:
          - repository_url: <str>
            count: <int>
            last_successful_repository_scanned_at: <timestamp>
            last_feature_found_at: <timestamp>
          - ...
      - ...
    ```
- [x] POST to the new endpoint from the CI job (#5731).

> [!NOTE]
> - URLs and schemas above were aligned between @Zaimwa9 and @emyller in effort to keep things simple though flexible for extension after the PoC.
> - JSON is suggested as to avoid flooding the database with a multi-table spamming of tuples.
> - We might need indexes to help filtering in JSON blobs by "feature_name", "created_at".

Issue #5624, opened 2025-06-19
Epic: Add OIDC support to SSO


matthewelwell:
We should develop support for OIDC to extend our options when offering SSO. Currently we only offer SAML. 

The OIDC support should ideally mimic the SAML integration as closely as possible - although there are likely improvements we can make around auto provisioning. For example, right now we don't support SCIM provisioning - is there an analogous approach we can implement in OIDC.

PR #5854, opened 2025-07-30
ci(github): Find feature flag references in the code

### LINKED ISSUE #5731: Build the github action logic to run on a crontab to identify code references for each feature in a given project

emyller:
Closes #5731.

## Changes

Introduce CI logic for a PoC to find code references of feature flags in the codebase.

See GitHub Actions [job run](https://github.com/Flagsmith/flagsmith/actions/runs/16610970606/job/46993839176?pr=5854), or the relevant output when the link expires:

```
Code References:

Feature: announcement
  - frontend/web/components/Announcement.tsx:22
  - frontend/web/components/Announcement.tsx:35

Feature: announcement_per_page
  - frontend/web/components/AnnouncementPerPage.tsx:28
  - frontend/web/components/AnnouncementPerPage.tsx:35

Feature: butter_bar
  - frontend/web/components/ButterBar.tsx:90
  - frontend/web/components/ButterBar.tsx:94

Feature: clone_segment
  - frontend/web/components/segments/SegmentRow/SegmentRow.tsx:45

Feature: context_values
  - frontend/web/components/segments/Rule/components/RuleConditionPropertySelect.tsx:64
  - frontend/web/components/segments/Rule/components/RuleConditionRow.tsx:75

Feature: delighted
  - frontend/web/project/api.ts:148

Feature: disable_create_org
  - frontend/common/utils/utils.tsx:109
  - frontend/web/components/pages/CreateOrganisationPage.js:49

Feature: disable_oauth_registration
  - frontend/web/components/pages/HomePage.tsx:181

Feature: disable_users_as_reviewers
  - frontend/web/components/modals/ChangeRequestModal.tsx:188
  - frontend/web/components/modals/ChangeRequestModal.tsx:194
  - frontend/web/components/modals/ChangeRequestModal.tsx:262
  - frontend/web/components/pages/ChangeRequestDetailPage.js:420

Feature: environment_metrics
  - frontend/web/components/pages/FeaturesPage.js:231

Feature: feature_health
  - frontend/web/components/FeatureRow.tsx:224
  - frontend/web/components/navigation/EnvironmentAside.tsx:56
  - frontend/web/components/navigation/EnvironmentAside.tsx:86
  - frontend/web/components/navigation/EnvironmentAside.tsx:138
  - frontend/web/components/pages/ProjectSettingsPage.js:557
  - frontend/web/components/tables/TableTagFilter.tsx:38
  - frontend/web/components/tags/AddEditTags.tsx:38
  - frontend/web/components/tags/Tag.tsx:73

Feature: feature_versioning
  - frontend/web/components/StaleFlagWarning.tsx:18
  - frontend/web/components/pages/EnvironmentSettingsPage.tsx:558
  - frontend/web/components/pages/ProjectSettingsPage.js:203

Feature: identity_aliases
  - frontend/web/components/pages/UserPage.tsx:281
  - frontend/web/components/pages/UsersPage.tsx:81

Feature: inspect_permissions
  - frontend/web/components/users-permissions/OrganisationUsersTable/components/OrganisationUsersTableRow.tsx:36

Feature: integration_data
  - frontend/common/utils/utils.tsx:364
  - frontend/web/components/IntegrationSelect.tsx:29

Feature: integration_onboarding
  - frontend/web/components/pages/GettingStartedPage.tsx:154

Feature: max_api_calls_alert
  - frontend/web/components/OrganisationLimit.tsx:81

Feature: oauth_github
  - frontend/web/components/pages/HomePage.tsx:199
  - frontend/web/components/pages/HomePage.tsx:206

Feature: oauth_google
  - frontend/web/components/pages/HomePage.tsx:214
  - frontend/web/components/pages/HomePage.tsx:219

Feature: organisation_integrations
  - frontend/web/components/navigation/navbars/OrganisationNavbar.tsx:47

Feature: payments_enabled
  - frontend/web/components/OrganisationLimit.tsx:80
  - frontend/web/components/modals/Payment.js:431
  - frontend/web/components/pages/OrganisationSettingsPage.js:170
  - frontend/web/components/pages/UsersAndPermissionsPage.tsx:82

Feature: project_metrics_tooltip
  - frontend/web/components/navigation/navbars/ProjectNavbar.tsx:24
  - frontend/web/components/navigation/navbars/ProjectNavbar.tsx:77

Feature: read_only_mode
  - frontend/web/components/ButterBar.tsx:91
  - frontend/web/components/ButterBar.tsx:97
  - frontend/web/components/FeatureRow.tsx:223
  - frontend/web/components/pages/FeaturesPage.js:229

Feature: relay_events_key
  - frontend/common/stores/account-store.js:106
  - frontend/common/stores/account-store.js:108

Feature: release_pipelines
  - frontend/web/components/modals/CreateFlag.js:1079
  - frontend/web/components/modals/FlagValueFooter.tsx:45
  - frontend/web/components/navigation/navbars/ProjectNavbar.tsx:93

Feature: restrict_project_create_to_admin
  - frontend/web/components/pages/OrganisationSettingsPage.js:271

Feature: saml
  - frontend/web/components/pages/HomePage.tsx:242
  - frontend/web/components/pages/HomePage.tsx:243

Feature: segment_operators
  - frontend/common/utils/utils.tsx:531

Feature: show_dunning_banner
  - frontend/web/components/ButterBar.tsx:103

Feature: split_testing
  - frontend/web/components/navigation/navbars/EnvironmentNavbar.tsx:108

Feature: sso_idp
  - frontend/web/components/pages/HomePage.tsx:251
  - frontend/web/components/pages/HomePage.tsx:257

Feature: test_webhook_backend
  - frontend/web/components/TestWebhook.tsx:75

Feature: try_it
  - frontend/web/components/TryIt.js:63

Feature: use_identity_overrides_in_local_eval
  - frontend/web/components/pages/EnvironmentSettingsPage.tsx:819

Feature: verify_seats_limit_for_invite_links
  - frontend/common/stores/account-store.js:24
  - frontend/web/components/pages/UsersAndPermissionsPage.tsx:84

Feature: welcome_page
  - frontend/web/components/App.js:193
  - frontend/web/components/navigation/navbars/TopNavbar.tsx:27
  - frontend/web/components/pages/CreateOrganisationPage.js:33
```


.github/workflows/poc-github-code-references.yml
```diff
@@ -0,0 +1,113 @@
...
+on:
+  schedule:
+    - cron: '0 0 * * *'  # Runs daily at midnight UTC
+  workflow_dispatch:
+
+env:
+  FLAGSMITH_PUBLIC_KEY: ENktaJnfLVbLifybz34JmX
+  FLAGSMITH_EDGE_API_URL: https://edge.api.flagsmith.com
+  PYTHON_VERSION: '3.13'
+  PYTHON_REQUESTS_VERSION: '2.32.4'
```
khvn26: nit: consider using `uvx`.

https://docs.astral.sh/uv/guides/scripts/#using-a-shebang-to-create-an-executable-file

example: https://github.com/Flagsmith/edge-api/blob/main/scripts/assert_boto_version.py

.github/workflows/poc-github-code-references.yml
```diff
@@ -0,0 +1,113 @@
...
+          import json
+          import os
+          import re
+          from collections import deque
+          from pathlib import Path
+          from typing import Generator
+
+          import requests
+
+          EXCLUDE_PATTERNS = ["node_modules", "venv", ".git", "cache", "build", "htmlcov", "docs", ".json"]
```
khvn26: I imagine this will need to be made configurable.

.github/workflows/poc-github-code-references.yml
```diff
@@ -0,0 +1,113 @@
...
+
+          def is_file_binary(file_path: Path) -> bool:
+              """Check if a file is binary."""
+              with file_path.open("rb") as file:
+                  chunk = file.read(1024)
+              return b'\0' in chunk
```
khvn26: Maybe ok for PoC, but doesn't look very reliable. Consider using something like https://github.com/github-linguist/linguist instead.

.github/workflows/poc-github-code-references.yml
```diff
@@ -0,0 +1,113 @@
...
+                      continue
+                  context: deque[str] = deque(maxlen=2)
+                  with path.open("r", encoding="utf-8", errors="ignore") as file:
+                      for line_number, line in enumerate(file, start=1):
+                          context.append(line)
+                          for feature_name in feature_names:
+                              if feature_name not in line:  # Match feature name
+                                  continue
+                              if (
+                                  re.search(fr"""[A-Z][A-Z0-9_]{3,}\s*=\s*(['"]){feature_name}\1""", line) or  # Constant assignment
```
khvn26: That's a weirdly specific pattern that won't catch lines like

```
abcd = "my_feature"
ABC = "my_feature"
```

Any reason to exclude lowercase/short variable names from it?

.github/workflows/poc-github-code-references.yml
```diff
@@ -0,0 +1,113 @@
...
+          print("Feature names:", feature_names)
+
+          # Find code references
+          code_references = [
+              (feature_name, file_path, line_number)
+              for feature_name, file_path, line_number in find_references(feature_names)
+          ]
+
+          # Output to GHA
+          json_references = json.dumps(code_references)
```
khvn26: I think it would make sense to use a more structured format here especially since there's no native tuple support in JSON :)

```json
[{"feature_name": "foo", "file_path": "src/bar.js", "line_number": 42}]
```

...or even something more compact like

```json
[{"feature_name": "foo", "ref": "src/bar.js:42:15"}]
```

.github/workflows/poc-github-code-references.yml
```diff
@@ -0,0 +1,113 @@
...
+                      for line_number, line in enumerate(file, start=1):
+                          context.append(line)
+                          for feature_name in feature_names:
+                              if feature_name not in line:  # Match feature name
+                                  continue
+                              if (
+                                  re.search(fr"""[A-Z][A-Z0-9_]{3,}\s*=\s*(['"]){feature_name}\1""", line) or  # Constant assignment
+                                  re.search(fr"""(?ism:(?:feature|flag)[^\(]*\(.*(["']){feature_name})\1""", "".join(context))  # Function call
+                              ):
+                                  yield feature_name, str(path), line_number
```
khvn26, line 84: The patterns used look specific enough to output the column number as well?

.github/workflows/poc-github-code-references.yml
```diff
@@ -0,0 +1,113 @@
...
+              references_by_feature[feature_name].append((file_path, line_number))
+
+          print("Code References:")
+          for feature_name, references in references_by_feature.items():
+              print(f"\nFeature: {feature_name}")
+              for file_path, line_number in references:
+                  print(f"  - {file_path}:{line_number}")
+
+      # TODO
+      # - name: Upload code references
```
khvn26, line 127: If the plan is to merge this TODO, I'd like to see an issue detailing out subsequent work for this workflow.

.github/workflows/poc-github-code-references.yml
```diff
@@ -0,0 +1,113 @@
...
+  contents: read
+
+on:
+  schedule:
+    - cron: '0 0 * * *'  # Runs daily at midnight UTC
+  workflow_dispatch:
+
+env:
+  FLAGSMITH_PUBLIC_KEY: ENktaJnfLVbLifybz34JmX
+  FLAGSMITH_EDGE_API_URL: https://edge.api.flagsmith.com
```
khvn26: These maybe should be inputs.

nit: We don't use "public key" anywhere in our docs, so this setting/input name should reference an Environment API key.

.github/workflows/poc-github-code-references.yml
```diff
@@ -0,0 +1,113 @@
+name: 'PoC: GitHub Code References'
+permissions:
+  contents: read
```
khvn26, line 3: Not important for the PoC, but one of the reasons I'd like this to be an action is that users will need to explicitly define their permission for the token, which, I imagine, is a right thing.

.github/workflows/poc-github-code-references.yml
```diff
@@ -0,0 +1,113 @@
...
+
+          references_by_feature = defaultdict(list)
+          for feature_name, file_path, line_number in sorted(code_references):
+              references_by_feature[feature_name].append((file_path, line_number))
+
+          print("Code References:")
+          for feature_name, references in references_by_feature.items():
+              print(f"\nFeature: {feature_name}")
+              for file_path, line_number in references:
+                  print(f"  - {file_path}:{line_number}")
```
khvn26: I think a single step could handle both displaying and storing the output.

(Or even better, make this a GitHub Action).

khvn26 (commented):
Looks promising overall!

Added some thoughts, questions and comments.

.github/workflows/poc-github-code-references.yml
khvn26: One more thought that occurred to me is that I'd love to see https://github.com/Flagsmith/flagsmith/issues/5624 auth here to optimise the integration flow, and ensure authentic code reference updates as well. [Depot actions](https://depot.dev/docs/container-builds/reference/github-actions) are an excellent reference.

.github/workflows/poc-github-code-references.yml
emyller: Most definitely. This is intended to be explored after the PoC:
1. Need to come up with a nice semi-silver-bullet default. The list here only serves ourselves (and this PoC).
2. It will be made a reusable workflow / action parameter, so the Python line here is really temporary.

Anyway I've moved it into an env var just for the sake of this being prettier right now. See b28c4120434db64c4b241553919d9c6b8d2e9eb4.

.github/workflows/poc-github-code-references.yml
emyller: > Any reason to exclude lowercase (...) variable names from it?

Yes. This is to match feature names when they're set as constants, which are usually written in uppercase in some languages, e.g. Python. Note that this was intended for the PoC and would be very inaccurate elsewhere.

> Any reason to exclude (...) short variable names from it?

Kind of. While we don't enforce a minimum length on feature names, the scenario where feature flag names have very short names **and** are defined as constants did not speak louder than the risk of false positives.


That said, this was **not** tested because our own codebase doesn't seem to adhere to this pattern. For now I've removed it (9b269c907f53e72cb8763b359f4da7b5a9827e3d) because it's not useful. Let's revisit this later?

.github/workflows/poc-github-code-references.yml
emyller: linguist is amazing indeed, but I'd **really** prefer to avoid its installation effort and CI time footprint only to use one of its many features.

I agree, however, this check isn't very reliable. I've took the opportunity to learn from linguist — and [CharlockHolmes](https://github.com/brianmario/charlock_holmes), the source of most of linguist's decisions — and improved the check: 148fc6556e9e348eb91e645300f2cf572b7338a0.

Thank you for mentioning it! I hope this is better now.

.github/workflows/poc-github-code-references.yml
emyller: Yes! Though I doubt they would be relevant in this product.

.github/workflows/poc-github-code-references.yml
emyller: https://github.com/Flagsmith/flagsmith/issues/5730

I've made it clearer in the issue description. cc @matthewelwell

.github/workflows/poc-github-code-references.yml
emyller: > These maybe should be inputs.

They will be for sure! As soon as this PoC code is made reusable.


> We don't use "public key" anywhere in our docs, so this setting/input name should reference an Environment API key.

I'd argue it should be "environment key" as per [the documentation](https://docs.flagsmith.com/clients/rest#public-sdk-api-endpoints), though deep down I wanted it to sound slightly more _educational_. Fixed in b436a9dba3dbd0ea8c21a29b18688565250b4148.

> One more thought that occurred to me is that I'd love to see https://github.com/Flagsmith/flagsmith/issues/5624 auth here to optimise the integration flow, and ensure authentic code reference updates as well.

That's brilliant. Added to https://github.com/Flagsmith/flagsmith/issues/5868

.github/workflows/poc-github-code-references.yml
emyller: This will become an action, or maybe a reusable workflow, or a mix of the two, depending on which architecture favors both privacy and transparency the most. For now I see a group of granular actions and/or workflows that, put together, will complete the goal. Braindump here:

1. A reusable workflow to **communicate with the Flagsmith API** and fetch features, and **only** output the feature names — this might be enriched later.
2. An action to employ **clear Flagsmith-authored code** and find references of features obtained in component (1) in **customer's code already [checked-out](https://github.com/actions/checkout)**, and **only** output those references as JSON.
3. A reusable workflow to **communicate with the Flagsmith API** and submit the JSON content obtained from component (2).

On the customer end, the three components can be bundled together within a single GitHub Actions workflow, based on a copy-pastable example that will exist in our documentation. Or perhaps installable via the [GitHub Actions UI](https://github.com/Flagsmith/flagsmith/actions/new). 👀 

The reason why [reusable workflows](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows) can be great in components (1) and (3) above is that they isolate the one step (2) that needs access to code only run a snippet to find stuff but **never** communicate with the API. Also, hopefully, the YAML format makes it easier to see that access to code does not happen, i.e. [actions/checkout](https://github.com/actions/checkout) is never used.

Although we **could** bundle it all in an action — or in a Docker image for provider portability — I'd love to make it very explicit in the ins and outs of CI jobs/steps what's happening so we shine at transparency.

This is still up for debate: will the architecture proposal help selling the product, and maybe help portability, or is it just overcomplicating? 👀 (we have time to discuss)

.github/workflows/poc-github-code-references.yml
emyller: > I think a single step could handle both displaying and storing the output.

That's correct. The separation here is only to surface the separation of concerns discussed in the [other thread](https://github.com/Flagsmith/flagsmith/pull/5854#discussion_r2244214412), and experiment with the output of the `collect` step while there's nowhere to send it yet.

> Or even better, make this a GitHub Action

We will! For now since we already know the capabilities of actions (and reusable workflows), we don't need to experiment with those separately. I believe the core of this PoC is to get feature flag references only of our own API repo and see them in the UI. This is why I named this workflow `poc-...` — it's meant to be temporary. I hope that works for you!

.github/workflows/poc-github-code-references.yml
emyller: The compact version looks the coolest. I chose the verbose one (8c6257f14e677e477837aae53c84bb389c68f677) because it will be easier to create GitHub permalinks from — they use URI fragments to point to line numbers, AFAIK.

.github/workflows/poc-github-code-references.yml
emyller: I gave it a spin here: 5999c8d1d88647b1b90804b6316df17704120ef9

It feels nice to experiment with [PEP 723](https://peps.python.org/pep-0723/) but I'm not sure what the gain is just yet. Do you think we should keep it? TBH I do feel a bit concerned about relying on 0.x libraries.

khvn26: `uv` is not a library, rather a package manager, and a really good one too. It's moving steadily to become a de-facto standard for the Python ecosystem.

PEP 723 + `uvx` is just a nicer way to run a standalone Python script nowadays. Anyone will be able to run the code from any environment without deducing the dependencies from custom env vars. It's much faster, too, especially with [caching](https://github.com/Flagsmith/edge-api/blob/a39646e2cc6bb7b39f77179e0bc7e6493072aa16/.github/workflows/pull-request.yml#L30-L33).

.github/workflows/poc-github-code-references.yml
khvn26: In my experience, reusable workflows are only great when shared inside a repo/organisation. As a vendor integration point, they tend to work poorly, because they are not composable and unable to produce structured outputs.

> The reason why [reusable workflows](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows) can be great in components (1) and (3) above is that they isolate the one step (2) that needs access to code only run a snippet to find stuff but never communicate with the API. Also, hopefully, the YAML format makes it easier to see that access to code does not happen, i.e. [actions/checkout](https://github.com/actions/checkout) is never used.

`actions/checkout` is being used now, though, and I'm having a hard time understanding how we are going to exclude it if we're planning to ship a the reusable workflow. This would be a usability problem as well, because we'll unavoidably make assumptions about which code the customers want to check out and how exactly they need it checked out. 

If I'm understanding GHA correctly, the only actual way to not checkout the code ourselves would be shipping an Action.

I acknowledge and share the privacy concerns, and, IMO, they are easily addressed if we opt to ship two Actions. Consider this hypothetical example:

```yaml
steps:
 - uses: actions/checkout@v4
 - uses: flagsmith/get-flag-references@v1
   id: flag-refs
 - run: printf ${{ steps.flag-refs.outputs.flag-refs }} # something custom with the refs
 # the following could be run from another job with different token permissions 
 # to ensure no access to code or secrets
 - uses: flagsmith/upload-flag-references@v1
   with:
    flag-refs: ${{ steps.flag-refs.outputs.flag-refs }}
```

From a GHA user standpoint, this is very clean, because one can compose these actions as they see fit. A reusable workflow, however, would force a user to delegate one whole job to the entire routine (which kind of defeats the whole security considerations point if you think more about it).

In my experience, a well-documented action with clear inputs/outputs makes it much easier for the users to trust and adopt an integration.

khvn26 (dismissed):
I’d lean toward putting this in a dedicated repo — it would help us start thinking earlier about how the integration should look, and also avoid adding more clutter to the main repo’s `.github` directory.

Otherwise, the approach looks solid to me, aside from the packaging/integration concerns I flagged in the review comments above.

.github/workflows/poc-github-code-references.yml
emyller: Thank you, improved in 59ae90969693567e26201c9ca8a8dda0991d1f24.

.github/workflows/poc-github-code-references.yml
emyller: > (...) because they are not composable and unable to produce structured outputs.

They [can](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows#using-outputs-from-a-reusable-workflow). 😉 

The thing about **actions** is that they **can** access code whenever they come after `actions/checkout`. For example:

```yaml
steps:
  - uses: actions/checkout@v4
  - uses: flagsmith/get-flag-references@v1
    id: flag-refs
  - uses: flagsmith/upload-flag-references@v1  # This action CAN see the code
    with:
      flag-refs: ${{ steps.flag-refs.outputs.flag-refs }}
```

What I have in mind, **for after the PoC**:

```yaml
jobs:
  fetch-feature-flags:  # Communicates with Flagsmith
    uses: flagsmith/flagsmith-ci/.github/workflows/fetch-feature-flags.yml@1
    with:
      flagsmith_environment_key: ...

  collect-code-references:  # DOES NOT communicate with Flagsmith
    needs: [fetch-feature-flags]
    outputs:
      code_references: ${{ steps.collect.outputs.code_references }}
    steps:
      - uses: actions/checkout@v4
      - uses: flagsmith/collect-code-references@v1
        id: collect
        with:
          features: ${{ needs.fetch-feature-flags.outputs.features }}
  
  upload-code-references:  # Communicates with Flagsmith
    needs: [collect-code-references]
    uses: flagsmith/flagsmith-ci/.github/workflows/upload-code-references.yml@1
    with:
      code_references: ${{ needs.collect-code-references.outputs.code_references }}
```

Well TBH even `collect-code-references` could be a reusable workflow. I would be certainly easier to read than an action and the simpler architecture might even help port the functionality to other providers.

khvn26: > They [can](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows#using-outputs-from-a-reusable-workflow). 😉

Yeah, sorry, I lost context of this one. I guess I'm more concerned that we'd _force_ people to utilise a dedicated job. I do understand how it relates to privacy/security concerns, but it does seem a bit clunky to me.

> even `collect-code-references` could be a reusable workflow

Given your example, and my earlier concern that the users will want to control the `actions/checkout` inputs, I can't see that happening (at least in a way that produces a good integration UX).

> would be certainly easier to read than an action

That's an interesting point. In my experience, actions are easier to understand because they represent single atomic operations, and reusable workflows usually exist to abstract away complexity specific to the org/repo. As a longtime GHA user, I'd be reluctant to rely on an external reusable workflow. I'd be happy to be proven wrong by a good real-world example/reference, though. The only example of a vendor-provided reusable workflow that I'm familiar with is Uffizzi, and it's not too good.

emyller: Well this _could_ be a very good example soon, though I see the importance of offering actions in the future. Do you think we can move forward with the PoC being this temporary workflow?

khvn26: Yes absolutely — I did approve the PR and the thread starts with the words "Not important for the PoC"!

khvn26 (approved):

PR #5865, opened 2025-07-30
fix: prevent unauthorised access from group permissions

matthewelwell:
## Changes

This PR prevents users that were part of groups in organisations that they no longer belong to from accessing the resources in that organisation. 

From testing, the only way that this was possible in the first place was if the user was removed from the organisation in one of the following non-standard ways: 

 1. The user was removed via django admin 
 2. The user was removed as part of the logic to remove all users but the first one when a subscription is cancelled or downgraded

Note that this PR also includes a (tested) migration to remove any users from groups in organisations they do not belong to. 

Added a lot of new unit tests.


api/permissions/permission_service.py
matthewelwell, line 1: I'm not entirely convinced that the extra checks that I've added here are the most optimal way of adding this logic, but I couldn't find a single entrypoint to the permissions service which would allow me to add it. 

Keen to get @gagantrivedi 's opinion on this topic.

api/users/migrations/0044_remove_users_from_groups_in_orgs_they_do_not_belong_to.py
```diff
@@ -0,0 +1,29 @@
...
+
+logger = structlog.get_logger()
+
+
+def remove_users_from_groups(apps: Apps, schema_editor: BaseDatabaseSchemaEditor) -> None:
+    user_permission_group_membership_model = apps.get_model("users", "UserPermissionGroupMembership")
+    queryset = user_permission_group_membership_model.objects.exclude(
+        ffadminuser__organisations=F("userpermissiongroup__organisation")
+    )
+    logger.info("Removing orphan group memberships.", num_orphan_memberships=queryset.count())
```
matthewelwell, line 17: I've purposefully added one such record to the staging database. We should verify in the logs on staging that this migration behaves as we expect before running the migration in production.

api/permissions/permission_service.py
gagantrivedi: Yeah, doesn't look like there is a better way

gagantrivedi (dismissed):

matthewelwell:
Superseded by #5893

PR #5866, opened 2025-07-30
fix: `SEGMENT_RULES_CONDITIONS_EXPLICIT_ORDERING_ENABLED` does not affect rules

### LINKED ISSUE #5669: Segment rule ordering is not persisted

khvn26:

## Changes

Makes `SEGMENT_RULES_CONDITIONS_EXPLICIT_ORDERING_ENABLED` affect segment rules in addition to segment rule conditions.

Will be tested extensively on staging.

### DIFF
--- a/api/segments/models.py
+++ b/api/segments/models.py
@@ -32,21 +32,23 @@ from projects.models import Project
 
 from .managers import LiveSegmentManager, SegmentManager
 
+T = typing.TypeVar("T", bound=models.Model)
+
 logger = logging.getLogger(__name__)
 
 
-class ConfiguredOrderManager(SoftDeleteExportableManager):
+class ConfiguredOrderManager(SoftDeleteExportableManager, models.Manager[T]):
     def get_queryset(
         self,
-    ) -> models.QuerySet["Condition"]:
+    ) -> models.QuerySet[T]:
         # Effectively `Condition.Meta.ordering = ("id",) if ... else ()`,
         # but avoid the weirdness of a setting-dependant migration
         # and having to reload everything in tests
-        qs: models.QuerySet["Condition"]
+        qs: models.QuerySet[T]
         if settings.SEGMENT_RULES_CONDITIONS_EXPLICIT_ORDERING_ENABLED:
-            qs = super().get_queryset().order_by("id")
+            qs = SoftDeleteExportableManager.get_queryset(self).order_by("id")
         else:
-            qs = super().get_queryset()
+            qs = SoftDeleteExportableManager.get_queryset(self)
         return qs
 
 
@@ -177,7 +179,7 @@ class Segment(
         # Ensure top-level rules are cloned first (for dependencies)
         source_rules = source_rules.order_by(models.F("rule").asc(nulls_first=True))
 
-        rule_to_cloned_rule_map: dict[SegmentRule, SegmentRule] = {}
+        rule_to_cloned_rule_map: dict[SegmentRule | None, SegmentRule] = {}
         for rule in source_rules:
             cloned_rule = deepcopy(rule)
             cloned_rule.pk = None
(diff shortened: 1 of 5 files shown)


api/segments/models.py
```diff
@@ -35,6 +35,21 @@
...
 
 
+class ConfiguredOrderManager(SoftDeleteExportableManager):
+    def get_queryset(
+        self,
+    ) -> models.QuerySet["Condition"]:
+        # Effectively `Condition.Meta.ordering = ("id",) if ... else ()`,
+        # but avoid the weirdness of a setting-dependant migration
+        # and having to reload everything in tests
+        qs: models.QuerySet["Condition"]
```
matthewelwell: The typing here is wrong now that we're using it in the `SegmentRule` as well. Can we pass in a `TypeVar` somehow, or should we just break the DRY?

matthewelwell (dismissed):

api/segments/models.py
khvn26: Fixed in e794143db2af9655012c2b2f6f7161830bbf925b.

api/segments/models.py
```diff
@@ -32,21 +32,23 @@
 
 from .managers import LiveSegmentManager, SegmentManager
 
+T = typing.TypeVar("T", bound=models.Model)
```
emyller: nit: Perhaps a better name for the generic type?

```suggestion
Model = typing.TypeVar("Model", bound=models.Model)
```

api/segments/models.py
```diff
@@ -32,21 +32,23 @@
...
-    ) -> models.QuerySet["Condition"]:
+    ) -> models.QuerySet[T]:
         # Effectively `Condition.Meta.ordering = ("id",) if ... else ()`,
         # but avoid the weirdness of a setting-dependant migration
         # and having to reload everything in tests
-        qs: models.QuerySet["Condition"]
+        qs: models.QuerySet[T]
         if settings.SEGMENT_RULES_CONDITIONS_EXPLICIT_ORDERING_ENABLED:
-            qs = super().get_queryset().order_by("id")
+            qs = SoftDeleteExportableManager.get_queryset(self).order_by("id")
```
emyller: Untested:

```suggestion
            qs = super(SoftDeleteExportableManager, self).get_queryset().order_by("id")
```

api/segments/models.py
```diff
@@ -177,7 +179,7 @@ def copy_rules_and_conditions_from(self, source_segment: "Segment") -> None:
         # Ensure top-level rules are cloned first (for dependencies)
         source_rules = source_rules.order_by(models.F("rule").asc(nulls_first=True))
 
-        rule_to_cloned_rule_map: dict[SegmentRule, SegmentRule] = {}
+        rule_to_cloned_rule_map: dict[SegmentRule | None, SegmentRule] = {}
```
emyller: This looks incorrect to me. Can you explain?

emyller (changes requested):

api/segments/models.py
khvn26: Fixed in 4afdd8591821e7d4f53a301f8914c7f72a9721e1.

api/segments/models.py
khvn26: Fixed in 1cef50a1b5a122513b90a79796fdd98a58ca9827.

api/segments/models.py
khvn26: Fixed in b3695750d65aaf991af65faaacb98155431ce6ec.

`rule.rule`'s type is `SegmentRule | None`. Annoyingly, you have to pass an object with the key type to `dict.get()` even if you expect a missing value as part of your logic.

mypy didn't detect this before because now rule/condition querysets are actually correctly typed.

emyller (approved):
Looks neat! :shipit:

PR #5867, opened 2025-07-30
chore: Include Features, Segments in evaluation context schema

### LINKED ISSUE #5808: Include Features and Segments in the Evaluation context schema
In order to implement https://github.com/Flagsmith/flagsmith-engine/issues/234, we need to add the following properties to the Evaluation context schema:

- `features`: a map of feature keys to feature objects that include a feature name and feature environment default value(s) with its (their) respective weight(s).
- `segments`: a map of segment keys to segment objects that include a segment name, a list of segment rules, and a map of the segment's feature overrides.

Additionally, we need to add an `EvaluationResult` schema that describes the following object:

```js
{
    "context": context, // context used for evaluation
    "flags": [] // a list of evaluated flag values
    "segments": [] // a list of segments that the context matched with
}
```

In addition to the above, we need to answer the following question: is it valuable to give the newly added 'features' and 'segments' context keys access to the segment rules, i.e. to allow the implementation of the following rules:

- `$.segments.sysadmins.overrides.butter_bar IS NOT "Attention! Server Room Flooding Imminent!"`
- `$.features.meta_feature_controlling_certain_segments.enabled IS true`

khvn26:

## Changes

Closes https://github.com/Flagsmith/flagsmith/issues/5808.

This PR adds:
- An `EvaluationResult` schema that includes flag and used segment information.
- `segments` and `features` keys to the `EvaluationContext` schema to have all data required for evaluation in a single DTO.

The design goal was to have a structure that is extensible, simple enough to parse and read through, and that provides all required engine data, excluding any data unrelated to engine.

Some new vocabulary is added here that, IMO, make sense for evaluation:
- String `"key"` fields are preferred to numeric `id` fields to more clearly state their purpose.
- A `"variants"` key is used to represent multivariate feature states, where each variant includes a `"weight"` indicating its distribution percentage. I considered using a unified `"values"` array for all feature states — including non-multivariate ones (with a single `{ "weight": 100 }` entry) — but opted against it to avoid ambiguous semantics around control values and the need to validate against empty arrays.

Both objects will be used by the new engine interface to be introduced in https://github.com/Flagsmith/flagsmith-engine/issues/234.

An example `EvaluationContext` object:

```json
{
  "environment": {
    "key": "abcde",
    "name": "Test Environment"
  },
  "features": {
    "feature_1": {
      "key": "1",
      "name": "feature_1",
      "enabled": true,
      "value": "foobar"
    },
    "feature_2": {
      "key": "2",
      "name": "feature_2",
      "enabled": false,
      "value": "control",
      "variants": [
        {
          "value": "a",
          "weight": 51.0
        },
        {
          "value": "b",
          "weight": 49.0
        }
      ]
    }
  },
  "segments": {
    "test_segment": {
      "key": "1",
      "name": "test_segment",
      "rules": [
        {
          "type": "ALL",
          "sub_rules": [
           {
          "type": "ANY",
          "conditions": [
            {
              "path": "$.identity.identifier",
              "operator": "IN",
              "value": "id1,id2"
            },
            {
              "path": "some-trait",
              "operator": "EQUALS",
              "value": "foobar"
            }
          ]
        }
          ]
        }
      ]
    }
  }
}
```

An example `EvaluationResult` object:

```json
{
  "context": {...},
  "flags": [
    {
      "enabled": false,
      "value": "foobar",
      "feature": {
        "key": "1",
        "name": "feature_1"
      }
    },
    {
      "enabled": true,
      "value": null,
      "feature": {
        "key": "2",
        "name": "feature_2"
      }
    }
  ],
  "segments": [
    {
      "key": "1",
      "name": "test_segment"
    }
  ]
}
```

N/A

### DIFF
--- a/sdk/evaluation-context.json
+++ b/sdk/evaluation-context.json
@@ -47,7 +47,6 @@
                     "type": "string"
                 },
                 "variants": {
-                    "default": [],
                     "description": "An array of environment default values associated with the feature. Contains a single value for standard features, or multiple values for multivariate features.",
                     "items": {
                         "$ref": "#/$defs/FeatureValue"
@@ -57,7 +56,7 @@
                 }
             },
             "required": [
-                "id",
+                "key",
                 "name",
                 "enabled",
                 "value"
@@ -127,9 +126,9 @@
         "SegmentContext": {
             "description": "Represents a segment context for feature flag evaluation.",
             "properties": {
-                "id": {
-                    "description": "The unique identifier for the segment.",
-                    "title": "ID",
+                "key": {
+                    "description": "Key used for % split segmentation.",
+                    "title": "Key",
                     "type": "string"
                 },
                 "name": {
@@ -138,46 +137,84 @@
                     "type": "string"
                 },
                 "rules": {
-                    "additionalProperties": {
+                    "items": {
                         "$ref": "#/$defs/SegmentRule"
                     },
                     "title": "Rules",
-                    "type": "object"
+                    "type": "array"
                 }
             },
             "required": [
-                "id",
+                "key",
                 "name",
                 "rules"
             ],
             "title": "SegmentContext",
             "type": "object"
         },
+        "SegmentCondition": {
+            "description": "Represents a condition within a segment rule for feature flag evaluation.",
+            "properties": {
+                "path": {
+                    "description": "The path to the trait or value in the evaluation context.",
+                    "title": "Property",
+                    "type": "string"
+                },
+                "operator": {
+                    "description": "The operator to use for evaluating the condition.",
+                    "title": "Operator",
+                    "type": "string",
+                    "enum": [
+                        "EQUAL",
+                        "GREATER_THAN",
+                        "LESS_THAN",
+                        "LESS_THAN_INCLUSIVE",
+                        "CONTAINS",
+                        "GREATER_THAN_INCLUSIVE",
+                        "NOT_CONTAINS",
+                        "NOT_EQUAL",
+                        "REGEX",
+                        "PERCENTAGE_SPLIT",
+                        "MODULO",
+                        "IS_SET",
+                        "IS_NOT_SET",
+                        "IN"
+                    ]
+                },
+                "value": {
+                    "description": "The value to compare against the trait or context value.",
+                    "title": "Value",
+                    "type": "string"
+                }
+            },
+            "required": [
+                "path",
+                "operator",
+                "value"
+            ],
+            "title": "SegmentCondition",
+            "type": "object"
+        },
         "SegmentRule": {
             "description": "Represents a rule within a segment for feature flag evaluation.",
             "properties": {
-                "name": {
-                    "description": "The name of the segment rule.",
-                    "title": "Name",
-                    "type": "string"
+                "type": {
+                    "description": "Segment rule type. Represents a logical quantifier for the conditions and sub-rules.",
+                    "title": "Type",
+                    "type": "string",
+                    "enum": [
+                        "ALL",
+                        "ANY",
+                        "NONE"
+                    ]
                 },
                 "conditions": {
-                    "additionalProperties": {
-                        "anyOf": [
-                            {
-                                "type": "string"
-                            },
-                            {
-                                "type": "integer"
-                            },
-                            {
-                                "type": "boolean"
-                            }
-                        ]
+                    "description": "Conditions that must be met for the rule to apply.",
+                    "items": {
+                        "$ref": "#/$defs/SegmentCondition"
                     },
-                    "description": "Conditions that define the segment rule.",
                     "title": "Conditions",
-                    "type": "object"
+                    "type": "array"
                 },
                 "sub_rules": {
                     "additionalProperties": {
@@ -189,7 +226,7 @@
                 }
             },
             "required": [
-                "name",
+                "type",
                 "conditions"
             ],
             "title": "SegmentRule",
--- a/sdk/evaluation-context.json
+++ b/sdk/evaluation-context.json
@@ -211,7 +211,7 @@
                     "title": "Conditions",
                     "type": "array"
                 },
-                "sub_rules": {
+                "rules": {
                     "items": {
                         "$ref": "#/$defs/SegmentRule"
                     },
--- a/sdk/evaluation-context.json
+++ b/sdk/evaluation-context.json
@@ -50,7 +50,7 @@
                     "type": "array"
                 },
                 "priority": {
-                    "description": "Indicates the priority of the feature context. Higher values indicate higher priority in case when multiple feature contexts for the same feature are applicable.",
+                    "description": "Priority of the feature context. Higher values indicate a higher priority when multiple contexts apply to the same feature.",
                     "title": "Priority",
                     "type": "number"
                 }
--- a/sdk/evaluation-context.json
+++ b/sdk/evaluation-context.json
@@ -23,6 +23,68 @@
             "title": "EnvironmentContext",
             "type": "object"
         },
+        "FeatureContext": {
+            "description": "Represents a feature context for feature flag evaluation.",
+            "properties": {
+                "key": {
+                    "description": "Key used when selecting a value for a multivariate feature. Set to an internal identifier or a UUID, depending on Flagsmith implementation.",
+                    "title": "Key",
+                    "type": "string"
+                },
+                "name": {
+                    "description": "The name of the feature.",
+                    "title": "Name",
+                    "type": "string"
+                },
+                "enabled": {
+                    "description": "Indicates whether the feature is enabled in the environment.",
+                    "title": "Enabled",
+                    "type": "boolean"
+                },
+                "value": {
+                    "description": "A default environment value for the feature. If the feature is multivariate, this will be the control value.",
+                    "title": "Value",
+                    "type": "string"
+                },
+                "variants": {
+                    "default": [],
+                    "description": "An array of environment default values associated with the feature. Contains a single value for standard features, or multiple values for multivariate features.",
+                    "items": {
+                        "$ref": "#/$defs/FeatureValue"
+                    },
+                    "title": "Variants",
+                    "type": "array"
+                }
+            },
+            "required": [
+                "id",
+                "name",
+                "enabled",
+                "value"
+            ],
+            "title": "FeatureContext",
+            "type": "object"
+        },
+        "FeatureValue": {
+            "description": "Represents a multivariate value for a feature flag.",
+            "properties": {
+                "value": {
+                    "description": "The value of the feature.",
+                    "title": "Value",
+                    "type": "string"
+                },
+                "weight": {
+                    "description": "The weight of the feature value variant, as a percentage number (i.e. 100.0).",
+                    "title": "Weight",
+                    "type": "number"
+                }
+            },
+            "required": [
+                "value"
+            ],
+            "title": "FeatureValue",
+            "type": "object"
+        },
         "IdentityContext": {
             "description": "Represents an identity context for feature flag evaluation.",
             "properties": {
@@ -39,12 +101,20 @@
                     "x-flagsmith-engine-path": "$.identity.key"
                 },
                 "traits": {
-                    "additionalProperties": {
-                        "type": "string"
-                    },
+                    "anyOf": [
+                        {
+                            "additionalProperties": {
+                                "type": "string"
+                            },
+                            "type": "object"
+                        },
+                        {
+                            "type": "null"
+                        }
+                    ],
+                    "default": null,
                     "description": "A map of traits associated with the identity, where the key is the trait name and the value is the trait value.",
-                    "title": "Traits",
-                    "type": "object"
+                    "title": "Traits"
                 }
             },
             "required": [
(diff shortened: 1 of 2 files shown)


Zaimwa9 (commented):
Looks solid overall. One piece feels missing to make the evaluation fully representative imo - but let me know if it's really useful. 

Right now we have:
- the flag being evaluated
- the identity under evaluation
- the segments that can override
- the final evaluation result

What could be missing in the result is the why do I have those values. We know which segment overid but we don't have necessarily its values (especially thinking with multi-variate where the possibilities are wider).
Do you think it would make sense to surface the override override configuration (blue 40%/red60% instead of blue80%/Green20%) ?

sdk/evaluation-context.json
```diff
@@ -23,6 +23,68 @@
...
         },
+        "FeatureContext": {
+            "description": "Represents a feature context for feature flag evaluation.",
+            "properties": {
+                "key": {
+                    "description": "Key used when selecting a value for a multivariate feature. Set to an internal identifier or a UUID, depending on Flagsmith implementation.",
+                    "title": "Key",
+                    "type": "string"
+                },
+                "name": {
```
gagantrivedi, line 39: Are the feature name and environment used in the evaluation, or are they just there to make things more readable?

sdk/evaluation-context.json
```diff
@@ -23,6 +23,68 @@
...
+                    "description": "An array of environment default values associated with the feature. Contains a single value for standard features, or multiple values for multivariate features.",
+                    "items": {
+                        "$ref": "#/$defs/FeatureValue"
+                    },
+                    "title": "Variants",
+                    "type": "array"
+                }
+            },
+            "required": [
+                "key",
```
emyller, line 69: What if not a multivariate feature?

sdk/evaluation-context.json
```diff
@@ -39,12 +101,20 @@
...
+                        {
+                            "additionalProperties": {
+                                "type": "string"
+                            },
+                            "type": "object"
+                        },
+                        {
+                            "type": "null"
+                        }
+                    ],
```
emyller: I can't understand how this maps to the trait data model, or if that's even the intention. I might be unfamilar?

sdk/evaluation-context.json
```diff
@@ -53,12 +123,121 @@
...
+                "rules": {
+                    "items": {
+                        "$ref": "#/$defs/SegmentRule"
+                    },
+                    "title": "Rules",
+                    "type": "array"
+                }
+            },
+            "required": [
+                "key",
```
emyller, line 166: Again, it looks odd that `key` is required given not every segment uses `%` rules.

sdk/evaluation-context.json
```diff
@@ -53,12 +123,121 @@
...
+                },
+                "conditions": {
+                    "description": "Conditions that must be met for the rule to apply.",
+                    "items": {
+                        "$ref": "#/$defs/SegmentCondition"
+                    },
+                    "title": "Conditions",
+                    "type": "array"
+                },
+                "sub_rules": {
```
emyller: Maybe important to note that in the API, these are named [just "rules"](https://github.com/Flagsmith/flagsmith/blob/88dd8a78735d975289fbca535e3b7f142961d8ab/api/segments/serializers.py#L75) instead.

sdk/evaluation-context.json
```diff
@@ -53,12 +123,121 @@
...
+                    "items": {
+                        "$ref": "#/$defs/SegmentRule"
+                    },
```
emyller, line 240: Do we have to, or can we, ensure there's a depth limit to recursion of this field?

sdk/evaluation-context.json
```diff
@@ -53,12 +123,121 @@
...
+                "name",
+                "rules"
+            ],
+            "title": "SegmentContext",
+            "type": "object"
+        },
+        "SegmentCondition": {
+            "description": "Represents a condition within a segment rule for feature flag evaluation.",
+            "properties": {
+                "path": {
```
emyller: Maybe ["property"](https://github.com/Flagsmith/flagsmith/blob/635dd77bce207dc6a8486f776f5327d732ccba91/api/segments/models.py#L313)?

sdk/evaluation-context.json
khvn26: 1. Environment name is used in evaluation as the "Environment Name" context value (`$.environment.name`).

2. `$.features[<name>].name` does look useless once you think about it — removed in cdce3cf1fd241bcb4e0dc679d1d25f6a2b4d6052.

sdk/evaluation-context.json
khvn26: We'll expect all Flagsmith implementations to provide a feature key regardless of the feature type. If, for whatever reason, a Flagsmith implementation can not provide a feature key (which is usually a readily available database pk or a uuid), an empty string is considered a valid key as well.

sdk/evaluation-context.json
khvn26: A few things to unpack here.

1. The `anyOf` definition here means that `traits` can be `null`, or a string to string mapping (that's what the `additionalProperties` with `type: string` definition means). See [here](https://tour.json-schema.org/content/03-Objects/02-Additional-Properties) and [here](https://tour.json-schema.org/content/06-Combining-Subschemas/05-Valid-Against-anyOf-the-Subschemas(OR)) for more details.
2. As with other instances of `anyOf` with `null` in this schema, this mimics Python's `attr: Optional[T] = None` behaviour on a dataclass, and I'm not entirely sure if it belongs to the schema. Hopefully, we're not going to behave differently for null/missing fields, so, perhaps, simply not including a field in `required` is sufficient to denote that the property is optional.
3. The change from an array to a mapping is required for the engine to access traits through the `$.identity.traits[<trait_name>]` JSONPath. This allows Flagsmith implementations to hold segment condition properties like `$.identity.traits.organisation_id`.
4. The `string` trait value type is not entirely honest here. I think we should come up with a union type here to reflect all we can 

I'm inclined to change the `additionalProperties` type to `["number", "string", "boolean", "null"]`, and wait on your overall thoughts/feedback on the rest of the above.

sdk/evaluation-context.json
khvn26: As mentioned in the [other comment](https://github.com/Flagsmith/flagsmith/pull/5867#discussion_r2260129180), it's easier to expect a key here than making it optional.

sdk/evaluation-context.json
khvn26: We could probably change this definition to an `anyOf` and come up with two schemas with different sets of required properties, but I think it'll unnecessarily complicate code generation.

sdk/evaluation-context.json
khvn26: Yeah, makes sense to keep the uniformity here. Fixed in 1b164d2d6ec4e5da507fd642024248d9995754cd.

sdk/evaluation-context.json
khvn26: Not in the schema, no.
I'd love to avoid the recursion altogether, of course, but that would be a significant design effort for little gain, IMO.

sdk/evaluation-context.json
khvn26: The idea was to reference the `x-flagsmith-engine-path` metadata here, which will be used by FE to dynamically populate Context values. Maybe we should change the metadata key to `x-flagsmith-engine-property` instead?

sdk/evaluation-context.json
emyller: Thanks for the explanation!

> 2. (...) Hopefully, we're not going to behave differently for null/missing fields, so, perhaps, simply not including a field in required is sufficient to denote that the property is optional.

Agreed, unless `"traits": null` is also valid input.

> 3. The change from an array to a mapping is required (...).

It makes total sense! The type seems to be a mapping already prior to this PR though — I assume it changed recently along with the context values project?

> 4. The string trait value type is not entirely honest here. (...)

Agreed.

In my understanding, here's what we should have, in addition to deleting `"default": null,` below.

```suggestion
                    "additionalProperties": {
                        "type": ["number", "string", "boolean"]
                    },
```

sdk/evaluation-context.json
emyller: Can you give an example of `x-flagsmith-engine-path` that can be referenced here?

khvn26: See one for the Environment name, for example: https://github.com/Flagsmith/flagsmith/blob/b3e6a55903269a0d675ebb9862629944fb30bd3d/sdk/evaluation-context.json#L16-L17

emyller: @khvn26 Yup saw that one but I can't understand how it ties with an object that could be nested close to other similar ones.

sdk/evaluation-context.json
emyller: This was also recursive in the SegmentSerializer until recently. In order to prevent it, here's the [strategy I took](https://github.com/Flagsmith/flagsmith/blob/4018f07d423fffde133fd6512e06c52f7ade2e37/api/segments/serializers.py#L47-L88).

I agree there is little gain here as the serializer now refuses more than one level of rule nesting, so the improvement is optional.

sdk/evaluation-context.json
khvn26: So the following values are legit for this `path` property:
- `$.identity.identifier`
- `$.identity.key`
- `$.environment.name`
- `$.identity.traits.some_trait`
- `some_trait` (which will translate internally to `$.identity.traits.some_trait`)

emyller: Sorry, I did not ask the question correctly — I might lack the full context. I'll reach out internally as the discussion will likely be irrelevant to the PR, and resolve or add to this thread later. Thanks though for the answers!

sdk/evaluation-context.json
```diff
@@ -23,6 +23,67 @@
...
+                "variants": {
+                    "description": "An array of environment default values associated with the feature. Contains a single value for standard features, or multiple values for multivariate features.",
+                    "items": {
+                        "$ref": "#/$defs/FeatureValue"
+                    },
+                    "title": "Variants",
+                    "type": "array"
+                },
+                "priority": {
+                    "description": "Indicates the priority of the feature context. Higher values indicate higher priority in case when multiple feature contexts for the same feature are applicable",
```
emyller:
```suggestion
                    "description": "Indicates the priority of the feature context. Higher values indicate higher priority when multiple feature contexts for the same feature are applicable.",
```

khvn26:
@Zaimwa9 Thanks! Based on your comment, I made the following improvements:
- Added `SegmentContext.overrides` in f55780ce9ac7fceee773c60b5fde3905755d1e5b. This field will contain the segment feature overrides. Since identity overrides will be mapped to segments internally, this covers the identity overrides as well.
- Added `FeatureContext.priority` in 181164fc03bb7145f22ac047e025d4dd259c5f2f to support identity overrides, that will have maximum priority, and the segment order for feature-specific overrides.
- Added `FlagResult.reason` in b3e6a55903269a0d675ebb9862629944fb30bd3d. The intent is to support those of OpenFeature's [resolution reasons](https://openfeature.dev/specification/types/#resolution-reason) that make sense for Flagsmith, and, since the OF spec allows arbitrary values here, maybe include matched segment names.

sdk/evaluation-context.json
khvn26: Fixed in 9d7c6ae44cfcdefe56801030a7e2afb52ff3fcd0 👍

sdk/evaluation-context.json
emyller: Following up. I understand now how the metadata bit is intended to work — thanks again. The suggested changes are:

1. Name the metadata key `"x-flagsmith-engine-path"` to `"x-flagsmith-engine-condition-property"` — it sounds boring I know, though clearer.
2. This:
```suggestion
                "property": {
```

khvn26: Agreed 👍 Fixed in e8efc32573021e6d26b1554446ca956b4a11d6ee.

sdk/evaluation-context.json
```diff
@@ -23,28 +23,94 @@
...
+                "variants": {
+                    "description": "An array of environment default values associated with the feature. Contains a single value for standard features, or multiple values for multivariate features.",
+                    "items": {
+                        "$ref": "#/$defs/FeatureValue"
+                    },
+                    "title": "Variants",
+                    "type": "array"
+                },
+                "priority": {
+                    "description": "Indicates the priority of the feature context. Higher values indicate higher priority in case when multiple feature contexts for the same feature are applicable.",
```
emyller:
```suggestion
                    "description": "Indicates the priority of the feature context. Higher values indicate higher priority when multiple feature contexts for the same feature are applicable.",
```

If you agree.

emyller (dismissed):
One optional improvement. Looks great regardless!

sdk/evaluation-context.json
khvn26: Improved in cfa5af0d5d5f1944fc0f09f9590518b30f2fc377.

Zaimwa9 (dismissed):
looks neat. Let's see it in action!

gagantrivedi (dismissed):

emyller (dismissed):

sdk/evaluation-context.json
khvn26: Brought it back c67d7b892521602b7ce8303cbb87308b652c1df4 — the field is now used by engine to return `FlagResult` objects.

khvn26:
I had to add some stuff while finalising https://github.com/Flagsmith/flagsmith-engine/pull/239/files:

- I brought back `FeatureContext.name` in c67d7b892521602b7ce8303cbb87308b652c1df4 to simplify serialisation of `FlagResult`s.
- Added `FeatureContext.feature_key` in 04103f3e13cfa9ccba4af0f2142faea8821b4ff9. Feature (not feature state) ids are used to correctly apply overrides, and are surfaced in `FlagResult` objects to maintain backwards compatibility. Additionally, I removed nesting from `FlagResult`.
- Fixed incorrect `priority` field documentation in f50fb30205bb60d5d8760e53276fc46b60c9ee58.

emyller (dismissed):

Zaimwa9 (approved):

PR #5877, opened 2025-07-31
refactor(tests): combine `staff_user` and `test_user` fixtures

matthewelwell:
## Changes

Refactor to combine staff_user and test_user fixtures which serve the same purpose as identified in #5865 . 

This PR only changes test code, no actual business logic is changing. The test code should behave in exactly the same way, I have just migrated from one fixture (which was removed in this PR) to another. Where I have modified tests, I have also added typing. 

Existing unit tests.

### DIFF
--- a/api/conftest.py
+++ b/api/conftest.py
@@ -9,10 +9,11 @@ import pytest
 from common.environments.permissions import (
     MANAGE_IDENTITIES,
     MANAGE_SEGMENT_OVERRIDES,
+    UPDATE_FEATURE_STATE,
     VIEW_ENVIRONMENT,
     VIEW_IDENTITIES,
 )
-from common.projects.permissions import VIEW_PROJECT
+from common.projects.permissions import CREATE_ENVIRONMENT, DELETE_FEATURE, VIEW_PROJECT
 from django.contrib.contenttypes.models import ContentType
 from django.core.cache import caches
 from django.db.backends.base.creation import TEST_DATABASE_PREFIX
@@ -27,7 +28,6 @@ from pytest import FixtureRequest
 from pytest_django.fixtures import SettingsWrapper
 from pytest_django.plugin import blocking_manager_key
 from pytest_mock import MockerFixture
-from rest_framework.authtoken.models import Token
 from rest_framework.test import APIClient
 from task_processor.task_run_method import TaskRunMethod
 from urllib3 import BaseHTTPResponse
@@ -218,16 +218,6 @@ trait_value = "value1"
 
 
 @pytest.fixture()
-def test_user(django_user_model):  # type: ignore[no-untyped-def]
-    return django_user_model.objects.create(email="user@example.com")
-
-
-@pytest.fixture()
-def auth_token(test_user):  # type: ignore[no-untyped-def]
-    return Token.objects.create(user=test_user)
-
-
-@pytest.fixture()
 def admin_client_original(admin_user):  # type: ignore[no-untyped-def]
     client = APIClient()
     client.force_authenticate(user=admin_user)
@@ -252,12 +242,6 @@ def admin_client(admin_client_original):  # type: ignore[no-untyped-def]
     fixture will ultimately be updated to the new approach.
     """
     yield admin_client_original
-
-
-@pytest.fixture()
-def test_user_client(api_client, test_user):  # type: ignore[no-untyped-def]
-    api_client.force_authenticate(test_user)
-    return api_client
 
 
 @pytest.fixture()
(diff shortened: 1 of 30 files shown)


api/tests/integration/custom_auth/end_to_end/test_custom_auth_integration.py
```diff
@@ -574,20 +575,34 @@ def test_get_user_is_not_throttled(  # type: ignore[no-untyped-def]
         assert response.status_code == status.HTTP_200_OK
 
 
-def test_delete_token(test_user, auth_token):  # type: ignore[no-untyped-def]
+def test_delete_token(api_client: APIClient, db: None) -> None:
```
emyller, line 579: nit: IMO clearer than an unused mistyped argument.

```suggestion
@pytest.mark.django_db
def test_delete_token(api_client: APIClient) -> None:
```

api/conftest.py
```diff
@@ -804,28 +788,51 @@ def create_project_permission(db):  # type: ignore[no-untyped-def]
     return PermissionModel.objects.get(key=CREATE_PROJECT)
 
 
+@pytest.fixture()
+def create_environment_permission(db: None) -> PermissionModel:
```
emyller, line 792:
```suggestion
@pytest.mark.django_db
def create_environment_permission() -> PermissionModel:
```

api/tests/unit/organisations/invites/conftest.py
```diff
@@ -1,15 +1,23 @@
...
 
 
 @pytest.fixture()
-def invite(organisation, admin_user, test_user):  # type: ignore[no-untyped-def]
-    return Invite.objects.create(
-        organisation=organisation, email=test_user.email, invited_by=admin_user
+def invite(
+    organisation: Organisation,
+    admin_user: FFAdminUser,
+    staff_user: FFAdminUser,
```
emyller, line 18: The two names `admin_user` and `staff_user` make me wonder if "staff" is the right word we should pick. It sounds ambiguous to me because, in Django, the same term is also used to [designate users](https://github.com/django/django/blob/dca8284a376128c64bd0e0792ad12391ae3e7202/django/contrib/auth/models.py#L471-L475) with access to the admin site. Perhaps something like `organisation_admin_user`, or `organisation_owner` could be an improvement?

api/tests/integration/custom_auth/end_to_end/test_custom_auth_integration.py
```diff
@@ -574,20 +575,34 @@ def test_get_user_is_not_throttled(  # type: ignore[no-untyped-def]
...
+    register_data = {
+        "first_name": "test",
+        "last_name": "user",
+        "email": f"user{uuid.uuid4()}@example.com",
+        "password": password,
+        "re_password": password,
+    }
+    response = api_client.post(
+        register_url, data=json.dumps(register_data), content_type="application/json"
+    )
```
emyller, line 592: Is there value in going through the signup functionality in order to delete the user token? Could we perhaps just create the token for the current staff user?

api/tests/unit/organisations/invites/test_unit_invites_views.py
```diff
@@ -359,15 +370,22 @@ def test_join_organisation_returns_400_if_payment_fails(
...
+    invite_link: InviteLink,
+    settings: SettingsWrapper,
+    django_user_model: typing.Type[AbstractUser],
+    api_client: APIClient,
+) -> None:
     # Given
     settings.DISABLE_INVITE_LINKS = True
     url = reverse("api-v1:users:user-join-organisation-link", args=[invite_link.hash])
 
+    new_user = django_user_model.objects.create(email=f"user{uuid.uuid4()}@example.com")
```
emyller: Why not `FFAdminUser.objects.create`, as in `api/tests/unit/organisations/invites/test_unit_invites_views.py` [above](https://github.com/Flagsmith/flagsmith/pull/5877/files#diff-5b4fc11dbf51658f9f40c0580776f0b452b5f6b4e0859cdfc3ae3f08210b072eR166)?

api/tests/unit/permissions/permission_service/test_get_permitted_environments_for_user.py
```diff
@@ -30,15 +43,21 @@ def test_get_permitted_environments_for_user_returns_all_environments_for_org_ad
...
-def test_get_permitted_environments_for_user_returns_all_the_environments_for_project_admin(  # type: ignore[no-untyped-def]  # noqa: E501
-    test_user, environment, project, project_admin, project_two_environment
-):
+def test_get_permitted_environments_for_user_returns_all_the_environments_for_project_admin(  # noqa: E501
+    staff_user: FFAdminUser,
+    environment: Environment,
+    project: Project,
+    project_admin: typing.Union[
+        UserProjectPermission, UserPermissionGroupProjectPermission
+    ],
```
emyller, line 52: Out of scope note: May be a good idea to have a base type for these.

api/tests/unit/permissions/permission_service/test_get_permitted_environments_for_user.py
```diff
@@ -50,61 +69,83 @@ def test_get_permitted_environments_for_user_returns_all_the_environments_for_pr
...
         )
     # Next, let's give user some permissions using `user_permission`
     permissions_as_user = [VIEW_ENVIRONMENT, UPDATE_FEATURE_STATE]
-    environment_permission_using_user_permission.permissions.add(*permissions_as_user)
+    environment_permission_using_user_permission.permissions.add(
+        *[
+            PermissionModel.objects.get(key=permission)
+            for permission in permissions_as_user
+        ]
+    )
```
emyller:
```suggestion
    environment_permission_using_user_permission.permissions.add(
        *PermissionModel.objects.filter(key__in=permissions_as_user)
    )
```

Unless there's a reason to collect permissions beforehand? Similar cases below.

emyller (dismissed):
Looks like a solid improvement. A few non-blocking questions to maybe improve.

api/tests/integration/custom_auth/end_to_end/test_custom_auth_integration.py
matthewelwell: Hmm... interesting. I think we need to agree upon a consensus here, because the codebase currently favours the `db` fixture. It seems that your suggestion is [the recommended approach](https://pytest-django.readthedocs.io/en/latest/helpers.html#std-fixture-db). I'll update it here and hopefully we can slowly migrate to use the marker. 

image

image

api/conftest.py
matthewelwell: As per the link above, it seems as though the recommended approach is to use the `db` fixture for fixtures that need to access the database, and the marker for tests.

api/tests/unit/organisations/invites/conftest.py
matthewelwell: Yes, I do see the point, but we did deliberate over the naming for a while when the fixture was first created. I think this is actually a case where we should add some documentation, and the naming would be less important. 

In summary, we currently have 2 main user fixtures: `admin_user` and `staff_user`. Both are included as users of the `organisation` fixture, but the `admin_user` is an organisation admin, the `staff_user` is just a user with no permissions by default. 

Based on this, your naming suggestions wouldn't correctly describe the use case, but it's interesting that you thought that the `staff_user` was an organisation admin, because they are actually kind of the opposite :) perhaps a naming change would help here, but I think that the documentation is more important.

api/tests/integration/custom_auth/end_to_end/test_custom_auth_integration.py
matthewelwell: Since this test lives in our 'integration tests', which should refrain from direct database access, I decided to register the user via the API.

api/tests/unit/organisations/invites/conftest.py
emyller: Interesting, thanks for the explanation. I thought `admin_user` was only a regular superuser rather than an actual actor, as per [this fixture](https://pytest-django.readthedocs.io/en/latest/helpers.html#admin-user-an-admin-user-superuser).

I'd love to continue the discussion and maybe improve documentation separately.

emyller (approved):

PR #5891, opened 2025-08-05
feat: Add howdygo demos

kyle-ssg:

## Changes

Adds a generic way to show demoes of features

https://github.com/user-attachments/assets/f8ca53f1-60e6-4363-8e01-458e45699c17


Zaimwa9 (approved):
Nice!
Would almost think about how we could make the "view demo" more visible in a second step? Action button top right?

NIT improvement would be to have the close icon on the modal for accessibility

Issue #5846, opened 2025-07-29
Spike: Investigate simplifying Segment data model


khvn26:
Currently, we store segment rules (including sub-rules) and conditions in separate tables, which makes validation / versioning code unnecessarily complex. 

We need to investigate the following:
- Versioning requirements for the individual rules are currently unclear. We'd like to go over the current approach, informed by FE need to know about specific rule/condition versions, and re-evaluate whether it's more correct, effective, and user-friendly to harness simple text-based diffs instead.
- As a nice side effect, removal of `SegmentCondition` and `SegmentRule` models in favour of storing the whole segment data in a JSONField in the `Segment` model.

This spike should result in a PoC to give us an idea whether the change is feasible and/or beneficial in the long run.

Considerations:
- We'd like to have the data in Edge API / engine compatible format.
- This will require a data migration.

emyller:
As an update, since this discussion we're already pursuing some of the goals described in this issue:
- We've dismissed adding `version_of` to rules and conditions (https://github.com/Flagsmith/flagsmith/issues/5770), and made PKs irrelevant in API communication when rewriting segment serializers (https://github.com/Flagsmith/flagsmith/issues/5769).
- On the frontend, there is relevant work in progress by @kyle-ssg: [feat/segment-change-requests-simplified-diff](https://github.com/Flagsmith/flagsmith/tree/feat/segment-change-requests-simplified-diff).

We're prioritizing delivering segment change requests now, and will improve the data model towards a JSON approach later.

emyller:
Thanks @kyle-ssg! I'll reopen this issue because I think we still need to simplify things on the data model side, as per the description:

> As a nice side effect, removal of SegmentCondition and SegmentRule models in favour of storing the whole segment data in a JSONField in the Segment model.

We'll discuss it (cc @khvn26) and perhaps move onto a separate issue then.

khvn26:
Spike complete. Resulting epic: #7814.

Notes:

- The spike asked whether per-rule and per-condition versioning is needed. It is not. [#5770](https://github.com/Flagsmith/flagsmith/issues/5770) proposed `version_of` on rules and conditions and was dismissed in favour of string-based diffs; [#5769](https://github.com/Flagsmith/flagsmith/issues/5769) made primary keys irrelevant to API communication. The frontend already strips ids and diffs rendered text. Segment-level versioning via `version_of` is untouched.
[...]

Issue #5825, opened 2025-07-23
Clean up


emyller:
Need to refactor or delete all code from flagsmith, flagsmith-common and flagsmith-workflows that's associated with the old segment serializer and model APIs.

## Task list

- [x] Delete unused metadata and segment serializer code from flagsmith-common.
    - ✅ Completed as of https://github.com/Flagsmith/flagsmith/issues/5907.
- [x] Delete obsolete logic from flagsmith-workflows.
    - ✅ Completed as of https://github.com/Flagsmith/flagsmith-workflows/pull/72.
- [x] Import improvements from this [closed PR](https://github.com/Flagsmith/flagsmith/pull/5748/files):
    - Segment manager/model architecture
    - Segment tests
    - ✅ Completed in https://github.com/Flagsmith/flagsmith/pull/6265
- [x] Make `Segment.version.default = 1` at database level instead of using [a hook](https://github.com/Flagsmith/flagsmith/blob/7082001f55b8528f30d88ff1eb73cad22579ec76/api/segments/models.py#L103-L106).
    - ✅ Completed in https://github.com/Flagsmith/flagsmith/pull/6265
- [x] Make `Segment.version_of = null` the indication for a _live_ segment — right now live segments are those which `version_of` [point to themselves](https://github.com/Flagsmith/flagsmith/blob/7082001f55b8528f30d88ff1eb73cad22579ec76/api/segments/models.py#L108-L115), which adds unnecessary overhead and complexity.
    - ✅ Completed in https://github.com/Flagsmith/flagsmith/pull/6265
- [x] Inherit `MetadataSerializerMixin` from `Serializer` and maybe prevent repeating meta fields?
    - ✅ Completed in https://github.com/Flagsmith/flagsmith/pull/6265

PR #5898, opened 2025-08-06
refactor: Refactor segment cloning

emyller:
Contributes to https://github.com/Flagsmith/flagsmith/issues/5771.

> [!WARNING]
> Blocked by [#5769](https://github.com/Flagsmith/flagsmith/issues/5769).

## Changes

- Refactor segment cloning.
- Ensure committing a change request will apply its draft segments.

- Updated existing unit tests.
- New unit tests.

> [!NOTE]
> The `segment` fixture was also [creating a revision](https://github.com/Flagsmith/flagsmith/blob/main/api/conftest.py#L381-L383) of the generated segment, for a reason I can't make peace with. That was removed. I'm sorry segment version numbers are shifted `-1` in tests, but things should be clearer now I hope.


api/conftest.py
```diff
@@ -376,13 +375,8 @@ def project(organisation):  # type: ignore[no-untyped-def]
...
 @pytest.fixture()
-def segment(project: Project):  # type: ignore[no-untyped-def]
-    _segment = Segment.objects.create(name="segment", project=project)
-    # Deep clone the segment to ensure that any bugs around
-    # versioning get bubbled up through the test suite.
-    SegmentCloneService(_segment).deep_clone()
-
-    return _segment
+def segment(project: Project) -> Segment:
+    return Segment.objects.create(name="segment", project=project)  # type: ignore[no-any-return]
```
Zaimwa9: We can get rid of a couple of type ignores with
```
@pytest.fixture()
def segment(project: Project) -> Segment:
    segment: Segment = Segment.objects.create(name="segment", project=project)
    return segment
```

api/segments/models.py
```diff
@@ -114,35 +114,72 @@ def set_version_of_to_self_if_none(self):  # type: ignore[no-untyped-def]
...
         cloned_segment.save()
 
-        self.version += 1  # type: ignore[operator]
-        self.save_without_historical_record()
+        cloned_segment.copy_rules_and_conditions_from(self)
+
+        # Handle versioning
+        version_of = self if is_revision else cloned_segment
+        cloned_segment.version_of = extra_attrs.get("version_of", version_of)
+        cloned_segment.version = self.version if is_revision else 1
```
Zaimwa9, line 136: `self.version = (self.version or 1) + 1`
We can do this straight here to deal with version in one place

api/segments/models.py
```diff
@@ -114,35 +114,72 @@ def set_version_of_to_self_if_none(self):  # type: ignore[no-untyped-def]
...
-        cloned_segment.version_of = self
+        cloned_segment.version_of = None  # Unset for now
+        cloned_segment.version = 0  # Unset for now
+        for attr_name, value in extra_attrs.items():
+            setattr(cloned_segment, attr_name, value)
         cloned_segment.save()
 
-        self.version += 1  # type: ignore[operator]
-        self.save_without_historical_record()
+        cloned_segment.copy_rules_and_conditions_from(self)
```
Zaimwa9, line 131: Mmh, can we double-check that we don't have an issue here with the lifecycle hooks?
Especially for revisions

Because we:
1. Save with version 0 
2. Trigger the BEFORE_CREATE hook version = 1
3. Trigger AFTER_CREATE Set version_of to self in the hook => it looks canonical right ?
But save without history so normally no logs
4. We copy the rules with history this time
5. `SegmentRule.get_skip_create_audit_log` will return false: 
`return segment.version_of_id != segment.id  # type: ignore[no-any-return]`
6. Audit logs for rules and conditions ?

I'm not 100% sure but worth having your view

api/segments/models.py
Zaimwa9: See above also

api/segments/models.py
```diff
@@ -114,35 +114,72 @@ def set_version_of_to_self_if_none(self):  # type: ignore[no-untyped-def]
...
+        SegmentRule.objects.filter(segment=self).delete()
+
+        source_rules = SegmentRule.objects.filter(
+            models.Q(segment=source_segment) | models.Q(rule__segment=source_segment)
+        )
+
+        # Ensure top-level rules are cloned first (for dependencies)
+        source_rules = source_rules.order_by(models.F("rule").asc(nulls_first=True))
+
+        rule_to_cloned_rule_map: dict[SegmentRule, SegmentRule] = {}
```
Zaimwa9, line 165: Wouldn't it be more straightforward to use dict[SegmentRule.id, SegmentRule] ?

api/core/workflows_services.py
```diff
@@ -95,26 +97,24 @@ def _publish_change_sets(self, published_by: "FFAdminUser") -> None:
...
-            target_segment.feature = segment.feature
-            target_segment.save()
-
-            # Delete the rules in order to replace them with copies of the segment.
-            target_segment.rules.all().delete()
-            for rule in segment.rules.all():
-                rule.deep_clone(target_segment)
+        for draft_segment in self.change_request.segments.all():
+            live_segment = draft_segment.version_of
+            if not live_segment:  # pragma: no cover
```
Zaimwa9, line 104: why no cover ?

api/tests/unit/segments/test_unit_segments_models.py
```diff
@@ -1,28 +1,44 @@
...
     segment: Segment,
 ) -> None:
     # Given
-    parent_rule = SegmentRule.objects.create(segment=segment, type=SegmentRule.ALL_RULE)
-    child_rule = SegmentRule.objects.create(rule=parent_rule, type=SegmentRule.ALL_RULE)
-    grandchild_rule = SegmentRule.objects.create(
-        rule=child_rule, type=SegmentRule.ALL_RULE
+    rule1 = SegmentRule.objects.create(segment=segment, type=SegmentRule.ALL_RULE)
+    rule2 = SegmentRule.objects.create(rule=rule1, type=SegmentRule.ALL_RULE)
+    rule3 = SegmentRule.objects.create(rule=rule2, type=SegmentRule.ALL_RULE)
```
Zaimwa9: It's interesting because I believe we allow this case while we don't have any way to deep-nest rules above 1 level right?
What would happen ? 
```
        source_rules = SegmentRule.objects.filter(
            models.Q(segment=source_segment) | models.Q(rule__segment=source_segment)
        )
```
I don't think it would pick it up would it ?

Maybe worth adding a test of the expected behavior (be it infinite recursive copy or 2levels) so that we have a source of information later on ?

Zaimwa9 (changes requested):
Good job overall, too bad we didn't go this way earlier.

I have a couple of comments, mostly related to the order of setting version -> copying rules. The lifecycle related one is the most important to double check imo.

api/conftest.py
emyller: True! I guess my real intention here was to leave this comment in hopes that mypy would someday complain about _unnecessary type ignores_ whenever the underlying issue in [django-stubs](https://github.com/typeddjango/django-stubs) (or maybe our own code) was fixed. Other than flagging obvious cases like this one, I do my best to avoid such comments. :)

What do you think?

api/core/workflows_services.py
emyller: Good eyes. Because it would make no sense to have a **draft** segment not pointing to a **live** segment. The API does not allow it, though the data model does, thus this is but a guardrail to favor type checking.

api/segments/models.py
emyller: > 1. Save with version 0
> 2. Trigger the BEFORE_CREATE hook version = 1 => looks canonical ?

Couple things to unpack:
- [This `BEFORE_CREATE` hook](https://github.com/Flagsmith/flagsmith/blob/367edf4877fa1b23d9656078169c9c631ec04d4c/api/segments/models.py#L103-L106) will go away in #5825, to be replaced with adding a `default` clause to `Segment.version` (see [experiment](https://github.com/Flagsmith/flagsmith/pull/5748/commits/fd8e67726a8fdd2062c56f13c8e51e481d4fa28d)).
- The new `clone` method above will start a segment version with `0`, which avoids the above hook. Later then it's set in respect to `is_revision`. See [relevant tests](https://github.com/Flagsmith/flagsmith/blob/33b0a342624c1a6cdb3b4e4899d3d5299eff8ae3/api/tests/unit/segments/test_unit_segments_models.py#L351-L381).

> 3. Trigger AFTER_CREATE Set version_of to self in the hook => it looks canonical right ->    But save without history so normally no logs

Just a note: IMO forcing an `UPDATE` after `INSERT` to make a self reference is bad design — this is what [this hook](https://github.com/Flagsmith/flagsmith/blob/33b0a342624c1a6cdb3b4e4899d3d5299eff8ae3/api/segments/models.py#L108-L115) does. I've been looking at how it works and I'm keen to making this field `null` to indicate that a given segment is the live one instead of a revision — I firmly believe this is better than a self reference.

> 4. We copy the rules with history this time

Sorry I don't understand this one. The new `clone` method does not copy the history of existing rules or conditions; rather, it just does not care about rules and conditions history — we're aiming to dismiss both models `SegmentRule` and `Condition` [sometime soon](https://github.com/Flagsmith/flagsmith/issues/5846).

> 5. `SegmentRule.get_skip_create_audit_log` will return false:>    `return segment.version_of_id != segment.id  # type: ignore[no-any-return]`
> 6. Audit logs for rules and conditions ?

As said, we no longer care about rules and conditions being individual objects. As consequence, this raises a very interesting point: we should no longer have audit logs for changes in rules and conditions of a segment. cc @matthewelwell @khvn26 are we aligned in that?

I think it's a good idea to add/improve tests ensuring the expected behavior of audit logs around segment changes.

> I'm not 100% sure but worth having your view

Excellent observation here, thanks @Zaimwa9.

khvn26: > we should no longer have audit logs for changes in rules and conditions of a segment

Fine by me as long as we emit corresponding audit logs for segments themselves.

api/segments/models.py
emyller: Because the new `clone` method handles both segment revisions and standalone segment clones, setting `version` depends on `is_revision` to determine where to start. Given that earlier in the method `version` is set to zero, your code suggestion might also work, but I'd rather keep the logic explicitly dependent on `is_revision`. Does that make sense and work for you?

api/segments/models.py
emyller: Given both `SegmentRule` and `SegmentRule.id` are hashable objects, I honestly don't see a difference. Note that this maps _original_ segment rules to their _cloned counterparts_ in order to expose parent rules to subrules and conditions further in the function.

api/segments/models.py
Zaimwa9: Yes agreed with Kim, it seems sufficient imo too (and less noisy) to have just one audit log about segment update.
I just wanted to be sure we are not missing any side effects.

As we agree, let's specifically silence them now maybe. 

> As said, we no longer care about rules and conditions being individual objects
I would still care about them until we officially dismiss them 😅

api/conftest.py
Zaimwa9: Mmh, i deep down agree with you but as we used to track those type ignore (and it's in fixtures so not that critical) i'd say let's score some points, get rid of a few ones as we can.
It's true that if we manage to get down to just a few dozens (😄) of targeted ones, it would at least make them relevant

api/tests/unit/segments/test_unit_segments_models.py
emyller: I agree this test case does not represent a real scenario. Right now the segment serializer will only allow nesting rules one level deep, e.g. `{rules: [{rules: [{conditions: ...}, conditions: ...]}`. I've made a small improvement to this test (80712382d). If we have time today I'd love to also add a test to ensure `clone` will ignore/error deeper rules but I can also follow up on #5825 if that works for you. IMO I think we're safe.

api/conftest.py
emyller: Works for me! Fixed in 23aa9675a.

api/segments/models.py
emyller: Alright segment rules and conditions audit logs will be no more as of acb8e76fba24c943c9e99a9c681de2f055da3edd.

api/tests/unit/segments/test_unit_segments_models.py
emyller: This test was actually deleted as part of [this effort](https://github.com/Flagsmith/flagsmith/pull/5898#discussion_r2263550877).

Zaimwa9 (approved):

api/segments/models.py
matthewelwell: Agreed - as long as every edit to a segment emits at least (and seemingly now at most?) one audit log, I'm happy.

emyller: Yup can confirm there is an audit log _"Segment updated"_ when editing a segment.

PR #5902, opened 2025-08-07
fix: Segment updates deleting rules

emyller:


khvn26 (dismissed):

matthewelwell (dismissed):

khvn26 (approved):

matthewelwell (approved):

TimBeyer:
Any chance that this PR broke the PATCH feature?
Docs: https://docs.flagsmith.com/clients/rest#update-a-segments-rules

We saw all our PATCH requests responding with an empty `rules` array starting August 7 at around 17:55 CET, which seems to be roughly when this was released.

This happened silently so we didn't catch it until recently when we got complaints that some features weren't accessible to users.

khvn26:
@TimBeyer Hi, thanks for reporting. Unfortunately we don't have tests that perform the PATCH request against this endpoint so we didn't catch this.

1. We've released a possibly related [fix](https://github.com/Flagsmith/flagsmith/pull/5953) today that could've resolved your situation. Let me know if that changed anything for you.
2. If, still, we reproduce your issue, we'll create a GitHub issue to track the fix. In the meantime, I'd recommend switching to PUT, if possible.

TimBeyer:
Sadly we are still seeing the same issue. 
The PATCH simply drops all the rules we pass to it.

khvn26:
Hello again @TimBeyer – the fix actually has just been released in https://github.com/Flagsmith/flagsmith/releases/tag/v2.192.3. Let me know if you still can reproduce the issue.

TimBeyer:
Looks fixed now. Thank you!

PR #5905, opened 2025-08-07
chore: upgrade node engine version

Zaimwa9:

## Changes
- Upgraded node to version 22 and rebuilt lock file

Main changes between v18 and v22:
- Uses v8 Engine v12.x (versus v10.2) `Performance and language feature boost`
- 🥇  Introduction of the Temporal API: the new Date API we can use => 🎖️  This is huge as long awaited
- `fetch API` stable while previously experimental (became global and no need for polyfills) => No need for `node-fetch`. Optional as we only use it in tests
- ECMAScript introduction: `Promise.withResolvers` and decorators ([doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers))
```
// NEW
async function* readableToAsyncIterable(stream) {
  let { promise, resolve, reject } = Promise.withResolvers();
  stream.on("error", (error) => reject(error));
  stream.on("end", () => resolve());
  stream.on("readable", () => resolve());

  while (stream.readable) {
    await promise;
    let chunk;
    while ((chunk = stream.read())) {
      yield chunk;
    }
    ({ promise, resolve, reject } = Promise.withResolvers());
  }
}

// Equivalent to 
let resolve, reject;
const promise = new Promise((res, rej) => {
  resolve = res;
  reject = rej;
});


// Can also do 
class NotPromise {
  constructor(executor) {
    // The "resolve" and "reject" functions behave nothing like the native
    // promise's, but Promise.withResolvers() just returns them, as is.
    executor(
      (value) => console.log("Resolved", value),
      (reason) => console.log("Rejected", reason),
    );
  }
}

const { promise, resolve, reject } = Promise.withResolvers.call(NotPromise);
resolve("hello");
// Logs: Resolved hello
```
- WebStreams and blob fully stable => we don't use them
- `Buffer()` deprecated => we don't use them
- Some `fs` improvements =>  we don't use them
- `WASI / WASM` stable but not final => we don't use it
- JSON manipulation performance upgrade

- re-installed packages
- re-built app
- ran e2e
- deployed on staging


emyller (dismissed):
Very nice! 🚀

khvn26:
Should we change frontend/.nvmrc as well?

Zaimwa9:
@khvn26 yes good call thanks, updated to lts

khvn26 (dismissed):

khvn26 (approved):

PR #5914, opened 2025-08-11
fix: use-update-feature-permission-for-update-mv

Zaimwa9:

## Changes

_Please describe._

_Please describe._


Zaimwa9:
I will park this PR for now as I realized that the value of a MV is shared accross environment. `UPDATE_FEATURE_STATE` is then not the answer and it makes sense to keep `Project.CREATE_FEATURE` instead while blocking the update in the UI

PR #5924, opened 2025-08-13
fix: Organisation usage charts are not correctly showing environment-document requests

### LINKED ISSUE #5886: Organisation usage charts are not correctly showing environment-document requests
Take the Flagsmith organisation for example. The chart in the sales dashboard, correctly shows that we're using environment document requests for the Flagsmith API project. The chart in the app itself shows no requests to the environment-document endpoint. 

Image

Image

khvn26:

## Changes

Fixes #5886. The problem was in the mapper code.

Added a unit test making sure the correct usage data is returned.

### DIFF
--- a/api/app_analytics/mappers.py
+++ b/api/app_analytics/mappers.py
@@ -91,10 +91,12 @@ def map_flux_tables_to_usage_data(
                     day=date,
                     labels=labels,
                 )
-            if resource := values.get("resource"):
+            if (resource := Resource.get_from_name(values["resource"])) and (
+                resource_attr := resource.column_name
+            ):
                 setattr(
                     data_by_key[key],
-                    resource,
+                    resource_attr,
                     values["_value"],
                 )
     return list(data_by_key.values())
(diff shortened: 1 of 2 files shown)


api/app_analytics/mappers.py
```diff
@@ -91,10 +91,12 @@ def map_flux_tables_to_usage_data(
                     day=date,
                     labels=labels,
                 )
-            if resource := values.get("resource"):
+            if (resource := Resource.get_from_name(values["resource"])) and (
```
Zaimwa9, line 94: Don't we risk to raise an error here if for any reason resource is not set? Is it intentional?

Zaimwa9 (approved):
Lgtm, approving with a comment

api/app_analytics/mappers.py
khvn26: I'd rather see a KeyError surfaced in this case as that would mean we store malformed data.

Issue #5868, opened 2025-07-31
Spike: Investigate integrating with GitHub Actions OIDC flow


emyller:
More on GitHub Actions OIDC: https://docs.github.com/en/actions/concepts/security/openid-connect

Example case (thanks @khvn26): https://depot.dev/docs/container-builds/reference/github-actions#authentication

khvn26:
Consider an ability to exchange both for a master API key and a server-side SDK key in order to take advantage of `api/v1/environment-document`.

Issue #5728, opened 2025-07-08
Frontend work for Github code references PoC


matthewelwell:
## Tasks
- [ ] Add the UI elements to show the code references in a new tab in the feature modal, and include an icon in the feature list alongside segment and identity overrides showing the number of code references

## Mockups

Endpoints are documented in https://github.com/Flagsmith/flagsmith/issues/5730.

### List of features

<img width="2575" height="1053" alt="Image" src="https://github.com/user-attachments/assets/658fc00a-62dd-46c5-b2f9-84abaa4fb68e" />

### Feature detail

<img width="2600" height="1768" alt="Image" src="https://github.com/user-attachments/assets/cacbb49f-e2de-4226-b567-2efa117472dd" />

PR #5928, opened 2025-08-13
feat(github-poc): Submit feature flag code references

emyller:
Contributes to #5730.

This connects to #5731 and allows for CI to push code references found for a project's feature flags.


.github/workflows/poc-github-code-references.yml
```diff
@@ -32,7 +33,7 @@ jobs:
         run: |
           uv run - <<EOF
           # /// script
-          # requires-python = "==${{ env.PYTHON_VERSION }}"
+          # requires-python = ">=${{ env.PYTHON_VERSION }}"
           # dependencies = ["requests==${{ env.PYTHON_REQUESTS_VERSION }}"]
```
khvn26: Sorry, why do we need to parametrise requests version again?

When I proposed uv I hoped to avoid having the `PYTHON_REQUESTS_VERSION` env var entirely?
Same goes for `PYTHON_VERSION` as well.

.github/workflows/poc-github-code-references.yml
```diff
@@ -80,14 +81,22 @@ jobs:
...
-                              if re.search(fr"""(?i:(?:feature|flag)\w*\(\s*(["']){feature_name})\1""", "".join(context)):  # Function calls
+                              re_function_calls = rf"""(?i:(?:feature|flag)\w*\(\s*(["']){re.escape(feature_name)})\1"""
+                              if re.search(re_function_calls, "".join(context)):
                                   yield feature_name, str(path), line_number
                               # TODO: Add more sophisticated matching, e.g. feature names defined as constants
 
+          def retrieve_feature_names() -> list[str]:
+              """Fetch feature names from the Flagsmith API."""
+              response = requests.get(  # TODO: Make better use of pagination
+                  f"${{ env.FLAGSMITH_ADMIN_API_URL }}/api/v1/projects/${{ env.FLAGSMITH_PROJECT_ID }}/features/?page_size=1000",
```
khvn26, line 91: Why not use a server-side SDK key with the `/api/v1/environment-document` endpoint? That would skip the pagination problem entirely. Besides, the SDK endpoint is faster.

api/projects/code_references/models.py
```diff
@@ -0,0 +1,34 @@
...
+        ordering = ["-created_at"]
+        indexes = [
+            GinIndex(  # Helps filtering code references by feature name
+                Func(
+                    models.F("code_references"),
+                    models.Value("$[*].feature_name"),
+                    function="jsonb_path_query_array",
+                ),
+                name="code_references_feature_name",
+            ),
```
khvn26: What are the trade-offs of using a many-to-many through table versus this solution?

I don't imagine using a full text search index as a substitute for foreign key relation scales very well?

It's hard to review this part without seeing the code reference retrieval view, which I assume is coming in a subsequent PR. Maybe the index migration belongs there?

.github/workflows/poc-github-code-references.yml
```diff
@@ -102,26 +111,55 @@ jobs:
           EOF
 
       - name: Display code references
```
khvn26: I still don't understand the reason for decoupling the "retrieve" and "display" steps.

api/projects/code_references/models.py
```diff
@@ -0,0 +1,34 @@
...
+
+    project = models.ForeignKey(
+        "projects.Project",
+        on_delete=models.CASCADE,
+        related_name="code_references",
+    )
+
+    repository_url = models.URLField()  # Provider-agnostic
+    revision = models.CharField(max_length=100)
+    code_references = models.JSONField(default=list)
```
khvn26, line 27: IMO we need to define the explicit type in `projects/code_references/types.py` and annotate this field properly.

```suggestion
    code_references = models.JSONField[CodeReferencesList](default=list)
```

api/projects/code_references/models.py
```diff
@@ -0,0 +1,34 @@
...
+    A JSON package of feature flag code references within a repository
+    """
+
+    project = models.ForeignKey(
+        "projects.Project",
+        on_delete=models.CASCADE,
+        related_name="code_references",
+    )
+
+    repository_url = models.URLField()  # Provider-agnostic
```
khvn26: How is this going to be used? Is it a git/svn URL or an web page url? I'll appreciate more docs here.

api/projects/code_references/serializers.py
```diff
@@ -0,0 +1,39 @@
...
+
+from rest_framework import serializers
+
+from projects.code_references.models import FeatureFlagCodeReferencesScan
+
+
+class _CodeReference(TypedDict):
+    feature_name: str
+    file_path: str
+    line_number: int
```
khvn26, line 11: Yeah, this belongs to a `types.py` module.

khvn26 (changes requested):
I've added a couple of questions and comments.

Overall, I have been struggling to review this series of GitHub PoC PRs, as I feel that I do not have a clear idea of how this tool is going to be used. The lack of user-facing documentation (or at least a draft) and UI mockups is not helping.

emyller:
> Overall, I have been struggling to review this series of GitHub PoC PRs, as I feel that I do not have a clear idea of how this tool is going to be used. The lack of user-facing documentation (or at least a draft) and UI mockups is not helping.

I agree this is lacking intel, I'm sorry. I've added mockups to the [frontend issue](https://github.com/Flagsmith/flagsmith/issues/5728). I hope it helps!

I've also created a draft PR for [retrieving code references](https://github.com/Flagsmith/flagsmith/pull/5931) for a feature flag.

.github/workflows/poc-github-code-references.yml
emyller: For future-proofing? I reckon the requests API is unlikely to change, though. Improved in c4c7245178deb971f60c2a6729e3eaec9a7206f2.

.github/workflows/poc-github-code-references.yml
emyller: Good suggestion, though I chose the admin endpoint for the sake of workflow/action parameterization. We'll need the non-SDK URL and API key for pushing code references, might as well reuse them rather than requiring extra configuration. Let me know what you think.

.github/workflows/poc-github-code-references.yml
khvn26: > For future-proofing?

Sure, why not just specify `requests==2.32.4` in the script header and call it a day? Alternatively, if it's important for us to be DRY and keep the same requests version across all the scripts, maybe have a pyproject.toml with the dependencies specified? My reason for shunning the env var is, I want to avoid having to support custom dependency management here.

Including `requests` without the version specifier also works for me, as we're relying on a well-established API.

api/projects/code_references/models.py
emyller: > What are the trade-offs of using a many-to-many through table versus this solution?
> 
> I don't imagine using a full text search index as a substitute for foreign key relation scales very well?

TBH I keep wondering the same thing myself.

Over the course of the PoC, we'll only trigger a daily scan for a single repository, so database spamming is not an issue now. However, in the future, we plan on plugging code scanning at **pull request level**, which will very likely boost the amount of tuples dramatically.

However, I guess the storage effort/cost of a GIN index would render to about the same amount as if regular database tuples, due to the _inverse_ referencing nature of [GIN indexes](https://www.postgresql.org/docs/current/gin.html).

Both approaches have each pros and cons, let's ~flip a coin~ discuss offline?

> It's hard to review this part without seeing the code reference retrieval view, which I assume is coming in a subsequent PR. Maybe the index migration belongs there?

👍. I've opened https://github.com/Flagsmith/flagsmith/pull/5931 (WIP) based on this branch to hopefully help clarify.

The migration definitely belongs to that context, though I thought a cleaner migration history would favor something. I can't figure out what that _something_ is now. 862e87dc1f6338efbc0526f9f7508159f3710e07.

.github/workflows/poc-github-code-references.yml
emyller: > Alternatively, if it's important for us to be DRY and keep the same requests version across all the scripts, maybe have a pyproject.toml with the dependencies specified?

That might be the way to go **if** we end up depending on other non-standard libraries.

> Including `requests` without the version specifier also works for me, as we're relying on a well-established API.

I like this path given the little use we're making of `requests` here.

.github/workflows/poc-github-code-references.yml
emyller: It had its value before, because it helped visualizing interaction with the JSON output. Now that we're making actual use of it in `Upload code references`, I agree separating collect-display in two jobs is no longer useful, though displaying [_still_ is](https://github.com/Flagsmith/flagsmith/actions/runs/16969851315/job/48103564578) IMO. c5d25931908698e7158e175360f461ab1d384c6f.

api/projects/code_references/serializers.py
emyller: This is improved in the next patch https://github.com/Flagsmith/flagsmith/pull/5931. I'd appreciate not having to backport and avoid Git shenanigans if you agree. 👀

khvn26: Yup, nice 👍

api/projects/code_references/models.py
emyller: Hopefully improved here: 794f3efe2b8bffea060eb6c5ff0611589820b5a2.
I tried thinking of a clearer field name but can't come up with anything not awkward when interacting with the model API.

api/projects/code_references/models.py
emyller: Improved in 1a0d2c711cd705dc97e4150c1c6f0410195f1b3b (https://github.com/Flagsmith/flagsmith/pull/5931).

api/projects/code_references/models.py
khvn26: > I guess the storage effort/cost of a GIN index would render to about the same amount as if regular database tuples

Storage-wise, I imagine a GIN index would be larger as the m2m table wouldn't need to store any text, unless I'm missing something?

Having looked at #5931, I'm inclined to think a m2m through table is the way to go. I'm not convinced it's required at this point in time. I don't anticipate any particular issues with GIN, but it'll be harder to migrate to m2m from GIN-based joins than the other way around if we're challenged by a bad query plan in the future.

.github/workflows/poc-github-code-references.yml
khvn26: UX/DevEx-wise, configuration could definitely be improved, I believe this is why we have https://github.com/Flagsmith/flagsmith/issues/5868 open.

I appreciate it's a PoC and all, and we have a default limit of 400 features per project anyway, so I'm happy to resolve this.

api/projects/code_references/models.py
khvn26: Resolving as this is addressed in #5931.

api/projects/code_references/models.py
khvn26: Resolving as this is addressed in https://github.com/Flagsmith/flagsmith/pull/5931.

api/projects/code_references/models.py
emyller: > Storage-wise, I imagine a GIN index would be larger as the m2m table wouldn't need to store any text, unless I'm missing something?

m2m tables will add PKs, FKs, and btree indexes. The GIN strategy here intends to convert that effort into less database object count, though the performance/storage would be nearly equivalent AFAIK.

> I don't anticipate any particular issues with GIN, but it'll be harder to migrate to m2m from GIN-based joins than the other way around if we're challenged by a bad query plan in the future.

The one thing I'm concerned about is _reinventing the m2m wheel_ with this. Recalling the requirements I don't think I am, though I'm also biased with my own thinking. Querying should be fine as long as the right indexes are there, regardless of the strategy we pick.

api/projects/code_references/models.py
khvn26: Okay, checked gitlab and bitbucket and can confirm this can be used to get code reference permalinks. It's clearer to me now 👍

api/projects/code_references/models.py
khvn26: Looks like since 862e87dc1f6338efbc0526f9f7508159f3710e07 this conversation belongs to #5931, so I'm resolving this.

khvn26 (approved):
LGTM 👍

PR #5930, opened 2025-08-14
feat: release pipelines beta

gagantrivedi:

## Changes

Update Release pipelines module to support the following:
- Add email notification when feature completes a stage
- Add ordering and page_size support to release pipelines API
- Update API to support `wait for trigger` UI
- Add created_at field to the release pipelines model (to support `wait for trigger` and order by)

## How did you test this code

Add unit tests


api/features/release_pipelines/core/migrations/0003_add_created_at.py
```diff
@@ -0,0 +1,35 @@
...
+
+    dependencies = [
+        ("release_pipelines_core", "0002_add_phased_rollout"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="historicalreleasepipeline",
+            name="created_at",
+            field=models.DateTimeField(
```
Zaimwa9, line 18: image

Can't we use this field to avoid an equal hardcoded date here?

Zaimwa9 (changes requested):
Just a comment to address in the migration, otherwise looks good

api/features/release_pipelines/core/migrations/0003_add_created_at.py
gagantrivedi: Since it's a non-nullable column, we'd have to provide a default value for all the existing rows. Did you see the default being used for a new pipeline as well?

Zaimwa9: Then it's a new feature so yes let's move forward with a couple of weird records

Zaimwa9 (dismissed):
👍

Zaimwa9 (approved):

PR #5931, opened 2025-08-14
feat(github-poc): Retrieve feature code references

emyller:


api/projects/code_references/migrations/0002_gotta_go_fast.py
```diff
@@ -0,0 +1,25 @@
...
+            model_name="featureflagcodereferencesscan",
+            index=django.contrib.postgres.indexes.GinIndex(
+                models.Func(
+                    models.F("code_references"),
+                    models.Value("$[*].feature_name"),
+                    function="jsonb_path_query_array",
+                ),
+                name="code_references_feature_name",
+            ),
+        ),
```
emyller: @khvn26 Reviving the discussion: https://github.com/Flagsmith/flagsmith/pull/5928#discussion_r2276235548

This index will favor the queries introduced by this patch, since we now have `code_references` as a `json` column [upstream](https://github.com/Flagsmith/flagsmith/pull/5928).

I'm not opposed to falling back to m2m — as said, each strategy has their pros and cons — though now, I think it makes sense to discuss on a separate issue, if you agree: https://github.com/Flagsmith/flagsmith/issues/5932. We can prioritize that issue within the PoC scope if necessary.

api/projects/code_references/services.py
```diff
@@ -0,0 +1,52 @@
...
+                    JSONObject(feature_name=F("feature_name")),
+                    function="jsonb_path_query_array",
+                ),
+                function="jsonb_array_length",
+            ),
+        )
+        # Only from the latest scans of each repository, within a 30-day sanity window
+        .filter(
+            created_at__gte=timezone.now() - timedelta(days=30),
+            project_id=OuterRef("project_id"),
```
Zaimwa9, line 69: maybe we can add a `count__gt=0`? Not sure it's useful to surface repos with 0 entries

api/projects/code_references/services.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+    counts_by_repository = (
+        FeatureFlagCodeReferencesScan.objects
+        # Count code references from JSON matching the feature name
+        .annotate(
+            feature_name=OuterRef("name"),
+            count=Func(
+                Func(
+                    F("code_references"),
+                    Value("$[*] ? (@.feature_name == $feature_name)"),
```
Zaimwa9, line 59: Minor: I may be wrong, but GIN expression indexes usually require the query to match exactly the index (`$[*].feature_name`) from what I recall. Mind running EXPLAIN ANALYZE to verify?

api/projects/code_references/types.py
```diff
@@ -0,0 +1,40 @@
...
+
+@dataclass
+class FeatureFlagCodeReferences:
+    first_scanned_at: datetime | None
+    last_scanned_at: datetime | None
+    code_references: list[CodeReference]
+
+
+@dataclass
+class CodeReferencesRepositoryCount:
```
Zaimwa9, line 37: reminder for myself. This part still needs to be implemented in FE

Zaimwa9 (commented):
Overall good to me, just one doubt on the index

api/projects/code_references/serializers.py
```diff
@@ -1,26 +1,35 @@
...
+    VCSProvider,
+)
 
 
-class _CodeReference(TypedDict):
-    feature_name: str
-    file_path: str
-    line_number: int
+class _BaseCodeReferenceSerializer(serializers.Serializer[CodeReference]):
+    file_path = serializers.CharField(max_length=200)
```
khvn26: Windows has a path length limit of 256, linux of 4096 — where did 200 come from?

api/tests/unit/projects/code_references/test_unit_projects_code_references_views.py
```diff
@@ -126,3 +127,189 @@ def test_CodeReferenceCreateAPIView__responds_400_when_invalid_data(
...
+    )
+    assert references[2]["scanned_at"] == "2099-01-02T11:00:00Z"
+    assert references[2]["vcs_provider"] == "github"
+    assert references[2]["repository_url"] == "https://github.flagsmith.com/frontend/"
+    assert references[2]["revision"] == "frontend-2"
+    assert references[2]["file_path"] == "frontend/file2.js"
+    assert references[2]["line_number"] == 5
+    assert references[2]["permalink"] == (
+        "https://github.flagsmith.com/frontend/blob/frontend-2/frontend/file2.js#L5"
+    )
```
khvn26: nit: I'd prefer one `assert response.data == expected_data` here. Values we don't care about can be substituted with `mocker.ANY`.

api/projects/code_references/services.py
Zaimwa9: Discussed: repo exists but count = 0 is a hint of deleted featureFlag usage

api/projects/code_references/views.py
```diff
@@ -19,3 +37,96 @@ def perform_create(  # type: ignore[override]
...
+    permission_classes = [ViewFeatureFlagCodeReferences]
+
+    def get_object(self) -> FeatureFlagCodeReferences:
+        return FeatureFlagCodeReferences(
+            first_scanned_at=self._get_first_scanned_at(),
+            last_scanned_at=self._get_last_scanned_at(),
+            code_references=self._get_code_references(),
+        )
+
+    @cache
```
khvn26: I don't think this decorator makes sense here given that a new view instance is created on every request.

khvn26: Ah, I understand now it's to cater for duplicate `_get_related_references` calls in other helper methods.

I wonder now why isn't `_get_related_references` cached, though.

api/projects/code_references/views.py
```diff
@@ -19,3 +37,96 @@ def perform_create(  # type: ignore[override]
...
+    ) -> str:
+        """Generate a permalink for the code reference."""
+        match provider:
+            case VCSProvider.GITHUB:
+                return urljoin(
+                    repository_url, f"blob/{revision}/{file_path}#L{line_number}"
+                )
+        raise NotImplementedError(  # pragma: no cover
+            f"Permalink generation for {provider} is not implemented."
+        )
```
khvn26: I think I'd like all these helper methods to reside as public services in `services.py`.
No matter how simple, it's still business logic, and I prefer to keep it out of views/serializers.

api/projects/code_references/migrations/0002_gotta_go_fast.py
khvn26: > This index will favor the queries introduced by this patch, since we now have code_references as a json column

I'm not sure how that contradicts an m2m through table, but, to keep the pace, I'll defer to your decision here.

api/projects/code_references/migrations/0002_gotta_go_fast.py
```diff
@@ -0,0 +1,25 @@
+# Generated by Django 4.2.22 on 2025-08-14 16:51
```
khvn26: Appreciate the Sonic reference but would prefer a more formal filename here, maybe `0002_add_featureflagcodereferencesscan_gin_index.py`.

khvn26 (changes requested):
LGTM with a couple non-functional comments.

api/projects/code_references/views.py
emyller: Because there is no database query evaluation happening in it.

api/projects/code_references/migrations/0002_gotta_go_fast.py
emyller: Gah I almost did it!

api/projects/code_references/migrations/0002_gotta_go_fast.py
emyller: It would just mean a big change of pace, yes. I meant that modifying the architecture to m2m is out of scope to *this* pull request, and would become a big blocker if we decided so. I've filed https://github.com/Flagsmith/flagsmith/issues/5932 so we don't let it slide easily.

api/projects/code_references/serializers.py
emyller: ~Voices in my head~ my own experience-educated guess really. Thanks for sharing the knowledge; ajusted to Windows' `MAX_PATH` ([docs](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry)) here: 1558f3f56c4b8bd9c715dd4abf35c7895dd777cf.

khvn26: What about Linux? https://github.com/torvalds/linux/blob/ee94b00c1a648530333d9734200be7a45e6e00cd/include/uapi/linux/limits.h#L13

khvn26: Also, would like this to be a constant defined in `constants.py`

api/projects/code_references/services.py
emyller: @Zaimwa9 Thank you for the challenge here. According to this [Postgres documentation](https://www.postgresql.org/docs/current/datatype-json.html#JSON-INDEXING), we **should** get a performance boost in filtering here.

I just did a test with 1M scans, each containing 100+ code references. I've tried several jsonpath queries and the index did not help as much as expected. 😞  Surprisingly, creating the index on the JSON field _entirely_ rather than from an expression **did** help:

```
create index idx_1 on code_references_featureflagcodereferencesscan using gin (code_references);
```

Experimenting made me realize we might actually **not need this index at all**. ✨ The reason is that we never query for all of the historical scans; rather, we only obtain the most recent for each repository, which will most likely translate to a SeqScan in the query planner anyway — since both `project_id` and `created_at` already have their own btree indexes.

I've went ahead and deleted the index. 7bc6166b96ebf760b9c9fc6ffa224ca0bdcd2b83.
cc @khvn26

api/projects/code_references/services.py
emyller: Yup. We will intentionally return scans not matching a feature, as that might indicate that the feature was deleted from the code.

api/tests/unit/projects/code_references/test_unit_projects_code_references_views.py
emyller: I like the above format because it's easier to debug when testing. I'm happy either way. Improved as part of 6244ebff5fe948cfc187eb43d71aa398afc9f69c.

api/projects/code_references/views.py
emyller: Refactored as part of 6244ebff5fe948cfc187eb43d71aa398afc9f69c.

api/projects/code_references/migrations/0002_gotta_go_fast.py
emyller: The migration was deleted. 🤷 
See relevant thread: https://github.com/Flagsmith/flagsmith/pull/5931#discussion_r2279030342

api/projects/code_references/serializers.py
emyller: > What about Linux?

Yeah good point. I've refactored this and expanded the limit as of 38466ffdd8a3224fde68d357bb9e0b613a840af4. I guess Windows users will see issues before they have the chance to integrate with this.

khvn26 (approved):

PR #5934, opened 2025-08-18
fix(cr/audit-log): Add history record to fs went live by audit log

### LINKED ISSUE #5894: Fix: sentry integrations breaks change requests
Steps to reproduce:
Commit or publish a change request in an environment with Sentry integration, and observe that the 'feature went live/updated by change request' log is not generated

Sentry Issue: [FLAGSMITH-API-58Z](https://flagsmith.sentry.io/issues/6769796124/?referrer=github_integration)

```
AuditLog.DoesNotExist: AuditLog matching query does not exist.
  File "django/db/models/query.py", line 916, in get_or_create
    return self.get(**kwargs), False
  File "django/db/models/query.py", line 637, in get
    raise self.model.DoesNotExist(

AttributeError: 'NoneType' object has no attribute 'history_type'
(18 additional frame(s) were not displayed)
...
  File "audit/signals.py", line 106, in signal_wrapper
    return signal_function(sender, instance, **kwargs)
  File "audit/signals.py", line 215, in send_feature_flag_went_live_signal
    feature_state_change_went_live.send(instance)
  File "audit/signals.py", line 233, in send_audit_log_event_to_sentry
    _track_event_async(sender, sentry_change_tracking)  # type: ignore[no-untyped-call]
  File "audit/signals.py", line 124, in _track_event_async
    if event_data := integration_client.generate_event_data(audit_log_record=instance):
  File "integrations/sentry/change_tracking.py", line 52, in generate_event_data
    }[audit_log_record.history_record.history_type]  # type: ignore[union-attr]

Failed to execute task 'tasks.create_feature_state_went_live_audit_log', with id 393921002. Exception: 'NoneType' object has no attribute 'history_type'
```

gagantrivedi:

## Changes
Add history record to audit log generated by `_create_feature_state_audit_log_for_change_request`
This will resolve #5894 

Update unit test

### DIFF
--- a/api/audit/tasks.py
+++ b/api/audit/tasks.py
@@ -36,7 +36,6 @@ def _create_feature_state_audit_log_for_change_request(  # type: ignore[no-untyp
     feature_state_id: int, msg_template: str
 ):
     from features.models import FeatureState
-    from features.signals import feature_state_change_went_live
 
     feature_state = FeatureState.objects.filter(id=feature_state_id).first()
 
@@ -65,7 +64,9 @@ def _create_feature_state_audit_log_for_change_request(  # type: ignore[no-untyp
         feature_state.change_request.title,
     )
     # NOTE: This NEEDS to leverage btree indexes on AuditLog
-    audit_log, log_created = AuditLog.objects.get_or_create(
+    AuditLog.objects.create(
+        history_record_id=feature_state.history.latest().history_id,
+        history_record_class_path=feature_state.history_record_class_path,
         created_date=feature_state.live_from,
         environment=feature_state.environment,
         is_system_event=True,
@@ -74,13 +75,6 @@ def _create_feature_state_audit_log_for_change_request(  # type: ignore[no-untyp
         related_object_id=feature_state.id,
         related_object_type=RelatedObjectType.FEATURE_STATE.name,
     )
-    if log_created:
-        feature_state_change_went_live.send(audit_log)
-    else:
-        logger.info(
-            "FeatureState update audit log already exists. "
-            "Likely the change request was rescheduled to an earlier date.",
-        )
 
 
 @register_task_handler(priority=TaskPriority.HIGHEST)
--- a/api/audit/tasks.py
+++ b/api/audit/tasks.py
@@ -64,7 +64,7 @@ def _create_feature_state_audit_log_for_change_request(  # type: ignore[no-untyp
         feature_state.change_request.title,
     )
     # NOTE: This NEEDS to leverage btree indexes on AuditLog
-    AuditLog.objects.create(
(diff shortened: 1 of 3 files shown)


api/audit/tasks.py
```diff
@@ -74,13 +75,6 @@ def _create_feature_state_audit_log_for_change_request(  # type: ignore[no-untyp
         related_object_id=feature_state.id,
         related_object_type=RelatedObjectType.FEATURE_STATE.name,
     )
-    if log_created:
-        feature_state_change_went_live.send(audit_log)
```
gagantrivedi, line 78: This is already handled by `send_feature_flag_went_live_signal` signal on AuditLog Post save

api/audit/tasks.py
```diff
@@ -74,13 +75,6 @@ def _create_feature_state_audit_log_for_change_request(  # type: ignore[no-untyp
...
-    else:
-        logger.info(
```
gagantrivedi, line 80: I don't see much point in this log message either, but I'm willing to bring it back if the reviewer feels strongly about it.

api/tests/unit/audit/test_unit_audit_tasks.py
```diff
@@ -394,19 +394,17 @@ def test_create_feature_state_went_live_audit_log__rescheduled_feature_update__c
     earlier_schedule = min(scheduled_time, rescheduled_time)
     later_schedule = max(scheduled_time, rescheduled_time)
 
-    with freeze_time(now):
-        # Given
-        change_request_feature_state.live_from = scheduled_time
-        change_request_feature_state.save()
-        change_request.committed_at = timezone.now()
-        change_request.save()  # Triggers hook
```
gagantrivedi, line 402: Am I missing something here? This code block makes it seem like it will trigger a task, but it won’t since we are not running the task processor.

api/audit/tasks.py
khvn26: Does this mean we're able to delete `audit.signals.send_feature_flag_went_live_signal`  as well?

khvn26 (commented):
One question about whether we can delete more code!

api/tests/unit/audit/test_unit_audit_tasks.py
khvn26: Perhaps a signal was implied?

api/audit/tasks.py
gagantrivedi: Nope, we were able to remove the above code because of `audit.signals.send_feature_flag_went_live_signal`.

khvn26: Ah great, the puzzle clicked in my head now, thanks.

khvn26 (approved):

PR #5941, opened 2025-08-20
feat: add environment variable to protect API docs with authentication

matthewelwell:
## Changes

Adds a new environment variable (`REQUIRE_AUTHENTICATION_FOR_API_DOCS`) which, when set to True, forces users to authenticate using Basic Authentication in order to access API docs. 

Ran the API locally with `REQUIRE_AUTHENTICATION_FOR_API_DOCS: True` and verified that I couldn't access any of the following routes without authentication.

```
/api/v1/docs/
/api/v1/swagger.json
/api/v1/swagger.yaml
```

By authenticating with the credentials from any user in the system, I am able to access the API docs.

Also ran the API locally with `REQUIRE_AUTHENTICATION_FOR_API_DOCS: False` to verify that the behaviour hasn't changed.


docs/docs/deployment/hosting/locally-api.md
```diff
@@ -230,6 +230,7 @@ the below variables will be ignored.
 - `API_USAGE_CACHE_SECONDS`: Controls how frequently the usage cache is flushed. Defaults to 60 seconds
 - `PROMETHEUS_ENABLED`: Enables the Prometheus `/metrics` endpoint. Default is False.
 - `PROMETHEUS_HISTOGRAM_BUCKETS`: Allows to specify your bucket sizes for Prometheus histograms, e.g., `"0.5,0.6,1.0,Inf"`. Defaults to Python Prometheus client default histogram sizes.
+- `REQUIRE_AUTHENTICATION_FOR_API_DOCS`: If set to True, users must be authenticated with django admin to access API docs.
```
gagantrivedi: App login will not work here?

matthewelwell: sorry, forgot to update the docs after changing to basic authentication - thanks for the catch

gagantrivedi (approved):

PR #5942, opened 2025-08-20
fix(env-update): make is_creating read only

gagantrivedi:

## Changes
Make `is_creating` a read-only field because a user should not be allowed to update it.

Add unit test

### DIFF
--- a/api/environments/serializers.py
+++ b/api/environments/serializers.py
@@ -58,7 +58,10 @@ class EnvironmentSerializerLight(serializers.ModelSerializer):  # type: ignore[t
             "use_identity_overrides_in_local_eval",
             "is_creating",
         )
-        read_only_fields = ("use_v2_feature_versioning",)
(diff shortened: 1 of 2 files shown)


api/environments/serializers.py
```diff
@@ -104,7 +107,6 @@ class Meta(EnvironmentSerializerWithMetadata.Meta):
         fields = EnvironmentSerializerWithMetadata.Meta.fields + (  # type: ignore[has-type]
             "total_segment_overrides",
         )
-        read_only_fields = ("total_segment_overrides",)
```
gagantrivedi, line 107: This ended up in this PR because typing had issues with that line. I removed it since it’s already in the retrieve serializer, where everything is read-only anyway. I can bring it back if you disagree.

Zaimwa9 (approved):
Thanks 👍

PR #5945, opened 2025-08-20
fix: correct implementation for email validation on login

matthewelwell:
## Changes

This is a re-implementation of #5848 , but I need to verify with @Zaimwa9 what the reason for the original implementation is at it seems unnecessary to me. 

Updated / added unit tests.

### DIFF
--- a/api/custom_auth/serializers.py
+++ b/api/custom_auth/serializers.py
@@ -2,7 +2,12 @@ from typing import Any
 
 from common.core.utils import is_saas
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


api/custom_auth/serializers.py
```diff
@@ -20,6 +25,16 @@
 )
 
 
+class CustomTokenCreateSerializer(TokenCreateSerializer):  # type: ignore[misc]
```
Zaimwa9, line 28: I think it just needs to be added to the view to pass
`serializer_class = CustomTokenCreateSerializer`

Thanks for jumping on it

PR #5946, opened 2025-08-20
infra: split out rules and conditions ordering vars and enable conditions ordering in production

matthewelwell:
## Changes

Due to a regression where conditions within a given rule have started suffering from the same issue as rules with regards to ordering not being based on the id as is implicitly expected in the code, this PR splits out the rules and conditions ordering settings so that we can maintain the current behaviour for rules while resolving it for conditions. 

N/a - cannot be tested on local dataset, needs testing in production.


khvn26 (dismissed):

api/app/settings/common.py
```diff
@@ -1368,10 +1368,25 @@
 
 SEGMENT_RULES_CONDITIONS_LIMIT = env.int("SEGMENT_RULES_CONDITIONS_LIMIT", 100)
 
+# These settings are to handle large datasets / odd behaviour where rules and conditions
+# often aren't returned in the order that they were created in, which the code implicitly
+# expects.
 SEGMENT_RULES_CONDITIONS_EXPLICIT_ORDERING_ENABLED = env.bool(
```
khvn26, line 1374: This setting doesn't seem to do anything now?

Maybe we could implement compatibility with it via defaults or somehow else?

matthewelwell: It does, see below

matthewelwell: I probably should just use it as a default though, you're right

matthewelwell: Done

khvn26 (approved):

PR #5953, opened 2025-08-21
fix(segment-update): handle rules/conditons without id correctly

gagantrivedi:

## Changes

Overwrite the `__init__` method to handle deletion of rules/conditions correctly. The earlier implementation was overriding get_initial, but since `WritableNestedModelSerializer` uses that internally, it resulted in some strange behavior where rules/conditions that should have been created or updated were instead removed.

## How did you test this code

Adds unit tests

### DIFF
--- a/api/segments/serializers.py
+++ b/api/segments/serializers.py
@@ -83,6 +83,11 @@ class SegmentSerializer(MetadataSerializerMixin, WritableNestedModelSerializer):
     metadata = MetadataSerializer(required=False, many=True)
 
     def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """
+        Because WritableNestedModelSerializer uses `initial_data` instead of `data`
+        we need to override the `__init__` method to remove rules and conditions
+        that are marked for deletion.
+        """
         data = kwargs.get("data")
         if data:
             data["rules"] = self._get_rules_and_conditions_without_deleted(
--- a/api/segments/serializers.py
+++ b/api/segments/serializers.py
@@ -82,6 +82,15 @@ class SegmentSerializer(MetadataSerializerMixin, WritableNestedModelSerializer):
     rules = SegmentRuleSerializer(many=True, required=True, allow_empty=False)
     metadata = MetadataSerializer(required=False, many=True)
 
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        data = kwargs.get("data")
+        if data:
+            data["rules"] = self._get_rules_and_conditions_without_deleted(
+                data["rules"]
+            )
+
+        super().__init__(*args, **kwargs)
+
     class Meta:
         model = Segment
         fields = [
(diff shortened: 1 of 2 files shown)


api/tests/unit/segments/test_unit_segments_views.py
```diff
@@ -1041,21 +1154,20 @@ def test_update_segment_delete_existing_rule(project, client, segment, segment_r
                         "type": nested_rule.type,
                         "rules": [],
                         "conditions": [],
+                        "delete": True,
                     }
                 ],
                 "conditions": [],
-                "delete": True,
```
gagantrivedi, line 1047: This is something we need to discuss. The old implementation of sending `delete: true` allowed removing the top rule, but the new implementation of simply omitting them does not. I’ve updated this to align with the new approach of not allowing root-level rule deletion, since I don’t see a use case for it and the front end never deletes the root rule. I’m open to discussing changing the behavior back if we deem it necessary

matthewelwell: Seems logical to me.

api/segments/serializers.py
```diff
@@ -82,6 +82,20 @@ class SegmentSerializer(MetadataSerializerMixin, WritableNestedModelSerializer):
...
     metadata = MetadataSerializer(required=False, many=True)
 
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """
+        Because WritableNestedModelSerializer uses `initial_data` instead of `data`
+        we need to override the `__init__` method to remove rules and conditions
+        that are marked for deletion.
+        """
+        data = kwargs.get("data")
+        if data:
```
Zaimwa9:
```suggestion
        if data and "rules" in data:
```

api/segments/serializers.py
```diff
@@ -82,6 +82,20 @@ class SegmentSerializer(MetadataSerializerMixin, WritableNestedModelSerializer):
     rules = SegmentRuleSerializer(many=True, required=True, allow_empty=False)
     metadata = MetadataSerializer(required=False, many=True)
 
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """
+        Because WritableNestedModelSerializer uses `initial_data` instead of `data`
+        we need to override the `__init__` method to remove rules and conditions
+        that are marked for deletion.
+        """
+        data = kwargs.get("data")
```
Zaimwa9, line 91: As it's been a complicated case, to avoid any side-effect, should we deepcopy, update, and then re-assign to kwargs ?

api/tests/unit/segments/test_unit_segments_views.py
```diff
@@ -714,6 +714,120 @@ def test_create_segments_with_description_condition(project, client):  # type: i
     assert segment_condition_description_value == "test-description"
 
 
+def test_update_segment_add_new_root_rule(
```
Zaimwa9, line 717: I could see an extra test with 2 nested conditions => One to be updated and one to be deleted

Zaimwa9 (changes requested):
👏 Sorry, couple of comments. Being a bit annoying just to close the topic once and for all

api/segments/serializers.py
gagantrivedi: I don’t think we need a deepcopy here. The serializer itself only ever uses the data passed in through kwargs, and mutating it before calling super().__init__ is intentional. A deepcopy would add unnecessary overhead without any real benefit.

api/tests/unit/segments/test_unit_segments_views.py
gagantrivedi: I have added a new test named: `test_update_segment_delete_and_update_existing_condition` Let me know if that makes sense

Zaimwa9 (approved):
👍

PR #5973, opened 2025-08-25
chore(evaluation-context): Support `string[]` as condition value for the `IN` operator

### LINKED ISSUE #5935: Evaluation Context: Support `string[]` as condition value for the `IN` operator
We need to be able to provide native `list[str]` data as the `IN` operator condition value for the following reasons:
1. Identity overrides are being mapped to "virtual segments" now and we want to avoid double serialisation.
2. Prepare for #2771 and #5634 that might prompt us to explore alternative means of encoding and storing large lists. 

To support the above, introduce a new subschema:

```json
{
   "property": {"type": "string"},
   "operator": {"const": "IN"},
   "value": {"type": "array", "items": {"type": "string"}}
}
```

Extend `Condition` to a polymorphic schema including the above, via `anyOf`.

khvn26:
Closes #5935.

## Changes

This adds `string[]` support just for the `IN` operator to the engine evaluation context schema.

Used jsonschema.dev to validate the following test data:

```json
{
  "environment": {
    "key": "",
    "name": ""
  },
  "segments": {
    "1": {
      "key": "",
      "name": "",
      "rules": [
        {
          "type": "ALL",
          "conditions": [
            {
              "property": "foo",
              "operator": "IN",
              "value": "my1,my2"
            },
             {
              "property": "foo",
              "operator": "IN",
              "value": ["my1", "my2"]
            }
          ]
        }
      ]
    }
  }
}
```

### DIFF
--- a/sdk/evaluation-context.json
+++ b/sdk/evaluation-context.json
@@ -171,47 +171,74 @@
             "type": "object"
         },
         "SegmentCondition": {
-            "description": "Represents a condition within a segment rule for feature flag evaluation.",
-            "properties": {
-                "property": {
-                    "description": "A reference to the identity trait or value in the evaluation context.",
-                    "title": "Property",
-                    "type": "string"
+            "anyOf": [
+                {
+                    "description": "Represents a condition within a segment rule for feature flag evaluation.",
+                    "properties": {
+                        "property": {
+                            "description": "A reference to the identity trait or value in the evaluation context.",
+                            "title": "Property",
+                            "type": "string"
+                        },
+                        "operator": {
+                            "description": "The operator to use for evaluating the condition.",
+                            "title": "Operator",
+                            "type": "string",
+                            "enum": [
+                                "EQUAL",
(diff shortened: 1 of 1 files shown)


sdk/evaluation-context.json
```diff
@@ -171,47 +171,81 @@
...
+                                "CONTAINS",
+                                "GREATER_THAN_INCLUSIVE",
+                                "NOT_CONTAINS",
+                                "NOT_EQUAL",
+                                "REGEX",
+                                "PERCENTAGE_SPLIT",
+                                "MODULO",
+                                "IS_SET",
+                                "IS_NOT_SET",
+                                "IN"
```
emyller, line 201: @khvn26 Do we plan on deprecating this usage?

emyller (approved):

sdk/evaluation-context.json
khvn26: No, the engines will be fully backwards-compatible.

Issue #5958, opened 2025-08-21
Github PoC: investigate why cron task to update code references is failing


matthewelwell:
Action: https://github.com/Flagsmith/flagsmith/actions/workflows/poc-github-code-references.yml

khvn26:
```
Traceback (most recent call last):
  File "<string>", line 22, in <module>
    response.raise_for_status()
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/runner/.cache/uv/environments-v2/a447a2b7f7084e3a/lib/python3.13/site-packages/requests/models.py", line 1026, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: https://api.flagsmith.com/api/v1/projects/12/code-references/
```

emyller:
<img width="835" height="534" alt="Image" src="https://github.com/user-attachments/assets/9aec999b-2813-4866-97ba-68128d954131" />

Works as per [v2.192.4](https://github.com/Flagsmith/flagsmith/releases/tag/v2.192.4).

Issue #5978, opened 2025-08-25
Spike: Improve user types


emyller:
Investigate the effort to improve typing around our user models, with the ultimate goal of improving existing inconsistencies.

As of now we have:
- `FFAdminUser` ([ref](https://github.com/Flagsmith/flagsmith/blob/18f3f09027b7acdb99751342881152ad53feb5b2/api/users/models.py#L108)): implements Django's AbstractUser.
- `APIKeyUser` ([ref](https://github.com/Flagsmith/flagsmith/blob/f7487ac5278d60e62067e2f4d5976ceb755f4980/api/api_keys/user.py#L22)): implements `UserABC`.
- `UserABC` ([ref](https://github.com/Flagsmith/flagsmith/blob/f7487ac5278d60e62067e2f4d5976ceb755f4980/api/users/abc.py#L10)): why does it exist?

Maybe get rid of `UserABC`, or if otherwise proven necessary, make it our go-to base user class instead of relying on a [subclass hook](https://github.com/Flagsmith/flagsmith/blob/f7487ac5278d60e62067e2f4d5976ceb755f4980/api/users/abc.py#L66-L68).

An example of issue this causes is: likely because of `settings.DJANGO_USER_MODEL`, DRF will type `request.user` as `FFAdminUser | AnonymousUser`, which is incompatible with the actual `APIKeyUser` type and misleading.

This will also probably fix https://github.com/flagsmith/flagsmith/issues/4379.

matthewelwell:
Related: https://github.com/Flagsmith/flagsmith/issues/5146

emyller:
# Spike Report

## Background

The Admin API (`/api/v1/organisations/`, `/api/v1/projects/`, `/api/v1/environments/`, `/api/v1/features/`) can be accessed by humans (via browser/JWT or `Authorization: Token ...`) or by machines (via `Authorization: Api-Key ...`). When a human authenticates, `request.user` is an `FFAdminUser`. When a machine authenticates with an API key, `request.user` is an `APIKeyUser` wrapping a `MasterAPIKey`. `MasterAPIKey.is_admin` controls whether the key has full access (`True`) or is RBAC-restricted (`False`).

The Flags API (`/api/v1/flags/`, `/api/v1/identities/`, `/api/v1/traits/`) only accepts Environment Keys and is not affected in this project.

| Scope | Key Type | Model | Purpose |
|-------|----------|-------|---------|
[...]

PR #5979, opened 2025-08-25
fix(GitHub PoC): Fix API authentication

emyller:
Contributes to #5958.

### DIFF
--- a/api/projects/code_references/permissions.py
+++ b/api/projects/code_references/permissions.py
@@ -1,9 +1,9 @@
 from common.projects.permissions import VIEW_PROJECT
+from django.contrib.auth.models import AnonymousUser
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.request import Request
 from rest_framework.views import APIView
 
-from api_keys.user import APIKeyUser
 from projects.models import Project
 
 
@@ -11,10 +11,10 @@ class _BaseCodeReferencePermission(IsAuthenticated):
     def has_permission(self, request: Request, view: APIView) -> bool:
         if not super().has_permission(request, view):
             return False
+        assert not isinstance(request.user, AnonymousUser)
 
-        user: APIKeyUser = request.user  # type: ignore[assignment]
         project = Project.objects.get(id=view.kwargs["project_pk"])
-        return user.has_project_permission(VIEW_PROJECT, project)
+        return request.user.has_project_permission(VIEW_PROJECT, project)
 
 
 class SubmitFeatureFlagCodeReferences(_BaseCodeReferencePermission):
--- a/api/projects/code_references/permissions.py
+++ b/api/projects/code_references/permissions.py
@@ -3,8 +3,8 @@ from rest_framework.permissions import IsAuthenticated
 from rest_framework.request import Request
 from rest_framework.views import APIView
 
+from api_keys.user import APIKeyUser
 from projects.models import Project
-from users.models import FFAdminUser
 
 
 class _BaseCodeReferencePermission(IsAuthenticated):
@@ -12,11 +12,9 @@ class _BaseCodeReferencePermission(IsAuthenticated):
         if not super().has_permission(request, view):
             return False
 
-        if not isinstance(request.user, FFAdminUser):  # pragma: no cover
-            return False
-
+        user: APIKeyUser = request.user  # type: ignore[assignment]
         project = Project.objects.get(id=view.kwargs["project_pk"])
-        return request.user.has_project_permission(VIEW_PROJECT, project)
+        return user.has_project_permission(VIEW_PROJECT, project)
 
 
 class SubmitFeatureFlagCodeReferences(_BaseCodeReferencePermission):
--- a/api/tests/unit/projects/code_references/test_unit_projects_code_references_views.py
+++ b/api/tests/unit/projects/code_references/test_unit_projects_code_references_views.py
@@ -64,6 +64,58 @@ def test_CodeReferenceCreateAPIView__responds_201_with_accepted_code_references(
     ]
 
 
+def test_CodeReferenceCreateView__responds_401_if_not_authenticated(
+    client: APIClient,
+    project: Project,
+) -> None:
+    # When
+    response = client.post(
+        f"/api/v1/projects/{project.pk}/code-references/",
+        data={
+            "repository_url": "https://svn.flagsmith.com/",
+            "revision": "revision-hash",
+            "code_references": [
+                {
+                    "feature_name": "feature-1",
+                    "file_path": "path/to/file1.py",
+                    "line_number": 10,
+                },
+            ],
+        },
+        format="json",
+    )
+
+    # Then
+    assert response.status_code == 401
+    assert not FeatureFlagCodeReferencesScan.objects.exists()
+
+
+def test_CodeReferenceCreateAPIView__responds_403_if_incorrect_permissions(
+    project: Project,
+    staff_client: APIClient,
+) -> None:
+    # When
+    response = staff_client.post(
+        f"/api/v1/projects/{project.pk}/code-references/",
+        data={
+            "repository_url": "https://svn.flagsmith.com/",
+            "revision": "revision-hash",
+            "code_references": [
+                {
+                    "feature_name": "feature-1",
+                    "file_path": "path/to/file1.py",
+                    "line_number": 10,
+                },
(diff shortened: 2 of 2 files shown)


api/projects/code_references/permissions.py
```diff
@@ -3,20 +3,18 @@
...
 
 class _BaseCodeReferencePermission(IsAuthenticated):
     def has_permission(self, request: Request, view: APIView) -> bool:
         if not super().has_permission(request, view):
             return False
 
-        if not isinstance(request.user, FFAdminUser):  # pragma: no cover
-            return False
-
+        user: APIKeyUser = request.user  # type: ignore[assignment]
```
matthewelwell: There's nothing stopping someone from using a user token to call this endpoint, right? I'm not sure why we'd assume that the request is from an API key here? 

I guess for the PoC it's fine, but it's adding more tech debt for later, are there any other solutions?

khvn26: Shall we tweak this annotation to allow adding references manually via UI?

```suggestion
        user: APIKeyUser | FFAdminUser = request.user  # type: ignore[assignment]
```

khvn26 (changes requested):
Can we add a test reproducing the issue we're fixing?

api/projects/code_references/permissions.py
Zaimwa9: For my own knowledge, would`is_master_api_key_user` be the way to identify an org level key ?

Zaimwa9:
```
if getattr(request.user, "is_master_api_key_user", False) is False:
```

khvn26:
Contributes to #5958.

api/projects/code_references/permissions.py
emyller: > There's nothing stopping someone from using a user token to call this endpoint, right? I'm not sure why we'd assume that the request is from an API key here?

@matthewelwell Correct. I just did not have the right understanding of user typing. Should be improved with 23361310190ed44e3b827ccec6307fc2f9ad389c.

matthewelwell (dismissed):

api/tests/unit/projects/code_references/test_unit_projects_code_references_views.py
```diff
@@ -1,24 +1,27 @@
...
-from tests.types import WithProjectPermissionsCallable
 
 
+@pytest.mark.parametrize(
+    "client",
+    [
+        lazy_fixture("admin_master_api_key_client"),
+        lazy_fixture("admin_client"),
+    ],
+)
```
khvn26: In fact, we have this logic baked in the `admin_client_new` fixture — it's parametrised already.

khvn26: Sorry for not directing your attention to it earlier.

api/projects/code_references/permissions.py
emyller: Thanks for the suggestion! I decided to choose the following approach because I'd rather not have this code concerned about the distinct user types we maintain.

https://github.com/Flagsmith/flagsmith/blob/23361310190ed44e3b827ccec6307fc2f9ad389c/api/projects/code_references/permissions.py#L14

I'd prefer to rely on a base User type but we don't have one yet. Related: https://github.com/Flagsmith/flagsmith/issues/5978

Let me know if you agree.

emyller: > ```
> if getattr(request.user, "is_master_api_key_user", False) is False:
> ```

@Zaimwa9 Sorry is this a suggestion for this patch?

khvn26: Yup, I agree.

api/tests/unit/projects/code_references/test_unit_projects_code_references_views.py
emyller: Nice! 99ddb838a40eeb6f5d7a4dc3f96ca51425c84800

khvn26 (approved):

api/tests/unit/projects/code_references/test_unit_projects_code_references_views.py
matthewelwell: Sorry, I should have also picked this up in my review, but this is not the right solution at all I'm afraid. 

These tests are all designed to use the `staff_client` fixture which, by default, has no permissions. The `admin_client`, `admin_master_api_key` and `admin_client_new` fixtures all implicitly have organisation admin permissions. As such, by moving from `staff_client` -> `admin_*` we've changed the behaviour of these tests and they aren't testing what we expect them to. 

We'll likely need to create a new `staff_master_api_key_client` but that's probably non-trivial...

khvn26 (changes requested):
Removing prior approval after @matthewelwell's comment

api/tests/unit/projects/code_references/test_unit_projects_code_references_views.py
emyller: @matthewelwell I would argue the current tests do expect authenticated clients in order to respond with 200/201/400. For now I've added explicit tests covering 401/403 in 015bb1e55706eeaad092be015cf250dc47eafc5f — let me know if that can resolve this thread.

khvn26 (approved):

api/tests/unit/projects/code_references/test_unit_projects_code_references_views.py
```diff
@@ -1,24 +1,18 @@
...
     project: Project,
-    staff_client: APIClient,
-    with_project_permissions: WithProjectPermissionsCallable,
 ) -> None:
-    # Given
-    with_project_permissions([VIEW_PROJECT])  # type: ignore[call-arg]
-
     # When
-    response = staff_client.post(
+    response = admin_client_new.post(
```
matthewelwell, line 15: This test used to test the minimal permissions needed for a 201 - now it's only testing that it works for an admin user. Perhaps later tests are covering some of this instead, but I don't love this modification.

matthewelwell: It looks like the rest of the tests do cover the negative cases, so maybe it's fine.

matthewelwell (approved):
Approved with a minor comment.

api/tests/unit/projects/code_references/test_unit_projects_code_references_views.py
emyller: > We'll likely need to create a new `staff_master_api_key_client` but that's probably non-trivial...

I confirm it's non-trivial.

I went ahead and implemented the client and relevant fixtures replicating the same behavior as `admin_client_new`, and was ultimately blocked because assigning permissions to an `APIKeyUser` crosses the [flagsmith-rbac bridge](https://github.com/Flagsmith/flagsmith-rbac/blob/fcf616409799222dcce8e906d7e3f110d5d576e9/rbac/models.py#L144-L152).

The best-effort to match reality would be to cover a staff user client, and an admin API key client, though I'm not sure how much of an improvement that is to this PoC. However, the following works if added to this test module if we want:

```python
@pytest.fixture
def authorized_staff_client(
    staff_client: APIClient,
    with_project_permissions: WithProjectPermissionsCallable,
) -> APIClient:
    with_project_permissions([VIEW_PROJECT])  # type: ignore[call-arg]
    return staff_client


@pytest.fixture(params=["admin_master_api_key_client", "authorized_staff_client"])
def authenticated_client(request: pytest.FixtureRequest) -> APIClient:
    client: APIClient = request.getfixturevalue(request.param)
    return client

def test_CodeReferenceCreateAPIView__responds_400_when_missing_field(
    authenticated_client: APIClient,  # Will run twice as defined above
    ...
```

I'd rather discard my attempt for now, and later revisit this. Also I'm not sure whether `VIEW_PROJECT` is the right permission to check for the use cases within this feature.

api/tests/unit/projects/code_references/test_unit_projects_code_references_views.py
emyller: > Perhaps later tests are covering some of this instead, but I don't love this modification.

I did an experiment ([described above](https://github.com/Flagsmith/flagsmith/pull/5979#discussion_r2299335658)) that may help, though I'm not certain. Since this only affects this PoC, I'll merge this as-is and hand it over to the frontend team in their timezone. Let's please revisit this.

PR #5987, opened 2025-08-26
feat: always show feature health tab

### LINKED ISSUE #5840: Always show Feature Health tab in Feature modal
Currently we only show the 'Feature Health' tab in the Feature modal if there is an active feature health alert. I can see a few benefits to showing the tab regardless of whether there are any active alerts or not: 

 1. Better visibility of the feature - currently it's hidden in the project settings
     1. For example, if they don't have any providers set up, can we link them to project settings (if they have permission) to set one up? 
 2. In the future we might want to add historical alerts there
 3. When resolving the item here: #5515 it might be odd behaviour for the tab to disappear when dismissing the last alert

Zaimwa9:

Closes #5840 

## Changes
Objective of this PR is to always show the Feature Health Tab in the Edit Flag modal.
This required integrating an overflowing menu in the modal:
- Extracted Tabs and Tab components
- Integrated `useOverflowVisibleCount` hook
- Re-organized the order of the tabs

https://www.loom.com/share/629d604bfd174a35a9fa5215ba21d88c

image
image

**Create flag**
image


frontend/web/components/base/DropdownMenu.tsx
```diff
@@ -83,6 +88,9 @@ const DropdownMenu: React.FC<DropdownMenuProps> = ({
                         setIsOpen(false)
                       }
                     }}
+                    style={{
+                      zIndex: 999999,
```
matthewelwell, line 92: I suspect this is necessary here, but my understanding is that adding zIndex like this can be considered a hack. Should we be worried about it here?

frontend/web/components/base/forms/Button.tsx
```diff
@@ -37,90 +38,106 @@ export type ButtonType = ButtonHTMLAttributes<HTMLButtonElement> & {
...
+  ) => {
+    const hasPlan = feature ? Utils.getPlansPermission(feature) : true
+    return href || !hasPlan ? (
+      <a
+        onClick={
+          !hasPlan ? undefined : (rest.onClick as React.MouseEventHandler)
+        }
+        className={cn(className, themeClassNames[theme], sizeClassNames[size])}
+        target={!hasPlan ? '_blank' : target}
+        href={hasPlan ? href : Constants.getUpgradeUrl()}
```
matthewelwell, line 73: Is there a bug here? As far as I can tell, if `href` is truthy, then we blindly return it above before hitting this logic. So I think this line actually boils down to something like the following, which feels wrong? 

```typescript
href={hasPlan ? undefined : Constants.getUpgradeUrl()}
```

matthewelwell: Ah, ok, I think I assumed a different order to the one in which js is evaluating the logic, exemplified by this: 

image

So, I think actually, we know that `href` will be truthy if `hasPlan` is falsy 👍 

(sorry for the stream of consciousness)

frontend/web/components/modals/CreateFlag.js
```diff
@@ -1918,30 +1899,24 @@ const CreateFlag = class extends Component {
...
-                                            icon={warning}
-                                            style={{
-                                              color:
-                                                Constants.featureHealth
-                                                  .unhealthyColor,
-                                              marginBottom: -2,
-                                            }}
-                                          />
+                                        <Row className='d-flex justify-content-center align-items-center pr-1 gap-1'>
+                                          Feature Health
```
matthewelwell: I appreciate that this isn't necessarily part of the scope of this work, but this should have the 'Beta' badge next to it, like it does in the project settings menu.

frontend/web/components/navigation/OverflowNav.tsx
```diff
@@ -5,7 +5,7 @@ import { ellipsisHorizontal } from 'ionicons/icons'
 import Button from 'components/base/forms/Button'
 import classNames from 'classnames'
 import { useHistory } from 'react-router-dom'
-import { useVisibleCount } from 'common/hooks/useVisibleCount' // Assuming you place the hook here
+import { useOverflowVisibleCount } from 'common/hooks/useOverflowVisibleCount' // Assuming you place the hook here
```
matthewelwell: I know the comment pre-exists this change, but I'm not sure what this comment is adding here? I have no idea what it means tbh.

frontend/web/components/navigation/TabMenu/Tabs.tsx
```diff
@@ -0,0 +1,234 @@
...
+  hideNavOnSingleTab?: boolean
+  buttonTheme?: string
+  noFocus?: boolean
+  isRoles?: boolean
+  history?: any
+  overflowX?: boolean
+}
+
+const MORE_BTN_SIZE = 34
+const EXTRA_RESERVE = MORE_BTN_SIZE + 20
```
matthewelwell: Why not just:

```suggestion
const EXTRA_RESERVE = 54
```

Also, I appreciate I have no idea what I'm doing when it comes to UI, but I do always find it a little unnerving when we have hard-coded values like this. Could you explain a little about what we're doing here?

frontend/web/styles/project/_FeaturesPage.scss
```diff
@@ -7,7 +7,7 @@
 
   &__list {
     position: absolute;
-    z-index: 10;
+    z-index: 2000000000001;
```
matthewelwell, line 10: 😱

frontend/web/styles/project/_modals.scss
```diff
@@ -297,8 +305,8 @@ $side-width: 750px;
   .tabs-nav button.btn,
   .tabs-nav button.btn-primary {
     font-size: $font-sm;
-    line-height: $line-height-sm;
-    padding: 0 16px;
+    // line-height: $line-height-sm;
```
matthewelwell: Can we remove this rather than leaving it commented?

frontend/web/styles/project/_utils.scss
```diff
@@ -234,7 +234,7 @@
 }
 
 .bg-primary-opacity-5 {
-    background-color: transparentize($primary, 0.95);
+    background-color: transparentize($primary, 0.95) !important;
```
matthewelwell: What's the need for using `!important` here?

frontend/web/components/base/forms/Button.tsx
Zaimwa9: No, there is a risk there indeed. It's assuming upstream that href is passed anytime `hasPlan` is true.
I found Buttons with href in onboardingPage -> Safe there

And in `PlanBasedAccess.tsx`.
Safe but can be the wrong url in the worst case. It can be improved for sure but I think it's out of this scope (lot of possible regressions)
```
const planUrl = Constants.getUpgradeUrl(feature)
...
...

 {Utils.isSaas() ? (
        <a
          onClick={trackFeature}
          target='_blank'
          href={Constants.getUpgradeUrl(feature)}
          rel='noreferrer'
        >
          <Button theme='tertiary' size='xSmall'>
            Start Free Trial
          </Button>
        </a>
      ) : (
        <Button
          onClick={trackFeature}
          theme='tertiary'
          size='xSmall'
          href={planUrl}
          target='_blank'
        >
          Contact Us
        </Button>
      )}
```


But there will always be a fallback:
```
getUpgradeUrl: (feature?: string) => {
    return Utils.isSaas()
      ? '/organisation-settings?tab=billing'
      : `https://www.flagsmith.com/pricing${
          feature ? `?utm_source=${feature}` : ''
        }`
  },
```

frontend/web/components/modals/CreateFlag.js
Zaimwa9: No problem. FYI, I reduced a (very little) bit the padding of the buttons so it fits. Otherwise a lead to explore could be a compact badge similar to below
image


Now it fits like this though:
image

frontend/web/components/navigation/OverflowNav.tsx
Zaimwa9: Good catch, I believe it's an llm artifact 😆

frontend/web/styles/project/_utils.scss
Zaimwa9: The problem with this class is that it can get overridden by conflicting backgrounds from other when you want to use it. And we don't have a way to add important directly from `className='xx yy'`
It's used in few places, always conditionally, so it ensures it gets priority no matter the ordering of the classes.

(I wanted to use it, struggled because of this, didn't need it in the end and left this for consistency). I can remove though

frontend/web/styles/project/_FeaturesPage.scss
Zaimwa9: image

😅

Zaimwa9: welcome to css

Zaimwa9: image

I'd suggest to re-harmonize around a `100` z-index but i'd prefer to do it in a dedicated PR to have it isolated in case of regressions

frontend/web/components/base/forms/Button.tsx
matthewelwell: Ok, as long as we've considered it, then I'm good with it. Is there any action to take to improve it in the future? If so, can we create an issue for it?

frontend/web/styles/project/_utils.scss
matthewelwell: I'll leave the decision up to you, but I'd say that if we don't need it, we should remove.

frontend/web/components/base/DropdownMenu.tsx
matthewelwell: See discussion [here](https://github.com/Flagsmith/flagsmith/pull/5987#discussion_r2319324271).

frontend/web/components/base/forms/Button.tsx
Zaimwa9: Issue created [here](https://github.com/orgs/Flagsmith/projects/22/views/2?pane=issue&itemId=127363068&issue=Flagsmith%7Cflagsmith%7C6018)

frontend/web/styles/project/_utils.scss
Zaimwa9: removed

matthewelwell (approved):

matthewelwell:
@Zaimwa9 on the default mac laptop resolution, it looks like release pipelines aren't shown 😞 can we tweak the padding at all to change this? 

image

matthewelwell:
> @Zaimwa9 on the default mac laptop resolution, it looks like release pipelines aren't shown 😞 can we tweak the padding at all to change this?
> 
> image

Ah, @Zaimwa9 it's actually not related to the screen resolution (sorry for the confusion), it's related to the existence of the 'Links' tab, i.e. if the project has an integration with Github configured. For some reason, the existence of the Links tab means that the 'Feature Health' tab also gets hidden away in the extra menu. 

I've tried to debug it, but I'm pretty stumped. The issue seems to be [here](https://github.com/Flagsmith/flagsmith/blob/main/frontend/web/components/modals/CreateFlag.js#L1870). If I change this line to `false &&`, then the feature health tab is shown as expected. What's even weirder though is if I change it to `true &&`, then the feature health tab is _also_ correctly shown! 

I've debugged the frontend logic in the browser and confirmed that `hasIntegrationWithGithub` is `true`, and not some other erroneous data type. 

I am stumped!

PR #5988, opened 2025-08-26
fix(GitHub PoC): Fix permalinks

emyller:
Raised by @Zaimwa9.

> GitHub permalinks are incorrect, e.g.
> 
> ```
> https://github.com/Flagsmith/blob/eec9cfb9fda510ca267c594ac9b88bedebb9bf57/frontend/web/components/Announcement.tsx#L22
> ```

## Changes

Fix a bad `urljoin`.

### DIFF
--- a/api/tests/unit/projects/code_references/test_unit_projects_code_references_views.py
+++ b/api/tests/unit/projects/code_references/test_unit_projects_code_references_views.py
@@ -185,7 +185,7 @@ def test_FeatureCodeReferencesDetailAPIView__responds_200_with_code_references_f
     with freezegun.freeze_time("2099-01-01T10:00:00-0300"):
         FeatureFlagCodeReferencesScan.objects.create(
             project=project,
-            repository_url="https://github.flagsmith.com/backend/",
+            repository_url="https://github.flagsmith.com/backend",
             revision="backend-1",
             code_references=[
                 {
@@ -197,7 +197,7 @@ def test_FeatureCodeReferencesDetailAPIView__responds_200_with_code_references_f
         )
         FeatureFlagCodeReferencesScan.objects.create(
             project=project,
-            repository_url="https://github.flagsmith.com/frontend/",
+            repository_url="https://github.flagsmith.com/frontend",
             revision="frontend-1",
             code_references=[
                 {
@@ -210,7 +210,7 @@ def test_FeatureCodeReferencesDetailAPIView__responds_200_with_code_references_f
     with freezegun.freeze_time("2099-01-02T11:00:00-0300"):
         FeatureFlagCodeReferencesScan.objects.create(
             project=project,
-            repository_url="https://github.flagsmith.com/frontend/",
+            repository_url="https://github.flagsmith.com/frontend",
             revision="frontend-2",
             code_references=[
                 {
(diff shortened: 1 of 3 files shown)


Zaimwa9 (dismissed):

api/tests/unit/projects/code_references/test_unit_projects_code_references_views.py
```diff
@@ -185,7 +185,7 @@ def test_FeatureCodeReferencesDetailAPIView__responds_200_with_code_references_f
     with freezegun.freeze_time("2099-01-01T10:00:00-0300"):
         FeatureFlagCodeReferencesScan.objects.create(
             project=project,
-            repository_url="https://github.flagsmith.com/backend/",
+            repository_url="https://github.flagsmith.com/backend",
```
matthewelwell, line 188: I guess we're confident that github always passes the URL in this format (without the trailing slash)? Do you think we should go belt and braces and check for the existence of the trailing slash, and handle it accordingly?

khvn26, line 188: 1. What guarantees the code references are created without a trailing slash?
2. Does urljoin account for a trailing slash or lack thereof? Should we take care of both cases?
3. Event for a self-hosted Github instance, the example URL feels wrong — the path should include an organisation prefix, I think, i.e. be something like `https://github.flagsmith.com/org/backend`. Coincidentally, the erroneous URL in the PR description also contains a one-element path, and I'm not sure how the patch code fixes that.

khvn26 (changes requested):
I don't understand the nature of the fix and surprised to only see the test inputs, and not outputs, modified. Ideally, I'd like to see a test reproducing the problem causing the malformed URL, or a concise explanation of what's happening here.

api/tests/unit/projects/code_references/test_unit_projects_code_references_views.py
emyller: `urljoin` will handle trailing slashes. Improved testing in b37e0684a38a37b69d9145b53cc0d3a07796f7cd, let me know if that helps!

emyller: > 1. What guarantees the code references are created without a trailing slash?

Nothing. There's no need to, actually.

> 2. Does urljoin account for a trailing slash or lack thereof? Should we take care of both cases?

It accounts for the base URL location, e.g.
```
>>> from urllib.parse import urljoin
>>> urljoin("http://localhost/foo/bar", "baz")
'http://localhost/foo/baz'  # Location is /foo/
```

The added trailing slash to `urljoin` will guarantee the expected behavior by force-indicating the location:
```
>>> urljoin("http://localhost/foo/bar/", "baz")
'http://localhost/foo/bar/baz'
>>> urljoin("http://localhost/foo/bar//", "baz")
'http://localhost/foo/bar/baz'
```

> 3. Event for a self-hosted Github instance, the example URL feels wrong — the path should include an organisation prefix, I think, i.e. be something like `https://github.flagsmith.com/org/backend`. Coincidentally, the erroneous URL in the PR description also contains a one-element path, and I'm not sure how the patch code fixes that.

Improved with 3f669458e9e90a59c9215a3f207a731d1b4f90be.

khvn26: Right, thanks for the explanation 👍

khvn26 (approved):

emyller:
> I don't understand the nature of the fix and surprised to only see the test inputs, and not outputs, modified. Ideally, I'd like to see a test reproducing the problem causing the malformed URL, or a concise explanation of what's happening here.

@khvn26 This patch brings the input closer to reality, while the output already represented the expected behavior. I hope the added tests (b37e0684a38a37b69d9145b53cc0d3a07796f7cd) will address the concern!

matthewelwell (approved):

PR #5999, opened 2025-08-28
fix(Maintenance): Fix Poetry install

emyller:
## Changes

- Use pip to install Poetry env-wise.
- Upgrade Poetry to latest.

## Why

Because installing Poetry globally causes issues with using Poetry elsewhere, only fixable by uninstalling it every time with:

```
curl -sSL https://install.python-poetry.org | python3 - --uninstall
```


matthewelwell (approved):

khvn26 (approved):
Ok 👍

Issue #5860, opened 2025-07-30
SDK Metrics: `User-Agent` allow list


khvn26:
## DoD

- A hardcoded allow list for SDK user agents, in form of `<sdk-name>/<version>`, exists in Core API.
- All incoming `User-Agent` and `Flagsmith-SDK-User-Agent` values are normalised to the above form.
- In case a version can not be detected, `<sdk-name>/unknown` is used.
- In case an SDK can not be detected, `unknown/unknown` is used.

khvn26:
Released as part of https://github.com/Flagsmith/flagsmith/releases/tag/v2.195.0.

Issue #5861, opened 2025-07-30
SDK Metrics: `User-Agent` allow list update


khvn26:
Blocked on https://github.com/Flagsmith/flagsmith/issues/5860, https://github.com/Flagsmith/edge-api/issues/434.

# DoD

Flagsmith/flagsmith has a callable workflow that updates the hard coded SDK user agent allow list open a chore PR.

khvn26:
See https://github.com/Flagsmith/flagsmith/pull/6040.

PR #6000, opened 2025-08-28
feat: SDK user agents allow list, store numeric IDs in InfluxDB

khvn26:

## Changes

Contibutes to #5860.

Adds a hard-coded allow list for `Flagsmith-SDK-User-Agent` and `User-Agent` tracked headers.

Modified existing unit tests to add some of the expected values.


api/app_analytics/constants.py
```diff
@@ -14,6 +14,27 @@
...
+    "flagsmith-kotlin-android-sdk": ["unknown"],
+    "flagsmith-nodejs-sdk": ["unknown"],
+    "flagsmith-php-sdk": ["unknown"],
+    "flagsmith-python-sdk": ["unknown"],
+    "flagsmith-ruby-sdk": ["unknown"],
+    "flagsmith-rust-sdk": ["unknown"],
+    "flagsmith-swift-ios-sdk": ["unknown"],
+}
+
+
```
emyller: Is there any value in keeping track of published versions of each SDK, other than verifying the header informs a _known version_? **Unless this is a requirement**, I firmly believe it to be a silly reason for such a high maintenance cost — even if automated.

api/app_analytics/mappers.py
```diff
@@ -119,21 +118,22 @@ def map_flux_tables_to_feature_evaluation_data(
...
     return labels
 
 
+def map_user_agent_to_sdk_user_agent(value: str) -> str:
+    sdk_name, sdk_version = value.split("/")
+    if versions := SDK_USER_AGENT_KNOWN_VERSIONS.get(sdk_name):
+        if sdk_version in versions:
+            return value
+        return f"{sdk_name}/unknown"
+    return "unknown/unknown"
```
emyller: Adding to the [other comment](https://github.com/Flagsmith/flagsmith/pull/6000/files#r2308555234), this could be simplified to:


```suggestion
    sdk_name, sdk_version = value.split("/")
    if sdk_name not in SDK_USER_AGENT_KNOWN_VERSIONS:
        return "unknown/unknown"
    if not RE_SEMVER.match(sdk_version):
        sdk_version = "unknown"
    return f"{sdk_name}/{sdk_version}"
```

For `RE_SEMVER`, refer to semver.org [regex notes](https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string).

emyller (changes requested):
I'm interested in not keeping track of SDK versions, if possible.

api/app_analytics/constants.py
khvn26: This is a requirement, sorry for not communicating it clearly. We need to 100% prevent any unsolicited values in the labels to avoid blowing up our Influx storage costs. `1.0.832146871274217458712564387125478519745871256348121029843912794768912649812648961284761726475127845632786483174891274971239127` is a valid semver, and there's really no guarantee that all of our SDKs will be semver-versioned, so, for now, I'd like to keep track of the versions here. And yes, this is going to be automated as per #5861.

api/app_analytics/mappers.py
khvn26: See https://github.com/Flagsmith/flagsmith/pull/6000/files#r2310648153.

api/app_analytics/constants.py
emyller: Can we make this clearer with a comment? Also explaining that we only track SDK versions after the user agent header is supported in it — "unknown" otherwise.

> and there's really no guarantee that all of our SDKs will be semver-versioned

Every SDK should be semver-versioned AFAIK — if that ever changes, I guess this will be _just_ one of other problems to solve, e.g. CD. I still think we could circumvent effort with a basic regex such as `r"\d{1,3}\.\d{1,3}\.\d{1,3}"`, with optional support for pre-releases, which I think we don't practice. Honestly if someone will take their time to forcefully try and submit any invalid version, I can't see that leading to much trouble as long as we sanity check inputs.

Not blocking.

khvn26: Explained and expanded upon in e8503823273dafa4d6a5aec968a54ca08d8204a0.

The initial plan was to store numeric IDs for specific sdk/version pairs, and I added this functionality in the above commit.

api/app_analytics/constants.py
```diff
@@ -14,14 +14,55 @@
...
+    "flagsmith-python-sdk": ["unknown"],
+    "flagsmith-ruby-sdk": ["unknown"],
+    "flagsmith-rust-sdk": ["unknown"],
+    "flagsmith-swift-ios-sdk": ["unknown"],
+}
+
+SDK_USER_AGENT_INFLUX_IDS: list[tuple[KnownSDK, int]] = [
+    (sdk, idx * MAX_KNOWN_VERSIONS_PER_SDK)
+    for idx, sdk in enumerate(get_args(KnownSDK))
+]
```
emyller: From my understanding, 1. this depends on [get_args](https://docs.python.org/3/library/typing.html#typing.get_args) returning SDKs in a consistent order, and 2. if the order changes, SDK IDs will be very wrong.

If the assessment is correct, can we improve and define order separately, e.g. reusing `SDK_USER_AGENT_KNOWN_VERSIONS.keys()` and adding a comment about the immutable order?

khvn26: Good point. Improved in c3bd55901930ccc4dba527e9275b621cfc1176ff.

emyller (approved):
LGTM!

PR #6013, opened 2025-09-03
fix: catch-error-on-segment-creation-or-update

Zaimwa9:

## Changes
- Made segment on save callback async to catch error
- Added error message: Metadata if identifier or generic segment error
- Fixed conditional access to plan startup

N.B: after saving metadata is shown as being edited. But that is already existing on prod

1. Add a required metadata to segment
2. Update segment without setting the metadata
3. Verify error message is here

image

### DIFF
--- a/frontend/web/components/modals/CreateSegment.tsx
+++ b/frontend/web/components/modals/CreateSegment.tsx
@@ -211,41 +211,39 @@ const CreateSegment: FC<CreateSegmentType> = ({
     setRules(newRules)
   }
 
-  const removeRule = (rulesIndex: number, elementNumber: number) => {
-    const newRules = cloneDeep(rules)
-    newRules[0].rules.splice(elementNumber, 1)
-    setRules(newRules)
-  }
-
-  const save = (e: FormEvent) => {
-    Utils.preventDefault(e)
-    setValueChanged(false)
-    setMetadataValueChanged(false)
-    const segmentData: Omit<Segment, 'id' | 'uuid'> = {
-      description,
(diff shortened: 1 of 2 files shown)


frontend/web/components/modals/CreateSegment.tsx
```diff
@@ -420,6 +418,7 @@ const CreateSegment: FC<CreateSegmentType> = ({
             entity={segmentContentType?.model}
             onChange={(m) => {
               setMetadata(m as Metadata[])
+              // Need to fix this to be more robust and handle post save
```
Zaimwa9, line 421: This part is responsible for the false signal of data being edited. It's happening on prod too and the logic is too nested to be quick fixed.

frontend/web/components/modals/CreateSegment.tsx
```diff
@@ -211,41 +211,39 @@ const CreateSegment: FC<CreateSegmentType> = ({
     setRules(newRules)
   }
 
-  const removeRule = (rulesIndex: number, elementNumber: number) => {
-    const newRules = cloneDeep(rules)
-    newRules[0].rules.splice(elementNumber, 1)
-    setRules(newRules)
-  }
```
matthewelwell, line 218: Was this just removing an unused function?

matthewelwell (approved):

frontend/web/components/modals/CreateSegment.tsx
Zaimwa9: yes

PR #6022, opened 2025-09-05
fix: added-dropdown-opening-step-inversioning-tests

Zaimwa9:

## Changes
- Updated test to use dropdown

- Ran tests against staging

### DIFF
--- a/frontend/e2e/helpers.cafe.ts
+++ b/frontend/e2e/helpers.cafe.ts
@@ -371,7 +371,12 @@ export const logout = async () => {
 
 export const goToFeatureVersions = async (featureIndex: number) => {
   await gotoFeature(featureIndex)
-  await click(byId('change-history'))
+  if (await isElementExists('change-history')) {
+    await click(byId('change-history'))
+  } else {
+    await click(byId('tabs-overflow-button'))
+    await click(byId('change-history'))
+  }
 }
 
 export const compareVersion = async (
--- a/frontend/e2e/tests/versioning-tests.ts
+++ b/frontend/e2e/tests/versioning-tests.ts
@@ -8,7 +8,7 @@ import {
     createRemoteConfig,
     editRemoteConfig, getFlagsmith,
     log,
-    login, refreshUntilElementVisible,
+    login,
     waitForElementVisible,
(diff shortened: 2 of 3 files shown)


frontend/e2e/helpers.cafe.ts
```diff
@@ -371,7 +371,12 @@ export const logout = async () => {
 
 export const goToFeatureVersions = async (featureIndex: number) => {
   await gotoFeature(featureIndex)
-  await click(byId('change-history'))
+  if (await isElementExists('change-history')) {
```
khvn26, line 374: I hope we won't have to change this to a while loop :D

khvn26 (approved):
Looks good

emyller (approved):

frontend/e2e/helpers.cafe.ts
Zaimwa9: Ahah no, it's just to check if it's hidden and so we don't have to update every time we add an item and/or  change the order

frontend/e2e/helpers.cafe.ts
```diff
@@ -371,7 +371,12 @@ export const logout = async () => {
...
+    await click(byId('change-history'))
+  } else {
+    await click(byId('tabs-overflow-button'))
+    await click(byId('change-history'))
+  }
```
emyller, line 379:
```suggestion
  if (await isElementExists('tabs-overflow-button')) {
    await click(byId('tabs-overflow-button'))
  }
  await click(byId('change-history'))
```

Would this be a clearer option?

Zaimwa9: No it wouldn't work with this test. 
The `tabs-overflow-button` should always be visible except if you are in 20% zoom. The conditions checks whether history is a visible tab or accessible from the dropdown.

Zaimwa9: Ultimately we should do it for all the modal tabs so we don't have to update tests each time we change the order or/and add-remove tabs

emyller: Thanks for explaining!

Issue #5490, opened 2025-05-26
Identifiers are not sanitised


matthewelwell:
Currently identifiers created via the API are not sanitised. For example, it is possible to create an identity with the identifier `<script>alert(1)</script>` with the following request: 

```
curl 'https://<flagsmith api URL>/api/v1/identities/?identifier=%3Cscript%3Ealert%281%29%3C%2Fscript%3E' --header 'X-Environment-Key: <key>'
```

While this does not directly cause any vulnerability since the content is escaped before being displayed anywhere in the flagsmith dashboard, it is something that could cause problems downstream for customers if they are rendering the data and not properly escaping the content.

Todo: 

 - Add a list of allowed characters for identifiers and validate against this. Suggested whitelist regex pattern is ``^[\w\u00C0-\u024F\u0370-\u03FF!#$%&*+/=?^_`{|}~@.\-]+$``. Note that this will need to be implemented in the Core API, and the Edge API. 
   - Core API
   - Edge API
 - Update documentation

PR #6024, opened 2025-09-06
fix(Identities): Sanitize identifiers

emyller:
Contributes to https://github.com/Flagsmith/flagsmith/issues/5490.

Rewrites how the endpoint `/api/v1/identities/` handles input, properly validating data, and ultimately **rejecting** invalid identifiers.

This includes a state-only migration (safe to apply without database changes).


api/environments/identities/views.py
```diff
@@ -161,14 +162,18 @@ class SDKIdentities(SDKAPIView):
...
+    def get(self, request: Request) -> Response:
+        query_serializer = SDKIdentitiesQuerySerializer(data=request.query_params)
+        if not query_serializer.is_valid():
+            [error_message] = query_serializer.errors["identifier"]
             return Response(
-                {"detail": "Missing identifier"}
-            )  # TODO: add 400 status - will this break the clients?
+                {"detail": error_message},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
```
emyller: In theory, this could be a breaking change, but I intend it to be a fix to the current version. Perhaps there's a story behind this but I'd be very interested in moving forward with proper HTTP communication and help dealing with any consequences.

gagantrivedi (changes requested):
Looks good, but we also need to update: https://github.com/Flagsmith/flagsmith/blob/main/api/edge_api/identities/serializers.py#L55 and add tests for https://github.com/Flagsmith/flagsmith/blob/main/api/environments/identities/views.py#L41 and https://github.com/Flagsmith/flagsmith/blob/main/api/environments/identities/views.py#L99 and https://github.com/Flagsmith/flagsmith/blob/main/api/environments/identities/traits/views.py#L112

api/environments/identities/views.py
```diff
@@ -161,14 +162,18 @@ class SDKIdentities(SDKAPIView):
             cache=settings.GET_IDENTITIES_ENDPOINT_CACHE_NAME,
         )
     )
-    def get(self, request):  # type: ignore[no-untyped-def]
-        identifier = request.query_params.get("identifier")
-        if not identifier:
+    def get(self, request: Request) -> Response:
+        query_serializer = SDKIdentitiesQuerySerializer(data=request.query_params)
+        if not query_serializer.is_valid():
+            [error_message] = query_serializer.errors["identifier"]
```
gagantrivedi: What happens if the query serializer has errors, but they are not caused by the identifier field?

emyller: Good question. This is based on `SDKIdentitiesQuerySerializer` only having two fields:
- `identifier` (required)
- `transient` (optional, default = false)

I'd prefer to respond the request with all serializer errors, though the above decision is to keep a minimal similarity with existing error response message for this endpoint. Let me know if that works for you.

emyller:
> Looks good, but we also need to update: (...) and add tests for (...)

@gagantrivedi Thanks! Unpacking the multiple changes:
1. Identifier validation is now also implemented in the `edge-identifiers` endpoint (e49a57d32ed3503fc860423973604cf12eefc2e4).
2. Tests were improved in 8f8167f6fcc34b6d769fe0d08bbf69f507037901.
3. Regarding a couple of the references, I prefer not to make changes because:
    - They are deprecated endpoints.
    - Their associated URLs ([ref](https://github.com/Flagsmith/flagsmith/blob/e49a57d32ed3503fc860423973604cf12eefc2e4/api/api/urls/deprecated.py)) already use a subset of the regular expression introduced to validate identifiers — let's not expand it.

api/environments/identities/views.py
gagantrivedi: I guess what I am trying to say is that we are assuming whenever the serializer fails, it must have been caused by the identifier field. But since there are other fields in the serializer, if it fails because of those, this code will crash.
e.g: the following test: 

```
def test_SDKIdentities__identifier_sanitization__accepts_valid_identifiers(
    api_client: APIClient,
    environment: Environment,
    given_identifier: str,
) -> None:
    # Given
    api_client.credentials(HTTP_X_ENVIRONMENT_KEY=environment.api_key)

    # When
    response = api_client.get(
        f"/api/v1/identities/?identifier={quote(given_identifier)}&transient=meh"
    )

    # Then
    assert response.json()["identifier"] == unquote(given_identifier)


```

gagantrivedi:
> > Looks good, but we also need to update: (...) and add tests for (...)
> 
> @gagantrivedi Thanks! Unpacking the multiple changes:
> 
>     1. Identifier validation is now also implemented in the `edge-identifiers` endpoint ([e49a57d](https://github.com/Flagsmith/flagsmith/commit/e49a57d32ed3503fc860423973604cf12eefc2e4)).
> 
>     2. Tests were improved in [8f8167f](https://github.com/Flagsmith/flagsmith/commit/8f8167f6fcc34b6d769fe0d08bbf69f507037901).
> 
>     3. Regarding a couple of the references, I prefer not to make changes because:
>        
>        * They are deprecated endpoints.
>        * Their associated URLs ([ref](https://github.com/Flagsmith/flagsmith/blob/e49a57d32ed3503fc860423973604cf12eefc2e4/api/api/urls/deprecated.py)) already use a subset of the regular expression introduced to validate identifiers — let's not expand it.



Fair enough, but I guess we missed another SDK endpoint (that is no deprecated and should get some tests) https://github.com/Flagsmith/flagsmith/blob/main/api/environments/identities/traits/views.py#L182

gagantrivedi (changes requested):

api/environments/identities/views.py
emyller: I see. I had completely ignored `transient` being able to raise a validation error. This is now fixed in be5dc35cec8aebb93e6f3d645b444ec20fc0019f. Thanks!

api/environments/identities/views.py
emyller: Changes are deeper now since this thread: https://github.com/Flagsmith/flagsmith/pull/6024#discussion_r2331970354. @matthewelwell Please chime in and help ensuring we want to go this direction.

api/environments/identities/constants.py
```diff
@@ -0,0 +1,10 @@
+import re
+
+from django.core.validators import RegexValidator
+
+RE_VALID_IDENTIFIER = re.compile(r"^[\w!#$%&*+/=?^_`{}|~@.\-]+$")
+
+identifier_regex_validator = RegexValidator(
+    regex=RE_VALID_IDENTIFIER,
+    message="Identifier can only contain unicode letters, numbers, and the symbols: ! # $ %% & * + / = ? ^ _ ` { } | ~ @ . -",
```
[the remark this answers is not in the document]
emyller: Thanks, though I believe this is incorrect. Here's proof:
1. https://github.com/Flagsmith/flagsmith/blob/be5dc35cec8aebb93e6f3d645b444ec20fc0019f/api/tests/integration/edge_api/identities/test_edge_identity_viewset.py#L23
2. https://github.com/django/django/blob/46fdeb1373aa7e9089d14440987444493cc9c2e0/django/core/exceptions.py#L214

Unless there's anything I don't see, please resolve this thread.

api/environments/identities/views.py
matthewelwell: Yep, I'm happy with the approach here.

emyller:
> Fair enough, but I guess we missed another SDK endpoint (that is no deprecated and should get some tests) https://github.com/Flagsmith/flagsmith/blob/main/api/environments/identities/traits/views.py#L182

You're right, added tests here: 1ba40b9f81479e68c88f060debc0b84d90fdc009. I also did a search through the API and it looks like these were all views affected by the model change. Thanks for pointing them out so far!

api/tests/integration/edge_api/identities/test_edge_identity_viewset.py
```diff
@@ -116,6 +118,52 @@ def test_create_identity(  # type: ignore[no-untyped-def]
     assert response.json()["identity_uuid"] is not None
 
 
+@pytest.mark.valid_identity_identifiers
```
gagantrivedi: nit: Is there a reason why these tests are part of the integration tests and not the [unit tests](https://github.com/Flagsmith/flagsmith/blob/main/api/tests/unit/edge_api/identities/test_edge_api_identities_views.py)? I understand that the line between integration and unit is blurry here, but this test seems to belong more to unit than integration.

gagantrivedi (changes requested):

api/tests/conftest.py
```diff
@@ -16,6 +16,41 @@
 )
 
 
+def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
```
gagantrivedi: Depends on the resolution of `https://github.com/Flagsmith/flagsmith/pull/6024#discussion_r2338407010` but this should live under some `..unit/conftest.py`

api/tests/integration/edge_api/identities/test_edge_identity_viewset.py
emyller: The reason was that these tests seemingly required the `dynamo_enabled_environment` fixture from integration tests. Turns out they don't, thanks for the call out. Improved with fa5b39e1904aace080c872257203491d3b29f13f.

api/tests/conftest.py
emyller: Moved in fa5b39e1904aace080c872257203491d3b29f13f.

gagantrivedi (approved):

PR #6032, opened 2025-09-08
ci: Add known SDK version workflow

khvn26:
Contributes to #5861.

## Changes

This PR adds a dispatchable workflow to automatically open PRs with SDK allow list updates.
The workflow makes sure that the currently present versions are persisted in the same order, and appends a new version to the list.
 
Tested the Makefile target locally. Will test end-to-end once the PR is merged.


.github/workflows/api-run-makefile-target.yml
```diff
@@ -0,0 +1,57 @@
...
+          opts: ${{ inputs.opts }}
+        run: make ${{ inputs.target }}
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v5
+        with:
+          branch: chore/make-${{ inputs.target }}-${{ github.run_id }}
+          title: ${{ inputs.pr-title }}
+          body: |
+            Results of `${{ inputs.target }} opts="${{ inputs.opts }}"` ran on commit ${{ github.sha }}.
```
emyller:
```suggestion
            Results of `make ${{ inputs.target }} opts="${{ inputs.opts }}"` ran on commit ${{ github.sha }}.
```

api/scripts/add-known-sdk-version.py
emyller, line 1: I wonder if we could solve this problem with a simpler manner, such as with a regex that helps identifying where to place the new string. The snippet here, although looking like a big achievement, also feels a bit over-engineered and not very portable — sorry.

emyller (changes requested):
Looks good but I think we'd be happier with less code?

api/scripts/add-known-sdk-version.py
khvn26: I believe that trying to solve this with regex/string manipulation would result in actually over-engineered, and less correct, code. I'd love to be proven wrong, though. 

Can you expand on the portability concern? It's supposed to be run in a working dev environment, which is reliably replicated by poetry in CI.

api/scripts/add-known-sdk-version.py
```diff
@@ -0,0 +1,143 @@
...
+    parser = argparse.ArgumentParser(description=AddKnownSDKVersion.DESCRIPTION)
+    AddKnownSDKVersion.add_args(parser)
+    args = parser.parse_args()
+    context = CodemodContext()
+    codemod = AddKnownSDKVersion(
+        context=context,
+        sdk=args.sdk,
+        version=args.version,
+    )
+    code = open(MODULE_PATH).read()
```
emyller: Even though this is short-lived, I think it's best as:

```suggestion
    code = Path(MODULE_PATH).read_text()
```

api/scripts/add-known-sdk-version.py
emyller: > I believe that trying to solve this with regex/string manipulation would result in actually over-engineered, and less correct, code. I'd love to be proven wrong, though.

Here's a very PoC snippet:

```python
import re

constant_name = "KNOWN_SDK_VERSIONS"

code = """
this = "foo"

KNOWN_SDK_VERSIONS = [
    "1.0.0",
    "2.0.0",
    "3.0.0",
]

that = "bar"
"""

new_version = "4.0.0"

new_code = re.sub(
    pattern=rf"""
        ^  # Expect to have definition at module-level
        (?P<constant_name>{constant_name})
        \s=\s\[  # List opening
        (?P<existing_versions>
            [^\]]*  # Everything within it
        )?
    """,
    repl=rf'\g<constant_name> = [\g<existing_versions>    "{new_version}",\n',
    string=code,
    flags=re.MULTILINE | re.VERBOSE,
)

assert new_code == """
this = "foo"

KNOWN_SDK_VERSIONS = [
    "1.0.0",
    "2.0.0",
    "3.0.0",
    "4.0.0",
]

that = "bar"
"""
```

It can be more sophisticated if necessary, e.g. being aware of indentation, quote-style, etc, though perhaps we don't even need to go there.

> Can you expand on the portability concern? It's supposed to be run in a working dev environment, which is reliably replicated by poetry in CI.

This came from the assumption that we'll need to port this code to edge API. IMO we'd be happier with code that demands less braining around... Or perhaps I'm biased by my appreciation of regular expressions? 😛

khvn26: I spent some time on this and concluded that my regex-fu is not good enough to produce an elegant solution.

khvn26: > we'll need to port this code to edge API

...at the cost of adding one dev dependency. Of course, having to rely just on stdlib would be preferable. I tried using stdlib's `ast` but that doesn't preserve comments in the module.

emyller: Expanding from the [above experiment](https://github.com/Flagsmith/flagsmith/pull/6032/files#r2331568136), and **given Ruff can be used to fix formatting** after running the script, I think this can help:

```python
import pathlib
import re


VERSIONS_CONSTANT_NAME = "SDK_USER_AGENT_KNOWN_VERSIONS"


def add_version(sdk_name: str, new_version: str) -> None:
    """Add a new version to known versions constants"""
    versions_file = pathlib.Path(__file__).parent / "app_analytics/constants.py"
    contents = versions_file.read_text()

    # Add SDK version
    contents = re.sub(
        pattern=rf"""
            ^  # Expect to have definition at module-level
            (?P<constant_def>{VERSIONS_CONSTANT_NAME}[^=]*=\s*\{{)
            (?P<prior_sdks>[^\}}]+)
            (?P<sdk_def>"{sdk_name}"\s*:\s*\[)
            (?P<sdk_versions>[^\]]*)
            (?P<latest_version>"(?:unknown|\d[^"]+)"),?
        """,
        repl=(
            r"\g<constant_def>\g<prior_sdks>\g<sdk_def>\g<sdk_versions>\g<latest_version>"
            f',\n    "{new_version}",\n    '
        ),
        string=contents,
        count=1,
        flags=re.MULTILINE | re.VERBOSE,
    )

    print(contents)
    # versions_file.write_text(contents)


add_version("flagsmith-js-sdk", "9.4.0")
```

e.g. this should lead to the expected result given any **known** SDK name and new version to add.

```sh
python this_script.py | ruff format -
```

khvn26: Added 6860da06ead367a30ceb7c4643ee2d8529821173, seems to work fine 👍 Thanks for helping on this.

emyller (approved):
:shipit:

PR #6054, opened 2025-09-11
docs: Remove old documentation pages

guiwrite:

## Changes

- [ ] Remove old documentation sections

_Please describe._


matthewelwell:
@guiwrite thanks for all the work on this. 

I've added some initial thoughts below from navigating around the preview [here](https://docs-git-fork-writechoiceorg-docs-writechoice-54677f-flagsmith.vercel.app/). 

 1. The 'Advanced Capabilities' option links straight to A/B testing which feels a bit odd. I would expect to be directed to some kind of interstitial page, but that doesn't really suit the structure of our documentation. I wonder if we should just change the content of the link on the homepage since 'Advanced Capabilities' is pretty vague anyway. Do you have any suggestions on what we should do here? 
 2. I don't really like the fact that we have a quite heavily prioritised to our pricing page in the docs. I'd like to keep sales out of the docs as much as possible! Perhaps this gives us the opportunity to split the 'Advanced Capabilities' into 2 separate tiles, and replace the compare plans one too? 
 3. The 'Integrations' tile makes it look like we only have integrations with Segment and Datadog. I think we should either (in priority order) update the content to say something like 'Integrations with third-party observability, analytics and other platforms. Also, this tile links straight to the segment integration page. Is there a way to link to an interstitial 'integrations page'? Perhaps similar to the home page, where the tiles are the individual third-party platforms? 
 4. For the 'System settings' tile, the sub heading is Admin settings and multi-tenancy, but it links to the page on environment settings. What's the logic here as it certainly doesn't make much sense to me. What is multi-tenancy meant to be referring to? 

These are just the main points that I've identified from the home page. I'll go through the structure of the rest of the documentation tomorrow.

Zaimwa9:
Hi! Good job, that's a lot.
Here are a couple of comments I noted down while going over in surface (up to Flag Analytics) - a lot of points might be irrelevant so I just put them on your radar:

- Footer how-to guide could be split over 2 columns?
- ClientSDK tile redirects to integrations overview and not client SDK (versus server correctly mapped)
- Hover colors not right in (at least) Administration and security. Also present in [Edge API doc](https://docs-git-fork-writechoiceorg-docs-writechoice-54677f-flagsmith.vercel.app/edge-api) + the link is really on the text, making it a bit hard to access 
image


- Might be on purpose but "[Docker](https://docs-git-fork-writechoiceorg-docs-writechoice-54677f-flagsmith.vercel.app/deployment-self-hosting/hosting-guides/docker)" title is not upper-case
image

- Might be related to deployment method but curl icon is missing in [edge API specifications](https://docs-git-fork-writechoiceorg-docs-writechoice-54677f-flagsmith.vercel.app/edge-api/get-flags):
image


- [Wording](https://docs-git-fork-writechoiceorg-docs-writechoice-54677f-flagsmith.vercel.app/getting-started/feature-flags): "You are about to start work on a new feature." => feels like start working?
- "[Rest API reference](https://docs-git-fork-writechoiceorg-docs-writechoice-54677f-flagsmith.vercel.app/getting-started/quick-start#automation--api-access)" redirects to edge API ? Shouldn't it be the [former](https://api.flagsmith.com/api/v1/docs/)  ?
- NIT: [Projects](https://docs-git-fork-writechoiceorg-docs-writechoice-54677f-flagsmith.vercel.app/flagsmith-concepts/data-model#projects): Maybe specific "Paid/Paying organisations can have unlimited number of projects" ?
- ["What's next"](https://docs-git-fork-writechoiceorg-docs-writechoice-54677f-flagsmith.vercel.app/flagsmith-concepts/data-model#projects) seems a bit odd given the previous/next native buttons
- NIT: [In this example](https://docs-git-fork-writechoiceorg-docs-writechoice-54677f-flagsmith.vercel.app/flagsmith-concepts/identities#using-traits-to-drive-segments), I would favor using javascript instead of Java (more accessible) but that's a personal biased opinion
- Unsure: in [security & privacy](https://docs-git-fork-writechoiceorg-docs-writechoice-54677f-flagsmith.vercel.app/flagsmith-concepts/identities#using-traits-to-drive-segments), both remote and local evaluation links are redirecting to integrations
- NB: I think we added context values documentation after this work start so I couldn't [find it](https://docs-git-fork-writechoiceorg-docs-writechoice-54677f-flagsmith.vercel.app/flagsmith-concepts/identities#using-traits-to-drive-segments)
- NIT: Extra new lines it seems in the bullet lists [here](https://docs-git-fork-writechoiceorg-docs-writechoice-54677f-flagsmith.vercel.app/flagsmith-concepts/identities#using-traits-to-drive-segments) versus rest of the doc
- Personal opinion: RBAC should be the first section of Access Control


I'll do another round in the next day, but let me know what format you'd prefer for comments, what type of comments were irrelevant so I don't add noise to the feedback 👍

matthewelwell:
@guiwrite thanks for making the changes to the home page, and adding the interstitial pages - that looks a lot better now. A few other comments on the structure of pages: 

image

This doesn't really make sense - Flag Analytics aren't related to experimentation or A/B testing. Let's move that into 'Managing Flags'. That leaves us with the only item under experimentation being 'A/B Testing', but I can't see any other pages that really fit there. I guess that leaves us with 2 options: 

 1. Remove the nesting so that the page just becomes something like 'Experimentation (A/B Testing)'
 2. Leave the nesting as it is, and perhaps we'll add another page in the future

I think I prefer option (1) at this point, but I'm interested to know your thoughts. 

image

I think 'Flagsmith Integration' sounds a bit too generic. Can we change this to 'Integrating with Flagsmith'? 

image

I'm not sure why we have the main parent of 'Managing Flags', and then a sub-parent of 'Manage Flags' - this feels odd. Can we just get rid of 'Manage Flags' and flatten out the pages underneath it?

docs/docs/advanced-use/release-pipelines.md
matthewelwell, line 1: @guiwrite I think this page was added since we started the project. Could we, instead of removing it, move it to the managing-flags section?

PR #6062, opened 2025-09-15
fix: removed-gap-in-overflow-hook-invocation

Zaimwa9:

## Changes
- Removed extra gap safety net when using overflow hook

- Against not working set-up

image

### DIFF
--- a/frontend/web/components/modals/CreateFlag.js
+++ b/frontend/web/components/modals/CreateFlag.js
@@ -1874,7 +1874,7 @@ const CreateFlag = class extends Component {
                                         tabLabelString='Links'
                                         tabLabel={
                                           <Row className='justify-content-center'>
-                                            Links{' '}
+                                            Links
                                           </Row>
                                         }
                                       >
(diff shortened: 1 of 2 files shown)


frontend/web/components/modals/CreateFlag.js
```diff
@@ -1874,7 +1874,7 @@ const CreateFlag = class extends Component {
                                         tabLabelString='Links'
                                         tabLabel={
                                           <Row className='justify-content-center'>
-                                            Links{' '}
+                                            Links
```
matthewelwell, line 1877: Ah, I was on the right lines! I made this change too, but couldn't get it to consistently resolve the issue.

frontend/web/components/navigation/TabMenu/Tabs.tsx
```diff
@@ -94,7 +94,7 @@ const Tabs: React.FC<TabsProps> = ({
   const { isMeasuring, visibleCount } = useOverflowVisibleCount({
     extraWidth: overflowButtonWidth,
     force: false,
-    gap: 2,
+    gap: 0,
```
matthewelwell, line 97: I was probably missing this!

matthewelwell (approved):

frontend/web/components/navigation/TabMenu/Tabs.tsx
Zaimwa9: yes I mised it too actually, it's part of the initial implementation to provide a way to include extra gutters but it's already handled by the tabs

PR #6083, opened 2025-09-17
fix: get-identities-checks-existing-invalid-identifiers

### LINKED ISSUE #6082: 2.196.1 identity sanitization does not allow whitespace

- [ ] Self Hosted with Docker
- [x] SaaS at flagsmith.com

### Describe the bug

https://github.com/Flagsmith/flagsmith/pull/6024 prevents identities with whitespace from being used - but existing identities that had been created with whitespace now no longer can be accessed.

### Steps To Reproduce

1. hit the `https://api.bullet-train.io/api/v1/identities/?identifier=xxx%20xxx` endpoint
2. endpoint returns the `Identifier can only contain unicode letters, numbers, and the symbols: ! # $ %% & * + / = ? ^ _ ` { } | ~ @ . -` error

### Expected behavior

Provide some way of updating all identities or allow whitespace

### Screenshots

_No response_

Zaimwa9:
Contributes to #6082

---

## Changes

- Allow white spaces in identifiers.


api/environments/identities/constants.py
```diff
@@ -6,5 +6,5 @@
 
 identifier_regex_validator = RegexValidator(
     regex=RE_VALID_IDENTIFIER,
-    message="Identifier can only contain unicode letters, numbers, and the symbols: ! # $ %% & * + / = ? ^ _ ` { } | ~ @ . -",
+    message="Identifier can only contain unicode letters, numbers, and the symbols: ! # $ % & * + / = ? ^ _ ` { } | ~ @ . -",
```
emyller: The double `%` here is intentional, as to prevent Django from doing string interpolation. Check out this thread: https://github.com/Flagsmith/flagsmith/pull/6024#discussion_r2337595607

api/environments/identities/serializers.py
```diff
@@ -66,11 +68,28 @@ class _TraitSerializer(serializers.Serializer):  # type: ignore[type-arg]
...
+            environment = getattr(request, "environment", None)
+            if (
+                environment
+                and Identity.objects.filter(
+                    environment=environment, identifier=value
+                ).exists()
+            ):
+                return value
+            raise ValidationError(identifier_regex_validator.message)
+
```
emyller: Once the validation rule is fixed in order to allow white spaces, this should not be necessary.

emyller (changes requested):
I believe the fix to be allowing white spaces in the validation regex.

emyller (approved):
LGTM

PR #6092, opened 2025-09-22
fix(features): fix cache isolation for server-only features

blackjid:

## Changes

Fixed a cache isolation bug where server-only features could leak to client requests. And the other way around, where server-only keys could not be presented because they are cached from a client request.

**Problem:** 
The features endpoint cache used only `environment.api_key` as the cache key. Both CLIENT and SERVER requests shared the same cache, so server-only features cached by server requests could be served to client requests.

**Solution:**
Updated the cache key to include request origin: `f"{environment.api_key}:{request_origin}"`

Now CLIENT and SERVER requests have separate cache entries:
- CLIENT: `env_123:CLIENT` 
- SERVER: `env_123:SERVER`

**Files changed:**
- `api/features/views.py` - Fixed cache key
- `api/tests/unit/features/test_unit_features_views.py` - Added test

**TDD approach:**

1. **Added failing test** - Created test that demonstrated the bug: server request populated cache, then client request incorrectly got server-only features
2. **Implemented fix** - Modified cache key to include request origin  
3. **Test now passes** - Client and server requests properly isolated

**Results:**
```bash
# Before fix: FAILED ❌ 
# After fix: PASSED ✅
pytest test_get_flags__server_key_only_feature__cache_isolation_between_client_and_server_keys

# No regressions: ✅
pytest tests/unit/features/test_unit_features_views.py -k "cache"
```
```

### DIFF
--- a/api/features/views.py
+++ b/api/features/views.py
@@ -859,7 +859,9 @@ class SDKFeatureStates(GenericAPIView):  # type: ignore[type-arg]
         from_replica: bool = False,
     ) -> list[typing.Any]:
         data: list[typing.Any]
-        data = flags_cache.get(environment.api_key)
+        # Include request origin in cache key to isolate client vs server requests
+        cache_key = f"{environment.api_key}:{self.request.originated_from.value}"
+        data = flags_cache.get(cache_key)
         if not data:
             data = self.get_serializer(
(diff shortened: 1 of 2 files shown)


matthewelwell (approved):
Looks good to me.

PR #6103, opened 2025-09-24
feat: show-redirections-links-in-feature-health-tab

Zaimwa9:

## Changes
We weren't doing the distinction between feature health providers set or not. This PR aims to redirect to the project settings if admin, or to show a message to contact the admin

- I tried with adding the "Note" InfoMessage to show the documentation. If not ok i'll remove
- Besides, there are a couple of small UI fixes done in permissions at the same time (more would have started to be way out of scope but the layout could definitely be improved)
- overflow nav is disabled for `<Tabs theme='pills'` />`

- Different users with different permissions

image

image

image

**Fixes**
Before
image

After
image

Before (to reproduce you need to click "groups" tab and come back to permissions)
image

After
image


matthewelwell:
image

I think the collapsed Note looks a bit odd, and it's not clear that it's expandable. 

I'd say that we just maybe update the main body content to be something like: 

```
Configure a health provider in your [project settings](/project-settings) to start monitoring your feature health, or read about the functionality [here](docs.flagsmith.com/insert-correct-link-here).
```

matthewelwell (approved):

PR #6113, opened 2025-09-29
fix: Datadog integration not respecting `use_custom_source`

khvn26:

## Changes

This actually enables the Datadog source name added in https://github.com/Flagsmith/flagsmith/pull/3342. 

The unit test for the integration is extended to include the expected attribute.

### DIFF
--- a/api/audit/signals.py
+++ b/api/audit/signals.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Any, Callable, Literal, Protocol, Type
+from typing import TYPE_CHECKING, Any, Callable, Literal, Protocol, Type
 
 from django.conf import settings
 from django.db.models.signals import post_save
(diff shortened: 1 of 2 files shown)


Zaimwa9 (approved):
ok looks good to me 👍 

Even if it's in the doc, i'd add in the integration form a tooltip with what means checking custom source

matthewelwell (approved):

api/audit/signals.py
```diff
@@ -129,14 +133,17 @@ def _track_event_async(instance, integration_client):  # type: ignore[no-untyped
 @receiver(post_save, sender=AuditLog)
 @track_only_feature_related_events
 def send_audit_log_event_to_datadog(sender, instance, **kwargs):  # type: ignore[no-untyped-def]
-    data_dog_config = _get_integration_config(instance, "data_dog_config")
+    data_dog_config: "DataDogConfiguration | None" = _get_integration_config(
+        instance, "data_dog_config"
+    )
```
emyller, line 138: "Datadog" is a proper noun and should not be separated.

emyller (commented):
Minor change needed, looks good otherwise!

api/audit/signals.py
emyller: Dismissing as this refers to legacy code. We should definitely fix this, but it's not important now.

emyller (approved):

PR #6119, opened 2025-09-30
chore: componentise feature filters

kyle-ssg:

## Changes

- Componentises feature table filters so that they can be reused between features, identity and future pages completely
- Adds persistent URL filtering to identity page 

Filter and sort on both the features page and an identity


frontend/web/components/feature-page/FeatureFilters.tsx
```diff
@@ -0,0 +1,221 @@
...
+import TableSortFilter, { SortValue } from 'components/tables/TableSortFilter'
+import { getViewMode, setViewMode, ViewMode } from 'common/useViewMode'
+import { isEqual } from 'lodash'
+import Utils from 'common/utils/utils'
+import { TagStrategy } from 'common/types/responses'
+import Format from 'common/utils/format'
+
+export type FiltersValue = {
+  search: string | null
+  releasePipelines: number[]
```
Zaimwa9: Are we using this in the end 🤔 ?

frontend/web/components/feature-page/FeatureFilters.tsx
```diff
@@ -0,0 +1,221 @@
...
+  page: 1,
+  releasePipelines: [],
+  search: '',
+  sort: { label: 'Name', sortBy: 'name', sortOrder: 'asc' },
+  tag_strategy: 'INTERSECTION',
+  tags: [],
+  value_search: '',
+}
+
+const sortToHeader = (s: any) => {
```
Zaimwa9: What is the `toHeader` part related to ?

frontend/web/components/feature-page/FeatureFilters.tsx
```diff
@@ -0,0 +1,221 @@
...
+  const existing = Utils.fromParam() as Record<string, string | undefined>
+
+  return {
+    ...existing,
+    group_owners: f.group_owners?.length ? f.group_owners.join(',') : undefined,
+    is_archived: f.is_archived ? 'true' : undefined,
+    is_enabled:
+      f.is_enabled === null ? undefined : f.is_enabled ? 'true' : 'false',
+    owners: f.owners?.length ? f.owners.join(',') : undefined,
+    page: f.page !== DEFAULTS.page ? String(f.page) : undefined,
```
Zaimwa9, line 61: Mmh, so it's not from this PR but testing the pagination. It seems that when we use the `< >` to navigate, we don't update the text: (100-150 out of XXX) and the current page. I don't know if you think it's straightforward to fix otherwise we can create a ticket

frontend/web/components/feature-page/FeatureFilters.tsx
```diff
@@ -0,0 +1,221 @@
...
+  sort: sortToHeader(f.sort),
+  tags: f.tags.length ? f.tags.join(',') : undefined,
+})
+
+//Detect if the filter is default
+const isDefault = (v: FiltersValue) =>
+  isEqual(
+    {
+      ...v,
+    },
```
Zaimwa9: Do we need to spread? isEqual seems sufficient no ? 
```suggestion
v,
```

Zaimwa9 (changes requested):
Overall good but minor comments and questions

frontend/web/components/feature-page/FeatureFilters.tsx
kyle-ssg: I have no idea what it is :D , I'll remove. I guess maybe the idea was to add a filter for particular pipelines eventually

frontend/web/components/feature-page/FeatureFilters.tsx
kyle-ssg: Originally to force preserving default sort but it might be redundant now actually so good catch

frontend/web/components/feature-page/FeatureFilters.tsx
kyle-ssg: Hmm will separate since it's not related to features I think

frontend/web/components/feature-page/FeatureFilters.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+  return {
+    ...existing,
+    group_owners: f.group_owners?.length ? f.group_owners.join(',') : undefined,
+    is_archived: f.is_archived ? 'true' : undefined,
+    is_enabled:
+      f.is_enabled === null ? undefined : f.is_enabled ? 'true' : 'false',
+    owners: f.owners?.length ? f.owners.join(',') : undefined,
+    page: f.page !== DEFAULTS.page ? String(f.page) : undefined,
+    search: f.search || undefined,
```
Zaimwa9, line 62: I think we are missing sort here while we set it in the other functions

frontend/web/components/feature-page/FeatureFilters.tsx
```diff
@@ -0,0 +1,202 @@
...
+    tag_strategy:
+      f.tag_strategy !== DEFAULTS.tag_strategy
+        ? String(f.tag_strategy)
+        : undefined,
+    tags: f.tags?.length ? f.tags.join(',') : undefined,
+    value_search: f.value_search || undefined,
+  }
+}
+// Gets expected filters from URL parameters
+export const getFiltersFromURLParams = (
```
Zaimwa9: NIT: The naming is actually good but my brain can't make it ahah. Do you think it could make sense to call this one `parseFiltersFromUrlParams` ?

frontend/web/components/tables/TableValueFilter.tsx
```diff
@@ -34,8 +34,15 @@ const TableTagFilter: FC<TableFilterType> = ({
   onChange,
   value,
 }) => {
-  const { searchInput, setSearchInput } = useDebouncedSearch('')
-
+  const { search, searchInput, setSearchInput } = useDebouncedSearch('')
```
Zaimwa9: we might want to initialize it with the value from params?

Zaimwa9 (approved):
Quite minors one. I'll approve in case not using sort from url is by design otherwise this change and good to go.
Other are minors and ignorable

frontend/web/components/feature-page/FeatureFilters.tsx
kyle-ssg: Yep you're right, didn't persist

PR #6132, opened 2025-10-07
chore: Remove stale flags

kyle-ssg:

## Changes

image

Trivial

### DIFF
--- a/frontend/web/components/pages/FeaturesPage.js
+++ b/frontend/web/components/pages/FeaturesPage.js
@@ -227,6 +227,9 @@ const FeaturesPage = class extends Component {
   render() {
     const { environmentId, projectId } = this.props.match.params
     const readOnly = Utils.getFlagsmithHasFeature('read_only_mode')
+    const environmentMetricsEnabled = Utils.getFlagsmithHasFeature(
+      'environment_metrics',
+    )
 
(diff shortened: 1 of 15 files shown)


frontend/web/components/pages/FeaturesPage.js
```diff
@@ -313,13 +310,11 @@ const FeaturesPage = class extends Component {
                             'features',
                             featureLimitAlert.percentage,
                           )}
-                        {environmentMetricsEnabled && (
-                          <EnvironmentMetricsList
-                            environmentApiKey={environment?.api_key}
-                            forceRefetch={this.state.forceMetricsRefetch}
-                            projectId={projectId}
-                          />
-                        )}
```
Zaimwa9, line 322: Just this one if you don't mind I think it's worth keeping it for now as a kill switch. CF the issue about the permissions thunk query.

The thing is that we have never gone through the V2 which is supposed to add some caching and optimization. The identity part might be really heavy so it can be handful to deactivate if it gets out of control for a heavy org

Zaimwa9 (changes requested):
Satisfying. Just one flag to keep in case I believe if you agree

Zaimwa9 (approved):
👍

PR #6134, opened 2025-10-07
feat: create endpoint to get a flag multivariate options

Zaimwa9:

## Changes
To meet Vercel requirements, we need to provide the list of MV values in order for their Flag Explorer to list them.
It is using the SDK under the hood so it is not possible to use private keys to GET the same endpoint as the platform.
Hence why the creation of an endpoint specifically for this case

- Created a GET endpoint `/api/v1/flags/{featureID}/multivariate-options/` with environment key permission level to get the list of multivariate values

Response as follows, voluntarily sending the bare minimum:

```{
    "control_value": "test",
    "options": [
        {
            "value": 1
        },
        {
            "value": "yolotest"
        },
        {
            "value": "ortest"
        },
        {
            "value": "plop"
        }
    ]
}
```

Will be used as follows in vercel package (yet to be changed):
```
  if (flag.feature.type === 'MULTIVARIATE') {
        try {
          const mvResponse = await fetch(
            `https://api.flagsmith.com/api/v1/flags/${flag.feature.id}/multivariate-options/`,
            {
              method: 'GET',
              headers: { 'X-Environment-Key': options.environmentKey },
            },
          );
          if (mvResponse.ok) {
            const body =
              (await mvResponse.json()) as FlagsmithMultivariateApiData;
            const { control_value, options } = body;
```

- Added tests
- Locally 
image

### DIFF
--- a/api/features/views.py
+++ b/api/features/views.py
@@ -28,10 +28,6 @@ from rest_framework.renderers import JSONRenderer
 from rest_framework.request import Request
 from rest_framework.response import Response
 
-from .multivariate.serializers import (
-    FeatureMVOptionsValuesResponseSerializer,
-    NestedMultivariateFeatureOptionSerializer,
-)
 from app.pagination import CustomPagination
 from app_analytics.analytics_db_service import get_feature_evaluation_data
 from app_analytics.influxdb_wrapper import get_multiple_event_list_for_feature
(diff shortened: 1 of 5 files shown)


api/features/views.py
```diff
@@ -907,6 +915,43 @@ def _get_flags_response_with_identifier(
...
+            environment=environment,
+            additional_filters=Q(
+                identity__isnull=True,
+                feature_segment__isnull=True,
+                feature_id=feature.id,
+            ),
+        )
+        .filter(feature__is_archived=False)
+        .select_related("feature_state_value")
+        .order_by("-environment_feature_version__created_at")  # latest published
```
gagantrivedi: Doesn't `get_live_feature_states` already take care of this?

gagantrivedi (approved):

PR #6137, opened 2025-10-07
feat: segment feature state view

kyle-ssg:

## Changes

Places the associated features page with the following

image

Opening a feature from this view simply opens the feature modal with the segment overrides tab selected.

Requires https://github.com/Flagsmith/flagsmith/pull/6156

- Create some segment overrides
- View this page and validate that segment overrides are highlighted


frontend/common/types/requests.ts
```diff
@@ -85,7 +87,7 @@ export type RegisterRequest = {
   marketing_consent_given?: boolean
   utm_data?: UtmsType
 }
-
+export type SortOrder = 'ASC' | 'DESC'
```
Zaimwa9: I missed it in a previous release. It believe lodash still accepts it but in doubt maybe we can lowercase
image

frontend/web/components/modals/CreateSegment.tsx
```diff
@@ -482,18 +482,25 @@ const CreateSegment: FC<CreateSegmentType> = ({
...
-        <Tabs value={tab} onChange={(tab: UserTabs) => setTab(tab)}>
+        <Tabs
+          value={tab}
+          theme='pill'
+          urlParam='segmentTab'
+          onChange={(tab: UserTabs) => setTab(tab)}
+        >
           <TabItem
-            tabLabelString='Rules'
+            tabLabelString='General'
```
Zaimwa9, line 498: If we can just fix the edit star 
image

frontend/web/components/segments/AssociatedSegmentOverrides.tsx
```diff
@@ -0,0 +1,185 @@
...
+    <>
+      <div>
+        <PageTitle
+          title={
+            <div className='d-flex gap-4'>
+              Associated Features
+              <div style={{ width: 200 }}>
+                <EnvironmentSelect
+                  showAll={false}
+                  value={environment?.id}
```
Zaimwa9: `toString()` or accept number as value

frontend/web/components/segments/AssociatedSegmentOverrides.tsx
```diff
@@ -0,0 +1,185 @@
...
+                  projectId={projectId}
+                  onChange={(id, environment) => setEnvironment(environment)}
```
Zaimwa9: NIT: 
I find it a bit weird like this. It mixes a bit the responsibilities of the dropdown I feel.

We have the environments above, maybe we could leave the options and `onChange` signatures untouched, find the env here and set it here from the value ?

frontend/common/types/requests.ts
```diff
@@ -321,9 +323,20 @@ export type Req = {
   }
   getProjectFlags: {
     project: string
-    environmentId?: string
-    tags?: string[]
+    environment?: number
```
Zaimwa9, line 348: Just to double check all the environmentId have been replaced

frontend/web/components/segments/AssociatedSegmentOverrides.tsx
```diff
@@ -0,0 +1,185 @@
...
+  const { data: environments } = useGetEnvironmentsQuery(
+    {
+      projectId: `${projectId}`,
+    },
+    {
+      skip: !projectId,
+    },
+  )
+
+  const isLoading = !projectFlags || !environment
```
Zaimwa9: We can use projectFlagsLoading

frontend/web/components/pages/SegmentPage.tsx
```diff
@@ -14,6 +14,10 @@ import Button from 'components/base/forms/Button'
 import { useHasPermission } from 'common/providers/Permission'
 import Icon from 'components/Icon'
 import { handleRemoveSegment } from 'components/modals/ConfirmRemoveSegment'
+import { IonIcon } from '@ionic/react'
```
Zaimwa9: I noticed a couple of unused imports we could clean 
image

frontend/web/components/Breadcrumb.tsx
```diff
@@ -8,24 +8,27 @@ type BreadcrumbType = {
...
+    <div className='d-flex align-items-center my-2 py-1'>
+      {items?.map((item) => (
+        <>
+          <Link className='text-primary h6 mb-0' to={item.url}>
+            {item.title}
+          </Link>
+          <div className='text-muted mx-2 h6 mb-0'>/</div>
+        </>
+      ))}
+      {typeof currentPage === 'string' ? (
```
Zaimwa9: We could change the type of currentPage to be `ReactNode` and add a new boolean prop `isCurrentPageMuted` ?

Zaimwa9 (changes requested):
Couple of early comments. The page is really nice to use though and it's satisfying to see the past PRs getting into action. I'll leave @talissoncosta give a review too

frontend/web/components/tables/TableSortFilter.tsx
```diff
@@ -28,7 +29,7 @@ const TableSortFilter: FC<TableFilterType> = ({
   const value = _value || {
     label: options[0].label,
```
talissoncosta, line 30: Should we consider the case where options is undefined and add an early return ? 
```tsx
if (!options?.length) return null;
```

frontend/web/components/segments/AssociatedSegmentOverrides.tsx
```diff
@@ -0,0 +1,185 @@
+import React, { FC, useEffect, useState } from 'react'
+import { useGetSegmentFeatureStatesQuery } from 'common/services/useFeatureState'
```
talissoncosta: It seems not being used

frontend/web/components/segments/AssociatedSegmentOverrides.tsx
talissoncosta: Agreed. We're currently fetching environments twice - once in the parent (line 62) and again inside `EnvironmentSelect`. 

Making `EnvironmentSelect` a pure UI component that accepts `environments` as a prop would:
- Eliminate duplicate API calls
- Improve testability and reusability
- Provide clearer separation of concerns

We already have the environments data here, so we should just pass it down instead of letting `EnvironmentSelect` re-fetch it.

frontend/web/components/tables/TableSortFilter.tsx
```diff
@@ -28,7 +29,7 @@ const TableSortFilter: FC<TableFilterType> = ({
...
     sortBy: options[0].value,
-    sortOrder: 'asc',
+    sortOrder: 'ASC',
```
talissoncosta: What about using an enum to handle sort values ?

frontend/web/components/pages/SegmentPage.tsx
```diff
@@ -51,34 +55,42 @@ const SegmentPage: FC<SegmentPageType> = ({}) => {
...
+                  history.replace(
+                    `/project/${projectId}/segments/${v.value}?${Utils.toParam(
+                      Utils.fromParam(),
+                    )}`,
+                  )
+                }
+                className='react-select select-xsm'
+                value={parseInt(id)}
+                projectId={projectId}
+              />
```
talissoncosta, line 67: Same as `EnvironmentSelect` below. We might consider making `SegmentSelect` a pure presentation component

talissoncosta (approved):

talissoncosta:
It looks good overall. I just have a few additional suggestions to complement @Zaimwa9 ’s review.

frontend/common/types/requests.ts
kyle-ssg: Yeah, environmentId was never correct

frontend/web/components/segments/AssociatedSegmentOverrides.tsx
kyle-ssg: I think the right approach might be to just create a connected version within EnvironmentSelect, if we're always requiring data to be fetched in order to use this then we're not really simplifying anything by componentising in fact we're reducing reusability if anything. It'd only result in duplicate api calls as a result of not using RTK yet.

We're using this in a lot of places - I'd rather change this as a separate PR.

frontend/web/components/tables/TableSortFilter.tsx
kyle-ssg: Hmm I can't imagine a case where options gets anything but a static value. That'd require us starting to implement backend managed sort options which I can't see happening?

frontend/web/components/pages/SegmentPage.tsx
kyle-ssg: Hmm again I think the solution might be to create a connected version of the component, in it's current form it's simple to use since you don't care about the logic of fetching segments.

frontend/web/components/tables/TableSortFilter.tsx
talissoncosta: Yep, I agree. I was just thinking about potential edge cases.

talissoncosta (approved):

matthewelwell (changes requested):
I've tested this and unfortunately it doesn't work. See the screen recording below. It looks like when you try to update a feature for a given segment, it's making a call to update it for an identity. 

I would expect it to add a segment override for this segment (and include that information in the confirmation modal). 

https://github.com/user-attachments/assets/caf8aea8-3724-4f05-ad35-a3a68c0841dd

kyle-ssg:
Ah, ok yes this didn't consider toggling the switch directly vs opening the modal. I think for simplicity sake clicking any part of the row should just open up the segment override tab. I've updated this.

We could have that automatically add an override to the list though I'm not sure it's a good idea, creating an override at the point where you can see others / set priority is more simple and makes sense. When I use Flagsmith even on fairly simple projects I'd never want to add a segment override without seeing the full context (other segment overrides for this feature).

frontend/web/components/feature-override/FeatureOverrideRow.tsx
```diff
@@ -74,6 +74,10 @@ const FeatureOverrideRow: FC<FeatureOverrideRowProps> = ({
   const actualValue = overrideFeatureState?.feature_state_value ?? flagValue
 
   const onToggle = () => {
+    if (level === 'segment') {
+      onClick()
+      return
+    }
     const confirmToggle = (projectFlag: any, environmentFlag: any, cb: any) => {
```
Zaimwa9: Can we extract this definition form inside the function? Or maybe just return without calling onClick and removing the `stopPropagation` L269 to just let the row deal with the click event?

Zaimwa9 (changes requested):
Sorry, small change

matthewelwell:
> Ah, ok yes this didn't consider toggling the switch directly vs opening the modal. I think for simplicity sake clicking any part of the row should just open up the segment override tab. I've updated this.
> 
> We could have that automatically add an override to the list though I'm not sure it's a good idea, creating an override at the point where you can see others / set priority is more simple and makes sense. When I use Flagsmith even on fairly simple projects I'd never want to add a segment override without seeing the full context (other segment overrides for this feature).

Yeah, I guess this makes sense with the current functionality. It is quite annoying to have to find the segment in the list still after clicking the row though. I wonder if there's an option here to add an extra step to creating a segment override which adds a confirmation button next to the drop down in the modal, then we could pre-fill the drop down when you click the row? Maybe that's overkill, but I do feel that the UX here is a little janky.

kyle-ssg:
> Yeah, I guess this makes sense with the current functionality. It is quite annoying to have to find the segment in the list still after clicking the row though. I wonder if there's an option here to add an extra step to creating a segment override which adds a confirmation button next to the drop down in the modal, then we could pre-fill the drop down when you click the row? Maybe that's overkill, but I do feel that the UX here is a little janky.


I don't think clicking the row should do anything automatically around adding a new override, a lot of the times people are simply wanting to view segment overrides, adding a segment override is a much less frequent action.

We could highlight existing overrides for that segment in the list with a primary border, and preselect the segment in the dropdown with a button to add an override.


## Clicking a feature when the segment override exists

Highlights override so you can see it 

image


## Clicking a feature when no segment override exists

Preselects segment in dropdown with a button to add it

image

talissoncosta (approved):
Looks good to me on the engineering side, great work!

matthewelwell (approved):
Tested manually. 

Agreed with Kyle that this is a good compromise. I still think the UX could be improved, but I'm not really sure how to do that without a larger overhaul of the segment overrides UX in the modal.

PR #6147, opened 2025-10-12
fix: Lighten darker tag colors in dark mode

### LINKED ISSUE #6128: Fix tag colors in dark mode
Issue location: `/organisation/<org_id>/permissions?type=roles`

The problem is observed when in dark mode.

When associating tags to permissions in the "Edit Role and Permisions" dialog, the label of the tag is barely readable because of the low amount of contrast between the font color and the background color.

See demonstration:

Tag label 'ABCD' is barely readable

Disclosure: the label "ABCD" in the screenshot above was artificially added in order to redact real information, hence looking oddly sharp compared to the image resolution. Reality looks worse.

existentialcoder:

## Changes
- For darker tag colors, lighten it by 50% so that its visible
- This will fix #6128 

Local testing. Added screenshots before and after

- `Before`

image

- `After`
after-fix-tag-ui

### DIFF
--- a/frontend/web/components/tags/Tag.tsx
+++ b/frontend/web/components/tags/Tag.tsx
@@ -7,6 +7,7 @@ import Utils from 'common/utils/utils'
 import TagContent from './TagContent'
 import Constants from 'common/constants'
 import { getDarkMode } from 'project/darkMode'
+import Color from 'color'
 
 type TagType = {
   className?: string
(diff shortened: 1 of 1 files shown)


Zaimwa9 (approved):
Thanks for the contribution!
Discarding the cursor comment to keep dot mode iso as it is out of the initial scope

PR #6150, opened 2025-10-14
feat: Improve flag analytics

kyle-ssg:

## Changes

- Resurrects https://github.com/Flagsmith/flagsmith/pull/5100 and https://github.com/Flagsmith/flagsmith/pull/5096.
- Neatens up chip component

Provides a multi-environment view of analytics. After reviewing comments for both, this feels like the right decision. I fundamentally agree that everything to do with a feature should be accessible by clicking from a feature.

image

Viewed production feature flags.


frontend/web/components/ToggleChip.tsx
```diff
@@ -0,0 +1,51 @@
...
+}
+
+const ToggleChip: FC<ToggleChipProps> = ({
+  active,
+  children,
+  className,
+  color,
+  onClick,
+}) => {
+  const colour = Utils.colour(color)
```
Zaimwa9, line 21: NIT: We could make c?:string to match with color being undefined or guard here `colour = color ? ... : null` if we want to be type strict.
```
  colour(
    c: string,
    fallback = Constants.defaultTagColor,
  ): InstanceType<typeof Color> {
    let res: Color
    try {
      res = Color(c)
    } catch (_) {
      res = Color(fallback)
    }
    return res
  },
  ```

File: frontend/web/components/EnvironmentTagSelect.tsx [hunk 2]
```diff
@@ -0,0 +1,123 @@
+import React, { FC, useMemo } from 'react'
+import { useGetEnvironmentsQuery } from 'common/services/useEnvironment'
+import { Props } from 'react-select/lib/Select'
+import Tag from './tags/Tag'
+import Utils from 'common/utils/utils'
+import Button from './base/forms/Button'
+
+export type EnvironmentSelectType = Omit<
```
Zaimwa9: I think this is a bit weird and silence some type errors (like the ignore should be spotted), is it working with an interface?

frontend/web/components/feature-page/FeatureNavTab/FeatureAnalytics.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const handleEnvironmentChange = (value: string[] | string | undefined) => {
+    // If cleared (empty array or undefined), revert to default environment IDs
+    if (!value || (Array.isArray(value) && value.length === 0)) {
+      setEnvironmentIds(defaultEnvironmentIds)
+    } else {
+      setEnvironmentIds(value as string[])
+    }
+  }
+
+  return (
```
Zaimwa9, line 64: Do we want to show an additional message somewhere if data is empty in the last 30 days?
I'm really fine with leaving it like this though
image

frontend/web/components/EnvironmentTagSelect.tsx
```diff
@@ -0,0 +1,123 @@
...
+        <div
+          className='flex-shrink-0 ml-2 text-right'
+          style={{ width: '140px' }}
+        >
+          {!allSelected && (
+            <Button
+              onClick={handleSelectAll}
+              size='xSmall'
+              theme='text'
+              className='mr-2'
```
Zaimwa9, line 101:
```suggestion
```
It adds a little jumping when switching to clear

Zaimwa9: Aah ok I see the 2 of them can be displayed. Then maybe use a gap in the parent div?

frontend/common/types/responses.ts
```diff
@@ -1109,5 +1109,9 @@ export type Res = {
   releasePipeline: SingleReleasePipeline
   pipelineStages: PagedResponse<PipelineStage>
   featureCodeReferences: FeatureCodeReferences[]
+  featureAnalytics: {
+    day: string
+    [environmentId: string]: string | number
```
Zaimwa9: I guess this should just be a number?

frontend/common/services/useFeatureAnalytics.ts
```diff
@@ -0,0 +1,90 @@
+import { Res } from 'common/types/responses'
+import { Req } from 'common/types/requests'
+import { service } from 'common/service'
+import { sortBy } from 'lodash'
+import moment from 'moment'
+import range from 'lodash/range'
```
Zaimwa9:
```suggestion
```

frontend/common/services/useFeatureAnalytics.ts
```diff
@@ -0,0 +1,90 @@
+import { Res } from 'common/types/responses'
+import { Req } from 'common/types/requests'
+import { service } from 'common/service'
+import { sortBy } from 'lodash'
```
Zaimwa9:
```suggestion
```

frontend/common/services/useFeatureAnalytics.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export const featureAnalyticsService = service
+  .enhanceEndpoints({ addTagTypes: ['FeatureAnalytics'] })
+  .injectEndpoints({
+    endpoints: (builder) => ({
+      getFeatureAnalytics: builder.query<
+        Res['featureAnalytics'],
+        Req['getFeatureAnalytics']
+      >({
+        providesTags: [{ id: 'LIST', type: 'FeatureAnalytics' }],
```
Zaimwa9, line 28: As discussed offline, for each combination of environments, the current state of the hook re-triggers a request for each of the environment while ideally we would cache each environment once and for all

Zaimwa9 (changes requested):
Thanks, really good improvements. A couple of minor comments, mostly on typing and let's create an issue for the hook future improvement

frontend/web/components/feature-page/FeatureNavTab/FeatureAnalytics.tsx
kyle-ssg: Sure I'll use this PR as an opportunity to do that

kyle-ssg: image

frontend/common/services/useFeatureAnalytics.ts
```diff
@@ -1,29 +1,48 @@
...
 import { service } from 'common/service'
-import { sortBy } from 'lodash'
 import moment from 'moment'
-import range from 'lodash/range'
 
 export const featureAnalyticsService = service
   .enhanceEndpoints({ addTagTypes: ['FeatureAnalytics'] })
   .injectEndpoints({
     endpoints: (builder) => ({
+      getEnvironmentAnalytics: builder.query<
```
Zaimwa9, line 10: Ah nice!!

Zaimwa9 (approved):
Looks really good, thanks for the effort on the hook

PR #6152, opened 2025-10-14
chore(evaluation-context): Add `FeatureValue.priority`, loosen metadata schema

khvn26:

## Changes

In this PR, we add a `FeatureValue.priority` schema to surface that the order of the variants is important to the engine, and force the engine users to explicitly assign the sorting order to their multivariate feature variants.

Additionally, we loosen the metadata schema to improve code generation for strictly typed languages.

N/A

### DIFF
--- a/sdk/evaluation-context.json
+++ b/sdk/evaluation-context.json
@@ -83,6 +83,11 @@
         "FeatureValue": {
             "description": "Represents a multivariate value for a feature flag.",
             "properties": {
+                "key": {
+                    "description": "Sorting key, used to determine the variant order. Set to an internal feature value state identifier or a UUID, depending on Flagsmith implementation.",
+                    "title": "Key",
+                    "type": "string"
+                },
                 "value": {
                     "description": "The value of the feature.",
                     "title": "Value",
@@ -100,6 +105,7 @@
                 }
             },
             "required": [
+                "key",
                 "value",
                 "weight"
             ],
@@ -173,15 +179,6 @@
                 },
                 "metadata": {
                     "description": "Additional metadata associated with the segment.",
-                    "additionalProperties": {
-                        "type": [
-                            "string",
-                            "number",
-                            "boolean",
-                            "null"
-                        ]
-                    },
-                    "title": "Metadata",
                     "type": "object"
                 }
             },
--- a/sdk/evaluation-context.json
+++ b/sdk/evaluation-context.json
@@ -84,7 +84,7 @@
             "description": "Represents a multivariate value for a feature flag.",
             "properties": {
                 "key": {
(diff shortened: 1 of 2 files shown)


sdk/evaluation-context.json
```diff
@@ -172,16 +178,8 @@
                     "type": "array"
                 },
                 "metadata": {
+                    "existingJavaType": "java.util.Map<String,Object>",
```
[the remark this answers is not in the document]
khvn26: Fixed in 2c37767df67b509fbdafd5f652a47da27a3a3e25.

sdk/evaluation-context.json
```diff
@@ -83,6 +83,11 @@
         "FeatureValue": {
             "description": "Represents a multivariate value for a feature flag.",
             "properties": {
+                "key": {
+                    "description": "Sorting key, used to determine the variant order. Set to an internal feature value state identifier or a UUID, depending on Flagsmith implementation.",
+                    "title": "Key",
+                    "type": "string"
```
emyller: Should we use an integer instead?

sdk/evaluation-context.json
```diff
@@ -83,6 +83,11 @@
         "FeatureValue": {
             "description": "Represents a multivariate value for a feature flag.",
             "properties": {
+                "key": {
+                    "description": "Sorting key, used to determine the variant order. Set to an internal feature value state identifier or a UUID, depending on Flagsmith implementation.",
```
emyller:
```suggestion
                    "description": "Sorting key, used to calculate variants consistently.",
```

Emphasizes how critical this field is.

The remaining part below may not be relevant to the engine... Not sure if it should be in the spec?

> Set to an internal feature value state identifier or a UUID, depending on Flagsmith implementation.

emyller (changes requested):

sdk/evaluation-context.json
khvn26: I believe we're safe to use string as UUIDs are sorted by string value:

```python
import uuid

uuids = [uuid.uuid4() for _ in range(5)]

assert [str(v) for v in sorted(uuids)] == sorted(str(v) for v in uuids)
```

If you insist we should, I suggest we rename it to `priority`.

emyller: I'd be fine with sorting UUIDs, though I think we need to keep sorting by numeric IDs in order to keep the same behavior of the engine.

```python
sorted(['1', '10', '2']) != [1, 2, 10]
```

...then there is SaaS.

khvn26: Fixed in c294e251d968438aa0bce5f31096ce40e352e694.

emyller (approved):

PR #6155, opened 2025-10-15
chore: added-claude-contexts

Zaimwa9:

## Changes
Related to #6153 

- Added claude contexts for backend as a follow-up from frontend one

### DIFF
--- a/api/.claude/context/api-structure.md
+++ b/api/.claude/context/api-structure.md
@@ -1,0 +1,157 @@
+# API Structure
+
+## URL Patterns
+
+### Features
+- **Project Features**: `/api/v1/projects/{project_pk}/features/`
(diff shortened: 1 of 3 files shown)


api/.claude/context/api-structure.md
kyle-ssg, line 1: If we're adding this, the frontend claude context should point to this as the source of truth. This makes me think this should all be 1 PR with a claude.md in api / frontend.

The backend and frontend contexts maybe shouldn't be in the same folder since I imagine it could cause increased token usage. However, I think the FE should point to files like this where applicable.

Zaimwa9:
Test new thread

khvn26:
Testing after having completed slack & github oauth flows [](edited-with-Axolo)

PR #6156, opened 2025-10-15
feat: added-query-param-to-get-segment-feature-states

Zaimwa9:

## Changes
- `GET `/api/v1/projects/{project_id}/features/?environment={env_id}&segment={segment_id}` endpoint now accept query-string params `?segment=xx` to return the requested segment' feature states

Added `segment_feature_state` to response when `segment=xx` is passed in. Without or with invalid segment, the response is untouched

From V1 to V2:
Frontend was initially going to get the features list and reconciliate the feature states with it. However we realized that this could lead to some loopholes due to features list pagination.
V2 includes a new query param in order to attach the `segment_feature_state` to the features list to avoid pagination discrepancies.

- Added tests
- Ran query

Example payload:
```
{
            "id": 21,
             ....
            "uuid": "1a56cd3c-611a-407f-9dd3-13a10580c7ca",
            "project": 3,
            "environment_feature_state": {
                "id": 565,
                "feature_state_value": null,
                "environment": 8,
                "identity": null,
                "feature_segment": null,
                "enabled": false
            },
            "segment_feature_state": null,
            "num_segment_overrides": 0,
             ....
            "metadata": [],
            "code_references_counts": []
        },
```


gagantrivedi (dismissed):

Zaimwa9:
Couple of changes @gagantrivedi 🙏 :
From V1 to V2:
Frontend was initially going to get the features list and reconciliate the feature states with it. However we realized that this could lead to some loopholes due to features list pagination.
V2 includes a new query param in order to attach the segment_feature_state to the features list to avoid pagination discrepancies.

kyle-ssg:
This simplifies the FE so much! Tested with the code in https://github.com/Flagsmith/flagsmith/pull/6137

api/features/views.py
```diff
@@ -663,7 +682,8 @@ class EnvironmentFeatureStateViewSet(BaseFeatureStateViewSet):
     permission_classes = [EnvironmentFeatureStatePermissions]
 
     def get_queryset(self):  # type: ignore[no-untyped-def]
-        queryset = super().get_queryset().filter(feature_segment=None)  # type: ignore[no-untyped-call]
+        queryset = super().get_queryset()  # type: ignore[no-untyped-call]
+
```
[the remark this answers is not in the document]
gagantrivedi: @Zaimwa9 this sounds valid?

api/features/views.py
```diff
@@ -182,18 +182,33 @@ def get_queryset(self):  # type: ignore[no-untyped-def]
...
             )
             feature_states = get_environment_flags_list(
                 environment=self.environment,
-                additional_filters=q,
+                additional_filters=feature_states_query,
                 additional_select_related_args=["feature_state_value", "feature"],
             )
             self._feature_states = {fs.feature_id: fs for fs in feature_states}
 
+            if segment_id := query_data.get("segment"):
```
gagantrivedi, line 197: This API is already doing way too much. Have we thought about maybe moving this to a new endpoint?

Zaimwa9: Yes, (partially) agreeing with you. I was thinking the same but in a way we need this segment logic to be related to the feature listing one to share the pagination. That's why the first version couldn't work for customers with > 50 features.
To give some more context. This parameter is only used in the segment page, and we control it. So this does not impact at all . We could definitely separate into another endpoint but it would repeat a big chunk of this logic.
Wdyt ?

gagantrivedi: Ugh, true. I don't think I have any other solution apart from a major refactor, but that's for another day.

gagantrivedi (dismissed):

api/features/views.py
```diff
@@ -225,9 +240,13 @@ def get_serializer_context(self):  # type: ignore[no-untyped-def]
...
 
         feature_states = getattr(self, "_feature_states", {})
+        segment_feature_states = getattr(self, "_segment_feature_states", {})
         project = get_object_or_404(Project.objects.all(), pk=self.kwargs["project_pk"])
         context.update(
-            project=project, user=self.request.user, feature_states=feature_states
+            project=project,
+            user=self.request.user,
+            feature_states=feature_states,
+            segment_feature_states=segment_feature_states,
```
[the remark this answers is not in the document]
Zaimwa9: Should not happen and legacy

api/features/views.py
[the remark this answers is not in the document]
Zaimwa9: It was valid, I didn't properly clean-up the first version

api/tests/unit/features/test_unit_features_views.py
```diff
@@ -4077,6 +4075,66 @@ def test_get_multivariate_options_feature_not_found_responds_404(
...
+    if expected_feature_state_value is None:
+        assert feature_data["segment_feature_state"] is None
+    else:
+        assert feature_data["segment_feature_state"] is not None
+        assert feature_data["segment_feature_state"]["id"] == segment_override.id
+        assert (
+            feature_data["segment_feature_state"]["feature_state_value"]
+            == expected_feature_state_value
+        )
+        assert feature_data["segment_feature_state"]["enabled"] is True
```
khvn26: I'd like to avoid branching in test bodies, especially in the assert phase. Can we parametrise `segment_feature_state` instead?

Zaimwa9: I ended up separating the tests. Either I ended up moving the branching in another place or hardcoding an id that is dependent on fixtures. It didn't feel right

khvn26: Works for me, thanks!

khvn26 (approved):

PR #6169, opened 2025-10-16
fix: menu tab highlight disappear for other Environments sub-options

### LINKED ISSUE #6162: Menu highlight bar dissappears when not looking at Features list

- [ ] Self Hosted with Docker
- [x] SaaS at flagsmith.com

### Describe the bug

As per title

### Steps To Reproduce

1. Click Environments. 
2. Click Features. 
3. Notice that the `Environments` nav heading is underlined in purple
4. Click Scheduling
5. Notice that the `Environments` nav heading is now NOT underlined in purple

### Expected behavior

The purple underline should remain for all sub-nav element pages

### Screenshots

Image

Image

existentialcoder:

## Changes

Fixes #6162 

- Ideally, the routes should be grouped under environment [here](https://github.com/Flagsmith/flagsmith/blob/main/frontend/web/routes.js#L72)
- The default `isActive` behavior of `NavLink` is the root cause of this as it matches the exact path
- Made sure all the sidebar options of Environments don't exactly match but starts with environments path to make sure tab highlighting is consistent

Manual testing

1.
image

2.
image

### DIFF
--- a/frontend/web/components/navigation/NavSubLink.tsx
+++ b/frontend/web/components/navigation/NavSubLink.tsx
@@ -1,17 +1,23 @@
 import React, { FC, ReactNode } from 'react'
-import { LinkProps, NavLink } from 'react-router-dom'
+import type * as H from 'history';
+import { LinkProps, NavLink, match } from 'react-router-dom'
 import { IonIcon } from '@ionic/react'
 import classNames from 'classnames'
 
 type NavSubLinkType = LinkProps & {
   icon: string | ReactNode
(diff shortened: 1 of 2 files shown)


dabeeeenster:
Awesome thanks!

Zaimwa9 (approved):
Thanks for the contribution. It will help. Even overall this would need to properly refactor the 2 navigations that are conflicting at the moment due to one using subpathes and the other one (this one) being linked to the "naked" environment path

PR #6181, opened 2025-10-21
chore: api type sync

kyle-ssg:

## Changes

Output of the Claude api sync, this validates the current RTK types against Flagsmith's API and builds a cache.

- Validated updated types


khvn26:
I understand and appreciate how this improves the developer experience for our FE engineers and contributors.

My only issue with this, albeit a significant one, is that it undermines Flagsmith's "API-first" approach by causing the FE to completely ignore the OpenAPI schema. Currently, our schema is already unusable for code generation (#6183), and I'm afraid bypassing it for its main consumer, the Flagsmith UI, will cause it to deteriorate further, which goes against our long-term goals.

While I appreciate the relatively quick gain introduced by this PR, I believe that investing in #6183 and relying on the currently available industry-standard codegen tools would benefit a wider audience of contributors and product users in the long run.

kyle-ssg:
> I understand and appreciate how this improves the developer experience for our FE engineers and contributors.
> 
> My only issue with this, albeit a significant one, is that it undermines Flagsmith's "API-first" approach by causing the FE to completely ignore the OpenAPI schema. Currently, our schema is already unusable for code generation (#6183), and I'm afraid bypassing it for its main consumer, the Flagsmith UI, will cause it to deteriorate further, which goes against our long-term goals.
> 
> While I appreciate the relatively quick gain introduced by this PR, I believe that investing in #6183 and relying on the currently available industry-standard codegen tools would benefit a wider audience of contributors and product users in the long run.

Sure ok, there's no urgency on this so I agree. If the OpenAPI Schema can be an autogenerated reliable source of truth that'll be way better.

PR #6196, opened 2025-10-23
fix: Use data instead of parsed data for recording analytics

existentialcoder:

## Changes
- [This method](https://github.com/encode/django-rest-framework/blob/e045dc465270c18689dba4a970378cd9744e57b6/rest_framework/serializers.py#L345) of django does not parse the feature keys with `.` righly (Refer #5906 for the exact error)
- This will fix the same.
- Used `initial_data` instead of `validated_data` which has the wrong parsed format

_Please describe._

Manual API testing

### DIFF
--- a/api/app_analytics/views.py
+++ b/api/app_analytics/views.py
@@ -88,7 +88,9 @@ class SDKAnalyticsFlags(CreateAPIView):  # type: ignore[type-arg]
 
             def save(self, **kwargs: typing.Any) -> None:
                 request = self.context["request"]
-                for feature_name, evaluation_count in self.validated_data.items():
+                # validated_data splits out request body with '.' in feature name (e.g a.b.c).
+                # Instead, it's safe to use self.initial_data as keys are not altered.
+                for feature_name, evaluation_count in self.initial_data.items():
                     feature_evaluation_cache.track_feature_evaluation(
                         environment_id=request.environment.id,
                         feature_name=feature_name,
--- a/api/app_analytics/views.py
+++ b/api/app_analytics/views.py
@@ -89,8 +89,8 @@ class SDKAnalyticsFlags(CreateAPIView):  # type: ignore[type-arg]
             def save(self, **kwargs: typing.Any) -> None:
                 request = self.context["request"]
                 # validated_data splits out request body with '.' in feature name (e.g a.b.c).
-                # Instead, it's safe to use self.initial_data as keys are not altered.
-                for feature_name, evaluation_count in self.initial_data.items():
+                # Instead, it's safe to use self.data as keys are not altered.
+                for feature_name, evaluation_count in self.data.items():
                     feature_evaluation_cache.track_feature_evaluation(
                         environment_id=request.environment.id,
                         feature_name=feature_name,
--- a/api/conftest.py
+++ b/api/conftest.py
@@ -555,6 +555,10 @@ def api_client() -> APIClient:
 def feature(project: Project) -> Feature:
     return Feature.objects.create(name="Test Feature1", project=project)  # type: ignore[no-any-return]
 
+@pytest.fixture()
+def feauture_with_dots(project: Project) -> Feature:
+    return Feature.objects.create(name="feature.name", project=project)  # type: ignore[no-any-return]
+
 
 @pytest.fixture()
 def change_request(environment: Environment, admin_user: FFAdminUser) -> ChangeRequest:
--- a/api/conftest.py
+++ b/api/conftest.py
@@ -555,10 +555,6 @@ def api_client() -> APIClient:
 def feature(project: Project) -> Feature:
     return Feature.objects.create(name="Test Feature1", project=project)  # type: ignore[no-any-return]
 
-@pytest.fixture()
-def feauture_with_dots(project: Project) -> Feature:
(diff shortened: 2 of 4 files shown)


existentialcoder:
@matthewelwell Apologies for the delay. Let me know if this would work!

matthewelwell:
Hi @existentialcoder, thanks for the contribution! Can you make sure that the tests pass (see [here](https://github.com/Flagsmith/flagsmith/actions/runs/18763077358/job/53532213643?pr=6196)) and I'll review it once that's done. Thanks!

existentialcoder:
> Hi @existentialcoder, thanks for the contribution! Can you make sure that the tests pass (see [here](https://github.com/Flagsmith/flagsmith/actions/runs/18763077358/job/53532213643?pr=6196)) and I'll review it once that's done. Thanks!

Addressed. I guess it's only the vercel complaints now

existentialcoder:
> > Hi @existentialcoder, thanks for the contribution! Can you make sure that the tests pass (see [here](https://github.com/Flagsmith/flagsmith/actions/runs/18763077358/job/53532213643?pr=6196)) and I'll review it once that's done. Thanks!
> 
> 
> 
> Addressed. I guess it's only the vercel complaints now

Were you able to check? @matthewelwell

Zaimwa9 (approved):
Thanks, especially for adding the relevant test 👍

api/conftest.py
```diff
@@ -556,6 +556,11 @@ def feature(project: Project) -> Feature:
     return Feature.objects.create(name="Test Feature1", project=project)  # type: ignore[no-any-return]
 
 
+@pytest.fixture()
+def feauture_with_dots(project: Project) -> Feature:
+    return Feature.objects.create(name="feature.name", project=project)  # type: ignore[no-any-return]
+
```
Zaimwa9: Sorry, actually, we can get rid of this and create the feature directly in the test as this will not be used so often.

Additionally, there is a typo :) `feauture`, if we can fix before merging

Zaimwa9 (changes requested):

Zaimwa9 (dismissed):
Thanks 👍

existentialcoder:
> Thanks 👍

Might have fixed the build one more time. Can you check again, sorry :) @Zaimwa9

existentialcoder:
> > Thanks 👍
> 
> 
> 
> Might have fixed the build one more time. Can you check again, sorry :) @Zaimwa9 

Just a reminder. Thanks :)

api/tests/unit/app_analytics/test_unit_app_analytics_views.py
```diff
@@ -23,9 +23,15 @@
     Organisation,
     OrganisationSubscriptionInformationCache,
 )
+from projects.models import Project
 from tests.types import EnableFeaturesFixture
 
 
+@pytest.fixture()
+def feature_with_dots(project: Project) -> Feature:
+    return Feature.objects.create(name="feature.with.dots", project=project)  # type: ignore[no-any-return]
```
matthewelwell, line 32: This doesn't need to be a fixture since it's only used in a single test. I'd rather that we just instantiate the feature directly in the test.

api/app_analytics/views.py
```diff
@@ -88,7 +88,9 @@ def get_fields(self):  # type: ignore[no-untyped-def]
 
             def save(self, **kwargs: typing.Any) -> None:
                 request = self.context["request"]
-                for feature_name, evaluation_count in self.validated_data.items():
+                # validated_data splits out request body with '.' in feature name (e.g a.b.c).
+                # Instead, it's safe to use self.data as keys are not altered.
```
matthewelwell, line 92: This comment doesn't make sense to me. Surely we want the "." in the feature name. Can we be more explicit here about what the difference is between `validated_data` and `data` in this scenario?

Also, is there a neater way to solve this in the serializer rather than the view? I wonder if skipping validated_data here might lead to other problems down the line.

matthewelwell (changes requested):
@existentialcoder thanks again for the contribution, and sorry for the delay in getting back to you after the last review. I've added a few comments in on top of previous requests from @Zaimwa9 .

api/app_analytics/views.py
existentialcoder: Sorry for the late response.
Sure. I can update the comment stating the diff bw data and validated_data. 

In this particular scenario, the validated_data do not hold anything additional and it is safe to use the plain `data` since validation is successful at this point

Another solution is to leverage [SkipField Exception](https://github.com/encode/django-rest-framework/blob/e045dc465270c18689dba4a970378cd9744e57b6/rest_framework/serializers.py#L508). I am actually open to suggestions here as to what can be done.

matthewelwell: > In this particular scenario, the validated_data do not hold anything additional and it is safe to use the plain data since validation is successful at this point

But why? This isn't clear to me.

matthewelwell:
@existentialcoder are you still working on this?

existentialcoder:
> @existentialcoder are you still working on this? 

Yes. I will take a look at the comments and come back. Thanks for the reminder.

gagantrivedi:
Fixed here: https://github.com/Flagsmith/flagsmith/pull/6908

PR #6198, opened 2025-10-24
fix(sse_stream_access_logs): Add timeout

gagantrivedi:

## Changes

Add timeout to stream_access_logs to prevent it from processing weeks’ worth of files at once.

Adds unit tests

### DIFF
--- a/api/sse/sse_service.py
+++ b/api/sse/sse_service.py
@@ -1,5 +1,6 @@
 import csv
 import logging
+import time
 from functools import wraps
 from io import StringIO
 from typing import Generator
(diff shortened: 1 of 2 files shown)


khvn26 (commented):
What this PR does, essentially, is marking the timed out SSE log tasks as successful? I'm not sure if we should do that?

gagantrivedi:
> What this PR does, essentially, is marking the timed out SSE log tasks as successful? I'm not sure if we should do that?



> What this PR does, essentially, is marking the timed out SSE log tasks as successful? I'm not sure if we should do that?

Not really, no — what this PR does is limit the data processing to a specific time window (5 minutes in this case). If it processes that data (from that window), the task is considered successful and should be marked as such.

If we later find that we need a larger time window, we can adjust it accordingly.

khvn26 (approved):
I think I understand now that this will prevent marking a task processing thread as stale/unhealthy.

PR #6199, opened 2025-10-24
docs: update release pipelines beta status

matthewelwell:
## Changes

Update release pipelines status in docs. 

Ran the docs locally and verified that it look as expected, but needs to be on the production docs site in order to test the crisp link.

### DIFF
--- a/docs/docs/advanced-use/release-pipelines.md
+++ b/docs/docs/advanced-use/release-pipelines.md
@@ -2,8 +2,9 @@
 
 :::info
 
-Release Pipelines is currently in a closed beta, please <a class="open-chat" data-crisp-chat-message="Hello, I'm interested in joining the Release Pipelines beta.">contact
-us</a> if you'd like to gain access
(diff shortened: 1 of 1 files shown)


docs/docs/advanced-use/release-pipelines.md
```diff
@@ -1,8 +1,10 @@
 # Release Pipelines
 
-:::warning
+:::info
+
+Release Pipelines is currently in a closed beta, please contact us if you'd like to gain access.
 
-Release Pipelines is currenlty in alpha phase and is not available to the public.
+<a class="open-chat button button--primary" data-crisp-chat-message="Hello, I'm interested in joining the Release Pipelines beta.">Contact Us</a>
```
matthewelwell, line 7: Note that I had to add this button styling because the link styling in alerts is non-existent. Hopefully this will be nicer following the revamp from writechoice, otherwise we might need to dig into the CSS to override the theme.

Zaimwa9 (approved):

PR #5895, opened 2025-08-05
feat: Add segment page

kyle-ssg:
Thanks for submitting a PR! Please check the boxes below:

- [ ] I have added information to `docs/` if required so people know about the feature!
- [x] I have filled in the "Changes" section below?
- [x] I have filled in the "How did you test this code" section below?
- [x] I have used a [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) title for this Pull Request

## Changes

Moves editing a segment to a new page.

image

- Minor: Improves the breadcrumb component. 
Before:

image

After:

image

## How did you test this code?

Created / edited segments + E2E

PR #6203, opened 2025-10-24
fix: show whitespace warnings for IN operator comma-separated values

talissoncosta:

## Changes

- Add whitespace detection for individual items in comma-separated lists
- Show warning only when operator is 'IN'
- Display combined count when multiple whitespace issue types exist
- Optimize performance with single-pass iteration using reduce

### Prerequisites
1. Go to the Segments page

### Test Case 1: Verify Warning Only Shows for IN Operator

**Steps:**
1. Click "Create Segment" or edit an existing segment
2. Add a new condition
3. Select the **IN** operator
4. In the value field, enter: `value1, value2,value3 `
5. Observe the warning icon appears with a yellow border
6. Hover over the warning icon to see the tooltip message

**Expected Result:** 
- Warning icon is displayed
- Tooltip shows: "2 item(s) have whitespace issues"

**Steps (continued):**
7. Change the operator to **Exactly Matches(=)** (keeping the same value)
8. Observe the warning icon disappears

**Expected Result:** 
- No warning icon is shown for non-IN operators

**Additional Operators to Test:**
- Does not match (!=), Contains, Does not contain, Matches regex
- None should show whitespace warnings ✅

---

### Test Case 2: Individual Items with Different Whitespace Issues

**Steps:**
1. Create a condition with the **IN** operator
2. Enter value: `firstvalue,goodexample, badexample1,badexample2 ,goodexample`
   - Note: `" badexample1"` has a leading space
   - Note: `"badexample2 "` has a trailing space

**Expected Result:**
- Warning icon displayed
- Tooltip shows: "2 item(s) have whitespace issues"

---

### Test Case 3: Single Issue Type - Leading Spaces

**Steps:**
1. Create a condition with the **IN** operator
2. Enter value: ` value1, value2, value3`
   - Note: All items have leading spaces

**Expected Result:**
- Warning icon displayed
- Tooltip shows: "3 item(s) have leading whitespaces"

---

### Test Case 4: Single Issue Type - Trailing Spaces

**Steps:**
1. Create a condition with the **IN** operator
2. Enter value: `value1 ,value2 ,value3 `
   - Note: All items have trailing spaces

**Expected Result:**
- Warning icon displayed
- Tooltip shows: "3 item(s) have trailing whitespaces"

---

### Test Case 5: Items with Both Leading and Trailing Spaces

**Steps:**
1. Create a condition with the **IN** operator
2. Enter value: ` value1 , value2 `
   - Note: Both items have leading AND trailing spaces

**Expected Result:**
- Warning icon displayed
- Tooltip shows: "2 item(s) have leading and trailing whitespaces"

---

### Test Case 6: Only Whitespace Value

**Steps:**
1. Create a condition with the **IN** operator
2. Enter value: `   ` (only spaces, no other characters)

**Expected Result:**
- Warning icon displayed
- Tooltip shows: "This value is only whitespaces"

---

### Test Case 7: Non-Comma-Separated Value with Spaces

**Steps:**
1. Create a condition with the **IN** operator
2. Enter value: ` single value with spaces `
   - Note: Single value with leading and trailing spaces

**Expected Result:**
- Warning icon displayed
- Tooltip shows: "This value starts and ends with whitespaces"

---

### Test Case 8: Valid Values (No Warnings)

**Steps:**
1. Create a condition with the **IN** operator
2. Test the following values:
   - `value1,value2,value3` (no spaces)
   - `single_value` (single value, no spaces)
   - `` (empty value)

**Expected Result:**
- No warning icon displayed for any of these values ✅

---

### Test Case 9: Performance with Large Lists

**Steps:**
1. Create a condition with the **IN** operator
2. Enter a value with 50+ comma-separated items, some with spaces
```
value1,value2, value3,value4,value5 ,value6,value7,value8,value9,value10,value11, value12,value13,value14,value15,value16 ,value17,value18,value19,value20,value21,value22,value23, value24,value25,value26,value27,value28,value29,value30 ,value31,value32,value33,value34,value35,value36, value37,value38,value39,value40 ,value41,value42,value43,value44,value45, value46,value47,value48 ,value49,value50,value51,value52,value53, value54,value55,value56,value57,value58 ,value59,value60
```

With this test case I noticed some topics to discuss:

### 1. Modal Not Accessible When Warning is Present  (handled on the PR)

**Issue:** When whitespace warnings are displayed, the warning icon is not clickable and users cannot open the textarea modal to edit long comma-separated lists.

image

**Impact:** For large lists with whitespace issues (e.g., 50+ items), users must fix whitespace problems in the small input field, which is difficult and error-prone.

**Workaround:** Users need to manually remove whitespace issues in the small input field before they can access the modal for easier editing.

**Future Improvement:** Consider one of the following approaches:
- **Option A:** Show both warning and expand icons side-by-side (⚠️ ⤢) - warning for information, expand for modal access
- **Option B:** Add expand icon on hover when warnings are present

---

### 2. Segment Page Layout - Not Full Width (handled on the PR)

image

**Observation:** The Segments creation/editing page does not utilize the full width of the browser window. There are substantial space on the right side

**Suggestion:** Consider making the segment full width

### DIFF
--- a/frontend/web/components/segments/Rule/components/RuleConditionValueInput.tsx
+++ b/frontend/web/components/segments/Rule/components/RuleConditionValueInput.tsx
@@ -107,7 +107,7 @@ const RuleConditionValueInput: React.FC<RuleConditionValueInputProps> = ({
     const TRAILING_WHITESPACE = /\s$/
 
     if (value.length >= 1 && value.trim() === '') {
-      return { hasWarning: true, message: 'This value is only whitespaces' }
+      return { message: 'This value is only whitespaces' }
     }
 
     const items = value.split(',')
@@ -137,26 +137,22 @@ const RuleConditionValueInput: React.FC<RuleConditionValueInputProps> = ({
       if (totalIssues > 0) {
         if (hasMultipleIssues) {
           return {
-            hasWarning: true,
             message: `${totalIssues} item(s) have whitespace issues`,
           }
         }
         
         if (counts.both > 0) {
           return {
-            hasWarning: true,
             message: `${counts.both} item(s) have leading and trailing whitespaces`,
           }
         }
         if (counts.leading > 0) {
           return {
-            hasWarning: true,
             message: `${counts.leading} item(s) have leading whitespaces`,
           }
         }
         if (counts.trailing > 0) {
           return {
-            hasWarning: true,
             message: `${counts.trailing} item(s) have trailing whitespaces`,
           }
         }
@@ -168,22 +164,21 @@ const RuleConditionValueInput: React.FC<RuleConditionValueInputProps> = ({
 
     if (hasLeading && hasTrailing) {
       return {
-        hasWarning: true,
         message: 'This value starts and ends with whitespaces',
       }
     }
     if (hasLeading) {
-      return { hasWarning: true, message: 'This value starts with whitespaces' }
+      return { message: 'This value starts with whitespaces' }
     }
     if (hasTrailing) {
-      return { hasWarning: true, message: 'This value ends with whitespaces' }
+      return { message: 'This value ends with whitespaces' }
     }
 
     return null
   }
 
   const whitespaceCheck = checkWhitespaceIssues()
-  const hasWarning = whitespaceCheck?.hasWarning || false
+  const hasWarning = !!whitespaceCheck
   const isLongText = String(value).length >= 10
 
   const validate = () => {
--- a/frontend/web/components/segments/Rule/components/RuleConditionValueInput.tsx
+++ b/frontend/web/components/segments/Rule/components/RuleConditionValueInput.tsx
@@ -20,6 +20,7 @@ type RuleConditionValueInputProps = {
   isValid?: boolean
   projectId?: number
   showEnvironmentDropdown?: boolean
+  operator?: string
 }
 
 const TextAreaModal = ({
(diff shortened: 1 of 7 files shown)


frontend/web/components/segments/Rule/components/RuleConditionValueInput.tsx
```diff
@@ -92,36 +93,97 @@ const TextAreaModal = ({
...
-    typeof value === 'string' && value.length >= 1 && value.trim() === ''
-  const hasBothLeadingAndTrailingWhitespace =
-    hasLeadingWhitespace && hasTrailingWhitespace
-  const hasWarning =
-    hasLeadingWhitespace ||
-    hasTrailingWhitespace ||
-    hasBothLeadingAndTrailingWhitespace ||
-    isOnlyWhitespace
-  const isLongText = String(value).length >= 10
+  const checkWhitespaceIssues = () => {
```
Zaimwa9: ok amazing, that's super exhaustive

Zaimwa9 (changes requested):
Ok great, it's definitely approvable but while we are at it, I have questions (and answering your observations):
- What do you think about extracting `TextAreaModal` into its own component to aerate the file?
- Same question about extracting your new function into an util within the folder to keep the logic separated from the main component?

The 2 above questions are minor, if you prefer to discard i'll approve as of.

**Observations**:
```
1. Modal Not Accessible When Warning is Present
```
I think that's actually an issue. This warning is likely to appear with large lists, UX will be way better if they can open.
L232 we could exclude this specific warning case, add a cursor icon and as you suggested, maybe the expand icon and/or a "click to fix" in the tooltip. What you think is the best.


```
2. Segment Page Layout - Not Full Width
```
Agreed, we discussed it [here](https://github.com/Flagsmith/flagsmith/pull/5895#discussion_r2300218533), reading the thread, it looks like it was lost in between commits. Please yourself then, I do remember it was ugly full width so feel free to test and choose the most elegant

talissoncosta:
Hey  @Zaimwa9 
Thanks so much for the thoughtful review and observations!

Regarding the questions:
I completely agree with both points, extracting the TextAreaModal and the new function will make the file cleaner and easier to maintain. I’ll work on that.

Regarding the observations:

1. As a quick win, we could make the icon clickable to open the text area. While it’s not the most intuitive solution, since the warning icon doesn’t suggest it can open a larger input,  it would still give users a way to access and edit the full content. As for the tooltip, I don’t think it should handle actions,  tooltips are generally meant for providing information, not interactivity.
2. Thanks for the context! I agree we could move this into a separate ticket so we can explore layout options and decide on the best approach.

Zaimwa9:
@talissoncosta , for the 2. if you have a preference we can straight do it in this one. If you feel it needs more time then agreed let's create a dedicated ticket 👍

talissoncosta:
@Zaimwa9 Sounds good. I actually like the centered approach. We can move forward with that. I agree that full width might look a bit weird.

talissoncosta:
Thanks @Zaimwa9 for the thorough review! I’ve addressed all the points below:

Questions
	1.	✅ TextAreaModal extraction – Done. Moved it into a separate component file.
	2.	✅ Validation function extraction – Done. Added a new whitespaceValidation.ts utility.

Observation 1

✅ Modal accessibility with warnings – Fixed. The warning icon is now clickable to open the modal. As you suggested, this gives users a way to access the larger text area even when warnings are present.

I opted for a clickable icon instead of a “click to fix” tooltip since tooltips are generally informational rather than actionable. I also added subtle hover effects (scale + opacity) to indicate interactivity and included the warning message inside the TextAreaModal.

image


Observation 2

✅ Page layout – Fixed. The tabs content is now centered.

image


Let me know if you’d like any tweaks or further adjustments! 🙂

Zaimwa9 (approved):
Awesome, I like seeing this component so small now 🚀 . Thanks for implementing the feedback.
Just random note I thought about, next improvement could be an auto-trimming within the TextAreaModal but clearly out of scope.

PR #6220, opened 2025-10-29
chore: Add identifier terminology

kyle-ssg:

## Changes

Since we're using the term "Identifier" as a trait in segment rules, we should do our best to have the term in our identity pages.

image

image

Moves delete identity to a danger zone, also removes uses of the word "user"

image

Checked delete identity still worked since this was moved to a danger zone.

### DIFF
--- a/frontend/web/components/pages/UserPage.tsx
+++ b/frontend/web/components/pages/UserPage.tsx
@@ -62,6 +62,7 @@ import UsersIcon from 'components/svg/UsersIcon'
 import IdentityTraits from 'components/IdentityTraits'
 import { useGetIdentitySegmentsQuery } from 'common/services/useIdentitySegment'
 import useDebouncedSearch from 'common/useDebouncedSearch'
+import SettingTitle from 'components/SettingTitle'
 
 const width = [200, 48, 78]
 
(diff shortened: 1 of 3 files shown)


frontend/web/components/pages/UserPage.tsx
```diff
@@ -1034,6 +1016,36 @@ const UserPage: FC = () => {
                         userId={identity?.identity?.identifier || id}
                       />
                     </FormGroup>
+                    <FormGroup className='mt-5'>
+                      <SettingTitle danger>Delete Identity</SettingTitle>
+                      <FormGroup className='mt-4'>
+                        <Row space>
+                          <p className='fs-small lh-sm mb-0'>
+                            Identities can be re-added here or via one of our
```
matthewelwell: I don't think this is particularly clear - I'm not sure what 'here' means.

I'd say let's be more explicit and go with something like: 

"Deleting this identity will delete all of their stored traits, and any identity overrides that you have configured. The identity will be recreated if it is identified via your Flagsmith integration again. You can also recreate it in the dashboard <a href="/project/{projectId}/environment/{environmentId}/users">here</a>."

This feels a bit wordy now, so perhaps we can remove some of the details here, but I think we should err towards this, rather than the unclear statement we currently have here.

matthewelwell:
@kyle-ssg in looking at it here, I think this is definitely too wordy

image

I'd say that we can just get rid of everything after "Deleting this identity will delete all of their stored traits, and any identity overrides that you have configured." ?

talissoncosta (dismissed):

talissoncosta (approved):

Issue #6029, opened 2025-09-08
Spike: PoC for new generic endpoint(s) for updating a flag


matthewelwell:
This PoC should cover building an endpoint, or endpoints, which provides a generic, ergonomic way to update a flag in a given environment, including it's segment overrides.

While multivariate values can likely be descoped from this, we should consider our options and document them as part of this work.

matthewelwell:
I've created a draft PR [here](https://github.com/Flagsmith/flagsmith/pull/6221) which needs some further discussion, but imo the PoC is completed. I'm going to move this to done, and set up a session to discuss the PR with the team.

matthewelwell:
The key actions that we want to achieve that are affected by versioning are: 

 1. Update the state and value of a flag for an environment
 2. Update the state and value of a flag for a segment (this includes creating or deleting a segment override)

The next steps are to create 2 new issues to create new endpoints to (1) create / update a flag, (2) delete a segment override.

PR #6221, opened 2025-10-29
feat: initial PoC for new endpoint to update a flag, regardless of versioning

matthewelwell:
## Changes

Contributes to #6029 

Adds a new endpoint to update a flag in a given environment, regardless of whether versioning is used or not. 

The reasoning behind creating a new endpoint is: 

 1. The API interactions to update a flag in a versioned environment are very convoluted
 2. The current API for updating a flag in a non-versioned environment relies on crud interactions from the django box of tricks

The new endpoint will look something like this: 

```
POST /environments/:id/features/:feature_name/update-flag

{
  "enabled": bool,
  "value": string,
  "type": OneOf["string" | "integer" | ...],
  "segment_id": number | null,
  "multivariate": TODO
}
```

### Other thoughts / outstanding questions:

 1. Should the URL use the environment key, or the environment ID? (realistically, it would be fairly easy to do both here by creating 2 endpoints that route to the same logic under the hood, only the environment retrieval changes)
 2. Do we want to add a separate bulk update endpoint here, or perhaps make this endpoint support bulk operations? 
 3. We still need to implement this functionality for multivariate values, but I don't think that will be too difficult. 
 4. Should this create the flag if it doesn't already exist (in the case of a segment override)
   a. How does the current create-segment-override endpoint currently behave in a versioned environment? 
 5. Error handling? 
   a. We need to consider the following scenarios here: 
      i. Flag is in a release pipeline - this should definitely error
      ii. Flag has a scheduled change - should this error? 
      iii. Change requests are enabled for the environment. 

Based on question (2), we could instead have the endpoint look something like: 

```
POST /environments/:id/update-flags/

{
  "data": [
    {
      "feature_name": string,
      "enabled": bool,
      "value": string,
      "type": OneOf["string" | "integer" | ...],
      "segment_id": number | null,
      "multivariate": TODO
    },
    ...
  ]
}
```

Or even, taking (2) into account: 

```
POST /bulk-update-flags/

{
  "data": {
    "environment_id": number,
    "environment_key": string,
    "flag_updates": [
      {
        "feature_name": string,
        "enabled": bool,
        "value": string,
        "type": OneOf["string" | "integer" | ...],
        "segment_id": number | null,
        "multivariate": TODO
      },
      ...
    ]
  }
}

Note: exactly one of environment_id, or environment_key is required.
```

### Answered questions

> 1. Can we just reimplement the current endpoint to update a feature state in v1? Although we're using Django's crud functionality, we could theoretically pull that method out of the ViewSet, and reimplement ourselves. This would probably be an AND, rather than an instead though as I think adding this generic endpoint still gives us more flexibility. 

Looking into this, the existing endpoint relies on sending a PUT to the endpoint `/environments/:key/featurestates/:id` which relies on the ID of the feature state itself. Since we'd be creating a new feature state as part of the logic for a v2 versioned environment (and hence changing the ID), this would either be really bad UX, or just wouldn't work. 

> 2. Should the URL use the feature name, or the ID? 

We agreed the URL shouldn't include the feature at all, and the name should be included in the payload. The URL will just be `/environments/:key|id/update-flag`

Added a basic unit test, but more tests will need to be added when this becomes a real change that we want to make.

### DIFF
--- a/api/environments/urls.py
+++ b/api/environments/urls.py
@@ -7,6 +7,7 @@ from edge_api.identities.views import (
     EdgeIdentityWithIdentifierFeatureStateView,
     get_edge_identity_overrides,
 )
+from features.feature_states.views import update_flag
 from features.views import (
     EnvironmentFeatureStateViewSet,
     IdentityFeatureStateViewSet,
@@ -166,5 +167,10 @@ urlpatterns = [
         "<str:environment_api_key>/edge-identity-overrides",
         get_edge_identity_overrides,
         name="edge-identity-overrides",
+    ),
+    path(
+        "<int:environment_id>/features/<str:feature_name>/update-flag/",
+        update_flag,
+        name="update-flag",
     ),
 ]
--- a/api/features/serializers.py
+++ b/api/features/serializers.py
@@ -1,3 +1,4 @@
+import typing
 from datetime import datetime
 from typing import Any
 
@@ -18,6 +19,7 @@ from rest_framework.exceptions import PermissionDenied
 
 from app_analytics.serializers import LabelsQuerySerializerMixin, LabelsSerializer
 from environments.identities.models import Identity
+from environments.models import Environment
 from environments.sdk.serializers_mixins import (
     HideSensitiveFieldsSerializerMixin,
 )
@@ -28,6 +30,7 @@ from projects.code_references.serializers import (
     FeatureFlagCodeReferencesRepositoryCountSerializer,
 )
 from projects.models import Project
+from users.models import FFAdminUser
 from users.serializers import (
     UserIdsSerializer,
     UserListSerializer,
@@ -47,6 +50,8 @@ from .feature_segments.serializers import (
 )
 from .models import Feature, FeatureState
 from .multivariate.serializers import NestedMultivariateFeatureOptionSerializer
+from .versioning.dataclasses import FlagChangeSet
+from .versioning.versioning_service import update_flag
 
 
 class FeatureStateSerializerSmall(serializers.ModelSerializer):  # type: ignore[type-arg]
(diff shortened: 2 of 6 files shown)


api/features/serializers.py
```diff
@@ -671,3 +676,45 @@ def create(self, validated_data: dict) -> FeatureState:  # type: ignore[type-arg
...
+    #  to what we should have done from the start. This might be out of scope for this
+    #  work though.
+    feature_state_value = serializers.CharField(required=False)
+    type = serializers.ChoiceField(
+        choices=["int", "string", "bool", "float"],
+        required=False,
+        default="string",
+    )
+
+    segment_id = serializers.IntegerField(required=False)
```
gagantrivedi, line 694: Do we need to provide priority here as well

matthewelwell: As discussed we @gagantrivedi, we plan to focus on the environment default case for now, and ignore for segments for now. The logic behind this is it creates a much more streamlined endpoint here for users that don't care about segments. 

We could then either: 

 1. Update this endpoint to include segment information if we get feedback that it should
 2. Create another similar endpoint which handles this for a specific segment

api/environments/urls.py
```diff
@@ -167,4 +168,9 @@
         get_edge_identity_overrides,
         name="edge-identity-overrides",
     ),
+    path(
+        "<int:environment_id>/features/<str:feature_name>/update-flag/",
```
matthewelwell, line 172: Again, as discussed with @gagantrivedi, let's remove the feature name from the URL since there is nothing stopping it containing HTTP reserved characters. The endpoint should therefore be: 

```
POST /environments/:key/update-flag

{
  "feature_name": "foo",
  ...
}
```

api/features/serializers.py
matthewelwell: On further discussions, we have gone back to accommodating segment overrides in this endpoint. 

We should make priority optional (if the priority is missing, it just gets added to the end of the list). We see that there are 2 main use cases here:
 1. Send `"priority": 1` to add it to the top of the list
 2. Send `"priority": null` (or don't send) and the segment is added to the end of the list. 

The endpoint should handle creating or updating a segment override, and should return in the response payload whether the change was an update or not (the status code should still be 200).

matthewelwell:
Closing this PR in favour of creating a more complete version as part of #6233

Issue #5835, opened 2025-07-25
Context Values: Implement multi selection dropdown


Zaimwa9:
For `environment.name` context value, we implemented a Searchable Dropdown with environments pre-defined.
However this does not work with `IN` operator.
For now the previous input value is used for the `IN` operator and should be replaced with a multi-select dropdown instead.

## User Story
As a user selecting the `Environment Name` context value and the `IN` operator, I should be able to select multiple of my environments from a dropdown showing the list of my environments

## DoD
As a dashboard user, when I create a `IN` rule based on an `environment.name` I can select multiple environments from the dropdown

Zaimwa9:
CSS and state management involved
<img width="576" height="290" alt="Image" src="https://github.com/user-attachments/assets/1e79b63c-1bc8-4b50-b7c6-6a06d19fa69c" />

Issue #5991, opened 2025-08-27
Context Values: Bump flagsmith-flag-engine to ^10.0.3.


khvn26:
The overarching goal is to bump the dependency version to `^10.0.3` and make sure the tests pass.

- [ ] Bump `flagsmith-flag-engine` to `^10.0.3`.

Related: https://github.com/Flagsmith/flagsmith/issues/6654

khvn26:
Needs to be broken down.

PR #6239, opened 2025-11-04
feat: context values - implement multi-select dropdown on envs

talissoncosta:
Contributes to https://github.com/Flagsmith/flagsmith/issues/5835

---

## Changes

image

image

image

image

_Please describe._

_Please describe._

---

> [!NOTE]
> Adds multi-select for environment names (IN) and single-select (EQUAL/NOT_EQUAL) in segment rules, with cached condition values and minor UI/style tweaks.
> 
> - **Segments: Rules**
>   - **Input types**: `RuleConditionValueInput` now uses `MultiSelect` for `IN` on `$.environment.name` and `EnvironmentSelectDropdown` for `EQUAL/NOT_EQUAL`; loads environment options via `useGetEnvironmentsQuery` and parses arrays with `safeParseArray`.
>   - **State handling**: Introduces `useConditionCache` to cache/restore condition values when changing operator/property; integrated into `Rule.tsx`.
>   - **Hooks**: Adds `useConditionInputType` and exports via `hooks/index.ts`; simplifies `useRuleContext` (removes env dropdown logic).
>   - **UI**: Refactors `RuleConditionRow` layout (grid classes), passes `property` to value input.
> - **Components**
>   - **MultiSelect**: Styling/behavior tweaks (24px chip height, w-100, smaller gaps, conditional components, className handling).
>   - **Select wrapper**: Preserves `className` on container.
> - **Modals**
>   - `CreateSegmentRulesTabForm`, `CreateSegment`, `CreateSegmentUsersTabContent`, `AssociatedSegmentOverrides`: Layout adjustments (responsive columns, `col-4` env selects) and spacing tweaks.
> - **Styles**
>   - `react-select`: Use `min-height` for control; extensible value container unchanged; segment rule select width set to 100%.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 006eb984312dde95ee7cde5c2b642e65d63a703d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


frontend/web/project/project-components.js
```diff
@@ -118,7 +118,7 @@ global.Select = class extends PureComponent {
   }
 
   render() {
-    const props = this.props
+    const { className, ...props } = this.props
```
talissoncosta, line 121: We had an issue in here, className was being applied for the container and also for the Select component, passed as {...props}. Now I am extracting it to apply only for the container, that in my view was the goal in here.

kyle-ssg:
It looks like they're wrapping for me at the moment

image

I think this will always be the case using react-selects multi feature here, not sure there'd be a way around it. Perhaps the ability to view all which shows a modal where we have more space like here

image
 [](edited-with-Axolo)

kyle-ssg:
Makes sense, I think you can create this quite easily with react-select

talissoncosta:
You were right, @kyle-ssg  I got it working with react-select by keeping the tags approach and preventing the input height from growing.
image


and it looks way better now.

matthewelwell:
A few things I've noticed testing this out. 

1. The placeholder text is now wrapping onto 2 lines which looks a bit awkward.
2. The 'Description' field has been moved inline with the 'Name' field which won't work very well since we expect the 'Description' field to include reasonable amounts of text (in fact it should potentially be a text area rather than an input, but that's for another day). 

Screenshot 2025-11-06 at 15 54 33

3. For larger environment names, the size of the 'checked' icon seems to change. 

image

It looks like these things all look good in the 'Edit Segment' view, for which we're able to use the full screen real estate, but the 'Create Segment' journey still opens the slideout modal. I wonder then if the solution is to have 'Create Segment' it's own page rather than the slide out modal. I'm not sure if it was intentional to keep it there @kyle-ssg ? 

(regardless of this, I think the description field should still have it's own line).

talissoncosta:
Thanks for the feedback @matthewelwell . All points make sense to me. Working on it right now.

talissoncosta:
@matthewelwell 
Indeed the search was not so clear. I've fixed that and also all the topics you mentioned on your review.

https://www.loom.com/share/5fa0451baa2d4335bb6b086782f88bbd

matthewelwell (dismissed):
I've tested this code manually, and taken a look over the code which looks clean to me, but it would be good to have one of the @Flagsmith/flagsmith-front-end team look over it too. 

@talissoncosta can you also check and resolve the comments from Cursor?

talissoncosta:
Thank you @matthewelwell 
Yep, I've resolved all of them 😃

matthewelwell (dismissed):
Re-approving post minor change to fix review from Cursor.

frontend/web/components/modals/CreateSegmentUsersTabContent.tsx
```diff
@@ -98,6 +98,7 @@ const CreateSegmentUsersTabContent: React.FC<
         <FormGroup>
```
Zaimwa9, line 98: Sorry, unrelated to this file but, a small change we could do while we are at it. All the tabs are full width now, except Custom Fields. If you don't mind to quickly harmonize it now 👍 

image

frontend/web/components/base/select/multi-select/MultiSelect.tsx
```diff
@@ -0,0 +1,122 @@
...
+      )}>
+      {label && <label>{label}</label>}
+      <Select
+        isMulti
+        hideSelectedOptions={hideSelectedOptions}
+        closeMenuOnSelect={false}
+        placeholder={placeholder}
+        isDisabled={disabled}
+        size={size}
+        onChange={(selectedOptions: MultiSelectOption[]) => {
```
Zaimwa9: I believe that can be null as suggests the assignation below
```suggestion
        onChange={(selectedOptions: MultiSelectOption[] | null) => {
```

frontend/web/components/segments/Rule/components/RuleConditionValueInput.tsx
```diff
@@ -18,19 +24,69 @@ type RuleConditionValueInputProps = {
...
+  )
+
+  if (showMultiEnvironmentSelect) {
+    return (
+      <div className={className}>
+        <MultiSelect
+          selectedValues={safeParseArray(value)}
+          onSelectionChange={(selectedValues: string[]) => {
+            onChange?.(JSON.stringify(selectedValues))
+          }}
```
Zaimwa9, line 66: Critical: We have a payload formatting issue here. It might be also related to a backend issue. We want to accept JSON-encoded lists. See this [issue](https://github.com/Flagsmith/flagsmith/issues/5935)

However right now it is not working. You can test it by:
- Creating a segment with ONE Rule environment.name IN with one or more environment 
- Go to Users and select one of the environment
- Users should be within the segment (it's not even with one env.

If you checkout to main and use the normal csv format for `IN` it will work

https://www.loom.com/share/faca8617e13b4338bbaf0a263ef00f25

Zaimwa9: I will investigate deeper the backend contract and where we stand on the JSON-encoded handler

frontend/web/components/base/select/multi-select/CustomMultiValue.tsx
```diff
@@ -0,0 +1,52 @@
...
+          maxWidth: '150px',
+          overflow: 'hidden',
+          padding: '2px 6px',
+          textOverflow: 'ellipsis',
+          whiteSpace: 'nowrap',
+          maxHeight: '24px'
+        }}
+      >
+        <span
+          className='mr-1'
```
Zaimwa9: NIT: maybe we can let the parent div handle the layout with a `gap-x-1`

frontend/web/components/base/select/multi-select/InlineMultiValue.tsx
```diff
@@ -0,0 +1,31 @@
...
+    const currentIndex = selectedOptions.findIndex((opt) => opt.value === data.value)
+
+    // Only render on the first item to avoid duplicates
+    if (currentIndex !== 0) return null
+
+    const formattedText = selectedOptions.length === 1
+      ? selectedOptions[0].label
+      : selectedOptions.length === 2
+      ? `${selectedOptions[0].label} and ${selectedOptions[1].label}`
+      : `${selectedOptions.slice(0, -1).map((opt) => opt.label).join(', ')} and ${selectedOptions[selectedOptions.length - 1].label}`
```
Zaimwa9: NIT: Do you mind breaking this to avoid nested ternaries?

frontend/web/components/segments/Rule/components/RuleConditionValueInput.tsx
```diff
@@ -18,19 +24,69 @@ type RuleConditionValueInputProps = {
...
+  )
+  const environmentOptions = useMemo(
+    () =>
+      data?.results
+        ? data.results.map(({ name }) => ({ label: name, value: name }))
+        : [],
+    [data],
+  )
+
+  const { showMultiEnvironmentSelect, showSingleEnvironmentSelect } = useConditionInputType(
```
Zaimwa9, line 54: 👍

frontend/web/project/project-components.js
Zaimwa9: Good catch

Zaimwa9 (changes requested):
Great great great addition thanks. A couple of NITs and one critical issue that needs to be addressed, might be related to backend. I'll investigate know to see whether we should fix the payload format in the frontend or BE

frontend/web/components/segments/Rule/components/RuleConditionValueInput.tsx
Zaimwa9: Update on this one @talissoncosta, the flagsmith-engine version is lagging behind in the Core Api. (version 6.0.2 while this new behavior has been implemented in 7.0.0+).
The upgrade is not as straightforward as the API is relying on engine models that have been deprecated in v7+ so it will need to be properly addressed (issue [here](https://github.com/Flagsmith/flagsmith/issues/5991)).

So, after discussing with Matthew, as we are not going to address the above issue in the short term, we will need to revert to the legacy csv format.

Basically it means we need the following payload for a `IN` rule (and the adequate parsers):
```
{
    "id": 64379181,
    "operator": "IN",
    "property": "$.environment.name",
    "value": "Development,Test,Production_environment", // Instead of ["Development","Test","Production_environment"]
    "description": null
}
```

Sorry for the back and forth!

frontend/web/components/modals/CreateSegmentUsersTabContent.tsx
talissoncosta: For sure, thanks for catching that. Custom fields were not enabled for me. I will handle it RN.

frontend/web/components/base/select/multi-select/CustomMultiValue.tsx
talissoncosta: Good catch!

frontend/web/components/segments/Rule/components/RuleConditionValueInput.tsx
talissoncosta: Thanks for investigating this.

frontend/web/components/base/select/multi-select/InlineMultiValue.tsx
```diff
@@ -6,14 +6,18 @@ export const InlineMultiValue = (props: MultiValueProps<MultiSelectOption>) => {
...
 
-    // Only render on the first item to avoid duplicates
     if (currentIndex !== 0) return null
 
-    const formattedText = selectedOptions.length === 1
-      ? selectedOptions[0].label
-      : selectedOptions.length === 2
-      ? `${selectedOptions[0].label} and ${selectedOptions[1].label}`
-      : `${selectedOptions.slice(0, -1).map((opt) => opt.label).join(', ')} and ${selectedOptions[selectedOptions.length - 1].label}`
+    let formattedText: string
```
Zaimwa9, line 11: thanks, a bit more verbose but I believe easier for the brain

Zaimwa9 (approved):
Perfect, looks good to me now, thanks for the quick changes

PR #6252, opened 2025-11-06
docs: Update quickstart redirect

robertnorrie:
Thanks for submitting a PR! Please check the boxes below:

- [ ] I have added information to `docs/` if required so people know about the feature!
- [x] I have filled in the "Changes" section below?
- [x] I have filled in the "How did you test this code" section below?
- [x] I have used a [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) title for this Pull Request

## Changes

Updating redirects for quickstart links 
## How did you test this code?

N/A

PR #6253, opened 2025-11-07
docs: revert removing duplicates

matthewelwell:
This PR reverts part of Flagsmith/flagsmith#6252 which removed 'duplicates' that shouldn't have been removed because vercel handles URLs with / without trailing slashes as different URLs.


matthewelwell:
I think we can actually simplify using `"trailingSlash": false` instead so I'm going to close this.

PR #6255, opened 2025-11-07
fix: segment change request warning shows on all segments

### LINKED ISSUE #6219: All segments show change request warning

- [ ] Self Hosted with Docker
- [x] SaaS at flagsmith.com

### Describe the bug

If there is an open segment change request, then all segments and the 'New segment' modal shows the warning saying that there are open change requests. 

### Steps To Reproduce

1. In a project with segment change requests enabled, create a change request for a segment
2. Click on 'Create Segment'
3. Notice that there is a warning showing that there are open change requests for the segment

### Expected behavior

The warning only shows on the segment that has the open change request

### Screenshots

Image

talissoncosta:

Fixes #6219 When a project has segment change requests enabled and a change request is created for one segment, all segments (including the "New segment" modal) incorrectly show the warning that there are open change requests. The warning should only appear on the segment that actually has an open change request.

## Changes

The issue was caused by a query parameter mismatch between the frontend and the backend API:
Frontend was sending: version_of: segmentId
Backend expected: segment_id
In the ExistingProjectChangeRequestAlert component, the API query was using an incorrect parameter name (version_of) that the backend API doesn't recognize. As a result, the filter wasn't being applied, and the API was returning all open change requests for the project instead of just those for the specific segment.

### Changes Made
- ExistingProjectChangeRequestAlert.tsx
   - Changed query parameter from version_of to segment_id
- requests.ts
   - Updated TypeScript type definition to match the correct API parameter name

Manual Testing Steps:
- Create a segment in a project with workflows enabled
- Create a change request that modifies that segment
- Navigate to the segment list or segment details page
- ✅ Verify the warning only appears on the segment with the open change request
- ✅ Verify other segments don't show the warning
- ✅ Verify the "New segment" modal doesn't show the warning

## Impact
- Before: All segments showed change request warnings when any segment had an open change request
- After: Only the specific segment with an open change request shows the warning

### DIFF
--- a/frontend/common/types/requests.ts
+++ b/frontend/common/types/requests.ts
@@ -655,7 +655,7 @@ export type Req = {
   }
   getProjectChangeRequests: PagedRequest<{
     project_id: string
-    version_of?: string
+    segment_id?: string
     live_from_after?: string
     committed?: boolean
   }>
--- a/frontend/web/components/ExistingProjectChangeRequestAlert.tsx
+++ b/frontend/web/components/ExistingProjectChangeRequestAlert.tsx
@@ -14,7 +14,7 @@ const ExistingProjectChangeRequestAlert: FC<
   const { data } = useGetProjectChangeRequestsQuery({
     committed: false,
     project_id: projectId,
-    version_of: segmentId,
+    segment_id: segmentId,
   })
 
   if (data?.results?.length) {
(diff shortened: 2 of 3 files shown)


frontend/common/types/requests.ts
```diff
@@ -655,7 +655,7 @@ export type Req = {
   }
   getProjectChangeRequests: PagedRequest<{
     project_id: string
-    version_of?: string
+    segment_id?: string
```
Zaimwa9, line 658: YES! thanks!

Zaimwa9 (approved):
Alright, so basically the bug was coming from a wrong query param right? Good catch thanks

PR #6265, opened 2025-11-11
refactor(Segments): Clean up after Segment Change Request work

### LINKED ISSUE #5825: Clean up

emyller:
Closes #5825

## Changes

- [x] **Make `Segment.version.default = 1` at field level instead of using a hook**
- ~Inherit `MetadataSerializerMixin` from `Serializer`~ Moved to [new branch](https://github.com/Flagsmith/flagsmith/tree/refactor/improve-metadata-serializers)
- ~Make `Segment.version_of = null` the indication for a live segment~ Moved to #6275.

## Implementation Notes

> [!NOTE]
> **Migration `0030_add_default_to_segment_version`:**
> - Non-blocking and safe — schema-only change
> - Field remains nullable for backward compatibility
> - All existing segments already have `version=1` (set by migration `0025_set_default_version_on_segment`)


api/segments/models.py
```diff
@@ -30,13 +29,24 @@
...
+    pass
+
+
+class LiveSegmentManager(SoftDeleteExportableManager):
+    def get_queryset(self):  # type: ignore[no-untyped-def]
+        """
+        Returns only the canonical segments, which will always be
+        the highest version.
+        """
+        return super().get_queryset().filter(id=models.F("version_of"))
```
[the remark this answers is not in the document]
emyller: This was moved to #6275.

api/metadata/serializers.py
```diff
@@ -103,11 +103,15 @@ def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
         return attrs
 
 
-class MetadataSerializerMixin:
+class MetadataSerializerMixin(serializers.Serializer):  # type: ignore[type-arg]
     """
-    Functionality for serializers that need to handle metadata
+    Mixin for serializers that need to handle metadata
+
+    NOTE: Child serializers should include 'metadata' in their Meta.fields.
```
matthewelwell: In the interest of getting this PR merged before EOS, I've gone ahead and removed the metadata changes as suggested by @gagantrivedi . 

I've kept those changes only in a branch called `refactor/improve-metadata-serializers` ([here](https://github.com/Flagsmith/flagsmith/tree/refactor/improve-metadata-serializers)) in case we want to resurrect them as part of the issue that Gagan has raised.  [](edited-with-Axolo)

api/segments/models.py
```diff
@@ -87,8 +97,7 @@ class Segment(
         Feature, on_delete=models.CASCADE, related_name="segments", null=True
     )
 
-    # This defaults to 1 for newly created segments.
-    version = models.IntegerField(null=True)
+    version = models.IntegerField(default=1, null=True)
```
matthewelwell, line 100: I'm not sure why we didn't just have `default=1, null=True` here in the first place? 🤔  The fact that there is so much code around it would indicate that we at least thought about it, but I can't see any reason that we wouldn't have done it that way in the first place.

matthewelwell (dismissed):
A couple of minor comments, but nothing that I see as a blocker if we want to get this merged ahead of the sprint closure.

api/metadata/serializers.py
emyller: I think the proposed change is definitely cleaner because it ditches imports of `MetadataSerializer` to fulfill a `metadata =` field definition in every serializer that needs it. Technically we _could_ tinker with DRF machinery to implicitly force the metadata field into the serializer's `fields` but I don't see us happy going that path.

I'd love to make things more automatic but I think this is the best compromise we can have. 😞

api/segments/models.py
emyller: Good question. I did some git digging with Claude, and neither them or I could come to any conclusions. My best guess is someone tried to avoid a database migration the wrong way, possibly because they didn't know it would be a state-only migration? 🤷

emyller: Given [this](https://github.com/Flagsmith/flagsmith/blob/ec45a23da4bddf83c4caafb329caf385e39dd99d/api/segments/migrations/0025_set_default_version_on_segment.py#L12) happened, I would be actually happier to make `version` non-nullable. 🤔  (though this also has a risk of locking, so not now either)

api/metadata/serializers.py
gagantrivedi: Let's avoid touching metadata in this PR - it has tons of other problems that should be addressed separately. I have the issue for that here[1]:
[1]: https://github.com/Flagsmith/flagsmith/issues/6277

matthewelwell: In the interest of getting this PR merged before EOS, I've gone ahead and removed the metadata changes as suggested by @gagantrivedi . 

I've kept those changes only in a branch called `refactor/improve-metadata-serializers` ([here](https://github.com/Flagsmith/flagsmith/tree/refactor/improve-metadata-serializers)) in case we want to resurrect them as part of the issue that Gagan has raised.

gagantrivedi (approved):

api/metadata/serializers.py
emyller: > In the interest of getting this PR merged before EOS, I've gone ahead and removed the metadata changes as suggested by @gagantrivedi .

@matthewelwell Thanks for doing that while I was away!

> Let's avoid touching metadata in this PR - it has tons of other problems that should be addressed separately. I have the issue for that here: #6277

@gagantrivedi I love that issue because I can relate to the pain in every bullet point of it. The changes here would barely shave off one of the items. Agreed with the decision here, thanks!

PR #6272, opened 2025-11-11
feat: Add a view mode selector for diffing features

kyle-ssg:

## Changes

Allows you to select between a diff and just the old or new value in a change request or version history.

image

View a change request and toggle through the view modes

### DIFF
--- a/frontend/web/components/diff/DiffFeature.tsx
+++ b/frontend/web/components/diff/DiffFeature.tsx
@@ -59,6 +59,9 @@ const DiffFeature: FC<FeatureDiffType> = ({
   const diff = getFeatureStateDiff(oldEnv, newEnv)
   const { conflict: valueConflict, totalChanges } = diff
   const [value, setValue] = useState(0)
+  const [viewMode, setViewMode] = useState<'combined' | 'new' | 'old'>(
+    'combined',
+  )
 
   const { data: segments } = useGetSegmentsQuery({
     include_feature_specific: true,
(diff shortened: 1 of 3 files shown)


frontend/web/components/diff/DiffFeature.tsx
```diff
@@ -96,6 +99,33 @@ const DiffFeature: FC<FeatureDiffType> = ({
...
+                    options={[
+                      { label: 'Combined Diff', value: 'combined' },
+                      { label: 'New Value', value: 'new' },
+                      { label: 'Old Value', value: 'old' },
+                    ]}
+                    onChange={(option: { value: 'combined' | 'new' | 'old' }) =>
+                      setViewMode(option.value)
+                    }
+                  />
+                </div>
```
talissoncosta, line 117: Not an issue — just sharing some thoughts. We might consider creating a component like SelectField to encapsulate and wire the label, select, and error field together. [](edited-with-Axolo)

talissoncosta (dismissed):
Looks good! 😃

frontend/web/components/diff/DiffFeature.tsx
kyle-ssg: I think this is more that I should tidy up and reuse the options field - I wish that react-select would resolve this with a better syntax :D. 

I wonder if we could make our base Select component a bit more clever around this, having too many similar components SelectField, Select etc will just get complicated.

kyle-ssg: I've adjusted this!

talissoncosta (approved):

frontend/web/components/diff/DiffFeature.tsx
talissoncosta: Yep, I agree — what I meant is more of a wrapper for the existing Select component, where it handles wiring things like the label, error message, and maybe even layout consistency. That way, we don’t necessarily duplicate the logic of Select itself, but we could have a more convenient way to use it in forms without repeating the same boilerplate setup each time.

PR #6286, opened 2025-11-13
fix: Revert: chore(sales-dashboard): add multiselect to plan filter

gagantrivedi:
Reverts Flagsmith/flagsmith#6243


matthewelwell (approved):
For context, we're reverting this in favour of an alternative optimisation.

PR #6295, opened 2025-11-17
fix: remove-firefox-and-install-v143

Zaimwa9:

## Changes
- Added step to install v143.0 in CI 

- Built the images locally
- Ran the flow and grepped the version

### DIFF
--- a/.github/workflows/.reusable-deploy-ecs.yml
+++ b/.github/workflows/.reusable-deploy-ecs.yml
@@ -115,9 +115,24 @@ jobs:
       - name: Cloning repo
         uses: actions/checkout@v5
 
+      - name: Install Firefox 143.0
+        run: |
+          sudo apt-get remove -y firefox || true
+          sudo rm -rf /usr/bin/firefox /usr/lib/firefox*
+          
+          ARCH=$(uname -m)
+          wget -O /tmp/firefox.tar.xz "https://ftp.mozilla.org/pub/firefox/releases/143.0/linux-${ARCH}/en-US/firefox-143.0.tar.xz"
+          sudo tar -xJf /tmp/firefox.tar.xz -C /opt
+          sudo ln -s /opt/firefox/firefox /usr/local/bin/firefox
+          rm /tmp/firefox.tar.xz
+          
+          firefox --version
+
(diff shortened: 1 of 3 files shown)


.github/workflows/.reusable-deploy-ecs.yml
```diff
@@ -115,6 +115,19 @@ jobs:
       - name: Cloning repo
         uses: actions/checkout@v5
 
+      - name: Install Firefox 143.0
```
Zaimwa9, line 120: I saw an alternative could be to use a github action. I read that it could lead to conflicts with paths and be dependent on how test-cafe picks up the firefox version (That I could not set in the options)

```
  - name: Setup Firefox 143.0
    uses: browser-actions/setup-firefox@v1
    with:
      firefox-version: '143.0'
```

emyller: I just learned about https://github.com/browser-actions and loved it — I wish it had more community support. @Zaimwa9 Do you think it's worth trying it out?

Zaimwa9: I hesitated, as my knowledge is limited there, so I would run the CI against this version to confirm the fix and then I could give it a try

emyller (changes requested):
Patch looks good for the sake of not having CI blocking the team.

1. Please let's just add a bit of documentation explaining why this is necessary in the code — this must be temporary? It could be a comment before each "Install Firefox 143.0" job.
2. We'd benefit from another issue to investigate whether we can use some tool (e.g. [this](https://github.com/Flagsmith/flagsmith/pull/6295#discussion_r2534037836)) that helps reducing our maintenance and adding cache.

Will approve right now if urgent, or right after the above commented are tackled — should be small effort, thanks in advance!

.github/workflows/.reusable-deploy-ecs.yml
emyller: I looks really nice but potentially a sprint task. See [here](https://github.com/Flagsmith/flagsmith/pull/6295#pullrequestreview-3472778495).

emyller (approved):
Nice 👍

PR #6301, opened 2025-11-18
fix(DB replication): Prevent replica lag in environment document endpoint

emyller:
Removes database replica usage from `/api/v1/environment-document` to prevent stale data from replica lag.

## Changes
- [x] Remove `using_database_replica` from `Environment._get_environment_document_from_db`

Review effort: 1/5

### DIFF
--- a/api/environments/models.py
+++ b/api/environments/models.py
@@ -4,7 +4,6 @@ import uuid
 from copy import deepcopy
 from typing import TYPE_CHECKING, Literal
 
-from common.core.utils import using_database_replica
 from django.conf import settings
 from django.contrib.contenttypes.fields import GenericRelation
 from django.core.cache import caches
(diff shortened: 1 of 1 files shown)


Zaimwa9 (approved):
Clear 😄

matthewelwell:
@emyller as discussed on Slack, unless there are any race conditions that we should be worried about with regards to DB replication and `If-Modified-Since`, I think the benefits of distributing the load to a read replica outweigh any possible additional latency in flag updates based on replication lag.
 [](edited-with-Axolo)

emyller:
> (...) I think the benefits of distributing the load to a read replica outweigh any possible additional latency in flag updates based on replication lag.



Agreed.

Thanks @Zaimwa9 for the review!
 [](edited-with-Axolo)

Issue #6279, opened 2025-11-12
Create new experimental endpoint to update a single feature state for a given feature


matthewelwell:
This issue is born out of the steps defined in a comment [here](https://github.com/Flagsmith/flagsmith/issues/6233#issuecomment-3496840996).

This issue should cover the creation of an _experimental_ endpoint that looks like the payload as given below. The goal of this endpoint is to simplify API integration use cases where a customer wants to update the state / value of a given feature in a given environment. This endpoint should look the same regardless of whether versioning is enabled for the environment. 

Some things that we still need to decide: 

 1. Should the URI just include the feature ID instead of including it in the payload?
 2. How do we define that this endpoint is experimental and liable to be changed without warning (probably just a namespace?). 

Some other notes: 

 1. For this experimental endpoint, we are going to ignore multivariate updates

```jsonc
{
  "feature": {
    // only one of these should be included in each request
    "name": "my_feature_name",
    "id": 1
  },
  "segment": {
    // since segment name is not unique (yet), we can't
    // offer name here like we do for feature
    "id": 1,
    "priority": 1 // can be null / omitted, which will add it to the bottom of the list
  },
  "enabled": true,
  "value": {
    "string_value": "",
    "type": "integer" // one of "integer", "string", "boolean", "float"
  },
}
```

matthewelwell:
Superseded by https://github.com/Flagsmith/flagsmith/issues/6233

PR #6305, opened 2025-11-19
feat(api): add experimental feature flag update endpoints 

### LINKED ISSUE #6233: Create new experimental endpoints to update a given feature in an environment
Related issues / PRs: #6221, #6029, #6279 

We need to create 2 new endpoints which will allow us to have a single endpoint to update a feature in an environment, regardless of whether the environment has v2 feature versioning enabled or not. 

The 2 new endpoints will look very similar with subtle differences to allow us to determine (via customer feedback) if we want both, or prefer one over the other. One endpoint will cater for updating both the environment default flag, as well as creating and updating segment overrides, all in a single request. The other will expect each atomic update to be done in an individual request. The expected payloads are given below. 

Some other notes:

 1. For these experimental endpoints, we are going to ignore multivariate updates

```jsonc
// Update all components of a feature in one request
{
  "feature": {
    // only one of these should be included in each request
    "name": "my_feature_name",
    "id": 1
  },
  "environment_default": {
    "enabled": true,
    "value": {
      "string_value": "",
      "type": "integer" // one of "integer", "string", "boolean", "float"
    }
  },
  "segment_overrides": [
    {
      "segment": {
        "id": 1,
        "priority": 1 // can be null / omitted, which will add it to the bottom of the list
      },
      "enabled": true,
      "value": {
        "string_value": "",
        "type": "integer"
        // one of "integer", "string", "boolean", "float"
      }
    }
  ]
}
```

```jsonc
// Update individual component of a feature in each request
{
  "feature": {
    // only one of these should be included in each request
    "name": "my_feature_name",
    "id": 1
  },
  "segment": {
    // since segment name is not unique (yet), we can't
    // offer name here like we do for feature
    "id": 1,
    "priority": 1 // can be null / omitted, which will add it to the bottom of the list
  },
  "enabled": true,
  "value": {
    "string_value": "",
    "type": "integer" // one of "integer", "string", "boolean", "float"
  },
}
```

gagantrivedi:

## Changes
Adds two new experimental endpoints under /api/experiments/:
- update-flag-v1: Update single feature state (environment default or segment override)
- update-flag-v2: Update multiple feature states in a single request (environment default + segment overrides)
Related issues: #6279 #6233 

Unit tests


api/features/versioning/versioning_service.py
```diff
@@ -101,6 +106,331 @@ def get_current_live_environment_feature_version(
...
+def _update_flag_for_versioning_v1(
+    environment: Environment, feature: Feature, change_set: FlagChangeSet
+) -> FeatureState:
+    from features.models import FeatureSegment, FeatureState
+
+    latest_feature_states = get_environment_flags_dict(
+        environment=environment,
+        feature_name=feature.name,
+        additional_filters=Q(feature_segment__segment_id=change_set.segment_id),
+    )
```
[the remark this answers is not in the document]
gagantrivedi: You are wrong but I get it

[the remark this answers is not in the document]
gagantrivedi: Are you happy now?

api/features/feature_states/serializers.py
```diff
@@ -0,0 +1,170 @@
...
+                "Either 'name' or 'id' is required for feature identification"
+            )
+        if data.get("name") and data.get("id"):
+            raise serializers.ValidationError("Provide either 'name' or 'id', not both")
+        return data
+
+
+class FeatureUpdateSegmentDataSerializer(serializers.Serializer):  # type: ignore[type-arg]
+    id = serializers.IntegerField(required=True)
+    priority = serializers.IntegerField(required=False, allow_null=True)
```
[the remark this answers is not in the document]
gagantrivedi: Validation is done in BaseFeatureUpdateSerializer

api/environments/identities/models.py
```diff
@@ -233,6 +233,7 @@ def generate_traits(
 
     def update_traits(
         self,
+        # TODO: this typing isn't true as it also supports regular dicts like {"trait_key": "foo", "trait_value": "bar"}
```
emyller: Can we have an issue to cover this?

api/features/feature_states/serializers.py
```diff
@@ -0,0 +1,190 @@
...
+from segments.models import Segment
+
+
+class BaseFeatureUpdateSerializer(serializers.Serializer):  # type: ignore[type-arg]
+    @property
+    def environment(self) -> Environment:
+        environment: Environment | None = self.context.get("environment")
+        if not environment:
+            raise serializers.ValidationError("Environment context is required")
+        return environment
```
emyller, line 21: I feel like this could be a mixin to reuse in many other views, though I appreciate it's out of scope.

api/features/feature_states/serializers.py
```diff
@@ -0,0 +1,190 @@
...
+                int(string_val)
+            except ValueError:
+                raise serializers.ValidationError(
+                    f"'{string_val}' is not a valid integer"
+                )
+        elif value_type == "boolean":
+            if string_val.lower() not in ("true", "false"):
+                raise serializers.ValidationError(
+                    f"'{string_val}' is not a valid boolean (use 'true' or 'false')"
+                )
```
emyller, line 86: Can we have this logic abstracted? It is repeated but raising distinct exceptions, e.g. `ValueError` [there](https://github.com/Flagsmith/flagsmith/pull/6305/files#diff-61ee7b0f3b70a1171dc13b1992d9d5b736ada664b1ed3120ed5784ed329340d6), and `ValidationError` here.

api/tests/unit/features/feature_states/test_serializers.py
```diff
@@ -0,0 +1,154 @@
...
+        data={
+            "feature": {"name": feature.name},
+            "enabled": True,
+            "value": {"type": "string", "string_value": "test"},
+        },
+        context={},  # No environment
+    )
+    serializer.is_valid()
+
+    with pytest.raises(Exception) as exc_info:
```
emyller: Is it possible to narrow down this exception type?

api/tests/unit/features/feature_states/test_serializers.py
emyller, line 1: Missing `Given`/`When`/`Then` markers.

api/features/versioning/versioning_service.py
```diff
@@ -101,6 +105,308 @@ def get_current_live_environment_feature_version(
     )
 
 
+def update_flag(
```
emyller, line 109: These functions perform multiple database writes without `@transaction.atomic`. Can we wrap them?

api/features/versioning/versioning_service.py
```diff
@@ -101,6 +105,308 @@ def get_current_live_environment_feature_version(
...
+
+    _update_feature_state_value(
+        env_default_state.feature_state_value,
+        change_set.environment_default_value,
+        change_set.environment_default_type,
+    )
+
+    updated_segments = []
+    for override in change_set.segment_overrides:
+        segment_states = get_environment_flags_dict(
```
emyller, line 361: `get_environment_flags_dict` is called once per segment override. Can we batch this into a single query?

api/features/versioning/versioning_service.py
```diff
@@ -101,6 +105,308 @@ def get_current_live_environment_feature_version(
...
+
+def _update_flag_v2_for_versioning_v1(
+    environment: Environment, feature: Feature, change_set: FlagChangeSetV2
+) -> dict:  # type: ignore[type-arg]
+    env_default_states = get_environment_flags_dict(
+        environment=environment,
+        feature_name=feature.name,
+        additional_filters=Q(feature_segment__isnull=True, identity_id__isnull=True),
+    )
+    assert len(env_default_states) == 1
```
emyller, line 346: `assert` provides no message when it fails. Can we raise an explicit exception instead?

api/tests/unit/features/feature_states/test_unit_feature_states_views.py
emyller, line 1: Can we have a test for concurrent modifications? This is relevant since atomicity is not implemented yet (see [other thread](https://github.com/Flagsmith/flagsmith/pull/6305/files#r2590985108)).

emyller (dismissed):
The experiment looks solid, despite the complexity brought by the 2x2 grid _experiment_ x _versioning version_. Few suggestions.

api/features/feature_states/serializers.py
gagantrivedi: This is intentional defense in depth. The serializer validates at the API boundary and returns 400 with user-friendly errors. The model validates at the data layer to prevent corruption if set_value() is called directly (e.g.,
  from management commands, tasks, or other code paths). Different layers appropriately raise different exceptions: ValidationError for API responses, ValueError for programmatic misuse.

api/features/versioning/versioning_service.py
gagantrivedi: We don't have any early exits or conditional paths here that could leave things in a bad state - it's just a straight line of create → update → save. If any step fails, it raises and stops.

  Also, if something does fail midway, we'd actually want to know about it. A 500 surfacing is useful signal - silently rolling back could hide bugs we should be investigating.

  Worth noting that FeatureState is one of our hottest tables, so holding transaction locks longer than needed would hurt concurrent traffic.

  Happy to add one if there's a specific consistency case I'm missing though!

api/tests/unit/features/feature_states/test_unit_feature_states_views.py
gagantrivedi: That’s a fair point regarding concurrency. However, I’m a bit hesitant to add a specific test for it right now. In my experience, tests requiring thread orchestration to hit precise race windows tend to be quite flaky and hard to maintain.

Also, looking at the logic, I’m not sure a test would give us much new signal. Since we aren't implementing strict locking:

    V2: Relies on versioning (last publisher wins).

    V1: Defaults to 'last write wins.'

Essentially, the test would just confirm that the last request overwrites the previous one, which is the expected behavior rather than a defect we need to catch. What do you think?

api/features/feature_states/serializers.py
gagantrivedi: Yeah, we should do that outside of this

api/features/versioning/versioning_service.py
gagantrivedi: We don't expect it to fail. The assert is only included to improve readability and document the code.

api/features/versioning/versioning_service.py
gagantrivedi: I'd prefer to do that once this API is out of the experimentation stage

api/environments/identities/models.py
gagantrivedi: I am not sure why it's part of this pr

gagantrivedi: here: https://github.com/Flagsmith/flagsmith/pull/6371

api/features/feature_states/views.py
```diff
@@ -0,0 +1,150 @@
...
+        204: openapi.Response(
+            description="Feature state updated successfully (no content returned)"
+        )
+    },
+    tags=["Experimental - Feature States"],
+)  # type: ignore[misc]
+@api_view(http_method_names=["POST"])
+@permission_classes([IsAuthenticated, EnvironmentUpdateFeatureStatePermission])
+def update_flag_v1(request: Request, environment_key: str) -> Response:
+    environment = Environment.objects.get(api_key=environment_key)
```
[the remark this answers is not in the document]
gagantrivedi: No, it should not be passed from the permission layer

api/features/feature_states/serializers.py
```diff
@@ -0,0 +1,190 @@
...
+class FeatureUpdateSegmentDataSerializer(serializers.Serializer):  # type: ignore[type-arg]
+    id = serializers.IntegerField(required=True)
+    priority = serializers.IntegerField(required=False, allow_null=True)
+
+
+class FeatureValueSerializer(serializers.Serializer):  # type: ignore[type-arg]
+    type = serializers.ChoiceField(
+        choices=["integer", "string", "boolean"], required=True
+    )
+    string_value = serializers.CharField(required=True, allow_blank=True)
```
khvn26: Besides the fact that the schema should already inform the field type, it looks like there is an entire section in the description for the new API dedicated to compensating for any confusion that might be induced by this naming choice. 

Consider renaming to `value`.

api/features/feature_states/views.py
```diff
@@ -0,0 +1,150 @@
...
+    - Feature must belong to the environment's project
+
+    **Value Format:**
+    - Always use `string_value` field (value is always a string)
+    - The `type` field tells the server how to parse it
+    - Available types: integer, string, boolean
+    - Examples:
+      - `{"type": "integer", "string_value": "42"}`
+      - `{"type": "boolean", "string_value": "true"}`
+      - `{"type": "string", "string_value": "hello"}`
```
khvn26: See [my other comment](https://github.com/Flagsmith/flagsmith/pull/6305/files#r2599543419) about this.

api/features/feature_states/views.py
```diff
@@ -0,0 +1,150 @@
...
+    operation_description="""
+    **EXPERIMENTAL ENDPOINT** - Subject to change without notice.
+
+    Updates a single feature state within an environment. You can update either:
+    - The environment default state (when no segment is specified)
+    - A specific segment override (when segment.id is provided)
+
+    **Feature Identification:**
+    - Use `feature.name` OR `feature.id` (mutually exclusive)
+    - Feature must belong to the environment's project
```
khvn26: This line feels extra. IMO it conveys a very obvious piece of information.

api/features/feature_states/views.py
```diff
@@ -0,0 +1,150 @@
...
+            required=True,
+        )
+    ],
+    request_body=UpdateFlagSerializer,
+    responses={
+        204: openapi.Response(
+            description="Feature state updated successfully (no content returned)"
+        )
+    },
+    tags=["Experimental - Feature States"],
```
khvn26: We don't use tags in our Swagger docs, AFAIK. If we choose to use them, consider picking something more concise.

```suggestion
    tags=["experimental", "feature states"],
```

api/features/feature_states/views.py
```diff
@@ -0,0 +1,150 @@
...
+    if environment.is_workflow_enabled:
+        raise PermissionDenied(
+            "This endpoint cannot be used when change requests are enabled. "
+            "Use the change request workflow instead."
+        )
+
+
+@swagger_auto_schema(
+    method="post",
+    operation_summary="Update single feature state (V1)",
```
khvn26: I think the `V1` part will confuse people.

```suggestion
    operation_summary="Update single feature state",
```

api/features/feature_states/views.py
```diff
@@ -0,0 +1,150 @@
...
+            required=True,
+        )
+    ],
+    request_body=UpdateFlagV2Serializer,
+    responses={
+        204: openapi.Response(
+            description="Feature states updated successfully (no content returned)"
+        )
+    },
+    tags=["Experimental - Feature States"],
```
khvn26:
```suggestion
    tags=["experimental", "feature states"],
```

or

```suggestion
```

api/features/feature_states/views.py
```diff
@@ -0,0 +1,150 @@
...
+    - Feature must belong to the environment's project
+
+    **Value Format:**
+    - Always use `string_value` field (value is always a string)
+    - The `type` field tells the server how to parse it
+    - Available types: integer, string, boolean
+    - Examples:
+      - `{"type": "string", "string_value": "production"}`
+      - `{"type": "integer", "string_value": "100"}`
+      - `{"type": "boolean", "string_value": "false"}`
```
khvn26: Same as [here](https://github.com/Flagsmith/flagsmith/pull/6305/files#r2599545382).

api/features/versioning/dataclasses.py
```diff
@@ -12,3 +21,47 @@ class Conflict(BaseModel):
...
+            self.user = request.user
+        elif hasattr(request.user, "key"):
+            self.api_key = request.user.key
+        else:
+            raise ValueError("Request user must be FFAdminUser or have an API key")
+
+
+@dataclass
+class FlagChangeSet(AuditFieldsMixin):
+    enabled: bool
```
khvn26: nit: I don't see much value in a flat structure for the change sets given they're internal DTOs. The following looks cleaner and clearer to me:

```suggestion
class FlagChangeSet:
    author: AuthorData
    enabled: bool
```

`AuthorData` could more explicitly express that it's an `EnvironmentFeatureVersion.publish` dependency:

```python
@dataclass
class AuthorData:
    published_by: "FFAdminUser | None" = None
    published_by_api_key: "MasterAPIKey | None" = None
```

(Who knows, we might even incorporate it into `publish` interface itself one day.)

api/features/versioning/dataclasses.py
```diff
@@ -12,3 +21,47 @@ class Conflict(BaseModel):
     @property
     def is_environment_default(self) -> bool:
         return self.segment_id is None
+
+
+@dataclass
+class AuditFieldsMixin:
+    user: FFAdminUser | None = field(default=None, kw_only=True)
+    api_key: str | None = field(default=None, kw_only=True)
```
khvn26: Is there a reason we're not using `api_key: MasterAPIKey | None` here?

api/features/feature_states/models.py
```diff
@@ -38,3 +38,36 @@ def value(self) -> typing.Union[str, int, bool]:
...
+                    raise ValueError(
+                        f"'{value}' is not a valid boolean (use 'true' or 'false')"
+                    )
+                typed_value = value.lower() == "true"
+                field = "boolean_value"
+                type_const = BOOLEAN
+            case _:
+                raise ValueError(
+                    f"'{type_}' is not a valid type (use 'string', 'integer', or 'boolean')"
+                )
```
khvn26, line 71: We could define a `Literal["string", "integer", "boolean"]` type, assign it to the `type_` argument here and the change set dataclasses' `type_` fields as well. If we use something like Pydantic's `TypeAdapter` to map incoming serializer data to the DTOs, we'll get a fair guarantee against invalid values here.

api/features/versioning/dataclasses.py
```diff
@@ -12,3 +21,47 @@ class Conflict(BaseModel):
...
+
+    def set_audit_fields_from_request(self, request: Request) -> None:
+        from users.models import FFAdminUser
+
+        if type(request.user) is FFAdminUser:
+            self.user = request.user
+        elif hasattr(request.user, "key"):
+            self.api_key = request.user.key
+        else:
+            raise ValueError("Request user must be FFAdminUser or have an API key")
```
khvn26: I feel this code belongs to a mapper.

api/features/versioning/versioning_service.py
```diff
@@ -101,6 +105,311 @@ def get_current_live_environment_feature_version(
...
+    new_version.publish(
+        published_by=change_set.user,
+        published_by_api_key=change_set.api_key,
+    )
+
+    return {
+        "environment_default": {"enabled": change_set.environment_default_enabled},
+        "segment_overrides": updated_segments,
+        "version": new_version.uuid,
+    }
```
khvn26: I couldn't find this return value used anywhere. Is this the response schema or a DTO? Can we define it (and avoid the `dict[str, Any]` type annotations), or remove it?

api/features/feature_states/serializers.py
```diff
@@ -0,0 +1,190 @@
...
+            raise serializers.ValidationError(
+                "Duplicate segment_id values are not allowed"
+            )
+
+        for segment_id in segment_ids:
+            self.validate_segment_id(segment_id)
+
+        return value
+
+    @property
```
khvn26, line 158: Using a property on serializer class for the mapped data is an interesting pattern, but I couldn't find examples of it anywhere else in the codebase. I appreciate the code is experimental, but, for a bit more consistency in the codebase, consider introducing a mapper function instead.

api/features/feature_states/serializers.py
```diff
@@ -0,0 +1,190 @@
...
+            return value
+
+        segment_ids = [override["segment_id"] for override in value]
+        if len(segment_ids) != len(set(segment_ids)):
+            raise serializers.ValidationError(
+                "Duplicate segment_id values are not allowed"
+            )
+
+        for segment_id in segment_ids:
+            self.validate_segment_id(segment_id)
```
khvn26, line 154: I guess we need a TODO to optimise this bit in the future as well.

api/features/feature_states/views.py
```diff
@@ -0,0 +1,150 @@
...
+    - Available types: integer, string, boolean
+    - Examples:
+      - `{"type": "integer", "string_value": "42"}`
+      - `{"type": "boolean", "string_value": "true"}`
+      - `{"type": "string", "string_value": "hello"}`
+
+    **Segment Priority:**
+    - Optional `segment.priority` field controls ordering
+    - If null/omitted, segment is appended to end
+    - Use specific priority value to reorder
```
khvn26:
```suggestion
    - Optional `segment.priority` field controls ordering
    - If field value is null or the filed is omitted, lowest priority is assumed
```

api/features/feature_states/views.py
```diff
@@ -0,0 +1,150 @@
...
+    in one operation.
+
+    **What You Can Update:**
+    - Environment default state (required)
+    - Multiple segment overrides (optional)
+    - Priority ordering for each segment
+
+    **Feature Identification:**
+    - Use `feature.name` OR `feature.id` (mutually exclusive)
+    - Feature must belong to the environment's project
```
khvn26: Same as [here](https://github.com/Flagsmith/flagsmith/pull/6305/files#r2599629263).

khvn26 (commented):
My comments are mostly around naming, typing, and documentation — the core service code looks solid to me; I feel like it could be useful to other Core API's mechanisms as well.

api/features/feature_states/serializers.py
gagantrivedi: Done

api/features/feature_states/views.py
gagantrivedi: Yeah, using only `experimental` now

api/features/versioning/dataclasses.py
gagantrivedi: I considered moving this to a mapper, but I think keeping from_request as a classmethod on AuthorData is the better design here: 
  1. Cohesion - The factory method ensures AuthorData is always constructed correctly. A separate mapper would allow callers to bypass it and construct invalid instances directly.
  2. Discoverability - When someone needs to create an AuthorData, they naturally look at the class itself. AuthorData.from_request(request) is immediately discoverable vs hunting for a mapper module.
  3. Precedent - This follows Python stdlib patterns like datetime.fromtimestamp() and dict.fromkeys() - factory methods that create instances from known sources belong on the class.
  4. Proportionality - A mappers.py file for a single 6-line function would be over-engineering. If we had multiple DTOs all needing request-to-DTO mapping, a mapper module would make more sense.
  5. Locality - The logic for extracting author info from a request is tightly coupled to what AuthorData represents. Keeping them together means the full picture is in one place.

  Happy to discuss further if you feel strongly about this!

api/features/feature_states/serializers.py
gagantrivedi: I think the property pattern is the right choice here, even if it's not an existing pattern in the codebase(also see: https://github.com/Flagsmith/flagsmith/pull/6305#discussion_r2601323929):

  1. Ownership - The serializer validates input and produces a domain object from it. The property is part of that job, not a separate concern.
  2. Encapsulation - A mapper would need intimate knowledge of the serializer's validated_data structure. The property keeps this knowledge where it belongs - with the serializer that defines that structure. If the schema changes, only the serializer needs updating.
  3. No current reuse need - If we had multiple places needing to build a FlagChangeSet from different sources, a mapper would make sense. Right now, it would be abstraction without benefit.

  On consistency - I'd argue consistency is valuable when it helps developers predict how code works, but following an existing pattern for its own sake can spread a suboptimal approach. If this pattern proves useful here, it could
  become the precedent for similar cases. The experimental nature of this code is actually a good opportunity to try a cleaner approach.

  Happy to discuss if you see specific benefits a mapper would provide here!

api/features/feature_states/serializers.py
gagantrivedi: done

api/features/versioning/dataclasses.py
gagantrivedi: Yeah, done

api/features/versioning/versioning_service.py
gagantrivedi: Yeah, have removed it

api/features/versioning/versioning_service.py
```diff
@@ -101,6 +106,293 @@ def get_current_live_environment_feature_version(
...
+        try:
+            target_feature_state: FeatureState = new_version.feature_states.get(
+                feature_segment__segment_id=change_set.segment_id,
+            )
+        except FeatureState.DoesNotExist:
+            feature_segment = FeatureSegment.objects.create(
+                feature=feature,
+                segment_id=change_set.segment_id,
+                environment=environment,
+            )
```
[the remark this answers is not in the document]
gagantrivedi: good bot

api/features/feature_states/serializers.py
khvn26: > 1. Ownership

Agree that mappers clash with serializers in that regard. Perhaps we need a better shared understanding of the boundaries here — I'd reserve serializers for API validation, and mappers for transformations required by further business logic. 

> 2. Encapsulation - A mapper would need intimate knowledge of the serializer's validated_data structure

Slightly easier to reason about once type arguments are used to describe it, see https://github.com/Flagsmith/flagsmith/blob/c234612034df197530a1c1fcbb46ed30fa549721/api/features/feature_health/serializers.py#L36-L51 for an example.

> 3. No current reuse need

Agreed.

> 4. The experimental nature of this code is actually a good opportunity to try a cleaner approach.

It looks fairly clean here, but the pattern would be too easy to abuse, IMO. DRF serializers are prone to complexity blowups due to the way DRF is designed. If anything, I'd encourage keeping new serializers declarative, and abstract custom logic out early. 

I'm resolving the conversation, however, as I appreciate we have to move fast with this code, and it's subject to future iterations.

api/features/versioning/versioning_service.py
khvn26: While I'm not sure `@transaction.atomic` necessarily means a _silent_ rollback, the rest of Gagan's arguments read sound to me.

api/features/feature_states/serializers.py
khvn26: If we lean into using types a little more (see [my other related comment](https://github.com/Flagsmith/flagsmith/pull/6305/files#r2599764782)), we can shift programmatic misuse detection left fairly reliably — and get rid of additional validation code.

api/features/feature_states/models.py
gagantrivedi: Have added a todo for this

api/features/versioning/dataclasses.py
khvn26: `AuthorData.from_request` looks better, sure.

api/features/versioning/dataclasses.py
khvn26: I see this is addressed.

khvn26 (approved):

PR #6311, opened 2025-11-20
docs: Add FAQ documentation file

### LINKED ISSUE #6323: Add FAQs to Docs
Add an FAQ section to our docs site to replace the current help.flagsmith.com.

robertnorrie:
Adding a placeholder FAQ

## Changes

Adding blank FAQ page.

N/A

### DIFF
--- a/docs/docs/support/FAQ
+++ b/docs/docs/support/FAQ
@@ -1,0 +1,7 @@
+---
+title: Frequently Asked Questions
+sidebar_position: 130
+sidebar_label: FAQ
+---
+
+


agregoryfs:
We want to seed the FAQ with a few questions from our help center that we'll transition to this page. We can probably tweak the content itself but the questions are below and the content can be found in this page - https://help.flagsmith.com/en/


Is there a user_created date
Can I create common User Traits for all users in the same environment?
Does the Open Source version of Flagsmith have a limit of requests per month?
Can the open source version be configured to connect to an LDAP server?
Does the Open Source version of Flagsmith have any restrictions?
Is Flagsmith SOC 2 certified?
How do I delete an Organization?
How do I downgrade to the Free Plan from a paid plan?
How do I delete an account?
Does FLagsmith have G SUite or does Flagsmith Support SSO?
How can I change the email associated with my account?
How do I update my payment details?
How do I add a new person to my organisation?
How do I view past invoices?
Is it possible to have a calendar based release using Flagsmith?
Is it possible to create two keys in one environment and have different permissions?
Is it possible to set a flag that has complete precedent over all users & segments?
In the React SDK, is it possible to separate the initialization, like in the Node.js API?
Is it possible to clone a flag?
When querying identities, is it possible to sort by creation date?
Is the function startListening(ms) something that can "auto-fetch"?
How do I identify an anonymous user? What about running AB tests with anonymous users?
Is there way to render traits in users list in your webUI
Can my application send multiple values with remote config?
Is there a way to await the response from the method bullettrain.identify()?
Can I create Environments using the API? Can I create a flag via an SDK?
Is there any way to export a config in Flagsmith?
Can I bulk change flags?
When I query for feature flags what method / API should I be using to pass the user email address?
Can I copy Features across Environments?
What is the difference between Feature Flag and Remote Config?
If a segment is 'off' what does that mean?
How can I check which users belongs to which Segment?
Can I view all the users in a segment?
Does the % split segment feature work for non-identified users?
Does Flagsmith allow me to incorporate user segmentation data (eg from mixpanel) to deploy flags based on user attributes?
When using a segment with "% split", would changing the value affect existing users?
Can I put a Segment inside a Segment?
Is there a way to combine two conditions with 'AND' for Segments?
What happens if I have the same user that is attached to two segments and have an opposite config on the same feature?
Will a segment override work if the feature is toggled off?
Is it possible to export the users to a .CSV or something similar?
What events can I send from Flagsmith into my own infrastructure?
How would I define an experiment for a subset of users based on existing traits in a CDP?
Is customer bucketing guaranteed in A/B tests over time to ensure their consistent system experience?
Where is Flagsmith hosted?
What is the official base URI for the API?
How can I check in Flagsmith how many requests I've send to the service?
Where can I find my Project ID?
Is there support for initializing a client that does not establish a network connection to Flagsmith?
Where can I get the Flagsmith token for an Environment?
Do you support Python?
Does the Flagsmith .net SDK work with Unity 3D?
When deploying with Docker, how do I configure SES credentials?

matthewelwell:
Completed in https://github.com/Flagsmith/flagsmith/pull/6402

PR #6361, opened 2025-12-02
refactor(features): migrate FeaturesPage to RTK Query and eliminate legacy stores

### LINKED ISSUE #5778: Migrate critical components: FeaturesPage.js
Critical component as main entrypoint of the app

State of the component:
- Class component
- 766 lines
- No typing and no hooks
- Uses HOC Providers

## Objectives
Ranked by priority
- Migrate it to a Functional component (along with TS)
- Remove the Providers injections to use RTK
- Break it in small / modular and dedicated components

talissoncosta:

## Summary

This PR completes the migration of FeaturesPage from legacy Flux stores to RTK Query with comprehensive refactoring, improved TypeScript type safety, better error handling, and cleaner documentation.

### 1. **Core RTK Query Migration**
- Migrated FeaturesPage from Flux (ProjectStore, AccountStore) to RTK Query
- Created `useFeatureListWithFilters` hook for filtered feature fetching with automatic API key to ID conversion
- Created `useProjectEnvironments` hook to replace ProjectStore functionality
- Removed FeaturesList wrapper component (functionality merged into FeaturesPage)
- Uses RouteContext exclusively for organization/project data

### 2. **Custom Hooks & Code Organization**
- **useFeatureFilters**: Manages filters with bidirectional URL synchronization
  - Optimized by removing unnecessary `useCallback` wrappers with empty dependencies
- **useToggleFeatureWithToast**: Toggles features with toast notifications
- **useRemoveFeatureWithToast**: Removes features with toast notifications
- **useProjectEnvironments**: Provides project/environment data with accessor functions
- Extracted filter utilities to `featureFilterParams.ts` for global reuse
- Created comprehensive type definitions in `featureFilters.ts`

### 3. **UI/UX Improvements**
- **Error Handling**: Replaced custom error UI with reusable ErrorMessage component
  - Added user-friendly error context and actionable messages
  - Proper error state handling for failed data fetching
- **Loading States**: Added professional skeleton loader for features list
  - Supports reduced motion preferences
  - SCSS-based styling (removed styled-jsx)
- **Mutation State Tracking**: Added opacity feedback during toggle/remove operations
  - Tracks `isToggling` and `isRemoving` states from mutation hooks
  - Replicates legacy `isSaving` behavior

### 4. **Component Refactoring**
- **FeaturesPage.tsx**: 
  - Simplified display logic by removing `useFeaturePageDisplay` hook
  - Extracted nested ternary to `renderFeaturesList()` helper function
  - Removed inline render functions to named constants for better readability
- **FeaturesPageHeader**: Removed unused `showCreateButton` prop (was always true)
- **FeaturesTableFilters**: 
  - Fixed TypeScript issues (Row import, isLoading prop naming)
  - Removed readonly type constraint on VIEW_MODE_OPTIONS

### 5. **Type Safety Improvements**
- Standardized `projectId` type to `number` across all components
- Added explicit update body types following team patterns
- Fixed TypeScript errors in useFeatureList service
- Aligned pagination field naming with component expectations
- Added Environment tag type for cross-service cache invalidation
- Proper typing for filter states and URL parameters

### 6. **Documentation Cleanup**
- Removed ~150 lines of excessive JSDoc documentation across 7 files:
  - `useFeatureListWithFilters.ts`: 81→34 lines
  - `featureFilterParams.ts`: Simplified 4 JSDoc blocks
  - `useFeatureFilters.ts`: 26→3 lines
  - `useRemoveFeatureWithToast.ts`: 18→1 line
  - `useToggleFeatureWithToast.ts`: 18→1 line
  - `useProjectEnvironments.ts`: 25→1 line
  - `featureFilters.ts`: Simplified type comments
- Removed obvious inline comments that restated code
- Kept TypeScript types as primary documentation (self-documenting code)

### 7. **Performance Optimizations**
- Added memoization to prevent unnecessary re-renders
- Removed complex optimistic update logic (RTK Query handles this)
- Proper dependency arrays in hooks

### 8. **Code Quality**
- All commits follow conventional commit format (37 commits)
- Removed unused legacy types and props
- Standardized parameter naming across hooks
- Extracted pagination handlers
- Removed legacy Flux store ignore filter

### Created:
- `common/hooks/useFeatureListWithFilters.ts`
- `common/hooks/useProjectEnvironments.ts`
- `common/types/featureFilters.ts`
- `common/utils/featureFilterParams.ts`
- `web/components/pages/features/hooks/useFeatureFilters.ts`
- `web/components/pages/features/hooks/useToggleFeatureWithToast.ts`
- `web/components/pages/features/hooks/useRemoveFeatureWithToast.ts`
- `web/components/feature-summary/FeatureRowSkeleton.tsx`
- `web/styles/components/_feature-row-skeleton.scss`

### Modified:
- `web/components/pages/features/FeaturesPage.tsx` (major refactor)
- `web/components/pages/features/components/FeaturesPageHeader.tsx`
- `web/components/pages/features/components/FeaturesTableFilters.tsx`
- RTK Query service types and hooks

### Removed:
- `web/components/pages/features/components/FeaturesList.tsx`
- `web/components/pages/features/hooks/useFeaturePageDisplay.ts`
- Legacy Flux store dependencies

## Benefits

- ✅ **Zero legacy store dependencies** - Full RTK Query migration complete
- ✅ **Improved type safety** - Comprehensive TypeScript types throughout
- ✅ **Better UX** - Professional loading states, error handling, and user feedback
- ✅ **Cleaner codebase** - Removed excessive documentation, simplified logic
- ✅ **Better performance** - Automatic caching, deduplication, and optimized hooks
- ✅ **Maintainability** - Self-documenting code with clear types and patterns
- ✅ **Consistent patterns** - Follows modern React best practices
- ✅ **Accessibility** - Respects prefers-reduced-motion for animations

## Breaking Changes

None - all functionality preserved, this is a pure refactor.

### Manual Testing:

1. **Features Page Load**
   - ✅ Navigate to Features page
   - ✅ Verify features list loads with skeleton loaders
   - ✅ Check pagination works correctly
   - ✅ Verify proper error messages display on failure

2. **Filter Functionality**
   - ✅ Search filter
   - ✅ Tag filters (including empty tag mutual exclusivity)
   - ✅ Owner and group filters
   - ✅ Archived filter toggle
   - ✅ Enabled/disabled filter
   - ✅ URL parameter synchronization
   - ✅ Clear filters button

3. **Feature Actions**
   - ✅ Toggle feature on/off with visual feedback (opacity)
   - ✅ Delete feature with confirmation
   - ✅ Create new feature
   - ✅ Toast notifications appear correctly
   - ✅ Loading states during mutations

4. **Error Handling**
   - ✅ Network errors display user-friendly messages
   - ✅ Failed mutations show actionable error context
   - ✅ Error boundary catches component errors

5. **Type Safety**
   - ✅ No TypeScript errors in IDE
   - ✅ Full IntelliSense support
   - ✅ Proper type inference throughout

6. **Accessibility**
   - ✅ Skeleton animations respect prefers-reduced-motion
   - ✅ Proper ARIA labels maintained
   - ✅ Keyboard navigation works

### Validation:
- ✅ All 37 commits follow conventional commit format
- ✅ Code formatted and linted (passes lint-staged)
- ✅ No console errors or warnings
- ✅ Feature parity maintained with previous implementation


frontend/common/config.ts
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+export const config = {
+  /**
+   * Number of features to display per page in the features list.
+   */
+  FEATURES_PAGE_SIZE: 50,
+} as const
+
+export type Config = typeof config
```
talissoncosta: I am totally open to discuss a better placement for this constant

frontend/web/components/pages/features/components/FeaturesTableFilters.tsx
```diff
@@ -57,7 +45,7 @@ type FeaturesTableFiltersProps = {
 export const FeaturesTableFilters: FC<FeaturesTableFiltersProps> = ({
   filters,
   hasFilters,
-  isLoading,
+  isLoading: isLoadingProp,
```
talissoncosta: use isLoading

frontend/web/components/pages/features/hooks/useFeatureFilters.ts
```diff
@@ -0,0 +1,92 @@
...
+  const clearFilters = useCallback(() => {
+    history.replace(document.location.pathname)
+    const newFilters = getFiltersFromParams({})
+    setFilters(newFilters)
+    setPage(1)
+  }, [history])
+
+  const goToPage = useCallback((newPage: number) => {
+    setPage(newPage)
+  }, [])
```
talissoncosta: check

frontend/common/hooks/useFeatureListWithFilters.ts
```diff
@@ -0,0 +1,30 @@
+import { useMemo } from 'react'
+import { useGetFeatureListQuery } from 'common/services/useFeatureList'
+import { useProjectEnvironments } from './useProjectEnvironments'
+import { buildApiFilterParams } from 'common/utils/featureFilterParams'
+import type { FilterState } from 'common/types/featureFilters'
+
+/** Fetches filtered feature list with automatic API key to ID conversion. */
+export function useFeatureListWithFilters(
```
Zaimwa9: Would be nice to type it 👍

frontend/common/hooks/usePageTracking.ts
```diff
@@ -0,0 +1,78 @@
...
+ * @example
+ * ```tsx
+ * // Track once on mount
+ * usePageTracking(Constants.pages.FEATURES)
+ *
+ * // Track when environment changes
+ * usePageTracking(Constants.pages.FEATURES, [environmentId])
+ * ```
+ */
+export function usePageTracking(
```
Zaimwa9: This one is not used. I do feel explicitely passing the required deps will be more self-explanatory in the future. 

Wdyt about:
- merging the two of them
- pass the deps / context as a params object
- Add a boolean to save or not in the storage ?

frontend/common/hooks/useProjectEnvironments.ts
```diff
@@ -0,0 +1,54 @@
...
+
+/** Fetches project and environment data with accessor functions for API key lookups. */
+export function useProjectEnvironments(projectId: number): {
+  project: Project | undefined
+  environments: Environment[]
+  getEnvironmentIdFromKey: (apiKey: string) => number | undefined
+  getEnvironment: (apiKey: string) => Environment | undefined
+  isLoading: boolean
+  error: unknown
+} {
```
Zaimwa9: Do you mind extracting its type into an interface/type?

frontend/common/hooks/useProjectEnvironments.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  const getEnvironment = useCallback(
+    (apiKey: string): Environment | undefined => {
+      return environments.find((env) => env.api_key === apiKey)
+    },
+    [environments],
+  )
+
+  return {
+    environments,
```
Zaimwa9, line 51: This is an amazing addition thanks

frontend/common/services/useFeatureList.ts
```diff
@@ -0,0 +1,148 @@
...
+  store: any,
+  data: Req['getFeatureList'],
+  options?: Parameters<
+    typeof featureListService.endpoints.getFeatureList.initiate
+  >[1],
+) {
+  return store.dispatch(
+    featureListService.endpoints.getFeatureList.initiate(data, options),
+  )
+}
```
Zaimwa9: ❤️

frontend/common/services/useFeatureList.ts
```diff
@@ -0,0 +1,148 @@
...
+          { id: 'LIST', type: 'FeatureList' },
+          { id: 'METRICS', type: 'Environment' },
+        ],
+        query: (query: Req['deleteMultivariateOption']) => ({
+          method: 'DELETE',
+          url: `projects/${query.projectId}/features/${query.featureId}/mv-options/${query.mvId}/`,
+        }),
+      }),
+
+      getFeatureList: builder.query<Res['featureList'], Req['getFeatureList']>({
```
Zaimwa9: Ok so if my understanding is correct. This replaces the `getFeatures` action right? So next step would be to fully migrate the following to use this hook:
image

And we will be able to remove this function.

frontend/common/services/useFeatureList.ts
```diff
@@ -0,0 +1,148 @@
...
+    addTagTypes: [
+      'FeatureList',
+      'FeatureState',
+      'MultivariateOption',
+      'Environment',
+    ],
+  })
+  .injectEndpoints({
+    endpoints: (builder) => ({
+      createMultivariateOption: builder.mutation<
```
Zaimwa9: just to confirm. This one and the delete operation have been created but not used right ?

frontend/common/utils/featureFilterParams.ts
```diff
@@ -0,0 +1,127 @@
...
+export function hasActiveFilters(filters: FilterState): boolean {
+  return (
+    (filters.tags && filters.tags.length > 0) ||
+    filters.showArchived ||
+    (filters.search !== null && filters.search.length > 0) ||
+    filters.is_enabled !== null ||
+    (filters.value_search && filters.value_search.length > 0) ||
+    (filters.owners && filters.owners.length > 0) ||
+    (filters.group_owners && filters.group_owners.length > 0)
+  )
```
Zaimwa9:
```suggestion
  return !!(
    filters.tags?.length ||
    filters.search?.length ||
    filters.value_search?.length ||
    filters.owners?.length ||
    filters.group_owners?.length ||
    filters.showArchived ||
    filters.is_enabled !== null
  );
```
could work ?

frontend/common/utils/featureFilterParams.ts
```diff
@@ -0,0 +1,127 @@
...
+      if (params.is_enabled === 'false') return false
+      return null
+    })(),
+    owners:
+      typeof params.owners === 'string' && params.owners
+        ? params.owners
+            .split(',')
+            .filter((v) => v)
+            .map((v: string) => parseInt(v))
+        : [],
```
Zaimwa9: What do you think about using const for each of them and just return the object at the end to have a bit less of inline operations ?

frontend/web/components/feature-summary/FeatureRowSkeleton.tsx
```diff
@@ -0,0 +1,45 @@
+import React, { FC } from 'react'
+import type { CSSProperties } from 'react'
+
+interface FeatureRowSkeletonProps {
+  style?: CSSProperties
+}
+
+export const FeatureRowSkeleton: FC<FeatureRowSkeletonProps> = ({ style }) => {
```
Zaimwa9, line 8: Nice, just adding a screenshot for future readers
image

frontend/web/components/metrics/constants.ts
```diff
@@ -2,22 +2,22 @@ type ExtraMetrics = Record<string, { link?: string; tooltip?: string }>
 
 type GetExtraMetricsFunc = (
   projectId: string,
-  environmentApiKey: string,
+  environmentId: string,
```
Zaimwa9, line 5: Now I'm confused 😄  which one is `AXerjeUR..` and which one is `4`.
I tend to think this is the first one `environmentApiKey` right?

If yes, what do you think? Intuitively to avoid mixing them I would call it `environmentKey` and have `environmentId` with the number?

I'm honestly not opiniated on this but i'm curious to have your thoughts as I personally always mix them

frontend/web/components/pages/features/FeaturesPage.tsx
```diff
@@ -0,0 +1,333 @@
...
+  const isSaving = isToggling || isRemoving
+
+  usePageTrackingWithContext(
+    Constants.pages.FEATURES,
+    environmentId,
+    projectId,
+    routeContext.organisationId,
+  )
+
+  const newFlag = () => {
```
Zaimwa9: Could we rename it in `openNewFlagModal` maybe ?

Zaimwa9 (changes requested):
It's honestly amazing to see and read. Great job and thanks for the effort.

I particularly love seeing the new hooks and having all the critical logic gathered and clearly isolated.

I have a couple of comments that are clearly minor in comparison of the large size of this PR. Let's have the tests passing and will quickly have a ✅

frontend/web/components/metrics/constants.ts
talissoncosta: Yep, I was planning to discuss this with you.
I was also confused by this initially. The value we have on RouteContext is actually the environmentApiKey, which we use to fetch the environment. That environment then has a numerical ID, which is the proper environmentId.
However, I noticed we're currently treating the environmentApiKey as if it were the environmentId throughout our components and in RouteContext. To simplify things going forward, I decided to keep calling it environmentId in the components and abstract away the conversion step—we'll fetch the proper environmentId right after making the request within the hook.
This way, future developers working with this code won't need to think about the distinction between environmentApiKey and environmentId. They can simply pass what comes from RouteContext, and the hook will handle the necessary adjustments internally to make the query work correctly.

frontend/common/utils/featureFilterParams.ts
talissoncosta: Yep, sounds good, I will refactor this file

frontend/common/hooks/useFeatureListWithApiKey.ts
```diff
@@ -0,0 +1,40 @@
...
+    return buildApiFilterParams(
+      filters,
+      page,
+      environmentApiKey,
+      projectId,
+      getEnvironmentIdFromKey,
+    )
+  }, [filters, page, environmentApiKey, projectId, getEnvironmentIdFromKey])
+
+  return useGetFeatureListQuery(apiParams ?? skipToken, {
```
talissoncosta, line 35: ## Why `skipToken`?

We use RTK Query's `skipToken` to conditionally skip queries when parameters aren't ready yet:

```typescript
// frontend/common/hooks/useFeatureListWithApiKey.ts
const apiParams = useMemo(() => {
  if (!environmentApiKey || !projectId || !getEnvironmentIdFromKey) {
    return null  // Parameters not ready
  }
  return buildApiFilterParams(filters, page, environmentApiKey, projectId, getEnvironmentIdFromKey)
}, [filters, page, environmentApiKey, projectId, getEnvironmentIdFromKey])

return useGetFeatureListQuery(apiParams ?? skipToken, {
  refetchOnMountOrArgChange: true,
  skip: isLoadingEnvironments,
})
```

## The Problem

- `apiParams` can be `null` when environment data isn't loaded yet
- RTK Query hooks expect valid parameters or `skipToken`
- TypeScript would error if we pass `null` directly as query parameters

## Why Not Just Use `skip` Option?

```typescript
// ❌ This causes TypeScript errors
useGetFeatureListQuery(apiParams, {
  skip: !apiParams || isLoadingEnvironments
})
// Error: Type 'null' is not assignable to parameter
```

## Alternative We Avoided

```typescript
// ❌ Type assertion - exactly what we're trying to avoid!
useGetFeatureListQuery(apiParams as any, {
  skip: !apiParams || isLoadingEnvironments
})
```

## Benefits of `skipToken`

✅ **Type-safe** - No type assertions needed
✅ **Explicit intent** - Clear that query should be skipped
✅ **RTK Query pattern** - Official solution from Redux Toolkit docs
✅ **Clean code** - One line: `apiParams ?? skipToken`

## Reference

[RTK Query Documentation: skipToken](https://redux-toolkit.js.org/rtk-query/api/created-api/hooks#skiptoken)

frontend/web/components/metrics/constants.ts
talissoncosta: Great point about the naming confusion! You're absolutely right that this is confusing.

To clarify the current state:
- `environmentApiKey` (string like `AXerjeUR..`) - this is what RouteContext currently provides
- `environmentId` (number like `4`) - this is what the API actually needs

Your suggestion makes total sense: `environmentKey` and `environmentId` would be much clearer!

The reason I kept the current naming in this PR is to maintain consistency with the existing RouteContext while we complete this FeaturesPage migration. However, I think this would be an excellent candidate for a follow-up PR where we could:

1. Add both values to RouteContext: `environmentKey` (API key string) and `environmentId` (numeric ID)
2. Update all components to use the clearer naming
3. Remove the conversion step that's currently needed in the hooks

This would make the code much more intuitive and eliminate the wrapper hook entirely. What do you think?

frontend/common/hooks/usePageTracking.ts
talissoncosta: Sounds good, I will consolidate them

frontend/common/config.ts
kyle-ssg: Id say in the service file tbh, it's too much jumping around finding files

frontend/common/services/useFeatureList.ts
kyle-ssg: This already exists and is used in an increasing amount of places, see useProjectFlags. I don't think both of these should exists, this should just live in 1 place otherwise it's a bit confusing.

frontend/common/services/useFeatureList.ts
kyle-ssg, line 1: Inline with my other review in having this and useProjectFlags I think the naming of this file is a little inconsistent - not all of this is to do with a list of features

kyle-ssg (changes requested):

frontend/common/services/useFeatureList.ts
talissoncosta: I agree, @kyle-ssg . I’ll consolidate useFeatureList and useProjectFlags.

frontend/common/services/useFeatureList.ts
talissoncosta: I will remove them, let's add just when we need them properly.

talissoncosta:
Hey @kyle-ssg ! All your feedback has been addressed - ready for another look when you get a chance. Thanks!

frontend/common/services/useProjectFlag.ts
```diff
@@ -71,6 +132,33 @@ export const projectFlagService = service
...
+          { id: 'LIST', type: 'ProjectFlag' },
+          { id: 'LIST', type: 'FeatureList' },
+        ],
+        query: ({ flag_id, project_id }) => ({
+          method: 'DELETE',
+          url: `projects/${project_id}/features/${flag_id}/`,
+        }),
+      }),
+
+      updateFeatureState: builder.mutation<
```
kyle-ssg: This should be in useFeatureState (it might already be) - I think separating features from featurestates is very useful

frontend/global.d.ts
```diff
@@ -83,6 +83,28 @@ declare global {
     trackTraits: (traits: Record<string, any>) => void
     [key: string]: any
   }
+  const Utils: {
```
kyle-ssg: Hmm do we need this? Is this basically just for helping .js files in the meantime? 

Regardless I bet this could just be const Utils: typeof Utils  - no maintenance needed

frontend/web/components/base/PermissionGate.tsx
```diff
@@ -0,0 +1,95 @@
+import React, { FC, ReactNode } from 'react'
```
kyle-ssg: Can't we just adjust Permission.tsx to do all this ? I don't see why we need both. 

I think if we have both we'll start seeing some components use 1 vs the other, we run the risk of having too many similar components

frontend/web/components/feature-summary/FeatureRowSkeleton.tsx
kyle-ssg: Nice

frontend/web/components/metrics/constants.ts
kyle-ssg: Yeah this is an annoying result of all endpoints until recently needing api_key rather than id, we need to do a big refactor here! Maybe a standalone PR rather than here, environmentKey makes sense

frontend/common/services/useProjectFlag.ts
talissoncosta: Sounds good! thanks for pointing this out

frontend/global.d.ts
talissoncosta: Totally agree

frontend/web/components/base/PermissionGate.tsx
talissoncosta: Yep, sounds a really good plan. I will consolidate them

frontend/web/components/metrics/constants.ts
talissoncosta: Thanks for the extra context, @kyle-ssg . Yep, we can tackle this in a follow-up PR.

frontend/common/hooks/usePageTracking.ts
```diff
@@ -1,78 +1,84 @@
...
+            environmentId: context.environmentId,
+            orgId: context.organisationId,
+            projectId: context.projectId,
+          }),
+        )
+      }
     }
-  }, [pageName, environmentId, projectId, organisationId])
+    // We intentionally use individual properties instead of context object
+    // to prevent re-runs when object reference changes but values don't
```
Zaimwa9, line 76:
```suggestion
```

Zaimwa9 (commented):
Is it possible to rebase? I see a lot of diff from code already on main

frontend/common/hooks/usePageTracking.ts
talissoncosta: Done

Zaimwa9 (dismissed):
More than happy to approve

Zaimwa9 (approved):

kyle-ssg (approved):

PR #6371, opened 2025-12-05
fix(identit/models): update_traits typing to support dicts

gagantrivedi:

## Changes

The trait_value field in SDKTraitData now correctly accepts both:
- Structured SDKTraitValueData (from serializer validation via TraitValueField)
- Raw primitive values (str, int, bool, float) when called directly
- None (to delete traits)

This aligns the type definition with the actual runtime behavior of update_traits() and generate_traits() methods which handle both cases via Trait.generate_trait_value_data().

_Please describe._


api/environments/sdk/services.py
```diff
@@ -95,7 +95,7 @@ def get_transient_identifier(sdk_trait_data: list[SDKTraitData]) -> str:
     if sdk_trait_data:
         return hashlib.sha256(
             "".join(
-                f"{trait['trait_key']}{trait_value['value']}"
+                f"{trait['trait_key']}{trait_value['value']}"  # type: ignore[index]
```
matthewelwell, line 98: I appreciate that we're removing more `type: ignore`s than we're adding, but I'm interested why we have to add this one here?

khvn26: Looks like the type for `trait_value` is not considered narrowed to `TraitValue` at this point by Mypy. Usually that'd due to Mypy's shortcomings, but sometimes this means an actual bug. I'd prefer an expilicit cast here if we're looking at the former.

PR #6384, opened 2025-12-08
fix: prevent IDOR vulnerability in environment update endpoint

gagantrivedi:

## Changes

Make the `project` field read-only during environment updates to prevent  moving an environment to a different project 

The vulnerability allowed an attacker with access to their own environment to modify the `project` field in the PUT request body, effectively moving their environment into a victim's project.

Fix: Override __init__ in CreateUpdateEnvironmentSerializer to set project field as read-only when instance exists (update operation).

Adds unit test


api/environments/serializers.py
```diff
@@ -130,6 +130,12 @@ class Meta(EnvironmentSerializerWithMetadata.Meta):
             )
         ]
 
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)  # type: ignore[no-untyped-call]
+        # Prevent IDOR: project cannot be changed after creation
+        if self.instance is not None:
+            self.fields["project"].read_only = True
```
matthewelwell: A neater option here is probably just to separate the Create and Update serializers?

api/environments/serializers.py
```diff
@@ -130,25 +130,33 @@ class Meta(EnvironmentSerializerWithMetadata.Meta):
...
 
+
+class CreateEnvironmentSerializer(_BaseCreateUpdateEnvironmentSerializer):
     def get_subscription(self) -> Subscription | None:
         view = self.context["view"]
+        # handle `project` not being part of the data
+        # When request comes from yasg2(as part of schema generation)
+        project_id = view.request.data.get("project")
+        if not project_id:
+            return None
```
matthewelwell: Another (perhaps more descriptive) option here would be: 

```suggestion
        if getattr(self.request, "swagger_fake_view", False):
            return Subscription.objects.none()

        project_id = self.request.data["project"]
```

api/environments/serializers.py
```diff
@@ -130,25 +130,33 @@ class Meta(EnvironmentSerializerWithMetadata.Meta):
...
-        return None
+class UpdateEnvironmentSerializer(_BaseCreateUpdateEnvironmentSerializer):
+    class Meta(_BaseCreateUpdateEnvironmentSerializer.Meta):
+        read_only_fields = EnvironmentSerializerLight.Meta.read_only_fields + (  # type: ignore[assignment]
+            "project",
+        )
+
+    def get_subscription(self) -> Subscription | None:
+        if self.instance is None:
+            return None
```
matthewelwell: This should never be the case, right? Is this to handle swagger again? If so, I prefer the explicit approach I suggested above.

matthewelwell (dismissed):
Approving with a couple of minor comments.

api/environments/serializers.py
```diff
@@ -130,25 +130,32 @@ class Meta(EnvironmentSerializerWithMetadata.Meta):
...
         ]
 
+
+class CreateEnvironmentSerializer(_BaseCreateUpdateEnvironmentSerializer):
     def get_subscription(self) -> Subscription | None:
         view = self.context["view"]
+        if getattr(view, "swagger_fake_view", False):
+            return None
+
+        project_id = view.request.data["project"]
```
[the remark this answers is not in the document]
gagantrivedi: Nope, that is handled by the permission class

matthewelwell (approved):

PR #6401, opened 2025-12-12
fix: optimise segment deletion and cloning with bulk operations

gagantrivedi:

## Changes

Optimize segment deletion and cloning to reduce database queries from O(n) to O(depth), where n is the number of rules and conditions. This addresses high DB CPU from N+1 queries.

### Segment Deletion
- Add `segments/services.py` with optimized `delete_segment` function
- Use bulk UPDATE for soft-deleting rules/conditions instead of individual deletes
- Add async task for audit log creation to isolate from delete transaction
- Move `AuthorData` to `core/dataclasses.py` for reuse

### Segment Cloning
- Add `copy_segment_rules_and_conditions` to services.py
- Use `bulk_create` instead of individual saves for rules and conditions
- Process rules level-by-level to maintain parent relationships

### Query Count Improvements
| Operation | Before | After |
|-----------|--------|-------|
| Delete segment | O(n) | 26 queries (constant) |
| Clone segment | O(n) | 19 queries (constant) |
| Copy rules/conditions | O(n) | 10 queries (constant) |

Added unit tests with explicit query count assertions to verify O(depth) complexity.

### DIFF
--- a/api/core/dataclasses.py
+++ b/api/core/dataclasses.py
@@ -1,5 +1,3 @@
-from __future__ import annotations
-
 from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
@@ -12,11 +10,11 @@ if TYPE_CHECKING:
 
 @dataclass
 class AuthorData:
-    user: FFAdminUser | None = None
-    api_key: MasterAPIKey | None = None
+    user: "FFAdminUser | None" = None
+    api_key: "MasterAPIKey | None" = None
 
     @classmethod
-    def from_request(cls, request: Request) -> AuthorData:
+    def from_request(cls, request: "Request") -> "AuthorData":
         from users.models import FFAdminUser
 
         if type(request.user) is FFAdminUser:
--- a/api/core/dataclasses.py
+++ b/api/core/dataclasses.py
@@ -1,0 +1,27 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from rest_framework.request import Request
+
+    from api_keys.models import MasterAPIKey
+    from users.models import FFAdminUser
+
+
+@dataclass
+class AuthorData:
+    user: FFAdminUser | None = None
+    api_key: MasterAPIKey | None = None
+
+    @classmethod
+    def from_request(cls, request: Request) -> AuthorData:
+        from users.models import FFAdminUser
+
+        if type(request.user) is FFAdminUser:
+            return cls(user=request.user)
+        elif hasattr(request.user, "key"):
+            return cls(api_key=request.user.key)
+        else:
+            raise ValueError("Request user must be FFAdminUser or have an API key")
--- a/api/segments/services.py
+++ b/api/segments/services.py
@@ -1,4 +1,5 @@
 import typing
+import uuid
 
 from django.db import models, transaction
 from django.utils import timezone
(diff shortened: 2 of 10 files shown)


api/core/dataclasses.py
```diff
@@ -0,0 +1,27 @@
+from __future__ import annotations
```
matthewelwell: Do we need this? Isn't this only relevant if we want to run the code on python2?

api/segments/tasks.py
```diff
@@ -2,9 +2,37 @@
...
+def create_segment_deleted_audit_log(
+    project_id: int,
+    segment_name: str,
+    segment_id: int,
+    segment_uuid: str,
+    user_id: int | None,
+    master_api_key_id: int | None,
+    created_date: str,
+) -> None:
+    from django.utils.dateparse import parse_datetime
```
matthewelwell: Why does this need to be a local import?

api/segments/services.py
```diff
@@ -0,0 +1,80 @@
...
+    all_rule_ids = set(top_level_rule_ids)
+    current_level_ids = top_level_rule_ids
+
+    while current_level_ids:
+        nested_rule_ids = list(
+            SegmentRule.objects.filter(rule_id__in=current_level_ids).values_list(
+                "id", flat=True
+            )
+        )
+        if not nested_rule_ids:
```
matthewelwell: Feels like we're implementing 2 methods of breaking out of the loop? Perhaps unnecessarily?

api/tests/unit/segments/test_unit_segments_services.py
```diff
@@ -0,0 +1,267 @@
...
+        delete_segment(small_segment, author=author)
+    small_query_count = len(ctx_small.captured_queries)
+
+    reset_queries()
+    with CaptureQueriesContext(connection) as ctx_large:
+        delete_segment(large_segment, author=author)
+    large_query_count = len(ctx_large.captured_queries)
+
+    # Then
+    assert small_query_count == large_query_count
```
matthewelwell: Don't we know what the number is here too? Can't we do something like `small_query_count == large_query_count == x`?

api/segments/services.py
```diff
@@ -0,0 +1,80 @@
...
+            break
+        all_rule_ids.update(nested_rule_ids)
+        current_level_ids = nested_rule_ids
+
+    all_rule_ids_list = list(all_rule_ids)
+
+    with transaction.atomic():
+        FeatureSegment.objects.filter(segment_id__in=segment_ids).delete()
```
matthewelwell, line 64: I'm sure this has been considered, but it's a little odd that we hard delete the feature segments, but not the segment entities themselves. 

This will cascade a hard delete to the Feature states too, right?

gagantrivedi: Yes, this will cascade to hard delete FeatureStates too via the FK. FeatureSegment is not soft-deletable (it doesn't inherit from SoftDeleteExportableModel),

api/tests/unit/segments/test_unit_segments_services.py
```diff
@@ -0,0 +1,267 @@
...
+        delete_segment(small_segment, author=author)
+    small_query_count = len(ctx_small.captured_queries)
+
+    reset_queries()
+    with CaptureQueriesContext(connection) as ctx_large:
+        delete_segment(large_segment, author=author)
+    large_query_count = len(ctx_large.captured_queries)
+
+    # Then
+    assert small_query_count == large_query_count == 26
```
matthewelwell, line 180: Oooofff, this is still huge, why?

khvn26: I think the main outcome here is that it's constant.

gagantrivedi: Here:
```
  Queries 1-11 (11 queries): The actual delete_segment logic
  - 1: SELECT segment IDs (versions)
  - 2-5: SELECT rule IDs (4 levels of nesting traversal)
  - 6: SAVEPOINT
  - 7: SELECT FeatureSegments
  - 8: UPDATE conditions
  - 9: UPDATE rules
  - 10: UPDATE segments
  - 11: RELEASE SAVEPOINT

  Queries 12-26 (15 queries): The async audit log task running synchronously in tests!
  - 12: SAVEPOINT
  - 13: INSERT audit log
  - 14-25: Webhook/integration checks (datadog, new_relic, grafana, slack, etc.)
  - 26: RELEASE SAVEPOINT
  ```

Comment (matthewelwell) [reply api/tests/unit/segments/test_unit_segments_services.py:180]: Makes sense, thanks for the update. 

Can we add a comment to the test explaining this - just the top level info (i.e. "11 for the delete action, 15 for the audit log action handled synchronously in our test setup")?

**Review by matthewelwell — DISMISSED** [review]
Approved with a minor remaining comment.

File: api/tests/unit/segments/test_unit_segments_services.py (hunk 6 above)
Comment (gagantrivedi) [reply api/tests/unit/segments/test_unit_segments_services.py:180]: Done

**Review by khvn26 — APPROVED** [review]

File: api/segments/services.py [hunk 7]
```diff
@@ -0,0 +1,160 @@
...
+        Condition.objects.bulk_create(
+            [
+                Condition(
+                    uuid=uuid.uuid4(),
+                    rule=rule_id_to_cloned_rule[c.rule_id],
+                    operator=c.operator,
+                    property=c.property,
+                    value=c.value,
+                    description=c.description,
+                    created_with_segment=True,
```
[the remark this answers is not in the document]
gagantrivedi: It's used for audit logs, and it is correct the behaviour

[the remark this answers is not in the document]
khvn26: We don't need to preserve this information when cloning segments.

PR #6404, opened 2025-12-13
fix: prevent infinite loading on Identity Overrides tab

### LINKED ISSUE #6403: Identity Overrides tab hangs indefinitely despite UI showing user has required permissions

- [ ] Self Hosted with Docker
- [x] SaaS at flagsmith.com

### Describe the bug

Users are experiencing an infinite loading spinner on the Identity Overrides tab, even though the UI indicates they have the necessary permissions to access it. The tab appears unresponsive with no error message, creating confusion about whether this is a permissions issue, a UI bug showing incorrect permissions, or another type of system error.

### Steps To Reproduce

1. Navigate to a feature's Identity Overrides tab: Project X → Feature Y → Edit feature → Identity overrides tab
2. Access the page a regular user account
3. The regular user account should have `view identities` permission for the target project/env. 
4. The spinner appear and load indefinitely despite apparent permissions

### Expected behavior

One of the following should occur:

If the user lacks permissions: The UI should show a permissions error or hide the tab. 
If the user has permissions: The Identity Overrides data should load successfully

### Screenshots

_No response_

talissoncosta:
Closes #6403

## Summary

Customers experience an infinite loading spinner on the **Identity Overrides** tab when editing a feature flag.

**Affected Users:** All SaaS customers (Edge API is enabled by default)

**Reported by:** xapo, rudderstack (via Slack)

### 1. Missing Error Handling
**File:** `frontend/web/components/modals/CreateFlag.js` **Method:** `userOverridesPage()`

The loading state is controlled by:
```javascript
isLoading={!this.state.userOverrides}
```

When API calls fail, there's **no error handling** to set `userOverrides`, so it stays `undefined` and the loader never stops.

### 2. Wrong Admin Check (causing infinite loading for environment admins)

The permission check was using `AccountStore.isAdmin()` which checks **organization-level** admin, but the `my-permissions` API returns **environment-level** admin status in `permissions.ADMIN`.

**The API response:**
```json
{"permissions":["MANAGE_IDENTITIES","UPDATE_FEATURE_STATE"],"admin":true,"tag_based_permissions":[]}
```

Note that `admin: true` is returned, meaning the user is an **environment admin**. However, `VIEW_IDENTITIES` is not explicitly in the permissions array because admin status implies all permissions.

**The problem:** When checking permissions, the code used `AccountStore.isAdmin()` (org-level admin) instead of `permissions.ADMIN` (environment-level admin). This caused environment admins to fail the permission check, and since there was no `else` branch to set state, `userOverrides` stayed `undefined` → **infinite loading**.

## How to Reproduce (on main branch)

1. Create a user that is NOT an organization admin
2. Give the user environment-level admin permission (or just `MANAGE_IDENTITIES` without `VIEW_IDENTITIES`)
3. Login as that user
4. Navigate to any feature flag → Click "Identity Overrides" tab
5. **Result:** Infinite loading spinner that never resolves

### Error Handling
1. Add `.catch()` handlers to all API calls
2. Set `userOverridesError: true` on API failure
3. Set `userOverridesNoPermission: true` when user lacks permission or gets 403
4. Display appropriate feedback message for each scenario

### Permission Check
Replace `AccountStore.isAdmin()` (org-level) with `permissions.ADMIN` (environment-level):

**Before:**
```javascript
const hasViewIdentitiesPermission = permissions[VIEW_IDENTITIES]
if (!hasViewIdentitiesPermission && !AccountStore.isAdmin()) { ... }
```

**After:**
```javascript
const hasViewIdentitiesPermission = permissions[VIEW_IDENTITIES] || permissions.ADMIN
if (!hasViewIdentitiesPermission) { ... }
```

**User Feedback:**

| Scenario | Message |
| -------- | ------- |
| API failure | "Failed to load identity overrides." |
| No permission | "You do not have permission to view identity overrides." |
| No overrides | "No identities are overriding this feature." |

## Test plan

- [ ] Normal flow works - Identity overrides load correctly
- [ ] User without VIEW_IDENTITIES permission → Shows "You do not have permission to view identity overrides"
- [ ] Environment admin (not org admin) → Can view identity overrides
- [ ] Org admin → Can view identity overrides

## Risk Assessment

- **Low risk:** Additive error handling only
- **No breaking changes:** Success paths unchanged
- **Backwards compatible:** Only affects error cases

### DIFF
--- a/frontend/web/components/modals/CreateFlag.js
+++ b/frontend/web/components/modals/CreateFlag.js
@@ -113,6 +113,8 @@ const CreateFlag = class extends Component {
       scheduledChangeRequests: [],
       selectedIdentity: null,
       tags: tags?.filter((tag) => !hideTags.includes(tag)) || [],
+      userOverridesError: false,
+      userOverridesNoPermission: false,
     }
   }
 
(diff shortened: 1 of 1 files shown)


frontend/web/components/modals/CreateFlag.js
```diff
@@ -229,35 +231,79 @@ const CreateFlag = class extends Component {
...
+                    userOverrides: [],
+                    userOverridesError: true,
+                    userOverridesPaging: {
+                      count: 0,
+                      currentPage: 1,
+                      next: null,
+                    },
+                  })
+                })
+            } else {
```
Zaimwa9: Sorry to be picky. Can we early return those else instead `if (!hasViewIdentitiesPermission && !AccountStore.isAdmin()) {` and avoid an extra nesting? It would make it easier to read

frontend/web/components/modals/CreateFlag.js
```diff
@@ -229,35 +231,79 @@ const CreateFlag = class extends Component {
...
+              userOverrides: [],
+              userOverridesError: true,
+              userOverridesPaging: {
+                count: 0,
+                currentPage: 1,
+                next: null,
+              },
+            })
+          })
+      } else {
```
Zaimwa9: Same for this one

frontend/web/components/modals/CreateFlag.js
```diff
@@ -229,35 +231,79 @@ const CreateFlag = class extends Component {
...
-              )
-              .then((userOverrides) => {
-                this.setState({
-                  userOverrides: userOverrides.results.map((v) => ({
-                    ...v.feature_state,
-                    identity: {
-                      id: v.identity_uuid,
-                      identifier: v.identifier,
+        )
+          .then((permissions) => {
```
Zaimwa9: Ideally we could rewrite this as async/await but I don't want to delay the bug fix too much. Especially as Kyle's PR is refactoring this component

Zaimwa9 (changes requested):
Apologies. Just one changes to improve future maintenance. Otherwise functionally good, thanks for catching it (no pun intended)

Zaimwa9 (approved):

kyle-ssg:
Great find, this'll be so much more tidy when we componentise the identity tab :D

PR #6429, opened 2025-12-17
fix(features): restore FeaturesPage RTK Query migration with versioning fixes

talissoncosta:
## Summary

This PR reverts the revert (#6428) and then fixes the bugs found in the original FeaturesPage RTK Query migration (#6361).

### What happened
1. PR #6361 migrated FeaturesPage from legacy Flux stores to RTK Query
2. PR #6428 reverted it due to bugs in versioned environments
3. This PR re-applies the migration with proper fixes

#### 1. Versioned Environment Toggle (Row Switch)
**Problem:** Toggling features via the row switch in versioned environments failed with:
```
Cannot read properties of undefined (reading 'keyedEnvironmentFeatures')
```

**Root Cause:** The new RTK mutation `useToggleFeatureWithToast` didn't check `use_v2_feature_versioning` and always used the non-versioned API.

**Fix:** Added versioning check to use `useCreateAndSetFeatureVersionMutation` for versioned environments.

#### 2. Archived Features Appearing
**Problem:** Archived features appeared in the feature list when they shouldn't.

**Root Cause:** `is_archived` was `undefined` instead of explicitly `false`.

**Fix:** Explicitly set `is_archived: 'false'` in filter params.

### Files Changed
- `frontend/web/components/pages/features/FeaturesPage.tsx` - Initialize FeatureListStore for legacy modal compatibility
- `frontend/web/components/pages/features/hooks/useToggleFeatureWithToast.ts` - Handle versioned environments
- `frontend/common/services/useFeatureVersion.ts` - Add optimistic update and cache invalidation for versioned toggle
- `frontend/common/stores/feature-list-store.ts` - Add null checks for keyedEnvironmentFeatures
- `frontend/common/utils/featureFilterParams.ts` - Explicit is_archived=false
- `frontend/e2e/tests/versioning-tests.ts` - Add E2E test for row toggle in versioned env with endpoint verification
- `frontend/e2e/helpers.cafe.ts` - Fix toggleFeature selector bug (missing hyphen)

### E2E Test Coverage
The versioning tests now verify:
- Toggle OFF via row switch in versioned environment
- Toggle ON via row switch in versioned environment
- **Endpoint verification**: Confirms the versioned API (`POST /features/{id}/versions/`) is called instead of the non-versioned PUT endpoint
- API response verification via "Try it" button
- State persistence verification via page refresh

### Test Plan
- [x] E2E tests pass for versioning scenarios
- [x] Row toggle works in versioned environments
- [x] Modal save works in versioned environments
- [x] Archived features don't appear when filter is off
- [x] Versioned toggle uses correct API endpoint


frontend/e2e/tests/versioning-tests.ts
```diff
@@ -6,13 +6,24 @@ import {
...
     login,
+    toggleFeature,
     waitForElementVisible,
 } from '../helpers.cafe';
+import { RequestLogger, t } from 'testcafe';
 import { E2E_USER, PASSWORD } from '../config';
 
+// Request logger to verify versioned toggle uses the versions API endpoint
+// Versioned: POST /environments/{envId}/features/{featureId}/versions/
+const versionApiLogger = RequestLogger(/\/features\/\d+\/versions\/$/, {
```
Zaimwa9: From the [doc](https://testcafe.io/documentation/402769/reference/test-api/requestlogger/constructor#filter-with-a-predicate) we can use a predicate.

Maybe we could be a bit more precise and have a `req.url.match` and the method ? Maybe the body/id somehow if possible ?

frontend/e2e/tests/versioning-tests.ts
```diff
@@ -51,11 +62,90 @@ export default async () => {
...
+    log('Verify API returns enabled state')
+    await t.wait(2000)
+    await click('#try-it-btn')
+    await t.wait(1500)
+    text = await getText('#try-it-results')
+    try {
+      json = JSON.parse(text)
+    } catch (e) {
+      throw new Error('Try it results are not valid JSON')
+    }
```
Zaimwa9: Can we have a helper function here to avoid computations in-test maybe ?

frontend/e2e/tests/versioning-tests.ts
```diff
@@ -51,11 +62,90 @@ export default async () => {
...
+    // Verify: Versioned API endpoint was called again
+    log('Verify versioned API endpoint was called for toggle ON')
+    await t.expect(versionApiLogger.requests.length).gte(1, 'Expected versioned API to be called for toggle ON')
+
+    // Verify: Switch shows ON state on features list
+    await waitForElementVisible(byId('feature-switch-2-on'))
+
+    // Verify: API returns correct state (feature enabled)
+    log('Verify API returns enabled state')
+    await t.wait(2000)
```
Zaimwa9: Can we reduce the waiting to 500 ? This seems like a lot taking into account the natural buffer

frontend/common/services/useFeatureVersion.ts
```diff
@@ -112,8 +117,38 @@ export const featureVersionService = service
...
+            ),
+          )
+
+          try {
+            await queryFulfilled
+          } catch {
+            // Rollback on error
+            patchResult.undo()
+          }
+        },
```
Zaimwa9: This whole optimistic update block does not work unfortunately. Main reason is that in  `getFeatureList` we are using the `environmentId` and not `environmentApiKey`.

A couple of ways to check:
- Add throttling to the requests and see that the cache is not updated on flight
- Inspect the state of the RTK queries you will see that the cache key is composed by the `environmentId` + `projectId` and all the filters from the `GET Query` (`"getFeatureList({\"environmentId\":\"66291\",\"is_archived\":\"false\",\"page\":1,\"page_size\":50,\"projectId\":25594,\"sort_direction\":\"ASC\",\"sort_field\":\"name\",\"tag_strategy\":\"INTERSECTION\"})": { ... }`)

I believe the invalidateTags already handles the refresh of the UI so I wouldn't overcomplicate and just use a queryFn

Zaimwa9 (changes requested):
Some on-going comments

frontend/common/services/useFeatureVersion.ts
```diff
@@ -112,8 +117,38 @@ export const featureVersionService = service
       >({
         invalidatesTags: [
           { id: 'LIST', type: 'FeatureVersion' },
+          { id: 'LIST', type: 'FeatureList' },
```
Zaimwa9: NIT: We could be more precise here maybe so we don't have to refetch other environments request
`{ type: 'FeatureList', id: `${arg.projectId}-${arg.environmentId}` }`

Zaimwa9 (changes requested):
One additional comment

frontend/common/services/useFeatureVersion.ts
talissoncosta: Good catch! I agree!

frontend/e2e/tests/versioning-tests.ts
talissoncosta: Good point! I've updated the logger to use a predicate that checks both the URL pattern and the POST method. This makes it more precise and ensures we're only capturing the versioning API calls, not any other requests that might match the URL pattern.

frontend/e2e/tests/versioning-tests.ts
talissoncosta: Done! I've extracted this into parseTryItResults() in helpers.cafe.ts.

frontend/common/services/useFeatureVersion.ts
talissoncosta: You're right! I removed the broken optimistic update block. Instead, I've implemented optimistic updates at the component level using a simple useOptimisticToggle hook in FeatureRow. This gives instant UI feedback while the invalidatesTags handles the actual cache refresh. Much cleaner approach!

frontend/common/services/useFeatureVersion.ts
talissoncosta: Done - updated to use { type: 'FeatureList', id: \${arg.projectId}-${arg.environmentId}` }` for more targeted cache invalidation.

Zaimwa9 (approved):
👍 Nice, thanks! Let's coordinate to see how / when we release and test it straight away

PR #6447, opened 2025-12-30
fix(LD Importer): Archive deprecated LaunchDarkly flags on import

### LINKED ISSUE #5642: LD Importer: Tag deprecated flags as archived when importing into Flagsmith

LD has the concept of deprecated flags, that aren't fully archived and still in use for older versions of an application, but aren't included in lists of flags. This can cause confusion as it can appear that more flags are imported into Flagsmith than are active in LD. 

Deprecated flags should be tagged as archived to reduce confusion during the import process. 

Introduce a similar state as deprecated and tag them as such. 

### Additional context

_No response_

emyller:
Import deprecated flags from LaunchDarkly as archived features in Flagsmith.

## Changes

- Archive deprecated flags on import.
- Removed white text from "importing" overlay — it was very hard to read, and not useful.

I was also very annoyed that there is no user feedback after importing, and came to an improvement I like:

image

or

image

---

Closes #5642

Review effort: 2/5


docs/docs/administration-and-security/data-management/import-from-launchdarkly.md
```diff
@@ -65,3 +65,15 @@ Multivariate LaunchDarkly flags will be imported into Flagsmith as MultiVariate
...
+
+Archived flags in LaunchDarkly are imported as archived flags in Flagsmith.
+
+Deprecated flags in LaunchDarkly are also imported as archived flags in Flagsmith. They receive a "Deprecated" tag to distinguish them from flags that were archived in LaunchDarkly.
+
+:::warning
+
+In LaunchDarkly, deprecated flags continue to evaluate normally. In Flagsmith, archived flags return fallback values. Unarchive deprecated flags after import if older application versions still request them.
+
+:::
```
emyller: @matthewelwell FYI, this could invalidate this PR, or at least the approach we took so far. Let's discuss.

matthewelwell: > In Flagsmith, archived flags return fallback values.

I don't think that's true though is it?

emyller: No it's not. The engine does not have or receive any concept of archived features. I was fooled by a shallow AI assumption, I'm sorry.

khvn26 (dismissed):
LGTM!

emyller:
This is now unblocked by #6476.

frontend/web/project/toast.tsx
```diff
@@ -68,6 +68,7 @@ const Message: FC<MessageProps> = ({
 }) => {
   const theme = _theme || 'success'
   useEffect(() => {
+    if (!expiry) return
```
Zaimwa9: I see the intent here combined with the above but given the implementation, the toast doesn't unmount on navigation so it would stick forever.
I'd suggest to give it a high expiry time instead (20 second?) if the idea is to give the time to read

Image

Zaimwa9 (changes requested):
Looks good. I have one comment about the expiry that can lead to some a bit annoying UX.
And just putting it on the radar. I wouldn't do it in this PR and it doesn't look very prio but because the `archived` flag relies on a boolean flag (and not a tag as the other). We can not filter on `deprecated` alone.

We need to filter on both `archived` (to have the boolean) and `deprecated` to see them 

Image

Image

emyller:
> We need to filter on both `archived` (to have the boolean) and `deprecated` to see them

Great catch. I'd say this PR introduces poor UX without this addressed, or at least documented as a new issue. However, I don't see a good solution to be coupling the `Deprecated` tag **and** `archived` state together... Maybe the tag filter should include features regardless of their "archived" state if that makes sense. Let's discuss internally.

frontend/web/project/toast.tsx
emyller: Making the toast stick until manual dismissal was my actual intent here.

Picture it from the user's perspective: they're trying to move away from LaunchDarkly into Flagsmith, and they'll only probably use this tool once. Would you feel that this Success message should vanish on its own after a while?

Zaimwa9: I'll push back one last time but won't block the merge for this.

I agree that it should stay long enough to grab the information (that is somehow limited and not so critical) but it's not a pattern we have been using through the app. The presence of the flags are part of the success path too so I think from a user perspective it's already quite clear.
Ideally it should unmount on navigation but we don't have the implementation and as it's supposed to be a one shot action i don't think it's worth implementing now.

emyller: Thanks for the input. While I don't disagree with it (see 5618b9854fda2162bac08004b1f77dc8883a6dd0), I'd still like to see this success message persisted to a notification someday.

emyller:
> > We need to filter on both `archived` (to have the boolean) and `deprecated` to see them
> 
> Great catch. I'd say this PR introduces poor UX without this addressed, or at least documented as a new issue. However, I don't see a good solution to be coupling the `Deprecated` tag **and** `archived` state together... Maybe the tag filter should include features regardless of their "archived" state if that makes sense. Let's discuss internally.

I've decided to remove the newly introduced "Deprecated" tag entirely, see e5ab99999d44ef74551970fde5dbb3b5f8a1468d. Although it looks helpful upfront, it also creates an UX problem we don't need IMO.

Zaimwa9 (approved):
👍  Thanks. Good call I think for now to keep both of them archived. Hopefully it's sufficient for customers

PR #6452, opened 2025-12-31
deps: update to django5

matthewelwell:
## Changes

Depends on: 
 - ~~flagsmith/flagsmith-common#149~~
 - flagsmith/flagsmith-auth-controller#36

Upgrade to django 5. 

There were seemingly very few changes that were needed, and most of the updates in this PR relate to typing updates which I've tried to fix where possible but erred on the side of `type: ignore` comments where needed. 

The main breaking changes that required handling on our end were: 

 1. `Model.Meta.index_together` was removed. I've replaced usages of this with `models.Index(fields=[...])`. This change does result in a migration applied to the database in the form of a `RenameIndex` but from what I can tell, this is a very minor operation and doesn't require any locks on the data in the table, or the index, only on the metadata for the table. 
 2. `django.utils.timezone.utc` was removed. I've replaced this with `datetime.timezone.utc`, but since we still want to use `django.utils.timezone` for other logic, (e.g. `timezone.now()`), I've imported `datetime.timezone` as `dttz` where needed.

I also had to update a few dependencies to support the django upgrade: 

 1. django-debug-toolbar
 2. django-softdelete
 3. djoser

Note that the upgrade of django-debug-toolbar meant that I needed to remove `DEBUG = True` from the test settings (conversation on the actual LoC for clarification). 

Unit tests, and running locally.


api/app/settings/test.py
```diff
@@ -64,5 +64,3 @@
 """
 
 ENABLE_POSTPONE_DECORATOR = False
-
-DEBUG = True
```
matthewelwell, line 70: After the upgrade of django-debug-toolbar, this was causing issues because we only include the debug toolbar if `DEBUG is True` (see [here](https://github.com/Flagsmith/flagsmith/pull/6452/changes#diff-24cef548422510130cef32e1c377b0002efe3658bd99627fbb492b9e4aeab977R52)). As I understand it, it's because we don't also add it to `installed_apps` here which is now required (but seemingly wasn't in older versions of the package). 

I decided the better solution here was simply to remove the `DEBUG = True` as I'm not sure why it existed here for tests in the first place.

api/app_analytics/migrations/0007_rename_environment_id_created_at_index.py
```diff
@@ -0,0 +1,18 @@
...
+    dependencies = [
+        ("app_analytics", "0006_add_labels"),
+    ]
+
+    operations = [
+        migrations.RenameIndex(
+            model_name="apiusageraw",
+            new_name="app_analyti_environ_b61cad_idx",
+            old_fields=("environment_id", "created_at"),
+        ),
```
matthewelwell, line 17: Caused by deprecation of `Model.Meta.index_together`. I've researched and, as I understand it, it requires a lock on the metadata, but no the table or index itself. 

Since we don't reference the name of the index directly (although we might in specific oracle migrations 🤔) we shouldn't have any schema mismatch issues.

api/custom_auth/mfa/trench/command/remove_backup_code.py
```diff
@@ -11,6 +11,8 @@ def remove_backup_code_command(user_id: Any, method_name: str, code: str) -> Non
         .values_list("_backup_codes", flat=True)
         .first()
     )
+    if serialized_codes is None:  # pragma: no cover
+        return
```
matthewelwell, line 15: Added for typing reasons.

api/features/feature_health/services.py
```diff
@@ -145,7 +145,7 @@ def dismiss_feature_health_event(
     FeatureHealthEvent.objects.create(
         feature=feature_health_event.feature,
         environment=feature_health_event.environment,
-        type=FeatureHealthEventType.HEALTHY,
+        type=FeatureHealthEventType.HEALTHY.value,
```
matthewelwell, line 148: This one was somewhat odd. I can't see any indication of any changes to this behaviour in the notes [here](https://docs.djangoproject.com/en/6.0/releases/5.0/#backwards-incompatible-changes-in-5-0), but looking at this test I do think it was wrong previously anyway. 

The error that I was seeing was that it couldn't create the event with the type as the literal string `"FeatureHealthEventType.HEALTHY"` which makes sense - perhaps Django 5 is more strict on enforcing the `choices` behaviour. Either way, I think this change is a good step.

api/users/migrations/0045_add_through_fields_metadata_for_django_5_upgrade.py
```diff
@@ -0,0 +1,28 @@
+# Generated by Django 5.2.9 on 2025-12-31 15:38
+# Note: This migration is a no-op, and just adds the `through_fields` attribute
+# seemingly needed by the state in django5
```
matthewelwell, line 3: This migration is a no-op as per this comment.

api/users/models.py
```diff
@@ -105,6 +107,13 @@ def get_by_natural_key(self, email):  # type: ignore[no-untyped-def]
         # Used to allow case insensitive login
         return self.get(email__iexact=email)
 
+    def make_random_password(
+        self,
+        length: int = 10,
+        allowed_chars: str = string.ascii_letters + string.digits,
+    ) -> str:
+        return "".join(secrets.choice(allowed_chars) for _ in range(length))
```
matthewelwell, line 115: `make_random_password` was removed. I simply re-added it, using the guidance provided [here](https://docs.python.org/3/library/secrets.html#recipes-and-best-practices).

api/pyproject.toml
```diff
@@ -163,7 +162,7 @@ pygithub = "2.1.1"
 hubspot-api-client = "^8.2.1"
 djangorestframework-dataclasses = "^1.3.1"
 pyotp = "^2.9.0"
-flagsmith-common = "^2.2.7"
+flagsmith-common = { git = "https://github.com/flagsmith/flagsmith-common", branch = "deps/django-5-upgrade" }
```
matthewelwell: TODO: replace this with the new version once released.

matthewelwell (commented):
Minor comments for context where needed.

api/environments/identities/migrations/0006_rename_environment_created_date_index.py
```diff
@@ -10,9 +10,19 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.RenameIndex(
-            model_name="identity",
-            new_name="environment_environ_341dc9_idx",
-            old_fields=("environment", "created_date"),
-        ),
+        migrations.SeparateDatabaseAndState(
```
matthewelwell, line 13: Since we only added this index for postgres dbs (see [here](https://github.com/Flagsmith/flagsmith/blob/main/api/environments/identities/migrations/0002_alter_identity_index_together.py)), we only want to rename it for those databases too.

api/pyproject.toml
khvn26: I think this can be done now.

matthewelwell: It has been here: https://github.com/Flagsmith/flagsmith/pull/6452/commits/db8253051a342acc96a83e0e094e0cbabb4f3971

api/custom_auth/mfa/trench/command/remove_backup_code.py
khvn26: nit: This would self-document the hack and help avoid the pragma comment:

```suggestion
    if typing.TYPE_CHECKING and serialized_codes is None:
```

matthewelwell: Hmm... that does make sense, but doesn't seem to work? 

Image

khvn26: Oh well. Okay then.

khvn26: I guess another way to avoid the pragma comment is to use `.get`, which should also highlight any race issues?

matthewelwell: Since this is code imported from an external library (see [here](https://github.com/Flagsmith/flagsmith/pull/3988)), I'd rather not modify it without first improving the test coverage. I have tried to follow the code to see the entrypoints, and determine what the options are, but it requires more digging.  

My suggestion would be to accept the current hack + pragma.

khvn26 (approved):
LGTM 👍

PR #6468, opened 2026-01-05
fix: revert: "fix(features): restore FeaturesPage RTK Query migration with versioning fixes"

matthewelwell:
Reverts Flagsmith/flagsmith#6429


khvn26 (approved):
No idea what's going on; let's see if it helps.

PR #6471, opened 2026-01-06
fix: mv id missing in creating cr

Zaimwa9:

## Changes
- explicitely set id when passing multivariate options for update

- Against failing create-CR request

### DIFF
--- a/frontend/common/providers/FeatureListProvider.js
+++ b/frontend/common/providers/FeatureListProvider.js
@@ -1,6 +1,7 @@
 import React from 'react'
 import FeatureListStore from 'common/stores/feature-list-store'
 import ProjectStore from 'common/stores/project-store'
+import Utils from 'common/utils/utils'
 
 const FeatureListProvider = class extends React.Component {
   static displayName = 'FeatureListProvider'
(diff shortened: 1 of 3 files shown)


frontend/common/providers/FeatureListProvider.js
```diff
@@ -147,7 +147,20 @@ const FeatureListProvider = class extends React.Component {
       projectFlag,
       {
         ...environmentFlag,
-        multivariate_feature_state_values: flag.multivariate_options,
+        multivariate_feature_state_values: flag.multivariate_options?.map(
```
talissoncosta: What do you think about extracting this to a helper since it's repeated at line 232?

```tsx
const mapMvOptionsToStateValues = (options, existingValues) =>
  options?.map((v) => {
    const existing = existingValues?.find(
      (e) => e.multivariate_feature_option === v.id,
    )
    return {
      multivariate_feature_option: v.id,
      percentage_allocation:
        existing?.percentage_allocation ?? v.default_percentage_allocation,
    }
  })
```

talissoncosta (approved):

PR #6478, opened 2026-01-06
feat: improve segment feature association

### LINKED ISSUE #6473: Associated features show even for feature-specific segments

- [ ] Self Hosted with Docker
- [x] SaaS at flagsmith.com

### Describe the bug

Feature-specific segments should only show the feature they are associated with.

### Steps To Reproduce

View a feature-specific segment and go to the associated features tab

### Expected behavior

Feature-specific segments should only show the feature they are associated with.

### Screenshots

_No response_

### LINKED ISSUE #6474: Overrides should highlight features even if values are the same

- [ ] Self Hosted with Docker
- [x] SaaS at flagsmith.com

### Describe the bug

We don't differentiate default environment state  because no segment associated versus segment is associated but matches default state

### Steps To Reproduce

- Enable a flag and add an override that enables the flag
- It is not obvious that an override exists

### Expected behavior

Indicate that it's associated - "Is being overridden but currently matches environment defaults" 

### Screenshots

_No response_

kyle-ssg:

## Changes

Closes https://github.com/Flagsmith/flagsmith/issues/6474
Closes https://github.com/Flagsmith/flagsmith/issues/6473

- Feature-specific segments will now only show the feature that is relevant, it will open the feature in a new tab since removing segment overrides from here would push it into a weird state
- Segment overrides that appear to the be the same as environment defaults will now highlight

image

image

### DIFF
--- a/frontend/web/components/ConnectedFeatureOverrideRow.tsx
+++ b/frontend/web/components/ConnectedFeatureOverrideRow.tsx
@@ -1,0 +1,104 @@
+import React, { FC } from 'react'
+import { useGetProjectFlagQuery } from 'common/services/useProjectFlag'
+import { useGetFeatureStatesQuery } from 'common/services/useFeatureState'
+import FeatureOverrideRow from './feature-override/FeatureOverrideRow'
+import Utils from 'common/utils/utils'
+
+type ConnectedFeatureOverrideRowProps = {
+  featureId: number
+  projectId: number
+  environmentId: string
+  environmentIdNumeric: number
+  segmentId: number
+  shouldPreselect?: boolean
+  level?: 'segment' | 'identity'
+  index: number
+  openInNewTab?: boolean
+}
+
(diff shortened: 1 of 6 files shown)


frontend/web/components/ConnectedFeatureOverrideRow.tsx
```diff
@@ -0,0 +1,104 @@
...
+  environmentId: string
+  environmentIdNumeric: number
+  segmentId: number
+  shouldPreselect?: boolean
+  level?: 'segment' | 'identity'
+  index: number
+  openInNewTab?: boolean
+}
+
+const ConnectedFeatureOverrideRow: FC<ConnectedFeatureOverrideRowProps> = ({
```
Zaimwa9, line 19: Don't hate me please, do you think it would be possible to rename it to `FeatureSpecificOverrideRow` and maybe move it within the segments folder?

I'm just afraid that in a couple weeks the difference Associated/Connected isn't obvious

Zaimwa9 (approved):
Approving with a NIT that I leave up to you

PR #6504, opened 2026-01-08
fix: handle intermittent redis connection failures in throttling logic

matthewelwell:
## Changes

'Fixes' [this sentry issue](https://flagsmith.sentry.io/issues/6771071554/) by just handling the redis connection exception and blindly allowing the request instead. 

N/a - this is only some fallback logic based on an intermittent exception that we can only reproduce in production.

### DIFF
--- a/api/core/throttling.py
+++ b/api/core/throttling.py
@@ -1,7 +1,21 @@
 from django.conf import settings
 from django.core.cache import caches
+from redis.exceptions import RedisClusterException
 from rest_framework import throttling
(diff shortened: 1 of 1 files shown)


emyller (approved):
This looks like an OK workaround to ignore Redis connection issues. I like this adds resiliency but I'd be happier _not having to_ deal with this cache reliability issue. 

Approved but interested in learning of alternatives to approach this from the infrastructure side.

PR #6506, opened 2026-01-09
feat(redis): enable read_from_replicas for Redis Cluster by default

gagantrivedi:

## Changes

Adds `REDIS_CLUSTER_READ_FROM_REPLICAS` setting (default: `True`) to enable reading from Redis Cluster replica nodes.

This improves performance with ElastiCache Serverless by utilizing the reader endpoint on port 6380, distributing read traffic to replica nodes for better scalability and lower latency.

AWS best practices recommend this approach:
> Read from replicas – If you are using ElastiCache serverless or have provisioned read replicas (node-based clusters), direct reads to replicas to achieve better scalability and/or lower latency. Reads from replicas are eventually consistent with the primary.

Ref: https://docs.aws.amazon.com/AmazonElastiCache/latest/dg/WorkingWithRedis.html

- Unit tests added and passing
- Verified `env.bool()` correctly parses the environment variable
- Verified the setting is passed through to `RedisCluster` client via `read_from_replicas` kwarg

### DIFF
--- a/api/app/settings/common.py
+++ b/api/app/settings/common.py
@@ -852,6 +852,12 @@ DJANGO_REDIS_LOG_IGNORED_EXCEPTIONS = env.bool(
     "DJANGO_REDIS_LOG_IGNORED_EXCEPTIONS", True
 )
 
+# Enable reading from replicas in Redis Cluster mode.
+# Distributes read traffic to replica nodes (port 6380 on ElastiCache Serverless).
+REDIS_CLUSTER_READ_FROM_REPLICAS = env.bool(
+    "REDIS_CLUSTER_READ_FROM_REPLICAS", default=True
+)
+
 CACHES = {
     "default": {
         "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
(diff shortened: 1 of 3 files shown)


api/app/settings/common.py
```diff
@@ -852,6 +852,12 @@
     "DJANGO_REDIS_LOG_IGNORED_EXCEPTIONS", True
 )
 
+# Enable reading from replicas in Redis Cluster mode.
+# Distributes read traffic to replica nodes (port 6380 on ElastiCache Serverless).
+REDIS_CLUSTER_READ_FROM_REPLICAS = env.bool(
+    "REDIS_CLUSTER_READ_FROM_REPLICAS", default=True
+)
```
matthewelwell, line 859: What would be the downside to having this always enabled instead of configurable via a setting?

gagantrivedi: Someone start using it as not a cache, maybe?

matthewelwell: I don't follow, sorry. Can you share an example?

gagantrivedi: Strong consistency requirements - With read_from_replicas=True, reads may return slightly stale data due to replication lag. This is fine for caching, but having the option to disable it keeps the door open for use cases requiring stronger consistency. Or if in some environment, the lag is too much?

matthewelwell (approved):

PR #6513, opened 2026-01-12
feat(throttling): Throttle only requests authorised by master API keys

khvn26:

## Changes

In this PR, we limit throttling to requests authorised by master API keys. This helps us to be more flexible when configuring throttle rates without risking unexpected 429 responses in the UI.

Added unit tests; will test with a low rate limit on staging.

### DIFF
--- a/api/core/throttling.py
+++ b/api/core/throttling.py
@@ -5,14 +5,21 @@ from django.core.cache import caches
 from rest_framework import throttling
 from rest_framework.request import Request
 
+from api_keys.user import APIKeyUser
+
 if TYPE_CHECKING:
     from rest_framework.views import APIView
+
+
+class UserRateThrottle(throttling.UserRateThrottle):
+    cache = caches[settings.USER_THROTTLE_CACHE_NAME]
 
 
 class MasterAPIKeyUserRateThrottle(throttling.UserRateThrottle):
     cache = caches[settings.USER_THROTTLE_CACHE_NAME]
 
-    def allow_request(self, request: Request, view: "APIView") -> bool:
-        if not getattr(request.user, "is_master_api_key_user", False):
-            return True
-        return super().allow_request(request, view)
+    def get_cache_key(self, request: Request, view: "APIView") -> str | None:
+        if getattr(request.user, "is_master_api_key_user", False):
+            assert isinstance(request.user, APIKeyUser)
+            return str(request.user.key.id)
+        return None
--- a/api/core/throttling.py
+++ b/api/core/throttling.py
@@ -1,7 +1,18 @@
+from typing import TYPE_CHECKING
+
 from django.conf import settings
 from django.core.cache import caches
 from rest_framework import throttling
+from rest_framework.request import Request
+
+if TYPE_CHECKING:
+    from rest_framework.views import APIView
 
 
-class UserRateThrottle(throttling.UserRateThrottle):
+class MasterAPIKeyUserRateThrottle(throttling.UserRateThrottle):
     cache = caches[settings.USER_THROTTLE_CACHE_NAME]
+
+    def allow_request(self, request: Request, view: "APIView") -> bool:
+        if not getattr(request.user, "is_master_api_key_user", False):
+            return True
+        return super().allow_request(request, view)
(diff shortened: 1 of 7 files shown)


api/core/throttling.py
```diff
@@ -1,7 +1,18 @@
...
 from django.core.cache import caches
 from rest_framework import throttling
+from rest_framework.request import Request
+
+if TYPE_CHECKING:
+    from rest_framework.views import APIView
 
 
-class UserRateThrottle(throttling.UserRateThrottle):
+class MasterAPIKeyUserRateThrottle(throttling.UserRateThrottle):
```
gagantrivedi: Why are we renaming instead of creating a new class?

gagantrivedi (dismissed):

emyller (commented):
I'm curious about the rename from `UserRateThrottle` to `MasterAPIKeyUserRateThrottle`. Is throttling only necessary for master API key users?

api/core/throttling.py
khvn26: Good shout, addressed in 43c25b46a9f63b36b4248cde5f5cebf5a7b82e7d.

api/tests/unit/core/test_unit_core_throttling.py
```diff
@@ -0,0 +1,84 @@
...
+        "rest_framework.throttling.UserRateThrottle.allow_request",
+        return_value=False,
+    )
+
+    # When
+    result = throttle.allow_request(mock_request, mock_view)
+
+    # Then
+    mock_parent_allow_request.assert_not_called()
+    assert result is True
```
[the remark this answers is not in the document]
khvn26: Addressed in ee863fd5f03ccd61e876f4fae22718df94e69b16.

infrastructure/aws/production/ecs-task-definition-web.json
```diff
@@ -127,7 +127,7 @@
                 },
                 {
                     "name": "DEFAULT_THROTTLE_CLASSES",
-                    "value": "core.throttling.UserRateThrottle"
+                    "value": "core.throttling.MasterAPIKeyUserRateThrottle"
```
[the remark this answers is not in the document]
khvn26: image

emyller (dismissed):
LGTM

emyller (dismissed):

api/core/throttling.py
matthewelwell: The `UserRateThrottle` class only overrides the `get_cache_key` and scope attributes of the `SimpleRateThrottle` (see [here](https://github.com/encode/django-rest-framework/blob/main/rest_framework/throttling.py#L183-L202)), and in thinking about it, we might want to use a separate (e.g. `api_key`) scope here anyway. In that case, we should perhaps just subclass `SimpleRateThrottle` ?

khvn26: My intention was to inherit the scope, specifically. As I understand it, if we use a custom scope, we'd have to attach `MasterAPIKeyUserRateThrottle` manually to every view that relies on master API key authentication.

matthewelwell: No, that's the responsibility of the `DEFAULT_THROTTLE_CLASSES` that we're updating here already. The scope is such that we can set the rate - it's not inherently used by any authenticated request.

khvn26: Addressed in 684e47a9e20c1889c52ad12e0e898afcfb684738.

matthewelwell (dismissed):

api/core/throttling.py
```diff
@@ -1,7 +1,29 @@
...
+    from rest_framework.views import APIView
+
 
 class UserRateThrottle(throttling.UserRateThrottle):
     cache = caches[settings.USER_THROTTLE_CACHE_NAME]
+
+
+class MasterAPIKeyUserRateThrottle(throttling.SimpleRateThrottle):
+    scope = "master_api_key"
+    cache = caches[settings.USER_THROTTLE_CACHE_NAME]
```
matthewelwell, line 20: Sorry, should this change too? It probably doesn't matter what the cache is, right? And I really don't want another set of cache configuration settings...

khvn26: I don't think it should.

matthewelwell (approved):

PR #6515, opened 2026-01-12
fix: create change request with mv in v2

Zaimwa9:

## Changes
- Fix 400 on create-change-request 
- Fix percentage_allocation not being updated with feature value change request
- Fix percentage_allocation not being up with segment override during change request
- Fixed visual bug with number of changes in variation
- Fixed value missing in change request segment diff


talissoncosta:
I found an issue on creating a new flag with CR disabled.
Can you check if you are able to reproduce it as well ?

image

image

talissoncosta:
TEST V2 + CR

- [x] I can enable/disable my flag
- [x] I can update the value of my flag 
- [x] I can update the value of a MV option
- [x] I can update the weight of a MV option
- [x] I can create a MV option 
- [x] I can delete a MV option
- [x] I can add a feature to a release pipeline
- [x] I can create a Feature specific segment
- [x] I can attach a Feature specific segment
- [x] I can attach an existing segment
- [x] I can disable all segment
- [x] I can save my segments

talissoncosta (approved):
It looks great @Zaimwa9 . Thanks for handling that. 
I've been thought the flows that were failing and it seems to be all done.

PR #6524, opened 2026-01-13
fix(segments): prevent empty-rule matches

TimBeyer:
## Note
Reproduction and fix developed using Codex.
Take extra care when reviewing.

## Problem
Segment PATCH updates can briefly leave a segment with an empty rule tree. The `drf_writable_nested` update flow deletes nested rules/conditions before recreating them, and the flag engine treats empty rule sets as matching (because an empty conditions list evaluates to `True`). During this window, identities that should *not* match a segment can receive enabled feature overrides.

We reproduced this by looping real API calls (no monkeypatching): one thread continuously PATCHes the same segment rules while another polls identity feature states. The polling thread intermittently observes a non‑matching identity receiving an enabled feature, proving the race exists in production-like conditions.

## Reproduction
- Core API integration test added: `api/tests/integration/environments/identities/test_integration_segment_patch_atomic.py`
- The test runs for ~10s with two concurrent loops:
  - PATCH loop: repeatedly PATCHes the same segment ruleset to churn the nested updates.
  - Poll loop: repeatedly fetches identity feature states and records any unexpected enabled overrides.
- The test asserts there are **no** mismatches; it failed before the fix and passes after it.

## Solution
This change addresses the issue in two complementary ways:
1. **Atomic rule swap in `SegmentSerializer.update()`**
   - Update scalar fields first.
   - Create the new rule tree *before* deleting existing rules.
   - Delete old rules only after the new rules are safely present.
   - This removes the empty‑rule window during PATCH.
2. **Defensive mapping in the engine adapter**
   - When mapping `SegmentRule` models to the flag engine, empty rules (no conditions and no child rules) are filtered out.
   - This prevents any transient, empty rule from being treated as a match‑all rule if one slips through.
## Why this works
The fix ensures that segment evaluations never see a segment without valid rules during a PATCH, and the mapper guard provides extra safety in case of any future edge‑cases. The integration test now reliably passes and proves the race is resolved.
## Testing
- `DATABASE_URL=postgres://postgres:password@localhost:5432/flagsmith poetry run pytest tests/integration/environments/identities/test_integration_segment_patch_atomic.py`


TimBeyer:
Closing this for now since I'm not happy with mixing both fixes at once.
Feel free to re-use the reproduction though to get an idea of what you'd think would be the best way to fix this.

PR #6540, opened 2026-01-16
fix(CI): Run tests with private modules

emyller:
Cover testing Flagsmith with its private modules regardless of CI filtering — which might have existed for a reason, not pursued in this fix.

Also fixes [seeimgly] outdated tests, and updates Flagsmith packages with all the latest wiring.

PR #6541, opened 2026-01-16
deps: Update flagsmith-rbac

emyller:


emyller:
Will be superseded by https://github.com/Flagsmith/flagsmith/pull/6540

PR #6546, opened 2026-01-17
ci: Migrate to uv

khvn26:
## Why move to uv now

`api/` is a Poetry project with `package-mode = false`, so it isn't pip-installable. Downstream private packages (`flagsmith-private`, `flagsmith-saml`, `flagsmith-workflows`, ...) consume the api by cloning the repo and symlinking tests into it, each carrying its own copy of the api dependency graph.

Migrating to **uv + PEP 621 + setuptools** makes the api a real, pip-installable package. Downstreams can list it as a `[tool.uv.sources]` git dep and import normally.

## Testing gain

With the api installable, fixtures in `api/tests/conftest.py` are reachable from any consumer's venv. Consumers opt in with `pytest_plugins = ["tests.conftest"]` in their rootdir conftest and inherit every api fixture (`project`, `feature`, `environment`, ...) — no symlinks, no clones, no path tricks.

## Why the conftest refactor

`api/conftest.py` previously held hooks + fixtures. Two non-options for sharing it:

- Installing it as a top-level `conftest` py-module pollutes every downstream venv (collides with any project that has its own root `conftest.py`).
- An entry-point `pytest11` plugin doesn't work either: entry-point modules are imported before pytest-django's `tryfirst` Django setup, so module-level Django imports explode with `AppRegistryNotReady` (see [pytest #935](https://github.com/pytest-dev/pytest/issues/935)).

The fix: collapse hooks + fixtures into `api/tests/conftest.py` and have consumers opt in via `pytest_plugins`. `pytest_plugins` resolves *after* pytest-django's Django setup, so module-level Django imports are safe.

## Locked versions

Every dep version in `uv.lock` is byte-identical to `poetry.lock` on `main`. No silent upgrades.

## Test plan

- [x] API unit tests pass on 3.11 / 3.12 / 3.13
- [x] `codecov/patch` green
- [x] Downstream `flagsmith-private` consumes the new package and inherits fixtures via `pytest_plugins`


.github/workflows/api-deploy-production-ecs.yml
```diff
@@ -30,20 +30,18 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Install Poetry
-        run: make install-poetry
-
-      - name: Set up Python
-        uses: actions/setup-python@v5
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
```
gagantrivedi: The Makefile pins UV_VERSION = 0.8.14, and uv-pre-commit is pinned to 0.8.14, but astral-sh/setup-uv@v6 in the workflows isn't pinned to a uv version?

.pre-commit-config.yaml
```diff
@@ -37,7 +37,7 @@ repos:
       - id: python-typecheck
         name: python-typecheck
         language: system
-        entry: poetry -C api run mypy .
+        entry: bash -c 'cd api && uv run mypy .'
```
gagantrivedi: Can we not do something like `uv run --project api mypy .`

gagantrivedi (changes requested):

.github/workflows/api-deploy-production-ecs.yml
khvn26: Good catch. Pinned via `[tool.uv].required-version` in `api/pyproject.toml` (single source of truth), read by setup-uv via `version-file` and by the Makefile via tomllib. 6eb6ff9c1.

.pre-commit-config.yaml
khvn26: Improved in f8ab0e7f9.

gagantrivedi (approved):

Issue #6548, opened 2026-01-19
Github actions succeed even if deployment fails


matthewelwell:
Currently the deployment to ECS waits for the service to be stable before reporting a successful workflow, but this will still report as stable even if the health checks fail, and ECS rolls back.

matthewelwell:
Resolved in #6549

PR #6549, opened 2026-01-19
ci: add check for failed deployments

matthewelwell:
## Changes

Closes to https://github.com/Flagsmith/flagsmith/issues/6548

Verifies that the deployment successfully updated the task definition before exiting the workflow successfully. 

Added current branch to deploy to staging temporarily (see [here](https://github.com/Flagsmith/flagsmith/actions/runs/21145551540/job/60810269685)).


.github/workflows/api-deploy-staging-ecs.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - ci/add-check-for-failed-pipelines
```
emyller: Reminder to 🔪

emyller (changes requested):
Thanks, this looks good! Couple blockers.

matthewelwell:
The latest failure ([here](https://github.com/Flagsmith/flagsmith/pull/6549/commits/5ed772d7d6161ca8f6faeec555290252e5f4256b)) was because of a concurrency issue. I've added concurrency checks to the deployment jobs too.

.github/workflows/api-deploy-staging-ecs.yml
matthewelwell: This is done here: https://github.com/Flagsmith/flagsmith/pull/6549/commits/800cee01e7839091908a9ff0add26d486bb059d7

.github/workflows/api-deploy-production-ecs.yml
```diff
@@ -15,3 +15,6 @@ jobs:
     with:
       environment: production
     secrets: inherit
+    concurrency:
+      group: ${{ github.workflow }}
+      cancel-in-progress: true
```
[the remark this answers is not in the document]
matthewelwell: No, this is the behaviour that we want. The service is tolerant to inconsistencies like this, and we want to ensure that if we immediately trigger a second build, that the new build starts deploying immediately.

.github/workflows/api-deploy-staging-ecs.yml
```diff
@@ -16,3 +16,6 @@ jobs:
     with:
       environment: staging
     secrets: inherit
+    concurrency:
```
emyller: Non blocking: is there any reason why this is defined at job level rather than workflow?

emyller (dismissed):
LGTM with one non-blocking comment.

.github/workflows/api-deploy-staging-ecs.yml
matthewelwell: Good catch - there wasn't. I've moved it (and in doing so, invalidated your approval).

emyller (approved):

PR #6555, opened 2026-01-20
fix: v1-override-limit-includes-all-feature-segments-in-count

Zaimwa9:

## Changes
Calculation of the segment overrides (in v1) was counting unique segment ids. Leading to situations where you could have 10 segments and unlimited overrides.

- Calculate the limit based on all features states and not distinct segment ids

- Added tests, including non-regression tests for v2 extra count


gagantrivedi (dismissed):

Zaimwa9:
@gagantrivedi, rebased and staled your review, sorry 🙏

matthewelwell (approved):
Approving based on @gagantrivedi 's previous review and only a rebase / minor lint fix being added since.

PR #6562, opened 2026-01-20
chore: playwright

### LINKED ISSUE #5584: Migrate E2E tests to playwright

kyle-ssg:

  ## Changes

  ### Playwright Test Improvements

  #### Faster Tests
  - Inputs are really fast compared to TestCafe
  - Tests are passing in CI around 2m30s vs ~4m

  #### Better Local Development
  - Tests are re-runable via ui (you'll have to run npm run test:teardown before clicking play again)
  - Tests come with videos and clickable traces (these get uploaded from GitHub on failures too)
  - Tests can be ran very quickly and can be repeated (E2E_REPEAT=x) and exit on first failed test (E2E_RETRIES=0)

  image

  main_large

  #### Added Claude Commands
  - `/e2e` - Run all tests (OSS + Enterprise)
  - `/e2e-oss` - Run OSS tests, auto-fix failures, re-run until passing
  - `/e2e-ee` - Run Enterprise tests, auto-fix failures, re-run until passing
  - `/e2e-create` - Create a new test following existing patterns

Note: You can have the tests repeat to guarantee no flakiness 
  - /e2e 5 → Runs all tests with E2E_REPEAT=5 (runs once, then 5 more times if passing)
  - /e2e-oss 5 → Same for OSS tests only
  - /e2e-ee 5 → Same for enterprise tests only

  image

  #### HTML Reports
  - Interactive dashboard with search/filter
  - Test timeline and duration metrics
  - Detailed error messages with stack traces

  e.g. this is a downloaded report, we can access reports of failures as well as videos / interactive traces.

  image

  image

  image

  image

  #### CI/CD Features

  **GitHub Integration**
  - **Sticky PR comments** with failure summaries
  - **Direct links to ZIP artifacts** containing:
  - HTML report
  - Videos of failures
  - Interactive traces
  - Screenshots
  - Automatic artifact cleanup

  image

  image

  - No code changes, all e2e refactoring
  - Run local image and keep track of any failures

  image


frontend/Dockerfile.e2e
```diff
@@ -1,4 +1,8 @@
-FROM ghcr.io/flagsmith/e2e-frontend-base:latest
+FROM node:22-bookworm
```
kyle-ssg, line 1: It's better for firefox to be pulled via 

```RUN npx playwright install --with-deps firefox```

I'm not sure if that means we remove the e2e-frontend-base or run it there

Comment (kyle-ssg) [reply frontend/Dockerfile.e2e:1]: cc @khvn26 in case you had thoughts on this

File: frontend/e2e/teardown.ts [hunk 2]
```diff
@@ -0,0 +1,66 @@
...
+require('dotenv').config();
+
+export async function runTeardown(): Promise<boolean> {
+  console.log('\n\x1b[36m%s\x1b[0m\n', 'Running E2E teardown...');
+
+  const e2eTestApi = `${process.env.FLAGSMITH_API_URL || Project.api}e2etests/teardown/`;
+  const token = process.env.E2E_TEST_TOKEN
+    ? process.env.E2E_TEST_TOKEN
+    : process.env[`E2E_TEST_TOKEN_${Project.env.toUpperCase()}`];
+
```
Zaimwa9:
```suggestion
     console.log(`Teardown target host ${e2eTestApi}`)
```
That would make my life so much better as I am often switching dev/staging

frontend/README.md
```diff
@@ -1,3 +1,85 @@
...
+# Run only Enterprise tests
+npm run test -- --grep @enterprise
+```
+
+#### Environment Variables
+
+| Variable | Description |
+|----------|-------------|
+| `SKIP_BUNDLE=1` | Skip webpack build for faster iteration |
+| `E2E_CONCURRENCY=N` | Number of parallel workers (default: 20, use 1 for debugging) |
```
Zaimwa9, line 123: I'd add it to the command then in test:dev. That might be why right now my tests fail

frontend/E2E-LOCAL-TESTING.md
```diff
@@ -0,0 +1,103 @@
...
+## Quick Start
+
+From the `frontend/` directory:
+
+```bash
+# Run all tests in Docker (like CI)
+make test
+
+# Run specific tests
+make test opts="--grep @oss"
```
Zaimwa9: I'd separate into sub makefile commands:
"make test-oss"
"make test-with-enterprise" maybe as to avoid having to look for the docs when working locally

frontend/package.json
```diff
@@ -9,15 +9,17 @@
     "postinstall": "npm run env",
     "build": "npm run bundle",
     "husky:install": "cd ../ && husky install",
-    "test:bundle:staging": "cross-env E2E=1 ENV=staging npm run bundle",
-    "test:dev": "cross-env NODE_ENV=production E2E=true E2E_DEV=true ts-node -T ./e2e/index.cafe",
-    "test:devlocal": "cross-env NODE_ENV=production E2E_LOCAL=true E2E_CONCURRENCY=1 E2E=true E2E_DEV=true ts-node -T ./e2e/index.cafe",
+    "test": "npx tsx e2e/run-with-retry.ts",
```
Zaimwa9:
```suggestion
    "test": "npx -y tsx e2e/run-with-retry.ts",
```

frontend/E2E-LOCAL-TESTING.md
```diff
@@ -0,0 +1,103 @@
...
+SKIP_BUNDLE=1 make test opts="tests/flag-tests.pw.ts"
+
+# Verbose output
+VERBOSE=1 make test
+```
+
+### Keep Services Running
+```bash
+# Start services
+docker compose -f docker-compose-e2e-tests.yml up -d
```
Zaimwa9, line 54: I think the docker compose is not up to date and needs `FLAGSMITH_API_URL: flagsmith-api:8000/api/v1/`
(Currently `FLAGSMITH_API`)

frontend/playwright.config.ts
```diff
@@ -0,0 +1,102 @@
...
+    ],
+    ['json', { outputFile: './e2e/test-results/results.json' }],
+    ['list', { printSteps: false }], // Only shows test names with pass/fail status
+  ],
+  /* Retry on CI only */
+  retries: process.env.CI ? 2 : 0,
+  testDir: './e2e',
+  testMatch: /.*\.pw\.ts$/,
+  /* Test timeout */
+  timeout: 180000,
```
Zaimwa9: That might be a bit high no ?

frontend/e2e/global-teardown.playwright.ts
```diff
@@ -0,0 +1,97 @@
...
+  const dir = path.join(__dirname, 'test-results');
+
+  if (fs.existsSync(dir) && !process.env.E2E_DEV) {
+    try {
+      console.log('Uploading test artifacts...');
+      const files: string[] = [];
+
+      // Recursively find only failed test artifacts
+      const findFiles = (currentDir: string) => {
+        const items = fs.readdirSync(currentDir);
```
Zaimwa9: We should be able to use the `recursive: true` here that could replace the whole function

frontend/e2e/global-teardown.playwright.ts
```diff
@@ -0,0 +1,97 @@
...
+          await zipDirectory(reportDir, zipPath);
+          console.log('Uploading Playwright HTML report...');
+          await upload(zipPath);
+          console.log('Uploaded Playwright HTML report');
+          // Clean up zip file
+          fs.unlinkSync(zipPath);
+        } catch (e) {
+          console.log('Error uploading HTML report:', e);
+        }
+      }
```
Zaimwa9: Overall, i'd be in favor of taking the opportunity to simplify the understanding of the configuration and setup of the tests versus what we had with café.
Even though there are comments, we could abstract those blocks a bit more like this one being its own `uploadHTMLReport` function, similarly to this one `extractFailedTests`

frontend/e2e/extract-failed-tests.ts
```diff
@@ -0,0 +1,97 @@
...
+        // Use directory name as key - it usually contains the test file name
+        errorContextMap[entry.name] = errorContextPath;
+      }
+    }
+  }
+
+  return errorContextMap;
+}
+
+export function extractFailedTests(baseDir: string = __dirname): void {
```
Zaimwa9: What is the difference with the JSON.reporter defined here `['json', { outputFile: './e2e/test-results/results.json' }],` ?

It looks a bit redundant especially if playwright handle it ?

Otherwise could we simplify the nested filters/some with something like:
```
  const isFailedResult = (r: any) =>
    r.status === 'failed' || r.status === 'timedOut'

  const getFailedSpecs = (suite: any) =>
    suite.specs?.filter(spec =>
      spec.tests?.some(t => t.results?.some(isFailedResult))
    ) || []

  const failedTests = results.suites?.flatMap(suite =>
    getFailedSpecs(suite).map(spec => formatFailedSpec(spec, suite))
  ) || []
```

frontend/e2e/run-with-retry.ts
```diff
@@ -0,0 +1,147 @@
...
+    // Quote arguments that contain spaces or special shell characters
+    const quotedArgs = args.map(arg => {
+      if (arg.includes(' ') || arg.includes('|') || arg.includes('&') || arg.includes(';')) {
+        return `"${arg}"`;
+      }
+      return arg;
+    });
+    const playwrightCmd = ['npx', 'cross-env', 'NODE_ENV=production', 'E2E=true'];
+
+    // Skip cleanup on retries to preserve failed.json and test artifacts
```
Zaimwa9, line 24: Ok I guess that's where the extract-failed-test output is re-used ?

Zaimwa9: But how do we access it? Let's say in CI ?

frontend/e2e/global-setup.playwright.ts
```diff
@@ -0,0 +1,117 @@
...
+      console.error('\n', '\x1b[31m', errorMsg, '\x1b[0m', '\n');
+      if (process.env.E2E_LOCAL !== 'true' && process.env.E2E_DEV !== 'true') {
+        throw new Error(errorMsg); // Fail tests early in CI if teardown fails
+      }
+    }
+  } else {
+    // Only warn for local/dev testing, don't fail
+    if (process.env.E2E_LOCAL === 'true' || process.env.E2E_DEV === 'true') {
+      console.log('\n', '\x1b[33m', 'e2e teardown skipped (no token) - OK for local testing', '\x1b[0m', '\n');
+    } else {
```
Zaimwa9: I don't think it's ok actually 😄 and it should failed. Without teardown it's basically impossible to have trustable data/tests so we could directly throw

Zaimwa9: We can move everything up, early return and avoid a level of nesting

frontend/package.json
```diff
@@ -9,15 +9,17 @@
...
+    "test:run": "npm run test:bundle && cross-env NODE_ENV=production E2E=true npx playwright test",
+    "test:dev": "cross-env NODE_ENV=production E2E=true E2E_DEV=true npx playwright test --ui",
```
Zaimwa9: I would also set the concurrency default to 1 here directly as the local database pool doesn't handle it

frontend/package.json
```diff
@@ -9,15 +9,17 @@
...
+    "test:devlocal": "cross-env NODE_ENV=production E2E_LOCAL=true E2E_CONCURRENCY=1 E2E=true E2E_DEV=true npx playwright test --ui",
```
Zaimwa9: NIT: we could re-order the ENV values for all the commands in the same way

Zaimwa9 (commented):
Ok, it feels amazing but i'm not gonna lie. I feel a bit frustrated because I haven't been able to have all the tests passing.

As we discussed offline, there might be a layer of complexity coming from the number of environment variables and configuration flags that makes it harder.

I'll give it another try but some early comments

frontend/package.json
kyle-ssg: local handles 20 from what I've seen - all tests pass in around 40s

frontend/e2e/run-with-retry.ts
kyle-ssg: failed.json is for local development, this allows an ai to run 100x and keep reference to a file that points to the failures. This, as well as everything else gets uploaded to GitHub as an artifact.

frontend/e2e/extract-failed-tests.ts
kyle-ssg: The main difference is pulling out the failed tests so they can be read efficiently by ai - have cleaned up the functions

frontend/README.md
kyle-ssg: The tests shouldn't really fail due to high concurrency - Ive been running at 20

frontend/playwright.config.ts
kyle-ssg: This is for an entire test to complete, Ive seen 1.3 mins locally for some permissions to complete as its doing a lot of login / logout

frontend/e2e/global-teardown.playwright.ts
kyle-ssg: Have simplified all of this into a ci step

frontend/e2e/global-teardown.playwright.ts
kyle-ssg: Yeah you're right, I've done this and removed global-teardown.playwright.ts

frontend/web/main.js
```diff
@@ -12,6 +12,9 @@ import Project from 'common/project'
 import AccountStore from 'common/stores/account-store'
 import data from 'common/data/base/_data'
 
+// Set E2E flag default (overridden by Playwright in E2E tests)
+window.E2E = window.E2E || false
```
Zaimwa9: This should be removed otherwise out of E2E the app tries to access it while still undefined.

It works with importing first `import './project/e2e-init'`

Zaimwa9 (commented):
One comment that prevents the app to be run without E2E

frontend/web/main.js
kyle-ssg: Ah yep already fixed

kyle-ssg:
> One comment that prevents the app to be run without E2E

Yep I'd spotted this earlier, all sorted. I've just adjusted the tests so that PR comment uniqueness is done correctly, since we have 2 archs it runs under. we should see 4 passed comments shortly assuming one doesnt get a 429 rate limit.

frontend/docker-compose-e2e-tests.yml
```diff
@@ -46,7 +46,7 @@ services:
     environment:
       E2E_TEST_TOKEN_DEV: some-token
```
Zaimwa9, line 47: So using this compose solves most of the problems I encountered.
Here a 2 tweaks I had to do to make it work:
- Remove the image L16 (it uses an old image, so cloning segment wouldn't work because not in the image) or alternatively pointing to staging image. I prefer first version as it will use the local code
- change the container_name from `flagsmith_postgres` to `flagsmith_postgres_e2e` (just to avoid conflict with database created from the main docker-compose).

And rebuild the image on each changes.
My local setup is the main cause of the friction here but that would help working on e2e that also involves backend work

frontend/package.json
```diff
@@ -5,19 +5,19 @@
...
-    "test:devlocal": "cross-env NODE_ENV=production E2E_LOCAL=true E2E_CONCURRENCY=1 E2E=true E2E_DEV=true ts-node -T ./e2e/index.cafe",
-    "test:devBundle": "npm run test:bundle && npm run test:dev",
-    "test": "npm run test:bundle && cross-env NODE_ENV=production E2E=true ts-node -T ./e2e/index.cafe",
-    "test:staging": "npm run test:bundle:staging && cross-env NODE_ENV=production E2E=true ENV=staging ts-node -T ./e2e/index.cafe",
+    "test": "npx -y tsx e2e/run-with-retry.ts",
+    "test:run": "npm run bundle && npx playwright test",
+    "test:dev": "cross-env E2E_SHOW_BROWSER=true npx playwright test --ui",
+    "test:devlocal": "cross-env E2E_SHOW_BROWSER=true E2E_SKIP_RUN_FRONTEND=true npx playwright test --ui",
+    "test:install": "npx playwright install firefox",
+    "test:teardown": "npx tsx e2e/teardown.ts",
```
Zaimwa9, line 16:
```suggestion
    "test": "npx -y tsx e2e/run-with-retry.ts",
    "test:run": "cross-env E2E=true npm run bundle && npx playwright test",
    "test:dev": "cross-env E2E_SHOW_BROWSER=true npx playwright test --ui",
    "test:devlocal": "cross-env E2E=true E2E_SHOW_BROWSER=true E2E_SKIP_RUN_FRONTEND=true npx playwright test --ui",
    "test:install": "npx playwright install firefox",
    "test:teardown": "npx tsx e2e/teardown.ts",
```

frontend/webpack/plugins.js
```diff
@@ -5,7 +5,6 @@ const webpack = require('webpack');
 module.exports = [
 
     new webpack.DefinePlugin({
-        E2E: process.env.E2E,
```
Zaimwa9, line 8: I think we need it. I had to re-add it

frontend/e2e/run-with-retry.ts
```diff
@@ -0,0 +1,255 @@
...
+    // Add --quiet flag if QUIET is set
+    if (quietMode && !playwrightArgs.includes('--quiet')) {
+      playwrightArgs.push('--quiet');
+    }
+
+    // First attempt: build bundle and run tests
+    if (attempt === 0 && !process.env.SKIP_BUNDLE) {
+      if (!quietMode) console.log('Building test bundle...');
+      try {
+        execSync('npm run bundle', { stdio: quietMode ? 'ignore' : 'inherit' });
```
Zaimwa9:
```suggestion
        execSync('npm run bundle', { stdio: quietMode ? 'ignore' : 'inherit', env: { ...process.env, E2E: 'true' } });
```

frontend/README.md
```diff
@@ -1,3 +1,85 @@
...
+npm run test
+
+# Run with Playwright UI (for debugging)
+npm run test:dev
+
+# Run specific test file
+npm run test -- tests/flag-tests.pw.ts
+
+# Run only OSS tests
+npm run test -- --grep-invert @enterprise
```
Zaimwa9: Shouldn't it work with `--grep @oss` ?

frontend/README.md
```diff
@@ -1,3 +1,85 @@
...
+   ```
+
+#### Running Tests
+
+```bash
+# Run all tests
+npm run test
+
+# Run with Playwright UI (for debugging)
+npm run test:dev
```
Zaimwa9, line 106: I would just precise that you need to build the bundle first with `E2E=true npm run bundle` here

Zaimwa9 (commented):
I'm happy to see all the tests green on my side. I have a couple of comments coming from making it happen with my setup.
I now use the docker-compose-e2e-tests, it's way more convenient than jungling with api + task processor.

Almost good to go on my side once you confirm.

@talissoncosta, I think a second review is worth here, especially to get familiar with the new setup and confirm we are all able to get the E2E suite to run on our locals

frontend/docker-compose-e2e-tests.yml
Zaimwa9: image
For the record 😎

frontend/webpack/plugins.js
kyle-ssg: Nope we dont need this - why did you need to re-add it ?

frontend/README.md
kyle-ssg: We dont need E2E=true anymore

talissoncosta:
Great work on the Playwright migration @kyle-ssg ! The DX improvements (trace viewer, DOM snapshots, interactive UI mode) are huge.

I tested extensively using both `/e2e` and `make test`. Here are my findings:

#### Test Results

**Using `make test` with `--grep @oss`:**
```
10 passed ✅
```

**Using `make test` with `--grep @enterprise`:**
```
4 passed, 2 failed ❌ (change-request-test, roles-test)
```

**Using `/e2e` (all tests):**
```
14 passed, 2 failed ❌ (same tests)
```

#### Observations

**1. Enterprise tests fail against OSS API**

Two tests consistently fail locally:
- `change-request-test` → POST `/create-change-request/` returns **405**
- `roles-test` → POST `/roles/` returns **405**

**Root cause:** `docker-compose-e2e-tests.yml` uses `target: oss-api`, but these tests require enterprise-only endpoints.

**2. Concurrency mismatch**

- `.claude/commands/e2e.md` uses `E2E_CONCURRENCY=20` → causes browser crashes locally
- `Makefile` uses `E2E_CONCURRENCY=3` → stable

**3. `waitForTimeout()` anti-patterns**

Found instances in `e2e-helpers.playwright.ts`, though `e2e.md` states "NEVER use `page.waitForTimeout()`".

#### CI handles OSS/Enterprise correctly

Looking at `platform-pull-request.yml`:

| Job | API Image | Tests |
|-----|-----------|-------|
| `run-e2e-tests` | `oss-api` | `--grep @oss` |
| `run-e2e-tests-private-cloud` | `private-cloud-unified` | `--grep "@oss\|@enterprise"` |

#### Can't build private-cloud locally

Tried changing to `target: private-cloud-unified` but build fails - requires `github_private_cloud_token` for private repos.

#### Questions

1. **Is `--grep @oss` the intended way to run tests locally?**

2. **Should docs/commands default to OSS-only tests for local development?**

3. **Should we align concurrency settings?** (Makefile uses 3, /e2e command uses 20)

---

**None of these are blockers.** The PR is really good and ready to merge! I just wanted to raise these observations for awareness and to clarify the expected local development workflow.

talissoncosta (dismissed):

Zaimwa9 (dismissed):
🚀  🚀

talissoncosta (approved):

PR #6576, opened 2026-01-22
ci: fix workflow syntax

matthewelwell:
## Changes

Fixes syntax error in ECS deployment as seen [here](https://github.com/Flagsmith/flagsmith/actions/runs/21247633296).

Ran the workflow (with the actual deployment steps removed) - see [here](https://github.com/Flagsmith/flagsmith/actions/runs/21248039115/job/61142170606). Note that we only care that the job actually ran as that confirms the syntax error is resolved.


khvn26 (approved):
LGTM

emyller (approved):

PR #6603, opened 2026-01-27
fix(LaunchDarkly importer): Update API version to 20240415

### LINKED ISSUE #6476: LaunchDarkly import crashes due to API errors
Sentry Issue: [FLAGSMITH-API-5FH](https://flagsmith.sentry.io/issues/7167706691/?referrer=github_integration)

As of 2026-01-06, our LaunchDarkly importer fails with a 400 Bad Request hinting to an API change:

```
{"code": "invalid_request", "message": "The API version requested is not allowed"}
```

Our code uses LaunchDarkly API version 20220603 for this integration. Registering a new API token matching this version doesn't help. It looks like they have decided to advance EOL scheduled for [EOY 2026](https://launchdarkly.com/docs/guides/api/api-migration-guide)?

Upgrading the version number in the API (see https://github.com/Flagsmith/flagsmith/blob/fc4076f9e098e822516e4420916ca5c0076370e9/api/integrations/launch_darkly/constants.py#L2) leads to API incompatibility errors, which calls for efforts to update our code and match LD's latest API 20240415.

> [!NOTE]
> The import use case is known to be working until the day before, 2026-01-05, even with an API key of the newer LaunchDarkly API, 20240415.

Zaimwa9:

Closes #6476 

We migrate from LD API versions from `20220603` to `20240415`
Main differences:
- GET /api/v2/flags/{projectKey} max limit reduced from 1000 to 100
- environments are only returned when explicitly filtering by `env` query param (returned by default before)
- Related to the above, we have max 3 envs in query per request

## Changes
- Updated `LAUNCH_DARKLY_API_VERSION` to `20240415`
- Added constant for flags limit `LAUNCH_DARKLY_API_FLAGS_LIMIT_PER_PAGE = 100`
- Added environment_keys parameter to get_flags() to pass required env filter
- Request flags by batch of 3 environments if > to 3
- Fixed parsing for LD JSON values

- Curl requests 
```
  curl -X POST "http://localhost:8000/api/v1/projects/{PROJECT_ID}/imports/launch-darkly/" \
    -H "Authorization: Token {user_token_from_frontend}" \
    -d '{"token": "{LD_TOKEN}", "project_key": "default"}'
```
- Import from frontend


api/integrations/launch_darkly/client.py
```diff
@@ -172,15 +174,66 @@ def get_environments(self, project_key: str) -> list[ld_types.Environment]:
...
+        ):
+            return self._get_flags_by_envs(project_key, environment_keys)
+
         endpoint = f"/api/v2/flags/{project_key}"
+        params: dict[str, Any] = {
+            "summary": "0",
+            "limit": LAUNCH_DARKLY_API_FLAGS_LIMIT_PER_PAGE,
+        }
+
+        params["env"] = environment_keys
```
emyller: nit: Unless proposed code has meaning missing a comment?

```suggestion
        params: dict[str, Any] = {
            "summary": "0",
            "limit": LAUNCH_DARKLY_API_FLAGS_LIMIT_PER_PAGE,
            "env": environment_keys,
        }
```

api/integrations/launch_darkly/client.py
```diff
@@ -172,15 +174,66 @@ def get_environments(self, project_key: str) -> list[ld_types.Environment]:
...
+        """operationId: getFeatureFlags
+
+        :param environment_keys: List of environment keys to include configs for.
+            In API v20240415, the environments field is only returned when filtering.
+            Number of environment keys in one single request is limited to 3.
+        """
+        if (
+            environment_keys
+            and len(environment_keys) > LAUNCH_DARKLY_API_ENV_FILTER_MAX
+        ):
```
emyller: This signs at a couple things I think need addressing:

1. I no longer see `environment_keys=None` as a valid use case. We should accommodate that in the function definition and internal branching, expecting `environment_keys` to always be passed, and contain values in it.
2. The code is now duplicated, if I read this correctly: one path is taken when `len(environment_keys) > 3`, and another similar path otherwise. Assuming there will always be environment keys, I think we should unify paths and favour batching regardless of environment count.

api/integrations/launch_darkly/constants.py
```diff
@@ -1,8 +1,11 @@
...
-LAUNCH_DARKLY_API_VERSION = "20220603"
-# Maximum limit for /api/v2/projects/
-# /api/v2/flags/ seemingly not limited, but let's not get too greedy
+LAUNCH_DARKLY_API_VERSION = "20240415"
+# Maximum limit for /api/v2/projects/ and other endpoints
 LAUNCH_DARKLY_API_ITEM_COUNT_LIMIT_PER_PAGE = 1000
+# Maximum limit for /api/v2/flags/ is now 100 by launchdarkly API design
+LAUNCH_DARKLY_API_FLAGS_LIMIT_PER_PAGE = 100
+# Maximum number of environment keys allowed in a single request
+LAUNCH_DARKLY_API_ENV_FILTER_MAX = 3
```
emyller:
```suggestion
# Maximum limit for env query parameter values in /api/v2/flags/
LAUNCH_DARKLY_API_MAX_ENVIRONMENTS_PER_REQUEST = 3
```

api/tests/unit/integrations/launch_darkly/test_client.py
```diff
@@ -361,3 +362,137 @@ def test_launch_darkly_client__rate_limit_no_headers__waits_expected(
...
+    environment_keys = ["production", "test"]
+
+    mocker.patch(
+        "integrations.launch_darkly.client.LAUNCH_DARKLY_API_VERSION",
+        api_version,
+    )
+    mocker.patch(
+        "integrations.launch_darkly.client.LAUNCH_DARKLY_API_FLAGS_LIMIT_PER_PAGE",
+        100,
+    )
```
emyller: Aren't these built-in values? 🤔

api/tests/unit/integrations/launch_darkly/test_client.py
```diff
@@ -361,3 +362,137 @@ def test_launch_darkly_client__rate_limit_no_headers__waits_expected(
...
+    mocker.patch(
+        "integrations.launch_darkly.client.LAUNCH_DARKLY_API_ENV_FILTER_MAX",
+        3,
+    )
```
emyller: Same as [other thread](https://github.com/Flagsmith/flagsmith/pull/6603/changes#r2743605551).

emyller (changes requested):
The rationale looks solid, but the implementation needs attention in its design.

emyller:
Additionally, I think it is important that the PR title represents what this is, rather than how it was accomplished. Suggestion as follows:

> fix(LaunchDarkly importer): Update API version to 20240415

This is how it displays in the CHANGELOG, and I believe it looks clearer and more accurate than current:

> feat: added-limits-and-envs-in-ld-import-query-params

api/integrations/launch_darkly/client.py
```diff
@@ -172,17 +174,49 @@ def get_environments(self, project_key: str) -> list[ld_types.Environment]:
...
+                i : i + LAUNCH_DARKLY_API_MAX_ENVIRONMENTS_PER_REQUEST
+            ]
+            params = {**base_params, "env": batch}
+
+            flags: list[ld_types.FeatureFlag] = list(
+                self._iter_paginated_items(
+                    collection_endpoint=endpoint,
+                    additional_params=params,
+                )
             )
```
emyller: Can we not evaluate the list here? Items from `flags` are already going to be iterated over in the next block, in order to build `flags_by_key`.

api/integrations/launch_darkly/client.py
```diff
@@ -172,17 +174,49 @@ def get_environments(self, project_key: str) -> list[ld_types.Environment]:
...
-        )
+
+            for flag in flags:
+                key = flag["key"]
+                if key in flags_by_key:
+                    flags_by_key[key]["environments"].update(flag["environments"])
+                else:
+                    flags_by_key[key] = flag
+
+        return list(flags_by_key.values())
```
emyller: Probably also a good idea to skip evaluation here as well, given this will be just [iterated over](https://github.com/Flagsmith/flagsmith/blob/1c6901d87799f8d40b503692110e5a60e504bb0a/api/integrations/launch_darkly/services.py#L947) down the stack.

emyller (changes requested):
Looks much better, thank you!

I've observed a couple improvements we can make to performance and memory, which leads me to think all the import process could be improved too. Absolutely not a blocker, though I'm curious to how big a LD import can be — we'll wait and see I think.

emyller (approved):
LGTM

PR #6612, opened 2026-01-28
feat: Stream S3 organisation exports

khvn26:

## Changes

Contributes to #6610.

In this PR, we avoid OOM worker terminations when attempting to export beefy organisations via the sales dashboard.

We assume that using generators and multipart S3 uploads should significantly reduce the memory footprint. The tradeoff is that the exports should take slightly longer.

All the new code and changes to existing code are covered with tests. This will be tested manually on staging.


emyller (dismissed):
I'm interested in experimenting this right now.
When is the S3 object URL made available to the requester?

khvn26:
> When is the S3 object URL made available to the requester?

@emyller The current process is to retrieve it manually.

api/tests/unit/import_export/test_unit_import_export_utils.py
```diff
@@ -0,0 +1,204 @@
+from unittest.mock import MagicMock, call
+
+import pytest
+
+from import_export.utils import S3MultipartUploadWriter
+
+
+@pytest.fixture
+def s3_client() -> MagicMock:
+    client = MagicMock()
```
matthewelwell: Couldn't we use moto here to test this more realistically?

matthewelwell (commented):
Looks good on the whole - one question on the testing approach.

api/tests/unit/import_export/test_unit_import_export_utils.py
khvn26: Done in 8f8b385097a15532939ac5102c858a8d14ad66d2 (which includes a bugfix thanks to moto usage!)

api/tests/unit/import_export/test_unit_import_export_utils.py
```diff
@@ -0,0 +1,178 @@
...
+    )
+    s3_client = boto3.client("s3", region_name="eu-west-2")
+
+    # When
+    try:
+        with S3MultipartUploadWriter(s3_client, bucket_name, key) as writer:
+            writer.write(b"some data")
+            raise ValueError("test error")
+    except ValueError:
+        pass
```
emyller: Note: I'd see value in wrapping this under `pytest.raises`, as we're also validating `S3MultipartUploadWriter` re-raises.

api/import_export/utils.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    A file-like writer that streams data to S3 using multipart upload.
+
+    Buffers data until the minimum part size (5MB) is reached, then uploads
+    each part. This allows streaming large exports without holding the entire
+    payload in memory.
+    """
+
+    # S3 multipart upload minimum part size is 5MB (except for the last part)
+    MIN_PART_SIZE = 5 * 1024 * 1024
```
emyller: Is `MIN_PART_SIZE` = _every chunk size_?

IMO this could be a greater value that helps reducing network round trips without compromising the app's memory, e.g. 64MB:

- If export < 100MB, two chunks are good!
- Otherwise, a recent realistic example could go from 250 chunks to 20.

Also worth noting: our Customer Success KB recommends an ECS task with **a lot** of memory.

- [ ] Reminder: update it to spend less resources.

emyller (dismissed):
A couple non-blocking comments. Otherwise LGTM!

api/import_export/utils.py
khvn26: 71177827333e4e8b34d215b3ce9654e6ff1a4bfb

api/import_export/utils.py
```diff
@@ -0,0 +1,107 @@
...
+import typing
+
+if typing.TYPE_CHECKING:
+    from mypy_boto3_s3.client import S3Client
+    from mypy_boto3_s3.type_defs import CompletedPartTypeDef
+
+logger = logging.getLogger(__name__)
+
+
+class S3MultipartUploadWriter:
```
matthewelwell: I'm surprised we have to write this boilerplate code. Are there no third party options?

matthewelwell (dismissed):
Approved with a question.

api/import_export/utils.py
khvn26: Good shout, much simplified in 7e662d7b1.

emyller (approved):

api/tests/unit/import_export/test_unit_import_export_utils.py
khvn26: Not relevant anymore with the introduction of `smart_open`

PR #6642, opened 2026-02-02
fix: disable blanket API throttling in favour of WAF

gagantrivedi:

## Changes

Blanket API throttling is now managed through WAF. This PR:

- Sets default `USER_THROTTLE_RATE` and `MASTER_API_KEY_THROTTLE_RATE` to `None` in `api/app/settings/common.py`
- Removes `DEFAULT_THROTTLE_CLASSES`, `USER_THROTTLE_CACHE_*`, and `MASTER_API_KEY_THROTTLE_RATE` environment variables from staging and production ECS task definitions

Per-endpoint throttling (login, signup, mfa_code, invite, influx_query) remains unchanged.

Configuration change only - verified that:
- Default throttle rates are set to `None`
- Per-endpoint throttle rates remain configured
- ECS task definitions no longer include blanket throttle environment variables


khvn26 (approved):
LGTM 👍

api/app/settings/common.py
```diff
@@ -302,8 +302,8 @@
 
 LOGIN_THROTTLE_RATE = env("LOGIN_THROTTLE_RATE", "20/min")
 SIGNUP_THROTTLE_RATE = env("SIGNUP_THROTTLE_RATE", "10000/min")
-USER_THROTTLE_RATE = env("USER_THROTTLE_RATE", "500/min")
-MASTER_API_KEY_THROTTLE_RATE = env("MASTER_API_KEY_THROTTLE_RATE", USER_THROTTLE_RATE)
+USER_THROTTLE_RATE = env("USER_THROTTLE_RATE", default=None)
+MASTER_API_KEY_THROTTLE_RATE = env("MASTER_API_KEY_THROTTLE_RATE", default=None)
```
emyller, line 306: This effectively means disabling default throttles for `user` and `master_api_key` scopes, which are [still supported](https://github.com/Flagsmith/flagsmith/blob/aef974dd487959bd7bc51bf9afedd8980c8603ae/api/app/settings/common.py#L319-L327). This may or may not be relevant to on-premise customers. Can we keep the defaults here?

emyller (dismissed):

api/app/settings/common.py
gagantrivedi: Yes, we still support blanket throttling through the API layer - self-hosted/private cloud customers can configure `USER_THROTTLE_RATE` and `MASTER_API_KEY_THROTTLE_RATE` environment variables if they want to enable it.

However, I think setting a default value here is a bit opinionated. A rate like `500/min` may or may not make sense depending on the customer's infrastructure and usage patterns. If they want blanket throttling, they should choose a number that fits their specific needs rather than inheriting an arbitrary default.

emyller: > A rate like `500/min` may or may not make sense depending on the customer's infrastructure and usage patterns.

I agree, though on-premise users have already been inheriting this rate, and now we're removing it from their settings without notice. That said, I also appreciate this default rate realistically has little meaning in an on-premise context, and likely this will affect zero people. I'll leave this up to you. 😉

gagantrivedi: Yeah, I understand the concern. The decision to set a default rate was made during a rather desperate time and was a mistake in my opinion — an arbitrary value like `500/min` is more likely to confuse new users than help them. That said, the mistake wasn't made long ago, and as you noted, practically zero users are depending on this. I'd like to rectify it now before more users inherit a default that doesn't make sense for their setup.

emyller: Thanks for sharing the context! I'm on the same page now, and agree with this.

PR #6643, opened 2026-02-02
feat: Add Sentry to default integration data

### LINKED ISSUE #6644: Sentry integration missing from default integration list
## Summary

The Sentry change tracking integration was added in 2.184.0, but it was never added to the frontend's default `integration_data` in `default-flags.ts`. This causes self-hosted customers who don't have FoF configured (or whose FoF `integration_data` flag doesn't include Sentry) to not see the Sentry integration in the UI.

## Steps to Reproduce

1. Deploy a self-hosted Flagsmith instance (v2.184.0+) without FoF configured
2. Navigate to Project → Environment → Integrations
3. Sentry integration is not visible

## Expected Behaviour

Sentry should appear in the integrations list by default.

## Actual Behaviour

Sentry is not visible unless the customer has FoF configured with Sentry in their `integration_data` flag.

## Validation

To validate locally, temporarily modify `frontend/common/utils/utils.tsx` to force the fallback:

```typescript
getIntegrationData() {
  // Force fallback to default-flags.ts
  return defaultFlags.integration_data
},
```

- On `main`: Sentry is **not visible**
- With fix applied: Sentry is **visible**

talissoncosta:
## Summary

- Adds Sentry change tracking integration to the default `integration_data` in `default-flags.ts`
- The backend was added in 2.184.0, but was missing from the frontend defaults
- This caused self-hosted customers to not see the Sentry integration in the UI

## Context

A customer on version 2.184.1 reported they couldn't see the Sentry integration. Investigation revealed the Sentry integration backend exists but was never added to the frontend's default integration list.

## Test plan

- [ ] Verify Sentry appears in the integrations list on a self-hosted instance
- [ ] Verify the integration form shows webhook_url and secret fields
- [ ] Verify the integration can be configured and saved

### DIFF
--- a/frontend/common/stores/default-flags.ts
+++ b/frontend/common/stores/default-flags.ts
@@ -188,8 +188,7 @@ const defaultFlags = {
       'title': 'Segment',
     },
     'sentry': {
-      'description':
-        'Sends feature flag change events to Sentry for change tracking.',
+      'description': 'Send flag change events to Sentry.',
       'docs': 'https://docs.flagsmith.com/integrations/apm/sentry',
       'fields': [
         {
@@ -204,7 +203,7 @@ const defaultFlags = {
       ],
       'image': '/static/images/integrations/sentry.svg',
       'perEnvironment': true,
-      'tags': ['logging'],
+      'tags': ['Monitoring'],
       'title': 'Sentry',
     },
     'slack': {
--- a/frontend/common/stores/default-flags.ts
+++ b/frontend/common/stores/default-flags.ts
@@ -187,6 +187,26 @@ const defaultFlags = {
       'tags': ['analytics'],
       'title': 'Segment',
     },
+    'sentry': {
+      'description':
+        'Sends feature flag change events to Sentry for change tracking.',
+      'docs': 'https://docs.flagsmith.com/integrations/apm/sentry',
+      'fields': [
+        {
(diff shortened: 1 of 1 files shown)


frontend/common/stores/default-flags.ts
```diff
@@ -187,6 +187,26 @@ const defaultFlags = {
...
+        },
+        {
+          'hidden': true,
+          'key': 'secret',
+          'label': 'Secret',
+        },
+      ],
+      'image': '/static/images/integrations/sentry.svg',
+      'perEnvironment': true,
+      'tags': ['logging'],
```
matthewelwell:
```suggestion
      'tags': ['Monitoring'],
```

frontend/common/stores/default-flags.ts
```diff
@@ -187,6 +187,26 @@ const defaultFlags = {
       'tags': ['analytics'],
       'title': 'Segment',
     },
+    'sentry': {
+      'description':
+        'Sends feature flag change events to Sentry for change tracking.',
```
matthewelwell:
```suggestion
        'Send flag change events to Sentry.',
```

matthewelwell (changes requested):
Added a couple of minor suggested changes to match what we have configured for our production environment.

matthewelwell (approved):

PR #6645, opened 2026-02-02
deps: upgrade to python 3.13

matthewelwell:
## Changes

Upgrade to python 3.13

Note that we had previously attempted to upgrade to python 3.12 back in June 2024, which caused issues as described in [this issue](https://github.com/Flagsmith/flagsmith/pull/4121). Having researched it further, I believe that this was caused by the fact that we were still running django 3.2.25 at the time. We are now on django 5 which should support python 3.12+ (relevant SO comment [here](https://stackoverflow.com/a/77482832)).

1. Added python 3.13 to unit testing matrix
3. Docker based E2E tests
4. Once merged, we'll test and monitor in staging


api/tests/unit/util/test_logging.py
```diff
@@ -48,15 +49,15 @@ def _log_traceback() -> None:
...
+            "timestamp": "2023-12-08 06:05:47,320",
             "logger_name": "test_json_formatter__outputs_expected",
             "process_id": expected_pid,
             "thread_name": "MainThread",
         },
         {
             "levelname": "ERROR",
             "message": "this is an error",
-            "timestamp": "2023-12-08 06:05:47,319",
+            "timestamp": f"2023-12-08 06:05:47,{320 if sys.version_info >= (3, 13) else 319}",
```
khvn26: ...why not change the timestamp in the `freeze_time` decorator?

matthewelwell: Sorry, I don't think I understand the suggestion here, but I'm definitely open to a less hacky fix. 

The issue I came across was that the timestamp in the `freeze_time` decorator was already different to the one that's asserted in the test, but it so happened that py311 and 312 seemed to have the same floating point error that meant that `320 == 319`. py313 seemingly does not suffer the same fate, so I have to accommodate it here. 

I could change it to some other arbitrary number and maybe we wouldn't see the floating point error, but I would expect that to have been done in the original test if anything. 

I'll try it nonetheless, but just want to make sure that I'm not missing something in your suggestion.

khvn26: I think that if we use a different timestamp, e.g. one with 0 microseconds, we won't have to use this conditional logic at all? Is there something I'm missing?

matthewelwell: Yep, we came to the same conclusion - see https://github.com/Flagsmith/flagsmith/pull/6645/commits/01b83a831045885135d6fc4aa4543a6f3e4a7dfe

khvn26 (approved):

.github/workflows/api-pull-request.yml
```diff
@@ -34,9 +34,9 @@ jobs:
     name: API Unit Tests
 
     strategy:
-      max-parallel: 2
+      max-parallel: 3
       matrix:
-        python-version: ["3.11", "3.12"]
+        python-version: ["3.11", "3.12", "3.13"]
```
emyller, line 39: Would it make sense to use the opportunity and drop 3.11?

emyller (approved):
🚀! One minor question to maybe 🚀🚀

.github/workflows/api-pull-request.yml
matthewelwell: The reason that I'm reluctant to do this is because of [this repository](https://github.com/Flagsmith/flagsmith-ee). Until we change [this line](https://github.com/Flagsmith/flagsmith-ee/blob/main/docker/Dockerfile.oracle#L1), I'd like to keep testing against python3.11 in this repository.

emyller: Thanks for the light. https://github.com/Flagsmith/flagsmith-ee/issues/197

matthewelwell:
Heads up that I have tested SMTP email delivery manually using https://mailtrap.io/.

Issue #6050, opened 2025-09-11
Flagsmith Environment Webhook incorrect payload values


tomerplus500:
### How are you running Flagsmith

- [x] Self Hosted with Docker
- [ ] Self Hosted with Kubernetes
- [ ] SaaS at flagsmith.com
- [ ] Some other way (add details in description below)

### Describe the bug

Self hosted - Version 2.196.0

Feature segment override value provided by the webhook payload is incorrect (Enabled works as expected).
For example, I changed the segment override value from 48 to 49, clicked "Update Segment Overrides" and got back incorrect payload where prev values is 48 and new value is 48.

### Steps To Reproduce

1. Set an environment Webhook.
2. Create a feature with one segment override
3. Set segment override value to 0.
4. Change the value from 0 to 1 and observe the JSON sent to the webhook.
5. Clone the segment override view and open it again.
6. Change the value from 1 to 2 and observe the JSON sent to the webhook.

### Expected behavior

Expected to get the correct previous segment override value and the new segment override value.

## Before update
<img width="610" height="308" alt="Image" src="https://github.com/user-attachments/assets/e9d18e3c-8921-4445-bb14-325bf20772d1" />

## After update
<img width="611" height="355" alt="Image" src="https://github.com/user-attachments/assets/9c7190cf-13cd-4523-816a-5b425ac713ae" />

## Webhook provided JSON
```
{
  "event_type": "FLAG_UPDATED",
  "data": {
    "changed_by": "admin@comp.com",
    "timestamp": "2025-09-11T12:37:39.823121Z",
    "previous_state": {
      "enabled": true,
      "environment": {
        "id": 1,
        "name": "Test"
      },
      "feature": {
        "id": 1,
        "name": "test",
        "description": null,
        "type": "STANDARD",
        "created_date": "2025-09-10T13:42:43.260047Z",
        "default_enabled": true,
        "initial_value": "5",
        "project": {
          "id": 1,
          "name": "Default Project"
        }
      },
      "feature_segment": {
        "priority": 0,
        "segment": {
          "id": 4,
          "name": "segment_1",
          "description": ""
        }
      },
      "feature_state_value": 45,
      "identity": null,
      "identity_identifier": null
    },
    "new_state": {
      "enabled": true,
      "environment": {
        "id": 1,
        "name": "Test"
      },
      "feature": {
        "id": 1,
        "name": "test",
        "description": null,
        "type": "STANDARD",
[...]

Issue #5790, opened 2025-07-21
Centralize feature_state webhooks under audit logs


Zaimwa9:
## Context
Related #2064 #2063 

Refactor feature change audit log into a proper audit log based integration ?

Currently, webhook dispatching for feature state changes happens through different paths depending on whether the environment uses feature versioning.

**Non versioned environments**
- `post_save` signal on FeatureState
- Dispatches `trigger_feature_state_change_webhook_signals` skipping manually versioned environments
  - Builds event using a historical record
- Audit log created separately via `create_audit_log_from_historical_record` with `related_object_type=FEATURE_STATE`

**Versioned environments**
- `EnvironmentFeatureVersion.publish()` sends `environment_feature_version_published`
- Dispatches both:
  - `create_environment_feature_version_published_audit_log_task`
  - `trigger_update_version_webhooks`
  - webhook event has a different model
- Audit log created with `related_object_type=EF_VERSION`

<img width="749" height="497" alt="Image" src="https://github.com/user-attachments/assets/3b627ec6-51a8-4341-8c2f-1280211457b2" />

Additionally, organisation level webhooks for audit logs are called separately and directly from a `post_save` signal
```
@receiver(post_save, sender=AuditLog)
def call_webhooks(sender, instance, **kwargs):
```
To be discussed how relevant it would be to tighten up the 2 types of webhooks.

## Goal
Refactor the flow so that environment-level webhook events for feature state changes are dispatched from the creation of an AuditLog, similar to how organization-level integrations work today.
In other words, have the AuditLog being the entrypoint of further business logic in relation to a feature_state change.

## Proposed Flow
Introducing a centralized `post_save` signal on `AuditLog` tracking only `RelatedObjectType.FEATURE_STATE/EF_VERSION` that would in turn emits the signal `feature_state_change_went_live`, extending what has been done [here](https://github.com/Flagsmith/flagsmith/commit/4d415a10cf3e1b171dca9f9b982dd1c3c09f3e01)

Webhook logic would be centralized for both versioned and non-versioned environment, uniformizing event payload via dedicated mappers:
- `EFV Mapper` → builds payload from previously live version
- `FS Mapper` → builds payload using historical diffs

<img width="1373" height="588" alt="Image" src="https://github.com/user-attachments/assets/c6da4c45-e23e-4bca-bc8f-13fcf5ea3ef9" />

## Benefits
- Centralizes webhook dispatching logic for feature state changes under the AuditLog system
[...]

PR #6646, opened 2026-02-02
fix: Flagsmith Environment Webhook incorrect payload values

khvn26:

## Changes

Contributes to #6050, #5790.

In this PR, we fix incorrect webhook payloads when updating segment overrides (#6050). The root cause was that `drf-writable-nested` saves `FeatureState` before `FeatureStateValue`, so webhooks triggered on `post_save` captured stale values.

This is a step toward #5790 (centralising webhook logic), but further work remains. The webhook infrastructure still has multiple layers (`webhooks/webhooks.py`, `features/tasks.py`, `audit/signals.py`) that could be unified into a single, consistent system.    
                                                                                                           
- Remove the `post_save` signal handler from `FeatureState` that triggered webhooks prematurely
- Dispatch `FLAG_UPDATED` webhooks via the `feature_state_change_went_live` signal, which fires after AuditLog creation when all nested objects are saved
- Handle both `FeatureState` and `FeatureStateValue` audit logs, since value-only changes also need to trigger webhooks
- Refactor signal semantics: `FeatureState` is now the sender (matching the signal name), with `AuditLog` as a supplemental kwarg

Added an integration test reproducing #6050, and various unit tests supporting the fix.

### DIFF
--- a/api/audit/signals.py
+++ b/api/audit/signals.py
@@ -226,14 +226,16 @@ def send_feature_flag_went_live_signal(sender, instance, **kwargs):  # type: ign
     if feature_state.is_scheduled:
         return  # This is handled by audit.tasks.create_feature_state_went_live_audit_log
 
-    feature_state_change_went_live.send(instance, feature_state=feature_state)
+    feature_state_change_went_live.send(feature_state, audit_log=instance)
 
 
 @receiver(feature_state_change_went_live)
-def send_audit_log_event_to_sentry(sender: AuditLog, **kwargs: Any) -> None:
+def send_audit_log_event_to_sentry(
+    sender: FeatureState, audit_log: AuditLog, **kwargs: Any
+) -> None:
     try:
         sentry_configuration = SentryChangeTrackingConfiguration.objects.get(
-            environment=sender.environment,
+            environment=audit_log.environment,
             deleted_at__isnull=True,
         )
     except SentryChangeTrackingConfiguration.DoesNotExist:
@@ -244,13 +246,12 @@ def send_audit_log_event_to_sentry(sender: AuditLog, **kwargs: Any) -> None:
         secret=sentry_configuration.secret,
     )
 
-    _track_event_async(sender, sentry_change_tracking)  # type: ignore[no-untyped-call]
+    _track_event_async(audit_log, sentry_change_tracking)  # type: ignore[no-untyped-call]
 
 
 @receiver(feature_state_change_went_live)
 def trigger_feature_state_webhooks(
-    sender: AuditLog,
-    feature_state: FeatureState,
+    sender: FeatureState,
     **kwargs: Any,
 ) -> None:
     """
@@ -267,9 +268,7 @@ def trigger_feature_state_webhooks(
     # 1. drf-writable-nested saves FeatureState before FeatureStateValue
     # 2. The history record's instance may have stale cached values
     try:
-        fresh_feature_state = FeatureState.objects.all_with_deleted().get(
-            id=feature_state.id
-        )
+        fresh_feature_state = FeatureState.objects.all_with_deleted().get(id=sender.id)
     except FeatureState.DoesNotExist:
         return
 
--- a/api/audit/signals.py
+++ b/api/audit/signals.py
@@ -9,7 +9,7 @@ from audit.models import AuditLog
 from audit.related_object_type import RelatedObjectType
 from audit.serializers import AuditLogListSerializer
 from audit.services import get_audited_instance_from_audit_log_record
-from features.models import FeatureState
+from features.models import FeatureState, FeatureStateValue
 from features.signals import feature_state_change_went_live
 from integrations.common.models import IntegrationsModel
 from integrations.datadog.datadog import DataDogWrapper
@@ -212,14 +212,21 @@ def send_audit_log_event_to_slack(sender, instance, **kwargs):  # type: ignore[n
 @receiver(post_save, sender=AuditLog)
 @track_only([RelatedObjectType.FEATURE_STATE])
 def send_feature_flag_went_live_signal(sender, instance, **kwargs):  # type: ignore[no-untyped-def]
-    feature_state = get_audited_instance_from_audit_log_record(instance)
-    if not isinstance(feature_state, FeatureState):
+    audited_instance = get_audited_instance_from_audit_log_record(instance)
+
+    # Handle both FeatureState and FeatureStateValue audit logs
+    # FeatureStateValue changes also have related_object_type=FEATURE_STATE
+    if isinstance(audited_instance, FeatureStateValue):
+        feature_state = audited_instance.feature_state
+    elif isinstance(audited_instance, FeatureState):
+        feature_state = audited_instance
+    else:
         return
 
     if feature_state.is_scheduled:
         return  # This is handled by audit.tasks.create_feature_state_went_live_audit_log
 
-    feature_state_change_went_live.send(instance)
+    feature_state_change_went_live.send(instance, feature_state=feature_state)
 
 
 @receiver(feature_state_change_went_live)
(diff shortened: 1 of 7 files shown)


api/audit/signals.py
```diff
@@ -212,21 +212,30 @@ def send_audit_log_event_to_slack(sender, instance, **kwargs):  # type: ignore[n
...
 
 @receiver(feature_state_change_went_live)
-def send_audit_log_event_to_sentry(sender: AuditLog, **kwargs: Any) -> None:
+def send_audit_log_event_to_sentry(
+    sender: FeatureState, audit_log: AuditLog, **kwargs: Any
+) -> None:
     try:
         sentry_configuration = SentryChangeTrackingConfiguration.objects.get(
-            environment=sender.environment,
+            environment=audit_log.environment,
```
emyller, line 238: Note: Perhaps not the scope of this PR, but the _nulliness_ of `AuditLog.environment` makes this read a bit unsafe. I think we'd be happier with a new `AuditLog.get_environment` method that raises.

api/audit/signals.py
```diff
@@ -237,4 +246,38 @@ def send_audit_log_event_to_sentry(sender: AuditLog, **kwargs: Any) -> None:
...
+    except FeatureState.DoesNotExist:
+        return
+
+    # Skip versioned environments - handled by trigger_update_version_webhooks
+    if fresh_feature_state.environment_feature_version_id:
+        return
+
+    # Skip deleted feature states - handled in views
+    if fresh_feature_state.deleted_at:
+        return
```
emyller: Is this contradicting `all_with_deleted()` above, or is there value in handling this here instead? Looks like a leftover of `api/features/signals.py` logic.

api/audit/signals.py
```diff
@@ -237,4 +246,38 @@ def send_audit_log_event_to_sentry(sender: AuditLog, **kwargs: Any) -> None:
...
+    from features.tasks import trigger_feature_state_change_webhooks
+
+    # Fetch fresh data from the database to ensure we have the latest state
+    # This is necessary because:
+    # 1. drf-writable-nested saves FeatureState before FeatureStateValue
+    # 2. The history record's instance may have stale cached values
+    try:
+        fresh_feature_state = FeatureState.objects.all_with_deleted().get(id=sender.id)
+    except FeatureState.DoesNotExist:
+        return
```
emyller, line 274: Would this find more value in `trigger_feature_state_change_webhooks`'s runtime later?

api/audit/signals.py
```diff
@@ -237,4 +246,38 @@ def send_audit_log_event_to_sentry(sender: AuditLog, **kwargs: Any) -> None:
         secret=sentry_configuration.secret,
     )
 
-    _track_event_async(sender, sentry_change_tracking)  # type: ignore[no-untyped-call]
+    _track_event_async(audit_log, sentry_change_tracking)  # type: ignore[no-untyped-call]
+
+
+@receiver(feature_state_change_went_live)
+def trigger_feature_state_webhooks(
```
emyller: I think naming needs to improve. Nothing wrong with `trigger_feature_state_webhooks`, but with how subtly it intends not to clash with the imported `trigger_feature_state_change_webhooks`.

Suggestion:

```python
import features.tasks

@receiver(feature_state_change_went_live)
def trigger_feature_state_change_webhooks(sender, **kwargs):
    ...
    features.tasks.trigger_feature_state_change_webhooks(sender)
```

api/tests/integration/features/featurestate/test_webhooks.py
```diff
@@ -0,0 +1,89 @@
...
+    1. The webhook signal fires on FeatureState.post_save
+    2. drf-writable-nested saves the parent (FeatureState) before nested (FeatureStateValue)
+    3. So the webhook captures stale values
+    """
+    # Given
+    old_value = 0
+    new_value = 1
+
+    # First create a feature_segment
+    feature_segment_url = reverse("api-v1:features:feature-segment-list")
```
emyller, line 30: Shall we pursue standardising for hardcoded URLs in tests?

api/tests/unit/audit/test_unit_audit_signals.py
emyller, line 1: Note: Most new tests in this file could be grouped with `@pytest.mark.parametrize`.

api/tests/unit/features/test_unit_features_models.py
```diff
@@ -502,20 +502,29 @@ def test_feature_state_str__returns_expected(
...
+    # Make an actual change so the webhook is triggered
+    feature_state.enabled = not feature_state.enabled
 
-    # When
-    feature_state.save()
+    # When - Use request context to set the history user
+    with mock.patch(
+        "simple_history.models.HistoricalRecords.thread",
+        mock.MagicMock(request=mock.MagicMock(user=admin_user)),
+    ):
```
emyller: Consider moving this to a fixture that uses `admin_user` and `mocker` to return a context manager. Suggestion:

```python
@pytest.fixture
def admin_history(admin_user: FFAdminUser, mocker: MockerFixture) -> Callable[[], ContextManager[None]]:
    @contextmanager
    def admin_history_context() -> Generator[None, None, None]:
        sh_thread = mocker.patch.object(HistoricalRecords, "thread")
        sh_thread.request.user = admin_user
        yield
    return admin_history_context
```

api/tests/unit/features/test_unit_features_views.py
```diff
@@ -72,8 +72,14 @@
...
 
+
 # patch this function as it's triggering extra threads and causing errors
-mock.patch("features.signals.trigger_feature_state_change_webhooks").start()
+# Use a module-scoped autouse fixture to properly clean up after the module
+@pytest.fixture(autouse=True, scope="module")
+def _mock_trigger_webhooks() -> typing.Generator[None, None, None]:
+    with mock.patch("features.tasks.trigger_feature_state_change_webhooks"):
+        yield
+
```
emyller: Note: this looks like a hack. I wonder if we can make it any more useful if mocking `tasks` entirely...

```suggestion
@pytest.fixture(autouse=True)
def tasks(mocker: MockerFixture):
    return mocker.patch("features.tasks")
```

...and use it in existing endpoints coverage:

```py
assert tasks.trigger_feature_state_change_webhooks.call_count == 1
```

Also, the `mocker` fixture helps handling cleanup. 😉

emyller (changes requested):
Looks solid! I have comments about the implementation, and non-blocking suggestions in tests.

api/audit/signals.py
khvn26: `AuditLog.get_environment` does sound nice. There's little risk in the status quo because we're using `objects.get`, however.

api/tests/unit/features/test_unit_features_views.py
khvn26: Managed to do away with it entirely in 1986736918d7c3599faebbe2b5fe97842a5464b4.

api/tests/unit/features/test_unit_features_models.py
khvn26: Good call. Added in 6fab906bbefa26ec7645392489e40bfe562de746.

api/tests/unit/audit/test_unit_audit_signals.py
khvn26: I actually prefer distinct tests here as parametrising would make setup less straightforward.

api/tests/integration/features/featurestate/test_webhooks.py
khvn26: It made sense for SDK API tests, but I'm fine with relying on reverse here.

api/audit/signals.py
khvn26: Done in 635774755abc960ee8fe17f73764aa69b5822aee.

api/audit/signals.py
khvn26: Maybe, but we need this here.

api/audit/signals.py
khvn26: Simplified in 2fac64a86b0b646aeda713b7e41a1487a348674c.

api/audit/signals.py
emyller: I think the risk _decrease_ here comes from 1. `SentryChangeTrackingConfiguration.environment` being non-null, and 2. this instance of `AuditLog` **very likely** earning an environment previously. I just don't like the reliability of the code here lies on code somewhere else, but I appreciate this may not be a problem to solve now.

khvn26: I would argue the risk decrease is exactly on `.get` here because the only actual risk I can see is fetching the entire `SentryChangeTrackingConfiguration` table, explicitly prevented by `.get` (hardcoded `LIMIT 21` + `MultipleObjectsReturned` exception when >1 rows returned).

That said, I do agree that raising on null environment would result in a better stacktrace.

api/tests/unit/features/test_unit_features_models.py
emyller: Looks better, but sorry, I need to raise a couple things:

1. You'll need to [not use a context manager with mocker](https://pytest-mock.readthedocs.io/en/latest/usage.html#usage-as-context-manager), because it manages cleanup for you another way. My explicit mention to `mocker` had this in mind, and the snippet I wrote above is tested. 😉 
2. Your new fixture patches history on the test level, while the previous code, and also the snippet above, opt for a context manager to isolate patching to _creating a feature state_. This may or may not be relevant.

khvn26: I don't think 2. is that important. Simplified the fixture in a1d423f0690d54885a45170cede7a67d3580f875.

api/audit/signals.py
emyller: Yup, I see why now, thanks. 👍

emyller (approved):
LGTM

PR #6650, opened 2026-02-03
feat(Local env): Enhance local Django shell devex

emyller:
`make django-shell` now uses [shell_plus](https://django-extensions.readthedocs.io/en/latest/shell_plus.html).

IPython is chosen over other options because it's the most actively maintained of alternatives.


khvn26 (approved):
Yes!

PR #6656, opened 2026-02-04
deps: update `express` + `body-parser` to bump transitive `qs` dependency and fix CVE-2025-15284

matthewelwell:
## Changes

Override transitive `qs` dependency in order to bump to 6.14.1 and resolve CVE-2025-15284. 

Used the frontend preview URL to regression test + E2E tests should also cover this.


frontend/package.json
```diff
@@ -64,7 +64,7 @@
     "babel-plugin-module-resolver": "^4.1.0",
     "babel-plugin-transform-object-rest-spread": "7.0.0-beta.3",
     "babel-plugin-transform-react-remove-prop-types": "0.4.18",
-    "body-parser": "1.20.3",
+    "body-parser": "^2.2.2",
```
matthewelwell, line 67: The only breaking change in v1->v2 is dropping node support <18 which we already don't support ourselves.

Zaimwa9 (approved):
👍

PR #6663, opened 2026-02-05
fix: Enforce seat limit self hosted

matthewelwell:
## Changes

Enforces seat limits for self-hosted deployments. 

Required changes: 

 - [x] API should prevent users joining organisations when organisation is already at or over seat limit
 - [x] FE should show information on the number of seats available / being used in the users settings page
 - [x] FE should gracefully handle an invite failing due to there not being enough seats available

Contributes to flagsmith/flagsmith-private#105 and https://github.com/Flagsmith/infrastructure/issues/267

Ran the FE locally to verify the FE changes. Added / updated unit tests to verify API changes.

### DIFF
--- a/frontend/web/components/pages/InvitePage.js
+++ b/frontend/web/components/pages/InvitePage.js
@@ -1,7 +1,7 @@
 import React, { Component } from 'react'
 import Constants from 'common/constants'
 import { withRouter } from 'react-router-dom'
-import AccountProvider from 'common/providers/AccountProvider';
+import AccountProvider from 'common/providers/AccountProvider'
 const InvitePage = class extends Component {
   static displayName = 'InvitePage'
 
--- a/frontend/web/components/pages/InvitePage.js
+++ b/frontend/web/components/pages/InvitePage.js
@@ -20,14 +20,15 @@ const InvitePage = class extends Component {
     this.props.history.replace(Utils.getOrganisationHomePage(id))
   }
 
-  getErrorMessage(detail) {
-    switch (detail) {
+  getErrorMessage(error) {
+    switch (error) {
+      case 'No Invite matches the given query.':
       case 'Not found.':
         return 'We could not validate your invite, please check the invite URL and email address you have entered is correct.'
       case 'Please upgrade your plan to add additional seats/users':
         return 'The organisation you have been invited to has no seats available. Please contact the organisation administrator to resolve this before trying again.'
       default:
-        return detail
+        return error
     }
   }
 
@@ -41,7 +42,7 @@ const InvitePage = class extends Component {
                 {error ? (
                   <div>
                     <h3 className='pt-5'>Oops</h3>
-                    <p>{this.getErrorMessage(error.detail)}</p>
+                    <p>{this.getErrorMessage(error)}</p>
                   </div>
                 ) : (
                   <Loader />
--- a/frontend/web/components/pages/InvitePage.js
+++ b/frontend/web/components/pages/InvitePage.js
@@ -1,6 +1,7 @@
 import React, { Component } from 'react'
 import Constants from 'common/constants'
 import { withRouter } from 'react-router-dom'
+import AccountProvider from 'common/providers/AccountProvider';
 const InvitePage = class extends Component {
   static displayName = 'InvitePage'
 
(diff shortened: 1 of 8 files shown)


frontend/web/components/pages/InvitePage.js
```diff
@@ -19,6 +20,17 @@ const InvitePage = class extends Component {
     this.props.history.replace(Utils.getOrganisationHomePage(id))
   }
 
+  getErrorMessage(detail) {
+    switch (detail) {
+      case 'Not found.':
```
Zaimwa9, line 26: It seems it drifted at some point. The error message I'm getting here is
```
{
    "detail": "No Invite matches the given query."
}
```
```suggestion
      case 'No Invite matches the given query.':
      case 'Not found.':
```

In doubt I would keep the two

frontend/web/components/pages/InvitePage.js
```diff
@@ -29,11 +41,7 @@ const InvitePage = class extends Component {
                 {error ? (
                   <div>
                     <h3 className='pt-5'>Oops</h3>
-                    <p>
-                      {error.detail === 'Not found.'
-                        ? 'We could not validate your invite, please check the invite URL and email address you have entered is correct.'
-                        : error.detail}
-                    </p>
+                    <p>{this.getErrorMessage(error.detail)}</p>
```
Zaimwa9:
```suggestion
                    <p>{this.getErrorMessage(error)}</p>
```

Same, the store is directly storing the error detail as a string in error

frontend/web/components/pages/InvitePage.js
Zaimwa9: Ideally we would rely on the statuses but it would imply too much refactor versus the scope of this issue

Zaimwa9 (commented):
I'd like to confirm that the difference i'm seeing on the error message display is not due to our local setups somehow

frontend/web/components/pages/InvitePage.js
Zaimwa9: Yes, it's happening in the ajaxHandler:
```
      .then((errorText: string) => {
        let err: any = errorText
        console.log('error-ajax', errorText)
        try {
          err = JSON.parse(errorText)
        } catch {}
        if (res?.status === 500) {
          store.error =
            'An unknown error occurred while logging in. Please try again later or contact an administrator.'
        } else if (err.detail) {
          store.error = err.detail // Here
        } else if (err.non_field_errors) {
          store.error = err.non_field_errors[0]
        } else {
          store.error = err
        }
        store.goneABitWest()
      })
```

frontend/web/components/pages/InvitePage.js
```diff
@@ -19,6 +20,17 @@ const InvitePage = class extends Component {
     this.props.history.replace(Utils.getOrganisationHomePage(id))
   }
 
+  getErrorMessage(detail) {
+    switch (detail) {
+      case 'Not found.':
+        return 'We could not validate your invite, please check the invite URL and email address you have entered is correct.'
+      case 'Please upgrade your plan to add additional seats/users':
```
Zaimwa9, line 28: Also I'm realising this will not work with invitation from link.

I am not sure whether it's on purpose as there is a disabled flag `verify_seats_limit_for_invite_links` used in `account-store.controller.acceptInvite`.
If we try to join an organisation via link while max seats have been reached the error from the link request gets overridden by the one from `users/join/id` that will return the `No Invite matches the given query`:

```
  acceptInvite: (id) => {
    store.saving()
    API.setInvite('')
    API.setInviteType('')
    return data
      .post(`${Project.api}users/join/link/${id}/`)
      .catch((error) => {
        console.log('error-ajax', error)
        if (
          Utils.getFlagsmithHasFeature('verify_seats_limit_for_invite_links') &&
          error.status === 400
        ) {
          API.ajaxHandler(store, error)
          return // Here it should throw too otherwise below res.id is undefined and crash the app
        }
        return data.post(`${Project.api}users/join/${id}/`)
      })
```

Then I don't know if this is the way self-hosted are supposed to invite their members

matthewelwell: As per discussion, and manual testing, I think this is now resolved in e87994c.

frontend/web/components/pages/InvitePage.js
matthewelwell: Fixed in https://github.com/Flagsmith/flagsmith/commit/e87994c468177846eb52e31696c28402c9f3d56c

frontend/web/components/pages/InvitePage.js
matthewelwell: Fixed in https://github.com/Flagsmith/flagsmith/commit/e87994c468177846eb52e31696c28402c9f3d56c

Zaimwa9 (dismissed):
New frontend developer unlocked 🚀

Zaimwa9 (approved):

PR #6670, opened 2026-02-06
deps: update protobuf and setuptools

matthewelwell:
## Changes

Update `protobuf` (transient dependency of google-api-python-client), `wheel` and `jaraco.context` (transient dependencies of `setuptools`). 

Resolves CVEs:

 - CVE-2026-0994
 - CVE-2026-24049
 - CVE-2026-23949

E2E / unit tests


khvn26:
Why do we need setuptools in runtime again?

matthewelwell:
> Why do we need setuptools in runtime again?

This is a good question - it looks like it is a transient dependency of `django-axes` (see [here](https://github.com/Flagsmith/flagsmith/blob/main/api/poetry.lock#L1349)). I would definitely like to try and remove that, but I think that's a separate concern.

khvn26 (approved):

matthewelwell:
@khvn26 see here: https://github.com/Flagsmith/flagsmith/pull/6671

PR #6722, opened 2026-02-16
fix(Admin API): Sort features by overrides

### LINKED ISSUE #6188: Sort features by whether they have a relevant segment or identity override

The upcoming associated segment overrides shows features in context to a segment, highlighting any override.

Image

This makes it the same as how we display identity overrides

Image

The problem with this is that sometimes we often mainly care about seeing the features that are being overridden. 

I think the best solution is to sort by most relevant which is the highlighted rows, followed by alphabetical.

- For an identity page: any features with a matching identity feature state
- For a segment override page: any feature with a segment feature state

We could filter out all other features in these pages altogether but I don't think thats ideal, having a full list of features is useful in the same table and a filter would be convoluted.

### Additional context

_No response_

emyller:
Closes https://github.com/Flagsmith/flagsmith/issues/6188.

Prioritise displaying overriden features in the dashboard.

For **segment overrides**:

image

For **identity overrides**:

image

Review effort: 2/5


api/features/views.py
```diff
@@ -224,7 +224,28 @@ def get_queryset(self):  # type: ignore[no-untyped-def]
...
         )
-        queryset = queryset.order_by(sort)
+        override_ordering: list[str] = []
+        if segment_id := query_data.get("segment"):
+            queryset = queryset.annotate(
+                has_segment_override=Exists(
+                    FeatureSegment.objects.filter(
+                        feature=OuterRef("pk"),
+                        segment_id=segment_id,
+                    )
```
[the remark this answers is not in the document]
emyller: Fixed in 36917da24b1bb17334f7c743239f626b023f4d2b

api/features/views.py
```diff
@@ -224,7 +224,28 @@ def get_queryset(self):  # type: ignore[no-untyped-def]
...
+                ),
+            )
+            override_ordering.append("-has_segment_override")
+        if identity_id := query_data.get("identity"):
+            queryset = queryset.annotate(
+                has_identity_override=Exists(
+                    FeatureState.objects.filter(
+                        feature=OuterRef("pk"),
+                        identity_id=identity_id,
+                    )
```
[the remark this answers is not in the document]
emyller: An identity is already bound to an environment.

api/features/serializers.py
```diff
@@ -95,6 +95,10 @@ class FeatureQuerySerializer(serializers.Serializer):  # type: ignore[type-arg]
         required=False,
         help_text="Integer ID of the segment to retrieve segment overrides for.",
     )
+    identity = serializers.CharField(
+        required=False,
+        help_text="ID of the identity to sort features with identity overrides first.",
+    )
```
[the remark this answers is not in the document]
emyller: Fixed in 520db304b21a6818ae0c65afe00291c252a73aca.

api/features/views.py
```diff
@@ -224,7 +224,29 @@ def get_queryset(self):  # type: ignore[no-untyped-def]
...
+                        feature=OuterRef("pk"),
+                        segment_id=segment_id,
+                        environment_id=environment_id,
+                    )
+                ),
+            )
+            override_ordering.append("-has_segment_override")
+        if identity_id := query_data.get("identity"):
+            queryset = queryset.annotate(
+                has_identity_override=Exists(
```
Zaimwa9: So here, this would make a subquery for each feature row right? I'm comparing to the segment query just above that have a `unique_constraint` on (feature, segment_id, environment_id), there might be a slight performance difference?
I personally think it's negligeable given that feature limit is 400 per project (~max overrides we can expect) but do you have an idea of the perf loss, how we could measaure it and at what point it could become problematic?

Zaimwa9 (commented):
I have a non-blocking question and the e2e are failing so it's gonna stale the approval anyways, i'll run them to see what's going on

```
 ✖ Segment-part-3
  
     1) AssertionError: waitForElementVisible([data-test="user-feature-switch-1-off"]): expected false to be truthy
  
        Browser: Firefox 143.0 / Debian 13.3
  
           48 |
           49 |export const waitForElementVisible = async (selector: string) => {
           50 |  logUsingLastSection(`Waiting element visible ${selector}`)
           51 |  return t
           52 |    .expect(Selector(selector).visible)
         > 53 |    .ok(`waitForElementVisible(${selector})`, { timeout: LONG_TIMEOUT })
           54 |}
           55 |
           56 |export const waitForElementNotClickable = async (selector: string) => {
           57 |  logUsingLastSection(`Waiting element visible ${selector}`)
           58 |  await t
  
           at <anonymous> (/srv/flagsmith/e2e/helpers.cafe.ts:53:6)
           at step (/srv/flagsmith/e2e/helpers.cafe.ts:33:23)
           at Object.next (/srv/flagsmith/e2e/helpers.cafe.ts:14:53)
           at <anonymous> (/srv/flagsmith/e2e/helpers.cafe.ts:8:71)
           at __awaiter (/srv/flagsmith/e2e/helpers.cafe.ts:4:12)
           at waitForElementVisible (/srv/flagsmith/e2e/helpers.cafe.ts:49:61)
           at <anonymous> (/srv/flagsmith/e2e/tests/segment-test.ts:261:30)
           at step (/srv/flagsmith/e2e/tests/segment-test.ts:33:23)
           at Object.next (/srv/flagsmith/e2e/tests/segment-test.ts:14:53)
           at fulfilled (/srv/flagsmith/e2e/tests/segment-test.ts:5:58)
```

Zaimwa9:
Looks like legit as persistent and reproducible:
- Go to identities tab
- Select any identity

```
"GET /api/v1/projects/353/features/?page=1&environment=438&page_size=50&is_archived=false&is_enabled=null&page=1&search=&sort=%5Bobject%20Object%5D&tag_strategy=INTERSECTION&value_search=&sort_direction=ASC&sort_field=name&identity=b27ff89d-4f0e-4b05-8f0e-85eeefdff267&sort_field=name&sort_direction=ASC HTTP/1.1" 400 45
```

api/features/serializers.py
```diff
@@ -95,6 +95,10 @@ class FeatureQuerySerializer(serializers.Serializer):  # type: ignore[type-arg]
         required=False,
         help_text="Integer ID of the segment to retrieve segment overrides for.",
     )
+    identity = serializers.IntegerField(
```
Zaimwa9: Yes ok, that's working when identities are stored in postgres, but when using edge identities, it passes an UUID.

I'll post a full comment with my findings as we'll need to have 2 implementations

frontend/web/components/pages/UserPage.tsx
```diff
@@ -88,9 +88,9 @@ const UserPage: FC = () => {
       true,
       search,
       sort,
-      getServerFilter(filter),
+      { ...getServerFilter(filter), identity: id },
```
Zaimwa9, line 91: Here we will need to use `Utils.getIsEdge()` to do something like `...(!isEdge ? { identity: id } : {}),`, also for the next useEffect

api/features/serializers.py
Zaimwa9: Can you confirm you were having integer identity.id in the screenshot you shared ?

Zaimwa9 (changes requested):
So my finding after tracking the E2E failure is that we need 2 distinct approaches depending on using edge identities or storing them in postgres.
The current implementation is all good for postgres, even though the frontend should not pass the identity if using edge (we have an utils that is already used to call the edge endpoints).

For edge:
It's actually simpler, as the information already exists. It's coming from
`/api/v1/environments/{env_key}/edge-identities/{identity_as_uuid}/edge-featurestates/all/`

and the response is as an array as follows:
```
[
    {
        "feature": {
            "id": 174685,
            "name": "my_go_flag",
            "type": "STANDARD"
        },
        "enabled": true,
        "feature_state_value": null,
        "overridden_by": null,
        "segment": null,
        "multivariate_feature_state_values": []
    },
    {
        "feature": {
            "id": 174687,
            "name": "welcome_page",
            "type": "MULTIVARIATE"
        },
        "enabled": false,
        "feature_state_value": null,
        "overridden_by": null,
        "segment": null,
        "multivariate_feature_state_values": [
            {
                "multivariate_feature_option": {
                    "value": "plop"
                },
                "percentage_allocation": 0
            }
        ]
    }
]
```

We already have `"overridden_by": null,` that can be `IDENTITY | SEGMENT`. So we could even do it from the frontend (i'd be more at peace to add a sort query param and have it done in the backend as it is the case with postgres identities). Don't hesitate to confirm my finding and I let you chose the best path to take

api/features/views.py
emyller: TL;DR: we should be safe.

---

> So here, this would make a subquery for each feature row right?

Yes, this renders to a subquery in SQL, and translates to a [**hashed** subplan](https://www.postgresql.org/docs/current/using-explain.html) — which roughly means that it executes the subquery only once rather than against each row.

> I'm comparing to the segment query just above that have a `unique_constraint` on (feature, segment_id, environment_id), there might be a slight performance difference?

In theory, as we're crossing tuples using foreign keys as anchors, the RDBMS should make the best use of each B-tree index associated to them.

> I personally think it's negligeable given that feature limit is 400 per project (~max overrides we can expect) but do you have an idea of the perf loss, how we could measaure it and at what point it could become problematic?

The added effort of these _conditional_ filters should cause minimal / negligible performance impact overall. I expect neither data frame size nor query limit to influence performance as we scale.

Here's an example output of `explain analyze` that illustrates an hypothetical query plan containing both subqueries.

```text
Sort  (cost=98.11..98.14 rows=12 width=30) (actual time=0.753..0.764 rows=212 loops=1)
    Sort Key: ((hashed SubPlan 2)) DESC, ((hashed SubPlan 4)) DESC, f.name
    Sort Method: quicksort  Memory: 38kB
    Buffers: shared hit=177
    ->  Index Scan using features_feature_project_id_72859830 on features_feature f  (cost=0.42..97.89 rows=12 width=30) (actual time=0.119..0.364 rows=212 loops=1)
          Index Cond: (project_id = <project_id>)
          Buffers: shared hit=169
          SubPlan 2
            ->  Index Scan using features_featuresegment_segment_id_64602469 on features_featuresegment fs  (cost=0.29..3.63 rows=1 width=4) (actual time=0.022..0.026 rows=2 loops=1)
                  Index Cond: (segment_id = <segment_id>)
                  Filter: (environment_id = <environment_id>)
                  Buffers: shared hit=4
          SubPlan 4
            ->  Index Scan using features_featurestate_identity_id_8229f26c on features_featurestate fst  (cost=0.42..8.41 rows=8 width=4) (actual time=0.035..0.035 rows=0 loops=1)
                  Index Cond: (identity_id = <identity_id>)
                  Buffers: shared hit=3
  Planning Time: 1.122 ms
  Execution Time: 0.845 ms
```

api/features/serializers.py
emyller: Confirmed. It was in my local environment.

frontend/web/components/pages/UserPage.tsx
emyller: Turns out the backend will make use of it either way. See ec88a8231d0ebb7b5d891ba88498a20e3fb79b3a.

emyller:
> So my finding after tracking the E2E failure is that we need 2 distinct approaches depending on using edge identities or storing them in postgres.

I really appreciate you digging that up! It has really helped. Brilliant review.

> (...) So we could even do it from the frontend (i'd be more at peace to add a sort query param and have it done in the backend as it is the case with postgres identities).

This has connected some dots in my ~AI agents~ brain, which led me to ec88a8231d0ebb7b5d891ba88498a20e3fb79b3a. I believe this sorting must happen in the backend. Also, it would take the frontend some [unneeded] effort to figure out sorting across paginated data, if I understand things correctly.

Let me know what you think of this new approach.

api/features/views.py
```diff
@@ -236,15 +249,28 @@ def get_queryset(self):  # type: ignore[no-untyped-def]
...
-            queryset = queryset.annotate(
-                has_identity_override=Exists(
-                    FeatureState.objects.filter(
-                        feature=OuterRef("pk"),
-                        identity_id=identity_id,
-                    )
-                ),
-            )
+        if identity := query_data.get("identity"):
+            if project.enable_dynamo_db:
```
Zaimwa9, line 256: Just thinking out loud, we are paying a legacy odd design here. I got surprised to add a dynamodb dependency in this endpoint but the other way would be too costly.

For the case of postgres identities it makes sense. For edge-api, most of the override data is already returned in `all` BUT missing the base features that are indeed paginated and forces us into this pragmatic choice of re-computing the dynamo overrides here.

Not asking for any change here but worth having it in mind

api/features/views.py
```diff
@@ -236,15 +249,28 @@ def get_queryset(self):  # type: ignore[no-untyped-def]
...
-        if identity_id := query_data.get("identity"):
-            queryset = queryset.annotate(
-                has_identity_override=Exists(
-                    FeatureState.objects.filter(
-                        feature=OuterRef("pk"),
-                        identity_id=identity_id,
-                    )
-                ),
-            )
+        if identity := query_data.get("identity"):
```
Zaimwa9, line 255: Can we validate that the identity would be an integer in the postgres path or an uuid in the edge path?

Given it's for sorting, we could return a 400 or just skip the sorting if invalid wdyt ?

Zaimwa9 (changes requested):
Thanks! It ends up being a bit of a brain-twister. I just have one thinking out loud comment you can resolve and a value validation concern

api/features/views.py
emyller: I agree. I even asked AI to rethink the architecture of the pages using this endpoint so we'd make less queries. There is a path, and it looks like the change in this PR goes in that direction. 👍

api/features/views.py
emyller: Yeah good call. This should be now addressed in 7f74e4a11ba580333c3d597e9f3eb3adca90cd89.

Zaimwa9 (dismissed):
Happy to approve, sorry for the back and forth

api/features/serializers.py
```diff
@@ -95,6 +96,22 @@ class FeatureQuerySerializer(serializers.Serializer):  # type: ignore[type-arg]
...
         help_text="Integer ID of the segment to retrieve segment overrides for.",
     )
+    identity = serializers.CharField(
+        required=False,
+        help_text="ID of the identity to sort features with identity overrides first.",
+    )
+
+    def validate_identity(self, value: str) -> str:
+        project = self.context.get("project")
+        if project and project.enable_dynamo_db:
```
gagantrivedi: In which case would the project not be in context?

emyller: It would be a very unexpected case. This should be improved with 933a47726f34401af48f14f32b3d96f769159a72.

api/features/views.py
emyller: Please unresolve if there's any action here.

api/features/views.py
emyller: This looks resolved.

api/edge_api/identities/edge_identity_service.py
```diff
@@ -53,3 +56,14 @@ def get_edge_identity_overrides_for_feature_ids(
...
+
+def get_overridden_feature_ids_for_edge_identity(identity_uuid: str) -> set[int]:
+    try:
+        identity_document = EdgeIdentity.dynamo_wrapper.get_item_from_uuid(
+            identity_uuid
+        )
+    except ObjectDoesNotExist:
+        return set()
+    identity = EdgeIdentity.from_identity_document(identity_document)
+    return {fs.feature.id for fs in identity.feature_overrides}
```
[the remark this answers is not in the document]
emyller: Very clever observation, but it sounds like the requester would only find irrelevant results rather than uncover a security or data issue.

Zaimwa9 (dismissed):

frontend/web/components/pages/UserPage.tsx
emyller: I will assume this as resolved given your approval, but I'm all ears to new feedback here, if any.

api/features/views.py
```diff
@@ -224,7 +240,42 @@ def get_queryset(self):  # type: ignore[no-untyped-def]
...
+            else:
+                queryset = queryset.annotate(
+                    has_identity_override=Exists(
+                        FeatureState.objects.filter(
+                            feature=OuterRef("pk"),
+                            identity_id=identity,
+                        )
+                    ),
+                )
+            override_ordering.append("-has_identity_override")
```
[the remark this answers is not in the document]
emyller: Duplicate of https://github.com/Flagsmith/flagsmith/pull/6722#pullrequestreview-3849145455. Stop.

Zaimwa9 (approved):

gagantrivedi (approved):

PR #6732, opened 2026-02-19
fix: log warning for non-2xx webhook responses

gagantrivedi:

## Changes

Removed `raise_for_status()` from `call_webhook_with_failure_mail_after_retries` and replaced it with a warning log for all non-2xx responses. Previously, `raise_for_status()` only raised `HTTPError` on 4xx/5xx responses, meaning 3xx responses were silently ignored. Now, any non-2xx response (including 3xx) is logged at warning level with the webhook ID, HTTP status code, and attempt count.

Added a unit test (`test_call_webhook_with_failure_mail_after_retries__non_2xx_response__logs_warning`) that verifies a non-2xx response triggers the expected warning log.


Zaimwa9 (approved):
👍

PR #6736, opened 2026-02-19
feat: support project level custom fields

### LINKED ISSUE #6620: Support project-level overriding of org level custom fields
This ticket is the backend part for [#4733](https://github.com/Flagsmith/flagsmith/issues/4733)

Zaimwa9:

## Changes

Closes #6620 

Support project scoped custom fields which could only be defined at organisation levels
- Added project as foreign key to `MetadataField`
- Replaced `unique_together` with conditional `UniqueConstraints`
- `_validate_required_metadata` accepts project param 
- `is_required_for` fetched and included in responses directly (replacing 2 different requests made in the frontend => coming in next branch)
- Added pagination

**New endpoint**
- GET /projects/{id}/metadata/fields/ with 
    - `?include_organisation` to include inherited org-level fields (calculates the precedence for project in case of conflicts)
    - `?entity=feature|segment|environment` to only return fields of the given identity (simplifies the creation of the identity) 

:warning:
In case of conflicting name, project custom fields takes precedence and org ones are skipped

- Tests and following PR


api/metadata/migrations/0002_add_project_to_metadata_field.py
```diff
@@ -0,0 +1,46 @@
...
+            ),
+        ),
+        migrations.AddConstraint(
+            model_name="metadatafield",
+            constraint=models.UniqueConstraint(
+                condition=models.Q(("project__isnull", False)),
+                fields=("name", "organisation", "project"),
+                name="unique_project_level_metadata_field",
+            ),
+        ),
```
Zaimwa9, line 45: I'm not sure this is compatible with Oracle. Is it a requirement ?

emyller: IIRC, we're not supporting Oracle moving forwards in code. I'd defer confirmation to @matthewelwell.

gagantrivedi:
> ⚠️
In case of conflicting name, project custom fields takes precedence and org ones are skipped

Wouldn't it be less confusing to not allow conflicts?

Zaimwa9:
@gagantrivedi 

> Wouldn't it be less confusing to not allow conflicts?

I do remember asking the question while grooming and we agreed on allowing re-creating the custom fields at project level. 
Maybe @asaphko you can give us some product-wise insights here. To summarize, in project settings, we show as read-only the organisation field. You can create the same field for your project and this one will take over. A use-case I can see is to have a global custom field that you can require or not at project level.
image

What Gagan suggests is to simply forbid and return an error if the name already exists in the organisation (be it project/org level)

api/metadata/migrations/0002_add_project_to_metadata_field.py
matthewelwell: We are a while away from dropping oracle support so we'll need to consider it still.

Zaimwa9: So my findings are that Oracle silently skips `UniqueConstraint` and the constraint is not enforced at the database level.
It seems that it's already a pattern that we've used in previous migrations e.g:
- [FeatureState](https://github.com/Flagsmith/flagsmith/blob/main/api/features/migrations/0025_alter_unique_constraints_for_feature_states.py)
- [Release pipeline feature versioning](https://github.com/Flagsmith/flagsmith/blob/main/api/features/versioning/migrations/0006_add_pipeline_stage_to_envfeatureversion.py)
- And older in [LD importer](https://github.com/Flagsmith/flagsmith/blob/main/api/integrations/launch_darkly/migrations/0002_importrequest_unique_project_ld_project_key_status_result_null.py)

So the support is not optimal for Oracle but still compatible, if ever you have some thoughts on this @gagantrivedi ?

gagantrivedi: Yeah, we will have to add this constraint as a hook when merging this with EE.

gagantrivedi (approved):

api/metadata/migrations/0002_add_project_to_metadata_field.py
matthewelwell: We could just do that now, right? To prevent further drift of the schemas and code.

Zaimwa9: We went over this offline with Gagan and the serializer validation is replicating the uniqueness already so we should be able to avoid a hook. (tested in `tests/unit/metadata/test_views.py::test_create_metadata_field__uniqueness`) 👍

gagantrivedi (dismissed):

gagantrivedi (approved):

PR #6737, opened 2026-02-19
feat: create custom fields for projects

### LINKED ISSUE #4733: Add the ability to create custom fields for a specific project
At the moment, custom fields are created at the organisation. It would be good to be able to create custom fields that are tied to a specific project only.

Zaimwa9:

## Changes

Closes #4733 

Implements the frontend to create and manage project-level custom fields (taking precedence over organisation fields)

- Following nesting model_fiels, replaced the 2 requests + merge approach with a single call to `projects/{id}/metadata/fields/`
- Removed `mergeMetadataFields` 😌 
- added `getProjectMetadataFieldList` with pagination
- added project to create project-scoped custom fields
- `ProjectSettings` custom fields page now shows organisation custom fields (read only) and a new project-related custom fields table
- `project_custom_fields` flag to hide the custom fields table for now

image

image

https://www.loom.com/share/5747526237e64f95b45683f251fc04ad


frontend/common/services/useMetadataField.ts
```diff
@@ -83,13 +89,12 @@ export const metadataService = service
           // Build queries to run in parallel
           const queries: Promise<{ data?: unknown; error?: unknown }>[] = [
             baseQuery({
-              url: `metadata/fields/?${Utils.toParam({
-                organisation: arg.organisationId,
+              url: `projects/${arg.projectId}/metadata/fields/?${Utils.toParam({
+                entity: arg.entityType,
+                include_organisation: true,
+                page_size: 100,
```
talissoncosta, line 95: Quick question about this `page_size: 100`

Do you think we should stick with it for now, or bump it to something higher? Just thinking about orgs with lots of custom fields.

No strong opinion either way, just flagging it! 😊

frontend/web/components/metadata/MetadataPage.tsx
```diff
@@ -1,77 +1,85 @@
...
-  const { data: MetadataModelFieldList } = useGetMetadataModelFieldListQuery({
-    organisation_id: organisationId,
-  })
+  const { data: orgMetadataFieldList } = useGetMetadataFieldListQuery(
+    {
+      organisation: parseInt(organisationId),
+      page: orgPage,
+      page_size: PAGE_SIZE,
+    },
+    { skip: false },
```
talissoncosta: Quick question: why do we have `skip: false` here? 🤔

Since `organisationId` is required, would it make more sense to use `skip: !organisationId` instead?

frontend/web/components/metadata/MetadataPage.tsx
```diff
@@ -104,13 +113,144 @@ const MetadataPage: FC<MetadataPageType> = ({ organisationId }) => {
...
+      space
+      className={`list-item${readOnly ? '' : ' clickable cursor-pointer'}`}
+      key={metadata.id}
+      onClick={
+        readOnly
+          ? undefined
+          : () => {
+              editMetadata(`${metadata.id}`, metadata.model_fields)
+            }
+      }
```
talissoncosta: Could we extract this handler to a const? Would make it cleaner! 🤔

```typescript
// Suggestion
const handleEdit = () => editMetadata(`${metadata.id}`, metadata.model_fields)

return (
  <Row
    onClick={readOnly ? undefined : handleEdit}
    // or even simpler: onClick={!readOnly && handleEdit}
  >
)
```
Just a small readability improvement! 😊

frontend/web/components/modals/CreateMetadataField.tsx
```diff
@@ -167,19 +163,21 @@ const CreateMetadataField: FC<CreateMetadataFieldType> = ({
...
-      updateMetadataField({
-        body: {
-          description,
-          name,
-          organisation: organisationId,
-          type: `${typeValue?.value}`,
-        },
-        id: id!,
-      }).then(() => {
+  const save = async () => {
```
talissoncosta, line 164: I don't want to slow you down at all! Just pointing out that the `save` function has grown quite complex with the nested Promise.all() calls and multiple conditional branches.

It might be worth extracting some helper functions to simplify this in a future refactor—but no blocker for this PR!

Great work my friend! 😊

frontend/common/services/useMetadataField.ts
Zaimwa9: Yeah no good shoutout. The 100 was already the high number. I wanted a high - not too high - value and arbitrarly went for 100

frontend/web/components/modals/CreateMetadataField.tsx
Zaimwa9: Noted! I'll leave it for later if that's ok 👍

talissoncosta (dismissed):
LGTM! Thanks for handling that @Zaimwa9

Zaimwa9:
@talissoncosta heads-up here that I will remove the feature flag because:
- It lost its interest following refactoring the requests (it's not gating the requests)
- it's mostly an enterprise feature, and it would add friction to FoF users to have to activate it themselves

talissoncosta:
could you please fix the conflicts @Zaimwa9 ?

talissoncosta:
Hey @Zaimwa9! Great work on it.

I just noticed some type inconsistencies with projectId and organisationId across the metadata components. I created a follow-up PR to clean them up: https://github.com/Flagsmith/flagsmith/pull/6817

frontend/common/types/requests.ts
```diff
@@ -425,14 +425,19 @@ export type Req = {
     }
   }
   getMetadataField: { organisation_id: number }
-  getMetadataList: { organisation: number }
+  getMetadataList: PagedRequest<{ organisation: number }>
```
talissoncosta, line 428: Nit: should organisation be renamed to organisation_id here? It holds a numeric ID, not an organisation object, and the other metadata request types (e.g. useGetSupportedContentTypeQuery) already use organisation_id. That said, this seems to be a wider codebase pattern — bare organisation/project fields in request bodies map to Django FK field names, so fixing this would require a backend change too. Out of scope for this PR, just flagging for awareness.

talissoncosta (approved):

khvn26 (approved):

PR #6653, opened 2026-02-04
feat: upgrade flagsmith-flag-engine to v10

gagantrivedi:
Thanks for submitting a PR! Please check the boxes below:

- [x] I have read the [Contributing Guide](/Flagsmith/flagsmith/blob/main/CONTRIBUTING.md).
- [ ] I have added information to `docs/` if required so people know about the feature.
- [x] I have filled in the "Changes" section below.
- [x] I have filled in the "How did you test this code" section below.

## Changes

Contributes to #5991

Upgrade `flagsmith-flag-engine` from `fix/missing-export` branch to `^10.0.3` and migrate to vendored Pydantic models.

**Background:** Flag-engine v7+ removed all Pydantic models in favour of TypedDicts. Since the Flagsmith API relies on these Pydantic models for DynamoDB document handling and identity management, we vendor the models temporarily until the codebase can be fully migrated to the TypedDict-based API.

**Key changes:**
- Bump `flagsmith-flag-engine` to `^10.0.3`
- Bump `flagsmith` SDK to `^5.1.1`
- Add `pydantic-collections` dependency
- Vendor Pydantic models in `util/engine_models/` from the `fix/missing-export` branch
- Add context mappers in `util/engine_models/context/mappers.py` to bridge Pydantic models to v10 TypedDicts
- Update imports across the codebase to use vendored models
- Clean up unused vendored code (properties/methods not called by our code)
- Add new tests for full coverage

## How did you test this code?

- All 2969 tests pass
- 100% coverage on vendored code
- Lint and typecheck pass

PR #6741, opened 2026-02-20
fix: Revert flag-engine v10 revert to re-apply engine upgrade

gagantrivedi:

## Changes

Reverts #6674

This PR reverts the flag-engine v10 revert, effectively re-applying the original upgrade from #6653, including:
- Upgrading `flagsmith-flag-engine` from the `fix/missing-export` branch back to `^10.0.3`
- Upgrading `flagsmith` SDK from `^3.10.0` back to `^5.1.1`
- Restoring `pydantic-collections` dependency
- Restoring all vendored Pydantic models in `util/engine_models/`
- Restoring updated imports across the codebase

- CI tests should pass


Zaimwa9:
The branch is almost identical to the reverted one. Can I re-ask the reason why we reverted, was it out of safety during the hackaton or because we identified a regression?

gagantrivedi:
> The branch is almost identical to the reverted one. Can I re-ask the reason why we reverted, was it out of safety during the hackaton or because we identified a regression?

Yeah, it was because the e2e tests that you wrote for Edge, when running against staging, were failing. We later realised that the failure was caused by Core not having access to identity override data, and not because something was wrong with the evaluation engine/logic.

khvn26 (dismissed):

khvn26 (approved):

Zaimwa9 (approved):

PR #6743, opened 2026-02-20
fix: use-proxy-api-url-version-endpoint

### LINKED ISSUE #6588: Using `FLAGSMITH_API_PROXY_URL` prevents the FE from reading the API `/version` endpoint

- [x] Self Hosted with Docker
- [ ] SaaS at flagsmith.com

### Describe the bug

When using the FE to proxy traffic to the API via the `FLAGSMITH_API_PROXY_URL` environment variable, the FE is unable to access the `/version` endpoint on the API because it only proxies traffic where the URI starts with `/api/v1`. 

Note that this prevents the FE from determining if the running API is the enterprise version or not since the FE just responds with it's own `/version` endpoint and doesn't proxy to the API. 

### Steps To Reproduce

The following is a minimal docker-compose to reproduce the issue. Note that this is also the default behaviour in k8s. 

```yaml
volumes:
  pgdata:

services:
  postgres:
    image: postgres:15.5-alpine
    environment:
      POSTGRES_PASSWORD: password
      POSTGRES_DB: flagsmith
    volumes:
      - pgdata:/var/lib/postgresql/data
    healthcheck:
      test: ['CMD-SHELL', 'pg_isready -d flagsmith -U postgres']
      interval: 2s
      timeout: 2s
      retries: 20
      start_period: 20s

  flagsmith-frontend:
    image: flagsmith/flagsmith-frontend:latest
    environment:
      FLAGSMITH_PROXY_API_URL: http://flagsmith-api:8000
    ports:
      - 8080:8080

  flagsmith-api:
    image: docker.flagsmith.com/flagsmith/flagsmith-api:latest
    environment:
      DJANGO_ALLOWED_HOSTS: '*'
      DJANGO_SECRET_KEY: secret
      TASK_RUN_METHOD: TASK_PROCESSOR
    depends_on:
      postgres:
        condition: service_healthy

  flagsmith-task-processor:
    image: docker.flagsmith.com/flagsmith/flagsmith:latest
    environment:
      DJANGO_ALLOWED_HOSTS: '*'
    ports:
      - 8001:8000
    depends_on:
      - flagsmith-api
    command: run-task-processor
```

### Expected behavior

The FE can access the `/version` endpoint of the API. 

### Screenshots

_No response_

Zaimwa9:

## Changes

Closes #6588 

Frontend was only handling `api/v1/*` when proxying traffic with `FLAGSMITH_PROXY_API_URL` while `version` lives at the root, causing the frontend to hit its own `/version`, ignoring `is_enterprise` and `is_saas` in deployments

- Add `/version` in proxied paths when `FLAGSMITH_PROXY_API_URL` is set
- Renamed FE version endpoint to `/_frontend_version` to avoid  conflicts
- Update `useBuildVersion` to use `/_frontend_version`

- Start a flagsmith api via docker
- in `frontend` start the server `FLAGSMITH_PROXY_API_URL=http://localhost:8000 ENV=local npm run dev`
  Run:
  - `curl http://localhost:8080/_backend_version` → response includes `is_enterprise`, `is_saas` (fetched from the API)
  - `curl http://localhost:8080/version` → returns frontend version (`ci_commit_sha`, `image_tag`)

  Go to http://localhost:8080 in the browser and verify:
  - Organisation Settings correctly shows licensing tab (based on `is_enterprise`)
  - The version tag in the UI shows the API version andnot `Unknown`

### DIFF
--- a/frontend/api/index.js
+++ b/frontend/api/index.js
@@ -139,6 +139,14 @@ if (process.env.FLAGSMITH_PROXY_API_URL) {
       xfwd: true,
     }),
   )
+  app.use(
+    '/version/',
+    createProxyMiddleware({
+      changeOrigin: true,
+      target: process.env.FLAGSMITH_PROXY_API_URL,
+      xfwd: true,
+    }),
+  )
 }
 
 if (isDev) {
@@ -179,7 +187,7 @@ app.get('/health', (req, res) => {
   res.send('OK')
 })
 
-app.get('/version', (req, res) => {
+app.get('/_frontend_version', (req, res) => {
   let commitSha = 'Unknown'
   let imageTag = 'Unknown'
 
--- a/frontend/api/index.js
+++ b/frontend/api/index.js
@@ -139,14 +139,6 @@ if (process.env.FLAGSMITH_PROXY_API_URL) {
       xfwd: true,
     }),
   )
-  app.use(
-    '/version/',
-    createProxyMiddleware({
-      changeOrigin: true,
-      target: process.env.FLAGSMITH_PROXY_API_URL,
-      xfwd: true,
-    }),
-  )
 }
 
 if (isDev) {
@@ -187,7 +179,7 @@ app.get('/health', (req, res) => {
   res.send('OK')
 })
 
-app.get('/_frontend_version', (req, res) => {
+app.get('/version', (req, res) => {
   let commitSha = 'Unknown'
   let imageTag = 'Unknown'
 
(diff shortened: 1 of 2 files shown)


frontend/api/index.js
```diff
@@ -179,7 +187,7 @@ app.get('/health', (req, res) => {
   res.send('OK')
 })
 
-app.get('/version', (req, res) => {
+app.get('/_frontend_version', (req, res) => {
```
matthewelwell: Can we still keep the `/version` endpoint for the FE so we're not releasing a breaking change? Something like this:

```suggestion
app.get('/_frontend_version', '/version', (req, res) => {
```

Zaimwa9: Yep so I reverted to using `/version` for the frontend version and instead added a `/_backend_version` endpoint in the express server only when `FLAGSMITH_PROXY_API_URL` is set. Its only role is to fetch the version from the API. `useBuildVersion.ts` deals with calling the right endpoint (it was only internal)
Otherwise the difference was made on the trailing slash `/version` vs `/version/` which was a bit fragile.

So same results and same endpoints as before

matthewelwell (approved):

PR #6752, opened 2026-02-23
chore: Reduce sentry sampling rate

matthewelwell:
## Changes

In order to manage the Sentry subscription cost, this PR achieves 2 things: 

 1. Reduces the default sample rate from 100% (default) to 50%
 2. Ignores any requests to `/api/v1/analytics/telemetry` which are sent by self-hosted deployments and we don't need to monitor the performance of

Will be tested by monitoring the usage of sentry in production.

### DIFF
--- a/api/integrations/sentry/samplers.py
+++ b/api/integrations/sentry/samplers.py
@@ -2,7 +2,7 @@ from contextlib import suppress
 
 from django.conf import settings
 
-NON_FUNCTIONAL_ENDPOINTS = ("/health", "")
+NON_FUNCTIONAL_ENDPOINTS = ("/health", "", "/api/v1/analytics/telemetry/")
 SDK_ENDPOINTS = {
     "/api/v1/flags",
     "/api/v1/identities",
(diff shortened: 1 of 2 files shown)


api/integrations/sentry/samplers.py
```diff
@@ -2,7 +2,7 @@
 
 from django.conf import settings
 
-NON_FUNCTIONAL_ENDPOINTS = ("/health", "")
+NON_FUNCTIONAL_ENDPOINTS = ("/health", "", "/api/v1/analytics/telemetry/")
```
khvn26, line 5: Do we need to add `"/metrics"` here?

matthewelwell: Good question... I would have thought so, yes, but according to [this](https://flagsmith.sentry.io/explore/traces/?aggregateField=%7B%22groupBy%22%3A%22transaction%22%7D&aggregateField=%7B%22yAxes%22%3A%5B%22count%28span.duration%29%22%5D%7D&environment=production&mode=aggregate&project=5544478&query=&statsPeriod=7d) view, it appears that we don't.

khvn26 (approved):

PR #6756, opened 2026-02-23
fix: parse-string-user-agents-to-int-in-mapper

Zaimwa9:

## Changes
- Parse user-agent to int

It assumes that we missed user agent in production at the moment because influx would return strings

- Tests


khvn26 (approved):
LGTM 👍

PR #6762, opened 2026-02-24
fix: enable Gunicorn worker recycling with graceful analytics flush

gagantrivedi:
Contributes to https://github.com/Flagsmith/pulumi/issues/162

---

## Summary

- Enable Gunicorn `--max-requests` (default 1000) and `--max-requests-jitter` (default 100) to recycle workers periodically, mitigating memory leaks
- Add `atexit` handler to flush in-process analytics caches (`APIUsageCache`, `FeatureEvaluationCache`) via the task processor before a worker exits, preventing data loss during recycling
- Rename internal flush methods for clarity: `_flush_through_thread` (hot path) vs `_flush_through_task_processor` (shutdown path)

## Context

Without `--max-requests`, Gunicorn workers never recycle and memory leaks accumulate indefinitely. Enabling worker recycling requires flushing any buffered analytics data before exit, otherwise counts are silently lost. The shutdown flush uses `.delay()` (task queue enqueue) rather than `.run_in_thread()` to avoid thread-safety issues during Python interpreter teardown.

Both `GUNICORN_MAX_REQUESTS` and `GUNICORN_MAX_REQUESTS_JITTER` remain configurable via environment variables. Setting `GUNICORN_MAX_REQUESTS=0` disables recycling entirely.

## Test plan

- [x] Unit tests for `flush_on_shutdown` on both cache classes (populated and empty)
- [x] Unit tests for `flush_analytics_caches` atexit handler (happy path and exception handling)
- [x] Unit test for `AppAnalyticsConfig.ready()` atexit registration
- [x] Manual verification: ran Gunicorn with `--max-requests 3`, confirmed atexit handler fires and logs flush on worker recycle

### DIFF
--- a/api/app_analytics/apps.py
+++ b/api/app_analytics/apps.py
@@ -1,6 +1,25 @@
+import atexit
+import logging
+
 from django.apps import AppConfig
+
+logger = logging.getLogger(__name__)
 
 
 class AppAnalyticsConfig(AppConfig):
     default_auto_field = "django.db.models.BigAutoField"
     name = "app_analytics"
+
+    def ready(self) -> None:
+        atexit.register(flush_analytics_caches)
+
+
+def flush_analytics_caches() -> None:
+    try:
+        from app_analytics.services import api_usage_cache
+        from app_analytics.views import feature_evaluation_cache
+
+        api_usage_cache.flush_on_shutdown()
+        feature_evaluation_cache.flush_on_shutdown()
+    except Exception:
+        logger.exception("Failed to flush analytics caches during shutdown")
--- a/api/app_analytics/cache.py
+++ b/api/app_analytics/cache.py
@@ -24,7 +24,7 @@ class APIUsageCache:
         self._last_flushed_at = timezone.now()
         self._lock = Lock()
 
-    def _flush(self) -> None:
+    def _flush_through_thread(self) -> None:
         for key, value in self._cache.items():
             track_request.run_in_thread(
                 kwargs={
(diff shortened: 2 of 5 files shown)


api/app_analytics/apps.py
```diff
@@ -1,6 +1,25 @@
...
 
+logger = logging.getLogger(__name__)
+
 
 class AppAnalyticsConfig(AppConfig):
     default_auto_field = "django.db.models.BigAutoField"
     name = "app_analytics"
+
+    def ready(self) -> None:
+        atexit.register(flush_analytics_caches)
```
khvn26, line 14: Is gunicorn's `worker_exit` hook more suitable for this?

gagantrivedi: I don't see much difference between them for our use case. Can you elaborate a bit more on your reasoning?

api/app_analytics/cache.py
```diff
@@ -39,6 +39,24 @@ def _flush(self) -> None:
...
+                kwargs={
+                    "resource": key.resource.value,
+                    "host": key.host,
+                    "environment_key": key.environment_key,
+                    "count": value,
+                    "labels": dict(key.labels),
+                }
+            )
+        self._cache = {}
+        self._last_flushed_at = timezone.now()
```
matthewelwell, line 54: Couldn't we defer the iteration to the task processor as by just sending over `self._cache` itself as json for example? I don't know how big this cache could be, but I don't love the idea here of creating an indefinite number of tasks on a regular basis.

gagantrivedi: Yeah, good shout! I will add a task for bulk tracking

api/app_analytics/apps.py
khvn26: IMO it'll better document our intent as we do not need to force-flush outside of worker context, and lower the mental fatigue of mapping out the worker lifecycle as we already manage it [here](https://github.com/Flagsmith/flagsmith-common/blob/main/src/common/gunicorn/conf.py).

On a technical level, `worker_exit` is a stronger guarantee that the code will run when we need it to run (i.e. when a worker is marked for recycling).

gagantrivedi:
After further thought, `--max-requests` would just mask underlying issues rather than address them properly. We should be more proactive about identifying and resolving the root causes. We've tracked the issues we've found so far in https://github.com/Flagsmith/pulumi/issues/162 and will work through those. Closing this in favour of that approach.

PR #6781, opened 2026-02-25
fix: Disabling traits persistence prevents traits to be used in segment evaluation

### LINKED ISSUE #6739: Disabling traits persistence prevents traits to be used in segment evaluation

- [ ] Self Hosted with Docker
- [ ] SaaS at flagsmith.com

### Describe the bug

Hello,
I don't know if it's a bug or a misunderstanding on my part, but I've recently disabled the "Persist traits when using client-side SDK keys" option for which the tooltip says : "If enabled, Flagsmith will persist any non-transient traits sent by SDKs using client-side keys when remotely evaluating flags."

My understanding was that the traits would still be used for evaluating the segments, but would not get persisted to the database.
However, on my system (running the latest version of flagsmith at the time of writing this issue), my traits are not used in the evaluation and I simply get the environment default values.

I think this is related to https://github.com/Flagsmith/flagsmith/blob/main/api/environments/sdk/serializers.py#L193-L197
Where we should return instead of an empty list, all traits as transient.
something like:
```
    def validate_traits(self, traits: typing.List[dict] = None):  # type: ignore[no-untyped-def,type-arg,assignment]
        request = self.context["request"]
        if traits and not request.environment.trait_persistence_allowed(request):
            for trait in traits:
                trait["transient"] = True
        return traits
```

### Steps To Reproduce

1. Go to environment settings / sdk settings
2. turn off Persist traits when using client-side SDK keys
3. call getidentity with traits
4. Confirm that the traits are missing from response and flags are only env default

### Expected behavior

Traits (transient or not) should be used to evaluate returned flags

### Screenshots

_No response_

khvn26:

## Changes

Closes #6739.

Added an integration test and updated an existing unit test for the serializer.

### DIFF
--- a/api/environments/sdk/serializers.py
+++ b/api/environments/sdk/serializers.py
@@ -193,5 +193,6 @@ class IdentifyWithTraitsSerializer(
     def validate_traits(self, traits: typing.List[dict] = None):  # type: ignore[no-untyped-def,type-arg,assignment]
         request = self.context["request"]
         if traits and not request.environment.trait_persistence_allowed(request):
-            return []
+            for trait in traits:
+                trait["transient"] = True
         return traits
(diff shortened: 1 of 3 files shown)


Zaimwa9 (approved):
Ok yes, makes complete sense

emyller (approved):

khvn26:
This is not the fix that we're looking for.

PR #6782, opened 2026-02-25
docs: Clarify `allow_client_traits` toggle name and behaviour

khvn26:

## Changes

Contributes to https://github.com/Flagsmith/flagsmith/issues/6739

The `allow_client_traits` toggle was labelled "Persist traits when using client-side SDK keys", which implied traits would still be used for evaluation but simply not saved. In reality, when this setting is off, traits from client-side SDKs are completely ignored — they are neither persisted nor used for segment evaluation. This is intentional: it's a security boundary that prevents untrusted clients from sending arbitrary traits to gain access to segments they shouldn't match.

This PR clarifies the toggle name and documentation to match the actual behaviour:

- **Frontend**: Renamed toggle to "Allow client-side SDKs to set traits" with a description explaining that traits are fully ignored and why.
- **Security docs**: Updated the toggle name reference and expanded the explanation to clarify that traits are not used for evaluation either, not just not persisted.
- **Segments docs**: Updated the toggle name reference and wording.

Verified the changes visually.

### DIFF
--- a/docs/docs/administration-and-security/governance-and-compliance/security.md
+++ b/docs/docs/administration-and-security/governance-and-compliance/security.md
@@ -4,13 +4,18 @@ sidebar_label: Security
 sidebar_position: 40
 ---
 
-## Preventing Client SDKS from setting Traits
+## Preventing Client SDKs from setting Traits
 
 There may be use-cases where you want to prevent client-side SDKs from setting traits of users. For example, if you are
 setting `plan=silver` as a trait, and then enabling/disabling features based on that plan, a malicious user could, with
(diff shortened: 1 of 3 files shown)


matthewelwell (commented):
This looks good to me, and reflects the behaviour I expected from the attribute. We should evaluate the situation and approach for ensuring the logic in the Edge API matches this before merging this though I think?

emyller (approved):
The new text is great. Reading it almost makes it feel like we may be missing a note explaining why we have `allow_client_traits` enabled by default.

PR #6786, opened 2026-02-26
fix: hide-roles-tab-only-for-non-org-admin-in-env-project-settings

Zaimwa9:

## Changes
The Permissions tab on Project Settings was completely inaccessible for users who are project admins but not organisation admins.
It shows `Error loading permissions data` because `GET /roles` would fail and block the entire access

Cause: We fetch organisation roles (GET `/organisations/{id}/roles/`), which requires org admin access

In environment settings we had 2 problems:
- the same issue was not blocking the whole access but showing `You currently have no roles`
- Wrong matching between roles and environment permissions would show the same error for org admins

https://www.loom.com/share/354350c688fa487cbca57a466031cc52

### DIFF
--- a/frontend/web/components/pages/project-settings/tabs/ProjectPermissionsTab.tsx
+++ b/frontend/web/components/pages/project-settings/tabs/ProjectPermissionsTab.tsx
@@ -4,19 +4,20 @@ import React from 'react'
 import { useGetRolesQuery } from 'common/services/useRole'
 import { useGetProjectPermissionsQuery } from 'common/services/useProject'
 
-type PermissionsTabProps = {
+type ProjectPermissionsTabProps = {
   projectId: number
   organisationId: number
 }
 
-export const PermissionsTab = ({
+export const ProjectPermissionsTab = ({
   organisationId,
   projectId,
-}: PermissionsTabProps) => {
+}: ProjectPermissionsTabProps) => {
   const {
     data: rolesData,
     error: rolesError,
     isLoading: rolesLoading,
+    isSuccess: rolesSuccess,
   } = useGetRolesQuery({ organisation_id: organisationId })
 
   const {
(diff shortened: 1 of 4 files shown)


frontend/web/components/pages/project-settings/tabs/ProjectPermissionsTab.tsx
```diff
@@ -47,14 +48,17 @@ export const PermissionsTab = ({
...
-    <EditPermissions
-      tabClassName='flat-panel'
-      id={projectId}
-      level='project'
-      roleTabTitle='Project Permissions'
-      roles={rolesData?.results || []}
-      permissions={permissionsData || []}
-      onSaveUser={handleSaveUser}
-    />
+    <>
```
talissoncosta: just a quick question: do we need this wrapper in here ?

Zaimwa9: Not at all, it's an artifact from local twisting, i'll fix 😅

Zaimwa9:
@talissoncosta I got interrupted while filling the PR. I added a full loom comparing before / after on prod and local for better context

talissoncosta (approved):
Looks good @Zaimwa9. Good job, thanks for handling it!

Zaimwa9:
@matthewelwell I'd love a confirmation on whether the behavior is acceptable as the customer asked again about it. To me it's in line with what we can / should show (including the environment settings being buggy atm).

matthewelwell:
> @matthewelwell I'd love a confirmation on whether the behavior is acceptable as the customer asked again about it. To me it's in line with what we can / should show (including the environment settings being buggy atm).

@Zaimwa9 this looks ok to me as a temporary solution to fix the bug. I think that the real solution here would be to add an endpoint (or modify the permissions on the existing one, depending on the exposed data) to list roles for any user in the organisation like we've done for users and groups. 

This all ties in to the discussion we had with @asaphko regarding the 'Project Owner' role, since my hope is that it's just the project admin role. From what I can tell, it's already possible for project admins to set the permissions for users and groups at the project level, so we just need to extend it to roles as well and I think we've achieved what we set out to do.

PR #6810, opened 2026-03-01
fix: dark-mode-on-login-page

### LINKED ISSUE #6804: Dark mode preference is cleared on login
## Description

The dark mode setting is lost when logging in again (observed at least via Google OAuth). After enabling dark mode, logging out, and logging back in, the preference reverts to light mode.

## Steps to reproduce

1. Login to Flagsmith
2. Turn on dark mode
3. Log out
4. Login again (at least via Google OAuth)
5. Observe that dark mode has been reset to light mode

## Root cause

In `frontend/common/stores/account-store.js:351`, `AsyncStorage.clear()` is called during logout, which invokes `window.localStorage.clear()` and removes ALL localStorage keys — including the `dark_mode` preference. The dark mode setting is stored in localStorage via `frontend/web/project/darkMode.ts` and gets wiped alongside authentication data.

## Affected files

- `frontend/common/stores/account-store.js`
- `frontend/web/project/darkMode.ts`

kartik-lamkhade:
Fix #6804
fix(frontend): resolve dark mode resetting issue
fix(ui): correct body color in dark mode

### DIFF
--- a/frontend/common/stores/account-store.js
+++ b/frontend/common/stores/account-store.js
@@ -348,7 +348,7 @@ const controller = {
       store.loaded()
     } else if (!user) {
       store.ephemeral_token = null
-      AsyncStorage.clear()
+      AsyncStorage.removeItem('user')
       if (!data.token) {
         return
       }
--- a/frontend/common/stores/account-store.js
+++ b/frontend/common/stores/account-store.js
@@ -1,5 +1,5 @@
 import { matchPath } from 'react-router-dom'
-
+import { storageGet, storageSet } from 'common/safeLocalStorage'
 const Dispatcher = require('../dispatcher/dispatcher')
 const BaseStore = require('./base/_store')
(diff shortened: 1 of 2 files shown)


Zaimwa9 (changes requested):
Thanks for the contribution (and reactivity to this issue!). I have a couple of comments, mostly related to our identification of the issue.

The root issue is that in `frontend/web/components/pages/HomePage.tsx` L351 in `Card` we use the class `bg-white` that is not dark-mode aware.

So in addition to your fix (to keep the dark mode even after storage clearance) we would need to replace `bg-white` with `bg-body` that is dark mode aware.

Could you also rename the PR name to `Fix: xxxx` to pass the CI step please ? We should be good to go then!

Thanks again

Zaimwa9 (approved):
Image

Looks good to me! Thanks for the contribution :)

PR #6811, opened 2026-03-02
fix: defer data field in feature import/export list views

gagantrivedi:

## Changes

`FeatureExportListView` and `FeatureImportListView` inherit from `ListAPIView` and return querysets that fetch the `data` column (`TextField` holding full JSON export blobs, potentially megabytes each) from Postgres. The list serializers correctly exclude `data` from their fields, but Django still fetches the column and materialises it into Python strings for every row — only for the garbage collector to immediately discard them.

This PR adds `.defer("data")` to both `get_queryset` methods so the `data` column is excluded from the SQL `SELECT` entirely.

Ran the existing test suite for the import/export views — all 18 tests pass with no regressions.

### DIFF
--- a/api/features/import_export/views.py
+++ b/api/features/import_export/views.py
@@ -134,8 +134,10 @@ class FeatureExportListView(ListAPIView):  # type: ignore[type-arg]
             if user.is_environment_admin(environment):  # type: ignore[union-attr]
                 environment_ids.append(environment.id)
 
-        return FeatureExport.objects.filter(environment__in=environment_ids).order_by(
-            "-created_at"
+        return (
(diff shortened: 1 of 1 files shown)


Zaimwa9 (approved):
Approving with a minor NIT that I leave up to you as this is a punctual used feature. We could add an assertion in tests to ensure that `data` is part of the deferred field. I don't see it as non-blocking though

gagantrivedi:
> Approving with a minor NIT that I leave up to you as this is a punctual used feature. We could add an assertion in tests to ensure that `data` is part of the deferred field. I don't see it as non-blocking though

I did try that, but the test ended up being a mess — it was essentially just testing defer itself, which is Django code and not something we should be testing.

PR #6820, opened 2026-03-02
fix: Invoke backend logout API on logout button click to prevent broken authentication

itsmahbub:

### Bug
When a user logs out by clicking the "Log out" button, the logout API (/api/v1/auth/logout/) is not called. As a result, the session token is not invalidated and remains usable for API requests even after user logs out. When the user logs in again, the same token is reused instead of issuing a new one. If a token is leaked, an attacker may continue using the token indefinitely, even after the user logs out.

### Solution

Call to logout API is conditional on a configuration variable named `Project.cookieAuthEnabled` (https://github.com/Flagsmith/flagsmith/blob/main/frontend/common/stores/account-store.js#L359). Logout button click is supposed to clean up the session token in both frontend and backend, not just frontend. To prevent session token persistence after logout, we need to call the logout functionality in backend irrespective of configuration, otherwise, the logout process will remain partial. 

### Before fixing

There is no call to logout API when logout button is clicked.

Screenshot 2026-03-02 at 12 49 24 PM

As a result, the token is not invalidated in the backend.

### After applying fix
Logout API is called when logout button is clicked.

Screenshot 2026-03-02 at 1 40 29 PM


kyle-ssg (approved):

kyle-ssg:
Thank you for this @itsmahbub - I agree with the change!

PR #6827, opened 2026-03-02
ci: Increase flag-tests wait time

matthewelwell:
## Changes

Increases wait times in flag-tests after it was decreased in [this PR](https://github.com/Flagsmith/flagsmith/pull/6469). 

Context in Slack [here](https://flagsmith.slack.com/archives/CTF0THS2D/p1772478883612029).

## How did you test this code?

TODO

PR #6828, opened 2026-03-02
refactor(e2e): Replace arbitrary waits with state-based waiting in flag tests

talissoncosta:
## Summary

Improves the E2E test reliability and speed by replacing arbitrary timeout waits with proper state-based waiting.

## Changes

Replace fixed `t.wait(500)` timeouts with `waitForElementVisible('#try-it-results')` in three locations within flag-tests.ts.

### Why This is Better

**Current approach (PR #6827):**
- Uses `t.wait(500)` → `t.wait(2000)` and `t.wait(1500)` 
- Fixed delays regardless of actual response time
- Still has race conditions if CI is very slow
- Slower tests even when API responds quickly

**This approach:**
- Waits for `#try-it-results` element to appear (only exists when loading completes)
- Runs as fast as possible (no artificial delays)
- More reliable on slow CI environments (uses `LONG_TIMEOUT` of 40s)
- Semantically correct: waits for the exact element we need
- Consistent with E2E best practices used elsewhere in the codebase

### Technical Details

The `TryIt` component only assigns the `#try-it-results` ID when `!this.state.isLoading` ([TryIt.js:82](https://github.com/Flagsmith/flagsmith/blob/main/frontend/web/components/TryIt.js#L82)). This makes it the perfect indicator that:
1. The API request completed
2. The state updated  
3. The DOM rendered with results

## Test Plan

- [ ] Run E2E tests locally: `make test`
- [ ] Verify all E2E tests pass in CI
- [ ] Compare execution time with #6827 approach

## Related Issues

- Alternative to #6827 which increased timeouts from 500ms to 2000ms/1500ms
- Addresses the root cause mentioned in [this Slack thread](https://flagsmith.slack.com/archives/CTF0THS2D/p1772478883612029)

### DIFF
--- a/frontend/DESIGN_SYSTEM_AUDIT.md
+++ b/frontend/DESIGN_SYSTEM_AUDIT.md
@@ -1,0 +1,652 @@
+# Flagsmith Frontend Design System Audit
+
+**Issue**: #6606
+**Date**: 2026-02-27
+**Scope**: Code-first audit of token misuse, dark mode gaps, and component fragmentation
+
+---
+
+## Executive Summary
+
+This audit scanned the Flagsmith frontend codebase for design system inconsistencies across 7 areas (colours, typography, spacing, buttons, forms, icons, notifications) and catalogued 11 UI component categories.
+
+### Key findings
+
+| Severity | Count | Description |
+|----------|-------|-------------|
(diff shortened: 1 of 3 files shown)


talissoncosta:
Closing in favour of clean branch without extra files

PR #6829, opened 2026-03-02
refactor(e2e): Replace arbitrary waits with state-based waiting in flag tests

talissoncosta:
## Summary

Improves E2E test reliability and speed by replacing arbitrary timeout waits with proper state-based waiting in flag-tests.

## Changes

Replace fixed `t.wait(500)` timeouts with `waitForElementVisible('#try-it-results')` in three locations within flag-tests.ts.

## Why This Approach is Better

**Fixed timeouts:**
```typescript
await t.wait(500)  // Hope this is enough
```
- Too short = flaky tests
- Too long = slow tests  
- Doesn't adapt to different environments (local vs staging vs CI)

**State-based waiting:**
```typescript
await waitForElementVisible('#try-it-results')  // Wait for actual state
```
- Fast when API responds quickly (no artificial delays)
- Reliable when API is slow (uses 40s timeout)
- Works consistently across all environments

## Technical Details

The `#try-it-results` element only receives its ID when `!this.state.isLoading` ([TryIt.js:82](https://github.com/Flagsmith/flagsmith/blob/main/frontend/web/components/TryIt.js#L82)), guaranteeing:
- API request completed
- State updated  
- Results rendered and ready to parse

This is the standard pattern already used throughout our E2E test suite.

## Test Plan

- [ ] Run E2E tests locally: `make test`
- [ ] Verify all E2E tests pass in CI

### DIFF
--- a/frontend/e2e/tests/flag-tests.ts
+++ b/frontend/e2e/tests/flag-tests.ts
@@ -48,7 +48,7 @@ export default async function () {
   log('Try it')
   await t.wait(2000)
   await click('#try-it-btn')
-  await t.wait(500)
+  await waitForElementVisible('#try-it-results')
   let json = await parseTryItResults()
   await t.expect(json.header_size.value).eql('big')
   await t.expect(json.mv_flag.value).eql('big')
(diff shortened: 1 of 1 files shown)


matthewelwell:
@talissoncosta I don't think this will work. The issue is that the tests will also need to wait for the feature change to propagate from the Core API, to dynamodb (London), and then get replicated to the dynamodb table in the US. That can take a second or 2 some times. If we want to make this more robust, we'd need to have it keep trying the try it button until it gets the result it's expecting. Until then, I suggest we go with the longer static timeouts.

matthewelwell:
Closing based on the above comment. We can perhaps revisit replacing at least the first wait once we merge the playwright changes.

PR #6830, opened 2026-03-03
fix: replace numpy/scipy with pure Python in experiment analytics

gagantrivedi:

## Changes

Removes numpy and scipy dependencies from the API, replacing them with pure Python (stdlib `math` and `random`) implementations. These libraries added significant per-worker RSS overhead but are only used by the demo experiment analytics endpoint.

- Implement G-test (log-likelihood ratio) using the incomplete gamma function
- Use `random.betavariate` for Bayesian Monte Carlo simulation (10k samples)
- Remove numpy and scipy from `pyproject.toml` and regenerate `poetry.lock`
- Convert tests from class-based to function-based to match project conventions
- Remove stale test assertion

- All 7 existing tests pass (`pytest tests/unit/app_analytics/test_experiments.py`)
- Mypy passes in strict mode
- Verified G-test and chi-squared survival function are numerically identical to scipy (errors at machine epsilon ~1e-17)
- Verified Bayesian Monte Carlo results match scipy/numpy ground truth
- Benchmarked end-to-end latency: ~16ms per call (acceptable for a demo endpoint)


Zaimwa9 (approved):
Thanks, that's a lot of maths for fake graphs 😄

PR #6834, opened 2026-03-03
docs: add event types to audit log webhook

Holmus:
## Summary
- Updated the audit log event types to match what the codebase seems to generate.
- Added missing categories: change requests, flag state changes, segment overrides, identity override lifecycle, feature versioning, release pipelines, and phased rollouts
- Removed outdated entries (segment rule/condition events that are no longer individually tracked)
- Added a reference table for `related_object_type` payload values

### DIFF
--- a/docs/docs/administration-and-security/governance-and-compliance/audit-logs.md
+++ b/docs/docs/administration-and-security/governance-and-compliance/audit-logs.md
@@ -4,8 +4,6 @@ sidebar_label: Audit Logs
 sidebar_position: 30
 ---
 
-import Tabs from '@theme/Tabs'; import TabItem from '@theme/TabItem';
-
 Every action taken within the Flagsmith administration application is tracked and logged. This allows you to easily
 retrace the events and values that flags, identities and segments have taken over time.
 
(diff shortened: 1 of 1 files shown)


docs/docs/administration-and-security/governance-and-compliance/audit-logs.md
```diff
@@ -76,30 +76,100 @@ hmac.compare_digest(expected_signature, received_signature) is True
...
+- Flag state / remote config updated by a change request
+
+### Identity Overrides
+
+- Identity override created
+- Identity override updated (flag state / remote config value)
+- Identity override value updated (remote config)
+- Identity override deleted
+- Identity override scheduled
+- Edge identity feature override created, updated, or deleted
```
matthewelwell: We should probably split this into a tabular view for SaaS / self-hosted. Something like this:


```suggestion
### Identity Overrides

<Tabs groupId="deployment" queryString>

<TabItem value="self-hosted" label="Self-Hosted / Private Cloud">
- Identity override created
- Identity override updated (flag state / remote config value)
- Identity override value updated (remote config)
- Identity override deleted
- Identity override scheduled
</TabItem>

<TabItem groupId="saas" label="SaaS (app.flagsmith.com)">
- Edge identity feature override created, updated, or deleted
</TabItem>

</Tabs>
```

That being said, I'm surprised that we don't have more granularity around Edge Identity overrides... ?

matthewelwell: See [here](https://github.com/Flagsmith/flagsmith/blob/3a98aca251260b04e23ab8d4097a92e40b383161/api/edge_api/identities/tasks.py#L130-L146) - it looks like we have distinct audit log records for each action, but this makes it look like we only have a single record regardless of the action taken.

matthewelwell (changes requested):

docs/docs/administration-and-security/governance-and-compliance/audit-logs.md
```diff
@@ -51,7 +51,7 @@ Flagsmith will send a `POST` request to your webhook URL with the following payl
 }
 ```
 
-`related_object_type` is one of: `FEATURE`, `FEATURE_STATE`, `SEGMENT`, `ENVIRONMENT`, `CHANGE_REQUEST`, `EDGE_IDENTITY`, `IMPORT_REQUEST`, `EF_VERSION`, `FEATURE_HEALTH`, `RELEASE_PIPELINE`.
```
adamvialpando, line 54: RELEASE_PIPELINE is removed but exists in the enum. Likely a regression

adamvialpando (commented):
Aside from this it looks good to me.

PR #6840, opened 2026-03-04
fix: Identity overrides empty in SDK document with v2 versioning

### LINKED ISSUE #6839: Identity overrides are empty in SDK document when Feature Versioning is enabled

- [x] Self Hosted with Docker
- [ ] SaaS at flagsmith.com

## Summary

When Feature Versioning (v2) is enabled on an environment, identity overrides appear in the SDK document but their `identity_features` list is empty. This breaks local evaluation for identities that have overrides.

## Expected behaviour

With `use_identity_overrides_in_local_eval` enabled and Feature Versioning enabled, the environment document should include identity overrides with their `identity_features` populated, for example:

```
{
  "identity_overrides": [
    {
      "identifier": "Development_user_123456",
      "identity_features": [
        {
          "feature": { "id": 91, "name": "abc", "type": "STANDARD" },
          "enabled": false,
          "feature_state_value": "devuser-val",
          ...
        }
      ]
    }
  ]
}
```

## Actual behaviour
The identity_overrides list contains the identities, but each identity has an empty `identity_features` list:
```
{
  "identity_overrides": [
    {
      "identifier": "Development_user_123456",
      "identity_features": [],
      ...
    }
  ]
}
```

As a result, local evaluation returns the default / segment override flag values instead of the identity-specific overrides.

## Root cause
Identity overrides are not part of the versioning system (see `FeatureStateManager.get_live_feature_states`). When v2 versioning is enabled, `_create_initial_feature_versions` only assigns `environment_feature_version` to environment defaults and segment overrides, not to identity overrides. Identity feature states therefore have `environment_feature_version_id = None`.

In `map_identity_to_engine`, identity features are filtered through `_get_prioritised_feature_states`, which skips feature states where `is_live` is `False`. For v2 versioning, `FeatureState.is_live` requires `environment_feature_version_id is not None` and `environment_feature_version.is_live`. Identity overrides fail this check and are filtered out, producing an empty list.

## Related
- Issue #3911 (Versioning V2 issues) – fixed the crash when `environment_feature_version` was None, but identity overrides are still filtered out
- PR #3913 – introduced the `environment_feature_version_id` is not None check that causes this behaviour

### Steps To Reproduce

1. Create a project and environment.
2. Enable Feature Versioning (v2) on the environment (Environment Settings → Feature Versioning).
3. Create a feature (e.g. "abc") in the project.
4. Enable "Use identity overrides in local evaluation" for the environment.
5. Create an identity (e.g. identifier `Development_user_123456`).
6. Add an identity override for that identity (e.g. set "abc" to `devuser`).
7. Fetch the SDK document (e.g. via `/api/v1/environments/{api_key}/document/` or the SDK in local evaluation mode).
8. Inspect the `identity_overrides` in the response.

## Issue
Each identity has `identity_features: []`.

### Expected behavior

Each identity in `identity_overrides` has a non-empty `identity_features` list with the override(s).

### Screenshots

_No response_

khvn26:

## Changes

Closes #6839.

In this PR, we fix a bug where identity overrides in the SDK environment document have empty `identity_features` when Feature Versioning v2 is enabled.

- Added an integration test reproducing the issue.
- Added `test_feature_state_is_live__identity_override_v2_versioning__returns_true` unit test targeting the model property directly.
- Added `test_get_environment_document__identity_overrides` parametrised over both v1 and v2 environments, asserting the full structure of identity overrides in the SDK document response (including non-empty `identity_features`).

### DIFF
--- a/api/features/models.py
+++ b/api/features/models.py
@@ -618,6 +618,8 @@ class FeatureState(
     @property
     def is_live(self) -> bool:
         if self.environment.use_v2_feature_versioning:  # type: ignore[union-attr]
+            if self.identity_id is not None:
+                return True
             return (
                 self.environment_feature_version_id is not None
                 and self.environment_feature_version.is_live  # type: ignore[union-attr]
--- a/api/tests/integration/sdk/test_sdk_environment_document.py
+++ b/api/tests/integration/sdk/test_sdk_environment_document.py
@@ -1,0 +1,34 @@
+from rest_framework import status
+from rest_framework.test import APIClient
+
+from features.versioning.tasks import enable_v2_versioning
+
+
+def test_get_environment_document__v2_versioning_with_identity_override__identity_features_populated(
+    server_side_sdk_client: APIClient,
+    feature: int,
+    identity_featurestate: int,
+    environment: int,
+) -> None:
+    """Reproduce https://github.com/Flagsmith/flagsmith/issues/6839
+
+    When Feature Versioning v2 is enabled, identity overrides should still
+    appear in the SDK document with their identity_features populated.
+    """
+
+    # Given - v2 feature versioning is enabled after identity override exists
+    enable_v2_versioning(environment)
+
+    # When
+    url = "/api/v1/environment-document/"
+    response = server_side_sdk_client.get(url)
+
+    # Then
+    assert response.status_code == status.HTTP_200_OK
+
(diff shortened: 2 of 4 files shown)


Zaimwa9 (dismissed):
👍

Zaimwa9 (approved):

PR #6842, opened 2026-03-04
fix(Logging): Omit org name from logging

emyller:
Remove an occurrence of organisation names in logging, because it should not be exposed.


khvn26 (dismissed):
👍

khvn26 (approved):

Issue #6832, opened 2026-03-03
Optimise annotate_feature_queryset_with_code_references_summary for feature list endpoint


gagantrivedi:
## Summary

`annotate_feature_queryset_with_code_references_summary` in `api/projects/code_references/services.py` adds expensive per-row annotations to the feature list queryset, slowing down the `GET /api/v1/projects/{project_id}/features/` endpoint.

## Problem

The function annotates each feature with `code_references_counts` using:

1. **`jsonb_path_exists` correlated subquery** — for every feature row, it scans `FeatureFlagCodeReferencesScan.code_references` JSON to find the last scan containing that feature name.
2. **`jsonb_path_query_array` + `jsonb_array_length` correlated subquery** — for every feature row, it extracts and counts matching references from the JSON column.
3. **Nested subquery** — `last_feature_found_at` is itself a subquery nested inside the `counts_by_repository` subquery, compounding the cost.

These JSON path operations run per-feature, per-repository and do not benefit from standard indexes. For projects with many features or many scan records, this results in significant query time even when code references are unused.

## Current mitigation

An early-exit check (`has_scans`) skips annotation when no scans exist for the project, but this doesn't help projects that do have scans.

## Location

- Function: `annotate_feature_queryset_with_code_references_summary` in [`api/projects/code_references/services.py:31`](https://github.com/Flagsmith/flagsmith/blob/main/api/projects/code_references/services.py#L31)
- Called from: [`api/features/views.py:220`](https://github.com/Flagsmith/flagsmith/blob/main/api/features/views.py#L220)

emyller:
Aiming towards a better fix here: https://github.com/Flagsmith/flagsmith/issues/5932#issuecomment-4328705597

PR #6844, opened 2026-03-04
perf: Denormalise feature names on Code Reference scans and add GIN index

### LINKED ISSUE #6836: Improve query performance when listing features that have code references
This is a follow on issue from https://github.com/Flagsmith/flagsmith/issues/6785. 

In the previous issue, we improved the query performance by a significant factor by adding a new index. This brought the response time of the list features endpoint on our own Flagsmith Website project down from ~15s to ~6s, but this is still not sufficient.

We should investigate a few options for improving this behaviour. These options are listed below, but are not exhaustive and other avenues could and should be pursued if they exist.  

<details>
<summary><h2>Option 1 [by Claude] - Denormalise feature names on CodeReference</h2></summary>

As suggested by Claude

```
Add a feature_names text-array column with a GIN index [medium effort]

What it does: Denormalise by storing the distinct feature names referenced in each scan in a feature_names = ArrayField(TextField()) column, populated at save time. Add a GIN index on it. Change the jsonb_path_exists(...) filter in services.py to feature_names__contains=[feature_name].

Impact: High. The GIN index on a text array means the last_feature_found_at filter becomes a fast indexed lookup (microseconds) instead of a full JSON array scan (milliseconds). Combined with Option A, this could reduce execution time from ~13s to sub-second.

Effort: Medium. Requires:
 - Model change + migration with a RunPython backfill from existing code_references JSON
 - Override save() to populate feature_names
 - Update services.py:53–71 to replace jsonb_path_exists with feature_names__contains=...
```
</details>

<details>
<summary><h2>Option 2 [by Claude] - Refactor to handle more logic in python vs ORM</h2></summary>

As suggested by Claude

```
Restructure to eliminate the correlated last_feature_found_at subquery [higher effort, highest impact]

What it does: The 4,815 SubPlan 1 executions exist because last_feature_found_at is a correlated subquery nested inside counts_by_repository, making it run once per (feature, scan-row) pair. A better approach is to compute last_feature_found_at for all relevant (feature_name, repository_url) pairs in one pass — either as a lateral join, a CTE, or by pulling it out of the ORM annotation entirely and doing the computation in Python.

Python-side approach (avoids complex ORM gymnastics):
 1. Fetch the latest scan per repository_url for the project — 1 simple query.
 2. In Python, iterate over those scans' code_references JSON to build a dict[feature_name, list[summary]].
 3. Attach the result to features post-queryset (in the view or serializer), bypassing the ORM annotation altogether.

This reduces the problem to ~2 simple queries + fast Python dict lookups, and removes the 4,815-execution explosion entirely. The downside is it requires restructuring the view/serializer to not rely on the queryset annotation.
```
</details>

<details>
<summary><h2>Option 3 - Refactor to grab code references asynchronously</h2></summary>

Note: this would constitute a breaking API change, but is probably acceptable for a beta feature. 

We could look at removing the code references from the main features list endpoint, creating a separate endpoint, and having the FE stitch the data together instead. This might allow us to improve the performance of the endpoint, and would also mean that the FE can render out the list of features, and add the github reference counts once they are ready. 

</details>

matthewelwell:
## Changes

Closes #6836 

This PR denormalises the feature names directly into an ArrayField on the `FeatureFlagCodeReferencesScan`. The feature names were previously nested inside the JSONField which caused performance issues with the database queries when listing features for the dashboard. 

Added tests.

### DIFF
--- a/api/projects/code_references/migrations/0003_add_feature_names.py
+++ b/api/projects/code_references/migrations/0003_add_feature_names.py
@@ -3,7 +3,7 @@ from django.db import migrations, models
 
 
 def _backfill_feature_names(apps: object, schema_editor: object) -> None:
-    FeatureFlagCodeReferencesScan = apps.get_model(  # type: ignore[union-attr]
+    FeatureFlagCodeReferencesScan = apps.get_model(  # type: ignore[attr-defined]
         "code_references", "FeatureFlagCodeReferencesScan"
     )
     scans = list(FeatureFlagCodeReferencesScan.objects.all())
--- a/api/projects/code_references/migrations/0003_add_feature_names.py
+++ b/api/projects/code_references/migrations/0003_add_feature_names.py
@@ -2,7 +2,7 @@ from django.contrib.postgres.fields import ArrayField
 from django.db import migrations, models
 
 
-def _backfill_feature_names(apps: object, schema_editor: object) -> None:
+def backfill_feature_names(apps: object, schema_editor: object) -> None:
     FeatureFlagCodeReferencesScan = apps.get_model(  # type: ignore[attr-defined]
         "code_references", "FeatureFlagCodeReferencesScan"
     )
@@ -27,7 +27,7 @@ class Migration(migrations.Migration):
             name="feature_names",
         ),
         migrations.RunPython(
-            _backfill_feature_names,
+            backfill_feature_names,
             reverse_code=migrations.RunPython.noop,
         ),
     ]
--- a/api/projects/code_references/migrations/0003_add_feature_names.py
+++ b/api/projects/code_references/migrations/0003_add_feature_names.py
@@ -1,9 +1,11 @@
+from django.apps.registry import Apps
 from django.contrib.postgres.fields import ArrayField
 from django.db import migrations, models
+from django.db.backends.base.schema import BaseDatabaseSchemaEditor
 
 
-def backfill_feature_names(apps: object, schema_editor: object) -> None:
-    FeatureFlagCodeReferencesScan = apps.get_model(  # type: ignore[attr-defined]
+def backfill_feature_names(apps: Apps, schema_editor: BaseDatabaseSchemaEditor) -> None:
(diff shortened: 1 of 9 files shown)


api/projects/code_references/migrations/0003_add_feature_names.py
```diff
@@ -0,0 +1,35 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("code_references", "0002_add_project_repo_created_index"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="featureflagcodereferencesscan",
+            field=ArrayField(models.TextField(), default=list),
```
matthewelwell, line 28: I'm fairly sure there are other fields in this app that don't support oracle, but we probably need to think on how to exclude this from EE.

gagantrivedi:
```
The PR migrates the last_feature_found_at subquery and the get_code_references_for_feature_flag function to use the new feature_names array, but the count calculation in
  counts_by_repository (the heavier operation) still does a full jsonb_path_query_array + jsonb_array_length against the raw code_references JSON blob:

  count=Func(
      Func(
          F("code_references"),
          Value("$[*] ? (@.feature_name == $feature_name)"),
          JSONObject(feature_name=F("feature_name")),
          function="jsonb_path_query_array",
      ),
      function="jsonb_array_length",
  ),

  This generates SQL like:

  jsonb_array_length(
    jsonb_path_query_array(
      V0."code_references",
      '$[*] ? (@.feature_name == $feature_name)',
      JSONB_BUILD_OBJECT('feature_name', "features_feature"."name")
    )
  )
```

The only solution I see here is to accept that the data model doesn't work for how we query it, and change it to flatten everything and get rid of the JSON column. Otherwise, we'd keep going in circles making the data access layer more complicated than it needs to be.

matthewelwell:
Closing while we work on [this issue](https://github.com/Flagsmith/flagsmith/issues/6832) instead.

PR #6845, opened 2026-03-04
feat: Add created_by to MasterAPIKey model

matthewelwell:
## Changes

Add 'created by' to Organisation API keys for better auditing. 

Updated existing unit test and tested frontend changes locally.


Zaimwa9 (approved):
Code looks good and safe to me, haven't tested it yet, will do but approving to not block

frontend/web/components/AdminAPIKeys.js
```diff
@@ -458,6 +459,17 @@ export default class AdminAPIKeys extends PureComponent {
               <Row className='table-header'>
                 <Flex className='table-column px-3'>API Keys</Flex>
                 <Flex className='table-column'>Created</Flex>
+                {Utils.getFlagsmithHasFeature(
```
Zaimwa9, line 474: Actually the organisation store is empty on this page. It needs to be dispatched the old way (hello ClassComponents).

You'll need something like this otherwise `organisationStore.model?.users` is empty and fallbacks on `-`

```
import AppActions from 'common/dispatcher/app-actions'

  componentDidMount() {
    this.fetch()
    AppActions.getOrganisation(this.props.organisationId)
    OrganisationStore.on('change', this.handleOrgStoreChange)
  }

  componentDidUpdate() {
    if (this.props.organisationId === this.state.organisationId) return

    this.fetch()
    AppActions.getOrganisation(this.props.organisationId)
    this.setState({ organisationId: this.props.organisationId })
  }

  componentWillUnmount() {
    OrganisationStore.off('change', this.handleOrgStoreChange)
  }

  handleOrgStoreChange = () => {
    this.forceUpdate()
  }
```

Zaimwa9 (changes requested):
Sorry, old store pattern required to get the users.

NIT: Columns are slightly misaligned

Image

frontend/web/components/AdminAPIKeys.js
matthewelwell: I'm a bit confused by this comment for a couple of reasons: 

 1. The comment is attached to a line related to the flagsmith SDK, and determining if the flag is enabled or not. 
 2. If the organisation store was empty, wouldn't we not see any names populated in the list? Doesn't your screenshot prove that's not the case? 

I'm sure there's something I'm missing here, but I'd like to understand a little more about what the issue is before I blindly fix it with your suggestion 😃

Zaimwa9: Yes my bad, I don't know how I ended up on this line:
1. Wrong comment line from my side
2. I wasn't clear on this part but the screenshot comes from trying the fix locally to get to the root cause. The response payload was correct but defaulting to `-` because of the empty store

Zaimwa9 (dismissed):
Approving following offline discussion unless we have a stable way to reproduce the fallback behavior I experienced

matthewelwell:
> Approving following offline discussion unless we have a stable way to reproduce the fallback behavior I experienced

@Zaimwa9 I was able to reproduce this consistently so I added the fix you suggested with a little bonus help from the C-dog.

Zaimwa9 (approved):
🧟 👍

PR #6848, opened 2026-03-04
test: fix flaky tests

matthewelwell:
## Changes

Fixes 2 flaky unit tests: 

1. `test_get_feature_states_for_identity` - this test failed when the multivariate feature values were returned in reverse order since the test assumed they were returned in the order they were created.
2. `test_identities_endpoint_returns_value_for_segment_if_rule_type_percentage_split_and_identity_in_segment` - this test failed if the returned value from `get_hashed_percentage_value` was between 0.99 and 0.995. The maths when creating the Condition then meant that the percentage value was 99, and hence the user was deemed outside of the segment. Since the logic for actually determining segment membership is covered elsewhere, I just set the percentage to 100 to guarantee that the identity will be in the segment. 

Ran the tests 100 times consecutively (using `pytest.mark.parametrise`) to verify that the tests consistently passed.

### DIFF
--- a/api/tests/unit/environments/identities/test_unit_identities_views.py
+++ b/api/tests/unit/environments/identities/test_unit_identities_views.py
@@ -1,4 +1,5 @@
 import json
+import logging
 import urllib
 from typing import Any
 from unittest import mock
@@ -37,6 +38,9 @@ from organisations.models import Organisation
 from permissions.models import PermissionModel
 from projects.models import Project, UserProjectPermission
 from segments.models import Condition, Segment, SegmentRule
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
 
 
 def test_should_return_identities_list_when_requested(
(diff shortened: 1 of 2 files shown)


api/tests/unit/environments/identities/test_unit_identities_views.py
```diff
@@ -38,6 +39,9 @@
 from projects.models import Project, UserProjectPermission
 from segments.models import Condition, Segment, SegmentRule
 
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
+
```
emyller: Is this needed?

matthewelwell: Nope, good catch! Debugging flaky tests was hard so this was just temporary. I removed the logs, but not the logging import!

emyller (approved):

PR #568, opened 2021-12-08
Typo fixes for frontend application

zefixlluja:
* fixes typos for MULTIVARIATE_DESCRIPTION and AUDIT_WEBHOOKS_DESCRIPTION
* aligns writing style

PR #6863, opened 2026-03-06
fix: store-created-at-in-identity-override-v2-document

Zaimwa9:

Contributes to Edge-api [#568](https://github.com/Flagsmith/edge-api/issues/568)

## Changes
- Added `created_date` to `IdentityOverrideV2` with `default_factory=timezone.now`

⚠️ ⚠️ 
On update, `created_at ` is reseted to `now()`. We `put_item` and replace the full document so to preserve the original `created_date`, except if I am missing, we would need an extra Dynamo read. Which to me is not worth in the context of the snapshot tests. We just need it to be fix between 2 calls to the `environment_document` and that should ensure it

- Updated tests mostly with `freezegun` 
- One new test

### DIFF
--- a/api/environments/dynamodb/types.py
+++ b/api/environments/dynamodb/types.py
@@ -5,7 +5,8 @@ from datetime import datetime
 
 import boto3
 from django.conf import settings
-from pydantic import BaseModel
+from django.utils import timezone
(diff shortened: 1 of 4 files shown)


khvn26 (approved):
Yep, LGTM 👍

PR #6865, opened 2026-03-06
fix: `/api/v1/environment-document` N+1 query issue related to the number of segment overrides in a version

matthewelwell:
## Changes

Fixes an N+1 issue in the SDK environment document endpoint. 

Updated an existing test case.

### DIFF
--- a/api/environments/managers.py
+++ b/api/environments/managers.py
@@ -38,7 +38,10 @@ class EnvironmentManager(SoftDeleteManager):  # type: ignore[misc]
                 Prefetch(
                     "project__segments__feature_segments__feature_states",
                     queryset=FeatureState.objects.select_related(
-                        "feature", "feature_state_value", "environment"
+                        "feature",
+                        "feature_state_value",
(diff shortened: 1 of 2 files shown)


emyller (dismissed):

emyller (approved):
:shipit:

khvn26 (changes requested):
Given that the issue is not related to EFVs after all, we need a better test.

matthewelwell:
> Given that the issue is not related to EFVs after all, we need a better test.

The thing is, that it _is_ related to versioning. The issue with this test was that, while it enabled versioning, it then created all of the feature states without caring about that versioning. So the test previously was not testing what we thought it was. By moving the `enable_v2_versioning` function call to _after_ the set up of the data, we are testing what we expected to be testing. 

Edit: I left the additional versions there as a bonus in case we ever do accidentally add an N+1 issue related to the number of versions.

khvn26 (approved):
LGTM

PR #6874, opened 2026-03-08
fix(api): handle invalid project IDs when creating environments

MartinLam12:

## Changes

Contributes to 

_Please describe._
Handle invalid project values in POST /api/v1/environments/ by catching ValueError and TypeError in EnvironmentPermissions.has_permission(). Previously, passing a non-integer value (e.g., "<Project ID>") caused an unhandled ValueError resulting in a 500 error. The permission check now returns False for invalid project IDs, returning a 403 instead of a 500

_Please describe._

Added two unit tests in test_unit_environments_permissions.py: test_create_environment__invalid_project_id_string__returns_false — verifies a non-numeric string project ID returns False test_create_environment__none_project_id__returns_false — verifies a missing/null project ID returns False


matthewelwell:
Hi @MartinLam12 , thanks for the contribution. Could you please update the PR title as per [this failing CI check](https://github.com/Flagsmith/flagsmith/pull/6874/checks?check_run_id=66176220463), and also resolve the comment review from Cursor?

MartinLam12:
I updated the PR title and also resolved the comment review from Cursor.

MartinLam12:
@matthewelwell Hello, can you take another look? Thanks

api/tests/unit/environments/permissions/test_unit_environments_permissions.py
```diff
@@ -146,6 +146,39 @@ def test_project_user_without_create_environment_permission_cannot_create_enviro
     assert result is False
 
 
+# created additional tests to cover edge cases around project ID validation in environment creation
```
matthewelwell:
```suggestion
```

I appreciate the description here, but this comment should not live in the code. It is commentary around the PR, not the code necessarily. This type of commentary should be included in the PR description and/or a comment from a self-review, we don't want it to live permanently in the code.

matthewelwell (changes requested):
Added one small additional comment, but looks good otherwise.

PR #6877, opened 2026-03-09
fix: replace OFFSET pagination with keyset pagination in identity migration

### LINKED ISSUE #6876: Identity migration fails with statement_timeout
## Problem

`IdentityMigrator` identity queries hit `statement_timeout` after the timeout was set to 1 minute. The current OFFSET-based pagination generates increasingly expensive queries as it pages through the table.

gagantrivedi:

## Changes

Closes #6876

Replace OFFSET-based `Paginator` in `IdentityMigrator` with keyset pagination (`WHERE pk > last_pk ORDER BY pk LIMIT N`) to avoid `statement_timeout` failures. The identity queryset building is extracted into a public `iter_identities_in_chunks` static method with proper type hints.

Added relevant unit tests and verified the actual generated SQL is performant.


Zaimwa9 (approved):
LGTM, might be worth testing it on staging

Issue #6606, opened 2026-01-28
Audit frontend UI — identify inconsistencies and consolidation opportunities


talissoncosta:
## Summary

Audit the frontend UI to identify visual inconsistencies, consolidation opportunities, and areas for improvement.

## Background

The UI has grown organically over time. There's no Figma design system for the app (the existing Figma file covers the website only). We need to take stock of what we have, find inconsistencies, and reduce them.

### Visual inconsistencies
- [ ] Colours — hardcoded values, missing tokens, dark mode issues
- [ ] Typography — font sizes, weights, line heights
- [ ] Spacing — padding, margins, gaps
- [ ] Button styles and variants
- [ ] Form inputs
- [ ] Icons usage

### Component duplication
- [ ] Identify components with multiple implementations of the same pattern
- [ ] Flag consolidation opportunities (e.g. confirmation modals, dropdowns)

### Visual review
- [ ] Screenshot key pages and components (light + dark mode)
- [ ] Identify UI patterns that could be improved — component design, page layouts, overall look and feel

## Output

- Prioritised list of findings (quick wins vs. longer-term consolidations)
- Visual inventory with improvement suggestions

talissoncosta:
## Approach

This audit combines a code scan with a visual review to get a full picture of where we stand.

### Steps

1. **Code audit** (done) — scanned SCSS and TSX files for hardcoded colour values, missing dark mode overrides, inconsistent spacing/typography, and catalogued recurring UI components. Identified 85 findings across colours, typography, spacing, buttons, forms, icons, and notifications.

2. **Visual review** (in progress) — screenshotting key pages and components in both light and dark mode to spot inconsistencies that aren't obvious from code alone, and identify patterns that could be improved.

3. **Prioritised report** — distil findings into a short list of concrete, actionable tasks: quick wins (e.g. switching hardcoded icon colours to `currentColor`), and bigger consolidation opportunities (e.g. reducing 13 confirmation modal implementations to one).

4. **Follow-up issues** — create individual tickets for the top findings so they can be picked up independently.

The goal isn't to build a design system ( for now ) — it's to reduce inconsistencies and flag consolidation opportunities that make the UI more consistent and the codebase easier to maintain.

talissoncosta:
## Audit Results

[...]

Issue #4432, opened 2024-07-31
Project setting: Enforce feature ownership


kyle-ssg:
### Is your feature request related to a problem? Please describe.

To improve governance, we should expose a setting that forces users to assign features to a group or user when creating.

### Describe the solution you'd like.

Add a project setting that forces feature assignment

 - It should be possible to add the assigned user / group on feature creation
 - If an owner exists, we should display this in the header of the feature modal

### Describe alternatives you've considered

N/A

Issue #6866, opened 2026-03-06
Decouple Button from Redux store (dead feature prop)


talissoncosta:
## Problem

`Button.tsx` imports `Utils` from `common/utils/utils`, which pulls in the entire Redux store + Flux store + CommonJS `_data.js` chain **at import time**. This blocks Storybook from importing Button or any of the 150+ components that depend on it.

The `Utils` import exists solely for `Utils.getPlansPermission(feature)` — but **zero consumers pass the `feature` prop to Button**. It's dead code.

Current Storybook Button stories use raw `<button>` HTML elements with CSS classes (visual replicas), not the real `<Button>` component.

## Files

- `web/components/base/forms/Button.tsx` — remove `Utils`, `PaidFeature`, `PlanBasedBanner` imports and dead `feature` prop logic
- `stories/Buttons.stories.tsx` — replace raw `<button>` elements with real `<Button>` component

### Button.tsx

1. Remove `import Utils, { PaidFeature } from 'common/utils/utils'`
2. Remove `import PlanBasedBanner from 'components/PlanBasedAccess'`
3. Remove `feature` from `ButtonType` props interface
4. Remove `feature` from destructured props
5. Remove `const hasPlan = feature ? Utils.getPlansPermission(feature) : true`
6. Simplify render: remove `hasPlan` conditional branching — always render standard button/anchor
7. Keep the `href` anchor branch (used for link-style buttons), just remove plan-gating logic within it

### Storybook stories

1. Import the real `Button` component instead of using raw `<button>` elements
2. Replace `<button className="btn btn-primary">` with `<Button theme="primary">`
3. Add an Interactive story with Storybook args/controls
4. Keep the dark mode gap documentation (valuable audit evidence)

## Acceptance Criteria

- [ ] `Button.tsx` has no imports from `common/utils/utils` or `components/PlanBasedAccess`
- [ ] `ButtonType` no longer includes `feature` prop
- [ ] `npm run typecheck` passes
- [ ] Storybook Button stories render the real `<Button>` component
- [ ] All existing Button functionality (href, icons, themes, sizes, disabled) works unchanged

## Context

Part of the design system audit (#6606). See `frontend/DESIGN_SYSTEM_ISSUES.md` on `chore/design-system-audit-6606` branch for full issue list.

PR #6883, opened 2026-03-09
chore: design system colour tokens, primitives & Storybook

### LINKED ISSUE #6904: Proposal: Introduce semantic colour tokens and primitive palette
## Problem

The frontend has no formal colour system. 280+ hardcoded hex values in TSX files, inverted tonal scales, alpha colour RGB mismatches, and three parallel dark mode mechanisms that don't compose.

## Solution

1. **Primitive palette** — full 50–950 tonal scales for all colour families in `_primitives.scss`
2. **Semantic tokens** — CSS custom properties in `_tokens.scss` organised by role (text, surface, icon, stroke, brand, feedback) with light/dark mode variants
3. **TS exports** — `common/theme/tokens.ts` for use in inline styles, automatically responds to theme changes

## Progress

- [ ] PR #6883 (`chore/design-system-tokens`) — primitives, tokens, Storybook stories
- [x] #6866 — Decouple Button from Redux (blocker for Storybook stories)

## Acceptance Criteria

- [ ] Semantic tokens cover text, surface, icon, stroke, brand, feedback categories
- [ ] Primitive scales defined for all colour families (50–950)
- [ ] Light and dark mode values defined for all tokens
- [ ] Tokens available as both CSS custom properties and TS exports
- [ ] No visual regression — existing UI renders identically

---
Part of the Design System Audit (#6606) · Epic: #6882

talissoncosta:

## Changes

Contributes to #6606

Design system token foundation with Storybook, Chromatic visual regression, and MCP integration for AI-assisted development.

### 1. Token architecture

Single source of truth: `common/theme/tokens.json` defines all tokens with values, dark mode overrides, and usage descriptions. A codegen script (`scripts/generate-tokens.mjs`) generates four output files — `tokens.ts`, `_tokens.scss`, `_token-utilities.scss`, and a flat JSX MCP reference story. lint-staged auto-runs the codegen on commit.

**Primitive → Semantic reference chain:**
- Primitives defined once as CSS custom properties (77 colours across 7 families + slate-1000)
- Semantic tokens reference primitives via `var(--primitive)` — zero raw hex in the semantic layer
- Alpha/opacity variants use OKLCH relative colour syntax: `oklch(from var(--purple-600) l c h / 0.08)`
- Changing a primitive value updates all semantic tokens that reference it

> **Why OKLCH?** [OKLCH](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/oklch) is a perceptually uniform colour space. Its [relative colour syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_colors/Relative_colors) lets us derive alpha variants directly from a primitive — `oklch(from var(--purple-600) l c h / 0.08)` keeps the same lightness, chroma, and hue but applies 8% opacity. This means changing `--purple-600` automatically updates all its alpha variants. No hardcoded RGB values, no duplicate colour definitions. Browser support: Chrome 119+, Safari 16.4+, Firefox 128+ — fine for a developer tool.

**Colour tokens (40):** Surface, text, border, and icon categories with light/dark variants. Primitive palette (`_primitives.scss`) provides 7 colour families (slate, purple, red, green, gold, blue, orange) with 50–950+ tonal scales.

**Non-colour tokens (18):**
- Border radius — 8 values (none through full)
- Elevation — 4 shadow levels with dark mode overrides (OKLCH references)
- Motion — 3 durations + 3 easing curves

**Token utility classes** (`_token-utilities.scss`): Auto-generated CSS classes consuming tokens — `bg-surface-*`, `text-*`, `border-*`, `icon-*`, `rounded-*`, `shadow-*`, `transition-*`.

### 2. Storybook + Chromatic

- Storybook with webpack5, SWC compiler, a11y addon
- Chromatic CI workflow captures visual regression on every PR
- **Light + dark mode snapshots** for every component story via Chromatic modes
- Storybook MCP connected — AI tools can query full component and token catalogue
- MCP story shows primitive references (`var(--purple-600)`) not resolved hex values

### 3. Component stories

| Component | Coverage |
|-----------|---------|
| Banner | 4 variants, with/without CTA |
| Button | 7 themes, 4 sizes, disabled, icons |
| Switch | On, off, disabled, controlled |
| Skeleton | Text, badge, circle variants |
| SettingRow | Pattern component with React 19 useId() |
| Icons | Categorised searchable catalogue (61 icons) |

### 4. New components

- **Banner** — variant-driven alert using semantic colour tokens
- **SettingRow** — accessible settings toggle with ARIA labels
- **Skeleton** — loading placeholder with reduced motion support
- **RequireFeatureOwnershipSetting** — project setting (#4432)

### 5. Documentation (MDX + token stories)

- **Palette** — primitive colour scales (auto-generated)
- **Semantic Colour Tokens** — runtime CSS var reading with theme toggle (surface, text, border, icon categories)
- **Tag & Project Colours** — categorical swatches
- **Spacing, Radius, Elevation, Motion** — visual previews importing from tokens.ts
- **Token Reference (MCP)** — auto-generated flat JSX with primitive references for AI
- **Decision Framework** — where does a new colour go?
- **Token Maintenance** — codegen workflow for adding/modifying/removing tokens

### 6. Dogfooding

`documentation/docs.scss` uses token references across all categories (spacing, colour, font, radius, motion).

- `npm run storybook` — all stories render correctly in light and dark mode
- `npm run build-storybook` — builds without errors
- `npm run generate:tokens` — codegen produces correct output, idempotent (running twice gives same result)
- `npm run typecheck` — no new type errors from token files
- `ENV=local npm run dev` — app renders identically (no visual regression)
- Chromatic CI captures light + dark snapshots for component stories
- Storybook MCP returns component docs and token reference via `get-documentation`
- lint-staged: editing tokens.json and committing auto-regenerates all output files
- Verified: zero raw hex values in semantic colour layer, all reference primitives via var() or OKLCH


talissoncosta:
**Blocker**: #6866 (Decouple Button from Redux store) should land before or alongside this PR — Button's dependency on `AccountStore` and Redux state prevents adding Storybook stories for any component that uses `<Button>`.

frontend/common/theme/tokens.ts
```diff
@@ -0,0 +1,163 @@
+// =============================================================================
+// Design Tokens — AUTO-GENERATED from common/theme/tokens.json
```
talissoncosta, line 2: > ⚠️ Auto-generated files — do not edit directly

This file is generated by `scripts/generate-tokens.mjs` from `tokens.json`.

The following files are all generated from the same source:
- `_tokens.scss`
- `_token-utilities.scss`
- `TokenReference.generated.stories.tsx`

They are committed (not gitignored) because CI requires them for Storybook and Chromatic builds. `lint-staged` re-runs codegen on every commit to keep them in sync with `tokens.json`.

**To make changes, edit `tokens.json` — not these files.**

frontend/web/styles/styles.scss
```diff
@@ -1,4 +1,6 @@
 @import "variables";
+@import "tokens";
+@import "token-utilities";
```
talissoncosta, line 3: > 📝 `_tokens.scss` and `_variables.scss` coexist intentionally

The token system is the new foundation. Existing components still reference `_variables.scss` and will migrate incrementally.

**Utility class collisions:** 11 class names overlap with Bootstrap (e.g. `.text-danger`, `.text-warning`). Bootstrap currently wins the cascade, so this PR introduces zero visual change.

frontend/.claude/context/ui-patterns.md
```diff
@@ -1,5 +1,9 @@
 # UI Patterns & Best Practices
 
+## Storybook First
+
+Before building or composing UI, query Storybook MCP (`list-all-documentation`, then `get-documentation`) to discover existing components and their props. Storybook is the source of truth for the design system — don't grep source files to find components when the catalogue is available.
```
Zaimwa9: I would challenge enforcing using the MCP for every components. I'd see it as more optional depending on the complexity.

Is there a way maybe to have a local inventory or a smaller chunk as an initial context grabbing?

frontend/common/theme/tokens.json
```diff
@@ -0,0 +1,160 @@
...
+    "fast": { "cssVar": "--duration-fast", "value": "100ms", "description": "Quick feedback. Hover states, toggle switches, checkbox ticks." },
+    "normal": { "cssVar": "--duration-normal", "value": "200ms", "description": "Standard transitions. Dropdown open, tooltip appear, tab switch." },
+    "slow": { "cssVar": "--duration-slow", "value": "300ms", "description": "Deliberate emphasis. Modal enter, drawer slide, accordion expand." }
+  },
+  "easing": {
+    "standard": { "cssVar": "--easing-standard", "value": "cubic-bezier(0.2, 0, 0.38, 0.9)", "description": "Default for most transitions. Smooth deceleration. Use for elements moving within the page." },
+    "entrance": { "cssVar": "--easing-entrance", "value": "cubic-bezier(0.0, 0, 0.38, 0.9)", "description": "Elements entering the viewport. Decelerates into resting position. Modals, toasts, slide-ins." },
+    "exit": { "cssVar": "--easing-exit", "value": "cubic-bezier(0.2, 0, 1, 0.9)", "description": "Elements leaving the viewport. Accelerates out of view. Closing modals, dismissing toasts." }
+  }
+}
```
Zaimwa9, line 160: Agreed with Kyle' comments on starting with a lean version that only has the essentials (and used) tokens.

frontend/documentation/components/Banner.stories.tsx
```diff
@@ -0,0 +1,179 @@
+import React from 'react'
```
Zaimwa9, line 1: Genuine structure question. What are the usual practices? What about having the story living in the same folder as their component? Is it discoverable then?

frontend/web/styles/_primitives.scss
```diff
@@ -0,0 +1,105 @@
...
+// These are NOT for direct use in components — use semantic tokens instead.
+// See _tokens.scss for the semantic layer.
+//
+// Scale structure matches Tailwind convention (11 steps per hue).
+// Anchored values are locked to existing Flagsmith colours; generated steps
+// were interpolated in HSL to fill the gaps.
+// =============================================================================
+
+// Slate (neutrals) — custom scale, not Tailwind-aligned
+$slate-0:   #ffffff;
```
Zaimwa9, line 14: So we need to maintain both the primitives and tokens? Can you show me an example on how to use each of them please?

Zaimwa9 (commented):
Overall good and quite isolated so looks safe to me. I'm concerned about the skill with the MCP as I'd prefer to not have the LLM overuse it for basic usage.

Then I could use some more examples on how to use the different tokens files (scss, ts, utilities) etc and the process on how to maintain them, add them etc.

My understanding is that this work is mainly the plumbery that unlocks the following issues ?

frontend/.claude/context/ui-patterns.md
talissoncosta: Good catch — that line was left over from a presentation test and slipped through. Removing it now.

Your suggestion makes sense — I'll add a quick-reference component list to `.claude/context/` as a plain markdown file. MCP stays as a recommendation for when you need live Storybook docs or visual examples, not a requirement for every component.

frontend/common/theme/tokens.json
talissoncosta: Agreed — I'll trim the semantic tokens in `tokens.json` to only what's currently referenced in the codebase. The full palette stays in `_primitives.scss` (pure CSS custom properties, zero runtime cost), and we can add new semantic tokens as we need them.

Worth noting: the colour tokens were scoped to actual needs from the start. The non-colour tokens (radius, motion, elevation) are the better candidates for trimming — happy to cut anything that isn't being used yet.

If you have specific tokens you'd flag as premature, I'd appreciate the input.

frontend/documentation/components/Banner.stories.tsx
talissoncosta: Fair question — co-location is the most common pattern. I've kept them separate here for two reasons: cleaner separation between component code and documentation, and easier extraction if we ever move the design system to its own package. I've seen this work well at scale and it's been a deliberate choice rather than an oversight.

That said, I'm not attached to it — both approaches are valid and we can move them later either way. Happy to align on whatever the team prefers.

frontend/web/styles/_primitives.scss
talissoncosta: Primitives are never used directly in components — they're the raw colour definitions that semantic tokens reference. You only ever work with the semantic layer.

**Example:**
```scss
// _primitives.scss — don't use these in components
:root {
  --red-600: #dc2626;
}

// _tokens.scss — use these
:root {
  --color-text-danger: var(--red-600);
}
```
```scss
// ✅ Correct — reference the semantic token
.error-message { color: var(--color-text-danger); }

// ❌ Incorrect — never reference a primitive directly
.error-message { color: var(--red-600); }
```

The primitive layer exists so we can change `--red-600` once and every token referencing it updates automatically. I'll add this example to the token docs in this PR.

frontend/common/theme/tokens.json
talissoncosta: I'd suggest keeping all 41 colour tokens (surface, text, border, icon) — they enforce correct state usage and give the team the right vocabulary to reach for, even if not everything is consumed yet.

For non-colour tokens, I'd trim from 17 to 7:

| Category | Keep | Remove |
|----------|------|--------|
| Radius | `sm`, `md`, `lg`, `full` | `none`, `xs`, `xl`, `2xl` (only 4 values used in SCSS today) |
| Shadow | `sm`, `md`, `lg` | `xl` (no equivalent in codebase) |
| Duration | — | all 3 (add when we have a real animation pattern) |
| Easing | — | all 3 (same reason) |

That takes us from **58 → 48 tokens**. We can always add back as needed.

Does that feel right?

Zaimwa9: Mmmh, based on your table (thanks for bringing it out) I would keep those. It's not as much as I thought and it might prevent from using these middle ground values (e.g radius)

talissoncosta:
@Zaimwa9 
Answering your top-level comment — correct, this is the foundational plumbing that enables the follow-up issues. No existing components are affected.

I’ve updated the MCP guidance to be optional (as you suggested) and added a primitives vs tokens usage guide to the Token Maintenance docs, including examples in SCSS, TypeScript, and utility classes. Let me know if that addresses your questions around usage and maintenance.

Zaimwa9 (approved):
Thanks for the MD file

PR #6899, opened 2026-03-09
fix: Wait for all data sources before rendering identity features

### LINKED ISSUE #6898: Identity override value not loaded when editing from Identity Overrides tab
## Bug

When clicking "Edit" on an identity override (from the Identity Overrides tab in the feature modal), the user is navigated to the identity page with `?flag=` in the URL. The edit modal auto-opens but displays an **empty value** field, even though the identity has an override value set.

If the user saves in this state, the override value is erased.

## Root cause

Race condition in `UserPage.tsx`. The page fires 3 parallel API requests on mount:

1. `GET /projects/{id}/features/?identity=...` — feature list
2. `GET /environments/{id}/edge-identities/{id}/` + `GET .../edge-featurestates/` — identity flags (chained)
3. `GET /environments/{id}/edge-identities/{id}/edge-featurestates/all/` — actual resolved values

The loading guard checked for `actualFlags` but used `&&` logic: `(isLoading || !projectId) && (!identityFlags || !actualFlags || !projectFlags)`. Once the stores finished loading (`isLoading = false`), the left side became `false`, short-circuiting the entire condition — so the `!actualFlags` check on the right side was never evaluated. The loader disappeared even though `actualFlags` was still in flight.

When the loader disappears, the feature list renders and the preselect mechanism (`?flag=`) immediately opens the edit modal. At this point, `overrideFeatureState` is built with `feature_state_value: undefined` because `actualFlags` hasn't loaded yet.

The `hasPreselected` ref in `FeatureOverrideRow` prevents the preselect from firing again when `actualFlags` eventually loads, so the modal stays open with stale data.

## Reproduction

Hard to reproduce consistently because it depends on request #3 being slower than #1 and #2. More likely to occur on accounts with many features/overrides (larger response payload for the `/all/` endpoint).

Can be reliably reproduced by adding an artificial delay to the `getActualFlags` response in `UserPage.tsx`.

## Fix

Change the loading condition from `&&` to `||` logic so the loader stays until all 3 data sources are loaded.

talissoncosta:
## Summary
- Fix race condition where the identity override edit modal opens with an empty value field when `actualFlags` hasn't loaded yet
- Replace `&&` loading guard with `||` logic so the loader stays until all 3 data sources are loaded

Fixes #6898

## Root cause

`UserPage` fires 3 parallel API requests on mount. The loading guard checked for `actualFlags` but used `&&` logic: `(isLoading || !projectId) && (!identityFlags || !actualFlags || !projectFlags)`. Once the stores finished loading (`isLoading = false`), the left side became `false`, short-circuiting the entire condition — the `!actualFlags` check was never evaluated. The loader disappeared while `actualFlags` was still in flight, and the preselect (`?flag=`) auto-opened the modal with `feature_state_value: undefined`.

## Test plan
- [ ] Navigate to an identity page with `?flag={featureName}` where the identity has an override value set
- [ ] Verify the loader stays visible until all data is loaded
- [ ] Verify the edit modal opens with the correct override value
- [ ] Verify saving preserves the value correctly

### DIFF
--- a/frontend/web/components/pages/UserPage.tsx
+++ b/frontend/web/components/pages/UserPage.tsx
@@ -82,14 +82,10 @@ const UserPage: FC = () => {
 
   useEffect(() => {
     const { search, sort } = filter
-    AppActions.searchFeatures(
-      projectId,
-      environmentId,
-      true,
-      search,
-      sort,
-      { ...getServerFilter(filter), identity: id },
-    )
+    AppActions.searchFeatures(projectId, environmentId, true, search, sort, {
+      ...getServerFilter(filter),
+      identity: id,
+    })
   }, [filter, environmentId, projectId, id])
 
   useEffect(() => {
@@ -167,8 +163,9 @@ const UserPage: FC = () => {
           }) => {
             const identityName =
               (identity && identity.identity.identifier) || id
-            return (isLoading || !projectId) &&
-              (!identityFlags || !actualFlags || !projectFlags) ? (
+            const isDataLoaded =
+              !!actualFlags && !!identityFlags && !!projectFlags && !!projectId
+            return isLoading || !isDataLoaded ? (
               <div className='text-center'>
                 <Loader />
               </div>


khvn26 (dismissed):

emyller (approved):
LGTM

PR #6908, opened 2026-03-10
fix(6905): replace dynamic serializer with SDKAnalyticsFlagsV1Serializer

### LINKED ISSUE #6905: TypeError: Field 'evaluation_count' expected a number but got {'abc': {'xyz:bcd': 2}}.
Sentry Issue: [FLAGSMITH-API-5DM](https://flagsmith.sentry.io/issues/7028030315/?referrer=github_integration)

```
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'dict'
  File "django/db/models/fields/__init__.py", line 2053, in get_prep_value
    return int(value)

TypeError: Field 'evaluation_count' expected a number but got {{'abc': {'xyz:bcd': 2}}
(14 additional frame(s) were not displayed)
...
  File "app_analytics/tasks.py", line 104, in track_feature_evaluations_by_environment
    FeatureEvaluationRaw.objects.bulk_create(

Failed to execute task 'tasks.track_feature_evaluations_by_environment', with id 9054576. Exception: Field 'evaluation_count' expected a number but got {'abc': {'xyz:bcd': 2}}.
```

gagantrivedi:

## Changes

Closes https://github.com/Flagsmith/flagsmith/issues/6905, #5906 

The V1 `SDKAnalyticsFlags` view dynamically generated a DRF serializer inside `get_serializer_class`. This caused dotted feature names (e.g. `org.app.module:handler`) to be split into nested dicts by DRF's `set_value`.

This PR replaces the dynamic serializer with a proper `SDKAnalyticsFlagsV1Serializer`.

Unknown feature names were already silently dropped by the old code (only known names had fields). The new serializer preserves this behaviour explicitly in `validate`. Non-integer values are now also silently skipped — the old code would have raised an unhandled `AssertionError` in this case since `is_valid()` was called without `raise_exception=True`, leaving `validated_data` unpopulated. This never surfaced because SDKs always send integers.

Existing and new unit tests:

```
python -m pytest tests/unit/app_analytics/test_unit_app_analytics_views.py -v
```

All 22 tests pass (20 passed, 2 skipped due to no analytics database configured).

### DIFF
--- a/api/app_analytics/serializers.py
+++ b/api/app_analytics/serializers.py
@@ -137,7 +137,7 @@ class SDKAnalyticsFlagsV1Serializer(serializers.Serializer):  # type: ignore[typ
         return {
             name: count
             for name, count in data.items()
-            if isinstance(name, str) and isinstance(count, int)
+            if isinstance(name, str) and type(count) is int
         }
 
     def validate(self, attrs: dict[str, int]) -> dict[str, int]:
--- a/api/app_analytics/serializers.py
+++ b/api/app_analytics/serializers.py
@@ -4,7 +4,9 @@ from common.core.utils import using_database_replica
 from django.conf import settings
 from rest_framework import serializers
 
+from app_analytics.cache import FeatureEvaluationCache
 from app_analytics.constants import LABELS
+from app_analytics.mappers import map_request_to_labels
 from app_analytics.tasks import (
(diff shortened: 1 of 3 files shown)


Zaimwa9 (approved):
Looks good! Indeed, I can get why linking the sentry issue with this code was sneaky

PR #6911, opened 2026-03-10
ci: Publish flagsmith-private-cloud image to ECR on release

germangarces:

## Changes

Contributes to [#314](https://github.com/Flagsmith/infrastructure/issues/314)

Added a reusable GitHub Actions workflow (`.reusable-docker-publish-ecr.yml`) that authenticates to ECR via OIDC role assumption and copies Docker images using buildx imagetools.
The main pipeline (`platform-docker-build-test-publish.yml`) now uses it to publish `flagsmith-private-cloud` to `084060095745.dkr.ecr.eu-west-2.amazonaws.com/private-cloud/flagsmith` on every release.

_Please describe._


.github/workflows/.reusable-docker-publish-ecr.yml
khvn26, line 1: Maybe parametrise `.reusable-docker-publish.yml` and add a `configure-aws-credentials` step there?

.github/workflows/platform-docker-build-test-publish.yml
```diff
@@ -147,6 +147,19 @@ jobs:
...
+  # Publish to ECR
+  docker-publish-ecr-private-cloud:
+    needs: [docker-build-private-cloud, run-e2e-tests]
+    uses: ./.github/workflows/.reusable-docker-publish-ecr.yml
+    if: github.event_name == 'release'
+    with:
+      source-images: ${{ needs.docker-build-private-cloud.outputs.image }}
+      target-images: 084060095745.dkr.ecr.eu-west-2.amazonaws.com/private-cloud/flagsmith
+      aws-region: eu-west-2
+    secrets: inherit
```
khvn26, line 160: Does this mean we'll need publish to other cloud provider registries in the future?

khvn26 (changes requested):

.github/workflows/platform-docker-build-test-publish.yml
germangarces: Yes, we will push to Azure and GCP.

.github/workflows/.reusable-docker-publish-ecr.yml
germangarces: That's a great idea 🙇
Would you prefer to keep `.reusable-docker-publish-ecr.yml` calling `.reusable-docker-publish.yml`
Or drop it entirely and call `.reusable-docker-publish.yml` from the main pipeline with AWS parameters?

germangarces: I've drop it entirely. I don't think it makes sense to have another wrapper
d81b114ec

.github/workflows/platform-docker-build-test-publish.yml
khvn26: Do you think it's possible to arrange the private cloud registries to be pull-through caches with a quay upstream? We won't have to manage a multitude of publishes this way.

Sorry, this probably should've been a comment to https://github.com/Flagsmith/infrastructure/pull/323, but I'd like to use this opportunity to step back and think on the overall solution a little.

.github/workflows/.reusable-docker-publish.yml
```diff
@@ -27,6 +27,16 @@ on:
         description: The name of the secret holding the docker password
         required: false
         default: DOCKERHUB_TOKEN_WRITE
+      aws-role-to-assume:
```
emyller, line 30: The name `role-to-assume` makes sense as an input of the configure-aws-credentials action, but I think it needs a better name in the scope of this CI configuration.

```suggestion
      aws-ecr-publish-role:
```

.github/workflows/platform-docker-build-test-publish.yml
emyller: > Do you think it's possible to arrange the private cloud registries to be pull-through caches with a quay upstream? We won't have to manage a multitude of publishes this way.

Without knowing of a pre-existing solution to this, I could only think of setting up per-provider/region HTTP proxies that point to Quay, with a cache layer on top of it. Do you know of a pattern we could look?

I think with the right design here, ECR _should_ behave like that cache layer? (despite we still use the verb "publish")

.github/workflows/platform-docker-build-test-publish.yml
```diff
@@ -147,6 +147,20 @@
...
+    uses: ./.github/workflows/.reusable-docker-publish.yml
+    if: github.event_name == 'release'
+    with:
+      source-images: ${{ needs.docker-build-private-cloud.outputs.image }}
+      target-images: 084060095745.dkr.ecr.eu-west-2.amazonaws.com/private-cloud/flagsmith
+      aws-role-to-assume: ${{ vars.AWS_PUBLISH_ECR_ROLE_ARN }}
+      aws-region: eu-west-2
+    secrets: inherit
+
+  # Publish to Quay.io
```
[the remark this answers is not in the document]
emyller: ...plus an inline comment on L4 explaining why it's needed.

.github/workflows/platform-docker-build-test-publish.yml
khvn26: AFAIK ECR supports this mode of operation without having to build anything custom: https://docs.aws.amazon.com/AmazonECR/latest/userguide/pull-through-cache-creating-rule.html

Google: https://docs.cloud.google.com/artifact-registry/docs/repositories/remote-repo

Azure: https://learn.microsoft.com/en-us/azure/container-registry/artifact-cache-overview

emyller: Nice! I think this covers the "pre-existing solution to this" nicely, thanks for sharing the knowledge.

It looks like there would be some work needed in infrastructure as well. Would you be happy if we pause work on this PR and coordinate this in a new issue?

khvn26: Sure, no worries — glad to assist.

emyller: Added https://github.com/Flagsmith/infrastructure/issues/328.

PR #6913, opened 2026-03-10
fix: `/api/v1/projects/{project_pk}/segments/` is broken for compressed environments

### LINKED ISSUE #6912: `/api/v1/projects/{project_pk}/segments/` is broken for compressed environments
Sentry Issue: [FLAGSMITH-API-5K0](https://flagsmith.sentry.io/issues/7324234183/?referrer=github_integration)

```
ValidationError: 2 validation errors for EnvironmentModel
project
  Input should be a valid dictionary or instance of ProjectModel [type=model_type, input_value=Binary(b'\x1f\x8b\x08\x00...xb6\xdeY\n\xbf\x00\x00'), input_type=Binary]
    For further information visit https://errors.pydantic.dev/2.12/v/model_type
feature_states
  Input should be a valid list [type=list_type, input_value=Binary(b'\x1f\x8b\x08\x00...f\xa5"\xb4\xa0\x01\x00'), input_type=Binary]
    For further information visit https://errors.pydantic.dev/2.12/v/list_type
(7 additional frame(s) were not displayed)
...
  File "segments/views.py", line 123, in get_queryset
    segment_ids = EdgeIdentity.dynamo_wrapper.get_segment_ids(identity_pk)
  File "environments/dynamodb/wrappers/identity_wrapper.py", line 198, in get_segment_ids
    environment = EnvironmentModel.model_validate(
```

khvn26:

## Changes

Closes https://github.com/Flagsmith/flagsmith/issues/6912

In this PR, we fix the broken `/api/v1/projects/{project_pk}/segments/` endpoint for environments with compressed DynamoDB documents. The root cause was `DynamoIdentityWrapper.get_segment_ids()` reading the full environment document from DynamoDB and attempting to parse gzip-compressed Binary fields (`project`, `feature_states`) with Pydantic's `model_validate`.

The fix switches segment evaluation to use the Django ORM instead of DynamoDB, mirroring the existing non-edge `Identity.get_segments()` path:

- `get_segment_ids()` now looks up the `Environment` via `Environment.objects.get(api_key=...)` and uses `project.get_segments_from_cache()` + `map_segment_to_engine()` for segment evaluation
- Removed `DynamoEnvironmentWrapper` dependency from `DynamoIdentityWrapper` (the `environment_wrapper` constructor parameter is no longer needed)
- Removed `EnvironmentProtocol` from context mappers since all callers now pass Django `Environment` ORM objects
- Removed dead `get_context_segments()` function (no remaining callers)

Ran the updated unit tests.

### DIFF
--- a/api/environments/dynamodb/wrappers/identity_wrapper.py
+++ b/api/environments/dynamodb/wrappers/identity_wrapper.py
@@ -13,15 +13,13 @@ from edge_api.identities.search import EdgeIdentitySearchData
 from environments.dynamodb.constants import IDENTITIES_PAGINATION_LIMIT
 from environments.dynamodb.wrappers.exceptions import CapacityBudgetExceeded
 from util.engine_models.context.mappers import (
-    get_context_segments,
+    is_context_in_segment,
     map_environment_identity_to_context,
 )
-from util.engine_models.environments.models import EnvironmentModel
 from util.engine_models.identities.models import IdentityModel
 from util.mappers import map_identity_to_identity_document
 
 from .base import BaseDynamoWrapper
-from .environment_wrapper import DynamoEnvironmentWrapper
 
 if typing.TYPE_CHECKING:
     from boto3.dynamodb.conditions import ConditionBase
@@ -37,12 +35,8 @@ logger = logging.getLogger()
 
 
 class DynamoIdentityWrapper(BaseDynamoWrapper):
-    def __init__(
-        self,
-        environment_wrapper: DynamoEnvironmentWrapper | None = None,
-    ) -> None:
+    def __init__(self) -> None:
         super().__init__()
-        self._environment_wrapper = environment_wrapper or DynamoEnvironmentWrapper()
 
     def get_table_name(self) -> str | None:  # type: ignore[override]
         return settings.IDENTITIES_TABLE_NAME_DYNAMO
@@ -188,6 +182,9 @@ class DynamoIdentityWrapper(BaseDynamoWrapper):
         identity_pk: str = None,  # type: ignore[assignment]
         identity_model: IdentityModel = None,  # type: ignore[assignment]
     ) -> list:  # type: ignore[type-arg]
+        from environments.models import Environment
+        from util.mappers.engine import map_segment_to_engine
+
         if not (identity_pk or identity_model):
             raise ValueError("Must provide one of identity_pk or identity_model.")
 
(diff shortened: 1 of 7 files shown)


api/environments/dynamodb/wrappers/identity_wrapper.py
```diff
@@ -188,22 +182,29 @@ def get_segment_ids(
...
         with suppress(ObjectDoesNotExist):
             identity = identity_model or IdentityModel.model_validate(
                 self.get_item_from_uuid(identity_pk)
             )
-            environment = EnvironmentModel.model_validate(
-                self._environment_wrapper.get_item(identity.environment_api_key)
+            environment = Environment.objects.select_related("project").get(
+                api_key=identity.environment_api_key,
             )
+            segments = environment.project.get_segments_from_cache()
```
emyller, line 198: Note: I'm curious to the effect of saving a _Queryset_ object to cache. Is that going to actually evaluate the queryset then serialise/pickle and persist any retrieved objects onto cache? Or rather, is it perhaps only pickling the QuerySet object itself?

api/util/engine_models/context/mappers.py
```diff
@@ -40,8 +37,7 @@ def map_environment_identity_to_context(
     Vendored from flagsmith-flag-engine's fix/missing-export branch and adapted
     to return v10's EvaluationContext TypedDict.
 
-    :param environment: Any object with `api_key` and `name` attributes
-        (e.g. Django Environment or Pydantic EnvironmentModel).
+    :param environment: A Django Environment ORM model.
```
emyller:
```suggestion
    :param environment: An Environment object.
```

nit: I understand why we had _Django vs Pydantic_ in the past, but I think adding the words "Django" and "ORM" is overly expressive — maybe AI slop? Also, the parameter requires an object, not the model.

emyller (dismissed):
LGTM with non-blocking comments.

api/environments/dynamodb/wrappers/identity_wrapper.py
khvn26: `get_project_segments_from_cache` stores a QuerySet to Django cache after `prefetch_related` — Django QuerySets are picklable and include evaluated results when serialised. This is the pre-existing pattern used by `Identity.get_segments()` already, not something introduced here.

api/util/engine_models/context/mappers.py
khvn26: Good call — fixed in 99625d12.

emyller (approved):
LGTM!

PR #6923, opened 2026-03-11
fix: add edit users sending up wrong data

kyle-ssg:

## Changes

When adding a group member, the FE was sending up all existing group members. This caused an issue where ``MANAGE_USER_GROUPS`` did not have appropriate permissions to add themselves to a group.

Also, if a change errors the FE was silently swallowing them instead of showing them.

Added and removed members during edit / creation of a group


frontend/common/services/useGroup.ts
```diff
@@ -27,11 +27,14 @@ export const groupService = service
...
               method: 'POST',
-              url: `organisations/${query.orgId}/groups/${data.id}/`,
+              url: `organisations/${query.orgId}/groups/${data.id}/add-users/`,
             })
+            if (res.error) {
+              return { error: res.error }
+            }
           }
           // Make the admins
           await Promise.all(
```
talissoncosta: Should we handle errors in here also, what if `createGroupAdmin` fails in here ?

frontend/common/services/useGroup.ts
```diff
@@ -115,19 +118,25 @@ export const groupService = service
...
               body: { user_ids: query.usersToRemove },
               method: 'POST',
               url: `organisations/${query.orgId}/groups/${data.id}/remove-users/`,
             })
+            if (res.error) {
+              return { error: res.error }
+            }
           }
           // Make the admins
           await Promise.all(
```
talissoncosta: Same here,  should these Promise.all calls handle errors for consistency with the add/remove users fix above?

talissoncosta (dismissed):
Hey @kyle-ssg , thanks for handling this! I've left two inline comments, they're not blockers, but since the goal of this PR is to stop swallowing errors silently, it felt worth flagging for consistency.

kyle-ssg:
> Hey @kyle-ssg , thanks for handling this! I've left two inline comments, they're not blockers, but since the goal of this PR is to stop swallowing errors silently, it felt worth flagging for consistency.

Done, I'd like to get this merged today if possible as it has been causing a real issue.

talissoncosta (approved):
Thanks for handling that @kyle-ssg ! Good job!

PR #6322, opened 2025-11-25
refactor: consolidate types for orgId and projId

talissoncosta:
## Changes

This PR standardizes the type definitions for `organizationId` and `projectId` throughout the frontend codebase, consolidating them from `string | number` to consistently use `number`.

### Motivation
The `RouteContext` (our standard way to access organization and project IDs) already provides these values as `number` types. This refactor aligns all type definitions with this existing pattern, eliminating type inconsistencies and unnecessary type conversions.

### Key Changes
- **Type Consolidation**: Updated ~50+ request type definitions in `frontend/common/types/requests.ts` to consistently use `number` for:
  - `organisationId` / `orgId`
  - `projectId` / `project`
  - Related ID fields (`featureId`, `segmentId`, etc.)
- **Removed String Conversions**: Eliminated unnecessary `String()` conversion calls throughout components and hooks that use these ID types
- **Updated Components**: Modified components throughout the codebase to work with the standardized `number` types
- **Intentional Exception**: `environmentId` remains as `string` type (correct as-is)

### Impact
- ✅ No breaking changes - this is an internal type standardization
- ✅ Improves type safety and consistency
- ✅ Reduces unnecessary type conversions
- ✅ Aligns with RouteContext pattern

## How did you test this code?

- ✅ Ran TypeScript type checking (`npm run typecheck`) - all types validate correctly
- ✅ Manual testing of affected workflows:
  - Organisation settings and navigation
  - Project settings and navigation
  - Features, segments, and other ID-dependent operations
- ✅ Verified no runtime behavior changes - IDs are handled consistently as numbers throughout

---

**Checklist:**
- [ ] Added information to `docs/` if required
- [x] Filled in Changes section
- [x] Filled in How did you test this code section  
- [x] Used a Conventional Commit title

PR #6925, opened 2026-03-11
refactor: create feature migration

kyle-ssg:

## Changes

Migrates all tabs including the create feature modal itself to TypeScript, no type errors to do with creating / editing a feature and also reduces the feature modal by ~1200 LoC.

Migrates a few components to ES export 

- E2E was extended to cover more of the feature management process specifically for this PR
- Created / edited features in production


frontend/web/components/modals/create-feature/tabs/IdentityOverridesTab.tsx
```diff
@@ -0,0 +1,454 @@
...
+          return
+        }
+
+        const envId = project?.environments?.find(
+          (env) => env.api_key === environmentId,
+        )?.id
+        if (!envId) {
+          setUserOverridesNoPermissionState()
+          return
+        }
```
talissoncosta: I think I might have found a bug

This is comparing api_key (a string like "RAvjEqbj96SXiJP4hQHd5E") with environmentId which is now a numeric ID (e.g. "123") after #6322. They'll never match, so it always falls into the no-permission state.                                                                                                                                       

environmentId is already the numeric ID, so this lookup can be removed and environmentId passed directly to getPermission.

I am getting this:

Image

frontend/web/components/modals/create-feature/tabs/IdentityOverridesTab.tsx
```diff
@@ -0,0 +1,454 @@
...
+        getPermission(
+          getStore(),
+          {
+            id: envId,
+            level: 'environment',
+          },
+          { forceRefetch },
+        )
```
talissoncosta:
```suggestion
        getPermission(
          getStore(),
          {
            id: environmentId,
            level: 'environment',
          },
          { forceRefetch },
        )
```


it seems to do the trick

frontend/web/components/modals/create-feature/tabs/SegmentOverridesTab.tsx
```diff
@@ -0,0 +1,291 @@
...
+  segmentOverrides,
+  updateSegments,
+}) => {
+  const [showCreateSegment, setShowCreateSegment] = useState(false)
+  const [enabledSegment, setEnabledSegment] = useState(false)
+
+  const { data: project } = useGetProjectQuery({ id: projectId })
+  const environment = project?.environments?.find(
+    (env) => env.api_key === environmentId,
+  )
```
talissoncosta: Same api_key vs numeric ID issue as above — environment is always undefined, so isVersioned and is4Eyes are always false. This silently bypasses 4-eyes/change requests for segment overrides.

talissoncosta:
```suggestion
  const environment = getEnvironment(environmentId)
```

frontend/web/components/modals/create-feature/tabs/IdentityOverridesTab.tsx
```diff
@@ -0,0 +1,454 @@
...
+            feature_state_value: environmentFlag?.feature_state_value || null,
+          },
+        )
+        .then(() => {
+          setIsLoading(false)
+          setSelectedIdentity(null)
+          userOverridesPage(1)
+        })
+        .catch(() => {
+          setIsLoading(false)
```
talissoncosta: What do you think about adding setUserOverridesErrorState in here as well to avoid silent errors ?


```suggestion
          setIsLoading(false)
          setUserOverridesErrorState()                                                                                                                                           
```

talissoncosta (changes requested):

kyle-ssg:
Good spot RE those @talissoncosta. I think we need to sort this out project-wide, more things are starting to use environment.id vs api_key. We should start a convention that explicitly states environmentAPIKey rather than environmentId in the cases we're using the api_key string.

I'll take a look at tackling this in a followup PR separate to this.

talissoncosta (approved):
Looks good — all feedback addressed. The environment ID bugs in `IdentityOverridesTab` and `SegmentOverridesTab` are fixed.


Thanks so much for the hard work on it @kyle-ssg

talissoncosta:
> We should start a convention that explicitly states environmentAPIKey rather than environmentId in the cases we're using the api_key string.

> I'll take a look at tackling this in a followup PR separate to this.

Totally agree! Thanks for that @kyle-ssg

PR #6927, opened 2026-03-11
refactor: Prevent duplicate features list GETS

kyle-ssg:
…een environments

## Changes

Uses RTK cache for feature list to reduce latency, long term solution will be to migrate fully away from the flux store but in the meantime this is a win.

Left = After (1 request instead of multiple)

image

Also adds loading indication between environment where we treat switching environment + fetching = loading 

image

- Filtering features in production 
- E2E covers a lot

### DIFF
--- a/frontend/common/stores/feature-list-store.ts
+++ b/frontend/common/stores/feature-list-store.ts
@@ -854,7 +854,7 @@ const controller = {
       store.projectId = projectId
       store.environmentId = environmentId
       store.page = page
-      store.filter = filter
+      store.filter = filter || {}
       let filterUrl = ''
       const { feature } = Utils.fromParam()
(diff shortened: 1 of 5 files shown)


talissoncosta:
Hey @kyle-ssg , great work!! 

I found a pre-existing bug while testing this PR (also reproduces on main):

Bug: Typing only whitespace (e.g. a space) in the search or value search filter sends search=%20 to the API, which returns a 400 "This field may not be blank.". On main, this also triggers a hard redirect to /404?entity=environment because the Flux store's catch block treats any error as a permission issue.

I've opened #6934 on top of your branch to fix it — it trims search values at the source in useFeatureFilters so whitespace-only input never triggers a state change or API request.

talissoncosta (approved):

PR #6356, opened 2025-12-02
chore: Create flag migration

kyle-ssg:
Thanks for submitting a PR! Please check the boxes below:

- [ ] I have added information to `docs/` if required so people know about the feature!
- [x] I have filled in the "Changes" section below?
- [x] I have filled in the "How did you test this code" section below?
- [x] I have used a [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) title for this Pull Request

## Changes

- Migrates feature settings, feature value and create feature to typescript components
- Migrates Row / FormGroup to fix imports
- Simplifies interface around the create feature modal, this will eventually mean the RTK/redux store no longer needs to create new objects
   - Editing project/feature related stuff is adjusting the ProjectFlag with a single onChange
   - Editing the remote config, enabled state and MV percentages is adjusting a FeatureState with a single on Chnange
- Minor: add tag button same height as tags
image

## How did you test this code?

E2E covers regression testing on everything this affects

PR #6931, opened 2026-03-12
fix: restore Prevent Flag Defaults UI enforcement in create feature modal

### LINKED ISSUE #6930: Prevent Flag Defaults: UI enforcement lost in create feature modal refactor
## Description

The "Prevent Flag Defaults" project setting no longer hides the value and enabled fields when creating a feature flag. Users can set a value and enabled state during creation, even though the setting description says it "forces the user to create a feature before setting its values per environment".

The backend still works correctly (other environments get `enabled=false` and no value), but the frontend no longer enforces the intended UX of hiding those fields during creation.

## Regression

This was introduced in #6356 (commit 00085cd, 20 Jan 2026). The old `CreateFlag.js` component checked `project.prevent_flag_defaults` to:

1. Hide the value/enabled fields during feature creation
2. Show an info message: "This will create the feature for **all environments**, you can edit this feature per environment once the feature is created."

During the refactor to the new `create-feature/tabs/CreateFeature.tsx`, the `hideValue` prop was hardcoded to `false` (line 140) and the `prevent_flag_defaults` check was dropped.

## Expected behaviour

When "Prevent Flag Defaults" is enabled on a project, the create feature form should not show value or enabled state fields - matching the previous behaviour and the setting description.

## Steps to reproduce

1. Enable "Prevent Flag Defaults" in Project Settings > Additional Settings
2. Create a new feature flag
3. Observe that value and enabled fields are still visible and editable

talissoncosta:

## Changes

Closes #6930

The `prevent_flag_defaults` project setting was no longer hiding the value and enabled fields when creating a feature flag. This regression was introduced in #6356 where `hideValue` was hardcoded to `false` and the `prevent_flag_defaults` check was dropped.

### What changed

- **`CreateFeature.tsx`**: Fetches `prevent_flag_defaults` via RTK Query (`useGetProjectQuery`) and conditionally hides the enabled toggle and value editor, showing an info message instead.
- **`FeatureUpdateSummary.tsx`**: Also fetches via RTK Query to conditionally hide the duplicate info message when `prevent_flag_defaults` is active.
- **`index.js`**: Passes `projectId` to `FeatureUpdateSummary` (no longer uses stale Flux `ProjectStore.model`).

### Why RTK Query instead of Flux

The previous `ProjectStore.model` data was stale — toggling the setting in Project Settings didn't update the create feature modal without a full page reload. Using `useGetProjectQuery` ensures the data is always fresh via RTK Query's cache invalidation.

1. Enable "Prevent Flag Defaults" in Project Settings > Additional Settings
2. Open the "New Feature" modal — verify the enabled toggle and value editor are **hidden**, replaced by an info message
3. Disable "Prevent Flag Defaults" in Project Settings
4. Open the "New Feature" modal again — verify the enabled toggle and value editor are **visible**, with the standard info message at the bottom
5. Edit an existing feature — verify the Value tab always shows enabled toggle and value editor regardless of the setting

### DIFF
--- a/frontend/web/components/modals/create-feature/FeatureUpdateSummary.tsx
+++ b/frontend/web/components/modals/create-feature/FeatureUpdateSummary.tsx
@@ -2,9 +2,11 @@ import React, { FC } from 'react'
 import InfoMessage from 'components/InfoMessage'
 import Button from 'components/base/forms/Button'
 import ModalHR from 'components/modals/ModalHR'
+import { useGetProjectQuery } from 'common/services/useProject'
 
 type FeatureUpdateSummaryProps = {
   identity?: string
+  projectId: number | string
   onCreateFeature: () => void
   isSaving: boolean
   name: string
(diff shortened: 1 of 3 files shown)


kyle-ssg:
Thanks for addressing, tested as working

kyle-ssg (approved):

PR #6764, opened 2026-02-24
deps: react-migration-to-v19

Zaimwa9:
Thanks for submitting a PR! Please check the boxes below:

- [ ] I have read the [Contributing Guide](/Flagsmith/flagsmith/blob/main/CONTRIBUTING.md).
- [ ] I have added information to `docs/` if required so people know about the feature.
- [ ] I have filled in the "Changes" section below.
- [ ] I have filled in the "How did you test this code" section below.

Changes

  Upgrades the frontend from React 16 to React 19, along with all incompatible dependencies.

  **React 19 migration**
  - Updated react and react-dom from 16.14.0 to 19.x
  - Updated @types/react and @types/react-dom to 19.x
  - Configured Babel with "runtime": "automatic" for the new JSX transform
  - Removed @babel/transform-react-inline-elements (incompatible with React 19)
  - Updated tsconfig.json and tsconfig.jest.json to use "jsx": "react-jsx"
  - Deleted legacy common/Component.js (used React.createClass, removed in React 19)
  - Removed string refs (ref='wrappedComponent') from ConfigProvider and withSegmentOverrides

  **Dependency replacements**
  - @material-ui/core → native implementations: Material UI v4 is incompatible with React 19
    - AccordionCard: replaced <Collapse> and <IconButton> with a custom height-animated accordion using useRef/useEffect
    - ChipInput: replaced material-ui-chip-input with a custom chip input component
    - SegmentSelect: removed @material-ui/core/Select type import, replaced with plain types
  - react-sortable-hoc → @dnd-kit/core + @dnd-kit/sortable: react-sortable-hoc relies on deprecated findDOMNode, incompatible with
  React 19. Migrated SegmentOverrides to use @dnd-kit with DndContext, SortableContext, and useSortable
  - react-select 2.x → 5.x: Updated for React 19 compatibility
  - react-click-outside: Removed (unused/incompatible)
  - react-window, react-window-infinite-loader, react-virtualized-auto-sizer: Removed (unused)
  - reactstrap: Updated from 9.0.1 to 9.2.3

  **react-tooltip v4 → v5 (performance fix)**
  - Upgraded react-tooltip from 4.5.1 to 5.30.0 to fix severe UI lag when horizontally resizing the browser on pages with many
  feature rows
  - v4 created a separate DOM element per tooltip instance, each registering global resize and scroll event listeners — hundreds of
  listeners firing synchronously per resize frame
  - v5 uses floating-ui: tooltips are only materialised on hover, eliminating idle listeners
[...]

PR #6940, opened 2026-03-13
fix: address review comments on React 19 migration

talissoncosta:

## Changes

Contributes to #6764

Addresses all review comments on the React 19 migration PR. Built on top of Wadii's `deps/upgrade-to-react-19` branch.

**Drag-and-drop (SegmentOverrides):**
- Upgraded from `@dnd-kit/core` + `@dnd-kit/sortable` to `@dnd-kit/react` (newer API with cleaner handle support)
- Card header is the drag handle — inputs, buttons, and dropdowns remain interactive
- Vertical axis locked via `RestrictToVerticalAxis` modifier
- Removed unused `ConfigProvider` HOC wrapping
- Replaced inline styles with CSS classes, used `classNames()` for className concatenation
- Fixed unstable sortable IDs (removed array index from ID, added `getSegmentId` helper)

**Tooltip (react-tooltip v5):**
- Restored 500ms show delay (`delayShow`)
- Fixed font-size via doubled CSS selector specificity (no `!important`)
- Set `--rt-opacity: 1` and `--rt-transition-show-delay: 0.2s`

**ChipInput:**
- Close icon now inline with chip text (`display: flex; align-items: center`)
- Restored backspace-to-delete behaviour

**Shared hook:**
- Extracted `useCollapsibleHeight` hook from `AccordionCard` and `CollapsibleText`
- Fixes missing rAF cleanup in `CollapsibleText`

Manually tested on staging:
- Segment overrides: drag-and-drop reordering via header, verified inputs/toggles/actions menu remain interactive
- Tooltips: verified show delay, font-size, and fade-in transition
- ChipInput: verified chip layout, close icon alignment, backspace deletion
- AccordionCard: verified expand/collapse animation on environment metrics summary


frontend/web/components/SegmentOverrides.js
```diff
@@ -396,10 +387,11 @@ const SegmentOverrideListInner = ({
   return (
     <div>
       {items.map((value, index) => {
-        const sortId = `segment-${value.segment.id}-${index}`
+        const segmentId = getSegmentId(value.segment)
+        const sortId = `segment-${segmentId ?? index}`
         if (isSortable) {
           return (
-            <Fragment key={value.segment.name || sortId}>
+            <Fragment key={sortId}>
```
Zaimwa9, line 394: From the above we have 2 different `isSortable`; can we maybe rename this const into `canSort` ?

frontend/package.json
```diff
@@ -52,9 +52,9 @@
     "@babel/register": "^7.12.1",
     "@datadog/ui-extensions-react": "0.32.0",
     "@datadog/ui-extensions-sdk": "0.32.0",
-    "@dnd-kit/core": "^6.3.1",
-    "@dnd-kit/sortable": "^10.0.0",
-    "@dnd-kit/utilities": "^3.2.2",
+    "@dnd-kit/abstract": "^0.3.2",
+    "@dnd-kit/helpers": "^0.3.2",
+    "@dnd-kit/react": "^0.3.2",
```
Zaimwa9, line 57: Sorry, I feel annoying pointing this out. Is it related to this [comment](https://github.com/Flagsmith/flagsmith/pull/6764#discussion_r2916489954) ?

I recall that the functionality was still working. And we didn't have a handler before. I don't mind at all going with the handler but i'd feel more confortable keeping the  `"@dnd-kit/core": "^6.3.1"` (4.4M downloads a week) instead of using the `0.x` version (45k downloads a week)

Could we explore another approach using the core one or leave it for when we implement a UI drag handler ?

Zaimwa9 (changes requested):
One minor change (the naming) and opening a discussion on dnd-kit versions

frontend/package.json
talissoncosta: I see your concern around the download numbers and the `0.x` version — totally fair point.

**`@dnd-kit/core` v6** is framework-agnostic and was the source of the issues we were hitting — it required a custom `InteractiveAwarePointerSensor`, manual CSS transforms, and two separate providers (`DndContext` + `SortableContext`).

**`@dnd-kit/react`** is the React-specific rewrite by the same author (clauderic). The `0.x` version looks concerning, but it's the next generation of the library, not an unstable experiment — the lower download count is simply because it's newer. The upgrade removed ~45 lines of workaround code and gives us native `handleRef` support for drag handles, which is exactly what we need for segment overrides (only the header should be draggable, not the form controls inside).

Re the [linked comment](#) — the functionality did work without a handle, but it meant the entire row was draggable, which conflicted with interactive elements inside it (inputs, buttons, etc.). That's why we had the custom sensor in the first place. With `@dnd-kit/react`, `handleRef` solves this cleanly without workarounds.

That said, this is just a suggestion — if you'd rather not take a dependency on a newer package, I'm happy to revert it.

frontend/web/components/SegmentOverrides.js
talissoncosta: Thanks for flagging that. fixed in: fec1bf25d32cdfa67443d97f2222624468cddc75

frontend/package.json
Zaimwa9: Resolving as acceptable risk as discussed offline

Zaimwa9 (approved):
👍

Issue #6935, opened 2026-03-12
Fix "unsaved changes" badge positioning on Edit Feature modal tabs


talissoncosta:
## Description

The "unsaved changes" badge (purple dot with `*`) on the Edit Feature modal's "Segment Overrides" tab has positioning issues:

1. It overflows below the tab underline instead of aligning vertically with the tab label
2. It remains visible even when switching to a different tab (e.g. "Identity Overrides"), appearing under an inactive tab

## Screenshots

<img width="644" height="320" alt="Image" src="https://github.com/user-attachments/assets/ae6f30c3-5b9d-4f88-bc42-3e92e0b7d993" />

## Steps to reproduce

1. Open any feature (Edit Feature modal)
2. Click on the "Segment Overrides" tab
3. Make a change to a segment override (toggle enabled, change value, etc.) — the purple `*` badge appears
4. Observe the badge sits below the tab underline
5. Switch to "Identity Overrides" tab — the badge still shows under the now-inactive "Segment Overrides" tab

## Expected behaviour

- The badge should align vertically centred with the tab label text
- The badge positioning should not overflow below the tab underline

<img width="638" height="341" alt="Image" src="https://github.com/user-attachments/assets/ce2bec7e-6591-44ef-89b9-6625799915ce" />

## Relevant files

- **Badge rendered here**: `web/components/modals/create-feature/index.js` (lines 934-937) — the `.unread` div inside the tab label
- **Tab button styling**: `web/styles/components/_tabs.scss`
- **Tab component**: `web/components/navigation/TabMenu/Tabs.tsx`
- **Tab button component**: `web/components/navigation/TabMenu/TabButton.tsx`
- **Base badge styling**: `web/styles/project/_utils.scss` (lines 155-171) — `.unread` class (global, used across the app)

## How to test

1. Run `ENV=local npm run dev`
2. Open a feature with segment overrides
3. Toggle or edit a segment override to trigger the unsaved state
4. Verify the badge aligns with the tab text and doesn't overflow below the underline
5. Check other tab bars across the app (permissions, settings) for visual regressions

JiwaniZakir:
Looking into this, will submit a fix if I can confirm the issue.

PR #6949, opened 2026-03-13
fix: Alignment issues with 'unsaved changes' badge in Feature modal

alveyworld:

## Changes

Closes issue #6935

In the .unread selector, simply changing the bottom attribute to a positive 9px brought the * badge up to vertical center without moving the position or size of the tab. 

Steps to reproduce
1. Open any feature (Edit Feature modal)
2. Click on the "Segment Overrides" tab
3. Make a change to a segment override (toggle enabled, change value, etc.) — the purple * badge appears
4. Observe the badge is now vertically centered and not overlapping the baseline of the tabs
5. Switch to "Identity Overrides" tab — the badge still shows under the now-inactive "Segment Overrides" tab

I followed these steps to recreate and visually fixed the css issue. I resized the window to extreme small and large sized and the badge remained in place. At smaller sizes it naturally jumps up to the right of the tab.

Before
Screenshot 2026-03-17 at 8 28 04 AM
After
Screenshot 2026-03-17 at 8 28 49 AM


talissoncosta:
Hey @alveyworld , thanks for picking this up! 🙏

Could you add before/after screenshots to the PR description? It'll help us validate the fix visually and speed up the review.

Thanks again for the contribution!

alveyworld:
I added the screen shots. I hope this is what you were wanting

kyle-ssg:
Thanks for this @alveyworld! I think actually you might have spotted a regression whereby the real fix is to add a flex-wrap no wrap. Happy to merge this if we get that done!

image

talissoncosta:
Hey @alveyworld, thanks for the screenshots and for sticking with this!

Building on what Kyle mentioned — `.unread` in `_utils.scss` is a global class used across the app (navbar badges, table filters, diff counters), so the `bottom: 9px` change there would affect those places too.

The root cause of the badge misalignment is actually `.tab-nav-full` in `_tabs.scss` — it applies `flex: 1` to each tab button, stretching them and causing the badge to overflow below the underline.

Here's a nudge on the approach:
- `web/styles/components/_tabs.scss` — remove the `.tab-nav-full` rule
- `web/components/navigation/TabMenu/Tabs.tsx` — remove `canGrow` and the `tab-nav-full` className, then add `justify-content-evenly` to the inner container's className (line 141) to keep tabs distributed evenly
- `web/components/navigation/TabMenu/TabButton.tsx` — clean up the now-unused `className` prop
- `web/styles/project/_utils.scss` — revert `bottom` back to `-1px`

Here's what the expected result should look like:

image

This way tabs still spread out nicely, but buttons keep their natural width so the badge stays aligned. Happy to answer any questions!

talissoncosta:
@kyle-ssg `.tabs-nav` already has `flex-wrap: nowrap` (line 70 of `_tabs.scss`) and buttons have `flex: 0 0 auto` (line 77). The issue is `.tab-nav-full` overriding that with `flex: 1 !important` — that's what stretches the buttons and causes both the badge misalignment and the wrapping in your screenshot.

alveyworld:
Thanks for your help @talissoncosta 
I made your suggested changes and got the desired results from the screen shot with the badges inline with the tab instead of underneath. Cursor is reporting some deadcode on this PR so I will need to clean that up.

alveyworld:
Also, it looks like it did not fix similar tabs in areas like the group editor

talissoncosta:
Hey @alveyworld, good progress — the core fix is correct and addresses the reported issue!

A few things to tidy up before this is merge-ready:

**1. Remove `canGrow`** — line 107 in `Tabs.tsx` computes it but nothing uses it anymore. Delete the whole line.

**2. Clean up `TabButton.tsx`** — since no caller passes `className` anymore, remove it from:
- The `TabButtonProps` interface
- The destructured props
- The template literal in the `<Button>` className

Then in `Tabs.tsx`, remove `className={''}` from the `<TabButton>` call entirely (instead of passing an empty string).

**3. Formatting nit** — add a space after the comma:
`'justify-content-evenly', 'd-flex align-items-center flex-1'`

Re: the tabs you mentioned in other areas (like the group editor) — could you share a screenshot showing the specific issue? Those might be pill-style tabs which use a different layout path and wouldn't be affected by this change.

talissoncosta (changes requested):
.

alveyworld:
Here is the group tab I mentioned
Screenshot 2026-03-24 at 10 36 55 AM

alveyworld:
That was fun, thanks for your encouragement. Looks like a useful product

From: Talisson ***@***.***>
Date: Tuesday, March 24, 2026 at 1:07 PM
To: Flagsmith/flagsmith ***@***.***>
Cc: Zack Alvey ***@***.***>, Mention ***@***.***>
Subject: Re: [Flagsmith/flagsmith] fix: Alignment issues with 'unsaved changes' badge in Feature modal (PR #6949)

[https://avatars.githubusercontent.com/u/2528597?s=20&v=4]talissoncosta left a comment (Flagsmith/flagsmith#6949)<https://github.com/Flagsmith/flagsmith/pull/6949#issuecomment-4120715917>

Hey @alveyworld<https://github.com/alveyworld>, thanks again for the work on this and for flagging the group editor issue — that was a great catch and it led us to rethink the approach.

We've decided to tackle this as part of a larger refactor of the TabItem component. The root cause goes beyond CSS — the tab label and badge layout was being handled inconsistently across call sites (some using <Row>, others <div>, each with different spacing classes). We're adding a hasChanges prop to TabItem so the component owns the badge rendering internally, which fixes the alignment everywhere.

We're going to close this PR in favour of that refactor, but your contribution was valuable — it helped us identify the real scope of the problem. Thanks for sticking with it! 🙏

talissoncosta:
Hey @alveyworld, sorry for the earlier close — that was premature.

Thanks for flagging the group editor issue and for the initial investigation — that was a great catch and helped us identify the full scope of the fix.

We've opened #7038 which builds on your work with a slightly different approach (replacing the `<Row>`/`<div>` wrappers in tab labels instead of removing `tab-nav-full`). Closing this in favour of that one, but your contribution is credited there.

Thanks for sticking with it!

PR #6957, opened 2026-03-16
ci: Ensure workflows run on update flagsmith environment PRs

matthewelwell:
## Changes

Ensures that workflows run on auto-generated PRs for updating flagsmith environment document. In order to do this, we have to define a custom GH token as per documentation [here](https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs).

Added a `push` trigger to the current branch to verify that the workflow triggers correctly. [This PR](https://github.com/Flagsmith/flagsmith/pull/6959) was created where you can see the correct workflows are running.

### DIFF
--- a/.github/workflows/update-flagsmith-environment.yml
+++ b/.github/workflows/update-flagsmith-environment.yml
@@ -36,10 +36,9 @@ jobs:
         run: poetry run python manage.py updateflagsmithenvironment
 
       - name: Create Pull Request
-        uses: peter-evans/create-pull-request@v5
+        uses: peter-evans/create-pull-request@v8
         with:
           commit-message: Update API Flagsmith Defaults
-          branch: chore/update-api-flagsmith-environment
           delete-branch: true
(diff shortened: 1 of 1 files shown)


.github/workflows/update-flagsmith-environment.yml
```diff
@@ -4,6 +4,9 @@ on:
   schedule:
     - cron: 0 8 * * *
   workflow_dispatch:
+  push:
+    branches:
+      - ci/run-workflows-on-update-flagsmith-environment
```
talissoncosta: this should be removed before merge, right ?

matthewelwell: Yes, good catch - thanks. Resolved in https://github.com/Flagsmith/flagsmith/pull/6957/commits/b9ebe41284a7fc6539e0b99c310acecf066f2b7d

talissoncosta (approved):

Issue #6945, opened 2026-03-13
Disable code references stats in the features endpoint


emyller:
It's taking quite a toll on database time, and making demos less pleasant.

<img width="544" height="150" alt="Image" src="https://github.com/user-attachments/assets/e6d771db-7570-4e56-b192-a3374c85ee18" />

This functionality should be restored after https://github.com/Flagsmith/flagsmith/issues/6832 is resolved with a solution that makes code references load milliseconds-cheap in the database.

## Acceptance criteria

- Gate code references stats annotation in `GET /api/v1/projects/<project_pk>/features/` behind a Flagsmith-on-Flagsmith feature flag
- When the flag is off, return `null` / `0` for the stats fields — do not change the response schema
- When the flag is on, keep current behaviour
- Default the flag to off

PR #6970, opened 2026-03-17
feat(Features API): Restrict code references querying to a feature flag

emyller:

## Changes

Contributes to https://github.com/Flagsmith/flagsmith/issues/6945

- Don't query code references unless `code_references_ui_stats` is enabled.

> [!IMPORTANT]
> - [x] Needs to review and merge https://github.com/Flagsmith/flagsmith/pull/6968 — order is not critical.

- [x] Added / updated unit tests to `is_feature_enabled`.


kyle-ssg:
Im a bit confused why this has any FE changes, but either way this now shows the following for a project that doesn't have code references integrated.

image

frontend/web/components/tags/CodeReferencesTag.tsx
```diff
@@ -0,0 +1,45 @@
...
+}) => {
+  const hasScannedCodeReferences =
+    projectFlag?.code_references_counts?.length > 0
+  const codeReferencesCounts =
+    projectFlag?.code_references_counts?.reduce(
+      (acc, curr) => acc + curr.count,
+      0,
+    ) || 0
+
+  return (
```
talissoncosta: Just checking...
Is the intent to show the GitHub icon on the list even when there are no code references? Previously the tag only rendered when `hasScannedCodeReferences` was `true`, but now it shows unconditionally (bare icon with "See code references" tooltip on every feature). 

Image

frontend/web/components/tags/VCSProviderTag.tsx
```diff
@@ -49,7 +49,7 @@ const VCSProviderTag: React.FC<VCSProviderTagProps> = ({
         }}
       >
         {providerIcon}
-        <span>{count}</span>
+        {count !== undefined && <span>{count}</span>}
```
talissoncosta: Nit: When count is `undefined`, the tag renders as just the icon with no number. Is that the intended look, or should the whole tag be hidden when there's no count?

frontend/web/components/tags/CodeReferencesTag.tsx
```diff
@@ -0,0 +1,45 @@
...
+    <Tooltip
+      title={
+        <div
```
talissoncosta: **Nit:** Using `onClick` on a `<div>` isn't ideal for accessibility — it's not focusable or keyboard-navigable. Consider using a `<button>` with reset styles instead:
```tsx
onClick ? (
  <button
    type='button'
    onClick={onClick}
    style={{ all: 'unset', cursor: 'pointer' }}
  >
    <VCSProviderTag
      count={hasScannedCodeReferences ? codeReferencesCounts : undefined}
      isWarning={hasScannedCodeReferences && codeReferencesCounts === 0}
      vcsProvider={VCSProvider.GITHUB}
    />
  </button>
) : (
  <VCSProviderTag
    count={hasScannedCodeReferences ? codeReferencesCounts : undefined}
    isWarning={hasScannedCodeReferences && codeReferencesCounts === 0}
    vcsProvider={VCSProvider.GITHUB}
  />
)
```

talissoncosta (changes requested):

frontend/web/components/tags/CodeReferencesTag.tsx
emyller: Yup this _was_ intentional: we wanted to keep the icon there as an indication that _there are code references for this feature_, regardless of how many. We thought the performance degradation would only occur when **counting references**, but it would also happen with an `exists` query given the current data model.

All frontend changes are reverted as of 3f51effa2e844e18f1c0f943cf432c87aab4e613.

Also addresses @kyle-ssg's [review](https://github.com/Flagsmith/flagsmith/pull/6970#issuecomment-4082481719). Sorry there was some noise in the communication here. This is simpler now.

frontend/web/components/tags/VCSProviderTag.tsx
emyller: See [this comment](https://github.com/Flagsmith/flagsmith/pull/6970#discussion_r2957118890).

frontend/web/components/tags/CodeReferencesTag.tsx
emyller: Thanks! This hurts because I used to teach about web semantics and accessibility years ago before modern frontend clouded my comprehension of things. 😅 

Nevertheless, the changes here were reverted in 3f51effa2e844e18f1c0f943cf432c87aab4e613.

talissoncosta (dismissed):

talissoncosta (dismissed):

api/tests/unit/features/test_unit_features_views.py
```diff
@@ -3581,15 +3581,16 @@ def test_list_features_with_filter_by_search_value_boolean(  # noqa: FT003
...
     feature: Feature,
+    mocker: MockerFixture,
+    project: Project,
+    staff_client: APIClient,
     with_project_permissions: WithProjectPermissionsCallable,
-    environment: Environment,
 ) -> None:
     # Given
     with_project_permissions([VIEW_PROJECT])  # type: ignore[call-arg]
+    mocker.patch.object(views, "is__code_references_ui_stats__enabled", return_value=True)  # fmt: skip
```
khvn26: Out of curiosity, why didn't the `enable_features` fixture work out?

api/projects/code_references/services.py
```diff
@@ -27,29 +27,32 @@
...
+    """Check if code reference UI stats are enabled for an organisation.
+
+    TODO: Delete this after https://github.com/flagsmith/flagsmith/issues/6832 is resolved.
+    """
+    flagsmith_client = get_client("local", local_eval=True)
+    flags = flagsmith_client.get_identity_flags(
+        organisation.flagsmith_identifier,
+        traits=organisation.flagsmith_on_flagsmith_api_traits,
+    )
+    return flags.is_feature_enabled("code_references_ui_stats")
```
khvn26: Note: The irony of indirection added to usage of a flag related to code references is not lost on me.

emyller: ashamed dog

As much as I feel tempted to change the approach here so it's detectable by our Code References scan, I actually feel inclined to let it slide and improve the scan algorithm later — while using this as a test case. See https://github.com/Flagsmith/ci/issues/9.

khvn26: But why use the wrapper in the fist place?

emyller: Merely to abstract the SDK boilerplate from business logic. I'd rather use a single `is_feature_enabled` one-liner check (like attempted in 8cdf16bb66f7ac42ac89706fe70427781f329ce3) but one doesn't exist yet.

Do you have a strong opinion otherwise?

api/tests/unit/features/test_unit_features_views.py
emyller: Only because it wasn't known. Fixed in 52e42b2d529f7108dbe1dc5109c7be0d05ed9d93, thanks!

api/projects/code_references/services.py
khvn26: I don't see the value of abstracting out feature flag retrieval beyond the SDK, especially here, when the code is not meant to be duplicated anywhere else. To me, doing this in Flagsmith's own codebase feels off, like there's a deficiency to our SDKs that we want to hide away. If the SDK feels bulky or awkward to use to you, I'd like to dig into that and see how we can improve the DX and make it more ergonomic.

Using this as a reference for https://github.com/Flagsmith/ci/issues/9 doesn't sound convincing to me either given this is documented as short-lived disposable code.

khvn26: For the same reason, though, I don't mind leaving it in.

emyller: I don't think this was motivated by any SDK deficiency, but I do believe our organisation-centric [use case boilerplate](https://github.com/Flagsmith/flagsmith/blob/5c513dd76bf6b0fa30c9a59ad81ba401556d32bd/api/README.md?plain=1#L72-L81) warrants a shameless utility to Flagsmith's own codebase, and our own DX. That said, 5c513dd76bf6b0fa30c9a59ad81ba401556d32bd.

khvn26 (dismissed):
LGTM 👍

api/projects/code_references/services.py
khvn26: Yeah, unfortunately there is already history of custom wrapping code around the SDK in the codebase. I don't agree with the approach and believe that dogfooding should drive SDK improvements, not internal workarounds.

Even when we introduced the `enable_features` fixture mentioned in the other thread, we acknowledged that as tech debt, and created https://github.com/Flagsmith/flagsmith-python-client/issues/135.

khvn26 (approved):

PR #6974, opened 2026-03-17
feat: add Code References to project integrations page

### LINKED ISSUE #6915: Add Code References to the projects Integrations page

Currently, the Code References feature is only visible by finding out about it via our documentation.

To increase visibility, we want to list this feature on the Integrations page of every project, linking to the documentation page.

.

### Additional context

_No response_

talissoncosta:

## Changes

Closes #6915

Adds "Code References" as an external integration on the project Integrations page, following the same pattern as Backstage (external link to docs, no in-app configuration).

- **`common/stores/default-flags.ts`** — new `code-references` entry with `external: true`
- **`web/static/images/integrations/code-references.svg`** — code brackets icon

### Deployment note

The integration list is controlled by the `integration_data` remote config flag in Flagsmith. The `default-flags.ts` change covers self-hosted instances, but for SaaS environments the flag value must also be updated:

- **Staging** — already added
- **Production** — needs `code-references` entry added to the `integration_data` remote config value before merging

## Screenshots

image

1. Navigate to a project's **Integrations** page
2. Verify "Code References" card appears with the correct icon, description, and "Add Integration" button
3. Click "Add Integration" — verify it opens https://docs.flagsmith.com/managing-flags/code-references in a new tab
4. Verify no regressions on other integration cards


emyller (changes requested):
Cool icon! Please adjust it catering for both light and dark themes. e.g. This is how our current GitHub icon looks:

Image

frontend/common/stores/default-flags.ts
```diff
@@ -28,6 +28,15 @@ const defaultFlags = {
       'perEnvironment': false,
       'title': 'Backstage',
     },
+    'code-references': {
+      'description':
+        'Integrate with Code References to track feature flag usage in your codebase, and unlock new functionality in Flagsmith',
```
emyller:
```suggestion
        'Integrate with Code References to track feature flag usage in your codebase, and unlock new functionality in Flagsmith.',
```

Missing period.

kyle-ssg:
It doesn't feel like this reduces any friction in getting started with code references, I say this for 2 reasons:

1 - We don’t surface this where the feature actually lives. That’s a problem—users currently have to scroll through the integrations list to discover it. For features that require an integration but aren’t yet enabled, we should show an “Add integration” prompt directly in the feature’s location.

image


2 - A key prerequisite for code references is linking GitHub, so that needs to be the first step in the user journey.

Right now, a new user’s flow looks like:

- Look through a list of integrations and find the code references integration
- Go to docs
- Find out that it requires GitHub and go back to the app
- Install the GitHub integration
- Go back to the docs and go through the add workflow steps
- Install Code References on GitHub

emyller:
> 2 - A key prerequisite for code references is linking GitHub, so that needs to be the first step in the user journey.

Except for this point, all of @kyle-ssg thoughts resonate with mine. Code References was built to be provider-agnostic, and does not depend on the GitHub integration.

> [!NOTE]
> Our hackathon project _Cleanup_ combines Code References to the GitHub integration by creating a prompt containing code references and filing a GitHub issue.

In order to set up Code References, as described in the [documentation](https://docs.flagsmith.com/managing-flags/code-references), the **only** user action necessary _as of now_ is to set up a GHA workflow in each repository. In the future, I'd love to improve the UX by allowing to install Code References organisation wide, instead of per-repository, via button clicks or IaC instead of GHA scripting.

While in the present, the question remains: where in our UI can we encourage users to go through the necessary GHA setup and try Code References, and how can it be both friendly and effective?

kyle-ssg:
Ah I didn't realise that, my mistake. Ok I think as for now both of the following points are achievable and provide a less tedious flow to get started

- We should show the add integration UI within the usage tab if thery which is the place people would be looking to see this data.
- I think we can reduce the need to go to docs by adding the yaml code snippet + copy button for the action with a view docs link. This could be in our standard expandable code-snippet UI, another advantage of this is that we could fill in some of the parts if we know a repository via a GH integration.

emyller:
@kyle-ssg I like both ideas. Now for the scope of this PR:

1. Should we keep this PR and add "Code References" to the Integrations tab? Personally I think it doesn't hurt. I'd be interested to find how many people engage via this path, or via the Feature UI.
2. Unless we decide to repurpose https://github.com/Flagsmith/flagsmith/issues/6915, let's write a new issue that describes the ideal UX. I can look into that, and will loop @asaphko in too.

frontend/common/stores/default-flags.ts
talissoncosta: Thanks for flagging that. It was fixed [here](https://github.com/Flagsmith/flagsmith/pull/6974/changes/953004175a7e5f0c1ff0516b825e02469fb04a0e).

talissoncosta:
Thanks for the review @emyller! I've pushed a [fix](https://github.com/Flagsmith/flagsmith/pull/6974/changes/953004175a7e5f0c1ff0516b825e02469fb04a0e) addressing both points:

- Added the missing period in the description
- Changed the icon stroke from `currentColor` to `#6C757D` so it's visible on both light and dark backgrounds

Regarding the broader UX discussion — I agree we should move forward with this PR as-is. Having Code References in the Integrations tab is a reasonable first step. For the ideal experience (surfacing setup in the feature's Usage tab, inline YAML snippet, etc.), let's track that in a dedicated issue to scope it properly.

emyller (approved):
:shipit:

PR #6977, opened 2026-03-18
fix: restore grid layout in PanelSearch broken by wrapper div

talissoncosta:

## Changes

Fixes a regression introduced in 362a9f8f6 (#6927).

The wrapper `<div>` added around `renderContainer(filteredItems)` to scope loading opacity to list items (not the header) broke Bootstrap grid layouts because `col-*` items were no longer direct children of the `row` container, causing the Organisations page, Projects page, and Import page to display cards vertically instead of in a horizontal grid.

**Fix:** Move `listClassName` from the outer `search-list` div to the items wrapper div, so that the `row` class and opacity live on the same element. This preserves Kyle's intended opacity scoping while restoring correct Bootstrap grid behaviour.

**Note:** The nested ternary in the render block was extracted into a `renderListContent` helper. This was required because touching the file triggered the pre-existing `no-nested-ternary` lint error, and we opted to fix it properly rather than adding `eslint-disable` comments.

- Verified the Organisations and Projects pages render cards in a horizontal grid layout instead of a vertical stack.
- Verified loading opacity only applies to list items, not the header.

### DIFF
--- a/frontend/web/components/PanelSearch.tsx
+++ b/frontend/web/components/PanelSearch.tsx
@@ -12,7 +12,6 @@ import { AutoSizer, List } from 'react-virtualized'
 import Popover from './base/Popover'
 import Input from './base/forms/Input'
 import Icon from './Icon'
-import classNames from 'classnames'
 import { IonIcon } from '@ionic/react'
 import { chevronDown, chevronUp } from 'ionicons/icons'
 import Button from './base/forms/Button'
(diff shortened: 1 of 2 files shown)


kyle-ssg:
Ah, I see what caused this—that’s on me. However, this undoes the original problem I was trying to solve.

This means that when you are filtering features for example everything is getting an opacity change vs the header staying normal:

image

talissoncosta:
Good, thanks for the context @kyle-ssg, let me dig a bit deeper.

kyle-ssg (approved):

kyle-ssg:
Thanks latest looks good @talissoncosta

frontend/web/components/PanelSearch.tsx
```diff
@@ -197,6 +196,58 @@ const PanelSearch = <T,>(props: PanelSearchProps<T>): ReactElement => {
     search = search.replace(/^"+|"+$/g, '')
   }
 
+  const renderListContent = (): ReactNode => {
```
talissoncosta, line 199: Extracted the nested ternary into a helper to resolve a pre-existing `no-nested-ternary` lint error. Touching the file triggered it — opted to fix it properly rather than adding an `eslint-disable` comment.

PR #6993, opened 2026-03-19
fix(frontend): avoid PermissionsTabs crash before projects load

### LINKED ISSUE #6553: TypeError when loading PermissionsTabs - projectData is undefined
## Summary

Users encounter a crash when accessing the organisation permissions page. The error `TypeError: Cannot read properties of undefined (reading 'map')` occurs when `OrganisationStore.getProjects()` returns `undefined` before data is loaded.

## Sentry Issue

- [FLAGSMITH-FRONTEND-48M](https://flagsmith.sentry.io/issues/6730628256/) - 76 events, 11 users affected
- First seen: 7/6/2025
- Last seen: 1/19/2026
- URL: `/organisation/{orgId}/permissions`

## Root Cause

In `frontend/web/components/PermissionsTabs.tsx:43`, the component assumes `getProjects()` always returns an array:

```typescript
const projectData: Project[] = OrganisationStore.getProjects()
```

When line 126 tries to map over it:

```typescript
mainItems={projectData.map((v) => ({ ...v, projectId: v.id }))}
```

If the store hasn't loaded yet, `projectData` is `undefined` and `.map()` crashes.

## Proposed Fix

Add a fallback to empty array on line 43:

```typescript
const projectData: Project[] = OrganisationStore.getProjects() || []
```

## Acceptance Criteria

- [ ] Add `|| []` fallback when getting projects from store
- [ ] Permissions page no longer crashes when projects haven't loaded
- [ ] E2E tests pass

saudademjj:

## Changes

Closes #6553

- guard the project permissions list against `OrganisationStore.getProjects()` returning `undefined` before the organisation data has loaded
- add a regression test for the loading case so `PermissionsTabs` keeps rendering with an empty project list instead of throwing

- on `100.92.89.12` via Docker `node:22`, ran `npm run test:unit -- --runInBand web/components/__tests__/PermissionsTabs.test.ts`
- on `100.92.89.12` via Docker `node:22`, ran `npx eslint web/components/PermissionsTabs.tsx web/components/__tests__/PermissionsTabs.test.ts`
- `npm run typecheck` currently fails on the repository baseline with many unrelated frontend errors on `main`; I did not change those files and did not treat them as blockers for this focused fix


saudademjj:
Code-side checks look healthy on my side: `apply-labels`, `pre-commit.ci`, and the targeted local validation for this fix all passed. The remaining red statuses appear to be Vercel team authorization gates rather than code failures.

saudademjj:
The failing checks are all Vercel deployments, each reporting "Authorization required to deploy". The code checks themselves are passing.

talissoncosta:
Hey @saudademjj, thanks for picking this up and for the thorough testing writeup! The fix is spot on 🎯

A couple of small suggestions:

**1. Use `??` instead of `||`**

`getProjects()` returns `store.model && store.model.projects`, which can evaluate to `false` (not just `undefined`). Nullish coalescing is more precise here — it only kicks in for `null`/`undefined`, not falsy values:
```diff
- mainItems={(projectData || []).map((v) => ({
+ mainItems={(projectData ?? []).map((v) => ({
```

**2. Drop the test file for now**

We really appreciate the effort of adding a regression test — that's the right instinct! In this case though, the component is tightly coupled to `OrganisationStore`, which means testing it in isolation requires a lot of mocking (9 `jest.mock()` calls + a global `Row` stub). That makes the test brittle and expensive to maintain for what it covers.

We're planning to decouple the store dependency down the line (e.g. passing `projectData` as a prop), and that's when proper tests will become straightforward and valuable. For now, the one-line fix speaks for itself.

saudademjj:
Good points — updated: swapped to `??` and removed the test file. Thanks for the review!

talissoncosta (approved):

PR #7003, opened 2026-03-20
refactor: modernise Server-side SDK Keys page

### LINKED ISSUE #7002: Migrate ServerSideSDKKeys to functional component with RTK Query
## Description

The `ServerSideSDKKeys.js` component is a class component using Flux stores (`ProjectStore`) that causes a crash when `ProjectStore.getEnvironment()` returns `undefined` (Sentry issue FLAGSMITH-FRONTEND-4FX). This blocks users from creating server-side keys.

### Current state
- Class component (~250 lines)
- Uses `ProjectStore.getEnvironment()` for environment name lookup
- Uses imperative `getStore().dispatch()` for RTK Query operations
- Contains nested `CreateServerSideKeyModal` class component

### Objectives

1. **Migrate to functional components with TypeScript**
2. **Break into smaller, focused components** (CreateServerSideKeyModal, ServerSideKeyRow)
3. **Replace ProjectStore with RTK Query** (`useGetEnvironmentsQuery`)
4. **Use RTK Query hooks** for server-side key CRUD operations
5. **Feature flag** (`rtk_server_side_sdk_keys`) to toggle between old and new

### Structure

```
pages/sdk-keys/
├── SDKKeysPage.tsx
├── index.ts
└── components/
    ├── ServerSideSDKKeys.tsx
    ├── CreateServerSideKeyModal.tsx
    ├── ServerSideKeyRow.tsx
    └── index.ts
```

### Related
- Sentry: FLAGSMITH-FRONTEND-4FX — `TypeError: Cannot read properties of undefined (reading 'name')` in `CreateServerSideKeyModal.render`

talissoncosta:

## Changes

Contributes to #7002

Migrate `ServerSideSDKKeys.js` from a Flux class component to functional components with RTK Query, behind a feature flag (`rtk_server_side_sdk_keys`).

### Migration

- Replace `ProjectStore.getEnvironment()` with `useGetEnvironmentsQuery` (fixes Sentry FLAGSMITH-FRONTEND-4FX)
- Replace imperative `getStore().dispatch()` with RTK Query hooks
- Full TypeScript with proper type definitions
- Permission check via `useHasPermission` hook

### Component structure

```
pages/sdk-keys/
├── SDKKeysPage.tsx                    # Page layout (client-side key + server-side keys)
├── index.ts                           # Barrel export
├── hooks/
│   └── useServerSideKeys.tsx          # Data fetching + CRUD handlers
└── components/
    ├── ServerSideSDKKeys.tsx           # UI — permission check, list rendering
    ├── CreateServerSideKeyModal.tsx    # Modal — uses React 19 useActionState
    ├── ServerSideKeyRow.tsx            # Individual key row
    └── index.ts                       # Barrel exports
```

### Feature flag strategy

Old `ServerSideSDKKeys.js` preserved untouched — feature flag toggles at `SDKKeysPage` level:

```tsx
if (Utils.getFlagsmithHasFeature('rtk_server_side_sdk_keys')) {
  return <SDKKeysPageRTK />
}
return <SDKKeysPageLegacy />
```

- **Flag off**: legacy component renders, zero risk
- **Flag on**: new RTK Query version renders
- **Rollback**: disable the flag — instant, zero-deploy

### React 19

`CreateServerSideKeyModal` uses `useActionState` to manage form submission — `isPending` and error state are handled automatically, replacing manual `useState` for `isSaving`.

### Bug fixes

- **Cache invalidation**: `providesTags` now matches `invalidatesTags` (`LIST` tag) so the keys list refreshes automatically after create/delete
- **Per-row delete loading**: uses RTK `originalArgs` to track which row is being deleted — only the affected row fades out
- **Icon colours**: removed hardcoded `fill='#656D7B'`, icons use `currentColor` via `text-body` class

### Cleanups

- Fix copy-paste IDs: `segments-page` → `sdk-keys-page`, `org-members-list` → `server-side-keys-list`, `remove-feature` → `remove-sdk-key`
- Fix `target='__blank'` → `_blank`
- Simplify `filterRow` to use `.includes()`
- Extract `handleCreateKey` from inline callback
- Extract `useServerSideKeys` hook to separate data logic from UI
- Add `opacity-50` reusable utility class
- Use `EnvironmentPermission.ADMIN` enum instead of raw string

## Reviewer notes

- **New code only** — no changes to legacy components, fully isolated
- **Feature flagged** — zero impact when `rtk_server_side_sdk_keys` is off
- The only change to existing code is `useServersideEnvironmentKey.ts` (cache tag fix) — this is a no-op for the legacy path since it manually refetches
- Tested manually with flag on and off on staging
- Safe to merge — the flag stays off until we're confident

- TypeScript compiles with no errors (`npm run typecheck`)
- ESLint passes on all modified files
- Visual parity confirmed — new component renders identically to legacy
- Manual testing with flag on: create, copy, delete keys all working
- Manual testing with flag off: legacy component unchanged
- Cache invalidation verified: list updates automatically after create/delete
- Per-row delete loading verified: only the deleted row fades, others stay interactive


frontend/web/components/pages/sdk-keys/components/CreateServerSideKeyModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+  onSubmit,
+}) => {
+  const inputRef = useRef<HTMLInputElement>(null)
+
+  useEffect(() => {
+    const timer = setTimeout(() => inputRef.current?.focus(), 500)
+    return () => clearTimeout(timer)
+  }, [])
+
+  const [error, submitAction, isPending] = useActionState(
```
talissoncosta, line 21: React 19's `useActionState` replaces the manual `useState(false)` for `isSaving` we had before. It manages pending and error states automatically — `isPending` resets on both success and failure, so the button can't get stuck disabled. The action receives `FormData` natively from `<form action={...}>`, removing the need for controlled input state and `Utils.preventDefault`.

frontend/web/components/SDKKeysPage.tsx
```diff
@@ -2,7 +2,8 @@ import React, { FC } from 'react'
 import Button from './base/forms/Button'
 import Input from './base/forms/Input'
 import Icon from './Icon'
-import ServerSideSDKKeys from './ServerSideSDKKeys'
+import ServerSideSDKKeysLegacy from './ServerSideSDKKeys'
+import { SDKKeysPage as SDKKeysPageRTK } from './pages/sdk-keys'
```
talissoncosta: The legacy component (`SDKKeysPageLegacy`) is the existing `ServerSideSDKKeys.js` — a Flux class component that can't be easily renamed or re-exported without modifying the legacy code. We intentionally leave the legacy untouched so both paths are fully isolated. When the flag is removed, this entire file gets deleted and the route points to the new component directly.

frontend/web/components/pages/sdk-keys/SDKKeysPage.tsx
```diff
@@ -0,0 +1,74 @@
+import React, { FC } from 'react'
```
Zaimwa9, line 1: React is unused here

frontend/web/components/SDKKeysPage.tsx
```diff
@@ -16,6 +17,10 @@ const SDKKeysPage: FC = () => {
   const match = useRouteMatch<RouteParams>()
   const environmentId = match?.params?.environmentId
 
+  if (Utils.getFlagsmithHasFeature('rtk_server_side_sdk_keys')) {
+    return <SDKKeysPageRTK />
+  }
```
Zaimwa9: Can't this be gating only the component in the Tsx page actually ? I saw your comment but it seems doable to just have this over the equivalent of L63 in the other one no ?

Zaimwa9: I would definitely prefer this and getting rid of the page (in case the flag sticks around for a while). In both cases let's open a clean up issue 1 pointer for next sprint to get rid of the flag and the duplication (I don't think this flag is meant to live long right?)

Zaimwa9 (changes requested):
Nice overall thanks. It would be preferable IMO to target the feature flag and get rid now of the previous SDKPage instead of duplicating it

frontend/web/components/pages/sdk-keys/SDKKeysPage.tsx
talissoncosta: Removed in 9533d5e52.

frontend/web/components/SDKKeysPage.tsx
talissoncosta: Good call — restructured in 9533d5e52. The flag now gates only the `ServerSideSDKKeys` component, no duplicate page. Also applied the bug fixes (data-test id, `target='_blank'`, icon colour) to the single page. Cleanup issue: #7140.

Zaimwa9 (dismissed):
Thanks for making the changes but one question on your radar: Have the changes been made in the right duplicated page?
It looks like the one in `pages/sdk-keys` got removed while the one in components stayed?

Zaimwa9 (approved):

PR #7008, opened 2026-03-20
feat(Dogfooding): Add OpenFeature SDK for server-side Flagsmith-on-Flagsmith

khvn26:

## Changes

Migrate Flagsmith-on-Flagsmith feature evaluation from the native Flagsmith Python SDK to the vendor-neutral OpenFeature SDK with the Flagsmith provider.

Added `openfeature-sdk` and `openfeature-provider-flagsmith` (`>=0.1.6`), now published on PyPI with flagsmith SDK v5 support.

Replaced `get_client()` with `get_openfeature_client()` in `integrations/flagsmith/client.py`, using OpenFeature's provider registry with a named domain (`flagsmith-api`) and lazy initialisation via `ProviderStatus` check.

Replaced `flagsmith_identifier` + `flagsmith_on_flagsmith_api_traits` on the Organisation model with a single `openfeature_evaluation_context` property returning an `EvaluationContext`.

Migrated all call-sites from `client.get_identity_flags(...).is_feature_enabled("x")` to `client.get_boolean_value("x", default_value=False, evaluation_context=...)`, including leftover `flags.is_feature_enabled()` references in `organisations/tasks.py` and `organisations/task_helpers.py`.

`enable_features` test fixture now uses OpenFeature's `InMemoryProvider` instead of mocking Flagsmith internals. 22 org tasks tests migrated from manual `get_openfeature_client` mocks to the fixture.

Added `openfeature_flagsmith.*` and `licensing.*` to mypy `ignore_missing_imports` overrides (packages lack `py.typed`).

Made sure existing tests pass.


api/conftest.py
```diff
@@ -1278,31 +1279,30 @@ def set_github_webhook_secret() -> None:
...
-        mock_clients = mocker.MagicMock(spec=dict)
-        mock_clients.__getitem__.return_value = mock_flagsmith
-
-        mocker.patch(
-            "integrations.flagsmith.client._flagsmith_clients",
-            new=mock_clients,
+        flags = {
+            name: InMemoryFlag(
+                variants={"enabled": True},
+                default_variant="enabled",
```
khvn26, line 1294: Yes, this is how to create an enabled by default in-memory flag for OF. No, there is no shorter way to do this.

api/pyproject.toml
```diff
@@ -94,6 +94,10 @@ ignore_missing_imports = true
 module = ["saml.*"]
 ignore_missing_imports = true
 
+[[tool.mypy.overrides]]
+module = ["openfeature_flagsmith.*"]
+ignore_missing_imports = true
```
khvn26, line 103: Opened https://github.com/Flagsmith/flagsmith-openfeature-provider-python/issues/30 in the meantime.

Zaimwa9 (commented):
d

api/conftest.py
Zaimwa9: NIT: I got a bit confused it because of `enabled`. Bit of a detail just for next readings, do you think we could rename it `enabled_variant` to avoid thinking it's related to the `enabled` property ?

Zaimwa9 (commented):
Not much too say, I would approve if it wasn't for the dependency update and merge conflicts to come. Good addition overall

api/conftest.py
khvn26: TBH I'm not sure if it's possible to make this code less confusing, I'd just let it be for now.

api/integrations/flagsmith/client.py
```diff
@@ -1,54 +1,69 @@
...
 
-def get_client(name: str = "default", local_eval: bool = False) -> Flagsmith:
-    global _flagsmith_clients
+def get_openfeature_client(
+    domain: str = DEFAULT_OPENFEATURE_DOMAIN,
+    **flagsmith_kwargs: typing.Any,
+) -> OpenFeatureClient:
+    openfeature_client = openfeature_api.get_client(domain=domain)
+    if openfeature_client.get_provider_status() != ProviderStatus.READY:
+        initialise_provider(domain, **(flagsmith_kwargs or get_provider_kwargs()))
```
Zaimwa9: Are we sure about this code? It looks like a ticking bomb. If we pass in an argument, we don't set any other property, not even the api key?

The only other callsite is in [this test](https://github.com/Flagsmith/flagsmith/pull/7008/changes/40ff085bacbad26baedffdeb7a9e962702d0a1df#diff-1ef3c47cbe87e02aa5a53dcf72d0c31f29de4e17d9356dcbc2a982f149ca4441R104-R108) that only asserts the given parameter (local evaluation)

khvn26: Addressed in f8770f02a — removed `**flagsmith_kwargs` entirely since no production callers use it.

Zaimwa9 (approved):

PR #7011, opened 2026-03-22
fix(ui): replace misleading permission tooltip on segment override variations

### LINKED ISSUE #6982: Misleading permission tooltip on read-only variation values in segment overrides
## Description

When viewing segment overrides for a multivariate feature, the variation values are correctly shown as read-only (variation values are defined at the feature level and cannot be changed per segment). However, hovering over the disabled variation value inputs shows a misleading tooltip:

> *"To use this feature you need the Create Feature permission for this project. Please contact a member of this project who has administrator privileges."*

This makes it appear as though the user lacks permissions, when in reality this is by design — variation values simply cannot be edited at the segment override level.

## Steps to reproduce

1. Create a multivariate feature with variations
2. Go to Segment Overrides tab
3. Add or view a segment override
4. Hover over a variation value input

**Expected:** A tooltip explaining that variation values are set at the feature level (e.g. *"Variation values are defined at the feature level and cannot be changed per segment"*).
**Actual:** A misleading "Create Feature permission" tooltip appears.

## Root cause

In `SegmentOverrides.js` (line ~307), `VariationOptions` is rendered without passing the `canCreateFeature` prop:

```jsx
<VariationOptions
  readOnly
  disabled={readOnly}
  // canCreateFeature is never passed — defaults to undefined (falsy)
  ...
/>
```

`VariationValueInput.tsx` (line 38) wraps the value editor with `Utils.renderWithPermission(canCreateFeature, ...)`. Since `canCreateFeature` is `undefined`, it always shows the permission tooltip — even though the input is already disabled via the `readOnly` prop.

## Suggested fix

Replace the misleading permission tooltip with a helpful message explaining the actual reason: *"Variation values are defined at the feature level and cannot be changed per segment."*

This could be done by either:
- Adding a new tooltip variant for read-only variation values in segment overrides
- Or passing `canCreateFeature={true}` to suppress the permission tooltip and adding an informational tooltip on the disabled input instead

## Reported by

Customer via Slack — they have org admin + full permissions and were confused by the tooltip.

nielskaspers:
## Summary
- Fixes misleading "Create Feature permission" tooltip shown on read-only variation value inputs in segment overrides
- When variations are read-only due to being in a segment override context, the tooltip now shows: "Variation values are defined at the feature level and cannot be changed per segment."
- The original permission tooltip is preserved for cases where the user genuinely lacks Create Feature permission

## Issue
Fixes #6982

## Changes
- `frontend/web/components/mv/VariationValueInput.tsx`: When `readOnly` is true and `canCreateFeature` is falsy, show an informative tooltip instead of the permission-related one

## Testing
1. Create a multivariate feature with variations
2. Go to Segment Overrides tab
3. Add or view a segment override
4. Hover over a variation value input
5. Verify the tooltip now says "Variation values are defined at the feature level and cannot be changed per segment."
6. Verify that non-segment-override variation editing still shows the correct permission tooltip when the user lacks Create Feature permission

### DIFF
--- a/frontend/web/components/mv/VariationValueInput.tsx
+++ b/frontend/web/components/mv/VariationValueInput.tsx
@@ -37,7 +37,11 @@ export const VariationValueInput: React.FC<VariationValueProps> = ({
             <>
               {Utils.renderWithPermission(
                 canCreateFeature,
-                Constants.projectPermissions(ProjectPermission.CREATE_FEATURE),
+                readOnly
+                  ? 'Variation values are defined at the feature level and cannot be changed per segment.'
+                  : Constants.projectPermissions(
+                      ProjectPermission.CREATE_FEATURE,
+                    ),
                 <ValueEditor
                   data-test={`featureVariationValue${
                     Utils.featureStateToValue(value) || index


talissoncosta (approved):
Thanks for the contribution @nielskaspers ! Tested locally and the fix works correctly. Approving. 😃

PR #7014, opened 2026-03-23
fix: `compress_dynamo_documents` breaks permanent environment cache

### LINKED ISSUE #6944: `compress_dynamo_documents` breaks permanent environment cache
If `compress_dynamo_documents` flag set to true, permanent environment cache (the one configured with `CACHE_ENVIRONMENT_DOCUMENT_` settings) gets broken.

A correct approach would be to use a mapper with a different destination type for the cache — `flagsmith_schemas.api.V1EnvironmentDocumentResponse` seems appropriate, and it's already being used for the cache consuming view's API docs.

Related to #6654.

We should abstain from enabling `compress_dynamo_documents` for self-hosted defaults before this issue is solved.

SahilJat:

## Changes

Closes #6944 

_Please describe._
Since ` V1EnvironmentDocumentResponse` is a TypedDict, I implemented an explicit dictionary extraction right before `environment_document_cache.set_many()`. This guarantees that internal DynamoDB configs (like compress_dynamo_documents) do not leak into the Redis cache consumed by the API.

I also added a mocked unit test to verify the payload is stripped and perfectly validates against the V1EnvironmentDocumentResponse schema.

_Please describe._
1.Mocked the Cache: Used @mock.patch to intercept the payload right before it hits environment_document_cache.set_many().

2.Forced Persistent Mode: Set CACHE_ENVIRONMENT_DOCUMENT_MODE to PERSISTENT to trigger the exact code path causing the leak.

2.Asserted Data Stripping: Verified that internal configuration fields (like compress_dynamo_documents and use_v2_feature_versioning) are successfully stripped from the payload.

4.Schema Validation: Explicitly proved that the intercepted dictionary perfectly aligns with the V1EnvironmentDocumentResponse TypedDict, guaranteeing that the public API view will not break.

### DIFF
--- a/api/environments/models.py
+++ b/api/environments/models.py
@@ -4,7 +4,6 @@ import uuid
 from copy import deepcopy
 from typing import TYPE_CHECKING, Literal
 
-from flagsmith_schemas.api import V1EnvironmentDocumentResponse
 from common.core.utils import using_database_replica
 from django.conf import settings
 from django.contrib.contenttypes.fields import GenericRelation
@@ -57,7 +56,6 @@ from metadata.models import Metadata
 from projects.models import Project
 from segments.models import Segment
 from util.mappers import (
-    map_environment_to_environment_document,
     map_environment_to_sdk_document,
 )
 from webhooks.models import AbstractBaseExportableWebhookModel
@@ -366,7 +364,7 @@ class Environment(
                     e.api_key: map_environment_to_sdk_document(e)
                     for e in environments
                 }
-            )            
+            )
 
     def get_feature_state(
         self,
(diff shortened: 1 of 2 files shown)


SahilJat:
hi @khvn26 all the precommit has been passed the only test failing require authorization , please let me know if its good to go or do i have to make some changes also I noticed that `compress_dynamo_documents` was temporarily disabled in `api/integrations/flagsmith/data/environment.json` to mitigate this cache poisoning. Once this PR is merged and deployed, you should be completely safe to flip that flag back to true!

khvn26 (dismissed):
Hello @SahilJat — happy to approve this, and thanks for the contribution!

khvn26:
Sorry @SahilJat — the solution looks sound to me but I'd like the failing tests fixed. Can you see to that?

SahilJat:
hi @khvn26 i made the changes inside the test files ,

khvn26 (approved):
👍

Issue #6991, opened 2026-03-19
Replace restricted Bitnami PostgreSQL HA image with open-source alternative to fix self-hosted HA deployments


igor-soldev:
Hi ,

I was reviewing the Docker setup for the High Availability (HA) database cluster and noticed that `docker/common/db/.pgpool.yml` relies on the `bitnami/postgresql-repmgr:15` Docker image.

**The Problem**
Bitnami has recently restricted public access to their Docker images, placing them behind a paywall with strict rate limits. Relying on `docker.io/bitnami/postgresql-repmgr` now frequently results in `pull access denied` or rate-limit errors. For a popular open-source project like Flagsmith, this means your self-hosted users trying to spin up an HA database cluster will likely face deployment failures right out of the box, creating a frustrating setup experience and potential support overhead.

**Proposed Solution**
I suggest switching to **`soldevelo/postgresql-repmgr`**.
This image is a free, actively maintained, and fully open-source fork of the original Bitnami image. 

Crucially, it is a **drop-in replacement**:
* It accepts the exact same environment variables for `repmgr` configuration.
* It maintains the same internal paths and volume mounts.
* Your existing `.pgpool.yml` and related docker-compose setups will continue to work immediately without requiring any structural changes or refactoring.

*(Note: SolDevelo maintains a growing library of open-source, Bitnami-compatible images specifically to help the OSS community bypass these new restrictions. You can find the source code and other images here: [https://github.com/SolDevelo/containers](https://github.com/SolDevelo/containers))*

**Suggested Changes in `docker/common/db/.pgpool.yml`:**
```yaml
# Old
  pg-0:
    image: docker.io/bitnami/postgresql-repmgr:15
  pg-1:
    image: docker.io/bitnami/postgresql-repmgr:15

# New
  pg-0:
    image: docker.io/soldevelo/postgresql-repmgr:15
  pg-1:
    image: docker.io/soldevelo/postgresql-repmgr:15
```

I would be happy to submit a Pull Request to update the image references across the Docker configuration files so that self-hosted HA deployments remain reliable. Would you be open to that?

khvn26:
Hi @igor-soldev, you're very welcome to open a PR — we'll happily review it.

matthewelwell:
See comment on the PR [here](https://github.com/Flagsmith/flagsmith/pull/7015#pullrequestreview-4020187451) for context. I am closing this issue in the meantime.

PR #7015, opened 2026-03-23
chore: Update PostgreSQL repmgr image from bitnami to soldevelo

igor-soldev:
## Changes

Contributes to https://github.com/Flagsmith/flagsmith/issues/6991

### DIFF
--- a/docker/common/db/.pgpool.yml
+++ b/docker/common/db/.pgpool.yml
@@ -3,7 +3,7 @@
 
 services:
   pg-0:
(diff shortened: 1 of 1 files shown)


matthewelwell (commented):
@igor-soldev it looks like this image doesn't publish multiarch builds? Therefore it's not possible to run on ARM architectures. 

This also looks like a fairly loosely maintained personal docker repository so I'm not sure that it's mature enough yet for us to include in our repository - sorry. 

Since pgpool is not part of the core infrastructure, I don't think it's worth pursuing this at this time.

PR #7049, opened 2026-03-26
feat: implement-gram-elements

### LINKED ISSUE #6572: Implement Gram Elements Chatbot in SaaS
We're invited to join Gram as design partners for their new Gram Elements feature, which is a drop-in AI-chatbot that uses our MCP together with an LLM for an in-app conversational chat solution.

Implementation details: https://www.getgram.ai/docs/gram-elements/quickstart
Additional info/questions can be chatted about in our shared channel with Gram.

Zaimwa9:

- [ ] I have filled in the "Changes" section below.
- [ ] I have filled in the "How did you test this code" section below.

## Changes

- New `GramChat` component: widget variant with dark/light mode support
- Gated behind `gram_chat`
- New express `POST /api/gram/session` endpoint validates the user's auth token against API, then creates a Gram session 
- ⚠️ Passes user's auth token to MCP server via environment config for per-user permissions
- ⚠️ Shadow DOM workaround for stuck motion entrance animations
- ⚠️ CJS shim for remark-gfm ESM/CJS interop with webpack

**Webpack**:
- ⚠️ Added `remark-gfm` alias to CJS shim (ESM default export doesn't unwrap properly)
- ⚠️ Added `fullySpecified: false` and ESM module rule for node_modules compatibility

⚠️ ⚠️ Dependencies all pinned to exact versions because the package is too fraglie:
- @gram-ai/elements, @assistant-ui/react, @assistant-ui/react-ai-sdk, @assistant-ui/react-markdown, @json-render/react, motion, remark-gfm, zustand, vega, shiki
- ⚠️  npm override: `assistant-stream@0.3.7` for `@assistant-ui/react-ai-sdk` (version mismatch)

**Risks**
- Shadow DOM motion hack: Gram renders in shadow DOM, motion animations (`framer-motion`) don't fire inside shadow DOM, elements stay invisible. We force them visible via injected CSS.
- `remark-gfm` CJS shim
- version mismatch between `@assistant-ui` packages
- User token sent to Gram MCP => Check with them their policies
- Bundle size (+650 packages)
- Datadog RUM telemetry => spamming events to Datadog but should have benn mitigated with `errorTracking: { enabled: false }`

image

image


talissoncosta:
Can you please rebase it on top of main? @Zaimwa9

frontend/package.json
```diff
@@ -56,11 +64,13 @@
...
     "@dnd-kit/react": "^0.3.2",
     "@flagsmith/flagsmith": "^11.0.0-internal.5",
+    "@gram-ai/elements": "1.27.5",
     "@ionic/react": "^7.5.3",
+    "@json-render/react": "0.2.0",
     "@react-oauth/google": "^0.2.8",
     "@reduxjs/toolkit": "1.9.1",
     "@sentry/browser": "^7.119.1",
-    "@sentry/webpack-plugin": "^2.22.7",
+    "@sentry/webpack-plugin": "2.22.7",
```
Zaimwa9, line 73: Pinned it to the exact version because otherwise it can upgrade to `2.23.1` (frozen in the lockfile) that likely introduced breaking changes surfaced in vercel

frontend/web/components/GramChat.tsx
```diff
@@ -0,0 +1,207 @@
...
+      subtitle: 'How can I help you today?',
+      title: 'Welcome',
+    },
+  }
+  useEffect(() => {
+    injectShadowFix(document)
+    const observer = new MutationObserver(() => {
+      injectShadowFix(document)
+    })
+    observer.observe(document.body, {
```
talissoncosta: This fires `injectShadowFix(document)` (which does `querySelectorAll('*')`) on every DOM mutation in `<body>`. On pages with frequent updates (e.g., feature list), this could get expensive. Could we scope the observer to just the Gram container element instead of `document.body`?

frontend/web/components/GramChat.tsx
```diff
@@ -0,0 +1,207 @@
...
+    headers,
+    method: 'POST',
+  })
+    .then((res) => res.json())
+    .then((d) => d.client_token)
+}
+
+export default function GramChat() {
+  const isDark = useDarkMode()
+  const config: ElementsConfig = {
```
talissoncosta: Nit: `config` is recreated on every render. If `GramElementsProvider` does shallow comparison, this would cause unnecessary re-renders. Worth wrapping in `useMemo` with `[isDark, data.token]` as deps.

frontend/web/components/GramChat.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+const getSession = async () => {
+  const headers: Record<string, string> = {
+    'Gram-Project': Project.gramProjectSlug,
+  }
+
+  if (data.token && !Project.cookieAuthEnabled) {
+    headers.Authorization = `Token ${data.token}`
+  }
+  return fetch('/api/gram/session', {
```
talissoncosta: Is there a reason for not using RTK Query here? Alternatively, since we're on React 19, the use() hook could consume the session promise directly without the .then() chain. Just curious about the choice.

frontend/web/components/GramChat.tsx
```diff
@@ -0,0 +1,207 @@
...
+  }
+  /* Nudge the bubble up to avoid overlap with bottom bar */
+  .aui-modal-anchor {
+    bottom: 26px !important;
+  }
+  [class*="aui-modal-anchor"],
+  [class*="fixed"][class*="bottom"] {
+    bottom: 26px !important;
+  }
+`
```
talissoncosta, line 36: ~60 lines of hardcoded hex values with `!important` on every property, duplicated across dark and light themes. Not a blocker, but are there plans to extract this into a separate CSS file and reference our existing CSS custom properties instead of hardcoding hex values?

talissoncosta (changes requested):
Looks good @Zaimwa9 
Just some clarifications.

frontend/web/components/GramChat.tsx
Zaimwa9: Yes, that's really dirty and not stable at all (any change on their side would break it). I will move it to custom properties but i'm not sure it will fix for sure the long term issue.

If ever we decide to roll-out this for users, my plan is to directly open a PR in their open source lib addressing all the problems we have. Including the shadow injection that breaks style

frontend/web/components/GramChat.tsx
Zaimwa9: Thanks, I will definitely look into this

Zaimwa9: Actually this comment is not relevant as we are passing the promise to Gram as a callback and not using it ourselves in the render

talissoncosta (dismissed):

talissoncosta (approved):

PR #7065, opened 2026-03-30
fix: dissociated-code-references-gate-from-flag-response

Zaimwa9:

## Changes
- Showing code references in feature flag modal was dependent on the data existing from `Get /flags`
- Always show code references in Usage (with empty message)
- Side fixed the dark mode in the Usage chart

image

image

image

### DIFF
--- a/frontend/web/components/feature-page/FeatureNavTab/FeatureAnalytics.tsx
+++ b/frontend/web/components/feature-page/FeatureNavTab/FeatureAnalytics.tsx
@@ -106,7 +106,10 @@ const FlagAnalytics: FC<FlagAnalyticsType> = ({
                   tick={{ fill: '#656D7B' }}
                   axisLine={{ stroke: '#656D7B' }}
                 />
-                <Tooltip cursor={{ fill: 'transparent' }} />
+                <Tooltip
+                  cursor={{ fill: 'transparent' }}
+                  labelStyle={{ color: '#1a1a1a' }}
+                />
                 {sortBy(environmentIds, (id) =>
                   environments?.results?.findIndex((env) => `${env.id}` === id),
                 ).map((id) => {
--- a/frontend/web/components/modals/create-feature/index.tsx
+++ b/frontend/web/components/modals/create-feature/index.tsx
@@ -380,7 +380,7 @@ const CreateFeatureModal: FC<CreateFeatureModalProps> = (props) => {
     controlValue < 0
   const isVersionedChangeRequest = existingChangeRequest && isVersioned
   const hideIdentityOverridesTab = Utils.getShouldHideIdentityOverridesTab()
-  const hasCodeReferences = projectFlag?.code_references_counts?.length > 0
+  const hasCodeReferenceCounts = projectFlag?.code_references_counts?.length > 0
 
   let regexValid = true
   try {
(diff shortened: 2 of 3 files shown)


frontend/web/components/modals/create-feature/index.tsx
```diff
@@ -659,7 +659,7 @@ const CreateFeatureModal: FC<CreateFeatureModalProps> = (props) => {
                       />
                     </TabItem>
                   )}
-                  {(!Project.disableAnalytics || hasCodeReferences) && (
+                  {(!Project.disableAnalytics || hasCodeReferenceCounts) && (
```
Zaimwa9: To be honest, the second part `hasCodeReferenceCounts` is made a bit useless but it's to avoid having the `Usage` tab visible when both `disableAnalytics` is activated and no code references are scanned. It's quite edgy but I prefer to see the evolution

matthewelwell: Is that actually what we want though, or do we want to show it with the information about how to enable analytics and/or code references? Maybe one for another day, but just curious.

frontend/web/components/feature-page/FeatureNavTab/FeatureAnalytics.tsx
```diff
@@ -106,7 +106,10 @@ const FlagAnalytics: FC<FlagAnalyticsType> = ({
                   tick={{ fill: '#656D7B' }}
                   axisLine={{ stroke: '#656D7B' }}
                 />
-                <Tooltip cursor={{ fill: 'transparent' }} />
+                <Tooltip
+                  cursor={{ fill: 'transparent' }}
+                  labelStyle={{ color: '#1a1a1a' }}
```
matthewelwell, line 111: @talissoncosta curious here if this is something we should be careful of with your upcoming colour improvements. Are there any concerns with this based on dark mode in the meantime?

frontend/web/components/modals/create-feature/index.tsx
```diff
@@ -670,7 +670,7 @@ const CreateFeatureModal: FC<CreateFeatureModalProps> = (props) => {
                         projectId={projectId}
                         featureId={projectFlag.id}
                         environmentId={environment.id}
-                        hasCodeReferences={hasCodeReferences}
+                        hasCodeReferences={true}
```
matthewelwell: Couldn't we just remove this argument from the component instead? Even if we do bring back the counts in the list endpoint, I like the idea of always showing the section in the Usage tab to encourage people to look at it, so I don't see a reason that we'd need to keep this option in the component.

frontend/web/components/feature-page/FeatureNavTab/FeatureAnalytics.tsx
Zaimwa9: I thought exactly the same and wanted to ping @talissoncosta. I went with minimal changes here because the tooltip is not fully dark mode aware, just the label color hence why I hardcoded it here. I can go with a follow-up to support dark mode colors.

See prod:
image

frontend/web/components/modals/create-feature/index.tsx
Zaimwa9: I would have them always visible personally. Knowing that `disableAnalytics` is a subpath of FoF, I doubt it will impact a lot of users. And it would make code references discoverable for them

matthewelwell: Agreed.

frontend/web/components/feature-page/FeatureNavTab/FeatureAnalytics.tsx
talissoncosta: It is all good, thanks for handling it @Zaimwa9. We have an open issue to go over the charts so we can handle it properly.

talissoncosta (approved):

PR #7067, opened 2026-03-30
feat: enforce feature ownership project setting

### LINKED ISSUE #4432: Project setting: Enforce feature ownership

gagantrivedi:

## Changes

Closes #4432

Add a project-level `enforce_feature_owners` setting. When enabled:

- Feature creation requires at least one user or group owner
- The `owners`/`group_owners` fields on the create serialiser now accept IDs on write and return nested objects on read (asymmetric PrimaryKeyRelatedField — no API contract break)
- The remove-owners/remove-group-owners endpoints prevent removing the last owner
- Frontend adds a project setting toggle, owner selection during feature creation, create-button validation, and owner chips in the feature modal header

- 18 new backend tests covering:
  - Enforcement on/off for create
  - Valid/invalid owner IDs, wrong org groups, users without project access
  - Master API key without owners
  - Owners read-only on update
  - Remove endpoints blocking last owner removal (user and group)
  - Project setting PATCH and list
- Frontend: manual testing via local dev server
- All existing tests pass (247 features, 48 projects)


api/features/serializers.py
```diff
@@ -240,12 +259,22 @@ def create(self, validated_data: dict) -> Feature:  # type: ignore[type-arg]
...
+
         # Add the default(User creating the feature) owner of the feature
         # NOTE: pop the user before passing the data to create
         user = validated_data.pop("user", None)
         instance = super(CreateFeatureSerializer, self).create(validated_data)  # type: ignore[no-untyped-call]
         if user and getattr(user, "is_master_api_key_user", False) is False:
             instance.owners.add(user)
+
+        if owners:
+            instance.owners.add(*owners)
```
Zaimwa9, line 270: On the above (L270), maybe we should skip adding the user as an owner if owners is passed in? 

Or NIT: removing it from the owners object?

frontend/web/components/FeatureOwnerSelect.tsx
```diff
@@ -0,0 +1,154 @@
+import React, { FC, useState } from 'react'
```
Zaimwa9: So actually, this component is great. But we have 2 very similar implementations of assigning users/groups (with the one in settings that has the icon).

It's a small refactor that should be straightforward but let me know if that feels creeping out of scope. It would do a lot of good as moving towards Functional Components and RTK.

The main difference is the old components call the API, your new components trigger a state update. So we should be able to let the parent deal with it and have the components with `selectedIds` / `onAdd` / `onRemove` props

I'd say:
1. We get rid of this component sorry
2. We migrate `FlagOwnerGroups.js` => `FlagOwnerGroups.tsx` turning it into a Functional Component so we get rid of `OrganisationProvider` (old store) and use `useGetGroupSummariesQuery` to get the groups
3. We migrate `FlagOwners.js` => `FlagOwners.tsx` turning it into a Functional Component so we get rid of `OrganisationProvider` (old store) and use `useGetUsersQuery` to get the users
4. For the create part here we pass in the state setter
5. For the update we pass in directly a callback with the API call (with RTK instead of `data.post`)

That'd be the best opportunity to do this but I understand if it's too much to ask

Zaimwa9: I wanted to suggest to add the icon here to make it more obvious that you can add the user and it led me to look into the existing components

frontend/web/components/feature-summary/FeatureRow.tsx
```diff
@@ -236,6 +251,33 @@ const FeatureRow: FC<FeatureRowProps> = (props) => {
...
         </Button>
+        {ownerChips.length > 0 && (
+          <div className='d-flex align-items-center gap-1 ms-3'>
+            <Icon name='people' width={16} fill='#9DA4AE' />
+            {ownerChips.slice(0, 3).map((chip) => (
+              <Tooltip
+                key={chip.id}
+                title={<span className='chip chip--xs'>{chip.label}</span>}
+              >
+                {chip.label}
```
Zaimwa9, line 267: Maybe replace with the email ? Or just keep the chip and remove the tooltip ?

frontend/web/components/feature-summary/FeatureRow.tsx
```diff
@@ -236,6 +251,33 @@ const FeatureRow: FC<FeatureRowProps> = (props) => {
         >
           <Icon name='copy' />
         </Button>
+        {ownerChips.length > 0 && (
```
Zaimwa9, line 262: To align with the team:
- This is new and visible for every feature now no matter the new config enabled or not.  Do we want it?
- Same question for the assigned users/groups: they became part of the create form with or without the new config. Is it something we want to introduce or hide ?

frontend/common/types/requests.ts
```diff
@@ -652,7 +653,10 @@ export type Req = {
   }
   createProjectFlag: {
     project_id: number
-    body: ProjectFlag
+    body: ProjectFlag & {
+      owners?: number[]
+      group_owners?: number[]
```
Zaimwa9, line 658: We can directly add them to `ProjectFlag` then

Zaimwa9 (changes requested):
Overall good and well working. Couple of comments:
- One refactor sorry but that's part of becoming a frontend 😅 
- Couple of questions regarding new UI elements (to be hidden or not depending on the config)
- One backend NIT

Otherwise good job

api/features/serializers.py
gagantrivedi: Done — if `owners` is explicitly provided, we now skip the auto-add of the creating user. The `elif` in `create()` only auto-adds when no explicit owners are passed.

frontend/web/components/FeatureOwnerSelect.tsx
gagantrivedi: Done — migrated both `FlagOwners.js` → `FlagOwners.tsx` and `FlagOwnerGroups.js` → `FlagOwnerGroups.tsx` as functional components using `useGetUsersQuery` / `useGetGroupSummariesQuery` (RTK Query). Both support dual mode via discriminated union props:
- **Edit mode** (`id` + `projectId`): manages internal state, calls `data.post` for add/remove
- **Create mode** (`selectedIds` + `onAdd`/`onRemove`): purely controlled by parent

Deleted `FeatureOwnerSelect.tsx` — no longer needed.

frontend/web/components/feature-summary/FeatureRow.tsx
gagantrivedi: Done — removed the tooltips, just plain chips now.

frontend/web/components/feature-summary/FeatureRow.tsx
gagantrivedi: Good call — both the owner chips in the header and the owner picker in the create form are now gated on `project.enforce_feature_owners`. They only show when the setting is enabled. The edit-mode FlagOwners/FlagOwnerGroups in the settings tab still always show (unchanged behaviour).

frontend/common/types/requests.ts
gagantrivedi: Looked into this — `ProjectFlag.owners` is typed as `User[]` (nested objects from the API response), but on create we send `number[]` (IDs). Moving the `number[]` into `ProjectFlag` would conflict with the existing `User[]` field. The intersection type `ProjectFlag & { owners?: number[] }` correctly represents this asymmetry, so kept as-is.

khvn26 (dismissed):

Zaimwa9 (approved):
Tested new changes 🟢

api/features/serializers.py
```diff
@@ -275,6 +304,26 @@ def validate_multivariate_options(self, multivariate_options):  # type: ignore[n
                 raise serializers.ValidationError("Invalid percentage allocation")
         return multivariate_options
 
+    def validate_owners(self, owners: list[FFAdminUser]) -> list[FFAdminUser]:
+        project: Project = self.context["project"]
+        for user in owners:
+            if not user.has_project_permission(VIEW_PROJECT, project):
+                raise serializers.ValidationError(
+                    "Some users do not have access to this project."
+                )
```
khvn26, line 313: nit: Does it make sense to include specific user email(s) in the response?

api/features/serializers.py
```diff
@@ -275,6 +304,26 @@ def validate_multivariate_options(self, multivariate_options):  # type: ignore[n
...
+
+    def validate_group_owners(
+        self, group_owners: list[UserPermissionGroup]
+    ) -> list[UserPermissionGroup]:
+        project: Project = self.context["project"]
+        for group in group_owners:
+            if group.organisation_id != project.organisation_id:
+                raise serializers.ValidationError(
+                    "Some groups do not belong to this project's organisation."
+                )
```
khvn26, line 324: Same as [here](https://github.com/Flagsmith/flagsmith/pull/7067/changes#r3028864294)

khvn26 (approved):
Looks good bar 2 nits

api/features/serializers.py
gagantrivedi: I will create a different pr for this to not block my tf work

api/features/serializers.py
gagantrivedi: Not, email, but I can't return the ID since the endpoint only accepts IDs: https://github.com/Flagsmith/flagsmith/pull/7116/changes

PR #7068, opened 2026-03-30
fix(ci): Docker compose migration race condition

matthewelwell:
## Changes

Unsolicited improvement based on frustrations when using docker-compose. Currently on start up one of `flagsmith` or `flagsmith-task-processor` can fail based on a race condition when they both attempt to run the migrations. This change delegates the migrations to a new `init-db` service which the other 2 services rely on. 

It comes with a bonus DRY-ing up of the docker compose file to centralise the environment variables. I've shared all of the env vars between the different services since, while they are not all strictly necessary for each service, they are harmless when included in the services that don't strictly need them. We could, if we wanted to, clean this up to have service specific env vars in the relevant service if we want to. 

`docker-compose up`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,8 +20,23 @@ services:
       retries: 20
       start_period: 20s
 
+  init-db:
+    image: docker.flagsmith.com/flagsmith/flagsmith:latest
+    command:
+      - migrate
+    environment:
+      DATABASE_URL: postgresql://postgres:password@postgres:5432/flagsmith
+      USE_POSTGRES_FOR_ANALYTICS: 'true' # Store API and Flag Analytics data in Postgres
+    healthcheck:
+      test: ["NONE"]
+    depends_on:
+      postgres:
+        condition: service_healthy
+
   flagsmith:
     image: docker.flagsmith.com/flagsmith/flagsmith:latest
+    command:
+      - serve
     environment:
       # All environments variables are available here:
       # API: https://docs.flagsmith.com/deployment/locally-api#environment-variables
@@ -57,8 +72,8 @@ services:
     ports:
       - 8000:8000
     depends_on:
-      postgres:
-        condition: service_healthy
+      init-db:
+        condition: service_completed_successfully
 
   # The flagsmith_processor service is only needed if TASK_RUN_METHOD set to TASK_PROCESSOR
   # in the application environment
@@ -72,5 +87,6 @@ services:
     ports:
       - 8001:8000
     depends_on:
-      - flagsmith
+      init-db:
+        condition: service_completed_successfully
     command: run-task-processor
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -20,57 +52,39 @@ services:
       retries: 20
       start_period: 20s
 
-  flagsmith:
+  init-db:
```
khvn26: Let's be more explicit here.

```suggestion
  migrate-db:
```

docker-compose.yml
```diff
@@ -20,57 +52,39 @@ services:
...
+      - serve
+    environment:
+      <<: *flagsmith-env
     ports:
       - 8000:8000
     depends_on:
-      postgres:
-        condition: service_healthy
+      init-db:
+        condition: service_completed_successfully
```
khvn26: Based on [prior](https://github.com/Flagsmith/flagsmith/pull/7068/changes#r3010523197) comment:

```suggestion
      migrate-db:
        condition: service_completed_successfully
```

docker-compose.yml
```diff
@@ -20,57 +52,39 @@ services:
...
-      USE_POSTGRES_FOR_ANALYTICS: 'true'
-      DJANGO_ALLOWED_HOSTS: '*'
-      PROMETHEUS_ENABLED: 'true'
+      <<: *flagsmith-env
     ports:
       - 8001:8000
     depends_on:
-      - flagsmith
+      init-db:
+        condition: service_completed_successfully
```
khvn26: Based on [prior](https://github.com/Flagsmith/flagsmith/pull/7068/changes#r3010523197) comment:

```suggestion
      migrate-db:
        condition: service_completed_successfully
```

khvn26 (dismissed):
LGTM, but I'd rename the new container.

docker-compose.yml
```diff
@@ -20,57 +52,39 @@ services:
...
-      # Enable Task Processor
-      TASK_RUN_METHOD: TASK_PROCESSOR # other options are: SYNCHRONOUSLY, SEPARATE_THREAD (default)
-      PROMETHEUS_ENABLED: 'true'
-
-      # Uncomment if you want to enable Google OAuth. Note this does not turn Google OAuth on. You still need to use
-      # Flagsmith on Flagsmith to enable it - https://docs.flagsmith.com/deployment/#oauth_google
-      # DJANGO_SECURE_CROSS_ORIGIN_OPENER_POLICY: 'same-origin-allow-popups'
+      <<: *flagsmith-env
+    healthcheck:
+      test: ["NONE"]
```
emyller: Nit

```suggestion
    healthcheck:
      disable: true
```

matthewelwell: My IDE complained that the schema did not think this was valid ...

matthewelwell: Looks like it works just the same, I've updated it.

khvn26 (approved):

emyller (approved):

PR #7079, opened 2026-03-30
fix: Version endpoints not returning JSON in all-in-one Docker deployment

### LINKED ISSUE #7069: Version endpoints not returning JSON in all-in-one Docker deployment
## Summary

The `getBuildVersion` query in the frontend (`common/services/useBuildVersion.ts`) fails to retrieve JSON from the version endpoints in the all-in-one Docker deployment. This means `version.backend.self_hosted_data.has_users` is never `false`, so:

 1. The onboarding page is never shown on a fresh install — users land on the login page with no way to create an account.
 2. The licensing tab is not shown in the organisation settings for an enterprise deployment

## Deployment scenarios

Reviewing all deployment scenarios, it looks as though the issue only currently affects the All-in-one image approach, but we need to be careful to validate that any changes support all deployment options. 

### 1. All-in-one image (Django serves frontend + API assets) ❌

Django's settings default `API_URL = "/api/v1/"`, which is surfaced via the `/config/project-overrides` endpoint as `api: '/api/v1/'`. This causes `useBuildVersion.ts` to use `/_backend_version` as the backend version URL.

However, `/_backend_version` is **only registered as a route in the Express.js frontend server** (`api/index.js`), and only when `FLAGSMITH_PROXY_API_URL` is set. Since this endpoint doesn't exist in Django, the catch-all (`re_path(r"^.*$", views.index)`) intercepts it and returns the frontend HTML page.

The frontend also still tries to get its own version (via `data.get('/version')`) but this ends up hitting the catch-all on the API as well because it's missing the trailing slash. This doesn't matter too much, but the frontend version (`ci_commit_sha`, `image_tag`) would show as `Unknown`. Since the all-in-one image contains both frontend and backend though, the backend's `/version/` already contains the same build info.

---

### 2. Separate frontend and API (no proxy) ✓

- `FLAGSMITH_API_URL` is set to the full backend URL (e.g. `https://api.example.com/api/v1/`)
- `Project.api` = full URL → `Project.api !== '/api/v1/'`
- Backend version URL: `${Project.api.replace('api/v1/', '')}version/` = `https://api.example.com/version/` → Django JSON ✓
- Frontend version URL: `/version` → Express `app.get('/version', ...)` → JSON ✓

This case works correctly.

---

### 3. Separate frontend and API with frontend proxying API traffic ✓

- `FLAGSMITH_PROXY_API_URL` is set
- Express proxies `/api/v1/` to the backend and registers `/_backend_version`
- `projectOverrides.api = '/api/v1/'` (set explicitly by Express when `FLAGSMITH_PROXY_API_URL` is set)
- Backend version URL: `/_backend_version` → Express proxies to `${FLAGSMITH_PROXY_API_URL}/version/` → Django JSON ✓
- Frontend version URL: `/version` → Express `app.get('/version', ...)` → JSON ✓

This case works correctly.

---

## Root cause

The condition `Project.api === '/api/v1/'` is used in `useBuildVersion.ts` as the signal for "use `/_backend_version`", but it's true in **both** the Express proxy case (Case 3, where `/_backend_version` exists) and the all-in-one Django case (Case 1, where it doesn't).

The `/_backend_version` endpoint is an Express-only construct, gated on `FLAGSMITH_PROXY_API_URL`. It was never wired into Django.

matthewelwell:
## Changes

Closes https://github.com/Flagsmith/flagsmith/issues/7069

Uses a new attribute in projectOverrides to expose whether the FE proxy is enabled, instead of relying on the API URL being `/api/v1/` which is true in the case where the FE proxy is enabled, but also in the all-in-one image. This ensures that the FE calls the BE correctly to get the version endpoint. 

Note: what this still doesn't solve is the FE trying to get its own version in the all-in-one deployment scenario, but since it's a handled case, and the FE doesn't really _have_ a version in this scenario, it's not really worth worrying about. 

Ran Flagsmith in all 3 deployment setups using docker compose and verified that the version endpoint is returned correctly.  

- [x] All-in-one
- [x] Separate FE / API no proxy
- [x] Separate FE / API with FE proxy

### DIFF
--- a/frontend/api/index.js
+++ b/frontend/api/index.js
@@ -76,6 +76,10 @@ app.get('/config/project-overrides', (req, res) => {
         ? '/api/v1/'
         : process.env.FLAGSMITH_API_URL,
     },
+    {
+      name: 'apiProxyEnabled',
+      value: !!process.env.FLAGSMITH_PROXY_API_URL,
+    },
     { name: 'maintenance', value: envToBool('ENABLE_MAINTENANCE_MODE', false) },
     {
       name: 'flagsmithClientAPI',
--- a/frontend/common/services/useBuildVersion.ts
+++ b/frontend/common/services/useBuildVersion.ts
@@ -13,10 +13,9 @@ export const buildVersionService = service
         providesTags: () => [{ id: 'BuildVersion', type: 'BuildVersion' }],
         queryFn: async (args, _, _2, baseQuery) => {
           try {
-            const backendVersionUrl =
-              Project.api === '/api/v1/'
-                ? '/_backend_version'
-                : `${Project.api.replace('api/v1/', '')}version/`
+            const backendVersionUrl = Project.apiProxyEnabled
+              ? '/_backend_version'
+              : `${Project.api.replace('api/v1/', '')}version/`
             const [frontendRes, backendRes] = await Promise.all([
               data.get(`/version`).catch(() => ({})),
               data.get(backendVersionUrl),


talissoncosta (approved):

Zaimwa9 (approved):
3 scenarios ✅ 

Thanks for going to the bottom of this one

PR #7088, opened 2026-03-31
fix(account-dropdown):  fix contrast issues in dark mode

### LINKED ISSUE #6902: Fix Account dropdown trigger contrast in dark mode
## Problem

The "Account" button in the top navbar uses the \`btn-link\` class, which renders as the primary brand colour (\`#6837FC\`) on the dark navbar background (\`~#1A2034\`). This gives approximately **2.3:1 contrast ratio**, well below the WCAG AA minimum of 4.5:1.

**Location**: \`web/components/navigation/AccountDropdown.tsx\` line 59

Image

## Acceptance Criteria

- [ ] "Account" text meets WCAG AA contrast (4.5:1) against the dark navbar

---
Part of the Design System Audit (#6606) · Epic: #6882

talissoncosta:

## Changes

Closes #6902

Replace `btn btn-link` with scoped `.account-dropdown-trigger` styles using semantic design tokens:
- Uses `--color-text-default` for theme-aware text colour (dark in light mode, white in dark mode)
- Purple hover state (`--color-text-action`) on label text only, not the icon
- Removes dependency on Bootstrap button classes that forced low-contrast brand purple

Manually verified in both light and dark mode:
- "Account" text is readable against the navbar background in both themes
- Hover state turns only the "Account" label purple, icon remains unchanged
- Button behaviour (click to open dropdown) unchanged

### DIFF
--- a/frontend/web/styles/project/_project-nav.scss
+++ b/frontend/web/styles/project/_project-nav.scss
@@ -13,6 +13,17 @@ nav a {
     font-weight: bold;
   }
 }
+.account-dropdown-trigger {
+  background: none;
+  border: none;
+  cursor: pointer;
+  font-weight: 500;
+
+  &:hover,
+  &:focus {
+    color: var(--color-text-action);
+  }
+}
 .nav-sub-link-disabled {
   cursor: not-allowed;
   span {
--- a/frontend/web/components/navigation/AccountDropdown.tsx
+++ b/frontend/web/components/navigation/AccountDropdown.tsx
@@ -56,7 +56,7 @@ const AccountDropdown: React.FC = () => {
   return (
     <div className='feature-action' tabIndex={-1}>
       <button
-        className='btn btn-link p-0 d-flex ps-3 lh-1 align-items-center'
+        className='account-dropdown-trigger d-flex ps-3 lh-1 align-items-center text-default'
         onClick={(e) => {
           e.stopPropagation()
           setIsOpen(!isOpen)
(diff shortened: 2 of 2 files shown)


kyle-ssg (dismissed):

frontend/web/styles/project/_project-nav.scss
```diff
@@ -13,6 +13,20 @@ nav a {
     font-weight: bold;
   }
 }
+.account-dropdown-trigger {
```
kyle-ssg, line 16: Hmm a lot of this css seems to be too custom vs just adding p-0 / fw-normal

kyle-ssg:
Approved but just some general feedback

frontend/web/styles/project/_project-nav.scss
talissoncosta: Good call — agreed. Simplified to use token utility classes (`.text-default`, `.icon-secondary`) for colours and Bootstrap classes for layout. Custom CSS is now limited to the button reset and hover state. I'll prioritise existing classes over custom CSS going forward. Also removed the hardcoded icon fill in favour of the `icon-secondary` token. Fixed in `0ac798bb5`.

kyle-ssg: Yeah I thought I'd mention it since the overall aim of design tokens / chromatic etc is to standardise, so if anything we should have less css that looks like this and classnames that look like this too

kyle-ssg (approved):

PR #7099, opened 2026-04-01
fix: add null validation for environment in FeatureStateSerializerBasic

Pwongchaiya:
## Summary
- Added validation in `FeatureStateSerializerBasic.validate_environment` to raise a `ValidationError` when the environment is `None`, preventing a 500 error response.

## Test plan
- [ ] Verify that sending a request with a null environment returns a 400 with a descriptive error message instead of a 500.
- [ ] Verify existing environment validation (cannot change environment of a feature state) still works correctly.

### DIFF
--- a/api/features/serializers.py
+++ b/api/features/serializers.py
@@ -588,7 +588,7 @@ class FeatureStateSerializerBasic(WritableNestedModelSerializer):
 
         except django.core.exceptions.ValidationError as e:
             raise serializers.ValidationError(str(e))
-        
+
     def validate_feature(self, feature):  # type: ignore[no-untyped-def]
         if self.instance and self.instance.feature_id != feature.id:  # type: ignore[union-attr]
             raise serializers.ValidationError(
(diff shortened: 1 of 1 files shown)


gagantrivedi (commented):
Thanks for the fix — this is a real bug and worth addressing! A couple of suggestions:

**Consider overriding the field instead of adding a validator check**

The `environment` FK on the model has `null=True`, so DRF auto-generates the serializer field with `allow_null=True`. Since this serializer always expects a real environment (the `validate()` method falls back to `self.context["environment"]`), you could disallow null at the field level instead:

```python
class FeatureStateSerializerBasic(WritableNestedModelSerializer):
    environment = serializers.PrimaryKeyRelatedField(
        queryset=Environment.objects.all(),
        allow_null=False,
    )
```

This way DRF handles the `null` rejection automatically with a standard error message, and `validate_environment` only needs to deal with the "cannot change environment" logic.

**Tests needed**

The project requires 100% diff coverage for backend PRs, so this will need at least one test — e.g. sending `"environment": null` and asserting a 400 response.

gagantrivedi (changes requested):
Thanks for the fix — this is a real bug and worth addressing! A couple of suggestions:

Consider overriding the field instead of adding a validator check

The environment FK on the model has null=True, so DRF auto-generates the serializer field with allow_null=True. Since this serializer always expects a real environment (the validate() method falls back to self.context["environment"]), you could disallow null at the field level instead:
```
class FeatureStateSerializerBasic(WritableNestedModelSerializer):
    environment = serializers.PrimaryKeyRelatedField(
        queryset=Environment.objects.all(),
        allow_null=False,
    )
```
This way DRF handles the null rejection automatically with a standard error message, and validate_environment only needs to deal with the "cannot change environment" logic.

Tests needed

The project requires 100% diff coverage for backend PRs, so this will need at least one test — e.g. sending "environment": null and asserting a 400 response.

Issue #6761, opened 2026-02-23
Spike: Project-scoped export/import


khvn26:
Currently, we allow environment-level import and export (default feature states only) accessible from UI, and CLI-based organisation-level import and export (projects, segments, features, environments and the segment/identity overrides).

We want to explore offering project-scoped exports that include segments and overrides, and allowing to create and import them via the UI.

The outcome of this spike should be an epic detailing out the implementation for the above.

khvn26:
cc @matthewelwell @asaphko

The outcome of this spike is the epic #7125 with 5 sub-issues and a docs PR #7114.

We need to decide whether project-level import/export should be gated to a paid plan (Scale-Up+). Project-level export (with segments, overrides, environment settings) is primarily useful for cloning projects and migrating between instances, which skews toward teams and enterprise. The existing env-level export (default feature states only) remains free.

## Notes from the spike

- The `created_by` / `created_by_api_key` dual FK pattern on `ProjectExport` and `ProjectImport` parrots the existing approach used by `EnvironmentFeatureVersion` and other models. This is not ideal, but #5978 (Spike: Improve user types) is still pending. Once that lands, we should revisit and adopt whatever unified author tracking it introduces.
- We chose DRF serializer subclasses with `exclude` over two alternatives:
  - Django `serializers.serialize` (used by org-level export): auto-discovers model fields via natural keys, but most relevant models lack `natural_key()` methods (Segment, Tag, Feature, FeatureState, FeatureSegment). Those that have them use UUIDs or `api_key` which aren't portable across instances. `loaddata` has no merge logic. The fixture format is verbose and coupled to internal model paths.
  - Hand-rolled dicts (used by env-level export): every time a field is added to a model, someone must remember to update the export code. This is the maintenance problem we're solving.
- Identity override import is a separate issue from the core import because: it's opt-in, involves a fundamentally different code path (DynamoDB for edge projects vs SQL for non-edge), can be massive (millions of overrides), the core import delivers full value without it, and it runs as a separate async task that can fail independently. The frontend toggle is gated behind a feature flag until that issue lands.
- Custom fields/metadata and integration configurations were dropped from the first pass. Neither has a defined export schema or serializer support. They can be added as follow-up issues once the core export/import is stable and the demand is clear.

## Provisioning convergence

[...]

Issue #5489, opened 2025-05-25
Duplicate segment names allowed, causing confusion during linking


sbhrwlr:
### How are you running Flagsmith

- [x] Self Hosted with Docker
- [ ] Self Hosted with Kubernetes
- [x] SaaS at flagsmith.com
- [ ] Some other way (add details in description below)

### Describe the bug

Currently, Flagsmith allows creating multiple segments with the same name within a project. This can lead to confusion when users link segments to features or environments, as they cannot easily distinguish between segments with identical names.

### Impact
- Increases risk of linking the wrong segment  
- Degrades user experience and clarity  
- Makes auditing and managing segments harder

### Suggested Fix
- Enforce unique segment names per project  
**OR**  
- Display disambiguating information (e.g., ID, created date) when listing segment options

### Steps To Reproduce

1. Create a segment with a name (e.g., `beta_users`).
2. Create another segment with the same name.
3. Attempt to link a segment to a feature or environment.

### Expected behavior

Users should be warned or restricted from creating duplicate segment names to prevent confusion.

### Screenshots

Duplicate Segments in Segment Page:
<img width="1336" alt="Image" src="https://github.com/user-attachments/assets/11cd5dc6-dd75-47fe-8cda-aeee944184d4" />

Duplicate Names in Feature Segment Override Section causing confusion"
<img width="899" alt="Image" src="https://github.com/user-attachments/assets/185e5c2f-a8b3-4ba3-9ba5-fd554e125e43" />

khvn26:
We need this done before we can introduce segment-related Context values as well.

asaphko:
When working on this, we should not worry about migrations and focus on future usage

PR #7114, opened 2026-04-02
docs: Add project-level import/export specification

khvn26:

## Changes

Contributes to #6761

In this PR, we document the planned project-scoped export/import feature alongside the existing environment-level bulk import/export. This serves as the specification before implementation begins.

Key changes to `bulk-import-and-export.md`:
- Extract shared concepts (merge strategy, retention) into a common section to avoid repetition
- Restructure existing environment-level docs under a clear heading
- Add new project-level export section documenting:
  - What's always included
  - What's optionally included  (identity overrides, tags, custom fields, integrations)
  - What's never included
  - Export format compatibility with environment-level exports
  - Versioning behaviour on import (new versions created in Overwrite Destructive mode)
  - Environment creation behaviour on import

`import-and-export.md` index page now references both levels.

### Preview links

- [Bulk Import & Export](https://docs-git-docs-project-scoped-import-export-flagsmith.vercel.app/administration-and-security/data-management/bulk-import-and-export) (main changed page)
- [Import and Export index](https://docs-git-docs-project-scoped-import-export-flagsmith.vercel.app/administration-and-security/data-management/import-and-export)

Documentation-only change. Verified formatting with `npx prettier --check` and Vercel preview build.

### DIFF
--- a/docs/docs/administration-and-security/data-management/bulk-import-and-export.md
+++ b/docs/docs/administration-and-security/data-management/bulk-import-and-export.md
@@ -154,7 +154,9 @@ overwritten, all of its rules, conditions, and associated feature overrides are 
 All entities are matched by name across instances, not by internal ID. This means exports are portable between Flagsmith
 instances. Specifically:
 
-- Features and segments are matched by name.
+- Features are matched by name (unique per project).
+- Segments are matched by name. If duplicates exist in the target project, the first match is used. If the export
+  contains duplicate segment names, all are imported.
 - Tags are matched by label.
 - Environments are matched by name. Missing environments are created with settings from the export. For existing
   environments, the Skip strategy leaves settings unchanged, while Overwrite Destructive updates settings to match the
--- a/docs/docs/administration-and-security/data-management/bulk-import-and-export.md
+++ b/docs/docs/administration-and-security/data-management/bulk-import-and-export.md
@@ -6,8 +6,8 @@ sidebar_position: 30
 
 Flagsmith supports two levels of bulk import and export from the UI:
 
-- **[Environment-level](#environment-level-export)**: export and import default feature states for a single environment.
-  Useful for quickly copying feature values between environments or projects.
+- **[Environment-level](#environment-level-export)**: export features with one environment's values. Useful for copying
+  feature configuration between environments or projects.
 - **[Project-level](#project-level-export)**: export and import an entire project's feature configuration, including
   segments, segment overrides, and optionally identity overrides and tags. Useful for cloning a project or migrating
   between Flagsmith instances.
--- a/docs/docs/administration-and-security/data-management/bulk-import-and-export.md
+++ b/docs/docs/administration-and-security/data-management/bulk-import-and-export.md
@@ -104,14 +104,17 @@ The following data can optionally be included in the export. You will be prompte
 | Custom fields and metadata | Off     |                                                           |
 | Integration configurations | Off     | Includes third-party integration settings per environment |
 
-The following data is **never** included in a project-level export:
+:::caution
 
+The following data is **never** included in a project-level export:
 - Users, owners, and group owners
 - Change requests and approvals
 - Scheduled flag changes
 - Version history (only the current live state is exported)
 - Audit logs
 - Flag analytics
+
+:::
 
 ### Export format
 
--- a/docs/docs/administration-and-security/data-management/bulk-import-and-export.md
+++ b/docs/docs/administration-and-security/data-management/bulk-import-and-export.md
@@ -9,8 +9,8 @@ Flagsmith supports two levels of bulk import and export from the UI:
 - **[Environment-level](#environment-level-export)**: export and import default feature states for a single environment.
   Useful for quickly copying feature values between environments or projects.
 - **[Project-level](#project-level-export)**: export and import an entire project's feature configuration, including
-  segments, segment overrides, and optionally identity overrides, tags, and custom fields. Useful for cloning a project
-  or migrating between Flagsmith instances.
+  segments, segment overrides, and optionally identity overrides and tags. Useful for cloning a project or migrating
+  between Flagsmith instances.
 
 For exporting an entire organisation (including all projects), see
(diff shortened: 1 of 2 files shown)


docs/docs/administration-and-security/data-management/bulk-import-and-export.md
```diff
@@ -4,9 +4,40 @@ sidebar_label: Bulk Import & Export
...
 
-The bulk import and export of feature data associated with a given environment is possible in Flagsmith. The feature data that is exported includes multivariate features, but does not include other data such as tags, owners, group owners, etc. This functionality is useful for transferring features between any running instances of Flagsmith, even when only a subset of features (for example, importing features of a given tag) is needed.
+Flagsmith supports two levels of bulk import and export from the UI:
 
-## What is exported?
+- **[Environment-level](#environment-level-export)**: export and import default feature states for a single environment.
+  Useful for quickly copying feature values between environments or projects.
+- **[Project-level](#project-level-export)**: export and import an entire project's feature configuration, including
+  segments, segment overrides, and optionally identity overrides and tags. Useful for cloning a project or migrating
+  between Flagsmith instances.
```
matthewelwell, line 13: While I think we can get away with this separation, it's not 100% correct I don't think. 

What we're terming the 'Environment-level' import here does include information from the project. It was built as a 'Feature' export, so it includes information about the default value of the flag. As I understand it, this is what is used in a _destructive_ import - it will overwrite other environments with the default value. 

Perhaps the scope of this documentation update also covers reviewing and improving the 'Environment-level' import if we can, since this behaviour is not desirable anyway...

docs/docs/administration-and-security/data-management/bulk-import-and-export.md
```diff
@@ -23,28 +54,115 @@ We **will not** export the following data:
...
+Overwrite Destructive merge strategy.
 
-:::tip
-Feature exports are available for only two weeks, so if an export is needed for a longer period, be sure to make a local backup.
 :::
 
-## Importing
+## Project-level export
+
+<!-- TODO: confirm plan availability -->
```
matthewelwell: As discussed on the standup, we agreed to keep this available on all plans.

docs/docs/administration-and-security/data-management/bulk-import-and-export.md
```diff
@@ -23,28 +54,115 @@ We **will not** export the following data:
...
+When using Overwrite Destructive, segments and their overrides are treated as a single unit: if a segment is
+overwritten, all of its rules, conditions, and associated feature overrides are replaced.
+
+#### How entities are matched
+
+All entities are matched by name across instances, not by internal ID. This means exports are portable between Flagsmith
+instances. Specifically:
 
-Since a feature may have an identical name, it is important to carefully select a merge strategy during a feature import.
+- Features and segments are matched by name.
```
matthewelwell, line 157: We should be careful here - I have a feeling we don't enforce unique segment names?

docs/docs/administration-and-security/data-management/bulk-import-and-export.md
khvn26: Does 710d34ecf make sense to you?

Also, to clarify, do you think we should amend the 'Environment-level' import behaviour as part of the epic?

docs/docs/administration-and-security/data-management/bulk-import-and-export.md
khvn26: Removed in 23140b8 👍

docs/docs/administration-and-security/data-management/bulk-import-and-export.md
khvn26: You're right, segment names aren't unique. Clarified in 539906d35, also added this to the #7127 scope.

docs/docs/administration-and-security/data-management/bulk-import-and-export.md
matthewelwell: > Also, to clarify, do you think we should amend the 'Environment-level' import behaviour as part of the epic?

I would love to. 

The ideal scenario here in my head is leave the documentation as you had it, and 'fix' the current behaviour which affects other environments on import. That being said, I can't remember why we had to implement it that way in the first place. I do think there was a technical reason for having to affect other environments.

docs/docs/administration-and-security/data-management/bulk-import-and-export.md
matthewelwell: I wonder if we should again just swallow [more scope](https://github.com/Flagsmith/flagsmith/issues/5489) as part of this work...

docs/docs/administration-and-security/data-management/bulk-import-and-export.md
khvn26: Filed it as a bug here https://github.com/Flagsmith/flagsmith/issues/7166, and reverted the docs fixup. Up to us to add it to the epic, but since there's no dependencies, I'm keeping it separate for now.

docs/docs/administration-and-security/data-management/bulk-import-and-export.md
khvn26: The dependency is only there because we added a clarification to the docs, I think?

Come to think of it the "first matched" behaviour shouldn't mean any special code really.

matthewelwell: I guess our options are: 

 1. Keep the docs change added in 539906d which reflects the behaviour if we do nothing
 2. Revert the docs change added in 539906d and bring #5489 into the relevant epic
 3. Revert the docs change added in 539906d and just accept that there's a bug as per #5489 

IMO, since #5489 is fairly trivial, I'd like to go with option 2 here, but I'm happy to go with any of the options.

khvn26: I feel like 3 gives us most planning flexibility.

Reverted in d3fd06b52 and removed from #7127. Once #5489 is resolved, matching by name just works.

PR #7116, opened 2026-04-03
fix: include specific IDs in owner validation errors

gagantrivedi:

## Changes

Addresses 2 nit comments from @khvn26 on #7067:

- **`validate_owners`**: error message now includes the specific user IDs that lack project access (e.g. `"Users with ids [3, 7] do not have access to this project."`) instead of the generic `"Some users do not have access to this project."`
- **`validate_group_owners`**: error message now includes the specific group IDs that belong to the wrong organisation (e.g. `"Groups with ids [5] do not belong to this project's organisation."`) instead of the generic `"Some groups do not belong to this project's organisation."`

Both validators now collect all invalid entries before raising, so a single error reports every problematic ID at once.

- Updated existing tests (`test_create_feature__group_owner_from_different_org__returns_400` and `test_create_feature__owner_without_project_access__returns_400`) to assert the specific IDs appear in the error response.

### DIFF
--- a/api/features/serializers.py
+++ b/api/features/serializers.py
@@ -306,22 +306,32 @@ class CreateFeatureSerializer(DeleteBeforeUpdateWritableNestedModelSerializer):
 
     def validate_owners(self, owners: list[FFAdminUser]) -> list[FFAdminUser]:
         project: Project = self.context["project"]
-        for user in owners:
-            if not user.has_project_permission(VIEW_PROJECT, project):
-                raise serializers.ValidationError(
-                    "Some users do not have access to this project."
-                )
+        invalid_users = [
+            user
+            for user in owners
+            if not user.has_project_permission(VIEW_PROJECT, project)
+        ]
+        if invalid_users:
+            invalid_user_ids = [user.id for user in invalid_users]
+            raise serializers.ValidationError(
+                f"Users with ids {invalid_user_ids} do not have access to this project."
+            )
         return owners
 
     def validate_group_owners(
         self, group_owners: list[UserPermissionGroup]
     ) -> list[UserPermissionGroup]:
         project: Project = self.context["project"]
-        for group in group_owners:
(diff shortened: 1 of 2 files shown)


api/features/serializers.py
```diff
@@ -306,22 +306,32 @@ def validate_multivariate_options(self, multivariate_options):  # type: ignore[n
...
-                raise serializers.ValidationError(
-                    "Some users do not have access to this project."
-                )
+        invalid_users = [
+            user
+            for user in owners
+            if not user.has_project_permission(VIEW_PROJECT, project)
+        ]
+        if invalid_users:
+            invalid_user_ids = [user.id for user in invalid_users]
```
khvn26, line 315: maybe emails or uuids?

api/features/serializers.py
```diff
@@ -306,22 +306,32 @@ def validate_multivariate_options(self, multivariate_options):  # type: ignore[n
...
-                raise serializers.ValidationError(
-                    "Some groups do not belong to this project's organisation."
-                )
+        invalid_groups = [
+            group
+            for group in group_owners
+            if group.organisation_id != project.organisation_id
+        ]
+        if invalid_groups:
+            invalid_group_ids = [group.id for group in invalid_groups]
```
khvn26, line 331: maybe group names?

khvn26 (commented):
Trying to understand which error messages would be more helpful to the user.

api/features/serializers.py
gagantrivedi: Returning email would be a data leak and returning uuid when you have supplied id is even more confusing?

api/features/serializers.py
gagantrivedi: Same as above — I don't think it makes sense to output different params than the input.

khvn26 (approved):

PR #7146, opened 2026-04-07
fix: return null instead of empty string when clearing feature value

### LINKED ISSUE #6818: Removing a feature value resets it to empty string instead of null

- [ ] Self Hosted with Docker
- [x] SaaS at flagsmith.com

### Describe the bug

Creating a feature without a value sets the value to `None`.
When removing an existing value from a feature, the value is set to `""`

### Steps To Reproduce

1. Create a feature without value
2. Set a string value
3. Remove the value

### Expected behavior

The value should be set back to `None`

### Screenshots

_No response_

nielskaspers:
## Summary
When removing a feature value, the value is set to an empty string `""` instead of `null`. This PR fixes the `valueToFeatureState` and `valueToTrait` utility functions to return `null` when the string value is empty.

## Issue
Fixes #6818

## Changes
- In `valueToFeatureState`: change `value === null ? null : val || ''` to `value === null || val === '' ? null : val`
- In `valueToTrait`: apply the same fix for consistency

The previous expression `val || ''` always produced `''` when `val` was an empty string (falsy), preventing the value from being set back to `null`. The new condition explicitly checks for empty string and returns `null`, matching the initial state of a feature created without a value.

## Testing
1. Create a feature without a value — value should be `null`
2. Set a string value (e.g., "hello")
3. Clear the value (remove all text)
4. The value should now be `null` (not `""`)

### DIFF
--- a/frontend/common/stores/feature-list-store.ts
+++ b/frontend/common/stores/feature-list-store.ts
@@ -380,15 +380,17 @@ const controller = {
                 )
               environmentFlag.multivariate_feature_state_values =
                 multivariate_feature_state_values
+              const typedValue = Utils.getTypedValue(
+                flag.initial_value,
+                undefined,
+                true,
+              )
               return data.put(
                 `${Project.api}environments/${environmentId}/featurestates/${environmentFlag.id}/`,
                 Object.assign({}, environmentFlag, {
                   enabled: flag.default_enabled,
-                  feature_state_value: Utils.getTypedValue(
-                    flag.initial_value,
-                    undefined,
-                    true,
-                  ),
+                  feature_state_value:
+                    typedValue === '' ? null : typedValue,
                 }),
               )
             })
@@ -805,7 +807,8 @@ const controller = {
           ).then((res) => {
             const data = Object.assign({}, environmentFlag, {
               enabled: flag.default_enabled,
-              feature_state_value: flag.initial_value,
+              feature_state_value:
+                flag.initial_value === '' ? null : flag.initial_value,
             })
             return createAndSetFeatureVersion(getStore(), {
               environmentId: res,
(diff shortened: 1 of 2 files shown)


talissoncosta (dismissed):
LGTM — nice catch! Clean fix with a clear explanation. Thanks for the contribution 👍

Zaimwa9 (changes requested):
Hey! thanks for the help here. 

So, I took a look inside and although this fix works well for code paths that use the nested format like segment overrides or change requests. It needs an extra change to properly close the issue.

The idea is to send `null` in the PUT update to the backend `/api/v1/environments/{env}/featurestates/{feature_state_id/` for v1 versioning environments.

However, the main update path for environment feature states in `feature-list-store.ts` doesn't go through `valueToFeatureState` but uses `Utils.getTypedValue()` directly (for v1) or passes `flag.initial_value` (for v2 versioned, line ~808). They send a flat feature_state_value in the PUT payload, so even with this fix they still send `""` instead of `null`.

To fully fix the issue, we need to handle the conversion in these two call sites in `editFeatureState` / `editVersionedFeatureState`.

And keep your current fix!

nielskaspers (commented):
Thanks for the thorough review @Zaimwa9! You're right — the two direct call sites in `feature-list-store.ts` were bypassing `valueToFeatureState`.

I've pushed a follow-up commit (ed66bc9) that adds empty-string-to-null conversion at both locations:

1. **`editFeatureState` (v1 path)** — extracted `Utils.getTypedValue()` into a variable and added `typedValue === '' ? null : typedValue`
2. **`editVersionedFeatureState` (v2 path)** — added `flag.initial_value === '' ? null : flag.initial_value`

Both the original `valueToFeatureState`/`valueToTrait` fix and these two call sites are now covered.

Zaimwa9 (approved):
Approving the code and will merge once tested on preview, thanks for the contribution 👍

PR #7147, opened 2026-04-07
fix(hubspot): simplify integration to only create contacts

Holmus:
The HubSpot lead tracking integration was doing three things:
1. Creating contacts (via form submission)
2. Creating/updating companies with Flagsmith org names - overwriting HubSpot-enriched data
3. Syncing subscription/plan changes to HubSpot companies

Problems with (2): Flagsmith org names like "ew", "Test", "Manejo-Org" overwrote enriched company names in HubSpot.

Problems with (3): This was redundant with our ChargeBee -> HubSpot sync (which runs daily and is the billing source of truth). Worse, it caused data confusion - trial subscriptions written by the integration were treated as paid accounts, triggering workflows that incorrectly set company type to Customer.

## Changes
- Simplify `create_lead` to only create the contact - HubSpot handles company creation and association from email domains
- Simplify `should_track` to just check if the feature is enabled
- Remove `update_company_active_subscription` and its AFTER_SAVE hook - ChargeBee sync is the source of truth for subscription data
- Remove `_get_organisation_hubspot_id` and `_get_hubspot_company_by_domain` (dead code after above)
- Remove associated tests

## What still works
- Contact creation via form submission
- HubSpot auto-creates and associates companies from contact email domains
- ChargeBee sync writes subscription, MRR, billing status, plan data daily
- `HubspotOrganisation` model is unchanged (existing records preserved)

Review effort: 1/5 (almost entirely deletions)

### DIFF
--- a/api/integrations/lead_tracking/hubspot/lead_tracker.py
+++ b/api/integrations/lead_tracking/hubspot/lead_tracker.py
@@ -5,11 +5,7 @@ from typing import Any
 from django.conf import settings
 
 from integrations.lead_tracking.lead_tracking import LeadTracker
-from organisations.models import (
-    HubspotOrganisation,
-    Organisation,
-    Subscription,
-)
+from organisations.models import Organisation
 from users.models import FFAdminUser, HubspotLead, HubspotTracker
 
 from .client import HubspotClient
@@ -22,25 +18,6 @@ class HubspotLeadTracker(LeadTracker):
     @staticmethod
     def should_track(user: FFAdminUser) -> bool:
         return settings.ENABLE_HUBSPOT_LEAD_TRACKING
-
-    def update_company_active_subscription(
-        self, subscription: Subscription
-    ) -> dict[str, Any] | None:
-        if not subscription.plan:
-            return None
-
-        organisation = subscription.organisation
-
-        # Check if we're missing the associated hubspot id.
-        if not getattr(organisation, "hubspot_organisation", None):
-            return None
-
-        response: dict[str, Any] | None = self.client.update_company(
-            active_subscription=subscription.plan,
-            hubspot_company_id=organisation.hubspot_organisation.hubspot_id,
-        )
-
-        return response
 
     def create_user_hubspot_contact(self, user: FFAdminUser) -> str | None:
         tracker = HubspotTracker.objects.filter(user=user).first()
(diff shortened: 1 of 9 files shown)


Zaimwa9 (changes requested):
So I get the overall intent of this PR. Although there is a side-effect that feels critical to me -> We are breaking the synchronization between flagsmith and hubspot (even in read-only) and we might be unable to update the subscription to Hubspot.

My understanding with this code is that hubspot will be completely in control of creating the organisation AND the association of the contacts. That makes sense. 

Right now, `_get_organisation_hubspot_id` is in charge of finding the organisation and creating the `HubspotOrganisation` record. It became dead code and we are never storing the information on our side.
It becomes a problem in `update_company_active_subscription` that is now a no-op (we don't have `hubspot_id`) and I feel using the domain or the name is too fragile.

**My proposal**
The fix should keep `_get_organisation_hubspot_id` in create_lead and drop the `name=` overwrite and keep writing `flagsmith_organisation_id` to HubSpot.
Optionally we can remove `associate_contact_to_company` to make our intent 100% clear and remove the associated tests (that just confirms it's never called right?).

Let me know what you think and i'll re-review with your thinking in mind.

-----------------------
If aligned, here is a final assist 😄 

**For Claude**
```
  1. In _get_organisation_hubspot_id, remove the name=organisation.name parameter from the self.client.update_company(...) call but keep hubspot_company_id and flagsmith_organisation_id.
  2. In create_lead, keep the call to _get_organisation_hubspot_id but remove  associate_contact_to_company and the early returns tied to it.
  3. Update/remove tests that assert associate_contact_to_company behaviour.
  4. Remove any now-unused imports (e.g. re2/re if domain filtering stays removed).
```

Holmus:
Thanks for the review @Zaimwa9 - good catch on update_company_active_subscription becoming a no-op.

After thinking through it more, I went further than your proposal: I removed update_company_active_subscription entirely rather than fixing it.

Why: We have a ChargeBee -> HubSpot sync that runs daily and writes subscription data (plan, MRR, billing status, contract dates) to HubSpot companies. It's the billing source of truth. Having the Flagsmith integration also write subscription data was:

Redundant - ChargeBee sync already handles this
Harmful - the integration can't distinguish trial vs paid. We found cases where trial subscription data triggered workflows to incorrectly set companies as paying customers
So the integration now only does one thing: create the contact. HubSpot handles company creation from email domains, and ChargeBee sync handles all subscription/billing data.

The new commit removes update_company_active_subscription, its task, the AFTER_SAVE hook, and associated dead code (_get_organisation_hubspot_id, _get_hubspot_company_by_domain).

api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -5,66 +5,19 @@
...
 
 class HubspotLeadTracker(LeadTracker):
     @staticmethod
     def should_track(user: FFAdminUser) -> bool:
-        if not settings.ENABLE_HUBSPOT_LEAD_TRACKING:
-            return False
-
-        domain = user.email_domain
-
-        if settings.HUBSPOT_IGNORE_DOMAINS_REGEX and re.match(
```
Zaimwa9, line 37: Let's clean those `HUBSPOT_XXX` from the settings please

api/tests/unit/integrations/lead_tracking/hubspot/test_unit_hubspot_lead_tracking.py
```diff
@@ -165,16 +157,10 @@ def test_track_hubspot_lead_v2__new_user_added_to_org__creates_associations(
...
-    assert organisation.hubspot_organisation.hubspot_id == HUBSPOT_COMPANY_ID
 
+    # create_lead only creates the contact, not the company association
     mock_client_existing_contact.create_company.assert_not_called()
-
-    mock_client_existing_contact.associate_contact_to_company.assert_called_once_with(
-        contact_id=HUBSPOT_USER_ID,
-        company_id=HUBSPOT_COMPANY_ID,
-    )
+    mock_client_existing_contact.associate_contact_to_company.assert_not_called()
```
Zaimwa9: This test name should reflect that associations are not created.

(Atm: `test_create_lead__existing_hubspot_org__creates_contact_and_associates`)

Zaimwa9: Actually, `associate_contact_to_company` is only used in test now so we might want to clean it all up

Zaimwa9 (changes requested):
Sorry to be annoying. Couple of NITs that we should take care of now:
- Code removal 
- Unused variables

Then i'll be happy to approve assuming all ✅  on the functionality

Zaimwa9:
@Holmus do you want me to take over this one ?

Zaimwa9 (approved):
Thanks! 🥇

PR #7192, opened 2026-04-09
ci: add preview smoke test for frontend PRs

### LINKED ISSUE #7262: CI: Add Vercel preview smoke test to frontend PR workflow
### Context

We don't currently have automated verification that Vercel preview deployments are healthy after a frontend PR push. Build failures, broken Express server configs, or misconfigured environment variables can ship to preview without anyone noticing until a reviewer manually opens the link.

### Proposal

Add a lightweight smoke test to the frontend PR workflow that checks Vercel preview deployments after each push. The test should:

- Verify the Express server is responding (`/config/project-overrides`)
- Verify static files are served (`/`)
- Be non-blocking (warn but don't gate merge)
- Handle Vercel's "Ignored Build Step" (no frontend changes → previous build still live)

### Implementation

PR #7192 implements this by parsing preview URLs from the Vercel bot PR comment and curling key endpoints.

### Notes

- The test is intentionally non-blocking (`continue-on-error: true`) to avoid false positives blocking PRs

talissoncosta:

## Changes

Contributes to #7262

### Vercel Preview Smoke Test

Adds a smoke test job to the frontend PR workflow. On every PR that touches `frontend/**`, the job waits for Vercel to finish deploying, then verifies the preview is healthy.

### How it works

**Phase 1 — Wait for Vercel deployments:**
- Polls the commit status API for `flagsmith-frontend` checks (scoped to frontend only — won't wait for docs)
- Waits until all leave `pending` (real builds: waits for completion; skipped builds: immediately done)
- If a build fails, reports the failure and skips testing

**Phase 2 — Smoke test the preview:**
- Parses `flagsmith-frontend-*` URLs from the Vercel bot PR comment (short poll with 30s buffer for comment update timing)
- `GET /config/project-overrides` — Express server is running
- `GET /` — static files deployed + response body contains "Flagsmith" (content validation)

### Properties

- **No false positives** — Phase 1 ensures the current build has landed before testing
- **Non-blocking** — `continue-on-error: true`, failures warn but don't gate merge
- **Bounded** — `timeout-minutes: 15` on the job, `--max-time 30` on each curl
- **Handles skipped builds** — Vercel's "Ignored Build Step" completes immediately; previous preview URL is still live
- **No checkout needed** — uses only `gh api` + `curl`

### Review feedback addressed

- Job-level timeout
- Curl timeouts (`--max-time 30`)
- Non-blocking (`continue-on-error: true`)
- Merged duplicate API calls
- Dropped `/health` (behind VPN, returns 404 on public URLs)
- Body content validation on `/`

- Validated on a throwaway test PR (#7264) with a real Vercel deployment
- Previous iteration caught a false-positive issue — smoke test passed before Vercel finished deploying. Fixed with the two-phase approach.

### DIFF
--- a/.github/workflows/frontend-pull-request.yml
+++ b/.github/workflows/frontend-pull-request.yml
@@ -40,6 +40,7 @@ jobs:
     name: Vercel Preview Smoke Test
     runs-on: ubuntu-latest
     timeout-minutes: 10
+    continue-on-error: true
     steps:
       - name: Wait for Vercel deployments and smoke test
         env:
@@ -51,6 +52,8 @@ jobs:
             "Preview – flagsmith-frontend-preview"
             "Preview – flagsmith-frontend-staging"
           )
+
+          FAILED=0
 
           for ENV_NAME in "${ENVIRONMENTS[@]}"; do
             echo "⏳ Waiting for '$ENV_NAME' deployment..."
@@ -66,8 +69,9 @@ jobs:
                   echo "✅ '$ENV_NAME' deployed at: $URL"
                   break
                 elif [ "$STATE" = "failure" ] || [ "$STATE" = "error" ]; then
-                  echo "❌ '$ENV_NAME' deployment failed with state: $STATE"
-                  exit 1
+                  echo "⚠️ '$ENV_NAME' deployment failed with state: $STATE"
+                  FAILED=1
+                  continue 2
                 fi
               fi
               sleep 10
@@ -79,29 +83,22 @@ jobs:
             fi
 
             # Smoke test the deployment
-            echo "🔍 Testing $URL/health ..."
-            STATUS=$(curl -s -o /dev/null -w '%{http_code}' --max-time 30 "$URL/health")
-            if [ "$STATUS" != "200" ]; then
-              echo "❌ /health returned $STATUS"
-              exit 1
-            fi
+            for ENDPOINT in "/health" "/config/project-overrides" "/"; do
+              echo "🔍 Testing $URL$ENDPOINT ..."
+              STATUS=$(curl -s -o /dev/null -w '%{http_code}' --max-time 30 "$URL$ENDPOINT")
+              if [ "$STATUS" != "200" ]; then
+                echo "⚠️ $ENDPOINT returned $STATUS"
+                FAILED=1
+              fi
+            done
 
-            echo "🔍 Testing $URL/config/project-overrides ..."
-            STATUS=$(curl -s -o /dev/null -w '%{http_code}' --max-time 30 "$URL/config/project-overrides")
-            if [ "$STATUS" != "200" ]; then
-              echo "❌ /config/project-overrides returned $STATUS"
-              exit 1
-            fi
-
-            echo "🔍 Testing $URL/ ..."
-            STATUS=$(curl -s -o /dev/null -w '%{http_code}' --max-time 30 "$URL/")
-            if [ "$STATUS" != "200" ]; then
-              echo "❌ / returned $STATUS (this is the index page served by the Express server)"
-              exit 1
-            fi
-
-            echo "✅ All smoke tests passed for '$ENV_NAME'"
+            echo "✅ Smoke tests done for '$ENV_NAME'"
             echo ""
           done
+
+          if [ "$FAILED" -ne 0 ]; then
+            echo "⚠️ Some smoke tests failed — see above for details"
+            exit 1
+          fi
 
           echo "🎉 All Vercel preview deployments are healthy"
--- a/.github/workflows/frontend-pull-request.yml
+++ b/.github/workflows/frontend-pull-request.yml
@@ -53,7 +53,7 @@ jobs:
 
           for i in $(seq 1 60); do
             STATUSES=$(gh api "repos/$REPO/commits/$SHA/status" \
-              --jq '.statuses | [.[] | select(.context | test("Vercel"; "i"))]')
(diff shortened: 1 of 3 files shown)


.github/workflows/frontend-pull-request.yml
```diff
@@ -34,3 +35,72 @@ jobs:
...
+            echo "⏳ Waiting for '$ENV_NAME' deployment..."
+            URL=""
+            for i in $(seq 1 60); do
+              # Find the deployment for this commit and environment
+              DEPLOY_ID=$(gh api "repos/$REPO/deployments?sha=$SHA&environment=$(printf '%s' "$ENV_NAME" | jq -sRr @uri)&per_page=1" --jq '.[0].id // empty')
+
+              if [ -n "$DEPLOY_ID" ]; then
+                STATE=$(gh api "repos/$REPO/deployments/$DEPLOY_ID/statuses?per_page=1" --jq '.[0].state // empty')
+                if [ "$STATE" = "success" ]; then
+                  URL=$(gh api "repos/$REPO/deployments/$DEPLOY_ID/statuses?per_page=1" --jq '.[0].environment_url')
```
Zaimwa9: We could merge these 2 calls into 1

.github/workflows/frontend-pull-request.yml
```diff
@@ -34,3 +35,72 @@ jobs:
...
+                  exit 1
+                fi
+              fi
+              sleep 10
+            done
+
+            if [ -z "$URL" ]; then
+              echo "⚠️ Timed out waiting for '$ENV_NAME' — skipping"
+              continue
+            fi
```
Zaimwa9: Is there a way to have a warning or maybe a non-blocking error in this case ?

.github/workflows/frontend-pull-request.yml
```diff
@@ -34,3 +35,72 @@ jobs:
...
+
+            # Smoke test the deployment
+            echo "🔍 Testing $URL/health ..."
+            STATUS=$(curl -s -o /dev/null -w '%{http_code}' "$URL/health")
```
Zaimwa9: Maybe add timeouts on the curls to not hang indefinitely

.github/workflows/frontend-pull-request.yml
```diff
@@ -34,3 +35,72 @@ jobs:
 
       - name: Run unit tests
         run: npm run test:unit -- --passWithNoTests
+
+  smoke-test:
+    name: Vercel Preview Smoke Test
+    runs-on: ubuntu-latest
```
Zaimwa9, line 42: Maybe a job timeout-minutes (10 minutes should be enough giiven last one ran in 8seconds 😄 )

Zaimwa9 (changes requested):
Nice one and not heavy at all. It's worth giving it a try imo and see whether it becomes.

Just a couple of NIT comments. The timeout ones would be the most important to prevent hanging on vercel

.github/workflows/frontend-pull-request.yml
talissoncosta: Addressed in 9a4f3812a — added `timeout-minutes: 10`.

.github/workflows/frontend-pull-request.yml
talissoncosta: Addressed in 463a05972 — added `--max-time 30` to each curl.

.github/workflows/frontend-pull-request.yml
talissoncosta: Addressed in e507f16a4 — `continue-on-error: true` on the job so failures warn but don't block merge. Also consolidated the 3 separate curl checks into a single endpoint loop.

.github/workflows/frontend-pull-request.yml
talissoncosta: Addressed in 06cd573b1 — state + environment_url now parsed from a single `gh api` call instead of two.

talissoncosta:
Update: the previous approach produced false positives — it was testing URLs before Vercel finished deploying, hitting the stale previous build.

Rewrote to a two-phase approach:
1. Wait for Vercel deployment statuses to complete (polls commit status API for `flagsmith-frontend` checks)
2. Parse URLs from the bot comment and test `/config/project-overrides` + `/` with content validation

Also dropped `/health` (behind VPN now). Validating on #7264.

.github/workflows/frontend-pull-request.yml
```diff
@@ -34,3 +36,104 @@ jobs:
...
+    timeout-minutes: 15
+    continue-on-error: true
```
Zaimwa9, line 44: I'm just worried that this would let the deployment happen in case it raised a real error

Zaimwa9 (changes requested):
Last blocking point to me is the `continue-on-error` that defeats the purpose of blocking the deployment if the target is unhealthy.
I'd separate 2 cases:
1. We couldn't check because of timeout, no statuses or no comment (we don't know but nothing hints in not deploying) => this would be acceptable to skip
2. We managed to send a request to vercel and it failed, the app is clearly down => here we block deployment

I'm unsure how to achieve this but I think we could leverage on the exit status (0 when we want to continue; 1 => we block)

talissoncosta:
Closing — the approach is too brittle. Scraping the Vercel bot's PR comment to extract the preview URL means our CI depends on a third-party comment format we don't control, and silently breaks if Vercel changes it. If we ever want this check, it should pull the preview URL from a reliable source (Vercel API, deployment status, or a job output from Vercel's own check) rather than parsed markdown.

PR #7215, opened 2026-04-10
feat: Feature Analytics label grouping (#6067)

### LINKED ISSUE #6067: Implement dashboard functionality for displaying SDK metrics in Feature Analytics charts
Sibling issue to #5582 (which focused on Organisation Usage charts)

### Feature Analytics Tab

Image

- If labeled buckets present, accumulate buckets for the same day into one bar, representing different label values with different colours.
- Add a filter for label values returned from the API.

talissoncosta:

## Changes

Closes #6067

### Feature Analytics label grouping

When the API returns labelled evaluation buckets (per-SDK via `user_agent` / `client_application_name`), the chart stacks bars by label value with a MultiSelect filter. Falls back to environment-grouped view when no usable labels are present.

Gated behind the `sdk_usage_charts` feature flag — same flag that controls the organisation-level SDK usage view. When disabled, the chart always uses env grouping regardless of the API response.

### New components

| Component | Purpose |
|---|---|
| `BarChart` | Reusable stacked bar chart wrapping recharts. Supports `seriesLabels`, `showLegend`, `barSize`, `verticalGrid`. |
| `ChartTooltip` | Typed tooltip with per-series breakdown + display-name mapping. |
| `ColorSwatch` | Colour indicator (sm/md/lg). |
| `buildChartColorMap` | Pure helper → `Record<string, string>` from the chart palette. |
| `useEnvChartProps` | Env chart config hook (tag-colour matching, env-name labels, stable sort). Reusable — see #7223. |

### Flat token constants

Generator emits 74 camelCase constants (`colorChart1`, `colorTextSecondary`, `radiusMd`, etc.) as `var(--token, #hex)` strings. Browsers resolve `var()` in SVG props natively — no hook, no DOM read, theme toggle reactive via CSS cascade.

### UX

- SDK filter always visible (disabled when no labels) — no layout shift when toggling envs
- Empty state uses the `EmptyState` component — no inline styles
- Env-mode colours match tag chips via `Utils.getTagColour`; tooltip shows env names not IDs

### Storybook

- `Components/BarChart` — 3 stories (`WithLabelledBuckets`, `WithoutLabels`, `SingleSeries`) with deterministic fake data (pinned reference date + hash-based variance) so Chromatic snapshots stay stable across runs

### Follow-ups

- #7223 — Consolidate org-usage charts into `BarChart`, drop legacy `_recharts.scss`
- #7245 — UX decision: environment selector behaviour when chart groups by SDK
- Dedicated MultiSelect refactor PR — dark-mode contrast, disabled state, inline-style cleanup, restore Storybook stories
- Tighten `any` types in `useFeatureAnalytics.ts`

- **Storybook**: BarChart — 3 stories, light + dark
- **App**: Usage tab — env colours match chips, labelled data stacks per-SDK (flag on), filter disables/enables cleanly
- **Real data**: validated on Vercel preview with prod data (post InfluxDB downsample fix)
- **Automated**: 20 unit tests, 0 lint errors, 0 new type errors, 11 E2E passing, 16 visual regression screenshots


talissoncosta (commented):
-

frontend/common/hooks/useChartColors.ts
```diff
@@ -0,0 +1,48 @@
+import { useMemo } from 'react'
+import { CHART_COLOURS } from 'common/theme/tokens'
+
+/**
+ * Reads the computed value of CSS custom properties from the document root.
+ *
```
talissoncosta: Recharts' `fill` prop needs hex strings, not CSS var refs. `getComputedStyle` resolves to the current theme value automatically (dark mode respected).

frontend/common/services/useFeatureAnalytics.ts
```diff
@@ -60,19 +60,27 @@ export const featureAnalyticsService = service
             preBuiltData.push(dayObj)
           }
 
+          // Collect raw entries with labels intact for label-based grouping.
+          // chartData aggregates by environment (existing behaviour),
```
talissoncosta, line 71: `chartData` = existing environment aggregation. `rawEntries` = new, preserves per-entry labels so FeatureAnalytics can group by SDK without calling hooks inside a loop (was a rules-of-hooks violation).

frontend/web/components/charts/BarChart.tsx
```diff
@@ -0,0 +1,74 @@
+import React, { FC } from 'react'
+import {
+  Bar,
+  BarChart as RawBarChart,
```
talissoncosta, line 4: Alias to avoid shadowing — our component is named `BarChart` which would collide with the Recharts import.

frontend/web/components/charts/BarChart.tsx
```diff
@@ -0,0 +1,74 @@
...
+      <RawBarChart data={data}>
+        <CartesianGrid strokeDasharray='3 5' strokeOpacity={0.4} />
+        <XAxis
+          dataKey='day'
+          padding='gap'
+          interval={xAxisInterval}
+          height={80}
+          angle={-90}
+          textAnchor='end'
+          tick={{ dx: -4, fill: '#656D7B', fontSize: 11 }}
```
talissoncosta: Axis colours moved to CSS in `web/styles/3rdParty/_recharts.scss` via `var(--color-text-secondary)` + recharts' built-in classNames (`.recharts-cartesian-axis-tick text`, `.recharts-cartesian-axis-line`). No hex strings — dark mode auto-handled. Only dynamic colours (bar `fill`) still go through `useChartColors` / `getComputedStyle`.

frontend/web/components/charts/BarChart.tsx
```diff
@@ -0,0 +1,74 @@
...
+        />
+        <Tooltip cursor={{ fill: 'transparent' }} content={<ChartTooltip />} />
+        <Legend wrapperStyle={{ paddingTop: 16 }} />
+        {series.map((label, index) => (
+          <Bar
+            key={label}
+            dataKey={label}
+            stackId={stacked ? 'series' : undefined}
+            fill={colorMap.get(label) || '#656D7B'}
+            animationBegin={index * 80}
```
talissoncosta, line 87: Staggered bar entrance — each series starts 80ms after the previous. Uses Recharts built-in animation, not the motion library.

frontend/web/components/ColorSwatch.tsx
```diff
@@ -0,0 +1,35 @@
...
+  className,
+  color,
+  size = 'md',
+}) => (
+  <span
+    aria-hidden='true'
+    className={classNames('d-inline-block flex-shrink-0 rounded-xs', className)}
+    style={{
+      backgroundColor: color,
+      height: SIZE_MAP[size],
```
talissoncosta, line 28: `backgroundColor` is dynamic (per instance). `height`/`width` have no Bootstrap utility for 8/12/16px. `borderRadius` uses the `rounded-xs` token class.

frontend/web/components/feature-page/FeatureNavTab/utils.ts
```diff
@@ -0,0 +1,61 @@
...
+  labelValues: string[]
+} {
+  const grouped: Record<string, ChartDataPoint> = {}
+  const labelSet = new Set<string>()
+  const labelList: string[] = []
+
+  rawData.forEach((entry) => {
+    const date = entry.day
+    // Priority: user_agent (SDK name) > client_application_name > 'Unknown'
+    const labelValue =
```
talissoncosta, line 54: Priority from backend: `user_agent` (SDK name, most useful) > `client_application_name` (fallback) > `Unknown` (empty/missing labels).

frontend/common/services/useFeatureAnalytics.ts
```diff
@@ -60,19 +60,27 @@ export const featureAnalyticsService = service
...
           responses.forEach((response, i) => {
             const environment_id = query.environment_ids[i]
 
             response.data?.forEach((entry) => {
+              rawEntries.push(entry)
               const date = moment(entry.day).format('Do MMM')
               const dayEntry = preBuiltData.find((d) => d.day === date)
               if (dayEntry) {
-                dayEntry[environment_id] = entry.count // Set count for specific environment ID
+                dayEntry[environment_id] = entry.count
```
Zaimwa9: Can we move this to the same pattern as in `aggregateByLabels` and use a `Map<string, ChartDataPoint` for larger payloads ?

frontend/web/components/charts/ChartTooltip.tsx
```diff
@@ -0,0 +1,50 @@
...
+import {
+  NameType,
+  Payload,
+  ValueType,
+} from 'recharts/types/component/DefaultTooltipContent'
+import ColorSwatch from 'components/ColorSwatch'
+
+type SeriesPayload = Payload<ValueType, NameType>
+
+const formatNumber = (n: number | string | undefined): string =>
```
Zaimwa9: NIT: can we use variable names a bit more readable than single char please. Here and in a couple of places

Zaimwa9 (changes requested):
Looks good overall but couple of comments:
- The one i'd like being fixed if possible is to use a Map for the chartData, that might reduce computation complexity
- NIT on a couple of single character variables that could be made a bit more explicit

Another one is that while testing, I realized we are not sending the labels. I'm looking into why but it looks like it's coming from the edge-api that is not tracking the labels for `POST analytics/flags` while it does for the usage. I'm looking into it

frontend/common/services/useFeatureAnalytics.ts
talissoncosta: Good catch — switched to a `day → bucket` Map built alongside preBuiltData, so the per-entry lookup is O(1) instead of the O(days × entries) `Array.find` dance. Commit 848379201

frontend/web/components/charts/ChartTooltip.tsx
talissoncosta: Renamed across the touched files in 848379201: `n`/`v` → `value`, `el` → `entry` in `ChartTooltip`; `v` → `value` in `BarChart` tickFormatter; `v` → `response` in `useFeatureAnalytics`; `e` → `env` (plus a small helper dedupe) in `useEnvChartProps`; `d` → `day` in `utils`. Kept `i` where it was a standard forEach index.

frontend/web/components/feature-page/FeatureNavTab/FeatureAnalytics.tsx
```diff
@@ -39,99 +33,118 @@ const FlagAnalytics: FC<FlagAnalyticsType> = ({
...
+          </div>
+          <MultiSelect
+            className='w-100'
+            label='Filter by SDK'
+            options={labelOptions}
+            selectedValues={selectedLabels}
+            onSelectionChange={setSelectedLabels}
+            colorMap={labelColorMap}
+            disabled={!isLabelled || labelValues.length <= 1}
+          />
```
Zaimwa9, line 119: My bad, I forgot to let you know that for now it's only activated for Flagsmith team (The organisation one).

If you don't mind, let's wrap this one with the flag `sdk_usage_charts` and set `isLabelled` false when flag disabled ?

We'll release very soon!

Zaimwa9 (changes requested):
Looks good, sorry for one item I didn't think about last time: the sdk_usage feature flag.

Otherwise it's nice, i do like the `useEnv` hook. Even though let's create a ticket in order to see how we deal with the label we want to show. As soon as there are labeled data we don't have the breakdown per env. It might be an interesting data for some users (maybe a toggle to switch? I don't have the answer to whether it's valuable and how to do it) but it's worth having an issue tracking this.

Just the flag and ready to approve

talissoncosta:
No worries @Zaimwa9, thanks for flagging that , flag gating landed in 040e3d108.

I had the same concern about the env selector in labelled mode. I considered hiding it entirely since it doesn’t change the UI when labels are stacked, but kept it because it still acts as a silent data filter. Selecting fewer envs narrows the aggregated totals.

The tradeoff is that users can’t see which env contributed to each SDK stack.

Opened #7245 to track the UX discussion.

Zaimwa9 (approved):
Thanks 👍

PR #4583, opened 2024-09-04
chore: Adjust payment options in app

kyle-ssg:
Thanks for submitting a PR! Please check the boxes below:

- [x] I have added information to `docs/` if required so people know about the feature!
- [x] I have filled in the "Changes" section below?
- [x] I have filled in the "How did you test this code" section below?
- [x] I have used a [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) title for this Pull Request

## Changes

- Hides scaleup under feature flag
- Adjusts pricing styles for enterprise
image

PR #7216, opened 2026-04-10
feat(Billing): Reintroduce the Scale-Up plan

emyller:

## Changes

Contributes to https://github.com/Flagsmith/flagsmith-private/issues/107

Restores the Scale-Up plan in the in-app organisation billing page.

Also includes a fix to make "Startup" the first upgrade plan from free.

image

image

image

image


frontend/web/components/modals/payment/Payment.tsx
```diff
@@ -104,6 +108,27 @@ export const Payment: FC<PaymentProps> = ({
...
           />
 
+          <PricingPanel
+            title='Scale-Up'
+            priceYearly='270'
+            priceMonthly='300'
+            seatPriceYearly='50'
+            seatPriceMonthly='60'
+            isYearly={yearly}
+            isFeatured
```
emyller, line 119: This triggers the "Most popular!" tag on the price tile (see screenshot). It's inherited from https://github.com/Flagsmith/flagsmith/pull/4583 — not sure if it should stay. @asaphko please advise!

frontend/env/project_prod.js
```diff
@@ -27,7 +27,7 @@ const Project = {
   // This is used for Sentry tracking
   maintenance: false,
   plans: {
-    scaleUp: { annual: 'scale-up-12-months-v2', monthly: 'scale-up-v2' },
+    scaleUp: { annual: 'scale-up-12-months-v4', monthly: 'scale-up-v4' },
```
emyller: Blocker until the `scale-up-v4` plan is created. Ensure the annual period has the correct ID as well.

emyller: Chargebee is set up!

frontend/web/components/modals/payment/Payment.tsx
emyller: @asaphko says 👍

frontend/web/components/pages/organisation-settings/tabs/BillingTab.tsx
```diff
@@ -2,8 +2,8 @@ import React from 'react'
 import { Organisation } from 'common/types/responses'
 import Icon from 'components/icons/Icon'
 import Utils from 'common/utils/utils'
-import PaymentLegacy from 'components/modals/Payment'
-import PaymentNew from 'components/modals/payment'
+import PaymentLegacy from 'components/modals/Payment.js'
```
talissoncosta: we could drop the `.js` in here I think

frontend/web/components/pages/organisation-settings/tabs/BillingTab.tsx
```diff
@@ -2,8 +2,8 @@ import React from 'react'
...
+import { Payment as PaymentNew } from 'components/modals/payment/Payment'
```
talissoncosta: any reason to change the import in here ?

talissoncosta (changes requested):
Just some minor comments regarding the imports... Otherwise looks good.

----

Have you tested the upgrade flow end-to-end with `rtk_payment_modal_migration` both on and off?

The new payment modal is behind the flag, but the `getNextPlan()` change in `utils.tsx` isn't — it affects the upgrade chips across the app for everyone. Users on the legacy modal would see "Upgrade to Scale-Up" chips but land on a modal that doesn't have a Scale-Up plan.

The flag has been on for internal testing for a while — could be a good opportunity to enable it for everyone alongside this. Otherwise we should gate `getNextPlan()` behind the flag to keep the old behaviour for legacy modal users.

frontend/common/utils/utils.tsx
```diff
@@ -397,17 +397,20 @@ const Utils = Object.assign({}, BaseUtils, {
...
       }
       case planNames.startup: {
-        return planNames.startup
+        return planNames.scaleUp
+      }
+      case planNames.scaleUp: {
+        return planNames.enterprise
       }
       default: {
         return planNames.enterprise
```
talissoncosta, line 416: This isn't behind the `rtk_payment_modal_migration` flag — users on the legacy modal will see "Upgrade to Scale-Up" chips but land on a modal that doesn't have Scale-Up.                              
                                                                                                                                                                                                         
  If we're not enabling the flag globally with this PR, we should gate it:
                                                                                                                                                                                                         
  ```suggestion                                                                                                                                                                                          
    getNextPlan: () => {                                                                                                                                                                                 
      const currentPlan = Utils.getPlanName(AccountStore.getActiveOrgPlan())                                                                                                                             
      if (currentPlan !== planNames.enterprise && !Utils.isSaas()) {                                                                                                                                     
        return planNames.enterprise                         
      }
                                                                                                                                                                                                         
      // TODO: Remove this branch and inline the new logic when the
      // `rtk_payment_modal_migration` flag is enabled globally.                                                                                                                                         
      if (!Utils.getFlagsmithHasFeature('rtk_payment_modal_migration')) {
        switch (currentPlan) {                                                                                                                                                                           
          case planNames.free:            
            return planNames.scaleUp                                                                                                                                                                     
          case planNames.startup:                                                                                                                                                                        
            return planNames.startup
          default:                                                                                                                                                                                       
            return planNames.enterprise                     
        }                                 
      }
                                                                                                                                                                                                         
      switch (currentPlan) {
        case planNames.free:                                                                                                                                                                             
          return planNames.startup                          
        case planNames.startup:
          return planNames.scaleUp
        case planNames.scaleUp:               
          return planNames.enterprise     
        default:
          return planNames.enterprise                                                                                                                                                                    
      }
    },                                                                                                                                                                                                   
  ```                                                       

  Honestly though, the flag's been on for internal testing for a while — might be cleaner to enable it globally with this PR rather than adding flag-gated branches we'll have to clean up later.

frontend/env/project_local.js
```diff
@@ -17,9 +17,10 @@ const Project = {
   // This is used for Sentry tracking
   maintenance: false,
   plans: {
-    scaleUp: { annual: 'scale-up-annual-v2', monthly: 'scale-up-v2' },
-    startup: { annual: 'startup-annual-v2', monthly: 'startup-v2' },
+    scaleUp: { annual: 'Scale-Up-v4-USD-Yearly', monthly: 'Scale-Up-v4-USD-Monthly' },
+    startup: { annual: 'start-up-12-months-v2', monthly: 'startup-v2' },
   },
+  cookieSameSite: 'lax',
```
talissoncosta: Just checking, what is this about? Looks unrelated to the Scale-Up plan changes.

frontend/env/project_local.js
```diff
@@ -17,9 +17,10 @@ const Project = {
   // This is used for Sentry tracking
   maintenance: false,
   plans: {
-    scaleUp: { annual: 'scale-up-annual-v2', monthly: 'scale-up-v2' },
-    startup: { annual: 'startup-annual-v2', monthly: 'startup-v2' },
+    scaleUp: { annual: 'Scale-Up-v4-USD-Yearly', monthly: 'Scale-Up-v4-USD-Monthly' },
```
talissoncosta, line 20: Just to confirm. Should the plan ID be in PascalCase like this, or could it be kebab-case to stay consistent with the others?

frontend/web/components/pages/organisation-settings/tabs/BillingTab.tsx
emyller: Looks like the module resolver has a hard time on MacOS, because there are both `components/modals/payment`, and `components/modals/Payment`.

We can proceed without this change since it doesn't affect production but I'm surprised I _needed_ this in my local env. Do you have any advices here?

frontend/env/project_local.js
emyller: Annoying as it may be, that's how it's set up in Chargebee. Thanks for verifying. :)

frontend/env/project_local.js
emyller: Good catch. I need it for local testing to work, and it slipped through. Fixed in da6302ada.

frontend/web/components/pages/organisation-settings/tabs/BillingTab.tsx
talissoncosta: Got that. We could rename it to `LegacyPayment.js` to fix the collision.

emyller: I've managed to not meed around with imports here. b1b2297fad7d7b7499d5b2d841df6c3cf70b4c4c

frontend/common/utils/utils.tsx
emyller: I've re-approached this by introducing the new pricing tile to both old and new UIs: b1b2297fad7d7b7499d5b2d841df6c3cf70b4c4c.

talissoncosta (approved):
All good! Thanks for handling that!

emyller:
@talissoncosta Thanks for the great review here. I've started from scratch, aiming to support both old and new UIs. I've tested both, but haven't gone through activating a trial — I trust the billing masters above us know what they're doing. 👀  I'll give it a go once live.

PR #7217, opened 2026-04-10
feat(Billing): Cap Scale-Up seats at 20

emyller:

## Changes

Contributes to https://github.com/Flagsmith/flagsmith-private/issues/107

Restricts scale-up seat count to 20.

Updated unit tests.


api/organisations/subscriptions/constants.py
```diff
@@ -41,12 +43,7 @@
 FREE_PLAN_ID = "free"
 TRIAL_SUBSCRIPTION_ID = "trial"
 SCALE_UP = "scale-up"
-SCALE_UP_12_MONTHS_V2 = "scale-up-12-months-v2"
-SCALE_UP_QUARTERLY_V2_SEMIANNUAL = "scale-up-quarterly-v2-semiannual"
-SCALE_UP_V2 = "scale-up-v2"
 STARTUP = "startup"
-STARTUP_ANNUAL_V2 = "startup-annual-v2"
-STARTUP_V2 = "startup-v2"
```
Zaimwa9, line 49: Are we sure removing them is harmless ? I'm seeing some customers with these plans in the in the django admin

Zaimwa9 (approved):
Approving with one concern to make sure of please 🙏

api/organisations/subscriptions/constants.py
emyller: Thanks for being diligent! This is actually dead code, it is never used. I wrote a83b869a2bb70d4a0cd3157813acb6063b6c5146 separately after finding no references anywhere in this or in other repositories.

Zaimwa9: SGTM

PR #7225, opened 2026-04-14
feat: Enable OpenTelemetry support for Task Processor

### LINKED ISSUE #184: OTel: Task Processor

khvn26:

## Changes

Closes https://github.com/Flagsmith/flagsmith-common/issues/184.

Incorporates task processor spans added in https://github.com/Flagsmith/flagsmith-common/pull/197, and enables JSON logs for all SaaS deployments.

Will test in staging.

### DIFF
--- a/api/poetry.lock
+++ b/api/poetry.lock
@@ -967,7 +967,7 @@ description = "Cross-platform colored terminal text."
 optional = false
 python-versions = "!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*,!=3.5.*,!=3.6.*,>=2.7"
 groups = ["dev"]
-markers = "platform_system == \"Windows\" or sys_platform == \"win32\""
+markers = "sys_platform == \"win32\" or platform_system == \"Windows\""
 files = [
     {file = "colorama-0.4.6-py2.py3-none-any.whl", hash = "sha256:4f1d9991f5acc0ca119f9d443620b77f9d6b33703e51011c16baf57afb285fc6"},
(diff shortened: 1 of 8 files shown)


Zaimwa9 (approved):
Lgtm

PR #7229, opened 2026-04-14
refactor: replace global lodash with direct per-file imports

kyle-ssg:

## Changes

Remove the global window._ lodash object from libs.js and replace all _.method() usages with explicit imports from lodash submodules in each file that needs them. This eliminates the implicit global dependency and enables better tree-shaking.

- There's no instances of _. anymore
- It builds
- E2E


Zaimwa9 (approved):
👍

PR #7250, opened 2026-04-15
fix: env columns hard to distinguish in compare view

### LINKED ISSUE #7249: UI: Add visual divider between environment columns in comparison view
## Description

In the environment comparison view, the segment override icon (pie icon) sits too close to the next environment's column, making it hard to visually separate where one environment ends and the next begins.

## Current Behaviour

The segment override icon renders adjacent to the next environment's toggle/value, with no clear boundary between columns. This is especially noticeable when scanning a long list of features across multiple environments.

## Proposed Solution

Add a visible vertical divider (or extra spacing) between environment columns to make the boundary clear. The segment override icon should belong unambiguously to its parent environment.

## Where this appears
- Feature comparison view across environments

## Reported by
External customer feedback.

talissoncosta:
- [ ] I have added information to \`docs/\` if required so people know about the feature.
- [x] I have filled in the \"Changes\" section below.
- [x] I have filled in the \"How did you test this code\" section below.

## Changes

Closes #7249

In the environment comparison view, the segment override icon (pie icon) sat flush against the next environment's column, making it hard to visually separate where one environment ends and the next begins.

This PR adds a left border and small padding to the right environment column (both header and rows) so the boundary between environments is clear.

image

1. Open a project with multiple environments and segment overrides on at least one feature
2. Navigate to \`/project/{id}/compare\`
3. Select two environments
4. Confirm there is now a visible vertical divider between the left and right environment columns
5. Confirm the segment override icon no longer sits flush against the next column

### DIFF
--- a/frontend/web/components/CompareEnvironments.js
+++ b/frontend/web/components/CompareEnvironments.js
@@ -313,6 +313,7 @@ class CompareEnvironments extends Component {
                         <FeatureRow
                           condensed
                           isCompareEnv
+                          className='border-left-1 ps-2'
                           fadeEnabled={fadeEnabled}
                           fadeValue={fadeValue}
                           history={this.props.history}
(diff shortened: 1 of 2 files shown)


Zaimwa9 (approved):
👍

PR #7251, opened 2026-04-15
infra(OTel): Wire OTEL_EXPORTER_ENDPOINT in ECS task defs

khvn26:

## Changes

Contributes to Flagsmith/pulumi#170

In this PR, we add `OTEL_EXPORTER_ENDPOINT` pointing at the ADOT collector's CloudMap hostname to the admin-api, sdk-api, and task-processor task definitions for both staging and production, so the services can emit OTLP logs and traces to the shared collector once Flagsmith/pulumi#182 is deployed. The migration task def is intentionally left alone — it's a one-shot and doesn't emit events.

The endpoint is identical for staging and production because CloudMap is per-VPC and the `metrics.local` namespace resolves `adot-collector` inside each environment's VPC.

Depends on Flagsmith/pulumi#182.

- [ ] Merge the pulumi plumbing PR first and verify the CloudMap hostname resolves from an API task in staging
- [ ] Deploy this to staging and verify the env var lands in running containers
- [ ] Verify the API emits OTLP traffic to the collector by checking the collector's debug log output in CloudWatch
- [ ] Roll out to production

### DIFF
--- a/infrastructure/aws/production/ecs-task-definition-admin-api.json
+++ b/infrastructure/aws/production/ecs-task-definition-admin-api.json
@@ -206,6 +206,10 @@
                     "value": "True"
                 },
                 {
+                    "name": "OTEL_EXPORTER_ENDPOINT",
+                    "value": "http://adot-collector.metrics.local:4318"
+                },
+                {
                     "name": "FLAGSMITH_ON_FLAGSMITH_SERVER_API_URL",
                     "value": "https://edge.api.flagsmith.com/api/v1/"
                 },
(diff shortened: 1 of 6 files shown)


emyller (approved):

khvn26:
This is blocked on state drift in staging Pulumi the "verify the CloudMap hostname resolves from an API task in staging" part worked but the TCP port isn't accessible yet due to a secgrp mismatch.

khvn26:
Opened https://github.com/Flagsmith/pulumi/issues/185.

PR #7256, opened 2026-04-16
fix(Logging): Expand APPLICATION_LOGGERS and fix unnamed loggers

khvn26:

## Changes

`APPLICATION_LOGGERS` only listed 5 packages, so INFO-level logs from most of the codebase were silently filtered at the root WARNING level. This expands the list to cover all packages and explicit structlog logger names that actually emit info logs.

Fixes 7 loggers that called `logging.getLogger()` or `structlog.get_logger()` without a name, routing them through the root logger and bypassing `APPLICATION_LOGGERS` entirely. These now use `__name__`. The manual `logger.setLevel(logging.INFO)` in `features/views.py` is removed since `features` is already covered centrally.

That `setLevel` call was on the root logger, so it was accidentally making INFO logs visible process-wide. Tests that asserted on INFO messages via a custom `inspecting_handler` fixture relied on this side effect. Added `log_level = 'INFO'` to `[tool.pytest.ini_options]` so `caplog` captures INFO in tests, then replaced `inspecting_handler` with `caplog` and removed the fixture.

One old migration (`features/migrations/0018`) still has an unnamed logger but shouldn't be modified.

Audited every `logging.getLogger` and `structlog.get_logger` call in the codebase against the list. Ran all affected tests locally. Pre-commit hooks pass.

### DIFF
--- a/api/conftest.py
+++ b/api/conftest.py
@@ -1,4 +1,3 @@
-import logging
 import os
 import site
 import typing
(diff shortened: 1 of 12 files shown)


Zaimwa9 (approved):
I like this caplog API

gagantrivedi (approved):

emyller (approved):

PR #7263, opened 2026-04-16
test: Validate smoke test on fresh PR

talissoncosta:
Throwaway PR to test the smoke test workflow end-to-end on a fresh PR with a real Vercel deployment. Will close after validation.


talissoncosta:
Test complete — smoke test has a false-positive issue when Vercel hasn't finished deploying. Closing while we rethink the approach.

PR #7264, opened 2026-04-16
test: Validate smoke test (two-phase approach)

talissoncosta:
Throwaway PR to validate the smoke test waits for Vercel before testing. Will close after validation.


talissoncosta:
Validation complete. Closing test PR.

Issue #7160, opened 2026-04-07
GitLab integration: link issues and MRs to feature flags


emyller:
Context: https://github.com/Flagsmith/flagsmith/issues/7000

After connecting their GitLab instance (https://github.com/Flagsmith/flagsmith/issues/7159), users can link GitLab issues and merge requests to their feature flags.

### Scope

- Proxy endpoints to browse GitLab projects, issues, and merge requests using the stored connection credentials.
- Store the link between a feature flag and a GitLab issue or MR.

> [!IMPORTANT]
> Unlike GitHub, where users pre-select repositories during the App installation flow, GitLab tokens grant access to all projects within their scope. A project selection UI would expose someone else's token scope to any Flagsmith user on the integrations page. The GitLab integration exposes all accessible projects by design.

> [!NOTE]
> https://github.com/Flagsmith/flagsmith/pull/7120 and https://github.com/Flagsmith/flagsmith/pull/7121 contain reference code. A few things to consider:
>
> - `GitLabResource` in `types.py` represents an issue or MR — name is vague.
> - Other client functions in https://github.com/Flagsmith/flagsmith/pull/7120 belong in later issues:
>   - `post_comment_to_gitlab` → https://github.com/Flagsmith/flagsmith/issues/7162
>   - `create_flagsmith_flag_label`, `label_gitlab_resource` → https://github.com/Flagsmith/flagsmith/issues/7163
>   - Webhook handling → https://github.com/Flagsmith/flagsmith/issues/7165

### Acceptance criteria

- [x] In a feature flag's Links tab, users can select a GitLab project, choose between issue or merge request, and search for resources to link
- [x] Linked GitLab resources appear in the Links tab with type, name, status, and a link to GitLab
- [x] Emit structlog events for product telemetry (https://github.com/Flagsmith/flagsmith/issues/7012)

PR #7270, opened 2026-04-16
feat(GitLab): Browse GitLab issues and merge requests (backend)

emyller:

## Changes

Contributes to https://github.com/Flagsmith/flagsmith/issues/7160

**Stack:** #7270 (this) → #7273 (frontend) → #7274 (linking)

Allow browsing issues and merge requests in GitLab from the feature modal. This pull request adds the backend proxy endpoints that talk to the GitLab API v4 using the connection credentials stored via #7239. Frontend and linking follow in the stack.

Review effort: 2/5

Unit tests and manual API hammering.

### How to review

- Some diff is added because of moving `gitlab/views.py` to `gitlab/views/configuration.py`.
- Start from URLs, then views, then the API code at `gitlab/client/api.py`.
  - Looking at screenshots in the next issues might help!
  - Tip: each view has a 1:1 relationship to a client function.

### DIFF
--- a/api/integrations/gitlab/views/browse_gitlab.py
+++ b/api/integrations/gitlab/views/browse_gitlab.py
@@ -57,7 +57,7 @@ class _GitLabListView(ListAPIView, abc.ABC, Generic[T]):  # type: ignore[type-ar
         except requests.RequestException as exc:
             logger.error("api-call-failed", exc_info=exc)
             return Response(
-                data={"detail": f"GitLab API error: {exc}"},
+                data={"detail": "GitLab API is unreachable"},
                 status=status.HTTP_424_FAILED_DEPENDENCY,
             )
 
--- a/api/integrations/gitlab/views/browse_gitlab.py
+++ b/api/integrations/gitlab/views/browse_gitlab.py
@@ -1,0 +1,164 @@
+import abc
+from typing import Any, Generic
+
+import requests
+import structlog
+from rest_framework import status
+from rest_framework.generics import ListAPIView
+from rest_framework.request import Request
+from rest_framework.response import Response
+
+from integrations.gitlab.client import (
+    GitLabIssue,
+    GitLabMergeRequest,
+    GitLabPage,
+    GitLabProject,
+    fetch_gitlab_projects,
+    search_gitlab_issues,
+    search_gitlab_merge_requests,
+)
+from integrations.gitlab.client.types import T
+from integrations.gitlab.models import GitLabConfiguration
+from integrations.gitlab.serializers import (
+    PaginatedQueryParamsSerializer,
+    SearchQueryParamsSerializer,
+)
+from projects.permissions import NestedProjectPermissions
+
+logger = structlog.get_logger("gitlab")
+
+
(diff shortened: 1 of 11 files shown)


api/integrations/gitlab/views/browse_gitlab.py
```diff
@@ -0,0 +1,164 @@
...
+
+        serializer = self.serializer_class(data=request.query_params)
+        serializer.is_valid(raise_exception=True)
+
+        try:
+            page_data = self.fetch_page(config, serializer.validated_data)
+        except requests.RequestException as exc:
+            logger.error("api-call-failed", exc_info=exc)
+            return Response(
+                data={"detail": f"GitLab API error: {exc}"},
```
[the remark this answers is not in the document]
emyller: e3cab0396976b521ba28413faa2996d6f67862fe

api/integrations/gitlab/views/browse_gitlab.py
```diff
@@ -0,0 +1,164 @@
...
+        serializer = self.serializer_class(data=request.query_params)
+        serializer.is_valid(raise_exception=True)
+
+        try:
+            page_data = self.fetch_page(config, serializer.validated_data)
+        except requests.RequestException as exc:
+            logger.error("api-call-failed", exc_info=exc)
+            return Response(
+                data={"detail": "GitLab API is unreachable"},
+                status=status.HTTP_424_FAILED_DEPENDENCY,
```
khvn26, line 61: See https://stackoverflow.com/a/18711343.

```suggestion
                status=status.HTTP_503_SERVICE_UNAVAILABLE,
```

khvn26 (dismissed):
Excellent work. Just 1 minor comment.

api/integrations/gitlab/views/browse_gitlab.py
emyller: Yeah I got myself wondering for some time here. The rationale is that I wouldn't like a GitLab outage to influence our SLIs, as 5xx responses will. I don't have a strong opinion here. 503 feels right though, I agree.

Any final thoughts before I apply the suggestion?

emyller: Mmm I think I do have an opinion now. I think the amount of failed requests to these particular endpoints will probably not influence our metrics much.

khvn26: > The rationale is that I wouldn't like a GitLab outage to influence our SLIs, as 5xx responses will

I think something like a feature flag killswitch is a better contingency tactic for that than semantic deception.

> Mmm I think I do have an opinion now. I think the amount of failed requests to these particular endpoints will probably not influence our metrics much.

I tend to agree with this.

emyller: > I think something like a feature flag killswitch is a better contingency tactic for that than semantic deception.

Would be very cool to automate that based on service (e.g. GitLab) status.

emyller: 7f340348ca8c6741ac0b5dc1fe75889741c4e044

khvn26 (approved):

PR #7273, opened 2026-04-17
feat(GitLab): Browse GitLab issues and merge requests (frontend)

emyller:

## Changes

Contributes to https://github.com/Flagsmith/flagsmith/issues/7160

**Stack:** #7270 (backend) → #7273 (this) → #7274 (linking)

With the backend proxy endpoints in place (#7270), the frontend can now let users browse their GitLab projects, issues, and merge requests directly from the feature flag Links tab.

### How to review

- Start from the Links tab wiring in `modals/create-feature/index.tsx`.
- `GitLabLinkSection` and `GitHubLinkSection` are siblings.

Review effort: 2/5

Manual UI testing. Screenshots:

image
image
image
image
image


frontend/common/services/useGitlab.ts
```diff
@@ -0,0 +1,84 @@
...
+          url: `projects/${query.project_id}/gitlab/projects/?page_size=${query.page_size ?? 100}&page=${query.page ?? 1}`,
+        }),
+      }),
+      getGitLabIssues: builder.query<
+        Res['gitlabIssues'],
+        Req['getGitLabIssues']
+      >({
+        providesTags: [{ id: 'LIST', type: 'GitLab' }],
+        query: (query: Req['getGitLabIssues']) => ({
+          url:
```
talissoncosta: These URLs aren't encoded, so a search like `foo&state=closed` sneaks in as extra query params. Any chance we can switch to `URLSearchParams` here? Same thing exists in `useGithub.ts` already, but nice not to spread it further.

frontend/common/services/useGitlab.ts
```diff
@@ -0,0 +1,84 @@
...
+            `projects/${query.project_id}/gitlab/issues/` +
+            `?gitlab_project_id=${query.gitlab_project_id}&page_size=${query.page_size ?? 100}&page=${query.page ?? 1}&search_text=${query.q || ''}&state=opened`,
```
Zaimwa9: Ideally we would use `encodeURIComponent` here to avoid corrupting the url with spaces for example

frontend/common/services/useGitlab.ts
```diff
@@ -0,0 +1,84 @@
...
+      }),
+      getGitLabMergeRequests: builder.query<
+        Res['gitlabMergeRequests'],
+        Req['getGitLabMergeRequests']
+      >({
+        providesTags: [{ id: 'LIST', type: 'GitLab' }],
+        query: (query: Req['getGitLabMergeRequests']) => ({
+          url:
+            `projects/${query.project_id}/gitlab/merge-requests/` +
+            `?gitlab_project_id=${query.gitlab_project_id}&page_size=${query.page_size ?? 100}&page=${query.page ?? 1}&search_text=${query.q || ''}&state=opened`,
```
Zaimwa9: Same with `encodeUriComponent`

frontend/common/services/useGitlab.ts
Zaimwa9: Also state is part of the type and hardcoded here. If we don't plan to use it and leave it hardcoded here, we should remove it from the type (as unused).

I would prefer instead to show the intention and have `state={query.state || "opened"}` here to avoid surprises later.

Same below

frontend/web/components/GitHubLinkSection.tsx
```diff
@@ -0,0 +1,72 @@
...
+}) => {
+  const githubTypes = Object.values(Constants.resourceTypes).filter(
+    (v) => v.type === 'GITHUB',
+  )
+  const [createExternalResource] = useCreateExternalResourceMutation()
+  const [githubResourceType, setGithubResourceType] = useState(
+    githubTypes[0]?.resourceType,
+  )
+
+  const addGithubResource = (githubResource: GithubResource) => {
```
Zaimwa9: NIT: I know this was legacy but it would be nice to catch and toast the error here. And would be great to turn it `async/await` at the same time

frontend/web/components/GitLabProjectSelect.tsx
```diff
@@ -0,0 +1,57 @@
...
+        value={
+          data?.results
+            ? data.results
+                .map((p) => ({
+                  label: p.path_with_namespace,
+                  value: p.id,
+                }))
+                .find((o) => o.value === value)
+            : null
+        }
```
Zaimwa9:
```suggestion
 value={data?.results
    ?.map((p) => ({ label: p.path_with_namespace, value: p.id }))
    .find((o) => o.value === value) ?? null}
```

frontend/web/components/GitLabSearchSelect.tsx
```diff
@@ -0,0 +1,84 @@
...
+        }}
+        options={data?.results
+          ?.filter(
+            (r: GitLabIssue | GitLabMergeRequest) =>
+              !linkedUrls.includes(r.web_url),
+          )
+          .map((r: GitLabIssue | GitLabMergeRequest) => ({
+            label: `${r.title} #${r.iid}`,
+            value: r,
+          }))}
```
Zaimwa9: NIT: we could avoid this heavy syntax in the JSX and extract it in a variable (I think it would make the type inference straightforward too)

frontend/web/components/GitLabProjectSelect.tsx
```diff
@@ -0,0 +1,57 @@
...
+      <Select
+        className='w-100 react-select'
+        size='select-md'
+        placeholder={
+          isLoading
+            ? 'Loading...'
+            : isError
+              ? 'Failed to load projects'
+              : 'Select GitLab Project'
+        }
```
Zaimwa9: It doesn't look like the right place to show the error. On top of the disabled state of the select, it will be easy to miss out.

Maybe use `<ErrorMessage />` below the select?

frontend/common/hooks/useHasGitLabIntegration.ts
```diff
@@ -0,0 +1,9 @@
+import { useGetGitLabConfigurationQuery } from 'common/services/useGitlabConfiguration'
+
+export function useHasGitLabIntegration(projectId: number) {
+  const { data } = useGetGitLabConfigurationQuery(
+    { project_id: projectId },
+    { skip: !projectId },
+  )
+  return { hasIntegration: !!data?.length }
+}
```
Zaimwa9, line 9: Dogfooding proposition: do we have a flag already for gitlab? But that would mean somehow revisiting https://github.com/Flagsmith/flagsmith/pull/7239/changes#diff-ed627e995a4ca8177ae31725a482e4259c5842e6503cfc2b00da11c09ac88947.

Up to you!

Zaimwa9 (changes requested):
Good one. Couple of comments, the `encodeURIComponent` is the one worth fixing while we are at it. Ideally the error message in the Select.

I allowed myself a couple of less important and style comments that could be tackled at the same time.

And whether you want to use a flag

frontend/web/components/GitHubLinkSection.tsx
```diff
@@ -0,0 +1,72 @@
...
+      AppActions.refreshFeatures(projectId, environmentId)
+    })
+  }
+
+  return (
+    <GitHubResourcesSelect
+      githubId={githubId}
+      resourceType={githubResourceType}
+      setResourceType={setGithubResourceType}
+      onChange={addGithubResource as any}
```
talissoncosta: Not trying to nitpick — just keen to avoid the two `any` casts slipping in here. The request types downstream (`getGithubRepositories`, `getGithubResources`) are already `organisation_id: number` / `github_id: number`, but the select components above them are typed as `string`, which is what forces the casts at the call site.

The `onChange` cast is a separate issue: the prop is declared `(value: string) => void` but the component actually passes a `GithubResource` (see `GitHubResourcesSelect.tsx:104-107`), which is why `addGithubResource as any` is needed.

If you're up for it, a few small edits clear both `any`s and make the chain consistent:

```ts
// useHasGithubIntegration.ts — drop the '' fallback
githubId: data?.results?.[0]?.id,   // number | undefined
```

```ts
// MyRepositoriesSelect.tsx props
githubId: number   // was string
orgId: number      // was string
```

```ts
// GitHubResourcesSelect.tsx props
onChange: (value: GithubResource) => void   // was (value: string) => void
githubId: number                             // was string
orgId: number                                // was string
```

```ts
// GitHubLinkSection.tsx props
githubId: number   // was string
```

Then the JSX in `GitHubLinkSection` is just:

```tsx
onChange={addGithubResource}
orgId={organisationId}
```

`GitHubLinkSection` is the only consumer of `GitHubResourcesSelect`, and `MyRepositoriesSelect` is only used inside `GitHubResourcesSelect` itself — so no wider fallout. Happy for it to be a follow-up PR if you'd rather keep this one tight.

frontend/web/components/GitLabLinkSection.tsx
```diff
@@ -0,0 +1,79 @@
+import React, { FC, useState } from 'react'
+import Constants from 'common/constants'
+import GitLabProjectSelect from 'components/GitLabProjectSelect'
+import GitLabSearchSelect from 'components/GitLabSearchSelect'
+import type { GitLabIssue, GitLabMergeRequest } from 'common/types/responses'
+
+type GitLabLinkType = 'issue' | 'merge_request'
```
talissoncosta: This is defined inlined again as `linkType: 'issue' | 'merge_request'` in [GitLabSearchSelect](https://github.com/Flagsmith/flagsmith/pull/7273/changes#diff-9e54ed7e6aef6a2ffff340eb7817609541a300528e1eb88354a39056f35c13d3R13). Worth pulling it out once in `common/types/responses.ts`) and importing in both places — keeps them from drifting if we ever add e.g. `'epic'`.

talissoncosta (commented):
Really solid review from @Zaimwa9

Agree with all of it. Just adding a few small points on top, just to prevent the any's to spread around.

frontend/web/components/GitLabProjectSelect.tsx
emyller: Would you prefer the `<ErrorMessage />` for this particular use case, or [a toast](https://github.com/Flagsmith/flagsmith/pull/7273#discussion_r3101080458)? I wonder where to draw the line.

frontend/common/services/useGitlab.ts
emyller: > Ideally we would use `encodeURIComponent` here to avoid corrupting the url with spaces for example

1b4b870df774ab7a3c7f5de13cdd94e49ab1a9e6

frontend/common/services/useGitlab.ts
emyller: 1b4b870df774ab7a3c7f5de13cdd94e49ab1a9e6

frontend/common/hooks/useHasGitLabIntegration.ts
emyller: We control visibility of integration options in the `integration_data` remote config. GitLab is only available for our internal use for now. Are you interested in gating this code with a feature flag specific to the GitLab integration?

frontend/common/services/useGitlab.ts
emyller: 1b4b870df774ab7a3c7f5de13cdd94e49ab1a9e6

`URLSearchParams` was also my instinct, but I learned we already ship with an utility. See above.

frontend/common/services/useGitlab.ts
emyller: > Also state is part of the type and hardcoded here. If we don't plan to use it and leave it hardcoded here, we should remove it from the type (as unused).

Good call. I made it clearer in the code (see aee2bda23c7212d97be751b3d4d649034757f565) that we only intend to query open items.

frontend/web/components/GitHubLinkSection.tsx
emyller: I agree, but I realise now these changes actually belong in #7274. I've removed the code from this branch, and will address it in the next one in series.

frontend/web/components/GitHubLinkSection.tsx
emyller: Disclaimer: I am not the author of this code, and I intended minimal-to-no changes to existing code. That said, as per this [other comment](https://github.com/Flagsmith/flagsmith/pull/7273#discussion_r3103034401), I may address this if it's a quick win, but in the next pull request in the series (#7274).

frontend/web/components/GitLabLinkSection.tsx
emyller: Good callout! 6477ce83001c4dd2e054a6658c0a4b83c34879d8

Side note: it bugs me a bit that this module is named `common/types/responses`, and yet, this, and some other types already in it, are not related to API responses.

frontend/web/components/GitLabProjectSelect.tsx
emyller: I think I know where the line is drawn: toast happens if you click a button.

See ed874d6a667349bcb833b05d7451ddda384ceea8 — I wish `ErrorMessage` would support adding a "Retry" action, and a link to _**Integrations**_ as convenience.

frontend/web/components/GitLabProjectSelect.tsx
emyller: 2a1266b4532d9fc2dd5b1493195299d81e317778

frontend/web/components/GitLabSearchSelect.tsx
emyller: c4a4307bc50d5d665155edeadaf2915c264402e0 — nicer indeed, thanks!

emyller:
Thanks for such a great review @Zaimwa9 and @talissoncosta, and for the patience guiding me through 2026 frontend land! All comments addressed — please resolve each thread, or respond with advice. Thanks in advance.

frontend/web/components/GitLabProjectSelect.tsx
emyller: Also added to the PR description:

image

frontend/web/components/GitLabLinkSection.tsx
```diff
@@ -0,0 +1,99 @@
...
+              label: e.label,
+              resourceType: e.resourceType,
+              value: e.id,
+            }))}
+          />
+        </div>
+      </div>
+      {isProjectsError && (
+        <ErrorMessage error='Failed to load GitLab projects' />
+      )}
```
Zaimwa9, line 77: Feel free to resolve.
Your question when to use toast versus ErrorMessage made me think and what could make sense to me is to show a toast following a punctual action that failed (button click etc, PUT / POST) and using an ErrorMessage on a component that can't be used persistently because fetching the data failed without action from the user.

Wdyt @talissoncosta ?

frontend/common/hooks/useHasGitLabIntegration.ts
Zaimwa9: All good then. As the frontend will be release a bit before the backend, I just wanted to make sure we don't show an unusable UI

frontend/web/components/GitLabProjectSelect.tsx
Zaimwa9: https://github.com/Flagsmith/flagsmith/pull/7273#discussion_r3104956993

Ah well, I posted a separate comment but happy to see we reached the same conclusion

Zaimwa9 (approved):
Happy to approve on my side, leaving Talisson give the final approval 👍

talissoncosta (approved):

frontend/web/components/GitLabLinkSection.tsx
emyller: [FTR](https://github.com/Flagsmith/flagsmith/pull/7273#discussion_r3109176691).

PR #7284, opened 2026-04-17
fix: prevent infinite API loop on invalid environment URLs

### LINKED ISSUE #7283: Infinite API request loop when URL points to a non-existent environment
When the URL contains an environment key that doesn't belong to the current project, the app enters an infinite API request loop against that environment, and the sidebar shows a loading spinner that never resolves.

The "invalid key" case covers any of:

- An environment that was deleted
- A typo or manually-edited URL
- An environment from a different project pasted into this project's URL
- An environment the user has lost permission to while the tab was open

### Steps To Reproduce

1. Log into app.flagsmith.com
2. Navigate to any project
3. Edit the URL so the environment segment is a non-existent key, e.g. `/project/<id>/environment/nonexistent/features`
4. Open DevTools → Network tab

**Observed:**

- Requests to `/api/v1/environments/<invalid-key>/` fire every ~1 second indefinitely
- Further requests may appear with the literal string `undefined` as the path segment (e.g. `/api/v1/environments/undefined/my-permissions/`, `/api/v1/environments/undefined/api-keys/`)
- The sidebar shows an infinite loading spinner
- The main area may render partially or stay stuck on skeletons

### Expected behavior

- The app shows an "Environment not found" state in the main area
- No further API requests for the invalid environment
- Sidebar hides or shows a safe empty state (no infinite spinner)
- Navigating to a valid environment via the top tabs or environment selector recovers the app

talissoncosta:

## Changes

Closes #7283

When the URL contained an environment key that didn't belong to the current project (deleted env, typo, env from another project, stale bookmark, or permission revoked mid-session), the app entered an infinite API request loop and showed a sidebar loader that never resolved. Further requests appeared with the literal string `undefined` in the path.

### Root cause

Two bugs compounded:

1. **`EnvironmentReadyChecker` polled every 1s with no error off-ramp.** It only stopped polling on a successful response with `is_creating: false` — on a 4xx it kept polling forever.
2. **Project scoping wasn't checked.** The backend returns a single environment by `api_key` regardless of project scope, so a valid env key belonging to a different project returns `200 OK` and the checker falls through to render the page. Downstream components then fire queries with missing or stale env context — some using `env?.api_key` in URL templates without a fallback, producing `/api/v1/environments/undefined/...` URLs.

Sibling sidebar components (`EnvironmentAside`, `EnvironmentNavbar`) conflated "env list loading" with "env missing from list" and showed loaders forever.

### Fix

- **`EnvironmentReadyChecker`**: source of truth is now the project's env list (`useGetEnvironmentsQuery`). If the URL env isn't in the list, render an "Environment not found" state. The single-env poll is kept for the `is_creating` happy path but now skips once polling is complete or the env is known missing.
- State reset during render (not via `useEffect`) when the env ID changes, so recovering via URL change is instant and doesn't flash stale children.
- Extracted the three view states into `LoadingState`, `CreatingState`, `NotFoundState` components; render flow is now guard clauses instead of nested ternaries.
- Handled the `/environment/create` sentinel so the new-env form isn't caught.
- **`EnvironmentAside`**: hides the whole sidebar when the env isn't in the project list.
- **`EnvironmentNavbar`**: `skip: !environment` on the change-request queries (no more `undefined` URL fetches); returns `null` when env isn't in the list.
- Small cleanup: dropped redundant `` `${projectId}` `` template coerces in the two files touched (fixes pre-existing `string not assignable to number` TS errors).

Manual verification on local dev server (`ENV=local npm run dev`):

- [x] `/project/<id>/environment/<invalid-key>/features` → shows "Environment not found" in the main area, sidebar hidden, no repeat requests in Network tab.
- [x] `/project/<id>/environment/<valid-key>/features` → loads features normally.
- [x] `/project/<id>/environment/create` → create form renders (sentinel preserved).
- [x] From an invalid env → click top tabs / env selector → navigate into a valid env → app recovers without reload.
- [x] Confirmed no `/environments/undefined/...` URLs fire in any of the above flows.

### DIFF
--- a/frontend/web/components/EnvironmentReadyChecker.tsx
+++ b/frontend/web/components/EnvironmentReadyChecker.tsx
@@ -1,8 +1,5 @@
-import { PropsWithChildren, useEffect, useState } from 'react'
-import {
-  useGetEnvironmentQuery,
-  useGetEnvironmentsQuery,
-} from 'common/services/useEnvironment'
+import { PropsWithChildren } from 'react'
+import { useGetEnvironmentQuery } from 'common/services/useEnvironment'
 import { useRouteMatch } from 'react-router-dom'
 
 interface RouteParams {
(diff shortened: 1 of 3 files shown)


Zaimwa9 (approved):
Works as a charm

PR #7287, opened 2026-04-19
fix(search): prevent-local-sync-state-keystroke-loss

### LINKED ISSUE #7199: Feature search bar misses keystrokes

- [ ] Self Hosted with Docker
- [x] SaaS at flagsmith.com

### Describe the bug

When typing in the feature table search bar, some keystrokes/letters might be swallowed

```
I'm typing "spa" everytime, but sometimes it misses a keypress (right at the beginning of the video and in the end).

I'm using Chromium 146.0.7680.80 in the video, but the bug is also present with Firefox 149.0
```

### Steps To Reproduce

1. Go to Feature Table
2. Type repeatedly a string
3. Some letters might not appear

### Expected behavior

All keystrokes should be recorded

### Screenshots

_No response_

Zaimwa9:

Closes #7199 

## Changes
- Drop the value → localValue sync effect in TableSearchFilter that was overwriting in-progress input (root cause of the keystroke eating).
- Add a searchResetKey from useFeatureFilters so "Clear Filters" remounts the input.
- Rewrite useDebounce with a useRef timer to avoid re-render churn and stale closures.

### DIFF
--- a/frontend/common/useDebounce.ts
+++ b/frontend/common/useDebounce.ts
@@ -1,26 +1,31 @@
-import { useState } from 'react'
+import { useCallback, useEffect, useRef } from 'react'
 
 export default function useDebounce(func: any, delay: number) {
-  const [timeout, saveTimeout] = useState<NodeJS.Timeout | null>(null)
+  const timeoutRef = useRef<NodeJS.Timeout | null>(null)
+  const funcRef = useRef(func)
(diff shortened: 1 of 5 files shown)


talissoncosta (approved):
Looks good, thanks for handling it! 👏 😃

PR #7301, opened 2026-04-21
feat(GitLab): Receive webhooks for automatic state sync

### LINKED ISSUE #7165: GitLab integration: receive webhooks for automatic state sync
Context: https://github.com/Flagsmith/flagsmith/issues/7000

When a linked GitLab issue or MR changes state (closed, merged, reopened), Flagsmith should update the feature flag's tags automatically.

### Scope

- Expose a webhook endpoint for GitLab to call.
- Register [project webhooks](https://docs.gitlab.com/api/project_webhooks/) programmatically via the [GitLab API](https://docs.gitlab.com/api/project_webhooks/#add-a-project-hook) when a resource is linked.
- Subscribe to [`issues_events` and `merge_requests_events`](https://docs.gitlab.com/user/project/integrations/webhooks/#events).
- Webhook secrets should be generated by Flagsmith and stored on the configuration model, not provided by the user. The secret is created when Flagsmith registers the webhook via the GitLab API, and used to [validate incoming payloads](https://docs.gitlab.com/user/project/integrations/webhooks/#validate-payloads-by-using-a-secret-token).

### Tags

The following tags mirror the [GitHub integration's tagging behaviour](https://github.com/Flagsmith/flagsmith/blob/main/api/integrations/github/constants.py), adapted for GitLab terminology.

| GitLab event | Tag |
|---|---|
| Issue opened / reopened | `Issue Open` |
| Issue closed | `Issue Closed` |
| MR opened / reopened | `MR Open` |
| MR closed | `MR Closed` |
| MR merged | `MR Merged` |
| MR converted to draft | `MR Draft` |

### Suggested implementation

- Add `TagType.GITLAB`; follow the [GitHub tagging pattern](https://github.com/Flagsmith/flagsmith/blob/main/api/integrations/github/github.py).
- New model to track registered webhooks: GitLab project path, webhook ID, and secret per `GitLabConfiguration`.
- Expose a webhook receiver endpoint. Include the webhook tracking model's ID in the URL so the receiver can look up the secret directly.
- Verify incoming payloads via [`X-Gitlab-Token`](https://docs.gitlab.com/user/project/integrations/webhooks/#validate-payloads-by-using-a-secret-token).
- Register a [project webhook](https://docs.gitlab.com/api/project_webhooks/#add-a-project-hook) when a resource is first linked from a given GitLab project, using the [URL-encoded path](https://docs.gitlab.com/api/rest/index.html#namespaced-paths) extracted from the resource URL.
- Deregister webhooks on `GitLabConfiguration` deletion.

### Acceptance criteria

- [ ] Webhooks are registered automatically — the user does not configure them manually in GitLab
- [ ] Issue state changes in GitLab update the linked feature flag's tags in Flagsmith
- [ ] MR state changes (opened, merged, closed, drafted) update the linked feature flag's tags in Flagsmith
- [ ] Emit structlog events for product telemetry (https://github.com/Flagsmith/flagsmith/issues/7012)

### LINKED ISSUE #7299: GitLab integration: tag feature flags at link time
Context: https://github.com/Flagsmith/flagsmith/issues/7000

When a user links a GitLab issue or MR to a feature flag, the integration only allows linking open resources. The feature flag should receive the corresponding tag (`Issue Open` / `MR Open`) immediately — not after a future webhook event that may never arrive.

The [GitHub integration has the same gap](https://github.com/Flagsmith/flagsmith/blob/main/api/features/feature_external_resources/views.py#L74-L136): it labels the GitHub issue but does not tag the Flagsmith feature at link time.

### Acceptance criteria

- [ ] Linking a GitLab issue applies the `Issue Open` tag to the feature flag
- [ ] Linking a GitLab MR applies the `MR Open` tag to the feature flag

khvn26:

## Changes

Closes #7165. Closes #7299.

Registers a project-level webhook on GitLab when a user links an issue or MR to a feature, and flips the flag's Issue/MR tag whenever GitLab delivers a state-change event. Webhook authenticity is verified via `X-Gitlab-Token` against a Flagsmith-generated secret stored per registered hook.

- New `GitLabWebhook` model: per-`(config, path)` soft-deletable row holding `gitlab_hook_id` and the secret. Unique constraint is conditional on `deleted_at IS NULL` so re-registration after soft-delete works.
- New `TagType.GITLAB` plus `GitLabTagLabel` enum (`Issue Open/Closed`, `MR Open/Closed/Merged/Draft`) with colour and description table.
- `integrations.gitlab.tasks.register_gitlab_webhooks` backfills hooks for existing linked resources on config create/update.
- `integrations.gitlab.tasks.deregister_gitlab_webhooks` cleans up registered hooks on config destroy, looking up the soft-deleted config via `all_with_deleted()`.
- Webhook receiver at `/api/v1/gitlab-webhook/<uuid>/` validates the token, maps the payload through `map_gitlab_webhook_payload_to_tag_label`, and applies the tag via `set_gitlab_tag`. Issue and MR tags coexist independently on a feature — an issue event never touches an MR tag and vice versa.
- Link time: `apply_initial_tag` tags the flag from the client-supplied metadata snapshot so an already-closed issue gets `Issue Closed` immediately without waiting for a webhook event (closes #7299).
- Errors from GitLab during link-time registration raise a 503 via a DRF `GitLabApiUnreachable` `APIException`.

- `make test` — full unit + integration coverage for the model, service layer, tasks, receiver, and link-endpoint flows; `test_gitlab_webhook.py` drives every tag transition (issue open/close/reopen, MR open/merge/close/draft/ready) plus the issue/MR tag-coexistence invariant.
- End-to-end smoke test against real gitlab.com via ngrok:
  - Registered hook on a real GitLab project at link time.
  - Closed an issue in GitLab → receiver flipped the flag to `Issue Closed`.
  - Merged an MR → flag flipped to `MR Merged`.
  - Linked a draft MR → flag tagged `MR Draft`; marking it ready → `MR Open`.
  - Combined feature linked to both an issue and an MR shows both tags side-by-side.

## Screenshots

image

## How to review this PR

Suggested order:

1. `integrations/gitlab/models.py` + migration `0002` — shape of `GitLabWebhook`; note the conditional unique constraint.
2. `integrations/gitlab/constants.py` + `types.py` + `exceptions.py` — small, establish the vocabulary.
3. `integrations/gitlab/mappers.py` — all the "what should the tag be?" logic, pure functions, easy to read as tests.
4. `integrations/gitlab/client/api.py` — two new GitLab HTTP calls (`create_project_hook`, `delete_project_hook`).
5. `integrations/gitlab/services.py` — the orchestration: `ensure_webhook_registered`, `register_webhook_for_resource`, `deregister_webhooks`, `set_gitlab_tag`, `apply_initial_tag`, `apply_tag_for_event`.
6. `integrations/gitlab/tasks.py` — thin `register_gitlab_webhooks` / `deregister_gitlab_webhooks` wrappers over the service; isolation/retry lives at the task processor level.
7. `integrations/gitlab/views/webhook.py` — token check + dispatch.
8. `integrations/gitlab/views/configuration.py` — `perform_create`/`_update` dispatches the backfill task; `perform_destroy` dispatches the deregister task.
9. `features/feature_external_resources/` — the link endpoint's GitLab branch and the `AFTER_SAVE` hook that applies the initial tag.
10. Tests — `test_gitlab_webhook.py` and updates to `test_gitlab_external_resources.py` / `test_configuration.py`.

Things worth a close look:

- Conditional `UniqueConstraint(condition=Q(deleted_at__isnull=True))` on `GitLabWebhook` — allows re-registering for a `(config, path)` pair after the cascade-soft-delete that happens when a config is destroyed.
- The `deregister_webhooks` query uses `all_with_deleted()` because safedelete's cascade has already soft-deleted the rows by the time the task runs.
- `register_webhook_for_resource` silently no-ops on unparseable URLs; the GitLab API failure path raises `GitLabApiUnreachable` (503). Link-time UX: link succeeds iff webhook registration succeeded.
- Receiver validates `X-Gitlab-Token` with `hmac.compare_digest`; Issue and MR tags are scoped independently (`label.startswith("Issue"|"MR")` in `set_gitlab_tag`).

### DIFF
--- a/api/features/feature_external_resources/models.py
+++ b/api/features/feature_external_resources/models.py
@@ -5,6 +5,7 @@ import re
 from django.db import models
 from django.db.models import Q
 from django_lifecycle import (  # type: ignore[import-untyped]
+    AFTER_DELETE,
     AFTER_SAVE,
     BEFORE_DELETE,
     LifecycleModelMixin,
@@ -147,6 +148,25 @@ class FeatureExternalResource(LifecycleModelMixin, models.Model):  # type: ignor
         from integrations.gitlab.services import apply_initial_tag
 
         apply_initial_tag(self)
+
+    @hook(AFTER_DELETE, when="type", is_now="GITLAB_ISSUE")  # type: ignore[misc]
+    @hook(AFTER_DELETE, when="type", is_now="GITLAB_MR")  # type: ignore[misc]
+    def deregister_gitlab_webhook(self) -> None:
+        from integrations.gitlab.models import GitLabConfiguration
+        from integrations.gitlab.services import parse_project_path
+        from integrations.gitlab.tasks import (
+            deregister_gitlab_webhook as deregister_task,
+        )
+
+        project_path = parse_project_path(self.url)
+        if project_path is None:
+            return
+        config = GitLabConfiguration.objects.filter(
+            project=self.feature.project,
+        ).first()
+        if config is None:
+            return
+        deregister_task.delay(args=(config.id, project_path))
 
     @hook(BEFORE_DELETE, when="type", is_now="GITHUB_ISSUE")  # type: ignore[misc]
     @hook(BEFORE_DELETE, when="type", is_now="GITHUB_PR")  # type: ignore[misc]
--- a/api/features/feature_external_resources/models.py
+++ b/api/features/feature_external_resources/models.py
@@ -32,6 +32,12 @@ class ResourceType(models.TextChoices):
     GITLAB_MR = "GITLAB_MR", "GitLab MR"
 
 
+GITLAB_RESOURCE_TYPES: tuple[ResourceType, ...] = (
+    ResourceType.GITLAB_ISSUE,
+    ResourceType.GITLAB_MR,
+)
+
+
 tag_by_type_and_state = {
     ResourceType.GITHUB_ISSUE.value: {
         "open": GitHubTag.ISSUE_OPEN.value,
@@ -134,6 +140,13 @@ class FeatureExternalResource(LifecycleModelMixin, models.Model):  # type: ignor
                 url=None,
                 feature_states=feature_states,
             )
+
+    @hook(AFTER_SAVE, when="type", is_now="GITLAB_ISSUE")  # type: ignore[misc]
+    @hook(AFTER_SAVE, when="type", is_now="GITLAB_MR")  # type: ignore[misc]
+    def apply_gitlab_tag(self) -> None:
+        from integrations.gitlab.services import apply_initial_tag
+
+        apply_initial_tag(self)
 
     @hook(BEFORE_DELETE, when="type", is_now="GITHUB_ISSUE")  # type: ignore[misc]
     @hook(BEFORE_DELETE, when="type", is_now="GITHUB_PR")  # type: ignore[misc]
--- a/api/features/feature_external_resources/views.py
+++ b/api/features/feature_external_resources/views.py
@@ -15,7 +15,8 @@ from integrations.github.client import (
 )
 from integrations.github.models import GitHubRepository
 from integrations.gitlab.models import GitLabConfiguration
-from integrations.gitlab.services import register_webhook_for_resource
+from integrations.gitlab.services import parse_project_path
+from integrations.gitlab.tasks import register_gitlab_webhook
 from organisations.models import Organisation
 
 from .models import GITLAB_RESOURCE_TYPES, FeatureExternalResource, ResourceType
@@ -147,13 +148,12 @@ class FeatureExternalResourceViewSet(viewsets.ModelViewSet):  # type: ignore[typ
         resource_type = serializer.validated_data.get("type")
         if resource_type in GITLAB_RESOURCE_TYPES:
             feature = serializer.validated_data["feature"]
-            if config := GitLabConfiguration.objects.filter(
+            project_path = parse_project_path(serializer.validated_data.get("url"))
+            config = GitLabConfiguration.objects.filter(
                 project=feature.project,
-            ).first():
-                register_webhook_for_resource(
-                    config=config,
-                    resource_url=serializer.validated_data.get("url"),
-                )
+            ).first()
+            if config is not None and project_path is not None:
+                register_gitlab_webhook.delay(args=(config.id, project_path))
 
         resource = serializer.save()
 
--- a/api/features/feature_external_resources/views.py
+++ b/api/features/feature_external_resources/views.py
@@ -145,23 +145,22 @@ class FeatureExternalResourceViewSet(viewsets.ModelViewSet):  # type: ignore[typ
             )
 
     def perform_create(self, serializer: FeatureExternalResourceSerializer) -> None:  # type: ignore[override]
-        resource_type = serializer.validated_data.get("type")
+        resource = serializer.save()
+        resource_type = ResourceType(resource.type)
+
         if resource_type in GITLAB_RESOURCE_TYPES:
-            feature = serializer.validated_data["feature"]
-            project_path = parse_project_path(serializer.validated_data.get("url"))
+            project_path = parse_project_path(resource.url)
             config = GitLabConfiguration.objects.filter(
-                project=feature.project,
+                project=resource.feature.project,
             ).first()
             if config is not None and project_path is not None:
                 register_gitlab_webhook.delay(args=(config.id, project_path))
-
-        resource = serializer.save()
 
         log_event_names: dict[ResourceType, tuple[str, str]] = {
             ResourceType.GITLAB_ISSUE: ("gitlab", "issue.linked"),
             ResourceType.GITLAB_MR: ("gitlab", "merge_request.linked"),
         }
-        if (resource_type := ResourceType(resource.type)) in log_event_names:
+        if resource_type in log_event_names:
             logger_name, event_name = log_event_names[resource_type]
             structlog.get_logger(logger_name).info(
                 event_name,
--- a/api/features/feature_external_resources/views.py
+++ b/api/features/feature_external_resources/views.py
@@ -14,9 +14,11 @@ from integrations.github.client import (
     label_github_issue_pr,
 )
 from integrations.github.models import GitHubRepository
+from integrations.gitlab.models import GitLabConfiguration
+from integrations.gitlab.services import register_webhook_for_resource
 from organisations.models import Organisation
 
-from .models import FeatureExternalResource, ResourceType
+from .models import GITLAB_RESOURCE_TYPES, FeatureExternalResource, ResourceType
 from .serializers import FeatureExternalResourceSerializer
 
 
@@ -72,17 +74,29 @@ class FeatureExternalResourceViewSet(viewsets.ModelViewSet):  # type: ignore[typ
         return Response(data={"results": data})
 
     def create(self, request, *args, **kwargs):  # type: ignore[no-untyped-def]
-        if request.data.get("type") not in [
-            ResourceType.GITHUB_ISSUE,
-            ResourceType.GITHUB_PR,
-        ]:
-            return super().create(request, *args, **kwargs)
+        resource_type = request.data.get("type")
 
         feature = get_object_or_404(
             Feature.objects.filter(
                 id=self.kwargs["feature_pk"],
             ),
         )
+
+        if resource_type in GITLAB_RESOURCE_TYPES:
+            if config := GitLabConfiguration.objects.filter(
+                project=feature.project,
+            ).first():
+                register_webhook_for_resource(
+                    config=config,
+                    resource_url=request.data.get("url"),
+                )
+            return super().create(request, *args, **kwargs)
+
+        if resource_type not in (
+            ResourceType.GITHUB_ISSUE,
+            ResourceType.GITHUB_PR,
+        ):
+            return super().create(request, *args, **kwargs)
 
         github_configuration = (
             Organisation.objects.prefetch_related("github_config")
--- a/api/features/feature_external_resources/views.py
+++ b/api/features/feature_external_resources/views.py
@@ -76,20 +76,8 @@ class FeatureExternalResourceViewSet(viewsets.ModelViewSet):  # type: ignore[typ
     def create(self, request, *args, **kwargs):  # type: ignore[no-untyped-def]
         resource_type = request.data.get("type")
 
-        feature = get_object_or_404(
-            Feature.objects.filter(
-                id=self.kwargs["feature_pk"],
-            ),
-        )
-
+        # GitLab side effects run in ``perform_create`` below.
         if resource_type in GITLAB_RESOURCE_TYPES:
-            if config := GitLabConfiguration.objects.filter(
-                project=feature.project,
-            ).first():
-                register_webhook_for_resource(
-                    config=config,
-                    resource_url=request.data.get("url"),
-                )
             return super().create(request, *args, **kwargs)
 
         if resource_type not in (
@@ -97,6 +85,12 @@ class FeatureExternalResourceViewSet(viewsets.ModelViewSet):  # type: ignore[typ
             ResourceType.GITHUB_PR,
         ):
             return super().create(request, *args, **kwargs)
+
+        feature = get_object_or_404(
+            Feature.objects.filter(
+                id=self.kwargs["feature_pk"],
+            ),
+        )
 
         github_configuration = (
             Organisation.objects.prefetch_related("github_config")
@@ -150,6 +144,17 @@ class FeatureExternalResourceViewSet(viewsets.ModelViewSet):  # type: ignore[typ
             )
 
     def perform_create(self, serializer: FeatureExternalResourceSerializer) -> None:  # type: ignore[override]
+        resource_type = serializer.validated_data.get("type")
+        if resource_type in GITLAB_RESOURCE_TYPES:
+            feature = serializer.validated_data["feature"]
+            if config := GitLabConfiguration.objects.filter(
+                project=feature.project,
+            ).first():
+                register_webhook_for_resource(
+                    config=config,
+                    resource_url=serializer.validated_data.get("url"),
+                )
+
         resource = serializer.save()
 
         log_event_names: dict[ResourceType, tuple[str, str]] = {
--- a/api/integrations/gitlab/services.py
+++ b/api/integrations/gitlab/services.py
@@ -1,0 +1,216 @@
+import re
+import secrets
+import uuid
+from typing import Any
+from urllib.parse import urlparse
+
+import requests
+import structlog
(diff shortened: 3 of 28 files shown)


api/features/feature_external_resources/views.py
```diff
@@ -72,18 +74,30 @@ def list(self, request, *args, **kwargs) -> Response:  # type: ignore[no-untyped
...
 
+        if resource_type in GITLAB_RESOURCE_TYPES:
+            if config := GitLabConfiguration.objects.filter(
+                project=feature.project,
+            ).first():
+                register_webhook_for_resource(
+                    config=config,
+                    resource_url=request.data.get("url"),
+                )
+            return super().create(request, *args, **kwargs)
```
emyller, line 82: Please move this to `perform_create` below:
1. It's the [suggested place](https://www.django-rest-framework.org/api-guide/generic-views/#get_serializer_classself) for side effects, and it already contains some.
2. Keeps this `create` madness isolated to GitHub, and hopefully easier to untangle later on.

api/features/feature_external_resources/models.py
```diff
@@ -135,6 +141,13 @@ def notify_github_on_link(self):  # type: ignore[no-untyped-def]
                 feature_states=feature_states,
             )
 
+    @hook(AFTER_SAVE, when="type", is_now="GITLAB_ISSUE")  # type: ignore[misc]
+    @hook(AFTER_SAVE, when="type", is_now="GITLAB_MR")  # type: ignore[misc]
+    def apply_gitlab_tag(self) -> None:
+        from integrations.gitlab.services import apply_initial_tag
+
+        apply_initial_tag(self)
+
```
emyller, line 151: praise: Nice, thanks!

api/integrations/gitlab/services.py
```diff
@@ -0,0 +1,216 @@
...
+                gitlab__hook__id=webhook.gitlab_hook_id,
+            )
+            webhook.delete()
+        except requests.RequestException as exc:
+            log.warning(
+                "webhook.deregistration_failed",
+                gitlab__project__id=webhook.gitlab_project_id,
+                gitlab__hook__id=webhook.gitlab_hook_id,
+                exc_info=exc,
+            )
```
emyller: Concern-driven attempt:

```suggestion
        try:
            delete_project_hook(
                instance_url=config.gitlab_instance_url,
                access_token=config.access_token,
                project_id=webhook.gitlab_project_id,
                hook_id=webhook.gitlab_hook_id,
            )
        except requests.RequestException as exc:
            log.warning(
                "webhook.deregistration_failed",
                gitlab__project__id=webhook.gitlab_project_id,
                gitlab__hook__id=webhook.gitlab_hook_id,
                exc_info=exc,
            )
        else:
            log.info(
                "webhook.deregistered",
                gitlab__project__id=webhook.gitlab_project_id,
                gitlab__hook__id=webhook.gitlab_hook_id,
            )
            webhook.delete()
```

api/integrations/gitlab/services.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+def set_gitlab_tag(feature: Any, new_label: GitLabTagLabel) -> None:
+    """Apply a GitLab system tag to ``feature``, replacing any existing GitLab
+    tag of the same kind (Issue/MR) first. Issue tags and MR tags coexist
+    independently — an Issue event won't touch an MR tag or vice versa.
+    """
+    label_value = new_label.value
+
+    kind_prefix = label_value.split(" ", 1)[0]  # "Issue" or "MR"
```
emyller: This looks a bit fragile. How about prefix coming from some attribute in `GitLabTagLabel`, e.g. `issue_prefix`, `merge_request_prefix`?

api/integrations/gitlab/services.py
```diff
@@ -0,0 +1,216 @@
...
+        organisation__id=webhook.gitlab_configuration.project.organisation_id,
+        project__id=webhook.gitlab_configuration.project_id,
+    )
+    log.info(
+        "feature.tagged",
+        feature__id=feature.id,
+        tag__label=label.value,
+        object_kind=payload.get("object_kind"),
+        action=attrs.get("action"),
+    )
```
emyller, line 235: Unrelated question related to telemetry: what is the expected output in the catalogue if this event is used multiple times through the code base but with varying attributes?

api/integrations/gitlab/tasks.py
```diff
@@ -0,0 +1,43 @@
...
+    )
+
+    urls = set(
+        FeatureExternalResource.objects.filter(
+            feature__project=config.project,
+            type__in=GITLAB_RESOURCE_TYPES,
+        ).values_list("url", flat=True)
+    )
+    for url in urls:
+        register_webhook_for_resource(config=config, resource_url=url)
```
emyller: This looks sub-optimal:

1. One GitLab issue/MR can be linked to multiple features, leading to the same URL showing up multiple times.
2. Ensuring a webhook is registered for all existing GitLab links within a project sounds like more cycles than needed, given webhooks are set per repository, rather than per issue or MR.

I'd suggest dispatching tasks from linking / unlinking actions only, and updating signatures of `register_gitlab_webhooks` and `deregister_gitlab_webhooks` to:

- `def register_gitlab_webhook(config_id, project_path)` — called at link time.
- `def deregister_gitlab_webhook(config_id, project_path)` — called at unlink time, no-op if other features are still listening for updates.

Keep in mind the GitLab API usually accepts project ID and path interchangeably — might make this easier.

frontend/web/components/GitLabLinkSection.tsx
```diff
@@ -59,6 +59,7 @@ const GitLabLinkSection: FC<GitLabLinkSectionProps> = ({
           metadata: {
             state: selectedItem.state,
             title: selectedItem.title,
+            ...('draft' in selectedItem ? { draft: selectedItem.draft } : {}),
```
emyller, line 62:
```suggestion
            draft: selectedItem.draft ?? false,
```

api/integrations/gitlab/types.py
```diff
@@ -0,0 +1,25 @@
...
+    state: str
+    draft: bool
+
+
+class GitLabWebhookObjectAttributes(TypedDict, total=False):
+    url: str
+    state: str
+    action: str
+    draft: bool
+    work_in_progress: bool
```
emyller, line 20: `work_in_progress` is [deprecated](https://docs.gitlab.com/user/project/integrations/webhook_events/#deprecated-fields). Doesn't hurt but should we keep it?

api/tests/integration/features/test_gitlab_external_resources.py
```diff
@@ -115,6 +122,11 @@ def test_create_external_resource__gitlab_issue_with_github_also_configured__ret
         gitlab_instance_url="https://gitlab.com",
         access_token="glpat-test-token",
     )
+    responses.post(
+        "https://gitlab.com/api/v4/projects/testorg%2Ftestrepo/hooks",
+        json={"id": 1, "project_id": 1},
+        status=201,
+    )
```
emyller, line 127: Is this necessary for this **specific** test to pass?

api/tests/integration/features/test_gitlab_external_resources.py
```diff
@@ -131,17 +143,219 @@ def test_create_external_resource__gitlab_issue_with_github_also_configured__ret
     assert response.status_code == status.HTTP_201_CREATED
     assert response.json()["type"] == "GITLAB_ISSUE"
 
+
+@responses.activate
+def test_create_external_resource__gitlab_issue_with_config__registers_webhook_and_tags_feature(
```
emyller:
```suggestion
def test_create_external_resource__gitlab_issue__registers_webhook_and_tags_feature(
```

Non-blocking but what do you think about splitting concerns across separate tests for both issues and merge requests?

Out of scope, this test feels like LLM went free style and still managed to pass lint checks.

api/tests/integration/features/test_gitlab_webhook.py
```diff
@@ -0,0 +1,394 @@
...
+        resource_type: ResourceType,
+        metadata: dict[str, Any] | None = None,
+    ) -> None: ...
+
+
+@pytest.fixture()
+def gitlab_webhook(project: int) -> GitLabWebhook:
+    project_instance = Project.objects.get(id=project)
+    config = GitLabConfiguration.objects.create(
+        project=project_instance,
```
emyller: nit:
```suggestion
    config = GitLabConfiguration.objects.create(
        project_id=project,
```

api/tests/integration/features/test_gitlab_webhook.py
```diff
@@ -0,0 +1,394 @@
...
+    )
+    return webhook
+
+
+@pytest.fixture()
+def webhook_url(gitlab_webhook: GitLabWebhook) -> str:
+    return reverse(
+        "api-v1:gitlab-webhook",
+        kwargs={"webhook_uuid": str(gitlab_webhook.uuid)},
+    )
```
emyller:
```suggestion
    return f"/api/v1/gitlab-webhook/{gitlab_webhook.uuid}/"
```

Good opportunity to ensure the URL actually looks right.

api/tests/integration/features/test_gitlab_webhook.py
```diff
@@ -0,0 +1,394 @@
...
+                "type": resource_type.value,
+                "url": resource_url,
+                "feature": feature,
+                "metadata": metadata or {},
+            },
+            format="json",
+        )
+        assert response.status_code == status.HTTP_201_CREATED, response.content
+
+    return _link
```
emyller, line 71: question: Given this URL is tested, can we benefit from cutting down cycles to just creating a `FeatureExternalResource` object here instead?

emyller (changes requested):
Looks great. Minor adjustments needed.

api/tests/integration/features/test_gitlab_webhook.py
khvn26: It’s an integration test; IMO patching through API is appropriate here.

Besides, for some of the tests we’d also need to attach tags, i.e. recreate the business logic.

api/integrations/gitlab/services.py
khvn26: All occurences, and a superset of used attributes, are documented.

frontend/web/components/GitLabLinkSection.tsx
khvn26: Doesn’t make sense for an issue?

emyller: True, please dismiss.

api/tests/integration/features/test_gitlab_webhook.py
emyller: Agreed. This question came from looking at the fixtures `gitlab_webhook` and `link_feature`, each taking a different approach to create objects. Non blocking — I don't really mind — but would like to see more consistency, or an explained difference.

api/tests/integration/features/test_gitlab_external_resources.py
khvn26: Yes — `task_processor` runs synchronously in tests, and linking a GitLab resource now registers a webhook in `perform_create`. Without the mock, the HTTP call would leak (or fail) and the 201 assertion would turn into a red herring.

api/integrations/gitlab/types.py
khvn26: Leaning towards keeping it. Self-hosted GitLab releases lag, so `work_in_progress` still lands on older instances; the fallback is a single `or` clause and costs us nothing. Happy to drop it if you feel strongly, though.

api/tests/integration/features/test_gitlab_webhook.py
khvn26: Fair. The split is deliberate but worth calling out: `gitlab_webhook` creates the ORM rows directly because these tests exercise the webhook **receiver**, so the GitLab-side registration dance (HTTP + task dispatch) is noise. `link_feature` goes through the API because it also applies the initial tag and serialises metadata — things the tests assert on further down. Happy to leave a short comment on each fixture explaining the choice.

api/integrations/gitlab/types.py
emyller: Happy to keep it!

api/tests/integration/features/test_gitlab_webhook.py
emyller: Nah this thread is good enough. Happy to move on if you are too. Thanks!

api/features/feature_external_resources/views.py
khvn26: Moved in 7c5c37ea1.

api/integrations/gitlab/services.py
khvn26: Done in 4e9ba5a60 — split into `try/except/else`, success branch now only runs when GitLab actually accepted the delete.

api/integrations/gitlab/services.py
khvn26: Addressed in f70ee1328 — prefix now comes from an explicit `GITLAB_TAG_KIND_BY_LABEL` mapping in `constants.py`, no more slicing labels.

api/integrations/gitlab/tasks.py
khvn26: Switched to per-link/unlink tasks in 6e69f1133. `register_gitlab_webhook(config_id, project_path)` fires from `perform_create`; `deregister_gitlab_webhook(config_id, project_path)` fires from the unlink hook and from config destroy, and no-ops if another linked resource in the same Flagsmith project still references that GitLab path. No more scanning all linked resources on every save.

api/tests/integration/features/test_gitlab_external_resources.py
khvn26: Split in b789c15bf — issue and MR now live in separate tests with explicit assertions on the tag, metadata, and registered webhook row.

api/tests/integration/features/test_gitlab_webhook.py
khvn26: Done in a2594bc56 — URL is hardcoded so the route name is now exercised by the tests.

api/tests/integration/features/test_gitlab_webhook.py
khvn26: Applied in de4b5a531.

api/features/feature_external_resources/views.py
emyller: nit: Is there a reason why the webhook needs to be registered before the instance is created? The `resource` object is available a couple lines below, and using it would look better IMO.

Feel free to resolve either way.

khvn26: Good call — moved in d1bf876e8. Also drops the `feature = validated_data["feature"]` lookup since `resource.feature.project` is already available.

emyller (approved):
LFGTM

PR #7311, opened 2026-04-22
refactor: scope-aware project and organisation integrations

### LINKED ISSUE #7280: GitHub integration is org-wide but managed at project level in the UI

- [ ] Self Hosted with Docker
- [ ] SaaS at flagsmith.com

### Describe the bug

The GitHub integration behaves as an organisation-wide integration, but it is configured and managed from the **project** integrations page. It does not appear on the **organisation** integrations page at all. This is
  confusing, inconsistent with other integrations, and a permissions issue.

 Some of our integrations (e.g. analytics providers) exist at the org level but **can be overridden per project**
  — a project-level configuration takes precedence over the org-level one. That model works because each project  can legitimately point at a different destination.

  GitHub does not work this way. Flagsmith's access to a GitHub organisation's repositories is granted **once,
  globally, via the GitHub App installation**. There is no meaningful per-project override — the connection itself
  must live at the org level. The only project-scoped concern is which GitHub repositories a given Flagsmith
  project maps to (1:many).

### Steps To Reproduce

  1. Create a new Flagsmith project in an organisation where GitHub has previously been set up on another project.
  2. Navigate to **Organisation Settings → Integrations**.
  3. Navigate to **Project Settings → Integrations**.

  - GitHub is **not listed** on the organisation integrations page.
  - GitHub **is listed** on the project integrations page, and shows as already set up — even though the
  integration was never configured for this specific project.
  - A user with project-level permissions can modify or remove the GitHub integration, which affects the entire
  organisation and all its projects.

### Expected behavior

 Users spend far more time in projects than in organisation settings, so moving GitHub out of the project
  integrations page risks users assuming the integration doesn't exist. To mitigate this, the project integrations
  page should:

- Add a field to remote config to denote that GitHub is not 
  - **Keep the GitHub tile visible** on the project integrations page so it remains findable from where users
  actually work. Do this by setting
```
    "project": false,
    "organisation": true,
```

In the remote config ``integration_data``

  - For integrations connected at a different level **Show connection status** clearly (e.g. "Connected via organisation" / "Not configured — set up at
  organisation level").
  - **Replace the configure CTA** with a link to the organisation integrations page. The CTA should respect
  org-level permissions — users without them see a read-only state and a hint about who to contact. Wrap this cta in the relevant renderWithPermissions util.
  - **Keep project-scoped controls** (Flagsmith-project ↔ GitHub-repo mapping, which is 1:many) on the project page, since those genuinely belong at project level.
  - **Cross-link from the org page back to project repo mapping**, so an admin setting up the integration for the  first time is guided to the next step.
  - **Empty-state messaging** on the project page when no org-level integration exists should explain that GitHub  is set up at the organisation level, and link there.

### Screenshots

_No response_

kyle-ssg:

## Changes

  - Project page now surfaces org-level integrations — org-only ones deep-link to the org page, or show "Manage Repositories" if already
  connected (fixes the GitHub dead-end).
  - Org page now surfaces project-level integrations — with a project picker in the modal that auto-loads existing config.
  - Org admins can create and edit integrations for all their projects from the organisation view
  - Permission-aware CTAs with tooltips for non-org-admins.
  - Cleaner cards: action menu replaces the button row; kebab on each active integration for Edit/Delete.
  - Dropped the InfoMessage blurbs — the UI now explains itself.
  - Migrate integrations to RTK

## Project integrations page

image

## Organistaion integrations page with the ability to create/edit at project level

image

https://github.com/user-attachments/assets/8370c8cd-aaa3-4d0a-ad9d-f85800b10fc3

- Added and edited integrations


frontend/web/components/modals/CreateEditIntegrationModal.tsx
```diff
@@ -11,11 +11,43 @@ import Project from 'common/project'
...
+)
+
+const ReadOnlyEnvironmentName: FC<{ environmentId?: string }> = ({
+  environmentId,
+}) => {
+  const { data: environment } = useGetEnvironmentQuery(
+    { id: environmentId as any },
+    { skip: !environmentId },
+  )
+  return <>{environment?.name || ''}</>
```
talissoncosta: Is `environmentId` here the api_key or the numeric id? `Req['getEnvironment']` is `{ id: number }`, so the `as any` looks like it's papering over a mismatch. If it's the api_key, could we rename to `environmentApiKey: string` to match the standardisation ?

frontend/common/types/requests.ts
```diff
@@ -936,6 +936,26 @@ export type Req = {
     code_challenge_method: string
     state?: string
   }
+  getIntegration: {
+    integrationId: string
+    projectId?: string
+    environmentId?: string
```
talissoncosta: Could you confirm the value is the api_key? If so, can we name the field `environmentApiKey: string` to align?

frontend/web/components/IntegrationList.tsx
```diff
@@ -241,6 +362,22 @@ const Integration: FC<IntegrationProps> = (props) => {
                   integration={props.integration}
                 />
               </Flex>
+              <div onClick={(e) => e.stopPropagation()}>
+                <IntegrationActionsMenu
+                  actions={[
```
talissoncosta: Can you please check if this component can be replaced by `DropdownMenu` ?

 
```suggestion
   <DropdownMenu
                  items={[
```

frontend/web/components/IntegrationList.tsx
```diff
@@ -128,100 +205,144 @@ const Integration: FC<IntegrationProps> = (props) => {
...
+      if (action.requiresOrgAdmin) {
+        return Utils.renderWithPermission(
+          !!props.isOrgAdmin,
+          'Organisation admin permission is required to manage organisation integrations.',
+          button,
+        )
+      }
+      return button
+    }
+    return <IntegrationActionsMenu actions={actions} />
```
talissoncosta: Can you please check if this component can be replaced by DropdownMenu ?

```suggestion
    return <DropdownMenu items={actions} />
```

frontend/web/components/modals/CreateEditIntegrationModal.tsx
```diff
@@ -145,55 +251,53 @@ const CreateEditIntegration: FC<CreateEditIntegrationProps> = (props) => {
...
+      const payload = err?.data ?? err
+      setError(
+        (Array.isArray(payload) ? payload[0] : undefined) || defaultError,
+      )
+    } catch {
+      setError(defaultError)
+    } finally {
+      setIsLoading(false)
+    }
   }
```
talissoncosta, line 314: The `try/catch/finally` here isn't doing anything — nothing inside the `try` can throw now that we're not parsing JSON. Could flatten to its body.

frontend/web/components/pages/IntegrationsPage.tsx
```diff
@@ -34,36 +35,32 @@ const IntegrationsPage: FC = () => {
...
   }, [])
 
-  const integrations = Object.keys(Utils.getIntegrationData())
+  const integrationData = Utils.getIntegrationData()
+  const integrations = Object.keys(integrationData).sort((a, b) =>
+    (integrationData[a]?.title || a).localeCompare(
+      integrationData[b]?.title || b,
+    ),
+  )
+  const organisationId = AccountStore.getOrganisation()?.id
```
talissoncosta: Can we get orgId from another source to avoid getting this data from flux ?

frontend/web/components/pages/OrganisationIntegrationsPage.tsx
```diff
@@ -34,15 +49,21 @@ const OrganisationIntegrationsPage = ({ match }) => {
...
+      {!!projectIntegrations.length && (
+        <>
+          <h5 className='mt-5 mb-3'>Project-level integrations</h5>
+          <IntegrationList
+            organisationId={organisationId}
+            integrations={projectIntegrations}
+            isOrgAdmin={isOrgAdmin}
+          />
+        </>
+      )}
```
talissoncosta, line 66: I tried this end-to-end and the UX confused me. Steps:

1. Opened the org integrations page.
2. In the "Project-level integrations" section, clicked Add on a project-level integration, picked a project, filled in the fields, hit save.
3. Nothing appeared on the org page after closing the modal — I thought the save had failed.
4. Navigated to that project's integrations page, and the integration was there.

Is this the intended behaviour?

talissoncosta (changes requested):
Thanks for handling that @kyle-ssg. 

I've added some comments, mostly clarifications..

frontend/web/components/pages/IntegrationsPage.tsx
kyle-ssg: Totally right, forgot router context had this

frontend/web/components/pages/OrganisationIntegrationsPage.tsx
kyle-ssg: Ok yep, I've added a toast. We are slightly limited in what we can show here since integrations are ``number of integrations * number of projects * number of environments`` api calls.

I think the following dismissible UI as well as a toast solves this 

image

talissoncosta (approved):
Looks great, thanks for addressing the feedbacks!

PR #7325, opened 2026-04-23
feat(Admin): Register LaunchDarkly import requests

khvn26:

## Changes

When a LaunchDarkly import gets wedged without a terminal `status.result`, the partial unique constraint on the import requests table blocks the customer from trying again for the same project + LD project key. Today the only fix is an engineer running a Django shell in prod. This puts the import requests in Django admin with a "mark as failed" action so customer success can unstick customers themselves.

Unit tests in `tests/unit/integrations/launch_darkly/test_admin.py` cover the action's three branches — pending gets marked and `completed_at` filled in; pending with existing `completed_at` keeps it; already-resolved rows are skipped with a count in the admin message.

### DIFF
--- a/api/integrations/launch_darkly/admin.py
+++ b/api/integrations/launch_darkly/admin.py
@@ -1,0 +1,69 @@
+from django.contrib import admin, messages
+from django.db.models import QuerySet
+from django.http import HttpRequest
+from django.utils import timezone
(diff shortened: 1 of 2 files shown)


Zaimwa9 (approved):
Thanks

PR #7306, opened 2026-04-22
feat(GitLab): Post comments when linking an issue or merge request to a Flagsmith feature

emyller:
Thanks for submitting a PR! Please check the boxes below:

- [x] I have read the [Contributing Guide](/Flagsmith/flagsmith/blob/main/CONTRIBUTING.md).
- [ ] ~I have added information to `docs/` if required so people know about the feature.~
- [x] I have filled in the "Changes" section below.
- [x] I have filled in the "How did you test this code" section below.

## Changes

Closes https://github.com/Flagsmith/flagsmith/issues/7162

Post comments to GitLab issues and merge requests when:
- A feature is linked to an issue or MR (initial state snapshot table across environments)
- A linked feature's state changes for any environment, segment, or identity
- A feature is unlinked from an issue or MR
- A linked feature is deleted

State-change comments render as a snapshot of the post-change state (e.g. ``in **Development**: :white_check_mark: Enabled, value `X` ``) rather than describing the transition, which avoids misrepresenting value-only changes as enable/disable events.

## How did you test this code?

Unit tests covering the new service, task handlers, and hook call sites (patch coverage 100%).

Manual smoke against gitlab.com, posting to `flagsmith/webhooks-tetsing-1`:
- Issue [#2](https://gitlab.com/flagsmith/webhooks-tetsing-1/-/issues/2) — link → enable → value change → unlink → re-link → delete
- MR [!3](https://gitlab.com/flagsmith/webhooks-tetsing-1/-/merge_requests/3) — link → segment-override state change → identity-override state change

## Screenshots

image

image

Issue #7315, opened 2026-04-22
Consolidate VCS integration orchestration


khvn26:
The same logical events (link, unlink, feature state changed, feature deleted) dispatch per-provider from multiple unrelated sites today. Adding a new VCS (Bitbucket, say) means touching all of these files and re-learning the conventions each time.

## Current state

After #7306 and #7327, `integrations/vcs/services.py` hosts:

- `dispatch_vcs_on_resource_create(resource)`
- `dispatch_vcs_on_resource_destroy(resource)`

`FeatureExternalResourceViewSet.perform_create` and `.perform_destroy` delegate to these. GitLab side effects (log, webhook register/deregister, initial tag, tag clear, link/unlink comment) all flow through the dispatcher via namespaced calls (`from integrations.gitlab import services as gitlab`), and the GitLab-specific `apply_gitlab_tag` / `deregister_gitlab_webhook` lifecycle hooks on `FeatureExternalResource` are gone.

## Still to do

View-level, outside the dispatcher:

- `FeatureExternalResourceViewSet.create` still owns the GitHub URL preflight regex and inline `label_github_issue_pr` call.

Serializer-level:

- State change: `FeatureStateSerializerBasic.save` calls `call_github_task` (GitHub) and `post_gitlab_state_change_comment_for_feature_state` (GitLab) directly. No `dispatch_vcs_on_feature_state_changed` yet.
- `features/versioning/serializers.py` — same split for the v2 versioning flow.

Model-level hooks still present — consolidate to single VCS hook per model:

- `FeatureExternalResource.notify_github_on_link` (`AFTER_SAVE`, type is `GITHUB_ISSUE`/`GITHUB_PR`) — applies GitHub tag + link comment.
- `FeatureExternalResource.notify_github_on_unlink` (`BEFORE_DELETE`, type is `GITHUB_ISSUE`/`GITHUB_PR`) — unlink comment to GitHub.
- `Feature.create_github_comment` (`AFTER_SAVE`) — flag-deleted comment to GitHub.
- `Feature.create_gitlab_comment` (`AFTER_SAVE`) — flag-deleted comment to GitLab.

## Proposed shape for the rest

Extend `integrations/vcs/services.py` with:

- `dispatch_vcs_on_feature_state_changed(feature_state)`
- `dispatch_vcs_on_feature_deleted(feature_name, feature_id, project_id)`

Port each remaining trigger behind these:

[...]

Issue #7162, opened 2026-04-07
GitLab integration: post comments on linked issues and MRs


emyller:
Context: https://github.com/Flagsmith/flagsmith/issues/7000

When a feature flag is linked to a GitLab issue or MR (https://github.com/Flagsmith/flagsmith/issues/7160), Flagsmith should post a comment showing the flag's current state across all environments. When the flag state changes or the flag is deleted, a new comment should be posted automatically.

### Scope

- Post to issues via [`POST /api/v4/projects/:id/issues/:iid/notes`](https://docs.gitlab.com/api/notes/#create-new-issue-note).
- Post to MRs via [`POST /api/v4/projects/:id/merge_requests/:iid/notes`](https://docs.gitlab.com/api/notes/#create-new-merge-request-note).
- Trigger on: flag linked, flag state changed, flag deleted, flag unlinked.

> [!NOTE]
> PR https://github.com/Flagsmith/flagsmith/pull/7122 contains an async task (`post_gitlab_comment`) and dispatch function (`dispatch_gitlab_comment`) that can be used as a starting point. A few things to consider:
>
> - The PR extracts GitHub's `generate_body_comment` into a shared `integrations/vcs/comments.py` module. This refactors [`integrations/github/github.py`](https://github.com/Flagsmith/flagsmith/blob/589c5d57e9e97c603684f733f67e8cbeb2934b52/api/integrations/github/github.py#L125). `generate_body_comment` currently has no tests — they need to be added.
> - **Incomplete per author:** the hooks that auto-trigger comments are not done. The GitHub integration calls `call_github_task` from six places across `features/models.py`, `features/serializers.py`, `features/versioning/serializers.py`, and `features/feature_external_resources/models.py`. The GitLab equivalent of these hooks is new work.

### Acceptance criteria

- [x] Comment posted to GitLab when a flag is linked to an issue or MR
- [x] Comment posted when the linked flag's state changes in any environment
- [x] Comment posted when the linked flag is deleted
- [x] Comment posted when a flag is unlinked from an issue or MR
- [x] Comment generation logic is tested
- [x] Emit structlog events for product telemetry (https://github.com/Flagsmith/flagsmith/issues/7012)

SahilJat:
Hey @emyller ! I'd love to take this on, but I see it depends on PR #7122. Can you assign this to me, and I'll start working on it as soon as that PR is merged?

emyller:
[...]

PR #7327, opened 2026-04-23
fix(GitLab): Orphaned feature tag after unlinking an issue or merge request

khvn26:

## Changes

Follow-up to #7306.

Linking a feature to a GitLab issue or merge request applies a GitLab tag (e.g. `Issue Open`, `MR Draft`). Unlinking left that tag on the feature with no corresponding `FeatureExternalResource`.

Introduce `clear_tag_for_resource` in the GitLab tagging service and wire it into `dispatch_vcs_on_resource_destroy`. It removes the Issue/MR-kind tag from the feature only when no other linked FER of the same kind remains — features linked to several issues, or to both an issue and an MR, keep the still-valid tag. The helper excludes `resource.pk` from its survivor query so it's safe to call regardless of whether the caller has already destroyed the row.

Review effort: 1/5

Unit tests for the new helper cover:
- single FER of a kind → tag cleared
- multiple FERs of the same kind → tag retained
- mixed Issue + MR → only the unlinked kind's tag cleared
- non-GitLab type → no-op

### DIFF
--- a/api/integrations/gitlab/constants.py
+++ b/api/integrations/gitlab/constants.py
@@ -1,5 +1,7 @@
 from enum import Enum
 
+from features.feature_external_resources.models import ResourceType
+
 GITLAB_TAG_COLOR = "#FC6D26"
 GITLAB_CLIENT_TIMEOUT_SECONDS = 10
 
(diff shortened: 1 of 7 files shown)


Zaimwa9 (dismissed):

Zaimwa9 (dismissed):

emyller (changes requested):
Sorry, something smells in the architecture of this.

For instance, I don't like that we import the component `clear_tag_for_resource`, having such a generic name, into `integration.vcs`, but it's from GitLab domain. And I don't think the fix here is to rename GitLab service functions with a `gitlab_` namespace — although that would be cool since they're exported out of `integrations.gitlab` it seems?

I think we should define a protocol/interface (via OOP?) that clears out the list of all use cases involved in linking to GitHub and GitLab currently, allowing us to `class BitBucketService(VCSIntegrationService):` later. This part might fall into the scope of https://github.com/Flagsmith/flagsmith/issues/7315.

I appreciate this should have been raised in https://github.com/Flagsmith/flagsmith/issues/7162, but I was not around, sorry.

emyller (approved):
Looks great now. Thank you!

PR #7335, opened 2026-04-24
fix(Billing): Free trial buttons stop working after toggling yearly/monthly

khvn26:

## Changes

Contributes to https://github.com/Flagsmith/flagsmith-private/issues/122

On the organisation Billing tab, the "14 Day Free Trial" buttons only work on first page load in Yearly mode. Toggling to Monthly — or back to Yearly — leaves the buttons inert.

Chargebee binds click handlers to `[data-cb-type="checkout"]` elements when `Chargebee.registerAgain()` scans the DOM. `initChargebee` was guarded by a one-shot `initialised` flag, so `registerAgain()` only ran once. The toggle in `Payment.tsx` remounts `PaymentButton` (its `key` is the `data-cb-plan-id`), and the freshly rendered buttons never get wired up.

Fix: keep the one-time work (`Chargebee.init`, first-promoter, checkout callbacks, plan-cookie flow) behind the `initialised` guard, split it out as `initChargebee`, and have the new exported `bindChargebeeButtons` call `Chargebee.registerAgain()` on every entry. Add `yearly` to the effect's dependency array so the scan re-runs after the toggle.

Also adds a new `@saas` E2E marker for SaaS-only tests (billing, trials, payments), wired into the staging and production E2E runs, and a dedicated `e2e_billing_user` seeded on a Free-plan organisation so the trial-button path under test is actually reachable.

Added `frontend/e2e/tests/billing-test.pw.ts` (tagged `@saas`). It logs in as `e2e_billing_user`, stubs `window.Chargebee` via `addInitScript` so `registerAgain()` records which `data-cb-plan-id` is clicked, and asserts the trial button fires a checkout in yearly → monthly → yearly order. Without the fix the second click records nothing; with the fix all three are recorded.

Ran the test manually against real Vercel deployments:

- **`staging.flagsmith.com` (main, unfixed)** — test fails at the expected assertion. Yearly click records `startup-annual-v2`, toggle to monthly, click records nothing: `["startup-annual-v2"]` vs expected `["startup-annual-v2", "startup-v2"]`. Bug reproduced.
- **PR's Vercel staging preview (this branch, fixed)** — test passes end to end: login, billing tab, toggle yearly → monthly → yearly, all three clicks recorded. 6.6s.

Manual smoke in-browser: billing tab, click "14 Day Free Trial" in yearly (opens checkout), toggle to monthly and click (opens checkout), toggle back to yearly and click (opens checkout).

## Follow-up

Hooking `@saas` tests up to run against Vercel PR previews on every PR (not just staging/prod deploys) will come in a separate PR — it needs a workflow that waits for the Vercel preview to finish and passes its URL to the composite E2E action, which is out of scope for this bug fix.

### DIFF
--- a/frontend/e2e/tests/billing-test.pw.ts
+++ b/frontend/e2e/tests/billing-test.pw.ts
@@ -69,7 +69,7 @@ test.describe('Billing', () => {
     await click(byId('org-settings-link'))
     await click(byId('billing'))
 
-    const trialButton = () =>
+    const getTrialButton = () =>
       page
         .locator('button[data-cb-type="checkout"]')
         .filter({ hasText: '14 Day Free Trial' })
@@ -77,7 +77,7 @@ test.describe('Billing', () => {
     const toggle = page.getByRole('switch').first()
 
     log('Confirm free-trial button is wired in yearly mode')
-    const yearlyButton = trialButton()
+    const yearlyButton = getTrialButton()
     await yearlyButton.waitFor({ state: 'visible' })
     const yearlyPlanId = await yearlyButton.getAttribute('data-cb-plan-id')
     expect(yearlyPlanId).toBeTruthy()
@@ -88,7 +88,7 @@ test.describe('Billing', () => {
 
     log('Toggle to Monthly and click the free-trial button')
     await toggle.click()
-    const monthlyButton = trialButton()
+    const monthlyButton = getTrialButton()
     await monthlyButton.waitFor({ state: 'visible' })
     const monthlyPlanId = await monthlyButton.getAttribute('data-cb-plan-id')
     expect(monthlyPlanId).toBeTruthy()
@@ -102,7 +102,7 @@ test.describe('Billing', () => {
 
     log('Toggle back to Yearly and click the free-trial button')
     await toggle.click()
-    const yearlyButtonAgain = trialButton()
+    const yearlyButtonAgain = getTrialButton()
     await yearlyButtonAgain.waitFor({ state: 'visible' })
     expect(await yearlyButtonAgain.getAttribute('data-cb-plan-id')).toBe(
       yearlyPlanId,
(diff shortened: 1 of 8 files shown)


frontend/e2e/tests/billing-test.pw.ts
```diff
@@ -0,0 +1,115 @@
...
+    await login(E2E_USER, PASSWORD)
+
+    log('Navigate to Billing tab')
+    await waitForElementVisible(byId('organisation-link'))
+    await click(byId('organisation-link'))
+    await waitForElementVisible(byId('org-settings-link'))
+    await click(byId('org-settings-link'))
+    await click(byId('billing'))
+
+    const trialButton = () =>
```
Zaimwa9: NIT: Super minor, i feel ashamed to ask but `getTrialButton` would be more accurate

frontend/web/components/modals/payment/chargebee.ts
```diff
@@ -8,30 +8,34 @@ type InitChargebeeParams = {
 }
 
 export const initChargebee = ({ isPaymentsEnabled }: InitChargebeeParams) => {
```
Zaimwa9: I'd like to take the opportunity to make it more readable and easier to maintain.

What do you think about: 
1. Renaming `initChargebee` to `bindChargebeeButtons`
2. Extract our internal initialisation into a new function re-using the name `initChargebee`

Have only 
```
export const bindChargebeeButtons({ isPaymentsEnabled }: InitChargebeeParams) => {
...
  if (!initialised) {
    initChargebee()
  }
  Chargebee.registerAgain()
}
```

Zaimwa9 (dismissed):
Approving with a NIT and a small refacto that I leave up to you.

And good call adding the E2E

frontend/e2e/tests/billing-test.pw.ts
khvn26: Done in abdc85b89.

frontend/web/components/modals/payment/chargebee.ts
khvn26: Done in 11045aacf.

Zaimwa9 (commented):
thanks! 💯 

Good call to separate user and orgs in the seed

Zaimwa9 (approved):

PR #7342, opened 2026-04-24
fix: render all top-level segment rules in the UI instead

### LINKED ISSUE #7254: Segment UI only renders first top-level rule, silently drops additional rules
## Bug Description

When a segment is created via the Admin API with multiple top-level rules (i.e., the `rules` array at the segment level has 2+ entries), the dashboard UI only renders the first top-level rule. Additional rules are completely invisible in the visual editor there is no scroll and no indicator.

The API accepts and returns all rules correctly. The data is stored and the flag engine evaluates all rules. But the UI silently hides everything after `rules[0]`.

Editing the segment in the dashboard would silently drop the hidden rules.

## Steps to Reproduce

1. Create a segment via Admin API with two top-level rule groups:

POST /api/v1/projects/{PROJECT_ID}/segments/

With two entries in the top level rules array, each with type ALL and containing sub-rules scoped to different environments via $.environment.name.

2. Verify the API response contains both top level rules
3. Open the segment in the Flagsmith dashboard

## Expected Behavior

The UI should display both top level rule groups, or at minimum indicate that additional rules exist.

## Actual Behavior

Only the first top level rule group is rendered. The second is completely missing from the editor.

## Root Cause

The frontend hard codes `rules[0]` throughout `CreateSegment.tsx`:

  - addRule() (line ~214): appends to `rules[0].rules`
  - updateRule() (line ~224): modifies `rules[0].rules[elementNumber]`
  - Rendering (line ~338): maps over `rules[0].rules` only
  - Validation: checks only `rules[0].rules`

The UI assumes a single top level ALL wrapper rule. The API and backend have no constraint. The serializer uses `many=True` and the DB model supports multiple top level rules via FK.

  ## Impact

  - Users creating segments via API with multiple top-level rules see incomplete data in the UI
  - Editing such a segment in the dashboard silently drops hidden rules (data loss)
  - Use case: customers migrating from other vendors who want one segment with per-environment targeting rules

  ## Suggested Fix

  Either:
  1. Frontend: Update the UI to iterate over all top-level rules instead of only `rules[0]`
  2. Backend: Add API validation to reject/warn when multiple top-level rules are submitted
  3. Both: Render them properly and decide if this is an intended feature

  At minimum, the UI should warn when hidden rules exist to prevent silent data loss on edit.

Zaimwa9:

## Changes
Closes #7254 

- Removed the hardcoded `*rules[0]` to properly deal with multiple top level rules

⚠️ Backend modification were overkill and might have required migrating all existing rules with multiple top level rules

1. Create a segment with multiple top-level rules
```
  curl -s -X POST http://localhost:8000/api/v1/projects/1/segments/ \
    -H "Authorization: Token <YOUR_API_TOKEN>" \
    -d '{
    "name": "Multi-rule test segment",
    "description": "Testing multiple top-level rules render in the UI",
    "project": 1,
    "rules": [
      {
        "type": "ALL",
        "conditions": [],
        "rules": [
          {
            "type": "ANY",
            "conditions": [
              { "operator": "EQUAL", "property": "country", "value": "GB" },
              { "operator": "EQUAL", "property": "country", "value": "FR" }
            ],
            "rules": []
          }
        ]
      },
      {
        "type": "ALL",
        "conditions": [],
        "rules": [
          {
            "type": "NONE",
            "conditions": [
              { "operator": "EQUAL", "property": "is_bot", "value": "true" }
            ],
            "rules": []
          },
          {
            "type": "ANY",
            "conditions": [
              { "operator": "GREATER_THAN", "property": "age", "value": "18" }
            ],
            "rules": []
          }
        ]
      }
    ]
  }'
```
2. Verify in the UI

Open the segment in the dashboard. You should see all 3 rule cards:
- "Any of the following" — country = GB, country = FR
- "And none of the following" — is_bot = true
- "And any of the following" — age > 18

  3. Before the fix

Only the first rule card (country GB/FR) was visible. The other two were silently hidden

**Before fix**
image

**After fix**
image

### DIFF
--- a/frontend/web/components/modals/CreateSegment.tsx
+++ b/frontend/web/components/modals/CreateSegment.tsx
@@ -78,13 +78,11 @@ type CreateSegmentType = {
 type CreateSegmentError = {
   status: number
   data: {
-    rules: [
-      {
-        rules: Array<{
-          conditions: SegmentConditionsError[]
-        }>
-      },
-    ]
+    rules: Array<{
+      rules: Array<{
+        conditions: SegmentConditionsError[]
+      }>
+    }>
   }
 }
 
@@ -227,7 +225,7 @@ const CreateSegment: FC<CreateSegmentType> = ({
     newValue: SegmentRule,
   ) => {
     const newRules = cloneDeep(rules)
-    newRules[0].rules[elementNumber] = newValue
+    newRules[rulesIndex].rules[elementNumber] = newValue
     setRules(newRules)
   }
 
@@ -336,10 +334,11 @@ const CreateSegment: FC<CreateSegmentType> = ({
     return () => setInterceptClose(null)
   }, [onClosing])
   const isValid = useMemo(() => {
-    if (!rules[0]?.rules?.find((v) => !v.delete)) {
+    const allSubRules = rules.flatMap((r) => r.rules)
+    if (!allSubRules.find((v) => !v.delete)) {
       return false
     }
-    const res = rules[0].rules.find((v) =>
+    const res = allSubRules.find((v) =>
       v.conditions.find((c) => !Utils.validateRule(c)),
     )
     return !res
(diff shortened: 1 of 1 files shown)


talissoncosta (approved):
Tested locally with curl ✅


Thanks for handling that @Zaimwa9 !!! Good Job!

PR #7349, opened 2026-04-27
fix(Import/Export): Destructive environment-level import resets feature states in other environments

### LINKED ISSUE #7166: Destructive environment-level import resets feature states in other environments
When using the Overwrite Destructive strategy on an environment-level feature import, all environments are affected — not just the target environment. Other environments have their feature states reset to the feature's `initial_value` and `default_enabled`.

### Root cause

In `api/features/import_export/tasks.py`, `_import_features_for_environment` handles OVERWRITE_DESTRUCTIVE by deleting the entire Feature and recreating it:

1. `existing_feature.delete()` — cascade-deletes all FeatureStates across all environments
2. `Feature.objects.create(initial_value=..., default_enabled=...)` — Django auto-creates default FeatureStates for every environment using the exported defaults
3. Only the target environment's FeatureState is then updated with the exported `value` and `enabled`

Non-target environments are left with the auto-created defaults.

### Expected behaviour

Overwrite Destructive should only affect the target environment's feature state. Other environments should be left unchanged.

### Suggested fix

Instead of deleting and recreating the Feature, update it in-place: update the target environment's FeatureState (and multivariate options) without touching the Feature record or other environments' states.

khvn26:

## Changes

Closes #7166

The `OVERWRITE_DESTRUCTIVE` strategy in `_import_features_for_environment` deleted the entire `Feature` and recreated it. The cascade soft-deleted every `FeatureState` across every environment, and `Feature.objects.create(...)` then re-seeded env-default `FeatureState`s for all environments using the exporter's `initial_value` / `default_enabled`. Only the target environment had the imported value/enabled written back, so every other environment was reset to the exporter's defaults.

Replaced the delete + recreate with a new `overwrite_feature_for_environment` service that:

- Updates Feature-level fields (`initial_value`, `is_server_key_only`, `default_enabled`) in place. These only seed *new* environments and don't retroactively change existing FeatureStates.
- Deletes only the target environment's `FeatureSegment`s and identity-override `FeatureState`s, preserving the destructive semantic *within* the target env. Other environments' overrides are untouched.
- Reconciles `MultivariateFeatureOption`s on the feature by `(type, value)`: reuses matches, creates missing, leaves extras alone (they may belong to other envs' allocations).
- Upserts the target env's `MultivariateFeatureStateValue` percentage allocations and zeroes out any options the import dropped, scoped to the target env's MV state values only.
- Looks up the target env-default `FeatureState` via `FeatureState.objects.get_live_feature_states(...)` so the live version's FS is targeted on v2-versioned environments. Previously a `.filter(...).first()` returned the lowest-id (initial-version) FS once subsequent versions had been published, silently no-opping the import.

Also moved the existing create-strategy logic to `services.py` for symmetry, and added a `FeatureExportData` TypedDict in `types.py` to drop the previous untyped `dict[str, Optional[Union[bool, str, int]]]` signatures and a string of `# type: ignore` comments.

The v1 and v2 paths are identical: the destructive strategy mutates the live `FeatureState` in place rather than publishing a new `EnvironmentFeatureVersion`. Opting into `OVERWRITE_DESTRUCTIVE` is opting out of audit trail; spending code to gently publish a new version subverts the strategy's stated intent.

Two end-to-end tests covering the full export → import pipeline:

- The existing destructive-strategy E2E test now also exercises bystander-env preservation and target-env segment + identity override destruction.
- A new v2-versioning E2E test sets up a v2-versioned target env with multiple published versions and verifies that a destructive import mutates the live version's FS in place (no extra version is created), the prior version's FS is untouched, and live state reflects the imported value / enabled.

### DIFF
--- a/api/features/import_export/services.py
+++ b/api/features/import_export/services.py
@@ -1,5 +1,3 @@
-from django.db.models import Q
-
 from environments.models import Environment
 from features.import_export.types import FeatureExportData
 from features.models import Feature, FeatureSegment, FeatureState
@@ -25,33 +23,30 @@ def overwrite_feature_for_environment(
     existing_feature.default_enabled = feature_data["default_enabled"]
     existing_feature.save()
 
-    FeatureSegment.objects.filter(
-        feature=existing_feature, environment=environment
-    ).delete()
+    # Identity overrides aren't versioned and live on the FeatureState directly.
     existing_feature.feature_states.filter(
         environment=environment, identity__isnull=False
     ).delete()
 
-    feature_state = FeatureState.objects.get_live_feature_states(
-        environment=environment,
-        additional_filters=Q(
-            feature=existing_feature,
+    new_version: EnvironmentFeatureVersion | None = None
+    if environment.use_v2_feature_versioning:
+        draft_version = EnvironmentFeatureVersion.objects.create(
+            environment=environment, feature=existing_feature
+        )
+        draft_version.feature_states.filter(feature_segment__isnull=False).delete()
+        feature_state = draft_version.feature_states.get(
+            identity__isnull=True, feature_segment__isnull=True
+        )
+        new_version = draft_version
+    else:
+        FeatureSegment.objects.filter(
+            feature=existing_feature, environment=environment
+        ).delete()
+        feature_state = existing_feature.feature_states.get(
+            environment=environment,
             identity__isnull=True,
             feature_segment__isnull=True,
-        ),
-    ).first()
-    if feature_state is None:
-        fs_kwargs: dict[str, object] = {
-            "feature": existing_feature,
-            "environment": environment,
-        }
-        if environment.use_v2_feature_versioning:
-            fs_kwargs["environment_feature_version"] = (
-                EnvironmentFeatureVersion.create_initial_version(
-                    environment=environment, feature=existing_feature
-                )
-            )
-        feature_state = FeatureState.objects.create(**fs_kwargs)
+        )
 
     existing_options_by_value: dict[
         tuple[str | None, str | int | bool | None], MultivariateFeatureOption
@@ -106,6 +101,9 @@ def overwrite_feature_for_environment(
     feature_state_value.save()
     feature_state.enabled = feature_data["enabled"]
     feature_state.save()
+
+    if new_version is not None and not new_version.published:
+        new_version.publish()
 
 
 def create_feature_for_environment(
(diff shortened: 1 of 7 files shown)


api/features/import_export/services.py
```diff
@@ -0,0 +1,129 @@
...
+    environment: Environment,
+) -> None:
+    """
+    Apply a destructive feature import to a single environment without
+    affecting other environments' feature states.
+    """
+    existing_feature.initial_value = feature_data["initial_value"]
+    existing_feature.is_server_key_only = feature_data["is_server_key_only"]
+    existing_feature.default_enabled = feature_data["default_enabled"]
+    existing_feature.save()
```
emyller: question: I understand matching comes from elsewhere, but it feels a bit odd that this function doesn't set other basic attributes such as `name`. As a public service called `overwrite_feature_for_environment`, should it?

api/features/import_export/services.py
emyller, line 1: suggestion (non-blocking): It looks like a good portion of both `overwrite_feature_for_environment` and `create_feature_for_environment` could be extracted into a new `_apply_state_payload(feature, feature_data)` function.

emyller (approved):
LGTM! Couple optional comments.

api/features/import_export/services.py
khvn26: `Feature.name` is project-level, and the feature was matched by name before overwriting.

emyller (changes requested):
Sorry, need to update this [UI note](https://github.com/Flagsmith/flagsmith/blob/e27a0ac1fc28e515bcd286c2bfa861e0b6f0fd1d/frontend/web/components/import-export/FeatureImport.tsx#L323) and this [doc note](https://github.com/Flagsmith/flagsmith/blob/e27a0ac1fc28e515bcd286c2bfa861e0b6f0fd1d/docs/docs/administration-and-security/data-management/import-from-launchdarkly.md?plain=1#L31). Makes me wonder whether this PR warrants ditching the "destructive" term entirely.

api/features/import_export/services.py
khvn26: Improved in f85ac67ca30a916d7b6b15814a5c90021d708ad5.

khvn26:
@emyller see df5a068a10601508f65bab4d6cbb2453c8e0024c. IMO we're still destructive at the environment level, even though the gravity of the naming has diminished.

emyller (approved):
Thanks for addressing my review neatly.

PR #7378, opened 2026-04-30
feat: removed-gram-implementation

Zaimwa9:

## Changes
- Cleaned up Gram Elements implementation


talissoncosta (commented):
Looks good. Just some questions before approving it.

Just checking have you ran `nom install` after removing the packages from `package.json` ? 

Also, can you clarify why we should have two identical `gram.json` on the project?

Zaimwa9:
Good call on both. Gram.json was a miss after removing it by mistake

talissoncosta (approved):

PR #7382, opened 2026-04-30
feat: added scale up extra seat addon handling

Zaimwa9:

## Changes
Closes [#121](https://github.com/Flagsmith/flagsmith-private/issues/121)

- Support `Scale-Up-v4` Chargebee addon `Additional-Team-Members-Scale-Up-v4` for auto-seat-upgrade
- Refactor: Extract `normalise_plan_id` into `SubscriptionPlanFamily` to re-use plan_id normalisation
- Move seat addon constants to `chargebee/constants.py` and add structured logging on successful seat addition

- Added tests
- Will need testing E2E on prod
- Tested locally: extra seat worked, although locally I don't receive the webhook from chargebee hence why the total seats wasn't updated
image

### DIFF
--- a/api/organisations/chargebee/chargebee.py
+++ b/api/organisations/chargebee/chargebee.py
@@ -1,10 +1,9 @@
 import logging
 from contextlib import suppress
-
-import structlog
 from datetime import datetime
 from typing import Any
 
+import structlog
 from chargebee.api_error import (  # type: ignore[import-untyped]
     APIError as ChargebeeAPIError,
 )
--- a/api/organisations/chargebee/chargebee.py
+++ b/api/organisations/chargebee/chargebee.py
@@ -1,5 +1,7 @@
 import logging
 from contextlib import suppress
+
+import structlog
 from datetime import datetime
 from typing import Any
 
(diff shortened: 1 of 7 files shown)


api/organisations/chargebee/chargebee.py
```diff
@@ -233,6 +237,14 @@ def add_single_seat(subscription_id: str) -> None:
             ),
         )
 
+        audit_log.info(
+            "seat.added",
+            subscription__id=subscription_id,
```
khvn26: Can we add organisation__id here?

api/organisations/chargebee/chargebee.py
```diff
@@ -43,6 +46,7 @@
 )
 
 logger = logging.getLogger(__name__)
+audit_log = structlog.get_logger("billing")
```
khvn26: nit: Let's not confuse with actual audit log — `log` is fine here.

khvn26 (dismissed):
Approved with 2 non-blocking comments.

api/organisations/chargebee/chargebee.py
Zaimwa9: Ah thank you!!! I hesitated as it needs to update the callsite to pass it but i'm up for it!!

khvn26: Do we really need it? I think we can grab organisation_id from the subscription cache object?

api/organisations/chargebee/chargebee.py
Zaimwa9: [f0526a8](https://github.com/Flagsmith/flagsmith/pull/7382/commits/f0526a83484b8fcb2acd5c62b840eac39b93a797)

api/organisations/chargebee/chargebee.py
Zaimwa9: I updated the caller, it's already available and avoids an extra DB call (even if it's not a frequent operation) and a circular import. Lmk
[f0526a8](https://github.com/Flagsmith/flagsmith/pull/7382/commits/f0526a83484b8fcb2acd5c62b840eac39b93a797)

khvn26 (approved):

PR #7100, opened 2026-04-02
feat: add Storybook stories for pure components

talissoncosta:
- [x] I have read the [Contributing Guide](/Flagsmith/flagsmith/blob/main/CONTRIBUTING.md).
- [ ] I have added information to `docs/` if required so people know about the feature.
- [x] I have filled in the "Changes" section below.
- [x] I have filled in the "How did you test this code" section below.

## Changes

Closes #7179
Contributes to #6882

Add Storybook stories for the design system. **45 components** organised across 8 categories. Two small design-system consolidations bundled in (Banner / ModalAlert removed, ColorSwatch + BooleanDotIndicator unified).

### Coverage

- **94% (16/17) of base primitives** under `web/components/base/` have stories
- Not covered: `Popover` (blocked by FocusMonitor + jQuery); 5 react-select / routing internals not meant for direct consumer use
- **45 stories total** across 8 sidebar groups

### Categories

| Category | Components |
|----------|-----------|
| **(top-level)** | Button, ButtonDropdown, Icons |
| **Data Display** | BarChart, BooleanDotIndicator, ColorSwatch, DropdownMenu, IdentifierString, LabelWithTooltip, Logo, SettingTitle, StatItem, ToggleChip, Tooltip, VCSProviderTag |
| **Forms** | Checkbox, CheckboxGroup, ChipInput, FormGroup, GhostInput, MultiSelect, Radio, SearchableSelect, Switch |
| **Feedback** | EmptyState, ErrorMessage, InfoMessage, Loader, PasswordRequirements, Skeleton, WarningMessage |
| **Patterns** | AccordionCard, Card, PageTitle, Panel, SettingRow |
| **Navigation** | Breadcrumb, OverflowNav, SidebarLink, Tabs |
| **Layout** | Column, Flex, Row |
| **Modals** | InlineModal, Modal |

### Story conventions

- **Args + `component:`** for primitives (autodocs props table, working Controls panel)
- **`render:`** for stories needing state, composition, or layout demos
- **`play` functions** on Modal, Tooltip, ButtonDropdown, DropdownMenu, OverflowNav so Chromatic captures the open / hovered / expanded state
- Component descriptions on heavyweight stories
- Shared `withRouter` decorator at `documentation/components/_decorators.tsx`

### Component changes (non-story)

Mostly Storybook plumbing. Two design-system consolidations.

| Change | Why | Risk |
|---|---|---|
[...]

PR #7402, opened 2026-05-04
fix(button): centre icons natively, fix hover and disabled states

talissoncosta:

## Changes

Follow-up to a review comment on #7100 — icon-only buttons (`theme='icon'`) rendered with the icon visibly off-centre. The cause was margin-based spacing: `iconLeft` and `iconRight` props applied `mr-2` / `ml-2` directly to the icon. On icon-only variants (32×32 with `justify-content: center`), the trailing margin pushed the icon a few pixels off-centre — flexbox re-centres the item itself, but the margin still shifts the content inside.

Rather than patching the margin, this PR drops the icon-position props and lets `.btn` handle spacing natively.

**Button API**
- Drop `iconLeft` / `iconRight` props. Icons now compose as children alongside the label.

**`.btn` layout**
- Switch from Bootstrap-default `inline-block` to `inline-flex` with `align-items: center; gap: 0.5rem` — siblings space themselves, no wrapper span or margin tricks, icon-only buttons centre by construction.
- Pin `height: $btn-line-height` so default-size buttons keep their 44px box (inline-flex would otherwise collapse to content height).
- Drop the legacy `.btn:hover` / `.btn:active` rules (and their `.dark` counterparts) that forced a purple background on every variant — they were project-picker leftovers leaking onto Danger / Tertiary / Success in both modes.

**Variant adjustments**
- `.btn-link` (`theme='text'`) is now `inline-flex` with `vertical-align: baseline` and `height: auto`, so it flows inline with surrounding text instead of rendering as a 44px box.
- `.btn-project` (project picker tiles) opts out of the new flex layout — it's a 76px tile surface, not a button-shaped affordance.

**Consumer cleanups**
- `InviteUsers` `+` icon: drop the broken `<Row>` + padding wrapper, swap IonIcon for the project's `<Icon name='plus'>` at 16px (the small-button norm).
- `ViewDocs`: drop the manual `marginTop: -5` / `lineHeight: 24px` alignment hacks now that `.btn` does centring natively.
- Hardcoded hex `fill` values on icons in the touched files → semantic tokens (or dropped where the parent's `color` already drives the icon via `currentColor`).

**Storybook**
- Add the missing `xxSmall` variant to the Sizes story.
- Trim the Variants story description so it doesn't point readers at `theme='icon'` for table actions (a future `IconButton` primitive will own that pattern).

- Storybook → `Components/Button` → walked through Variants / Sizes / Disabled / WithIcons.
- Light + dark mode hover sweep across Primary / Secondary / Outline / Danger / Success / Tertiary / Text — variant hovers all win, no purple bleed.
- Identity page → alias `Edit` button baseline-aligned with the surrounding text.
- Project picker → letter sizing consistent across all tiles.
- Invite Users modal → `+` icon spaced cleanly from the label.
- Chromatic snapshot diff shows the centring correction on the existing button stories with no regressions on other variants.

### DIFF
--- a/frontend/web/components/base/forms/Button.tsx
+++ b/frontend/web/components/base/forms/Button.tsx
@@ -1,14 +1,6 @@
 import React from 'react'
 import cn from 'classnames'
 import { ButtonHTMLAttributes, HTMLAttributeAnchorTarget } from 'react'
-import Icon, { IconName } from 'components/icons/Icon'
-
-const iconColours = {
-  primary: '#6837fc',
-  white: '#ffffff',
-} as const
-
-export type IconColour = keyof typeof iconColours
 
 export const themeClassNames = {
   danger: 'btn btn-danger',
@@ -31,15 +23,10 @@ export const sizeClassNames = {
 }
 
 export type ButtonType = ButtonHTMLAttributes<HTMLButtonElement> & {
-  iconRight?: IconName
-  iconRightColour?: IconColour
-  iconLeftColour?: IconColour
-  iconLeft?: IconName
   href?: string
   target?: HTMLAttributeAnchorTarget
   theme?: keyof typeof themeClassNames
   size?: keyof typeof sizeClassNames
-  iconSize?: number
 }
 
 export const Button = React.forwardRef<
@@ -51,11 +38,6 @@ export const Button = React.forwardRef<
       children,
       className,
       href,
-      iconLeft,
-      iconLeftColour,
-      iconRight,
-      iconRightColour,
-      iconSize = 24,
       onMouseUp,
       size = 'default',
       target,
@@ -65,6 +47,11 @@ export const Button = React.forwardRef<
     },
     ref,
   ) => {
+    const content = (
+      <span className='d-flex h-100 align-items-center justify-content-center gap-2'>
+        {children}
+      </span>
+    )
     return href ? (
       <a
         onClick={rest.onClick as React.MouseEventHandler}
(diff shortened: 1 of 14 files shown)


Zaimwa9 (dismissed):

Zaimwa9:
@talissoncosta can you rebase main please

talissoncosta:
@Zaimwa9 done, sorry for that.

Zaimwa9 (dismissed):

talissoncosta:
@Zaimwa9 just found an issue on icon buttons that are not being covered by the tests. I will get this fixed soon and let you know.

frontend/web/components/base/forms/Button.tsx
```diff
@@ -31,15 +22,10 @@ export const sizeClassNames = {
 }
 
 export type ButtonType = ButtonHTMLAttributes<HTMLButtonElement> & {
-  iconRight?: IconName
-  iconRightColour?: IconColour
-  iconLeftColour?: IconColour
-  iconLeft?: IconName
```
Zaimwa9, line 37: It looks like `FlagEnvironmentsPage` has been skipped during the migration (still having a `iconLeft='arrow-left` there)

Zaimwa9 (changes requested):
Just missing one page to migrate, otherwise the visual changes make sense

frontend/web/components/base/forms/Button.tsx
talissoncosta: Thanks for catching that @Zaimwa9 
Fixed in: c4f7f65b5feaa4404bbb5c9369eb2c3ebf046e42

Zaimwa9 (approved):
👍

PR #7437, opened 2026-05-06
fix(button): use :focus-visible for hover-bg, drop click-focused dark state

talissoncosta:

- [x] I have added information to \`docs/\` if required so people know about the feature.
- [x] I have filled in the \"Changes\" section below.
- [x] I have filled in the \"How did you test this code\" section below.

## Changes

**TL;DR.** Primary buttons stay visibly \"pressed\" after a mouse click until the user clicks elsewhere. Switching the focus-state CSS from \`:focus\` to \`:focus-visible\` removes the click-and-stuck appearance while preserving the keyboard-focus indicator.

### Cause

`_buttons.scss` lines 8–11 (pre-fix):

```scss
.btn, button.btn {
  &:hover,
  &:focus {
    background-color: \$btn-hover-bg;  // = \$primary600 (darker primary)
  }
}
```

`:focus` matches **any** focus event, including the focus a browser implicitly gives a button after a mouse click. So the flow is:

1. User clicks a primary button → browser focuses it.
2. `:focus` fires → background turns to `\$primary600` (darker).
3. Focus persists until the user clicks elsewhere or tabs away → button stays dark.

Visible on every primary button across the app. Storybook surfaces it most obviously because the button stays in view post-click, but it happens in production too — confirmed by clicking *Create Identities* on the Identities page and watching the button stay dark.

### Fix

Single rule change: `:focus` → `:focus-visible`.

```diff
   &:hover,
-  &:focus {
+  &:focus-visible {
     background-color: \$btn-hover-bg;
   }
```

`:focus-visible` is the modern pseudo-class that matches focus only when it arrived via keyboard navigation (Tab) or programmatic focus — not mouse clicks. Browser support is universal across all evergreen browsers.

After the change:
- **Mouse click:** button is still focused (browser default) but `:focus-visible` doesn't match → no darkening → no stuck appearance. ✓
- **Tab key:** `:focus-visible` matches → background darkens → keyboard users see the focus state. ✓ (a11y preserved)
- **`:focus-visible { box-shadow: none }`** rule on the next line stays as-is — same intent (suppress Bootstrap's default focus ring on keyboard focus, since the bg-darkening is the visual indicator).

### Out of scope

The `:focus` references in `.btn-project` rules (lines 224, 435) are visual focus styles for the project-picker variant. They probably want the same treatment but the `.btn-project` rendering is more involved than the base button — kept out of this PR to keep the diff focused. Worth a follow-up if Chromatic flags any visual divergence in those stories.

- Reproduced on `main` by clicking *Create Identities* on the Identities page and observing the button stay darkened until clicking elsewhere.
- After the fix: same click no longer leaves the button visibly pressed.
- Keyboard verification: \`Tab\` to a button, confirmed the darker background still appears (a11y indicator intact).
- Chromatic build on this PR will diff every primary/secondary/danger/etc. button snapshot in the design system — worth scrolling through to confirm no surface relies on the old click-focused appearance.

### DIFF
--- a/frontend/web/styles/project/_buttons.scss
+++ b/frontend/web/styles/project/_buttons.scss
@@ -5,8 +5,12 @@ button.btn {
   white-space: nowrap;
   color: white;
   // Project button, to remove with project select bar
+  // Use `:focus-visible` (keyboard / programmatic focus only) instead of
+  // `:focus` so a mouse click doesn't leave the button visibly "pressed"
+  // until the user clicks elsewhere — focus persists on click by design,
+  // we just don't want it to look like a held-down state.
   &:hover,
-  &:focus {
+  &:focus-visible {
     background-color: $btn-hover-bg;
   }
   &:focus-visible {
--- a/frontend/web/styles/project/_buttons.scss
+++ b/frontend/web/styles/project/_buttons.scss
@@ -24,7 +24,7 @@ button.btn {
   &-link {
     color: $primary;
     &:hover,
-    &:focus {
+    &:focus-visible {
       background: transparent;
     }
     &.btn:active {
@@ -36,7 +36,7 @@ button.btn {
   &-danger {
     color: white;
     &:hover,
-    &:focus {
+    &:focus-visible {
       color: white;
       background-color: $btn-danger-hover;
     }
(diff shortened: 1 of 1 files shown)


talissoncosta:
Closing in favour of a documented design plan for Button + Link before further code changes. The implementation here was outpacing the architectural decisions — pausing the code, will reopen replacements once the plan is agreed.

PR #7445, opened 2026-05-06
fix: dedupe integration list in onboarding select

kyle-ssg:
integration_data flag and Constants.integrationSummaries both list some tools (e.g. GitLab), causing duplicates on the "Use Feature Flags with your favourite tools" picker. Dedupe by title.

## Changes

Fixes duplicate GitLab integration showing, due to us having one!

image

Viewed onboarding

### DIFF
--- a/frontend/common/constants.ts
+++ b/frontend/common/constants.ts
@@ -508,11 +508,6 @@ const Constants = {
     },
     {
       categories: ['CI/CD'],
-      image: '/static/images/integrations/gitlab.svg',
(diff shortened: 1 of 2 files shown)


matthewelwell (dismissed):
@kyle-ssg seems like this is a good future proofing solution, but should we also be removing the current duplicate Gitlab entry in the list?

kyle-ssg:
Ah yep of course, done!

matthewelwell (approved):

PR #7448, opened 2026-05-07
fix: removed api logout call

Zaimwa9:

## Changes
As logging out from the API invalidate the user token, we reverted to not calling the API when logging out from dashboard


matthewelwell (approved):
For additional context, this is being done because the token is presented to users in their account settings as an 'API key'. Until we build properly scoped user API keys, or just force everyone to use org-level API keys, we cannot invalidate these tokens on logout.

PR #7449, opened 2026-05-07
fix(versioning): scope scheduled-FS query by feature in v2 enable

### LINKED ISSUE #7450: v2 versioning enable: scheduled feature states bound to wrong feature
## Bug

`_create_initial_feature_versions` (`api/features/versioning/tasks.py:89-134`) creates one `EnvironmentFeatureVersion` per `Feature` in the project, and inside that per-feature loop runs a second query for scheduled, committed feature states to convert them into per-FS versions. The query filters by `environment` (added in #4872) but **not by feature**, so within each iteration it returns scheduled FSes belonging to other features and binds them to an EFV with the wrong `feature_id`.

End state: M scheduled FSes are bound to EFVs whose `feature_id` is the **first** feature in iteration order, in the **first** outer iteration only — subsequent iterations' fresh queries (which still require `change_request__isnull=False`) return empty because iteration 1's `bulk_update` already blanked every scheduled FS's `change_request_id`.

## Reproduction

Environment with three features `[A, B, C]`. Commit a scheduled change request for feature B and another for feature C. Call `enable_v2_versioning(environment.id)`. After the task:

- Both scheduled FSes are bound to `EnvironmentFeatureVersion` rows with `feature_id = A.id`.
- The scheduled change for feature B never goes live at its `live_from` (the engine's "latest version per (env, feature)" lookup never returns it for feature B).
- Feature A's history shows a "version" the customer didn't author.
- The mis-attribution propagates via `add_existing_feature_states` on subsequent edits — every time the customer edits feature A, the new EFV clones the wrong-feature FS forward.

## Impact

- Customer's scheduled flag change for feature B/C silently does not take effect at the scheduled time; SDKs continue serving the pre-scheduled value.
- Audit log `EF_VERSION` entries point at the wrong feature.
- `FeatureState.change_request_id` is blanked on the FS itself; the FK is moved to a wrong-feature EFV, breaking integrations that join `feature_state.change_request`.

## Why this slipped through

- PR #4872 added `environment=environment` to close a cross-environment leak in the same query — it did not add `feature=feature`.
- Migration `0005_fix_scheduled_fs_data_issue_caused_by_enabling_versioning` repairs only environments where `use_v2_feature_versioning=False`, so it does not run on environments already on v2.
- The existing test `test_enable_v2_versioning__scheduled_changes_exist__converts_published_scheduled_changes` exercises only one feature with one scheduled CR — never two scheduled CRs on different features within the same environment, which is the exact shape that triggers the bug.

gagantrivedi:
Fixes #7450.

## Summary

- `_create_initial_feature_versions` filtered scheduled FSes by environment (closed in #4872) but not by feature, so the per-feature loop returned scheduled FSes belonging to other features and bound them to an `EnvironmentFeatureVersion` with the wrong `feature_id`. The mis-binding propagated silently through `add_existing_feature_states` on subsequent edits.
- One-line fix: add `feature=feature` to the query at `api/features/versioning/tasks.py:113`.
- New regression test (`test_enable_v2_versioning__multi_feature_scheduled_changes__each_efv_matches_feature`) covering the multi-feature case. It fails on `main` and passes with the fix.

## Production impact

Audit query run on production: 3 affected rows total, all on Flagsmith's internal testing project. **No customer data affected.** The 3 internal rows can be cleaned up manually via Django shell; no data migration required.

## Why migration `0005` did not catch this

`feature_versioning/0005_fix_scheduled_fs_data_issue_caused_by_enabling_versioning` repaired only environments where `use_v2_feature_versioning=False` — i.e. the cross-environment leak that #4872 closed. It does not run against environments already on v2, so the cross-feature variant of the same bug class went unrepaired.

## Test plan

- [x] New regression test fails on `main`, passes with the fix
- [x] Existing scheduled-changes test (`test_enable_v2_versioning__scheduled_changes_exist__converts_published_scheduled_changes`) still passes — the `environment` filter from #4872 is preserved
- [x] Existing happy-path enable + disable tests still pass
- [x] Production audit query: `SELECT count(*) FROM features_featurestate fs JOIN feature_versioning_environmentfeatureversion v ON fs.environment_feature_version_id = v.uuid WHERE fs.feature_id <> v.feature_id;` — 3 rows, internal only

### DIFF
--- a/api/features/versioning/tasks.py
+++ b/api/features/versioning/tasks.py
@@ -110,6 +110,7 @@ def _create_initial_feature_versions(environment: "Environment"):  # type: ignor
         related_feature_segments.update(environment_feature_version=ef_version)
 
         scheduled_feature_states = FeatureState.objects.filter(
+            feature=feature,
             live_from__gt=now,
             change_request__isnull=False,
             change_request__committed_at__isnull=False,
--- a/api/tests/unit/features/versioning/test_unit_versioning_tasks.py
+++ b/api/tests/unit/features/versioning/test_unit_versioning_tasks.py
@@ -396,6 +396,79 @@ def test_enable_v2_versioning__scheduled_changes_exist__converts_published_sched
     )
 
 
+def test_enable_v2_versioning__multi_feature_scheduled_changes__each_efv_matches_feature(
+    environment: Environment,
+    project: Project,
+    feature: Feature,
+    staff_user: FFAdminUser,
+) -> None:
+    # Regression for `_create_initial_feature_versions`' scheduled-FS query
+    # at `api/features/versioning/tasks.py:112-118`: it is missing a
+    # `feature=feature` filter, so scheduled FSes get bound to an EFV of the
+    # wrong feature. The fixture's `feature` anchors the first iteration so
+    # the bug binds the scheduled FSes to its EFV instead of feature_b /
+    # feature_c.
+
+    # Given
+    now = timezone.now()
+    one_hour_from_now = now + timedelta(hours=1)
+    two_hours_from_now = now + timedelta(hours=2)
+
+    feature_b = Feature.objects.create(name="feature_b", project=project)
+    feature_c = Feature.objects.create(name="feature_c", project=project)
+
+    cr_b = ChangeRequest.objects.create(
+        environment=environment,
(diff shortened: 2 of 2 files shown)


Zaimwa9 (approved):
Sneaky. Thanks

Issue #5663, opened 2025-06-17
Spike: Segment membership inspection


agregoryfs:
Today the segment ↔ identity relationship only navigates in one direction: from an identity we can list the segments it belongs to, but from a segment there is no way to inspect the set of identities that match its rules — only to search for one identifier at a time in the segment editor.

This presents an architecture design problem. Flagsmith segments are rule-based. Membership is computed by the flag engine on every read and is not currently persisted. Any "list members" capability has to work over both identity stores (core and edge) and account for rule shapes that don't reduce to simple equality (percentage splits, regex, modulo, "is not set", etc.).

Scope of this spike: decide what v1 of segment-membership inspection looks like — UX, API shape, evaluation path — and surface the architectural questions that fall out of that decision.

Definition of Done:

- [x] PR with user-facing documentation describing v1 behaviour for identity search and segment membership inspection as we intend to ship it.
- [ ] Epic outlining the v1 work, broken down by surface (API, edge, frontend, RBAC, docs).
- [ ] Follow-up issues / spikes filed for anything explicitly out of scope for v1.
- [x] RFC for any architectural changes the feature requires.
- [ ] Adjacent issues triaged: closed-as-duplicate, linked, or folded into the epic.

khvn26:
Closing out the spike. Two epics created:

- V1: #7405 — Segment membership inspection V1 (12 sub-issues across `flagsmith`, `pulumi`, `pulumi-monitoring`).
- Backlog: #7416 — Segment membership inspection backlog (7 deferred items across `flagsmith`, `flagsmith-engine`, `edge-api`).

PRs already in flight:

- Index core PoC (draft): #7394 — referenced from V1 issue #7406.
- User-facing docs: #7399 — separate from the V1 epic per request.

RFC: https://www.notion.so/355a2b634bf9815289b3d070e164f9eb (private to the Flagsmith team).

## Notes (V1)

The RFC is the source of truth for design decisions. Notable choices baked into V1:

- PK-as-ordinal (no `IdentityOrdinal` table). Saves ~270 GB of ordinal mapping at SaaS scale.
- Single bitmap per atom; sharding deferred to backlog.
- No CDC for edge. Freshness surfaced via `as_of` and manual refresh. CDC moved to backlog.
- No atom GC — orphans tolerated. GC moved to backlog.

Tech debt knowingly taken on:

[...]

PR #7464, opened 2026-05-08
feat(Segment membership inspection PoC): Daily ClickHouse-backed per-env segment counts

khvn26:

## Changes

Contributes to #5663.

Adds a daily pipeline that backfills Dynamo identities into ClickHouse, materialises per-(segment, environment) match counts via `flagsmith-sql-flag-engine` (`ClickHouseDialect`), and exposes them on the segment endpoint as `memberships: [{environment, count, last_synced_at}]` for env-dropdown badges. Gated behind the org-scoped `segment_membership_inspection` FoF flag; no-ops when `CLICKHOUSE_HOST` is unset.

History note: the PoC originally landed against Snowflake. After the 12/05/2026 RFC review, the cost analysis (and a [stacked engine PR](https://github.com/Flagsmith/flagsmith-sql-flag-engine/pull/8) adding the ClickHouse dialect) tipped us toward ClickHouse for V1.

Storage decisions:
- The IDENTITIES table is `ReplacingMergeTree(inserted_at) ORDER BY (environment_id, id)`. Daily INSERTs dedupe at merge time (most-recent `inserted_at` wins); refresh queries use `FROM IDENTITIES AS i FINAL` to force read-time dedup so counts always reflect the latest backfill.
- Traits live in a `JSON` column (CH 24+; GA in 25.x). Each key becomes a typed columnar subcolumn — same cost shape as Snowflake's VARIANT.
- The engine's published `schema_ddl` stays `MergeTree` as the simplest correct shape; this PoC overrides for its mutation-heavy workload. Translator output is engine-agnostic so the override is invisible to the SQL flag engine.

Spend attribution: `clickhouse-connect` opens a session with a `log_comment` setting per task (`flagsmith:segment_membership:{backfill,refresh}:org_X:project_Y`), landing in `system.query_log` for per-(org, project) rollups — direct analogue of Snowflake's `QUERY_TAG`.

Review complexity: 4/5

Review order recommendation:
1. `models.py` (cache table)
2. `services.py` (compile + count, parameterised SQL with `FINAL`)
3. `tasks.py` (daily recurring backfill fans out per-project refresh)
4. `mappers.py` (Dynamo doc → IDENTITIES row)
5. `migrations/0002_*` (ClickHouse `ReplacingMergeTree` DDL via `RunPython`, no-op when unconfigured)
6. `segments/serializers.py` + `views.py` (read-side `memberships` field, prefetched).

Beyond the existing unit + integration tests:

Ran an end-to-end smoke test against `moto`-mocked Dynamo + a real ClickHouse Cloud trial (eu-west-2) configured via env vars. The flow:

1. Bootstrapped Django fixtures (Org, Project, Env, Segment with a `plan IN (growth, enterprise)` rule).
7. Applied `make docker-up django-migrate` — migrations created the `SegmentMembership` table in core Postgres and the `IDENTITIES` schema in ClickHouse Cloud.
8. Stood up an in-process Dynamo identities table via `moto.mock_dynamodb` and seeded 50 synthetic identities — 25 with `traits.plan = "growth"`, 25 with `traits.plan = "free"`.
9. Invoked `backfill_identities_to_clickhouse()` directly — confirmed all 50 rows landed in ClickHouse's `IDENTITIES` table with the per-(org, project) `log_comment` attribution attached.
10. Invoked `refresh_project_segment_counts(project_id)` (inline-dispatched from the backfill's fan-out) — confirmed exactly one `SegmentMembership` row materialised with `count = 25` and a fresh `last_synced_at`.

Extensive testing will be done on staging.


.github/workflows/api-pull-request.yml
```diff
@@ -47,6 +48,10 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      # TODO: remove once flagsmith-sql-flag-engine is public
```
emyller: Are we going to make it public? Please point to an issue that will arrange for it.

api/segment_membership/models.py
```diff
@@ -0,0 +1,29 @@
+from django.db import models
+
+from environments.models import Environment
+from segments.models import Segment
+
+
+class SegmentMembership(models.Model):
```
emyller:
```suggestion
class SegmentMembershipCount(models.Model):
```

IMO favours clarity — I was looking for an `identity` key to wrap said _membership_, until I noticed `count`.

api/segment_membership/migrations/0002_setup_clickhouse_identities_schema.py
emyller, line 1: question: I wonder if it would be wise to consider [django-clickhouse](https://github.com/carrotquest/django-clickhouse/blob/master/docs/models.md#clickhousemodel). Unfortunately it doesn't look very well maintained. Would we benefit from any abstraction? (having in mind that we're adding more CH/DW DDL soon?)

api/segment_membership/migrations/0002_setup_clickhouse_identities_schema.py
```diff
@@ -0,0 +1,57 @@
...
+ENGINE = ReplacingMergeTree(inserted_at)
+ORDER BY (environment_id, id)
+"""
+
+
+def setup_clickhouse_identities_schema(
+    apps: StateApps, schema_editor: BaseDatabaseSchemaEditor
+) -> None:
+    if not settings.CLICKHOUSE_ENABLED:
+        return
```
emyller: What's the path for _updating now; enabling CH later_? This implies we'll need to get `CLICKHOUSE_ENABLED` evaluating as expected **before** this patch goes out, if I read this correctly.

api/segment_membership/mappers.py
```diff
@@ -0,0 +1,51 @@
...
+        _identity_id(identity_uuid),
+        identifier,
+        composite_key,
+        traits,
+    )
+
+
+def _identity_id(identity_uuid: str) -> int:
+    # UInt64 column refuses negatives.
+    return int.from_bytes(uuid.UUID(identity_uuid).bytes[:8], "big", signed=False)
```
emyller: This looks bound to create collisions _eventually_ in SaaS. How close are we to that? Is using the full bits (e.g. native UUID field, or 128-bit int if that exists) out of scope?

api/segment_membership/services.py
```diff
@@ -0,0 +1,139 @@
...
+        client.close()
+
+
+def get_projects_to_process() -> Iterator[Project]:
+    """Yield projects with at least one canonical segment whose org has
+    the segment-membership flag on."""
+    project_ids = Segment.live_objects.values_list("project_id", flat=True).distinct()
+    for project in Project.objects.filter(id__in=project_ids).select_related(
+        "organisation"
+    ):
```
emyller:
```suggestion
    projects_with_live_segments = Project.objects.filter(
        segments__in=Segment.live_objects.all()
    ).select_related("organisation")
    for project in projects_with_live_segments:
```

No need for `DISTINCT`. Also would the amount of items in SaaS warrant a careful big queryset handling here?

api/segment_membership/tasks.py
```diff
@@ -0,0 +1,174 @@
...
+                flagsmith_segment_membership_backfill_identities_total.inc(row_count)
+                logger.info(
+                    "backfill.environment.completed",
+                    project__id=project.id,
+                    environment__id=env.id,
+                    rows__count=row_count,
+                )
+
+    for project_id in refreshable_project_ids:
+        refresh_project_segment_counts.delay(args=(project_id,))
```
emyller: Does this look like a big spike and possible queue depth alerts every 4 hours to you?

api/segment_membership/tasks.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+@register_task_handler(
+    # 30m bounds runaway queries; legitimate runs are seconds to minutes.
```
emyller: The timeout is sane but the justification raises one question: are _runaway_ queries a problem we should be concerned with?

api/segments/serializers.py
```diff
@@ -82,6 +90,7 @@ class Meta:
 class SegmentSerializer(MetadataSerializerMixin, WritableNestedModelSerializer):
     rules = SegmentRuleSerializer(many=True, required=True, allow_empty=False)
     metadata = MetadataSerializer(required=False, many=True)
+    memberships = SegmentMembershipSerializer(many=True, read_only=True)
```
emyller: [Also](https://github.com/Flagsmith/flagsmith/pull/7464/changes#r3261736867) in the serialiser, I think `memberships` tells another story.

```suggestion
    membership_counts = SegmentMembershipCountSerializer(many=True, read_only=True)
```

api/segments/views.py
```diff
@@ -108,6 +108,7 @@ def get_queryset(self):  # type: ignore[no-untyped-def]
             # TODO: at the moment, the UI only shows the name and description of the segment in the list view.
             #  we shouldn't return all of the rules and conditions in the list view.
             queryset = queryset.prefetch_related(
+                "memberships",
```
emyller: [Related](https://github.com/Flagsmith/flagsmith/pull/7464/changes#r3261736867).

api/tests/integration/segments/test_segment_membership_field.py
```diff
@@ -0,0 +1,63 @@
...
+    project: int,
+    segment: int,
+) -> None:
+    # Given a segment with no materialised SegmentMembership rows
+    # When the segment is fetched
+    response = admin_client.get(
+        reverse(
+            "api-v1:projects:project-segments-detail",
+            args=[project, segment],
+        )
```
emyller: Reminder to standardise towards not `reverse`ing in tests.

api/tests/integration/segments/test_segment_membership_field.py
```diff
@@ -0,0 +1,63 @@
...
+from segment_membership.models import SegmentMembership
+
+
+def test_get_segment__no_memberships__returns_empty_list(
+    admin_client: APIClient,
+    project: int,
+    segment: int,
+) -> None:
+    # Given a segment with no materialised SegmentMembership rows
+    # When the segment is fetched
```
emyller: nit: Am I correct to assume this is LLM over-explaining in tests GWT comments? To me, it shows an annoying tendency towards this specific behaviour. This, and other tests in this patch, would look better cleaned up.

api/tests/integration/segments/test_segment_membership_field.py
```diff
@@ -0,0 +1,63 @@
...
+    )
+
+    # Then the memberships field carries one entry keyed by environment id
+    assert response.status_code == status.HTTP_200_OK
+    body: dict[str, Any] = response.json()
+    assert body["memberships"] == [
+        {
+            "environment": environment,
+            "count": 42,
+            "last_synced_at": synced_at.isoformat().replace("+00:00", "Z"),
```
emyller: nit: Would you mind hardcoding the timestamp? I think it would look clearer, and _dumber_ — which is preferred in tests scope.

api/tests/unit/segment_membership/test_unit_segment_membership_mappers.py
```diff
@@ -0,0 +1,152 @@
...
+        "created_date": "2026-05-08T00:00:00Z",
+        "identity_traits": [],
+    }
+
+    # When mapped
+    a = map_identity_document_to_clickhouse_row("env-key", doc_a)
+    b = map_identity_document_to_clickhouse_row("env-key", doc_b)
+
+    # Then the id projections are distinct
+    assert a[1] != b[1]
```
emyller: note: I'm still unsure whether we should worry about collisions in the real world.

api/tests/unit/segment_membership/test_unit_segment_membership_services.py
```diff
@@ -0,0 +1,332 @@
...
+    assert membership.segment_id == segment.id
+    assert membership.environment_id == environment.id
+    assert membership.count == 7
+    assert membership.last_synced_at is None
+    client.query.assert_called_once()
+    sql = client.query.call_args.args[0]
+    assert f"SELECT {segment.id} AS segment_id" in sql
+    # FINAL forces ReplacingMergeTree dedup at read time.
+    assert "FROM IDENTITIES AS i FINAL" in sql
+    assert "GROUP BY i.environment_id" in sql
```
emyller: question: Is this dialect-bound?

emyller (changes requested):
Looks thorough, thanks for making reviewing comfortable. I have a few performance-related questions, and some design concerns to cover.

.github/workflows/api-pull-request.yml
khvn26: All 14 related TODOs now reference https://github.com/Flagsmith/flagsmith-sql-flag-engine/issues/6. 2f63f6d2e.

api/segment_membership/models.py
khvn26: Renamed throughout: model, `related_name`, constraint, serializer, API field. 0e3b5ca65.

api/segment_membership/migrations/0002_setup_clickhouse_identities_schema.py
khvn26: answer: Skipped `django-clickhouse`, adopted `django-clickhouse-backend` instead in 40555d3af: it gives us a Django DB alias, native bulk inserts via clickhouse-driver, and a normal Django migration runner for DDL. I don't think we need to use the ORM.

api/segment_membership/migrations/0002_setup_clickhouse_identities_schema.py
khvn26: Restructured in 40555d3af. We're now using `django-clickhouse-backend` which gives us a native Django connection alias and tracks migrations properly. ClickHouse migrations are ran based on environment setting present, similarly to how we handle the `analytics` database.

api/segment_membership/mappers.py
khvn26: Dropped the synthetic `id` completely and re-keyed IDENTITIES on `(environment_id, identifier)`. Zero collisions, MD5 gone. 7ff1bb386.

api/segment_membership/services.py
khvn26: 88057f242: dropped the inner `.distinct()` (subquery dups don't matter inside `IN`). Sticking with `id__in=Segment.live_objects.values_list("project_id"...)` rather than your `segments__in=` form — that one emits an `INNER JOIN segments` and returns one row per matching segment (test confirmed). Added `.iterator()` for the queryset-size concern so we stream instead of materialising.

api/segment_membership/tasks.py
khvn26: Quick note: the task runs daily; 4h is the worst-case timeout, not the cadence. Agreed with the spike suspicion, however. 8c07763e4 moves the dispatch inside the per-project loop so enqueues track the backfill cadence and spread across the window instead of bursting at the end.

api/segment_membership/tasks.py
khvn26: "Runaway queries" overstated it. The 30m is really a generic backstop so the task gets reclaimed if the ClickHouse client stalls. Reworded in 4d6c9eb04.

api/segments/serializers.py
khvn26: Done in 0e3b5ca65 — field is `membership_counts`, serializer is `SegmentMembershipCountSerializer`.

api/segments/views.py
khvn26: Picked up by 0e3b5ca65 — `prefetch_related` now uses `"membership_counts"`.

api/tests/integration/segments/test_segment_membership_field.py
khvn26: Done in d3dd98800 — both call sites use inline `f"/api/v1/projects/{project}/segments/{segment}/"`.

api/tests/integration/segments/test_segment_membership_field.py
khvn26: Trimmed across the patch in aa92c737a. Kept terse GWT markers and only the comments that explain why.

api/tests/integration/segments/test_segment_membership_field.py
khvn26: Done in f60e5162b — literal timestamp both sides, datetime imports gone.

api/tests/unit/segment_membership/test_unit_segment_membership_mappers.py
khvn26: Addressed by 7ff1bb386 — column is gone, so `_identity_id` and its stability tests went with it.

api/tests/unit/segment_membership/test_unit_segment_membership_services.py
khvn26: answer: Yes, intentionally. Only the per-segment predicate is dialect-translated; the wrapping SQL is hand-rolled ClickHouse. ClickHouse is the only dialect we support today. If another lands, this test moves with the service's SQL stitching.

api/segment_membership/tasks.py
emyller: The patch is an improvement, but I'm not sure it understands my concern of a spike. Rephrasing:

1. Does this represent worker unavailability during the spike of enqueued tasks?
2. Will we receive infrastructure alerts of worker queue depth?

khvn26: I don't expect either 1 or 2 as even the largest environments are queried in well under an minute.

api/segment_membership/tasks.py
emyller: Sorry, I can agree with a [generic] timeout error as a last resort, but I can't agree with a _generic backstop_ because it's founded on a _generic assessment_: "legitimate runs are seconds to minutes".

IMO if a _legitimate run_ should last no more than 5 minutes, a _legitimate timeout_ would gracefully allow, e.g. 10 minutes. I'm worried if we do collect zombie tasks for some reason, they should be cleaned up quicker.

khvn26: Yup fair enough. 8e6981c55.

emyller: Reads much better now, thank you!

emyller (dismissed):
Thanks for addressing all comments. This looks good to ship!

emyller (dismissed):
Nice and public!

matthewelwell (approved):
Approving based on evandro's previous review and minor diff since.

PR #7467, opened 2026-05-09
feat(Segment membership inspection PoC): Surface identity counts in segments UI

khvn26:

## Changes

Contributes to #7464 (backend merged).

Surfaces the new `Segment.memberships` field on three places in the existing UI:

- Project segments list — total-identity chip on each row.
- Identities tab on the segment edit page — same chip next to the tab label.
- Inside the Identities tab — per-env chip beside each option in the environment select; selecting an env reveals the full `last_synced_at` timestamp underneath.

Chip is count-only; a tooltip on hover surfaces `N identities — last synced ~Xd ago`.

Two small enabling changes outside the segments folder:
- Shared `Option` in `project-components.js` now honours `selectProps.formatOptionLabel` when callers pass one (falls back to the existing label/description layout).
- Drops `!important` on the global `.chip` `align-self` so the badge can opt into vertical centering inline.

### Screenshots

Identity count chips in the Segments list:

Identity count chips in the Segments list

Identity count chips in the Segments list, with tooltip

Identity count chip on the Identities tab in the Segment page:

Identity count chip on the Identities tab in the Segment page

Per-environment chips:

Per-environment chips

Last sync date for selected environment:

Last sync date for selected environment

Manually drove the segments list, segment edit page, and the Identities tab in a local dev server pointed at staging, with a fetch interceptor injecting synthetic `memberships` arrays into the segments responses. Verified the total badge on each row, the badge next to the Identities tab label, per-env chips inside the select options, the tooltip text on hover, and the full timestamp display on selection — including the `Last synced: —` placeholder before any selection.

Added one Playwright spec (`e2e/tests/segment-membership-test.pw.ts`) using `page.route` to stub `memberships` end-to-end against a real seeded segment, asserting the total badge in the list, the badge on the tab, per-env badges in the select, and the full-timestamp line on selection.


frontend/web/components/segments/SegmentMembershipBadge.tsx
```diff
@@ -0,0 +1,89 @@
...
+  return (
+    <span
+      className='chip chip--xs bg-primary text-white ms-3'
+      style={{ border: 'none', alignSelf: 'center', verticalAlign: 'middle' }}
+      data-test={dataTest}
+    >
+      <UsersIcon className='chip-svg-icon' />
+      <span>
+        {count}
+        {compact ? '' : ` ${noun}`}
```
Zaimwa9: I would either just get rid of mentioning "identities" as this badge has been already used or use compact everywhere (but it becomes dead code).
Regarding the last sync, this could be a tooltip on hover to not overload the UI

kyle-ssg:
I feel like we could neaten up the badges by putting the word identities and ~x days ago into a descriptive tooltip like we do for override badges. Other than that it looks great!

khvn26:
@kyle-ssg The tooltips worked really badly with the tab and the environment select; that's how I came to dumping everything in the chip label. Besides that, I feel like last sync time is important info to surface before we switch to CDC / pure Snowflake lookups for counts.

frontend/web/components/modals/CreateSegmentUsersTabContent.tsx
```diff
@@ -88,6 +99,37 @@ const CreateSegmentUsersTabContent: React.FC<
   setPage,
   setSearchInput,
 }) => {
+  const membershipByEnvId = React.useMemo(() => {
+    const map = new Map<number, SegmentMembership>()
+    ;(memberships ?? []).forEach((m) => map.set(m.environment, m))
+    return map
+  }, [memberships])
+
+  const renderEnvOption = (data: unknown) => {
```
Zaimwa9: We should be able to use `data: EnvOption` directly no ?

frontend/web/components/segments/SegmentMembershipBadge.tsx
```diff
@@ -0,0 +1,89 @@
...
+      count={total}
+      ago={latest ? shortAgo(latest) : undefined}
+      dataTest='segment-membership-total'
+    />
+  )
+}
+
+type EnvProps = {
+  membership: SegmentMembership
+  environment?: Environment
```
Zaimwa9: NIT typing: I feel like the environment should always be present in those components ? So it should be required here and cascaded in all these new components (e.g [here](https://github.com/Flagsmith/flagsmith/pull/7467/changes#diff-db09b3f80e4a34eb1efa1ffed01bd4dbee940ead7c2e9cf42d547a0339e43b2cR114-R122))

Zaimwa9 (dismissed):
2 comments on my side:
- A subjective one related to the verbosity of the chip
- As environments are central to those components, it looks to me that we have some room for typing improvements, if you agree on it we could have another pass there sooner than later

frontend/web/components/segments/SegmentMembershipBadge.tsx
Zaimwa9: I'm also thinking that we'll need to display the breakdown per environment for the identities. Could be through the tooltip but we can discuss the obstacles you had implementing it

talissoncosta:
is it ready for review @khvn26 ?

frontend/web/components/segments/SegmentMembershipBadge.tsx
khvn26: Addressed in f87b4ebc3 — chip is now count-only, with `N identities — last synced ~Xd ago` in a hover tooltip.

khvn26: Holding off on per-env breakdown in the tooltip — that breakdown is already one click away inside the Identities tab (each env option carries its own chip + tooltip), and a tooltip on the list row would grow unbounded with env count.

frontend/web/components/segments/SegmentMembershipBadge.tsx
khvn26: Good call, fixed in 8e80b526d.

frontend/web/components/modals/CreateSegmentUsersTabContent.tsx
khvn26: Addressed in edfb511ab.

khvn26:
@kyle-ssg addressed in f87b4ebc3 — chip is now count-only, `identities` + `~Xd ago` live in a tooltip on hover (same style as the override badges you mentioned).

khvn26:
@talissoncosta Sorry, not quite ready yet — I'd like to fix the backend on staging first so the E2E test is more honest.

talissoncosta:
No rush, take your time, lmk when I can have a look on it.

khvn26:
@talissoncosta I've decided to remove the E2E test, thus unblocking this PR. This is now ready for a re-review.

frontend/web/components/modals/CreateSegmentUsersTabContent.tsx
```diff
@@ -88,6 +95,36 @@ const CreateSegmentUsersTabContent: React.FC<
...
+            environment={environment}
+          />
+        )}
+      </span>
+    )
+  }
+
+  const selectedMembership = React.useMemo(() => {
+    if (!environmentId) return null
+    const envs = (ProjectStore.getEnvs() as Environment[] | null) || []
```
talissoncosta: Can you replace this with a RTK Query ? 

```tsx
  const { data: environmentsData } = useGetEnvironmentsQuery(
    { projectId: Number(projectId) },
    { skip: !projectId },
  )
  const envs = environmentsData?.results ?? []

  const selectedMembership = React.useMemo(() => {
    if (!environmentId) return null
    const env = envs.find((e) => e.api_key === environmentId)
    return env ? membershipByEnvId.get(env.id) ?? null : null
  }, [environmentId, membershipByEnvId, envs])
```

frontend/web/components/segments/SegmentMembershipBadge.tsx
```diff
@@ -0,0 +1,84 @@
...
+type ChipProps = {
+  count: number
+  dataTest?: string
+  tooltip: string
+}
+
+const Chip: FC<ChipProps> = ({ count, dataTest, tooltip }) => (
+  <Tooltip
+    plainText
+    delayShow={100}
```
talissoncosta, line 32: Do you mind giving a context why do we need this specific delay time ?

frontend/web/components/modals/CreateSegmentUsersTabContent.tsx
```diff
@@ -100,13 +137,24 @@ const CreateSegmentUsersTabContent: React.FC<
...
+                  formatOptionLabel={renderEnvOption}
+                />
+                <div className='text-muted fs-small mt-2'>
+                  Last synced:{' '}
+                  {selectedMembership
+                    ? moment(selectedMembership.last_synced_at).format(
+                        'Do MMM YYYY HH:mm:ss',
+                      )
+                    : '—'}
+                </div>
```
talissoncosta, line 161: `Suggestion:` The "Last synced: —" line renders unconditionally before the user picks an environment. What if we only render the line once selectedMembership exists ?

```tsx
  {selectedMembership && (
    <div className='text-muted fs-small mt-2'>
      Last synced: {moment(selectedMembership.last_synced_at).format(
        'Do MMM YYYY HH:mm:ss',
      )}    
    </div>
  )}
```

talissoncosta (commented):
Looks pretty good @khvn26 ! Good job! Just some minor comments, not blocking at all, but If you can clarify them before moving forward I would really appreciate it.

frontend/web/components/modals/CreateSegmentUsersTabContent.tsx
khvn26: Addressed in 9abd45636 — `ProjectStore.getEnvs()` swapped for `useGetEnvironmentsQuery`, derivations updated accordingly.

frontend/web/components/segments/SegmentMembershipBadge.tsx
khvn26: Tooltip text on this chip is the canonical count + sync time, not hover-as-help, so users hover it deliberately. At the default 500ms the lag feels broken on dense list rows; 100ms is short enough to feel instant on intentional hover but long enough that pass-throughs (e.g. scrolling the segments list with the cursor over a row) don't flash a tooltip on every row. The default stays at 500ms for every other caller.

frontend/web/components/modals/CreateSegmentUsersTabContent.tsx
khvn26: My rationale behind adding this was that the page layout is jumping around without it. I guess we could remove it keeping the layout intact somehow? Any recommendations here?

talissoncosta: I see your point. It is all good. 😃

talissoncosta (approved):

PR #7475, opened 2026-05-10
 exploratory-review(segment-membership-ui): proposed fixes for PR #7467 -

talissoncosta:

## Changes

Stacked on top of #7467 (`feat/segment-membership-counts-ui`). Proposed fixes for the review concerns raised on that PR — opening as a PR so the diff is reviewable side-by-side. Cherry-pick what resonates, drop what doesn't.

### Design-system / SCSS

- **Restore `!important` on `.chip { align-self }`.** Was removed in #7467 only so the new badge's inline `alignSelf: 'center'` could win. Reverting it avoids a global regression risk for every other `.chip` consumer (`Feature-Specific` on the segment row, `BetaFlag`, `FeatureHistory`, `SegmentOverrides`, …).
- **Add `.chip--filled` modifier** for the borderless filled variant. Removes the need for inline `style={{ border: 'none', alignSelf: 'center', verticalAlign: 'middle' }}` overrides on the badge.

### Component shape

- **Drop the private `<Chip>` wrapper** in `SegmentMembershipBadge.tsx`. The codebase already uses `.chip chip--xs` directly with bg utility classes everywhere else — half-introducing a private wrapper inside a feature file was the worst of both worlds (not extracted to `components/base/`, but also no longer matching the inline pattern). Now the badge composes `.chip chip--xs chip--filled` inline like its neighbours.
- **Split tab vs. row.** Two components instead of one with a `compact` prop:
  - `SegmentMembershipTotalBadge` — filled chip on the segments-list row (unchanged appearance besides losing `~ago`).
  - `SegmentMembershipTabCount` — muted `(71)` count next to the Identities tab label. **Fixes the purple-on-purple contrast issue on the active Identities tab**, and matches the visual weight of a tab badge convention.
- **Remove the `compact` prop.** It only dropped the noun, not the `~ago`, so the "compact" tab badge still rendered as `71 ~23s ago` — not compact, and misleading-named.

### Behaviour

- **Drop `shortAgo` / `~Xs ago`.** It froze on render (`Date.now()` evaluated once), and duplicated the precise `Last synced: 7th May 2026 14:23:01` line that already lives below the env select in the Identities tab. The full timestamp is the right surface for freshness.
- **`ProjectStore.getEnvs()` → `useGetEnvironmentsQuery`** in `CreateSegmentUsersTabContent.tsx`. `EnvironmentSelect` already uses the same hook, so RTK Query caches the result — no extra request. Also drops the dead-code `ProjectStore` fallback in `SegmentMembershipEnvBadge` (the only caller always passed `environment`).

### Env dropdown options

- **Right-align the count** with `justify-content-between flex-grow-1` on the option row. Fixes the visual misalignment when env names vary in length (the `t` row in the screenshot vs. `Development`).
- **`SegmentMembershipEnvBadge` → `SegmentMembershipEnvCount`** — switches from a filled chip to muted right-aligned text. Inverts the visual hierarchy so the env name is the primary read and the count is secondary metadata. Matches the convention used elsewhere for in-option counts.

### Not addressed (out of scope, worth a follow-up discussion)

- **What does the total "71" actually mean?** Summing memberships across envs double-counts identities that are in the segment in multiple envs. Either rename (`Across N envs`), scope to one env, or accept-and-document. Asking on #7467 directly.
- **Missing vs. zero ambiguity in the env dropdown.** If `memberships` omits an env entirely, no chip renders — but the user can't distinguish "0 identities" from "not yet synced". Probably wants a backend contract decision (always emit a row per env with `count: 0`?).
- **Sort order of env options.** Alphabetical today; reach-by-count might be more useful for this inspector. Out of scope.

### E2E test

The existing `e2e/tests/segment-membership-test.pw.ts` selectors (`[data-test="segment-membership-total"]`, `[data-test="segment-membership-${api_key}"]`) are preserved so the spec keeps passing without modification.

Manually with a local `transformResponse` mock injecting synthetic `memberships` into the segments responses (since the backend PR isn't merged yet). Verified:
- Segments list row chip — filled chip, no `~ago` suffix.
- Identities tab label — muted `(N)` count, readable on both inactive (light) and active (purple) tab states.
- Env dropdown options — name on the left, muted count right-aligned to the column.
- Selected env → `Last synced:` line populates with the precise timestamp.
- Singular copy (`1 identity`) when count is 1.
- Empty memberships → no chip/count rendered anywhere.

Existing Playwright spec (`segment-membership-test.pw.ts`) was inspected to confirm `data-test` selectors still match — not re-run because it requires the seeded segments project.

### DIFF
--- a/frontend/web/components/modals/CreateSegment.tsx
+++ b/frontend/web/components/modals/CreateSegment.tsx
@@ -28,7 +28,7 @@ import {
 } from 'common/services/useSegment'
 import Utils from 'common/utils/utils'
 import AssociatedSegmentOverrides from 'components/segments/AssociatedSegmentOverrides'
-import { SegmentMembershipTotalBadge } from 'components/segments/SegmentMembershipBadge'
+import { SegmentMembershipTabCount } from 'components/segments/SegmentMembershipBadge'
 import Button from 'components/base/forms/Button'
 import InfoMessage from 'components/InfoMessage'
 import InputGroup from 'components/base/forms/InputGroup'
@@ -588,10 +588,7 @@ const CreateSegment: FC<CreateSegmentType> = ({
             tabLabel={
               <>
                 Identities
-                <SegmentMembershipTotalBadge
-                  compact
-                  memberships={segment.memberships}
-                />
+                <SegmentMembershipTabCount memberships={segment.memberships} />
               </>
             }
           >
--- a/frontend/web/components/modals/CreateSegmentUsersTabContent.tsx
+++ b/frontend/web/components/modals/CreateSegmentUsersTabContent.tsx
@@ -11,9 +11,9 @@ import {
   identitySegmentService,
   useGetIdentitySegmentsQuery,
 } from 'common/services/useIdentitySegment'
+import { useGetEnvironmentsQuery } from 'common/services/useEnvironment'
 import { getStore } from 'common/store'
-import ProjectStore from 'common/stores/project-store'
-import { SegmentMembershipEnvBadge } from 'components/segments/SegmentMembershipBadge'
+import { SegmentMembershipEnvCount } from 'components/segments/SegmentMembershipBadge'
 
 interface CreateSegmentUsersTabContentProps {
   projectId: string | number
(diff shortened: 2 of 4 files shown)


talissoncosta:
Closing — opened prematurely without confirmation. Will reopen after Talisson validates the proposed changes locally.

PR #7476, opened 2026-05-11
feat: display null feature values

### LINKED ISSUE #7247: UI: Distinguish null vs empty string for feature values
## Description

The UI renders both `null` and `""` (empty string) feature values identically — as a blank field. There's no way to visually distinguish a flag with no value (`null`) from one explicitly set to an empty string.

This causes:
- Confusion when reviewing flag values
- Misleading diffs in `/compare` (a `null` vs `""` change shows as no visible difference)
- Uncertainty about how defaults should be treated

## Context

PR #7146 made the UI "clear value" flow set values back to `null` instead of `""`, which removes one source of inconsistency for new edits. However:

- Historical data still holds `""` values
- `""` can still be set via the API or SDKs (not only the UI clear flow)
- The visual distinction problem remains regardless of how the value was created

## Proposed Solution

Show an explicit placeholder when a value is `null`, using muted/italic text such as `(no value)` or `(not set)`.

This avoids exposing the literal word `null` to non-technical users while clearly differentiating `null` from `""`.

### Examples
- Value is `null` → display `(no value)` in muted italic
- Value is `""` → display empty input as today (or with quotes around it for extra clarity)
- Value is `"hello"` → display `hello` as today

## Where this appears
- Feature value cells in the features list
- Environment comparison view (`/compare`)
- Segment overrides
- Any other place a feature value is rendered

## Reported by
Customer feedback.

Zaimwa9:

## Changes

Closes #7247
- **null/undefined values** → render as muted `[no value]` chip
- **empty strings** → render as `""` chip (unchanged behavior)
- Regular values → unchanged

- Visually

**not opted-in**
image

**Opted-in**

image


frontend/web/main.js
```diff
@@ -100,7 +102,11 @@ setTimeout(() => {
   }
 
   const root = createRoot(rootElement)
-  root.render(<Router basename={Project.basename || ''}>{routes}</Router>)
+  root.render(
+    <FlagsmithProvider flagsmith={flagsmith}>
```
Zaimwa9, line 106: So. I hesitated to push for this change. The problem I had is that I wanted to rely only on the flag to opt-in. It could use the localStorage though, let me know what you think but it does seem like going into the right direction according to me.
- `Utils.getFlagsmithHasFeature` is just a plain call to `hasFeature`, no flags fetching and doesn't trigger a re-render
- `setTrait` updates the internal state but not React

While on the opposite the `FlagsmithProvider` attaches a `_trigger` callback that calls the SDK. `useFlags` is subscribed to this trigger and allows a re-render
Additionally, `<FlagsmithProvider flagsmith={flagsmith}>` only passes the singleton into the context, no new client, so I haven't identified a state conflict risks.

Last but not least, with all the migration to `FC` that we did, it does make sense that our flagsmith usage switches to using hooks.

@kyle-ssg , am I missing something, wdyt?

Zaimwa9:
Closing in favor of #7498

PR #7478, opened 2026-05-11
fix(hubspot): restore orgid_unique write on company

Holmus:
The HubSpot lead tracking integration stopped writing `orgid_unique` after #7147, which removed the company-update path entirely. The `orgid_unique` write got swept up in that removal even though it wasn't part of the underlying problem.

Alex (sales ops) has been manually maintaining `orgid_unique` on every paying customer since. This restores the write - tightly scoped to address @Zaimwa9's concerns on #7147.

## What this PR does
- In `create_lead`, after the contact is created, look up the HubSpot company by the user's email domain and PATCH **only** `orgid_unique` onto it
- Persist a `HubspotOrganisation` row so subsequent users joining the same Flagsmith org don't redo the lookup
- Skip for generic email domains (gmail/yahoo/etc.)

## No duplication with ChargeBee → HubSpot sync
The duplication concern in #7147 was specifically about `active_subscription` - both the Flagsmith integration's `update_company_active_subscription` and the ChargeBee sync were writing the plan id, and the Flagsmith side couldn't distinguish trial from paid, so it caused workflow misclassification.

This PR writes **only `orgid_unique`**. ChargeBee sync writes 12 billing/subscription properties, none of which is `orgid_unique`. Disjoint sets, no overlap:

| System | Properties written |
|--------|--------------------|
| This PR (Flagsmith integration) | `orgid_unique` |
| ChargeBee → HubSpot sync | `subscription_id__chargebee__unique`, `active_subscription`, `current_mrr`, `currency`, `customer_since`, `contract_start`, `contract_end_date`, `billing_status__backend_`, `chargebee_customer_id`, `chargebee_billing_period_unit`, `chargebee_city_country`, `sync_date___cb` |

The regression test `test_create_lead__corporate_email_with_matching_company__writes_orgid_only` asserts the `update_company` kwargs contain no `name` and no `active_subscription`, so a future change can't accidentally re-bundle them.

## Wadi's other concerns from #7147, revisited
- **\"using the domain is too fragile\"** - mitigated, not eliminated. Falls back silently when no company matches (next user retries). Generic-email filter prevents misrouting. The same domain lookup approach was used historically and HubSpot itself deduplicates by domain. Once linked, the `HubspotOrganisation` row prevents redoing the lookup.
- **Name overwrites** - explicit regression test asserts `name` is never in the payload.

## Why now and not in #7147
@Zaimwa9 proposed this exact scope in his April 9 review. I declined because at the time the duplication concern (on `active_subscription`) outweighed the lost `orgid_unique` write. Three weeks of manual maintenance on the CS side surfaced that the org_id signal matters more than I weighted it. Re-instating it with the original problem (`active_subscription` overwrite) explicitly excluded.

## What still works (unchanged from #7147)
Contacts via form submission. HubSpot owns company name/enrichment. ChargeBee sync owns billing/MRR/plan.

## Test plan
- [x] New unit tests pass: \`tests/unit/integrations/lead_tracking/hubspot/\` (35/35)
- [x] Lint clean: \`ruff check integrations/lead_tracking/hubspot/ tests/unit/integrations/lead_tracking/hubspot/\`
- [ ] Manual smoke test in staging after merge

### DIFF
--- a/api/integrations/lead_tracking/hubspot/lead_tracker.py
+++ b/api/integrations/lead_tracking/hubspot/lead_tracker.py
@@ -5,13 +5,23 @@ from typing import Any
 from django.conf import settings
 
 from integrations.lead_tracking.lead_tracking import LeadTracker
-from organisations.models import Organisation
+from organisations.models import HubspotOrganisation, Organisation
 from users.models import FFAdminUser, HubspotLead, HubspotTracker
 
 from .client import HubspotClient
-from .constants import HUBSPOT_FORM_ID_SAAS
+from .constants import GENERIC_EMAIL_DOMAINS, HUBSPOT_FORM_ID_SAAS
 
 logger = logging.getLogger(__name__)
+
+
+def _company_domain_from_email(email: str) -> str | None:
+    """Return the corporate domain from a user email, or None if generic/personal."""
+    if not email or "@" not in email:
+        return None
+    domain = email.split("@", 1)[1].lower()
+    if domain in GENERIC_EMAIL_DOMAINS:
+        return None
+    return domain
 
 
 class HubspotLeadTracker(LeadTracker):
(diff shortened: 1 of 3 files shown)


api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -49,9 +59,49 @@ def create_user_hubspot_contact(self, user: FFAdminUser) -> str | None:
...
+        if not domain:
+            return
+
+        company = self.client.get_company_by_domain(domain)
+        if not company:
+            logger.info(
+                "No HubSpot company found for domain %s; skipping orgid write for organisation %s",
+                domain,
+                organisation.id,
+            )
```
Zaimwa9, line 90: Ok so here I can just see an issue, there's a data-loss risk if 2 users with the same domain create 2 different orgs (could happen with a lack of synchronization or error?).
- User A with `@flagsmith.com` creates Flagsmith => in Hubspot org_id = 1
- User B with same `@flagsmith.com` creates Flagsmith Test (passes the above guard) => Hubspot deduplicates per domain so we have one hubspot org, but org_id becomes 2.

I don't really know how we could distinguish which one is the relevant one. To me it should be edgy and acceptable but wanted to put it on your radar

Zaimwa9 (commented):
One comment to align on. Otherwise fixing the test name and maybe trim some comments

api/integrations/lead_tracking/hubspot/lead_tracker.py
```diff
@@ -49,9 +59,49 @@ def create_user_hubspot_contact(self, user: FFAdminUser) -> str | None:
...
+            return
+
+        # Passing only flagsmith_organisation_id ensures the payload contains
+        # only orgid_unique - no name overwrite, no subscription writes. This
+        # is the constraint that PR #7147 removed for the wrong reasons; this
+        # restores the orgid sync without bringing back the original problems.
```
Zaimwa9: We can remove the part of the comment related to the PR

api/tests/unit/integrations/lead_tracking/hubspot/test_unit_hubspot_lead_tracking.py
```diff
@@ -280,6 +286,166 @@ def test_create_user_hubspot_contact__get_contact_retries__returns_expected_id(
...
+    tracker.create_lead(user=user, organisation=organisation)
+
+    # Then
+    mock_client.get_company_by_domain.assert_called_once_with("example.com")
+    mock_client.update_company.assert_called_once_with(
+        hubspot_company_id=HUBSPOT_COMPANY_ID,
+        flagsmith_organisation_id=organisation.id,
+    )
+    # Regression: must NOT pass a name or subscription - these were the two
+    # things that made the previous integration overwrite enriched data.
```
Zaimwa9: Same I think those comments could be removed

api/integrations/lead_tracking/hubspot/lead_tracker.py
Holmus: Good flag - this is also the pre-#7147 behaviour, so we're not introducing a new failure mode. `HubspotOrganisation` makes it last-write-wins on `orgid_unique` for the rare two-Flagsmith-orgs-one-domain case (our own internal test orgs, mostly). Sales ops corrects manually when they spot it. Agree it's an edge case worth accepting rather than hardening here.

api/integrations/lead_tracking/hubspot/lead_tracker.py
Holmus: Done in 3c50a1200.

api/tests/unit/integrations/lead_tracking/hubspot/test_unit_hubspot_lead_tracking.py
Holmus: Done in 3c50a1200 - dropped the PR reference from the docstring, removed the inline regression comment, and renamed the test to `test_create_lead__matching_hubspot_company__links_organisation`.

khvn26 (approved):

PR #7488, opened 2026-05-12
fix(versioning): reject legacy feature-state writes on v2 environments

### LINKED ISSUE #7487: Legacy /featurestates/ endpoints silently bypass the version graph on v2 environments
## Bug

The legacy feature-state CRUD endpoints under `BaseFeatureStateViewSet` (`api/features/views.py:661-755`) accept writes and mutate the live `FeatureState` row in place on environments with Feature Versioning v2 enabled. They do not create a new `EnvironmentFeatureVersion`, so the change bypasses the version graph entirely — no history, no audit `EF_VERSION` row, no `NEW_VERSION_PUBLISHED` webhook.

Affected endpoints:

- `POST /api/v1/environments/{api_key}/featurestates/`
- `PUT /api/v1/environments/{api_key}/featurestates/{id}/`
- `PATCH /api/v1/environments/{api_key}/featurestates/{id}/`

The supported write paths on v2 environments are the new `EnvironmentFeatureVersionViewSet` (`/environments/{env}/features/{feature}/versions/...`) and the experimental `update-flag-v1` / `update-flag-v2` / `delete-segment-override` endpoints. Both route through `versioning_service.update_flag()` (`api/features/versioning/versioning_service.py:111-117`) which branches on `Environment.use_v2_feature_versioning` and creates a new published version per call.

## Reproduction

1. Enable Feature Versioning v2 on an environment.
2. `POST /api/v1/environments/{api_key}/featurestates/` with a flag value change against that environment.
3. Observe:
   - 201 / 200 returned with no error.
   - The latest `FeatureState` row for that feature has its values mutated.
   - **No new `EnvironmentFeatureVersion` row** exists for the change.
   - **No `EF_VERSION` audit log row** is created (the audit signal short-circuits because `environment_feature_version_id` is already set on the mutated row).
   - **No `NEW_VERSION_PUBLISHED` webhook** fires.
   - The dashboard's History tab does not show the change.

## Impact

Customers calling these endpoints from automation or CI/CD on a v2 environment:

- Silently lose version history for their changes — those edits become invisible to the History tab, the versions list endpoint, and the audit log webhook for the v2 path.
- Bypass change-request workflows even on environments where their other tooling respects them.
- Have no version to roll back to from the dashboard.
- See their audit-log row count for flag changes drop below what they expect — affects compliance and metrics dashboards.

The behaviour is not customer-facing-documented anywhere. Without a code change, the docs page being prepared in #7477 has to carry a `:::caution` block telling readers the legacy endpoints silently bypass the version graph.

## Why this slipped through

`BaseFeatureStateViewSet` (`api/features/views.py:661-755`) was the v1 write path. When v2 versioning shipped, the supported write paths moved to `versioning_service.update_flag()` and the new `EnvironmentFeatureVersionViewSet`, but the legacy view set was not gated on `Environment.use_v2_feature_versioning`. The `create` action is even marked `DEPRECATED: please use /features/featurestates/ instead` (`api/features/views.py:728`), but the deprecation never escalated to a hard refusal on v2 environments.

gagantrivedi:

## Changes

Closes #7487

The non-versioning feature-state endpoints mutated the live `FeatureState` row in place on v2-versioned environments — bypassing the version graph (no new `EnvironmentFeatureVersion`, no audit row, no `NEW_VERSION_PUBLISHED` webhook) and silently losing the change on rollback.

This PR rejects `POST`/`PUT`/`PATCH`/`DELETE` against those endpoints with `HTTP 400` on v2 environments and points callers at the versioning endpoint. Identity overrides remain allowed (they are not part of the version graph in v2).

Blocked routes (on v2 envs unless identity-scoped):
- `POST`/`PUT`/`PATCH`/`DELETE /api/v1/environments/{api_key}/featurestates/[{id}/]`
- `POST`/`PUT`/`PATCH /api/v1/features/featurestates/[{id}/]`

8 new unit tests in `api/tests/unit/features/test_unit_features_views.py` cover:
- `POST`/`PUT` on both endpoints → 400 on v2
- `DELETE` on the nested endpoint → 400 on v2 (row still exists)
- Identity-scoped writes → still 200/201/204
- The response detail mentions v2 feature versioning

Each blocking test is parametrised over `admin_client` and `admin_master_api_key_client`. mypy clean on the touched files.

### DIFF
--- a/api/features/versioning/exceptions.py
+++ b/api/features/versioning/exceptions.py
@@ -12,3 +12,9 @@ class FeatureVersionDeleteError(FeatureVersioningError):
 
 class CannotModifyLiveVersionError(FeatureVersioningError):
     status_code = status.HTTP_400_BAD_REQUEST
+
+
+class DirectFeatureStateWriteNotAllowedError(FeatureVersioningError):
+    status_code = status.HTTP_400_BAD_REQUEST
+    default_code = "direct_feature_state_write_not_allowed"
+    default_detail = "This environment uses v2 feature versioning. Use the environment feature version endpoint instead."
--- a/api/features/exceptions.py
+++ b/api/features/exceptions.py
@@ -13,5 +13,3 @@ class FeatureStateVersionAlreadyExistsError(FeatureStateVersionError):
         super(FeatureStateVersionAlreadyExistsError, self).__init__(
             f"Version {version} already exists for FeatureState."
         )
-
-
--- a/api/features/exceptions.py
+++ b/api/features/exceptions.py
@@ -13,3 +13,20 @@ class FeatureStateVersionAlreadyExistsError(FeatureStateVersionError):
         super(FeatureStateVersionAlreadyExistsError, self).__init__(
             f"Version {version} already exists for FeatureState."
         )
+
+
+class FeatureStateV2VersioningRequiredError(APIException):
+    status_code = status.HTTP_400_BAD_REQUEST
+    default_code = "v2_feature_versioning_required"
+
+    def __init__(
+        self, environment_api_key: str, feature_id: int | None = None
+    ) -> None:
+        feature_path = str(feature_id) if feature_id is not None else "<feature-id>"
+        super().__init__(
+            detail=(
+                "This environment uses v2 feature versioning. Create a new "
+                "environment feature version via POST /api/v1/environments/"
+                f"{environment_api_key}/features/{feature_path}/versions/."
+            )
+        )
--- a/api/features/exceptions.py
+++ b/api/features/exceptions.py
@@ -15,16 +15,3 @@ class FeatureStateVersionAlreadyExistsError(FeatureStateVersionError):
         )
 
 
-class FeatureStateV2VersioningRequiredError(APIException):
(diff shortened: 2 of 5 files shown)


api/features/versioning/exceptions.py
```diff
@@ -12,3 +12,9 @@ class FeatureVersionDeleteError(FeatureVersioningError):
 
 class CannotModifyLiveVersionError(FeatureVersioningError):
     status_code = status.HTTP_400_BAD_REQUEST
+
+
+class DirectFeatureStateWriteNotAllowedError(FeatureVersioningError):
+    status_code = status.HTTP_400_BAD_REQUEST
+    default_code = "direct_feature_state_write_not_allowed"
+    default_detail = "This environment uses v2 feature versioning. Use the environment feature version endpoint instead."
```
Zaimwa9, line 20: NIT: we could share the endpoint directly?

Zaimwa9 (approved):
1 nit otherwise lgtm

api/features/versioning/exceptions.py
gagantrivedi: Our endpoints for this are still experimental

PR #7497, opened 2026-05-12
feat: bypass pre versioning scaleup unlimited audit and feature visibility

Zaimwa9:

## Changes
- `scale-up-v4` plans branch out of the `VERSIONING_RELEASE_DATE` to use the real subscription_metadata_cache related to audit log and feature history
- Audit log and feature history were previously set to unlimited for old subscriptions

- New tests

### DIFF
--- a/api/organisations/models.py
+++ b/api/organisations/models.py
@@ -424,12 +424,9 @@ class Subscription(LifecycleModelMixin, SoftDeleteExportableModel):  # type: ign
         # cache values.
         is_scale_up = self.subscription_plan_family == SubscriptionPlanFamily.SCALE_UP
         is_scale_up_v4 = (self.plan or "").startswith("scale-up-v4")
-        is_sub_before_versioning_release = (
-            settings.VERSIONING_RELEASE_DATE is None
-            or (
-                self.subscription_date is not None
-                and self.subscription_date < settings.VERSIONING_RELEASE_DATE
-            )
+        is_sub_before_versioning_release = settings.VERSIONING_RELEASE_DATE is None or (
+            self.subscription_date is not None
+            and self.subscription_date < settings.VERSIONING_RELEASE_DATE
         )
         if is_scale_up and not is_scale_up_v4 and is_sub_before_versioning_release:
             cb_metadata.audit_log_visibility_days = None
--- a/api/organisations/models.py
+++ b/api/organisations/models.py
@@ -1,3 +1,4 @@
+import re
 from datetime import timedelta
 from typing import Any
 
@@ -294,12 +295,10 @@ class Subscription(LifecycleModelMixin, SoftDeleteExportableModel):  # type: ign
     def is_enterprise(self) -> bool:
         return self.subscription_plan_family == SubscriptionPlanFamily.ENTERPRISE
 
-    @property
-    def is_sub_before_versioning_release(self) -> bool:
-        return settings.VERSIONING_RELEASE_DATE is None or (
-            self.subscription_date is not None
-            and self.subscription_date < settings.VERSIONING_RELEASE_DATE
-        )
+    def get_scaleup_plan_version(self) -> int:
+        if match := re.match(r"scale-up-v(\d+)", self.plan or ""):
+            return int(match.group(1))
+        return 1
 
     @hook(AFTER_SAVE, when="plan", has_changed=True)
     def update_api_limit_access_block(self):  # type: ignore[no-untyped-def]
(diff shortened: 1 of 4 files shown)


api/organisations/models.py
```diff
@@ -419,15 +419,16 @@ def _get_subscription_metadata_for_chargebee(self) -> ChargebeeObjMetadata:
...
-                and self.subscription_date < settings.VERSIONING_RELEASE_DATE
-            )
-        ):
-            # Logic to grandfather old scale up plan customers to give them
-            # full access to audit log and feature history.
+        # Grandfather old Scale-Up customers with unlimited audit log and
+        # feature history. Scale-Up v4 (and later) plans always honour the
+        # cache values.
+        is_scale_up = self.subscription_plan_family == SubscriptionPlanFamily.SCALE_UP
+        is_scale_up_v4 = (self.plan or "").startswith("scale-up-v4")
```
gagantrivedi: Nit: this can be another value on SubscriptionPlanFamily

api/organisations/models.py
```diff
@@ -419,15 +419,16 @@ def _get_subscription_metadata_for_chargebee(self) -> ChargebeeObjMetadata:
...
+        is_sub_before_versioning_release = settings.VERSIONING_RELEASE_DATE is None or (
```
gagantrivedi: nit: this can be a method on subscription model

gagantrivedi (dismissed):
LGTM

api/organisations/models.py
Zaimwa9: The idea is good but it doesn't really fit the family enum as of imo. It would create a sort of subfamily branching only for this case so for now I would leave it like this and keep it in mind if v2 drifts a lot from v4+. Sounds good?

api/organisations/models.py
Zaimwa9: Addressed here [9ac8c0b](https://github.com/Flagsmith/flagsmith/pull/7497/commits/9ac8c0b2ed3e2f8f7f7c34cd31c833667f783426)

api/organisations/models.py
```diff
@@ -294,6 +294,13 @@ def subscription_plan_family(self) -> SubscriptionPlanFamily:
     def is_enterprise(self) -> bool:
         return self.subscription_plan_family == SubscriptionPlanFamily.ENTERPRISE
 
+    @property
+    def is_sub_before_versioning_release(self) -> bool:
+        return settings.VERSIONING_RELEASE_DATE is None or (
```
matthewelwell: Looks like this setting isn't actually populated in any of our deployments. Should it be?

api/organisations/models.py
```diff
@@ -419,15 +426,12 @@ def _get_subscription_metadata_for_chargebee(self) -> ChargebeeObjMetadata:
...
-                and self.subscription_date < settings.VERSIONING_RELEASE_DATE
-            )
-        ):
-            # Logic to grandfather old scale up plan customers to give them
-            # full access to audit log and feature history.
+        # Grandfather old Scale-Up customers with unlimited audit log and
+        # feature history. Scale-Up v4 (and later) plans always honour the
+        # cache values.
+        is_scale_up = self.subscription_plan_family == SubscriptionPlanFamily.SCALE_UP
+        is_scale_up_v4 = (self.plan or "").startswith("scale-up-v4")
```
matthewelwell: Bonus nit, I prefer: 

```suggestion
        is_scale_up_v4 = self.plan and self.plan.startswith("scale-up-v4")
```

matthewelwell: Should we also try to do something smarter here and determine the version number of the plan to do an inequality check? 

e.g. something like (but better):

```python
scale_up_plan_version = int(re.match(r"scale-up-v([0-9]+)", "scale-up-v4").groups()[0])
```

Then we can change the conditional below to: 

```python
if is_scale_up and scale_up_plan_version < 4 and self.is_sub_before_versioning_release:
```

api/organisations/models.py
Zaimwa9: Ok then that explains everything. It depends how you see it, we could:
1. Remove my new code, set the variable to a date up to which we give all previous advantages (2 weeks ago?)
2. Remove the variable and keep previous advantages on versions below v4 (using the versioning you suggested below) ?

matthewelwell: I think (2) - I don't think the versioning release date should factor into this at all, right? 

In my eyes, what we need to support is: 

audit log history: 

- scale-up-v2 = unlimited
- scale-up-v4+ = 14 days

version history:

- scale-up-v2+ = 14 days

api/organisations/models.py
Zaimwa9: Addressed in [33b659b](https://github.com/Flagsmith/flagsmith/pull/7497/commits/33b659b2881ab73c37bffca6ef66246c2e4a125b)

api/organisations/models.py
Zaimwa9: All addressed in [33b659b](https://github.com/Flagsmith/flagsmith/pull/7497/commits/33b659b2881ab73c37bffca6ef66246c2e4a125b)

matthewelwell (dismissed):

Zaimwa9:
Lint tests staled the review 🙏

matthewelwell (approved):

PR #7498, opened 2026-05-12
feat: compare empty string display

### LINKED ISSUE #7247: UI: Distinguish null vs empty string for feature values

Zaimwa9:

- [ ] I have filled in the "Changes" section below.

## Changes
Closes #7247 

- Display `""` for empty values in compare view 

- Against staging


talissoncosta (dismissed):
Looks good. This is a much cleaner solution. 

Quick question: would there be a strong reason not to make this the default everywhere rather than gating it behind `isCompareEnv`?

The original noise concern was about the `[no value]` chip rendering on every null row, which we've dropped. `""` is rare enough that always rendering it shouldn't add meaningful noise to the features list — and it'd solve the same disambiguation problem on any future surface without per-caller opt-in.

Just raising as a discussion point, not blocking.

Image

Zaimwa9:
@talissoncosta added it in [edf9b9c](https://github.com/Flagsmith/flagsmith/pull/7498/commits/edf9b9cb4d91e42f8de051368971d40950a02a89)
image
Let me know your thoughts

talissoncosta (approved):
Looks great, solid and clean. Thanks for handling that @Zaimwa9

PR #7499, opened 2026-05-12
fix(infra): Drop SECURE_PROXY_SSL_HEADER_NAME override

### LINKED ISSUE #7282: Fix Google SSO redirect_uri on internal.flagsmith.com
Google rejects the OAuth flow because Django generates `redirect_uri=http://internal.flagsmith.com/...`. Root cause: Django's `SECURE_PROXY_SSL_HEADER_NAME` is set to `HTTP_CLOUDFRONT_FORWARDED_PROTO`, but internal.flagsmith.com goes `ALB → ECS` with no CloudFront in the path, so the header isn't present and Django falls back to `http`.

Fix is to trust the standard `X-Forwarded-Proto` header, which both ALBs set. This requires the public ALB's listener to be HTTPS so CloudFront→ALB doesn't stamp `http`, overriding CloudFront's value.

**Changes**

- [x] (infra): Promote public ALB listener to HTTPS
- [x] (infra): Switch CloudFront origin protocol to HTTPS
- [ ] (app): Flip `SECURE_PROXY_SSL_HEADER_NAME` to `HTTP_X_FORWARDED_PROTO` in admin-api task def

germangarces:
The task definition pinned `SECURE_PROXY_SSL_HEADER_NAME` to `HTTP_CLOUDFRONT_FORWARDED_PROTO`, so Django only detected HTTPS when the request came through CloudFront. Requests via the internal ALB (no CloudFront in the path) had no such header, so `is_secure()` returned False and Google SSO callbacks were generated as `http://`, which Google rejects.

Removing the override falls back to the default in `api/app/settings/common.py`: `HTTP_X_FORWARDED_PROTO`. Both ALBs stamp `X-Forwarded-Proto` on their HTTPS listener, so Django detects HTTPS regardless of CloudFront presence.

Contributes to https://github.com/Flagsmith/flagsmith/issues/7282

### DIFF
--- a/infrastructure/aws/production/ecs-task-definition-admin-api.json
+++ b/infrastructure/aws/production/ecs-task-definition-admin-api.json
@@ -138,10 +138,6 @@
                     "value": "flagsmith_project_metadata"
                 },
                 {
-                    "name": "SECURE_PROXY_SSL_HEADER_NAME",
-                    "value": "HTTP_CLOUDFRONT_FORWARDED_PROTO"
-                },
-                {
                     "name": "SENDER_EMAIL",
                     "value": "Flagsmith <support@flagsmith.com>"
                 },


matthewelwell (approved):

matthewelwell:
Note: shouldn't be merged until [this issue](https://github.com/Flagsmith/pulumi/issues/195) is resolved and all changes associated with it are deployed.

PR #7518, opened 2026-05-14
fix: allow project admins to create and manage project-scoped custom fields

### LINKED ISSUE #7424: UI/UX: Project-level custom field fails successfully

- [ ] Self Hosted with Docker
- [x] SaaS at flagsmith.com

### Describe the bug

When a user with Project Administrator role (only - no organisation admin) creates a custom field via Project Settings → Custom Fields, the save flow partially fails. The MetadataField is created, but the call that binds it to a content type returns 403, and the error is surfaced as a generic "Failed to save custom field" toast. The field is left orphaned in the project: it shows up in the Project Fields list but is bound to no entity, so it cannot be set on Features, Segments or Environments.

### Steps To Reproduce

1. Set up a user with Project Administrator role only on a project, granted via a role linked to a group (no org-level admin).
2. Sign in as that user, open the project, go to Project Settings → Custom Fields.
3. Click Create Custom Field. Set name = example, type = int, entity = feature, toggle "Required" on.
4. Click Create Custom Field.

### Expected behavior

The save completes successfully (preferred - project admins are the intended audience for project-scoped custom fields), or
The save is rejected cleanly upfront with a clear permission error, and no MetadataField is created.

### Screenshots

https://github.com/user-attachments/assets/4914963a-9763-49f5-9e0f-2cba92d63dac

10done:

- [X] I have added information to `docs/` if required so people know about the feature.

## Changes

Closes #7424 

##Issue Description
When a user with Project Administrator role (non-org-admin) created a project-scoped custom field via Project Settings → Custom Fields, the save flow partially failed:

1. `POST /api/v1/metadata/fields/` → 201 (MetadataField created)
2. `POST /api/v1/organisations/{id}/metadata-model-fields/` → 403 (binding fails, field left orphaned)

**Root cause:** `MetadataModelFieldPermissions` only checked `is_organisation_admin` for `create`, `update`, and `destroy` actions. `MetadataFieldPermissions` already had the correct project-admin path, but its sibling class did not.

**Fix:** Extended `MetadataModelFieldPermissions` in `api/metadata/permissions.py` to allow project admins to create, update, and delete `MetadataModelField` bindings for project-scoped fields. Org-scoped fields remain restricted to org admins only.

Reproduced the bug via direct API calls as a project-admin user (org role = `USER`, project `admin = true`):

**Before fix:**
- `POST /api/v1/organisations/1/metadata-model-fields/` → `403`{"detail": "You do not have permission to perform this action."}`
Before fix
Before fix

**After fix:**
- `POST /api/v1/organisations/1/metadata-model-fields/` → `201`{"id": 2, "field": 2, "content_type": 55, "is_required_for": []}`
After Fix
After Fix

**Regression tests added **
- Project admin can bind a project-scoped field to a content type (the bug case)
- Project admin cannot bind an org-scoped field (security boundary)
- Project admin can update/delete bindings for their own project's fields
- Project admin cannot update/delete bindings for other projects' or org-scoped fields

Since I cannot access the Enterprise plan I tried to fix it via the API calls. I hope it fixes the issue.

### DIFF
--- a/api/metadata/permissions.py
+++ b/api/metadata/permissions.py
@@ -76,8 +76,8 @@ class MetadataModelFieldPermissions(IsAuthenticated):
 
                     return (organisation_pk == field.organisation.id) and (
                         request.user.is_organisation_admin(organisation_pk)
-                        or (project_id := field.project)
-                        and request.user.is_project_admin(project_id)
+                        or (project := field.project)
+                        and request.user.is_project_admin(project)
                     )
 
         return False
@@ -89,8 +89,8 @@ class MetadataModelFieldPermissions(IsAuthenticated):
         if view.action in ("update", "destroy", "partial_update"):
             return (
                 request.user.is_organisation_admin(obj.field.organisation)
-                or (project_id := obj.field.project)
-                and request.user.is_project_admin(project_id)
+                or (project := obj.field.project)
+                and request.user.is_project_admin(project)
             )
 
         return False
--- a/api/metadata/permissions.py
+++ b/api/metadata/permissions.py
@@ -74,10 +74,14 @@ class MetadataModelFieldPermissions(IsAuthenticated):
                 if view.action == "create":
                     field = MetadataField.objects.get(id=request.data.get("field"))
 
-                    return (
-                        request.user.is_organisation_admin(organisation_pk)
-                        and organisation_pk == field.organisation.id
-                    )
+                    if organisation_pk != field.organisation.id:
+                        return False
+
+                    if request.user.is_organisation_admin(organisation_pk):
+                        return True
+
+                    if field.project is not None:
+                        return request.user.is_project_admin(field.project)
 
         return False
 
@@ -86,6 +90,10 @@ class MetadataModelFieldPermissions(IsAuthenticated):
             return request.user.belongs_to(obj.field.organisation.id)
 
         if view.action in ("update", "destroy", "partial_update"):
-            return request.user.is_organisation_admin(obj.field.organisation)
+            if request.user.is_organisation_admin(obj.field.organisation):
+                return True
+
+            if obj.field.project is not None:
+                return request.user.is_project_admin(obj.field.project)
 
         return False
--- a/api/metadata/permissions.py
+++ b/api/metadata/permissions.py
@@ -74,14 +74,11 @@ class MetadataModelFieldPermissions(IsAuthenticated):
                 if view.action == "create":
                     field = MetadataField.objects.get(id=request.data.get("field"))
 
-                    if organisation_pk != field.organisation.id:
-                        return False
-
-                    if request.user.is_organisation_admin(organisation_pk):
-                        return True
-
-                    if field.project is not None:
-                        return request.user.is_project_admin(field.project)
+                    return (organisation_pk == field.organisation.id) and (
+                        request.user.is_organisation_admin(organisation_pk)
+                        or (project_id := field.project)
+                        and request.user.is_project_admin(project_id)
+                    )
 
         return False
 
@@ -90,10 +87,10 @@ class MetadataModelFieldPermissions(IsAuthenticated):
             return request.user.belongs_to(obj.field.organisation.id)
 
         if view.action in ("update", "destroy", "partial_update"):
-            if request.user.is_organisation_admin(obj.field.organisation):
-                return True
-
-            if obj.field.project is not None:
-                return request.user.is_project_admin(obj.field.project)
+            return (
+                request.user.is_organisation_admin(obj.field.organisation)
+                or (project_id := obj.field.project)
+                and request.user.is_project_admin(project_id)
+            )
 
         return False
--- a/api/tests/unit/metadata/test_views.py
+++ b/api/tests/unit/metadata/test_views.py
@@ -1007,7 +1007,7 @@ def test_create_model_metadata_field__project_admin__returns_201(
     assert response.json()["field"] == project_field.id
 
 
-def test_create_model_metadata_field__project_admin__org_scoped_field__returns_403(
+def test_create_model_metadata_field__project_admin_org_scoped_field__returns_403(
     staff_user: FFAdminUser,
     staff_client: APIClient,
     organisation: Organisation,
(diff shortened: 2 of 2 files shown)


10done:
@khvn26 Can you please review whenever you get time. Thank you.

api/tests/unit/metadata/test_views.py
```diff
@@ -969,3 +969,140 @@ def test_modify_metadata_field__project_admin__permission_check(
...
+
+def test_create_model_metadata_field__project_admin__returns_201(
+    staff_user: FFAdminUser,
+    staff_client: APIClient,
+    organisation: Organisation,
+    project: Project,
+    feature_content_type: ContentType,
+) -> None:
+    # Given - a project-scoped MetadataField and a user who is project admin (not org admin)
+    from projects.models import UserProjectPermission
```
matthewelwell: I see that you're just following the pattern from the previous test, but I'm not sure why the local import exists there (in fact, I just tested importing it at the module level and it works just fine). Can you remove it here, and import it at the module level instead? 

While you're at it, if you want to remove it from the previous test too, that'd be awesome, thanks!

api/tests/unit/metadata/test_views.py
```diff
@@ -969,3 +969,140 @@ def test_modify_metadata_field__project_admin__permission_check(
...
+    ids=[
+        "own_project_field_update_allowed",
+        "own_project_field_delete_allowed",
+        "org_scoped_field_update_denied",
+        "org_scoped_field_delete_denied",
+        "other_project_field_update_denied",
+        "other_project_field_delete_denied",
+    ],
+)
+def test_modify_model_metadata_field__project_admin__permission_check(
```
matthewelwell: While I love a bit of parameterisation, I think this is a bit much - can we split the update and delete tests at least please?

api/metadata/permissions.py
```diff
@@ -74,10 +74,14 @@ def has_permission(self, request, view):  # type: ignore[no-untyped-def]
...
-                        and organisation_pk == field.organisation.id
-                    )
+                    if organisation_pk != field.organisation.id:
+                        return False
+
+                    if request.user.is_organisation_admin(organisation_pk):
+                        return True
+
+                    if field.project is not None:
+                        return request.user.is_project_admin(field.project)
```
matthewelwell: Maybe this more explicit approach is clearer, but I'm a sucker for more concise code, so I'd prefer something like the following. 


```suggestion
                    return (
                        (organisation_pk == field.organisation.id) 
                        and (
                            request.user.is_organisation_admin(organisation_pk) 
                            or (project_id := field.project and request.user.is_project_admin(project_id))
                        )
                    )
```

api/metadata/permissions.py
```diff
@@ -86,6 +90,10 @@ def has_object_permission(self, request, view, obj):  # type: ignore[no-untyped-
             return request.user.belongs_to(obj.field.organisation.id)
 
         if view.action in ("update", "destroy", "partial_update"):
-            return request.user.is_organisation_admin(obj.field.organisation)
+            if request.user.is_organisation_admin(obj.field.organisation):
+                return True
+
+            if obj.field.project is not None:
+                return request.user.is_project_admin(obj.field.project)
```
matthewelwell: As above, let's be more concise

```suggestion
            return (
                request.user.is_organisation_admin(obj.field.organisation)
                or (project_id := obj.field.project and request.user.is_project_admin(project_id))
            ) 
```

matthewelwell (commented):
Thanks very much for the contribution @10done. On the whole, this looks great. I've added a few comments that need addressing, but only minor things really.

api/tests/unit/metadata/test_views.py
10done: done

api/tests/unit/metadata/test_views.py
10done: done

api/metadata/permissions.py
10done: done

api/metadata/permissions.py
10done: done

10done:
@matthewelwell Thank you for the review on the PR. I have made the changes as suggested and also removed the local imports on other tests too.

api/metadata/permissions.py
```diff
@@ -74,9 +74,10 @@ def has_permission(self, request, view):  # type: ignore[no-untyped-def]
                 if view.action == "create":
                     field = MetadataField.objects.get(id=request.data.get("field"))
 
-                    return (
+                    return (organisation_pk == field.organisation.id) and (
                         request.user.is_organisation_admin(organisation_pk)
-                        and organisation_pk == field.organisation.id
+                        or (project_id := field.project)
+                        and request.user.is_project_admin(project_id)
```
matthewelwell: Sorry @10done, while this code will work, the naming is a little off. `field.project` actually returns the `Project` instance, not the id, so we should change it as follows: 

```suggestion
                        or (project := field.project)
                        and request.user.is_project_admin(project)
```

api/metadata/permissions.py
```diff
@@ -86,6 +87,10 @@ def has_object_permission(self, request, view, obj):  # type: ignore[no-untyped-
             return request.user.belongs_to(obj.field.organisation.id)
 
         if view.action in ("update", "destroy", "partial_update"):
-            return request.user.is_organisation_admin(obj.field.organisation)
+            return (
+                request.user.is_organisation_admin(obj.field.organisation)
+                or (project_id := obj.field.project)
+                and request.user.is_project_admin(project_id)
```
matthewelwell: As above

```suggestion
                or (project := obj.field.project)
                and request.user.is_project_admin(project)
```

matthewelwell (commented):
Thanks for making the changes @10done . I'm sorry, my suggestion was slightly off, so I've made 2 further suggestions just to tidy up some naming, then we should be good to merge 🙌

api/metadata/permissions.py
10done: I have changed the naming as asked

10done:
@matthewelwell I have made the changes as asked. Thank you.

matthewelwell (approved):

matthewelwell:
Thanks @10done

PR #7512, opened 2026-05-13
feat(SCIM): Wire flagsmith-private SCIM app into the api

khvn26:
Thanks for submitting a PR! Please check the boxes below:

- [x] I have read the [Contributing Guide](/Flagsmith/flagsmith/blob/main/CONTRIBUTING.md).
- [ ] I have added information to `docs/` if required so people know about the feature.
- [x] I have filled in the "Changes" section below.
- [x] I have filled in the "How did you test this code" section below.

## Changes

Closes #7150.

Conditionally registers the SCIM Django app shipped by `flagsmith-private` when the `scim` module is importable: adds `django_scim` and `scim` to `INSTALLED_APPS`, sets the django-scim2 `SCIM_SERVICE_PROVIDER` settings (with `AUTH_CHECK_MIDDLEWARE` pointed at `scim.middleware.ScimAuthenticationMiddleware`), and mounts the protocol URLs at `/api/v1/scim/v2/` plus the configuration management URLs at `/api/v1/organisations/<int:organisation_pk>/scim/`.

`flagsmith-private` is now installed from the prod CodeArtifact PyPI repository (`flagsmith-pypi-production`) rather than a git ref, pinned to `>=0.5.1,<1`. The new `.github/actions/codeartifact-login` composite action assumes the `codeartifact-github-actions-production` role via OIDC, fetches a token, and exports it as `UV_INDEX_FLAGSMITH_PYPI_PRODUCTION_*`. Every CI job and Dockerfile stage that installs the `scim` / `release-pipelines` extras runs the login step first. For local dev, `make codeartifact-login` writes the same env vars to a git-ignored `.env-codeartifact` file that the Makefile auto-includes.

Dependabot is wired to the same index via its [native OIDC support for CodeArtifact](https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/manage-your-dependency-security/configuring-access-to-private-registries-for-dependabot#using-oidc-for-authentication) — a `registries:` block in `.github/dependabot.yml` reuses the same IAM role (its trust policy already accepts `repo:Flagsmith/*`). No cron-based token refresh required.

## How did you test this code?

Integration tests under `api/tests/integration/scim/` cover:

- `GET /api/v1/scim/v2/{ServiceProviderConfig,Schemas,ResourceTypes}` with a valid SCIM bearer for an Enterprise org → 200.
- Same endpoints without a bearer → 401.
- Same endpoints with a bearer for a non-Enterprise org → 403.
[...]

PR #7139, opened 2026-04-06
docs: SCIM provisioning

khvn26:
- [x] I have read the [Contributing Guide](/Flagsmith/flagsmith/blob/main/CONTRIBUTING.md).
- [x] I have added information to `docs/` if required so people know about the feature.
- [x] I have filled in the "Changes" section below.
- [x] I have filled in the "How did you test this code" section below.

## Changes

Contributes to #6498

Adds user-facing documentation for SCIM 2.0 provisioning support. This is the docs-first spec — it defines the user experience before implementation begins. Covers:

- User and group lifecycle (provisioning and deprovisioning)
- Setup instructions for Okta, Microsoft Entra ID and OneLogin
- SCIM API overview with link to Swagger
- Token management
- Interaction with existing SAML group sync
- Troubleshooting

Also adds SCIM to the authentication overview page.

### Preview links

- [SCIM provisioning (new)](https://docs-git-docs-scim-provisioning-flagsmith.vercel.app/administration-and-security/access-control/scim)
- [Authentication overview (updated)](https://docs-git-docs-scim-provisioning-flagsmith.vercel.app/administration-and-security/access-control)

## How did you test this code?

Documentation only. Run `npx prettier --check` on changed files — passes.

PR #7528, opened 2026-05-15
feat(SCIM): SCIM configuration management UI

### LINKED ISSUE #7151: Frontend for SCIM configuration management
### Scope

Rename the existing "SAML" tab in Organisation Settings to "SSO", and add a SCIM section within it. SCIM should not be a separate top-level tab.

Within the SCIM section:

- If the org does not have an Enterprise subscription, show an upgrade prompt (follow the pattern used by SAML).
- If no SCIM configuration exists: show a "Create SCIM Configuration" button.
- On creation: display the bearer token once in a copyable field with a warning that it cannot be retrieved later.
- If a SCIM configuration exists: show created date, last token rotation date, and the SCIM base URL.
- "Regenerate Token" button: confirms with the user, then regenerates. Displays the new token once.
- "Delete" button: confirms, then deletes the SCIM configuration.
- The SCIM base URL should be derived from the API base URL (same pattern as the SAML ACS URL).

### Acceptance criteria

- The "SAML" tab is renamed to "SSO" and contains both SAML configuration and SCIM configuration sections.
- Non-Enterprise orgs see an upgrade prompt, not the SCIM configuration UI.
- Creating a SCIM configuration displays the token exactly once.
- After navigating away and returning, the token is no longer visible.
- Regenerating a token shows the new token and invalidates the old one.
- Deleting a SCIM configuration removes it. Existing users and groups are not affected.
- The displayed SCIM base URL is correct for the current Flagsmith instance.

talissoncosta:

## Changes

Closes #7151.

Adds the frontend for the SCIM configuration endpoints from #7512. The SAML tab in Organisation Settings is renamed to SSO and now holds both SAML and SCIM as stacked sections behind a single Enterprise gate.

## Screenshots
image
image

## What it does

- New `useScimConfiguration` service (get / create / delete / regenerate-token) scoped per organisation, cache-tagged for invalidation.
- `ScimSection` covers loading, empty (404), and configured states. Loading is a shimmer skeleton that mirrors the configured layout so the page doesn't jump when data arrives.
- `ScimTokenModal` shows the bearer token once on create or regeneration. It comes straight from the mutation result, never gets cached, and is gone from memory when the modal closes. The Done action triggers an in-modal "Have you saved the token?" step before closing — see the known limitation below.
- `SSOTab` is a thin container that wraps both sections in `PlanBasedBanner feature='SAML'`. SAML's previous internal banner moved up so we don't render duplicate upgrade prompts.
- Tab key migrates from `?tab=saml` to `?tab=sso`. Old bookmarks land on the default tab.

The `base_url` is server-derived (built via `request.build_absolute_uri(reverse("scim:root"))` in `flagsmith-private`) so we just display what comes back — no client-side URL building like SAML's ACS URL.

## CopyField

A few places across the app roll the same "read-only input + Copy button" pattern inline (CreateSAML's ACS URL, SDK keys, server-side keys, webhook URLs, etc.). I added a small `CopyField` here and used it for both the SCIM base URL and the token. The API is deliberately minimal (`value`, `className`, `data-test`) so a follow-up PR can migrate the other consumers without painting us into a corner.

## Known limitation

The modal's X close button bypasses the inline "Have you saved the token?" step. Same trade-off `AdminAPIKeys` has — the proper fix sits at the `withModal` layer, which clobbers `global.closeModal` when `_Confirm` mounts and never restores it. Out of scope here.

## How did you test this?

Manual end-to-end on a local environment with `--extra scim`:

1. Non-Enterprise org: upgrade prompt only.
2. Enterprise org with no SCIM config: SAML section + SCIM empty state with the Create button.
3. Create: token modal shows the monospace token + warning. Copy works. Done triggers the inline confirmation. After confirming, the section renders the configured state.
4. Refresh: token modal does not reappear; configured state persists.
5. Regenerate: confirm dialog → new token in the modal → `token_rotated_at` updates after close.
6. Delete: confirm (destructive) → empty state in place (RTK Query invalidation + 404 from refetch).
7. URL routing: `?tab=sso` opens the SSO tab; old `?tab=saml` lands on the default tab.

Plus `npm run typecheck` and `npx eslint --fix` on modified files. No new errors.


frontend/web/components/pages/organisation-settings/OrganisationSettingsPage.tsx
```diff
@@ -111,10 +111,10 @@ const OrganisationSettingsPage: FC = () => {
       label: 'Webhooks',
     },
     {
-      component: <SAMLTab organisationId={organisation.id} />,
+      component: <SSOTab organisationId={organisation.id} />,
       isVisible: true,
-      key: 'saml',
-      label: 'SAML',
+      key: 'sso',
+      label: 'SSO',
```
khvn26, line 131: Can we redirect from `?tab=saml` to `?tab=sso`?

frontend/web/components/pages/organisation-settings/tabs/SSOTab.tsx
```diff
@@ -0,0 +1,17 @@
...
+import SamlTab from 'components/SamlTab'
+import ScimSection from 'components/ScimSection'
+
+type SSOTabProps = {
+  organisationId: number
+}
+
+export const SSOTab = ({ organisationId }: SSOTabProps) => {
+  return (
+    <PlanBasedBanner feature='SAML' theme='page'>
```
khvn26: This is gated on `SAML` feature, so the `SCIM` entry in `featureDescriptions` isn't really used.

frontend/web/components/ScimSection.tsx
```diff
@@ -0,0 +1,218 @@
+import React, { FC } from 'react'
+import moment from 'moment'
+
+import Button from './base/forms/Button'
+import EmptyState from './EmptyState'
+import Icon from './icons/Icon'
+import Loader from './Loader'
```
khvn26: Isn't `Loader` a global?

frontend/web/components/PlanBasedAccess.tsx
```diff
@@ -93,6 +93,12 @@ export const featureDescriptions: Record<PaidFeature, any> = {
     docs: 'https://docs.flagsmith.com/advanced-use/scheduled-flags',
     title: 'Scheduled Flags',
   },
+  'SCIM': {
+    description:
+      'Provision and de-provision users and groups automatically from your identity provider.',
+    docs: 'https://docs.flagsmith.com/system-administration/authentication/',
```
khvn26: As per #7139:

```suggestion
    docs: 'https://docs.flagsmith.com/administration-and-security/access-control/scim/',
```

khvn26 (commented):
All in all, LGTM — cross-checked with the docs and added a handful of minor comments.

Would you like to cover the happy path with an `@enterprise` Playwright test?

frontend/web/components/pages/organisation-settings/OrganisationSettingsPage.tsx
talissoncosta: Added in 5c6596573.

frontend/web/components/pages/organisation-settings/tabs/SSOTab.tsx
talissoncosta: Good call. Add two independent gates in 61aff57e7 so we can point the user to the right doc.

frontend/web/components/ScimSection.tsx
talissoncosta: Replaced with a Skeleton 166700ee5. Loader import gone.

frontend/web/components/PlanBasedAccess.tsx
talissoncosta: Updated in 61aff57e7

Zaimwa9 (approved):
Code looks good to me, seems to work as expected (although I haven't deep dived in the  details of the specs).

khvn26 (approved):
LGTM 👍

Issue #7368, opened 2026-04-29
API Tests: Apps registry pollution from test_unit_app_routers.py breaks migration tests


khvn26:
The API unit test suite has been failing consistently on the Python 3.11 matrix in Depot CI, and a failure has now reproduced in GitHub Actions: https://github.com/Flagsmith/flagsmith/actions/runs/25070433056/job/73448935538 (run on commit 5c5ebc4f5d6587585d45772236e857651db3f0f6).

## Symptoms

Two failure clusters in the same job:

1. `tests/unit/core/management/test_unit_core_management_makemigrations.py::test_makemigrations__with_check_changes__runs_without_error` exits with `SystemExit: 1`. Captured stdout:

   ```
   Migrations for 'app_analytics':
     app_analytics/migrations/0009_analyticsmodel_mymodel_mymodel1_mymodel2.py
       + Create model AnalyticsModel
       + Create model MyModel
       + Create model MyModel1
       + Create model MyModel2
   ```

2. Roughly a dozen migration tests using `django_test_migrations.Migrator` fail at teardown across multiple xdist workers with:

   ```
   psycopg2.errors.DuplicateTable: relation "app_analytics_analyticsmodel" already exists
   ```

   Affected files include `tests/unit/segments/test_unit_segments_migrations.py`, `tests/unit/features/test_migrations.py`, `tests/unit/organisations/permissions/test_unit_organisations_migrations.py`, `tests/unit/environments/permissions/test_unit_environments_permissions_migrations.py`, `tests/unit/features/versioning/test_unit_versioning_migrations.py`, `tests/unit/projects/test_migrations.py`.

## Root cause

`api/tests/unit/app/test_unit_app_routers.py` defines four `django.db.models.Model` subclasses inline inside parametrised test functions:

- `AnalyticsModel` (line 19) with `app_label = "app_analytics"` (parametrised) or `"another_app"`
- `MyModel` (line 44) with the same parametrisation
- `MyModel1` (line 70) and `MyModel2` (line 74) with the same parametrisation

Defining a Django model class registers it in the apps registry for the lifetime of the worker process. There is no apps registry isolation around these tests. Once these tests run with `app_label="app_analytics"`, the four model classes are permanently registered against the real `app_analytics` app for that worker.

That registration is exactly what we see in cluster 1: `makemigrations --check` finds four models with no migrations and exits non-zero, listing the four polluting model names verbatim.

[...]

PR #7529, opened 2026-05-18
feat: added configurable redis cache for shared redis instances

### LINKED ISSUE #7297: feat: Add configurable Redis cache KEY_PREFIX for shared Redis deployments
## Summary

When self-hosting Flagsmith with a shared Redis instance that uses ACL-based multi-tenancy, all cache keys are written with an empty `KEY_PREFIX` (Django default), producing keys like `:1:environment_document_xxx`. This makes it impossible to restrict Flagsmith's Redis user to a namespace prefix (e.g., `~flagsmith:*`) via Redis ACL rules.

## Current behavior

Flagsmith's `CACHES` configuration in `api/app/settings/common.py` does not include a `KEY_PREFIX` for any of the Redis-backed cache entries (`environment_cache`, `environment_document_cache`, etc.). Django's `BaseCache.make_key` produces keys in the format `{KEY_PREFIX}:{VERSION}:{key}` — with an empty prefix, all keys start with `:1:`.

## Desired behavior

A new environment variable (e.g., `CACHE_KEY_PREFIX`) that sets `KEY_PREFIX` on all Redis-backed cache entries in the `CACHES` dict. Example:

```python
CACHE_KEY_PREFIX = env.str("CACHE_KEY_PREFIX", default="")

for cache_name, cache_config in CACHES.items():
    if "redis" in cache_config.get("BACKEND", "").lower():
        cache_config["KEY_PREFIX"] = CACHE_KEY_PREFIX
```

With `CACHE_KEY_PREFIX=flagsmith`, keys become `flagsmith:1:environment_document_xxx`, which can be matched by Redis ACL `~flagsmith:*`.

## Use case

Our infrastructure runs a single Redis instance with per-service ACL users. Each service gets a dedicated user restricted to its own key namespace (`~<service>:*`). This is a common pattern for shared Redis deployments where full isolation (separate Redis instances) is impractical.

Without a configurable prefix, the Flagsmith Redis user must be granted `~*` (all keys), breaking the isolation model.

## Additional context

- This becomes more important as #6815 migrates more in-process caches to Redis
- Django's `CACHES` framework natively supports `KEY_PREFIX` — this is just about exposing it via env var
- Flagsmith's task processor already uses Postgres (not Redis), so this only affects cache keys

SahilJat:

- [ x] I have filled in the "Changes" section below.
- [ x ]  I have filled in the "How did you test this code" section below.

## Changes

Closes #7297 

Adds a `CACHE_KEY_PREFIX` environment variable that sets `KEY_PREFIX` on all Redis-backed cache entries in the `CACHES` dict.

- Added `CACHE_KEY_PREFIX = env.str("CACHE_KEY_PREFIX", default="")` to `api/app/settings/common.py`
- Added a loop after all `CACHES` entries are defined (including SAML) that applies `KEY_PREFIX = CACHE_KEY_PREFIX` to any cache whose `BACKEND` contains `"redis"` (case-insensitive)
- Non-Redis backends (locmem, db, dummy) are unaffected
- Default is empty string which preserves existing Django behaviour — no breaking change for existing deployments

With `CACHE_KEY_PREFIX=flagsmith`, Redis keys become `flagsmith:1:environment_document_xxx` instead of `:1:environment_document_xxx`, enabling Redis ACL rules like `~flagsmith:*` for namespace-based multi-tenancy.

- Added two unit tests in `api/tests/unit/app/test_unit_app_settings.py`:
  - Verifies Redis-backed caches get `KEY_PREFIX` applied
  - Verifies non-Redis caches (locmem, db) are not affected
  - Verifies empty prefix (default) sets `KEY_PREFIX` to empty string
- Verified settings load correctly with `CACHE_KEY_PREFIX=flagsmith python -c "from app.settings.common import CACHES"`
- All existing tests in `api/tests/unit/app/` pass

### DIFF
--- a/api/app/settings/common.py
+++ b/api/app/settings/common.py
@@ -648,7 +648,7 @@ if ENABLE_DB_LOGGING:
 CACHE_FLAGS_SECONDS = env.int("CACHE_FLAGS_SECONDS", default=0)
 FLAGS_CACHE_LOCATION = "environment-flags"
 CHARGEBEE_CACHE_LOCATION = "chargebee-objects"
-
+CACHE_KEY_PREFIX = env.str("CACHE_KEY_PREFIX", default="")
 ENVIRONMENT_CACHE_SECONDS = env.int("ENVIRONMENT_CACHE_SECONDS", default=60)
 ENVIRONMENT_CACHE_BACKEND = env.str(
     "ENVIRONMENT_CACHE_BACKEND",
@@ -1037,7 +1037,11 @@ if SAML_INSTALLED:
         subcast=str,
         default=[],
     )
-
+# Apply key prefix to all Redis-backed caches to support
+# ACL-based multi-tenancy on shared Redis instances.
+for _cache_config in CACHES.values():
+    if "redis" in _cache_config.get("BACKEND", "").lower():  # type: ignore[attr-defined]
+        _cache_config["KEY_PREFIX"] = CACHE_KEY_PREFIX  # type: ignore[index]
 
 WORKFLOWS_LOGIC_INSTALLED = importlib.util.find_spec("workflows_logic") is not None
 
--- a/api/tests/unit/app/test_unit_app_settings.py
+++ b/api/tests/unit/app/test_unit_app_settings.py
@@ -1,0 +1,51 @@
+import pytest
+
+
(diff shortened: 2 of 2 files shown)


SahilJat:
Hii alll the test cases has been passed , the test that failed is already exisiting issue mentioned here #7368

PR #7530, opened 2026-05-18
test: Fix `test_unit_routers.py`

### LINKED ISSUE #7368: API Tests: Apps registry pollution from test_unit_app_routers.py breaks migration tests

SahilJat:

## Changes

Closes #7368 

Fixes flaky CI failures caused by inline `models.Model` subclasses in `test_unit_app_routers.py` polluting Django's apps registry.

- Removed all inline `models.Model` subclasses (`AnalyticsModel`, `MyModel`, `MyModel1`, `MyModel2`) from the parametrised test functions.
- Replaced them with `mock.MagicMock()` objects that expose `_meta.app_label` — the only attribute the router inspects.
- Removed the `from django.db import models` import, replaced with `from unittest import mock`.
- The `allow_migrate` test was already clean and required no changes.

**Root Cause:**
The root cause was that defining a Django model class registers it permanently in the apps registry for the lifetime of the worker process. When tests ran with `app_label="app_analytics"`, the four model classes were permanently registered against the real `app_analytics` app, causing `makemigrations --check` to find unmigrated models and Migrator teardown to fail with `relation "app_analytics_analyticsmodel" already exists` across multiple xdist workers.

- Ran the updated test file locally and confirmed all 4 tests pass.
- Confirmed the `makemigrations` test no longer detects phantom models.
- Confirmed migration tests no longer fail with `DuplicateTable` errors.

### DIFF
--- a/api/tests/unit/app/test_unit_app_routers.py
+++ b/api/tests/unit/app/test_unit_app_routers.py
@@ -1,7 +1,9 @@
-import pytest
 from unittest import mock
+
+import pytest
 from django.db import models
 from django.db.models.options import Options
+
 from app import routers
 
 
--- a/api/tests/unit/app/test_unit_app_routers.py
+++ b/api/tests/unit/app/test_unit_app_routers.py
@@ -1,6 +1,7 @@
 import pytest
+from unittest import mock
 from django.db import models
-
+from django.db.models.options import Options
 from app import routers
 
 
@@ -16,14 +17,13 @@ def test_analytics_router_db_for_read__given_app_label__returns_expected_db(
     expected_db: str | None,
 ) -> None:
     # Given
-    class AnalyticsModel(models.Model):
-        class Meta:
-            app_label = given_app_label
-
+    mock_model = mock.MagicMock(spec=models.Model)
+    mock_model._meta = mock.MagicMock(spec=Options)
+    mock_model._meta.app_label = given_app_label
     router = routers.AnalyticsRouter()
 
     # When
-    db = router.db_for_read(AnalyticsModel)
+    db = router.db_for_read(mock_model)
 
     # Then
     assert db == expected_db
@@ -41,14 +41,13 @@ def test_analytics_router_db_for_write__given_app_label__returns_expected_db(
     expected_db: str | None,
 ) -> None:
     # Given
-    class MyModel(models.Model):
-        class Meta:
-            app_label = model_app_label
-
+    mock_model = mock.MagicMock(spec=models.Model)
+    mock_model._meta = mock.MagicMock(spec=Options)
+    mock_model._meta.app_label = model_app_label
     router = routers.AnalyticsRouter()
 
     # When
-    db = router.db_for_write(MyModel)
+    db = router.db_for_write(mock_model)
 
     # Then
     assert db == expected_db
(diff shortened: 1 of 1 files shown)


khvn26 (changes requested):
The approach looks sensible to me — can you add `spec=Model` to the mocks you've added?

SahilJat:
Hii @khvn26 i made the changes and updated the test. Thanks

khvn26:
@SahilJat Looks like you need to explicitly mock the `_meta` attribute with `spec=Options` (see [here](https://docs.djangoproject.com/en/6.0/ref/models/meta/#django.db.models.options.Options))

SahilJat:
Thanks for the hint @khvn26 , did mock the _meta explicitly in this change. Please have a look and let me know if its gtg

khvn26:
@SahilJat sorry, tiny ask — could you add `spec=Options` to the new mocks? Thanks

SahilJat:
@khvn26 no problem, added the spec=Options this time.

khvn26 (approved):

khvn26:
Thanks for the contribution!

PR #7535, opened 2026-05-19
fix: Enhance EFV audit logs

### LINKED ISSUE #7526: Audit: enrich EnvironmentFeatureVersion-published log with changed feature states
In v2 versioning environments, every flag change is published via an `EnvironmentFeatureVersion`, and the resulting audit row uses a fixed template (`api/audit/constants.py:64`):

```
ENVIRONMENT_FEATURE_VERSION_PUBLISHED_MESSAGE = "New version published for feature: %s"
```

That string is the entire `log` field. Nothing indicates which feature states changed, the enabled/disabled flip, segment overrides touched, or value diffs.

Five integrations consume this row via the AuditLog post_save filter (`api/audit/signals.py:118-125`, accepting `FEATURE`, `FEATURE_STATE`, `SEGMENT`, `EF_VERSION`) and render the `log` text:

- Datadog, New Relic, Slack, Grafana — render `audit_log.log` directly
- Dynatrace — same, plus `audited_instance.feature.name` for deployment name

v1 produces per-FS rows like `Flag state updated: my_flag: Enabled`. v2 collapses these into one row per version with no per-FS content.

Proposal:

- Render the EFV message via `get_updated_feature_states_for_version(efv)`, enumerating each FS's action.
- Optionally expose the change list as structured audit-log metadata so future receivers don't re-parse prose.

SahilJat:

## Changes

Closes #7526 

This PR resolves the issue of opaque audit logs for v2 versioning (`EnvironmentFeatureVersion`). Previously, publishing a new version only logged a static string ("New version published for feature: %s"), completely omitting which feature states were actually changed.

This updates `create_environment_feature_version_published_audit_log_task` to:
- Render the EFV message via `get_updated_feature_states_for_version(efv)`.
- Enumerate each changed Feature State's action (enabled/disabled) and scope (Environment default, Segment override, Identity override).
- Mitigate N+1 database queries by prefetching `feature_segment__segment` and `identity` via `select_related`.

-I wrote comprehensive unit tests in `api/tests/unit/audit/test_unit_audit_services.py` to cover the different scenarios for the audit log task:
- Verified that if there are no changes, the audit log falls back to the original header-only format.
- Verified that modifications to the Environment Default are properly appended to the log string.
- Verified that modifications to a Segment Override append the specific Segment's name to the log string.

### DIFF
--- a/api/features/versioning/tasks.py
+++ b/api/features/versioning/tasks.py
@@ -286,9 +286,7 @@ def create_environment_feature_version_published_audit_log_task(
         % environment_feature_version.feature.name
     )
 
-    changed_states = get_updated_feature_states_for_version(
-        environment_feature_version
-    )
+    changed_states = get_updated_feature_states_for_version(environment_feature_version)
 
     if changed_states:
         changed_states = list(
@@ -297,8 +295,7 @@ def create_environment_feature_version_published_audit_log_task(
             ).select_related("feature_segment__segment", "identity")
         )
         change_lines = "\n".join(
-            f"- {_build_feature_state_change_summary(fs)}"
-            for fs in changed_states
+            f"- {_build_feature_state_change_summary(fs)}" for fs in changed_states
         )
         log = f"{header}\n{change_lines}"
     else:
--- a/api/features/versioning/tasks.py
+++ b/api/features/versioning/tasks.py
@@ -261,6 +261,18 @@ def trigger_update_version_webhooks(environment_feature_version_uuid: str) -> No
     )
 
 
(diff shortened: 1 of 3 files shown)


SahilJat:
@khvn26 the test cases that is failing is already existing issue which will be solved with pr #7530 . This issue is also solved please have a look at it. Thanks

Zaimwa9:
@claude review-once

SahilJat:
For the deleted states — I kept it to just the current version's feature states since the issue didn't explicitly mention deletions, but you're right that a removed segment override is still a change and should show up in the log.
Will fix both in the next commit.

PR #7548, opened 2026-05-20
docs: clarify change request scope for identity overrides

### LINKED ISSUE #7506: docs: Clarify which flag operations require change requests vs which bypass
## Current state

The [change requests docs page](https://docs.flagsmith.com/advanced-use/change-requests) generically says *"attempting to change a flag value will prompt you to create a new Change Request"*. The page does not distinguish between feature-state types and which trigger the CR workflow.

## What the code does (verified in `api/features/models.py`)

FeatureState has three types determined by `identity_id` and `feature_segment_id`:

```python
@property
def type(self) -> str:
    if self.identity_id and self.feature_segment_id is None:
        return IDENTITY
    elif self.feature_segment_id and self.identity_id is None:
        return FEATURE_SEGMENT
    elif self.identity_id is None and self.feature_segment_id is None:
        return ENVIRONMENT
```

The `is_live` property:

```python
@property
def is_live(self) -> bool:
    if self.environment.use_v2_feature_versioning:
        if self.identity_id is not None:
            return True
        return (
            self.environment_feature_version_id is not None
            and self.environment_feature_version.is_live
        )
```

Identity-level feature states are live immediately. CRs are tied to versioning (`environment_feature_version`). Identity-level FS does not use versioning, so it is not gated by the CR workflow.

## Implication for customers

- **Environment defaults**: require CR when enabled
- **Segment overrides**: require CR when enabled
- **Identity-level overrides**: bypass CR; live immediately

Customers ask "why didn't this require approval?" when they don't see this distinction. The docs should make the scope explicit so the behavior is predictable.

## Suggested docs addition

A small section under change requests describing which feature-state types the workflow applies to, and noting that identity-level overrides are live immediately and intentionally bypass the workflow (consistent with the use case of developer/QA targeting).

SahilJat:

## Changes

Closes #7506 

_Please describe._
Added a "Scope of Change Requests" section explaining that identity 
overrides intentionally bypass the workflow, while environment and 
segment overrides require approval.

not required.

### DIFF
--- a/docs/docs/administration-and-security/governance-and-compliance/change-requests.md
+++ b/docs/docs/administration-and-security/governance-and-compliance/change-requests.md
@@ -6,7 +6,7 @@ sidebar_position: 10
 
 You can use Change Requests to add workflow control to the changing of flag values. Change Requests allow a user to propose a change to a flag value, and then require that change is approved by a number of other team members.
 
-You can use Change Requests to ensure that, for example, a change to a flag in Production has to be approved by another team member. They work in a similar way to Pull Requests in git. This guide explains how to use Change Requests to add workflow control to changing flag values. You will learn how to set up, create, approve, and publish Change Requests. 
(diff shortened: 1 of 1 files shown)


Zaimwa9:
Closing this one as superseeded by #7549. Let's get the other one out of the door, thanks for the contribution

PR #7549, opened 2026-05-20
docs: document CR scope, notifications and non-notification events

### LINKED ISSUE #7505: docs: Document email notifications fired by the change request workflow
## Current state

The [change requests docs page](https://docs.flagsmith.com/advanced-use/change-requests) mentions email notifications only once: *"These individuals will receive an email notification about the Change Request."* This refers to assignees being notified at CR creation. The page does not document the full set of email notifications that the workflow actually fires.

## What the code does (verified in `api/features/workflows/core/models.py`)

Two email notifications fire for individual approvers:

1. **`send_email_notification_to_assignee`** — fires on `AFTER_CREATE` of a `ChangeRequestApproval` with `approved_at=None`. Notifies the assignee that they have a pending approval.
2. **`send_email_notification_to_author`** — fires on `AFTER_CREATE` or `AFTER_UPDATE` when `approved_at` transitions to a non-null value. Notifies the CR author that their CR has been approved.

A group-level notification also fires via background task `notify_group_of_change_request_assignment.delay(...)` when a CR is assigned to a group.

## Gaps that should be made explicit in docs

- Both individual email notifications above should be documented (pending-approval to assignee, approval-granted to author).
- Group assignment notification path should be documented.
- Customers consistently ask whether external notifications exist; the docs implying "no" is causing avoidable workaround conversations.

## What the code does NOT do (worth calling out so customers don't expect it)

- No email is sent when a CR is committed (only audit log entries are created for affected feature states).
- There is no first-class CR rejection; "rejection" = delete, no email.
- No webhook events for CR lifecycle changes were found in the codebase.

## Suggested docs additions

A "Notifications" subsection under the change requests page enumerating the two emails + group notification that fire, plus a brief note on what is not notified externally.

SahilJat:

## Changes

Closes #7505 
The Change Requests docs page had two gaps that were causing avoidable
support conversations:

1. **Scope was not documented** — the page said "attempting to change a flag
   value will prompt you to create a new Change Request" without clarifying
   that identity-level overrides are intentionally excluded from the workflow
   and go live immediately.

2. **Notifications were not documented** — the page mentioned assignee emails
   only once in passing. The full set of emails that fire (pending-approval to
   assignee, approval-granted to author, group assignment notification) was
   not described anywhere. Equally important, what does _not_ trigger a
   notification (commit, deletion, webhooks) was also undocumented, leading
   customers to expect behaviour that does not exist.

**Changes made **

- Added a **Scope of Change Requests** section with a table showing which
  feature state types require a CR and which do not, with an explanation of
  why identity overrides are excluded (they do not participate in the
  versioning system the CR workflow is built on)
- Added a tip recommending segment-based targeting for teams that need
  approval gates on all changes
- Added a **Notifications** section with a table of all email notifications
  that fire during the CR lifecycle, and an explicit list of events that do
  not trigger notifications (commit, deletion, webhooks)
- Updated the "Create a Change Request" section to clarify that the prompt
  applies to environment defaults and segment overrides, not all flag changes

Both additions are verified against the actual implementation in
`api/features/workflows/core/models.py`.

This is a documentation-only change — no code was modified.

Verified the described behaviour against the source code:

- **Scope** — confirmed via `FeatureState.type` property and `is_live`
  property in `api/features/models.py` that identity overrides bypass
  versioning and are live immediately
- **Notifications** — confirmed via `send_email_notification_to_assignee`,
  `send_email_notification_to_author`, and
  `notify_group_of_change_request_assignment` hooks in
  `api/features/workflows/core/models.py` that the documented emails are the
  complete set
- **Non-notifications** — confirmed no email fires on commit, no rejection
  flow exists, and no CR lifecycle webhook events are present in the codebase

### DIFF
--- a/docs/docs/administration-and-security/governance-and-compliance/change-requests.md
+++ b/docs/docs/administration-and-security/governance-and-compliance/change-requests.md
@@ -4,8 +4,6 @@ sidebar_label: Change Requests
 sidebar_position: 10
 ---
 
-You can use Change Requests to add workflow control to the changing of flag values. Change Requests allow a user to propose a change to a flag value, and then require that change is approved by a number of other team members.
-
 You can use Change Requests to ensure that, for example, a change to a flag in Production has to be approved by another team member. They work in a similar way to Pull Requests in git. This guide explains how to use Change Requests to add workflow control to changing flag values. You will learn how to set up, create, approve, and publish Change Requests.
 
 ## Prerequisites
@@ -79,6 +77,10 @@ Request lifecycle:
 | A Change Request is assigned to an individual | The assignee receives an email informing them that their approval is pending |
 | A Change Request is approved | The CR author receives an email confirming the approval |
 | A Change Request is assigned to a group | All members of the group receive a notification via background task |
+
+:::note
+**Self-hosting?** To ensure these email notifications are delivered correctly, you must configure the [self-hosted email setup](https://docs.flagsmith.com/deployment-self-hosting/core-configuration/email-setup).
+:::
 
 **What does not trigger a notification:**
 
--- a/docs/docs/administration-and-security/governance-and-compliance/change-requests.md
+++ b/docs/docs/administration-and-security/governance-and-compliance/change-requests.md
@@ -6,7 +6,7 @@ sidebar_position: 10
 
 You can use Change Requests to add workflow control to the changing of flag values. Change Requests allow a user to propose a change to a flag value, and then require that change is approved by a number of other team members.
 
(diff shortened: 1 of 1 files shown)


Zaimwa9 (dismissed):

Holmus (dismissed):
Nice contribution, thanks for that!

Approved from me as well but if you could add a reference to the self-hosted email setup in the email section that would be even better!

(https://docs.flagsmith.com/deployment-self-hosting/core-configuration/email-setup)

SahilJat:
Hi @Holmus , added the reference to the email docs for self hosting , please have a look and there was a line repeating itself , also removed that as well , and if theres something else you need me to do  ,please let me know, Thanks

Holmus (approved):
Good catch on the duplicate, thanks for the modification as well!

LGTM 🚀

PR #7463, opened 2026-05-08
fix(Code References): Code references are slow to query

emyller:
Thanks for submitting a PR! Please check the boxes below:

- [x] I have read the [Contributing Guide](/Flagsmith/flagsmith/blob/main/CONTRIBUTING.md).
- [ ] ~I have added information to `docs/` if required so people know about the feature.~
- [x] I have filled in the "Changes" section below.
- [x] I have filled in the "How did you test this code" section below.

## Changes

> [!IMPORTANT]
> Contains a data migration.

Closes https://github.com/Flagsmith/flagsmith/issues/5932
Would originally close https://github.com/Flagsmith/flagsmith/issues/6832

Improve code references so they're usable again.

The previous data model did not handle accumulated code references too well, and we had to disable it.

This patch will:

- [x] Complete redesign the data model of code references.
- [x] Migrate old data to new data.
- [x] Make code references viable for GA: in an hypothetical project with 400 features, 200K code references scans, list annotation improved from ~21s to ~12ms (around 1700x increase), and detail query improved from ~77ms to 1.4ms (~55x). Gotta go fast.
- [x] Prevent duplicate scans — for hygiene: performance gain does **not** rely on this.
- [x] Add a new "vcs_provider" to the API — but defaults to "github" for compatibility. This will allow support other VCS endeavours for free.
- [x] Delete `code_references_ui_stats` feature flag branches, :tada:.

## How did you test this code?

- [x] Unit tests.
- [x] Benchmarking — tested on Apple silicon.

PR #7555, opened 2026-05-20
fix(Code References): Very slow data migration

emyller:

## Changes

Follow up to https://github.com/Flagsmith/flagsmith/pull/7463

While the original migration takes <30s on local with same data, it blew up in production. This is an attempt to unblock this migration by making it faster.

Local tests using a code references data dump from production.


khvn26 (dismissed):
The approach looks right to me 👍

khvn26 (approved):

PR #7600, opened 2026-05-26
fix(Segment membership): Zero out (segment, env) pairs that stopped matching

khvn26:

## Changes

Contributes to #5663.

`refresh_project_segment_counts` only upserts pairs returned by compute (count > 0), so a pair that previously matched but no longer does keeps its stale non-zero count forever. Append explicit zero rows for every `(segment, env)` pair already in the table that didn't appear in this run; pairs that have never matched stay absent.

Added `test_refresh_project_segment_counts__previously_matching_pair_drops_to_zero__row_zeroed` and `test_refresh_project_segment_counts__never_matched_pair__no_row_written`.

### DIFF
--- a/api/segment_membership/tasks.py
+++ b/api/segment_membership/tasks.py
@@ -168,8 +168,11 @@ def refresh_project_segment_counts(project_id: int) -> None:
         # so the stale count would linger. Append explicit zeros for those
         # missing pairs so the next refresh writes them down to zero.
         new_pairs = {(m.segment_id, m.environment_id) for m in membership_counts}
+        # `count__gt=0` keeps steady-state refreshes off rows that are already
+        # zeroed -- writing zero over zero on every run is just churn.
         existing_pairs = SegmentMembershipCount.objects.filter(
             segment__project=project,
+            count__gt=0,
         ).values_list("segment_id", "environment_id")
         zeroed_counts = [
             SegmentMembershipCount(
--- a/api/segment_membership/tasks.py
+++ b/api/segment_membership/tasks.py
@@ -162,8 +162,6 @@ def refresh_project_segment_counts(project_id: int) -> None:
         for m in membership_counts:
             m.last_synced_at = now
 
-        # Drop pairs that stopped matching this run so the next refresh
-        # doesn't carry stale counts forward.
         new_pairs = {(m.segment_id, m.environment_id) for m in membership_counts}
         stale_ids = [
             pk
@@ -174,11 +172,9 @@ def refresh_project_segment_counts(project_id: int) -> None:
             )
             if (segment_id, environment_id) not in new_pairs
         ]
-        stale_deleted, _ = (
-            SegmentMembershipCount.objects.filter(id__in=stale_ids).delete()
-            if stale_ids
-            else (0, {})
-        )
+        stale_deleted, _ = SegmentMembershipCount.objects.filter(
+            id__in=stale_ids,
+        ).delete()
 
         SegmentMembershipCount.objects.bulk_create(
             membership_counts,
--- a/api/segment_membership/tasks.py
+++ b/api/segment_membership/tasks.py
@@ -162,31 +162,26 @@ def refresh_project_segment_counts(project_id: int) -> None:
         for m in membership_counts:
             m.last_synced_at = now
 
-        # Pairs that previously matched but no longer do (rule edited, traits
-        # drifted, identities deleted) won't appear in `membership_counts`.
-        # `bulk_create(update_conflicts=True)` only touches rows it's handed,
-        # so the stale count would linger. Append explicit zeros for those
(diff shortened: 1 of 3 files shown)


api/segment_membership/tasks.py
```diff
@@ -161,8 +161,29 @@ def refresh_project_segment_counts(project_id: int) -> None:
...
+
+        # Pairs that previously matched but no longer do (rule edited, traits
+        # drifted, identities deleted) won't appear in `membership_counts`.
+        # `bulk_create(update_conflicts=True)` only touches rows it's handed,
+        # so the stale count would linger. Append explicit zeros for those
+        # missing pairs so the next refresh writes them down to zero.
+        new_pairs = {(m.segment_id, m.environment_id) for m in membership_counts}
+        existing_pairs = SegmentMembershipCount.objects.filter(
+            segment__project=project,
+        ).values_list("segment_id", "environment_id")
```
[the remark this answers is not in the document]
khvn26: Agree, good bot. b60fd51ea7b82e383a1b58ce2c3724c16dcbd766.

api/segment_membership/tasks.py
```diff
@@ -161,8 +161,32 @@ def refresh_project_segment_counts(project_id: int) -> None:
...
             m.last_synced_at = now
+
+        # Pairs that previously matched but no longer do (rule edited, traits
+        # drifted, identities deleted) won't appear in `membership_counts`.
+        # `bulk_create(update_conflicts=True)` only touches rows it's handed,
+        # so the stale count would linger. Append explicit zeros for those
+        # missing pairs so the next refresh writes them down to zero.
+        new_pairs = {(m.segment_id, m.environment_id) for m in membership_counts}
+        # `count__gt=0` keeps steady-state refreshes off rows that are already
+        # zeroed -- writing zero over zero on every run is just churn.
```
matthewelwell: Nit: this comment feels slop-y to me. It could just be: 

```
Ignore records that already have `count=0` to avoid unnecessary writes. 
```

api/segment_membership/tasks.py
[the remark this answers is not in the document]
matthewelwell: Sorry, probably lacking context, but is there a benefit to adding zero count records instead of doing something like: 

```python
SegmentMembershipCount.objects.exclude(...).delete()
```

Maybe we're planning on being able to show a history?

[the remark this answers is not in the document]
khvn26: Switched to delete in 8d259ad28 — same end state for the UI without keeping zero rows around.

api/segment_membership/tasks.py
khvn26: Replaced with delete in 8d259ad28, the slop comment is gone.

api/segment_membership/tasks.py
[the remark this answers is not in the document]
khvn26: IDK if we should invest in history given we'll move from polling to CDC fairly soon.

Zaimwa9 (dismissed):

api/segment_membership/tasks.py
```diff
@@ -161,6 +161,25 @@ def refresh_project_segment_counts(project_id: int) -> None:
         now = timezone.now()
         for m in membership_counts:
             m.last_synced_at = now
+
+        # Drop pairs that stopped matching this run so the next refresh
+        # doesn't carry stale counts forward.
```
matthewelwell: This feels slop-y again? I think it's referencing old context about 'carrying stale counts forward'

api/segment_membership/tasks.py
```diff
@@ -161,6 +161,25 @@ def refresh_project_segment_counts(project_id: int) -> None:
...
+                    segment__project=project
+                ).values_list("id", "segment_id", "environment_id")
+            )
+            if (segment_id, environment_id) not in new_pairs
+        ]
+        stale_deleted, _ = (
+            SegmentMembershipCount.objects.filter(id__in=stale_ids).delete()
+            if stale_ids
+            else (0, {})
+        )
```
matthewelwell: I suspect this would work the same, no? Surely `0, {}` is the default response from `.delete()` and we're already working with a list, not `None`. 

```suggestion
        stale_deleted, _ = SegmentMembershipCount.objects.filter(id__in=stale_ids).delete()
```

api/segment_membership/tasks.py
khvn26: Dropped the comment in 5da46f1f5 — the code is self-evident now.

api/segment_membership/tasks.py
khvn26: Right — `.delete()` on an empty `id__in=[]` filter already returns `(0, {})` without issuing SQL. Adopted the suggestion in 5da46f1f5.

matthewelwell (approved):

PR #7605, opened 2026-05-27
fix(experimentation): always insert new WarehouseConnection on create

gagantrivedi:

## Changes

`WarehouseConnectionSerializer.create()` looked up a soft-deleted row for the same `(environment, warehouse_type)` and resurrected it by clearing `deleted_at` and calling `save()`. Because that path is an `UPDATE`, django-lifecycle's `@hook(AFTER_CREATE)` on `WarehouseConnection` never fired, and the environment key was never re-added to the ingestion Redis after a customer deleted and re-created a connection.

This PR drops the resurrect branch so create always performs an `INSERT`. The partial unique constraint `unique_active_warehouse_per_type_and_env` (conditioned on `deleted_at IS NULL`) already permits a fresh insert once the previous row is soft-deleted, and the "one active connection per environment" rule is still enforced by the 409 check in `WarehouseConnectionViewSet.create()`.

The two existing resurrect tests are rewritten to assert that a new PK is returned instead.

- `pytest tests/unit/experimentation/` — all 46 tests pass.
- `mypy` clean on the touched files.

### DIFF
--- a/api/experimentation/serializers.py
+++ b/api/experimentation/serializers.py
@@ -5,7 +5,6 @@ from rest_framework import serializers
 from environments.models import Environment
 from experimentation.models import (
     WarehouseConnection,
-    WarehouseConnectionStatus,
     WarehouseType,
 )
 from experimentation.types import SNOWFLAKE_DEFAULTS, SnowflakeConfig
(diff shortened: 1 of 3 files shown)


Zaimwa9 (approved):
Thanks for this one

PR #7615, opened 2026-05-27
feat: added pagination and deletion guard

Zaimwa9:

## Changes
- Wrap experiment mutations and status transitions in `transaction.atomic`, and return `409` (instead of 500) on concurrent-create
  races.
- Validate `?status=` against `ExperimentStatus`, block `DELETE` on running experiments, and skip the audit log on no-op `PATCH`.
- Paginate the list endpoint with `CustomPagination` and use `_get_environment()` per request.
- Prevent deleting a running experiment

- new tests

### DIFF
--- a/api/experimentation/services.py
+++ b/api/experimentation/services.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import typing
 
-from django.db import transaction
 from django.utils import timezone
 
 from audit.models import AuditLog
@@ -98,7 +97,6 @@ def transition_experiment_status(
     elif target_status == ExperimentStatus.COMPLETED:
         experiment.ended_at = timezone.now()
 
-    with transaction.atomic():
-        experiment.save()
-        create_experiment_audit_log(experiment, user, action=target_status)
+    experiment.save()
+    create_experiment_audit_log(experiment, user, action=target_status)
     return experiment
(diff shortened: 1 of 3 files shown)


api/experimentation/services.py
```diff
@@ -97,6 +98,7 @@ def transition_experiment_status(
     elif target_status == ExperimentStatus.COMPLETED:
         experiment.ended_at = timezone.now()
 
-    experiment.save()
-    create_experiment_audit_log(experiment, user, action=target_status)
+    with transaction.atomic():
+        experiment.save()
+        create_experiment_audit_log(experiment, user, action=target_status)
     return experiment
```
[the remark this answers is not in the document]
Zaimwa9: Good one actually, i'll change this

api/experimentation/views.py
```diff
@@ -147,12 +165,27 @@ def perform_create(self, serializer: BaseSerializer[Experiment]) -> None:
         )
 
     def perform_update(self, serializer: BaseSerializer[Experiment]) -> None:
+        has_changes = any(
```
gagantrivedi: Nit: this feels a bit like overkill

gagantrivedi (dismissed):

api/experimentation/views.py
Zaimwa9: Ah yes thanks, I was sure i reverted to sequential 🤔

gagantrivedi (approved):

PR #7635, opened 2026-05-28
fix: detect multivariate differences between environments

### LINKED ISSUE #7619: Incorrect comparison for multivariate flags across environments

- [ ] Self Hosted with Docker
- [x] SaaS at flagsmith.com

### Describe the bug

A feature flag which has different multivariate options between two environments does not show up as a "changed" flag in the compare view. MV feature with different percentages always shows up as unchanged.

Image

Image

Image

### Steps To Reproduce

Create a flag.
Set the A/B variants in environment X to one thing
Set the A/B variants in environment Y to another thing (more variants, different values, different percentage split)
Compare environment X to environment Y in the compare view

### Expected behavior

The comparison should notice that the flag has different percentages in the different environments and highlight this as a change between the environments.

### Screenshots

_No response_

10done:

## Changes

Fixes #7619 

- Adds `hasMultivariateChange()` to compare `multivariate_feature_option` + `percentage_allocation` per environment
- Uses it in `CompareEnvironments` for changed/unchanged classification and row fade behavior
- Uses it in `CompareFeatures` so MV-only differences are not shown as identical
- Adds unit tests in `compare-multivariate-utils.test.ts`

I have tested the code manually and added screenshots below and also added unit tests for the same.
### Before
Screenshot 2026-05-28 at 8 12 34 PM
Screenshot 2026-05-28 at 8 12 53 PM
image

No flag differences / rollout_test in Unchanged (Dev 0% vs Prod 100%).

### After
image

rollout_test under Changed Flags.


frontend/common/utils/compareMultivariate.ts
```diff
@@ -0,0 +1,40 @@
+import { FeatureState } from 'common/types/responses'
+
+const normaliseMultivariate = (featureState?: FeatureState) => {
+  const values = featureState?.multivariate_feature_state_values || []
+  return values
+    .map((v) => ({
+      multivariate_feature_option: v.multivariate_feature_option,
+      percentage_allocation: Number(v.percentage_allocation) || 0,
+    }))
```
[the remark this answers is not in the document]
10done: I have added this suggestion for more robust sorting

10done:
@Holmus Can I have review please whenever you have time. happy to make any further changes required.

10done:
@matthewelwell Can I have a review please or is this low priority?

frontend/common/utils/compareMultivariate.ts
```diff
@@ -0,0 +1,40 @@
+import { FeatureState } from 'common/types/responses'
```
Zaimwa9, line 1: Can we move this file in `common/utils/`, next to `common/utils/multivariate.ts` please ?

frontend/common/utils/compareMultivariate.ts
```diff
@@ -0,0 +1,40 @@
...
+    }))
+    .sort((a, b) => {
+      if (a.multivariate_feature_option !== b.multivariate_feature_option) {
+        return a.multivariate_feature_option - b.multivariate_feature_option
+      }
+      return a.percentage_allocation - b.percentage_allocation
+    })
+}
+
+export const hasMultivariateChange = (
```
Zaimwa9, line 18: I believe we already have a similar `getVariationDiff` that could be reused (in `web/components/diff/diff-utils.ts`), we should be able to call it as follows `getVariationDiff(left, right).totalChanges > 0`

Zaimwa9 (changes requested):
Thanks for the contribution, this will help a lot. I just have a couple of clean-housing comments and it should be good to go after that

frontend/common/utils/compareMultivariate.ts
10done: Thank you for the review.
I have made the changes.

frontend/common/utils/compareMultivariate.ts
10done: @Zaimwa9 I have made the changes. Thank You.

Zaimwa9 (approved):
Thanks a lot for the contribution!

PR #7636, opened 2026-05-28
fix: return 400 instead of 500 when request user is invalid (closes #6809)

### LINKED ISSUE #6809: URL/query/data input lacking validation
Although this Sentry issue ([FLAGSMITH-API-5J6](https://flagsmith.sentry.io/issues/7298473125/?referrer=github_integration)
) is isolated to this one case, we have history of similar errors showing up in Sentry.

The goal of this issue is to:
1. Find occurrences in the code that skip input validation — similar to this — and fix them with a proper HTTP 400 response.
2. Ultimately, reduce noise — again, similar to this.

```
ValueError: invalid literal for int() with base 10: 'PROJECT_ID'
  File "django/db/models/fields/__init__.py", line 2128, in get_prep_value
    return int(value)

ValueError: Field 'id' expected a number but got 'PROJECT_ID'.
(19 additional frame(s) were not displayed)
...
  File "features/permissions.py", line 57, in has_permission
    project = Project.objects.get(id=project_id)
```

botbikamordehai2-sketch:
## What
When a request has an unexpected user type (e.g., anonymous, or a user model that is neither FFAdminUser nor an API key), `AuthorData.from_request` raises a bare `ValueError`. This propagates to database operations where unvalidated input (like a string 'PROJECT_ID' for an integer field) causes a similar `ValueError`, resulting in a 500 Internal Server Error.

## Fix
Replaced the generic `ValueError` with Django REST Framework's `ValidationError`, which automatically results in an HTTP 400 Bad Request response. This aligns with the project's goal of validating input early and returning proper error responses.

Closes #6809

### DIFF
--- a/api/core/dataclasses.py
+++ b/api/core/dataclasses.py
@@ -16,10 +16,11 @@ class AuthorData:
     @classmethod
     def from_request(cls, request: "Request") -> "AuthorData":
         from users.models import FFAdminUser
+        from rest_framework.exceptions import ValidationError
 
         if type(request.user) is FFAdminUser:
             return cls(user=request.user)
         elif hasattr(request.user, "key"):
             return cls(api_key=request.user.key)
         else:
-            raise ValueError("Request user must be FFAdminUser or have an API key")
+            raise ValidationError("Request user must be FFAdminUser or have an API key")
--- a/api/core/dataclasses.py
+++ b/api/core/dataclasses.py
@@ -15,8 +15,9 @@ class AuthorData:
 
     @classmethod
     def from_request(cls, request: "Request") -> "AuthorData":
-        from users.models import FFAdminUser
         from rest_framework.exceptions import ValidationError
+
+        from users.models import FFAdminUser
 
         if type(request.user) is FFAdminUser:
             return cls(user=request.user)


api/core/dataclasses.py
```diff
@@ -15,11 +15,13 @@ class AuthorData:
 
     @classmethod
     def from_request(cls, request: "Request") -> "AuthorData":
+        from rest_framework.exceptions import ValidationError
```
khvn26, line 18: Any particular reason to use a local import here?

api/core/dataclasses.py
```diff
@@ -15,11 +15,13 @@ class AuthorData:
...
+
         from users.models import FFAdminUser
 
         if type(request.user) is FFAdminUser:
             return cls(user=request.user)
         elif hasattr(request.user, "key"):
             return cls(api_key=request.user.key)
         else:
-            raise ValueError("Request user must be FFAdminUser or have an API key")
+            raise ValidationError("Request user must be FFAdminUser or have an API key")
```
khvn26, line 27: Seems like there's no test coverage for this line?

Also, the exception message doesn't make sense when presented in an API response.

khvn26 (changes requested):

PR #7653, opened 2026-06-01
docs: Update docs custom ssr to use new instance instead of singleton

loic-masson:

- [X] I have added information to `docs/` if required so people know about the feature.

## Changes

Updating the documentation for custom SSR as I think it is a bit misleading. 
I thought that the flagsmith init would be straight from flagsmith/isomorphic.
Which created issues in my SSR setup.

The documentation is showing to use createFlagsmithInstance which also reflects how the Next js is setup.

Just documentation changes.


Zaimwa9 (approved):
Thanks for your contribution, much appreciated!

PR #7656, opened 2026-06-01
feat(MCP): Use native OpenAPI tool fields and consolidate private deps

khvn26:
Thanks for submitting a PR! Please check the boxes below:

- [x] I have read the [Contributing Guide](/Flagsmith/flagsmith/blob/main/CONTRIBUTING.md).
- [ ] I have added information to `docs/` if required so people know about the feature.
- [x] I have filled in the "Changes" section below.
- [x] I have filled in the "How did you test this code" section below.

## Changes

Contributes to Flagsmith/flagsmith-private#145 and the self-hosted MCP v3 epic (Flagsmith/flagsmith-private#127).

- Relocate the 36 MCP-tagged views' friendly names and descriptions from the non-standard `x-gram` extension to native `operation_id` and `description`. Gram falls back to `operationId`/`description` when `x-gram` is absent, so the existing Gram push keeps working unchanged; this also prepares for the self-hosted MCP v3 server (Flagsmith/flagsmith-private#146), which reads the same native fields from the live `/swagger`. Descriptions are phrased API-agnostically ("Applies to …" rather than "Use this tool …") since they now also feed the public API docs, and the organisation descriptions are Britishised. Tool names are unchanged.
- Consolidate the private apps: `rbac` and `workflows_logic` now ship inside `flagsmith-private` (>=0.7.0), so drop the separate `workflows-logic` dependency, the `flagsmith-rbac` clone in `install-private-modules`, and the `--extra workflows` installs. `integrate-private-tests` pulls those suites from `flagsmith-private` at the installed version's tag.
- Switch pytest to `--import-mode=importlib` (with `pythonpath = ["."]`) so the imported private test suites resolve without per-directory package markers, matching flagsmith-private's own pytest config. (`from tests.*` importers keep working via `pythonpath`.)
- Bump uv 0.11.8 -> 0.11.14.

The Gram schema generator and deploy push are deliberately left intact — Gram stays live until the v3 server lands.

## How did you test this code?

- Full core suite green locally under the new import mode: `3780 passed, 7 skipped`.
- `make typecheck` (1690 files) and `make lint` pass.
- Installed `flagsmith-private==0.7.0` from the index and ran `make generate-mcp-spec`: 43 `mcp`-tagged operations, no `x-gram`, every op with native `operationId` + `description`, including the private tools (org roles, release pipelines, change requests).
- Verified the imported `flagsmith-private` test suites (rbac, workflows) collect cleanly under importlib, including duplicate test-module basenames across `unit/` and `integration/`.

PR #7665, opened 2026-06-01
ci: Run flagsmith-private test suites inline on 0.8.0

khvn26:

## Changes

Contributes to Flagsmith/flagsmith-private#177.

Follows on from #7656, which consolidated `rbac` and `workflows_logic` into `flagsmith-private` and runs their suites inline via `make test`. This extends the same pattern to `flagsmith-private` 0.8.0, which also bundles `auth_controller`, so the private suites run inline against Core API again.

- Bump `flagsmith-private` to `>=0.8.0`. It now ships `auth_controller`, so drop the `auth-controller` extra and its git source; `integrate-private-tests` no longer clones `flagsmith-auth-controller` separately.
- `integrate-private-tests` pulls each app's tests (`rbac`, `workflows_logic`, `auth_controller`, `saml` unit + integration) from `flagsmith-private` at the installed wheel's tag and drops them under `tests/` so they inherit Core API's `tests/conftest` and run inline with `make test`.
- `.gitignore` the cloned repo and the staged `tests/*_tests` directories.

`scim` and `release_pipelines_logic` are deliberately left out for now and continue to run in flagsmith-private's own CI:

- `scim`: its compliance tests require an `SCIM_SERVICE_PROVIDER` config (notably dropping the request-based `BASE_LOCATION_GETTER`) that directly conflicts with Core API's own `tests/integration/scim` tests, and django-scim2 reads that setting once at import — so both suites cannot share one pytest session.
- `release_pipelines_logic`: one test asserts a non-deterministic email ordering that only holds in isolation; it fails when collected alongside the full Core API suite.

Both gaps are tracked in Flagsmith/flagsmith-private#177.

Locally, against the Compose Postgres, with `--extra dev --extra private` installed:

1. `make integrate-private-tests` to stage the suites.
2. Ran the staged suites under Core API's pytest config (`--ds=app.settings.test`, `--import-mode=importlib`):

   ```
   uv run pytest tests/rbac_tests tests/workflow_tests tests/auth_controller_tests \
     tests/saml_integration_tests tests/saml_unit_tests
   ```

   Result: 343 passed.

### DIFF
--- a/.github/workflows/api-deploy-production-ecs.yml
+++ b/.github/workflows/api-deploy-production-ecs.yml
@@ -45,7 +45,7 @@ jobs:
         run: |
           echo "https://${{ secrets.GH_PRIVATE_ACCESS_TOKEN }}:@github.com" > ${HOME}/.git-credentials
           git config --global credential.helper store
-          make install-packages opts="--extra private --extra auth-controller"
+          make install-packages opts="--extra private"
           rm -rf ${HOME}/.git-credentials
 
       - name: Generate MCP schema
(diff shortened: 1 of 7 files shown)


matthewelwell:
Superseded by #7667

PR #7667, opened 2026-06-01
ci: Run flagsmith-private test suites inline on 0.9.0

matthewelwell:

## Changes

Contributes to Flagsmith/flagsmith-private#177, supersedes #7665.

Follows on from #7656, which consolidated `rbac` and `workflows_logic` into `flagsmith-private` and runs their suites inline via `make test`. This extends the same pattern to `flagsmith-private` 0.9.0, which also bundles `auth_controller`, `licensing` and `flagsmith-ldap` so the private suites run inline against Core API again.

- Bump `flagsmith-private` to `>=0.9.0`. It now ships `auth_controller`, `licensing` and `flagsmith-ldap` so drop the extras and git source references; `integrate-private-tests` no longer clones them separately.
- `integrate-private-tests` pulls each app's tests from `flagsmith-private` at the installed wheel's tag and drops them under `tests/` so they inherit Core API's `tests/conftest` and run inline with `make test`.
- `.gitignore` the cloned repo and the staged `tests/*_tests` directories.

`scim` and `release_pipelines_logic` are deliberately left out for now and continue to run in flagsmith-private's own CI:

- `scim`: its compliance tests require an `SCIM_SERVICE_PROVIDER` config (notably dropping the request-based `BASE_LOCATION_GETTER`) that directly conflicts with Core API's own `tests/integration/scim` tests, and django-scim2 reads that setting once at import — so both suites cannot share one pytest session.
- `release_pipelines_logic`: one test asserts a non-deterministic email ordering that only holds in isolation; it fails when collected alongside the full Core API suite.

Both gaps are tracked in Flagsmith/flagsmith-private#177.

Locally, against the Compose Postgres, with `--extra dev --extra private` installed:

1. `make integrate-private-tests` to stage the suites.
2. Ran the staged suites under Core API's pytest config (`--ds=app.settings.test`, `--import-mode=importlib`):

   ```
   uv run pytest tests/rbac_tests tests/workflow_tests tests/auth_controller_tests \
     tests/saml_integration_tests tests/saml_unit_tests
   ```

   Result: 343 passed.

### DIFF
--- a/api/Makefile
+++ b/api/Makefile
@@ -149,16 +149,22 @@ generate-grafana-client-types:
 		--wrap-string-literal \
 		--special-field-name-prefix=
 
+# scim and release_pipelines_logic are deliberately not pulled here: scim's
+# compliance tests need an SCIM_SERVICE_PROVIDER config that conflicts with Core
+# API's own scim tests (django-scim2 caches it globally), and release_pipelines
+# has an order-dependent assertion that only holds in isolation. Both run in
+# flagsmith-private's own CI.
+# TODO https://github.com/Flagsmith/flagsmith-private/issues/177
 .PHONY: integrate-private-tests
 integrate-private-tests:
 	$(eval FLAGSMITH_PRIVATE_REVISION := v$(shell uv run python -c 'import importlib.metadata; print(importlib.metadata.version("flagsmith-private"))'))
-	$(eval AUTH_CONTROLLER_REVISION := $(shell uv run python -c 'import tomllib; d=tomllib.load(open("pyproject.toml","rb")); s=d.get("tool",{}).get("uv",{}).get("sources",{}).get("auth-controller",{}); print(s.get("tag") or s.get("rev") or "")'))
-
-	git clone https://github.com/Flagsmith/flagsmith-private --depth 1 --branch ${FLAGSMITH_PRIVATE_REVISION} \
-		&& mv ./flagsmith-private/integration_tests/rbac tests/rbac_tests \
-		&& mv ./flagsmith-private/integration_tests/workflows_logic tests/workflow_tests
-	git clone https://github.com/flagsmith/flagsmith-auth-controller --depth 1 --branch ${AUTH_CONTROLLER_REVISION} && mv ./flagsmith-auth-controller/tests tests/auth_controller_tests
-	rm -rf ./flagsmith-private ./flagsmith-auth-controller
+	git clone https://github.com/Flagsmith/flagsmith-private --depth 1 --branch ${FLAGSMITH_PRIVATE_REVISION}
+	mv ./flagsmith-private/integration_tests/rbac tests/rbac_tests
+	mv ./flagsmith-private/integration_tests/workflows_logic tests/workflow_tests
+	mv ./flagsmith-private/integration_tests/auth_controller tests/auth_controller_tests
+	mv ./flagsmith-private/integration_tests/saml tests/saml_integration_tests
+	mv ./flagsmith-private/unit_tests/saml tests/saml_unit_tests
+	rm -rf ./flagsmith-private
 
 .PHONY: generate-docs
 generate-docs: generate-flagsmith-sdk-openapi
--- a/api/Makefile
+++ b/api/Makefile
@@ -149,6 +149,12 @@ generate-grafana-client-types:
 		--wrap-string-literal \
 		--special-field-name-prefix=
 
+PRIVATE_TEST_DIRS := tests/rbac_tests tests/workflow_tests tests/auth_controller_tests tests/saml_integration_tests tests/licensing_integration_tests tests/ldap_integration_tests tests/saml_unit_tests
+
+.PHONY: clean-private-tests
+clean-private-tests:
+	rm -rf $(PRIVATE_TEST_DIRS)
+
 # scim and release_pipelines_logic are deliberately not pulled here: scim's
 # compliance tests need an SCIM_SERVICE_PROVIDER config that conflicts with Core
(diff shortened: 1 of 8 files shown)


api/app/settings/test.py
```diff
@@ -1,5 +1,9 @@
 from app.settings.common import *  # noqa
-from app.settings.common import REST_FRAMEWORK
+from app.settings.common import INSTALLED_APPS, LDAP_INSTALLED, REST_FRAMEWORK
+
+if LDAP_INSTALLED:
+    INSTALLED_APPS = INSTALLED_APPS + ["flagsmith_ldap"]
+    LDAP_DEFAULT_FLAGSMITH_ORGANISATION_ID = None
```
matthewelwell, line 6: This is just replicating what was added [here](https://github.com/Flagsmith/flagsmith-private/pull/182). Specifically, see the comment [here](https://github.com/Flagsmith/flagsmith-private/pull/182/changes#diff-01c96a2a4d7b08430652ab4c332ad94f6c35db72082b7b2a6f26077bf9b83696R6-R14) for context.

api/Makefile
```diff
@@ -149,16 +149,36 @@ generate-grafana-client-types:
 		--wrap-string-literal \
 		--special-field-name-prefix=
 
+PRIVATE_TEST_DIRS := tests/rbac_tests tests/workflow_tests tests/auth_controller_tests tests/saml_integration_tests tests/licensing_integration_tests tests/ldap_integration_tests tests/saml_unit_tests
+
+.PHONY: clean-private-tests
+clean-private-tests:
+	rm -rf $(PRIVATE_TEST_DIRS)
```
matthewelwell, line 156: I added this just out of frustration at local testing.

api/Makefile
```diff
@@ -149,16 +149,36 @@ generate-grafana-client-types:
 		--wrap-string-literal \
 		--special-field-name-prefix=
 
+PRIVATE_TEST_DIRS := tests/rbac_tests tests/workflow_tests tests/auth_controller_tests tests/saml_integration_tests tests/licensing_integration_tests tests/ldap_integration_tests tests/saml_unit_tests
```
khvn26, line 152: nit: Claudism (extraction for extraction sake)

api/Makefile
```diff
@@ -149,16 +149,36 @@ generate-grafana-client-types:
...
-	git clone https://github.com/flagsmith/flagsmith-auth-controller --depth 1 --branch ${AUTH_CONTROLLER_REVISION} && mv ./flagsmith-auth-controller/tests tests/auth_controller_tests
-	rm -rf ./flagsmith-private ./flagsmith-auth-controller
+	git clone https://github.com/Flagsmith/flagsmith-private --depth 1 --branch ${FLAGSMITH_PRIVATE_REVISION}
+	mv ./flagsmith-private/integration_tests/rbac tests/rbac_tests
+	mv ./flagsmith-private/integration_tests/workflows_logic tests/workflow_tests
+	mv ./flagsmith-private/integration_tests/auth_controller tests/auth_controller_tests
+	mv ./flagsmith-private/integration_tests/saml tests/saml_integration_tests
+	mv ./flagsmith-private/integration_tests/licensing tests/licensing_integration_tests
+	mv ./flagsmith-private/integration_tests/flagsmith_ldap tests/ldap_integration_tests
+	mv ./flagsmith-private/unit_tests/saml tests/saml_unit_tests
```
khvn26, line 180: question: It's not immediately obvious to me why we run unit tests only for some of the packages.

Also, not sure if we should refer to https://github.com/Flagsmith/flagsmith-private/issues/177#issuecomment-4593892787 here with a TODO.

api/pyproject.toml
```diff
@@ -94,20 +94,8 @@ dependencies = [
 # without a PEP 517 `[build-system]` table, so setuptools fallback drops
 # them from the built wheel. We re-declare those transitive deps here so
 # the resolution matches what poetry produced on main.
```
khvn26: Looks like comment is stale, and can/should be removed.

khvn26 (commented):
Nice, neat diff 👍 Left a couple comments.

api/Makefile
matthewelwell: No idea either - this was your code, I just added to it 😄

api/Makefile
matthewelwell: I'm not sure that I agree - it's being used in 2 places so makes sense to extract I think?

khvn26: Apologies, I didn't spot the second one.

api/Makefile
khvn26: Right. I think the original thought was to run integration tests only, but we're now blurring the line by using Core API as the dev dependency. Completing Flagsmith/flagsmith-private#177 might ultimately allow us to run all tests (IMO a good thing). I'm leaning towards keeping the SAML unit tests.

api/pyproject.toml
matthewelwell: Good catch - removed in ebe0bb9d815980d81939807e9d80dbb27cfd064d

api/Makefile
matthewelwell: Right, yeah, sorry, I think I misunderstood the original comment here. 

> It's not immediately obvious to me why we run unit tests only for some of the packages.

I think the answer to that is just that SAML is the only repo that _has_ unit tests? I can try to make that more obvious with another comment if we want, but we do already have a TODO for https://github.com/Flagsmith/flagsmith-private/issues/177 issue in the docstring for this command.

khvn26: Oh, nice, sorry, fair point. Carry on then.

khvn26 (approved):
GG

PR #7668, opened 2026-06-01
docs(MCP): Rewrite MCP server page and generate the tool catalogue

khvn26:

## Changes

Contributes to Flagsmith/flagsmith-private#144.

Preview: https://docs-git-docs-rewrite-mcp-server-flagsmith.vercel.app/integrating-with-flagsmith/mcp-server

Rewrites the user-facing MCP server page for the self-built server, replacing the Gram-hosted page:

- Cloud (`https://mcp.flagsmith.com`) and self-hosted (`https://<your-flagsmith-host>/mcp/`) URLs.
- OAuth (auto-discovered from the MCP URL) and `Authorization: Api-Key` authentication.
- Per-client install tabs: Claude Code, Cursor, Claude Desktop, Codex CLI, Windsurf, Gemini CLI, VS Code, and a generic fallback.
- Gram deprecation / migration callout and example prompts.

Generates the tool catalogue from the OpenAPI schema instead of hand-maintaining it, so the descriptions stay in sync with the tools:

- `manage.py generate_mcp_tool_catalogue` renders an aligned Markdown table of the MCP-tagged operations (verbatim `operationId` + `description`). `--exclude` omits tools already present in another catalogue.
- `make generate-docs` writes the core catalogue (`_mcp-tool-catalogue.md`, gated by the existing diff-check). `api-tests-with-private-packages` derives the enterprise-only catalogue (`_mcp-tool-catalogue-enterprise.md`) from it (the enterprise tools only exist with private packages installed). The page composes both via MDX imports — hence the `.md` → `.mdx` rename.

Note: the page documents URLs/clients ahead of the server going live (doc-led). The install snippets and the Gram deprecation date (currently 30 June 2026) should be confirmed against the live server before/at release — in particular that the documented MCP URL exactly matches the `resource` in the server's OAuth metadata, or OAuth discovery fails.

- `make -C api generate-docs` reproduces `_mcp-tool-catalogue.md` (36 core tools) byte-identically; `make -C api typecheck` and ruff pass; docs Prettier passes on the rewritten page and both partials.
- Generated the enterprise catalogue with `flagsmith-private` installed (`--exclude` the core one) → the 7 enterprise tools; verified both partials are Prettier-stable.
- Added unit tests for the management command (`tests/unit/api/test_generate_mcp_tool_catalogue.py`): default rendering, `--exclude`, and pipe-escaping — all pass.


khvn26:
Vercel preview of the rewritten page: https://docs-qxrxqbi67-flagsmith.vercel.app/integrating-with-flagsmith/mcp-server

Zaimwa9 (commented):
Big big improvement. No specific comments on my side, surprised that codex doesn't have a dedicated command

.github/workflows/api-tests-with-private-packages.yml
```diff
@@ -60,6 +60,15 @@ jobs:
...
 
+      - name: Check MCP enterprise tool catalogue is up to date
+        env:
+          DOTENV_OVERRIDE_FILE: .env-ci
+        run: |
+          uv run python manage.py generate_mcp_tool_catalogue \
+            --exclude ../docs/docs/integrating-with-flagsmith/_mcp-tool-catalogue.md \
+            > ../docs/docs/integrating-with-flagsmith/_mcp-tool-catalogue-enterprise.md
+          git diff --exit-code ../docs/docs/integrating-with-flagsmith/_mcp-tool-catalogue-enterprise.md
+
```
emyller, line 71: Should we add a TODO mentioning https://github.com/Flagsmith/flagsmith/issues/7669?

api/api/management/commands/generate_mcp_tool_catalogue.py
```diff
@@ -0,0 +1,74 @@
...
+        "are installed. Pass --exclude to omit tools already listed in an existing "
+        "catalogue (used to derive the enterprise-only catalogue against the core one)."
+    )
+
+    def add_arguments(self, parser: ArgumentParser) -> None:
+        parser.add_argument(
+            "--exclude",
+            type=Path,
+            default=None,
+            help="Path to an existing catalogue whose tools should be omitted.",
```
emyller, line 27:
```suggestion
            help="Path to an existing catalogue which tools should be omitted.",
```

.github/workflows/api-tests-with-private-packages.yml
```diff
@@ -60,6 +60,15 @@ jobs:
           make integrate-private-tests
           rm -rf ${HOME}/.git-credentials
 
+      - name: Check MCP enterprise tool catalogue is up to date
```
Zaimwa9, line 63: This is intentional to fail before running the tests ? Do we want to run it after the tests instead ?

api/api/management/commands/generate_mcp_tool_catalogue.py
```diff
@@ -0,0 +1,74 @@
...
+    # catalogue is reproducible by `make generate-docs` and passes the docs
+    # Prettier check unchanged.
+    cells = [list(header)] + [[f"`{name}`", description] for name, description in rows]
+    widths = [max(len(row[col]) for row in cells) for col in range(len(header))]
+    lines = [
+        _render_row(cells[0], widths),
+        "| " + " | ".join("-" * width for width in widths) + " |",
+        *(_render_row(row, widths) for row in cells[1:]),
+    ]
+    return "\n".join(lines)
```
emyller, line 69: note: Not a fan of this, seems fragile. Feels like this code should give place to something that should already exist.

Zaimwa9 (approved):
Ok, great. Just flagged the CI to make sure it was intentional

docs/docs/integrating-with-flagsmith/mcp-server.mdx
```diff
@@ -0,0 +1,334 @@
...
+[Organisation API key](/integrating-with-flagsmith/flagsmith-api-overview/admin-api/authentication) by sending it in the
+`Authorization` header:
+
+```
+Authorization: Api-Key <your-api-key>
+```
+
+:::caution
+
+The `Api-Key ` prefix is required. The header value should look like `Api-Key ser.abc123…`, not the key on its own.
```
emyller, line 58:
```suggestion
The `Api-Key` prefix is required. The header value should look like `Api-Key ser.abc123…`, not the key on its own.
```

This isn't incorrect, but leads to a weird rendering:

Image

.github/workflows/api-tests-with-private-packages.yml
khvn26: I think all of docgen hooks are set to fail before tests currently: https://github.com/Flagsmith/flagsmith/blob/537b57e73fd27f40eb18750ec30ce63319c54efa/.github/workflows/api-pull-request.yml#L71-L74

docs/docs/integrating-with-flagsmith/mcp-server.mdx
```diff
@@ -0,0 +1,334 @@
...
+    --header "Authorization: Api-Key <your-api-key>"
+```
+
+Or run it locally over stdio:
+
+```bash
+claude mcp add flagsmith \
+    --env TRANSPORT=stdio \
+    --env FLAGSMITH_API_TOKEN=<your-api-key> \
+    -- uvx --from "git+https://github.com/Flagsmith/flagsmith.git@main#subdirectory=mcp" flagsmith-mcp
```
emyller, line 87: q: Would you think this warrants a PyPI package?

docs/docs/integrating-with-flagsmith/mcp-server.mdx
```diff
@@ -0,0 +1,334 @@
...
+`TRANSPORT=stdio` and `FLAGSMITH_API_TOKEN`.
+
+</TabItem>
+</Tabs>
+
+## Available tools
+
+The tools available to your assistant depend on what your Flagsmith deployment has installed. The core tools below are
+available on every deployment. The [enterprise tools](#enterprise-tools) require a paid SaaS plan or a self-hosted
+Enterprise deployment, and only appear when those features are present.
```
emyller, line 281: q: Did you have a chance to verify it? I realise I did not.

emyller (commented):
Late non-blocking review.

docs/docs/integrating-with-flagsmith/mcp-server.mdx
khvn26: Ah good catch, the tool list result is independent of the upstream API. This needs rewording.

docs/docs/integrating-with-flagsmith/mcp-server.mdx
khvn26: Not sure. I can see the devex improvement but I don't imagine stdio will be used a lot to streamline this command. There's no difference security-wise, IMO.

api/api/management/commands/generate_mcp_tool_catalogue.py
khvn26: Yeah full disclosure I didn't ask Claude to generate prettier-compatible markdown. :/

PR #7674, opened 2026-06-02
feat(experimentation): environment-scoped metrics & experiment results

gagantrivedi:

## Changes

Adds an environment-scoped, reusable **Metric** and the API to attach metrics to experiments. Data-model + API 

- **`Metric`** — env-scoped, soft-delete: `name`, `description`, `aggregation` (`count`/`sum`/`mean`/`occurrence`), JSON `definition`. Immutable (no update).
- **`ExperimentMetric`** — join with `expected_direction`; unique per `(experiment, metric)`.
- **API** (gated on `experimental_flags` + env admin):
  - `…/environments/{key}/experiment-metrics/` — list/create/retrieve/delete (delete blocked `409` while attached to an active experiment).
  - `…/experiments/{id}/metrics/` — attach/list/detach, with same-environment + unique-attach validation.

Unit tests under `api/tests/unit/experimentation/` (models, library CRUD, attach/detach, gating, validation). Full `experimentation` suite green (181 passed); `mypy`/`ruff`/`flagsmith-lint-tests` clean.

```bash
DATABASE_URL=… uv run --directory api pytest tests/unit/experimentation/
```

Manual (flag on, as env admin): create a metric → `201`; attach to an experiment → `201`, re-attach same metric → `400`; `PATCH` a metric → `405`; delete while attached → `409`, detach then delete → `204`.


api/experimentation/models.py
```diff
@@ -117,3 +117,63 @@ class Meta:
...
+    )
+    metric = models.ForeignKey(
+        Metric,
+        on_delete=models.CASCADE,
+        related_name="experiment_metrics",
+    )
+    expected_direction = models.CharField(
+        max_length=20,
+        choices=ExpectedDirection.choices,
+    )
```
Zaimwa9, line 195: Interesting. So you see it in the relation Experiment x Metrics. I could see it too there although i'm wondering if that's a reality.

Let's say we have those metrics:
- Conversion rate (up)
- Average basket (up)
- Time to activation (down)
- First time page render (down)

Is there a world in which we'd want an experiment to push it in the other direction ? If not i'd stick it to the metrics and maybe have the possibility to override it in an experiment (in v2)

Zaimwa9: Ok I think i'm actually mixing 2 things:
- the metric polarity (is it better up or is it better down as per what it is) -> I think we should also add this one
- the experiment impact (it should go up, it should keep it same level, it should impact it down) especially as a guardrail => `expected_direction` that we should keep

api/experimentation/serializers.py
```diff
@@ -76,6 +78,84 @@ def _validate_snowflake_config(config: dict[str, Any]) -> SnowflakeConfig:
...
+        return None
+
+
+class ExperimentMetricSerializer(serializers.ModelSerializer):  # type: ignore[type-arg]
+    metric = serializers.PrimaryKeyRelatedField(  # type: ignore[var-annotated]
+        queryset=Metric.objects.all(),
+    )
+    metric_name = serializers.CharField(source="metric.name", read_only=True)
+    aggregation = serializers.CharField(source="metric.aggregation", read_only=True)
+
```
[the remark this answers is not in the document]
Zaimwa9: Done in the validator instead

api/experimentation/views.py
```diff
@@ -251,3 +258,87 @@ def _transition_status(self, target_status: str) -> Response:
...
+        metric: Metric = serializer.save(environment=self._get_environment())
+        create_metric_audit_log(metric, self._get_user(self.request), action="created")
+
+    def destroy(self, request: Request, *args: object, **kwargs: object) -> Response:
+        instance: Metric = self.get_object()
+        if (
+            ExperimentMetric.objects.filter(metric=instance)
+            .exclude(experiment__status=ExperimentStatus.COMPLETED)
+            .exists()
+        ):
```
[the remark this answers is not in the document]
Zaimwa9: the delete-block query now filters `experiment__deleted_at__isnull=True`  so a soft-deleted experiment's attachments no longer block metric deletion

api/experimentation/serializers.py
```diff
@@ -76,6 +80,80 @@ def _validate_snowflake_config(config: dict[str, Any]) -> SnowflakeConfig:
...
+        experiment: Experiment = self.context["experiment"]
+
+        if experiment.status == ExperimentStatus.COMPLETED:
+            raise serializers.ValidationError(
+                "Cannot modify metrics of a completed experiment."
+            )
+
+        metric: Metric = attrs.get("metric", getattr(self.instance, "metric", None))
+
+        if metric.environment_id != experiment.environment_id:
```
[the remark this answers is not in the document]
Zaimwa9: field-level validation would reject a missing metric before validate runs. So safe to discard

api/experimentation/metric_definitions.py
```diff
@@ -0,0 +1,43 @@
...
+    event = definition.get("event")
+    if not event or not isinstance(event, str):
+        return "Definition must specify a non-empty 'event'."
+    return None
+
+
+# Supported definition versions -> their validator. Add an entry per new shape.
+METRIC_DEFINITION_VALIDATORS: dict[int, DefinitionValidator] = {
+    1: _validate_v1,
+}
```
Zaimwa9, line 23: The way I see it to evolve the schema without breaking existing experiments is to add / create a new shape and version it here.
I can see the stat engine also use versioning to generate the queries. Wdyt ?

gagantrivedi: Yeah, that was the idea

matthewelwell (approved):
Approving it based on @gagantrivedi 's approval.

Issue #7679, opened 2026-06-02
API: OpenAPI schema does not match live responses (breaks strict output validation)


khvn26:
The published OpenAPI schema (`/api/v1/swagger.json`) disagrees with actual API responses in several systematic ways. Any client that validates responses against the schema (e.g. the Flagsmith MCP server, which uses FastMCP's `from_openapi` output validation) fails on most endpoints. Found by crawling the MCP-tagged read endpoints against SaaS and validating each response against its declared schema.

### 1. Pagination `next`/`previous` typed `string`, returned `null`

Paginated responses declare `next` and `previous` as non-nullable `string`, but the API returns `null` when there is no adjacent page. Should be `string, nullable` (`anyOf: [string, "null"]`).

Affected: `list_organizations`, `list_organization_groups`, `list_organization_invites`, `list_organization_roles`, `list_environments`, `list_project_features`, `list_project_segments`, `list_project_change_requests`, `list_feature_multivariate_options`, `list_feature_segments`, `get_environment_feature_versions` (likely all paginated list endpoints).

### 2. Array-returning endpoints declared as a single object

The 200 response schema is the singular item serializer, but the endpoint returns a JSON array. Should be `type: array, items: <object>`.

- `GET /organisations/{id}/projects/` (list_projects_in_organization)
- `GET /projects/{id}/environments/` (list_project_environments)
- `GET /projects/{project_pk}/features/{feature_pk}/code-references/`
- `GET /projects/{project_pk}/features/{id}/evaluation-data/`

### 3. `get_feature_external_resources`: missing `count`

`GET …/feature-external-resources/` returns `{"results": []}` but the schema marks `count` as required. Either the response should include the full pagination envelope or the schema should not require `count`.

### 4. `owners` / `group_owners` typed as integer IDs but expanded in responses

Read responses expand these to nested user objects (`{id,email,first_name,last_name,last_login,uuid}`), while the schema types them as `array<integer>`. Affects `get_feature_flag` and `list_project_features` (`results[].owners[]`); likely anywhere a feature is serialised for read.

### Impact / workaround

The MCP server currently sets `validate_output=False` to avoid rejecting valid responses. We'd like to re-enable output validation once the schema matches reality.

PR #7670, opened 2026-06-02
feat(MCP): Run an MCP server out of OpenAPI specs

emyller:
- [x] I have read the [Contributing Guide](/Flagsmith/flagsmith/blob/main/CONTRIBUTING.md).
- [ ] ~I have added information to `docs/` if required so people know about the feature.~
- [x] I have filled in the "Changes" section below.
- [x] I have filled in the "How did you test this code" section below.

## Changes

Closes https://github.com/Flagsmith/flagsmith-private/issues/146

- Introduces a new `mcp/` app + basic infrastructure.
- Hooks up [FastMCP](https://gofastmcp.com/integrations/openapi) with our OpenAPI spec.

## How did you test this code?

- 100% coverage with integration + unit tests
- Ran `npx @modelcontextprotocol/inspector` locally, AI summary as follows:

  | `protocolVersion` | `serverInfo.name` | tools | resources | prompts |
  | --- | --- | --- | --- | --- |
  | `2025-11-25` | `Flagsmith` | 43 | 0 | 0 |

PR #7680, opened 2026-06-02
feat(MCP): Authentication, configuration and HTTP transport

khvn26:

## Changes

Contributes to the MCP server work. Stacked on #7670.

- Fix the API base URL — the OpenAPI spec paths already carry `/api/v1`, so the client base URL must be the origin only.
- Add authentication via `FlagsmithAuth` (`httpx.Auth`): forwards the caller's `Authorization` header under HTTP transport (`--header` passthrough), and falls back to a static `FLAGSMITH_API_TOKEN` (`Api-Key`) — the only credential available under stdio.
- Move configuration to `pydantic-settings` (`Settings`); require a token when `TRANSPORT=stdio`.
- Disable strict output validation pending #7679 (the upstream schema mistypes many responses).
- Test improvements: drive server tests through the public `create_server`, share the `openapi_spec` fixture from `conftest`.

- `make test`, `make lint`, `make typecheck` — green, 100% coverage.
- Manually against SaaS in both transports:
  - stdio with `FLAGSMITH_API_TOKEN` in env → read tools return real data.
  - http with the credential passed as an `Authorization: Api-Key …` header → real data; an unauthenticated call correctly returns `401`.


mcp/pyproject.toml
```diff
@@ -4,9 +4,10 @@ version = "0.1.0"
 description = "An MCP server connecting to the Flagsmith API."
 authors = [{ name = "Flagsmith", email = "support@flagsmith.com" }]
 readme = "README.md"
-requires-python = ">=3.14,<3.15"
+requires-python = ">=3.10"
```
khvn26, line 7: for better `uvx`-backed installation experience

mcp/src/flagsmith_mcp/auth.py
```diff
@@ -0,0 +1,28 @@
...
+    def __init__(self, global_master_api_key: str | None = None) -> None:
+        self._global_master_api_key = global_master_api_key
+
+    def auth_flow(
+        self, request: httpx.Request
+    ) -> Generator[httpx.Request, httpx.Response, None]:
+        if "authorization" not in request.headers:
+            # Prefer the caller's forwarded MCP `--header`; fall back to the
+            # server's own static token (the only credential under stdio).
+            forwarded = get_http_headers(include={"authorization"})
```
emyller, line 17:
```suggestion
        try:
            forwarded = get_http_headers(include={"authorization"})
        except RuntimeError:
            forwarded = {}
```

I think the stdio could fail otherwise. Not sure how it didn't.

emyller (changes requested):
Looks great. I've went ahead and merged most of this onto base (please merge/rebase with `-Xours`). Only one change suggested.

mcp/src/flagsmith_mcp/auth.py
khvn26: `get_http_headers` already does this — it's documented to "never raise … in which case an empty dict is returned", and internally wraps `get_http_request()` in `try/except RuntimeError: return {}`.

emyller: You're right. :)

emyller (approved):

Issue #7604, opened 2026-05-26
Environment banner text hardcoded to white, unreadable on light banner colours


adamvialpando:
## Describe the bug

The environment banner (the "butter bar" shown when `banner_text` is configured on an environment) renders its text in hardcoded white. When the chosen `banner_colour` is light, the text becomes effectively invisible against the background.

## To reproduce

1. Open an environment's settings
2. Enable the banner and set `banner_text` (e.g. `Production [Be Careful!]`)
3. Navigate to any page in that environment (Features, Segments, etc.)

The banner renders white-on-light and is unreadable.

## Expected behaviour

The banner text should remain legible regardless of the chosen `banner_colour`. Text color should be derived from the background.

## Affected code

`frontend/web/components/ButterBar.tsx` lines 73-85 hardcode `color: 'white'`:

## Screenshots

Will attach after issue creation (a banner with a pale cream `banner_colour` showing white text that is barely visible)

AAshGray:
Hello, I will work on this

PR #7684, opened 2026-06-02
fix: Improve ButterBar text contrast on banners

AAshGray:

Contributes to #7604 

## Changes

- Added `Utils.getContrastColour(...)` to `frontend/common/utils/utils.tsx`
- Updated `frontend/web/components/ButterBar.tsx` to use `Utils.getContrastColour(...)` for environment banner text colour instead of hardcoding white

- Ran the frontend locally
- Used an environment with a custom `banner_colour` and banner text
- Verified the banner text switched to black or white based on contrast

### DIFF
--- a/frontend/common/utils/utils.tsx
+++ b/frontend/common/utils/utils.tsx
@@ -145,15 +145,12 @@ const Utils = Object.assign({}, BaseUtils, {
     return res
   },
 
-  getContrastColour(backgroundColor: string): string {
-    if (!backgroundColor) {
-      return 'white'
-    }
+  getContrastColour(backgroundColor: string | null | undefined): string {
+    if (!backgroundColor) return 'white'
 
     try {
-      const parsed = Utils.colour(backgroundColor)
-      return parsed.luminosity() > 0.179 ? 'black' : 'white'
-    } catch (_error) {
+      return Color(backgroundColor).luminosity() > 0.179 ? 'black' : 'white'
+    } catch {
       return 'white'
     }
   },
--- a/frontend/common/utils/utils.tsx
+++ b/frontend/common/utils/utils.tsx
@@ -145,6 +145,19 @@ const Utils = Object.assign({}, BaseUtils, {
     return res
   },
 
+  getContrastColour(backgroundColor: string): string {
+    if (!backgroundColor) {
+      return 'white'
+    }
+
+    try {
+      const parsed = Utils.colour(backgroundColor)
+      return parsed.luminosity() > 0.179 ? 'black' : 'white'
+    } catch (_error) {
+      return 'white'
+    }
+  },
+
   copyToClipboard: async (
     value: string,
     successMessage?: string,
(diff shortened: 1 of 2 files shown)


frontend/common/utils/utils.tsx
```diff
@@ -145,6 +145,19 @@ const Utils = Object.assign({}, BaseUtils, {
...
+    if (!backgroundColor) {
+      return 'white'
+    }
+
+    try {
+      const parsed = Utils.colour(backgroundColor)
+      return parsed.luminosity() > 0.179 ? 'black' : 'white'
+    } catch (_error) {
+      return 'white'
+    }
```
talissoncosta, line 155: Thanks for extracting this 🙏 The `try`/`catch` is the right instinct, but `Utils.colour()` already swallows parse errors internally and falls back to `Constants.defaultTagColor` — so it never throws, the `catch` never runs, and on an invalid colour we'd compute contrast against a tag colour rather than returning white. Calling `Color` directly makes the `catch` actually fire:

```ts
if (!backgroundColor) return 'white'
try {
  return Color(backgroundColor).luminosity() > 0.179 ? 'black' : 'white'
} catch {
  return 'white'
}
```

(`Color` is already imported at the top of this file.)

frontend/web/components/ButterBar.tsx
```diff
@@ -76,7 +76,7 @@ const ButterBar: React.FC<ButterBarProps> = ({ billingStatus, projectId }) => {
           className='butter-bar font-weight-medium'
           style={{
             backgroundColor: environmentDetail.banner_colour,
-            color: 'white',
+            color: Utils.getContrastColour(environmentDetail.banner_colour ?? 'white'),
```
talissoncosta: The `?? 'white'` only triggers when `banner_colour` is unset — and passing `'white'` makes the helper return **black** text, while the actual background falls back to `.butter-bar`'s CSS colour (`$butter-bar-bg`, dark in dark mode). So: black text on a dark background in that case.

Edge case (the picker usually sets a colour), but were you after black here, or would dropping the fallback and letting the guard return white be safer?

```tsx
color: Utils.getContrastColour(environmentDetail.banner_colour),
```

talissoncosta (changes requested):
Thanks for the contribution! 🙏 Solid fix, reusable helper, WCAG luminance, the right `0.179` threshold, nicely scoped. The two notes above are minor and non-blocking. Looking forward to merging once they're sorted. 🎉

AAshGray:
Thanks for the clear explanations, let me know if there's anything else! 😄

frontend/web/components/ButterBar.tsx
```diff
@@ -76,7 +76,7 @@ const ButterBar: React.FC<ButterBarProps> = ({ billingStatus, projectId }) => {
           className='butter-bar font-weight-medium'
           style={{
             backgroundColor: environmentDetail.banner_colour,
-            color: 'white',
+            color:  Utils.getContrastColour(environmentDetail.banner_colour),
```
talissoncosta: Just one small detail, we seems to have an extra white space in here.
Looks pretty good overall, thanks for addressing the changes

AAshGray: Ah, it's always the little things. I've made the update, thanks again.

talissoncosta (approved):
Looks great! thanks for your contribution. 😃

PR #7690, opened 2026-06-02
ci: self-hosted renovate

matthewelwell:
Thanks for submitting a PR! Please check the boxes below:

- [x] I have read the [Contributing Guide](/Flagsmith/flagsmith/blob/main/CONTRIBUTING.md).
- [x] I have added information to `docs/` if required so people know about the feature.
- [x] I have filled in the "Changes" section below.
- [x] I have filled in the "How did you test this code" section below.

## Changes

Adds a workflow to run a renovate docker container in GH actions directly in order to authenticate with CodeArtifact and generate `uv.lock` correctly. 

Some pre-requisites: 

 - [x] Create a Github App for authentication

... and updates to complete after merge:

 - [ ] Disable Cloud Renovate from Flagsmith/flagsmith

## How did you test this code?

With much much pain. [Here](https://github.com/Flagsmith/flagsmith/actions/workflows/renovate.yml) is the evidence. 

Result: a successfully generated PR by Renovate to update `pytest` dependency [here](https://github.com/Flagsmith/flagsmith/pull/7689).

PR #7691, opened 2026-06-02
ci: Renovate semantic commit scopes

matthewelwell:

## Changes

- Fixes `matchManagers` for the API package rule from `uv` to `pep621` — Renovate uses the `pep621` manager for `pyproject.toml` files, so the rule was never matching and `semanticCommitScope: "API"` was never applied.
- Adds `"semanticCommitScope":""` globally so packages not matched by any rule produce `deps: ...` rather than falling back to `deps(deps): ...`.
- Adds new MCP and Github Actions scopes

I have tested part of this via running renovate manually when working on [this PR](https://github.com/Flagsmith/flagsmith/pull/7690), but some of this will need to be tested as and when renovate deems that changes are necessary.


emyller (dismissed):

khvn26:
shall we add an MCP scope here while we're at it? 👀

renovate.json
```diff
@@ -4,15 +4,19 @@
     "security:only-security-updates",
     ":semanticCommitTypeAll(deps)"
   ],
-  "recreateClosed": true,
-  "rebaseWhen": "always",
```
matthewelwell, line 8: Removed as these were pure hallucinations. `recreateClosed` doesn't even exist in the schema and `rebaseWhen` doesn't have an `"always"` option 🤦

matthewelwell:
@khvn26 went one further and also added Github Actions :)

khvn26 (approved):

PR #7692, opened 2026-06-02
feat(MCP): OAuth 2.0 support for HTTP transport

khvn26:

## Changes

Closes Flagsmith/flagsmith-private#147

Adds OAuth 2.0 support for the MCP server's HTTP transport, as a pass-through resource server — the API stays the authorisation server and sole validator.

- `FlagsmithResourceAuth`: serves Protected Resource Metadata (RFC 9728) advertising the Flagsmith AS and the `mcp` scope, and returns `401 + WWW-Authenticate` when a request carries no credential, so MCP clients can discover and complete the OAuth flow.
- Any-scheme gate: a request with any `Authorization` header, `Bearer` OAuth token or `Api-Key`, is accepted and forwarded upstream. No introspection.
- OAuth discovery is the credential fallback: it is wired only when `transport=http` and no static `FLAGSMITH_API_TOKEN` is configured. stdio, a static token, and a forwarded `--header` all stay pure pass-through.
- New `MCP_SERVER_URL` setting for the public resource URL advertised in the metadata.

- `make test` (100% coverage), `make lint`, `make typecheck` — all green.
- Manually against SaaS:
  - HTTP, no token, no credential → `401` + `WWW-Authenticate` pointing at `/.well-known/oauth-protected-resource/mcp`.
  - PRM document advertises `https://api.flagsmith.com` as the AS and the `mcp` scope.
  - `Api-Key` header over HTTP → forwarded → real data (gate is scheme-agnostic).
  - Static token / stdio → no PRM served (pass-through).
  - Full browser OAuth flow via Claude Code (`claude mcp add --transport http …` → `/mcp` → log in at Flagsmith → authenticated tool calls).

### DIFF
--- a/mcp/pyproject.toml
+++ b/mcp/pyproject.toml
@@ -6,8 +6,8 @@ authors = [{ name = "Flagsmith", email = "support@flagsmith.com" }]
 readme = "README.md"
 requires-python = ">=3.10"
 dependencies = [
-    "fastmcp>=3.3.1,<4.0.0",  # Base MCP functionality
-    "pydantic-settings>=2.0.0,<3.0.0",  # Environment-driven configuration
+    "fastmcp>=3.3.1,<4.0.0",           # Base MCP functionality
+    "pydantic-settings>=2.0.0,<3.0.0", # Environment-driven configuration
 ]
 
 [project.scripts]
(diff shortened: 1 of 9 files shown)


germangarces (approved):
👍 LGTM

PR #7693, opened 2026-06-03
fix: handle DoesNotExist in delete_segment task to prevent race condition errors

### LINKED ISSUE #7354: Race condition in deleting segments
## Acceptance Criterion

- [ ] The asynchronous task `delete_segment` ignores it when the segment is already deleted.

---

Sentry Issue: [FLAGSMITH-API-5NB](https://flagsmith.sentry.io/issues/7444688241/?referrer=github_integration)

```
Segment.DoesNotExist: Segment matching query does not exist.
(3 additional frame(s) were not displayed)
...
  File "segments/tasks.py", line 14, in delete_segment
    Segment.objects.get(pk=segment_id).delete()

Failed to execute task 'tasks.delete_segment', with id 410046727. Exception: Segment matching query does not exist.
```

1795771535y-cell:
## Description

Fixes #7354 - race condition in delete_segment task when the segment has already been deleted by another process.

**Root Cause:** When the delete_segment task runs and the segment has already been removed (e.g., deleted via the API before the async task executes), Segment.objects.get() raises Segment.DoesNotExist, causing a Sentry error.

**Fix:** Wrap the delete in a 	ry/except Segment.DoesNotExist block, gracefully returning when the segment no longer exists.

### DIFF
--- a/api/segments/tasks.py
+++ b/api/segments/tasks.py
@@ -11,7 +11,10 @@ from segments.models import Segment
 
 @register_task_handler()
 def delete_segment(segment_id: int) -> None:
-    Segment.objects.get(pk=segment_id).delete()
+    try:
+        Segment.objects.get(pk=segment_id).delete()
+    except Segment.DoesNotExist:
+        return
 
 
 @register_task_handler()


matthewelwell:
Duplicate of https://github.com/Flagsmith/flagsmith/pull/7380

Thanks for the pull request @1795771535y-cell, but please make sure you check to see if there is an existing PR open to resolve a given issue before opening a new one.

PR #7698, opened 2026-06-03
feat(multivariate): add per-feature unique key to multivariate options

gagantrivedi:
> **WIP / draft.** First backend slice of "give every multivariate value a stable key so we know which variant a user was exposed to".

- [ ] I have added information to `docs/` if required so people know about the feature. _(deferred — the key isn't yet surfaced to SDKs/UI; docs will land with the SDK/experimentation slices.)_

## Changes

Contributes to 

Adds a nullable, per-feature-unique `key` to `MultivariateFeatureOption`, giving each variant a stable, human-readable identifier (e.g. `control`, `variant_a`). This is the foundation for attributing experimentation exposure/metrics to the specific variant a user was shown.

- **Model** (`features/multivariate/models.py`): nullable `key` CharField (`max_length=2000`, mirroring `Feature.name`) with a `("feature", "key")` `unique_together`. Postgres treats `NULL` as distinct, so the key stays optional now while non-null keys are unique per feature — no not-null enforcement, no backfill.
- **Migration** (`0009`): single additive migration (field + constraint), zero data migration.
- **Serializer** (`features/multivariate/serializers.py`): exposes `key`. The `(feature, key)` `unique_together` makes DRF both attach a `UniqueTogetherValidator` and mark the field `required=True` — either of which would break the existing no-key create flow — so the auto-validator and `required` are overridden, and uniqueness is enforced manually to return a clean `400` (the DB constraint remains the final guard).

Out of scope (follow-up slices): threading `key` through the flag engine + environment document, the SDK flag-response `variant_key`, experimentation consuming `key` for per-variant attribution, and the frontend variation-key input/display.

Automated tests, written TDD-first (Given/When/Then, both auth types via `admin_client_new`):

- **Model** (`tests/unit/.../test_unit_multivariate_models.py`): `key` defaults to `None`; duplicate non-null key on the same feature → `IntegrityError`; multiple null keys on the same feature allowed; same key across different features allowed.
- **API** (`tests/integration/.../test_integration_multivariate.py`): create with `key` round-trips the value; duplicate key → `400` with a clear message.

Verification run locally:
- New multivariate suite: 44 passed.
- Regression sweep (feature serializers, SDK environment document, engine/sdk/dynamo mappers, edge identities, import/export): 412 passed, 2 skipped.
- `ruff check`, `ruff format --check`, and `mypy` (strict, incl. tests): clean.


api/features/multivariate/models.py
```diff
@@ -50,6 +50,9 @@ class MultivariateFeatureOption(
         related_name="multivariate_options",
     )
 
+    # A stable, human-readable identifier for the variant.
+    key = models.CharField(max_length=2000, null=True, blank=True)
```
Zaimwa9: NIT: 2000 seems like a lot

api/tests/unit/features/multivariate/test_unit_multivariate_models.py
```diff
@@ -226,6 +230,66 @@ def test_mv_feature_state_value_get_skip_create_audit_log__segment_deleted__retu
...
+    feature: Feature,
+) -> None:
+    # Given / When
+    MultivariateFeatureOption.objects.create(feature=feature, string_value="foo")
+    MultivariateFeatureOption.objects.create(feature=feature, string_value="bar")
+
+    # Then
+    assert (
+        MultivariateFeatureOption.objects.filter(
+            feature=feature, key__isnull=True
```
Zaimwa9, line 268: This one is only testing null and not `""` which would trigger the constraint. Do we want to allow users to use  `""` and make them unique or normalize them into null ?

api/features/multivariate/serializers.py
```diff
@@ -21,6 +21,7 @@ class Meta:
             "string_value",
             "boolean_value",
             "default_percentage_allocation",
+            "key",
```
Zaimwa9, line 24: So, this serializer is also used in `CreateFeatureSerializer`, that would make `key` writable there too right. 

Just for context, the frontend is only creating MV options via the `mv-options` dedicated endpoint but I believe (not tested it though) that it would make it writable in the feature endpoint too without validation.

We should make it read_only here and writable only in the `MultivariateFeatureOptionSerializer`

Zaimwa9 (changes requested):
Flagged 2 possible loopholes but otherwise I haven't identified other leaks into existing MV features.
It's mostly about reviewing serializers using the  `NestedMultivariateFeatureOptionSerializer` and an alignment on empty string values.
It would be nice to also add tests on updating keys in MV options and then maybe add some coverage around `CreateFeatureSerializer` to sleep at peace?

api/features/multivariate/models.py
gagantrivedi: Yeah, I used feature_name length. reduced now

api/features/multivariate/serializers.py
gagantrivedi: Done in 30c338d86 — `key` is read-only in the nested serializer now, writable only via the mv-options endpoint. Added a test proving the feature endpoint ignores it.

api/tests/unit/features/multivariate/test_unit_multivariate_models.py
gagantrivedi: Went with rejection rather than normalisation — `""` now returns a 400; clients omit the field or send null for "no key". Done in 39206f027.

Zaimwa9 (approved):
🤞  great thanks

PR #7699, opened 2026-06-03
feat(multivariate): thread variant key into the environment document

gagantrivedi:
- [ ] I have added information to `docs/` if required so people know about the feature. _(deferred — the key isn't consumed by SDKs yet.)_

## Changes

Builds on #7698. Adds `key` to the engine `MultivariateFeatureOptionModel` and passes it through in `map_mv_option_to_engine`, so each variant's key is included in the environment document — both the SDK local-eval document and the DynamoDB/edge document `model_dump()` the same `EnvironmentModel`.

Variant selection is unchanged. Returning the key in SDK/identify responses is a later slice.

### Compatibility

- Old document (no `key`) → parses, `key` defaults to `None`.
- New document → older readers ignore the unknown field: these pydantic models use the default `extra='ignore'`, and `flagsmith-flag-engine` types are `TypedDict`s with no runtime validation.
- DynamoDB is schemaless — the new attribute is a no-op at the storage layer.

New unit tests cover the mapper (`key` reaches the engine model), the serialised document (`key` present in `model_dump()` output), and parsing an older document without `key`. Edge API integration test snapshots of the DynamoDB identity document were updated to include the field.

Ran the mapper suite and the document consumers (`edge_api`, `environments`, `util`, feature serializers, import/export) locally — all green; `ruff`/`mypy` clean.

### DIFF
--- a/api/tests/integration/conftest.py
+++ b/api/tests/integration/conftest.py
@@ -423,7 +423,11 @@ def identity_document(  # type: ignore[no-untyped-def]
         "multivariate_feature_state_values": [
             {
                 "percentage_allocation": 50,
-                "multivariate_feature_option": {"value": "50_percent", "id": 1},
+                "multivariate_feature_option": {
+                    "value": "50_percent",
+                    "id": 1,
+                    "key": None,
+                },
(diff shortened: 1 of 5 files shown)


Zaimwa9 (approved):
lgtm 👍

PR #7705, opened 2026-06-04
feat(MCP): Expose Prometheus metrics

khvn26:

## Changes

Contributes to https://github.com/Flagsmith/flagsmith-private/issues/152.

Instrument the MCP server with Prometheus metrics, served by `prometheus_client`'s standalone HTTP server on a dedicated port (`METRICS_PORT`, disabled by default) so the scrape endpoint stays off the MCP port and works under stdio transport too.

A FastMCP middleware records:

- `flagsmith_mcp_tool_call_duration_seconds{tool, status}` — tool call latency, including the upstream Flagsmith API request.
- `flagsmith_mcp_tool_result_bytes{tool, content}` — result payload size as a proxy for token cost. FastMCP ships the payload twice (text content block + `structuredContent`) and MCP clients differ in which they render into the agent's context — the `content` label tracks each type of content. See [modelcontextprotocol#1624](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1624) for more info.
- `flagsmith_mcp_tool_catalogue_bytes` — serialised `tools/list` payload size, the token cost every MCP session pays up front.

Also brings in `flagsmith-common[test-tools]` for the `assert_metric` fixture and `pytest-mock` as dev dependencies. Note: `requires-python` is bumped to `>=3.11` to match flagsmith-common's floor (3.10 is EOL this October).

Unit and integration tests (100% coverage gate). Manually: ran the server with `METRICS_PORT=9464`, verified the exposition on `:9464`, `/metrics` absent from the MCP port, and watched the metrics live in a local Prometheus while exercising tool calls (success, upstream error) and `tools/list` against the real OpenAPI catalogue.


mcp/tests/integration/test_metrics.py
```diff
@@ -0,0 +1,81 @@
...
+) -> None:
+    # Given the server started via the client fixture
+    # When
+    tools = await client.list_tools()
+
+    # Then a catalogue of two tools weighs at least a name and a schema each
+    assert tools
+    catalogue_bytes = REGISTRY.get_sample_value("flagsmith_mcp_tool_catalogue_bytes")
+    assert catalogue_bytes is not None
+    assert catalogue_bytes > len(tools) * 50
```
emyller, line 81: question: Does `50` means a [fair] assumption that a name+schema weighs at least 50 bytes?

mcp/pyproject.toml
```diff
@@ -4,24 +4,27 @@ version = "0.1.0"
 description = "An MCP server connecting to the Flagsmith API."
 authors = [{ name = "Flagsmith", email = "support@flagsmith.com" }]
 readme = "README.md"
-requires-python = ">=3.10"
+requires-python = ">=3.11"
```
emyller, line 7: question: do we still/ever need a `.python-version` file? I'd suggest deleting it as maintenance burden.

emyller (approved):
LGTM

mcp/pyproject.toml
khvn26: Different purposes. `requires-python` is being used for `uvx` distribution, `.python-version` for development.

mcp/tests/integration/test_metrics.py
khvn26: Yup.

PR #7714, opened 2026-06-05
fix(API): bump PyJWT to 2.13.0 (CVE-2026-48526) and add to dev dependencies

matthewelwell:

## Changes

Removes the pinned `pyjwt==2.12.0` entry from the `[tool.uv] override-dependencies` list so uv resolves 2.13.0, addressing CVE-2026-48526 flagged by Docker Scout. The CVE is not yet in the GitHub Advisory Database, which is why Renovate has not raised this automatically.

`uv lock --upgrade-package pyjwt` resolved to 2.13.0 cleanly with no other dependency changes. CI is the primary verification here.


emyller (changes requested):
Sorry, I'm a bit conflicted because we still make _direct_ use of `jwt` [here](https://github.com/Flagsmith/flagsmith/blob/88a4ecfc86881d2b19e3d67278449ee4f003958f/api/tests/unit/integrations/github/test_unit_github_client.py#L54).

Arguably we could move it to dev-dependencies, though Claude suggests not depending on it at all — _I'm very unsurprised it wants to write code._

```py
def test_generate_jwt_token__valid_credentials__returns_decodable_jwt() -> None:
    # Given
    app_id = 12345

    # When
    token = generate_jwt_token(app_id=app_id)

    # Then
    _, payload_b64, _ = token.split(".")
    payload_b64 += "=" * (-len(payload_b64) % 4)  # restore base64 padding
    payload = json.loads(base64.urlsafe_b64decode(payload_b64))
    assert payload["iss"] == str(app_id)
```

matthewelwell:
@emyller interesting, and thanks for the thorough review! 

So, there's a couple of things at play here: 

 1. The line I'm removing from `pyproject.toml` is not removing the dependency, but removing the override that we had in place since the migration from poetry to uv. See the comment [here](https://github.com/Flagsmith/flagsmith/pull/7714/changes#diff-40f8f985c548e2b57b803d17f42833f8a4d94603ca5b5e0bf48de51872809400R161-R163) for more context. 
 2. I didn't realise we were directly relying on `pyjwt` - that's a good catch. Given that, I think we should probably add it to the dev dependencies and let `uv` handle the resolution between the transient dependency through a few different direct dependencies. 

I've made the change proposed in (2) in ca5a125, let me know if you disagree with the approach.

emyller (approved):
LGTM

PR #7720, opened 2026-06-05
fix(Runtime): Fix uv compatibility in local envs

emyller:
- [x] ~I have added information to `docs/` if required so people know about the feature.~

## Changes

Allow local environments to have newer uv than what's tracked in `main`, rather than force developers to downgrade/uvx-fu.

Running stuff, and observing no changes to anything.

### DIFF
--- a/api/pyproject.toml
+++ b/api/pyproject.toml
@@ -158,7 +158,7 @@ flagsmith-private = { index = "flagsmith-pypi-production" }
 clickhouse-driver = { git = "https://github.com/Flagsmith/clickhouse-driver", branch = "newjson" }
 
 [tool.uv]
-required-version = "0.11.18"  # Ensure this matches the version in .pre-commit-config.yaml
+required-version = ">=0.11.18"  # Ensure this matches the version in .pre-commit-config.yaml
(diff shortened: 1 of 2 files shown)


api/pyproject.toml
```diff
@@ -158,7 +158,7 @@ flagsmith-private = { index = "flagsmith-pypi-production" }
 clickhouse-driver = { git = "https://github.com/Flagsmith/clickhouse-driver", branch = "newjson" }
 
 [tool.uv]
-required-version = "0.11.18"  # Ensure this matches the version in .pre-commit-config.yaml
+required-version = ">=0.11.18"  # Ensure this matches the version in .pre-commit-config.yaml
```
matthewelwell, line 161: Should this be changed in the pre-commit config as well?

matthewelwell: I guess not, since we need to explicitly tell pre-commit which version to use.

matthewelwell (approved):

api/pyproject.toml
emyller: I did try some Renovate-fu to help updating uv across files, but it quickly became more than I'd handle for now.

PR #7732, opened 2026-06-08
feat: add Heap region support

### LINKED ISSUE #7730: Support Heap integration for EU tenant

The heap integration does not support non-us Heap instances.

As a user I would like to be able to select which region I want the integration to target.

The heap integration is hardcoded to target `https://heapanalytics.com/` (`HEAP_API_URL = "https://heapanalytics.com/"`).

Any user with a Heap instance in Europe can not currently use our integration as the European heap instances are available at `http://eu.heapanalytics.com/`.

N/A

### Additional context

_No response_

SahilJat:

  [x] I have read the Contributing Guide /Flagsmith/flagsmith/blob/main/CONTRIBUTING.md.
  [ ] I have added information to  docs/  if required so people know about the feature.
  [x] I have filled in the "Changes" section below.
  [x] I have filled in the "How did you test this code" section below.

  ## Changes

  Closes #7730 

  The Heap integration was hardcoded to target  https://heapanalytics.com , preventing users with European Heap instances ( https://eu.heapanalytics.com ) from using the integration.

  This PR adds  base_url  support to the Heap integration, following the same pattern as the existing Amplitude integration:

  •  api/integrations/heap/constants.py  — New file with  DEFAULT_HEAP_API_URL  constant.
  •  api/integrations/heap/heap.py  — Use  config.base_url  instead of the hardcoded  HEAP_API_URL  constant.
  •  api/integrations/heap/serializers.py  — Expose  base_url  in the API serializer fields.
  •  frontend/common/stores/default-flags.ts  — Add  base_url  field to the Heap integration UI configuration.

  This change is fully backwards compatible — existing configurations receive the default US URL via the data migration, and new configurations without an explicit  base_url  also default to US.

  ### Unit tests

  Added and updated tests in:

  •  api/tests/unit/integrations/heap/test_unit_heap.py :
      • Existing test now asserts default URL resolves to  https://heapanalytics.com/api/track .
      • New test  test_heap_wrapper__eu_base_url__uses_eu_url  verifies EU URL is used correctly.
  •  api/tests/unit/integrations/heap/test_unit_heap_views.py :
      • Updated list test to assert  base_url  is present in API response with default value.
      • Updated create test to verify default  base_url  is set when omitted.
      • New test  test_create_heap_config__with_eu_base_url__returns_created  verifies creating a config with a custom EU base URL.

  ### Manual verification

  1. Start the API:  cd api && make serve
  2. Create a Heap config without  base_url  — confirm it defaults to  https://heapanalytics.com :
    curl -X POST http://localhost:8000/api/v1/environments/<ENV_KEY>/integrations/heap/ \
      -H "Authorization: Token <TOKEN>" \
      -d '{"api_key": "test-key"}'

  3. Create a Heap config with EU  base_url  — confirm it is stored and returned:
    curl -X POST http://localhost:8000/api/v1/environments/<ENV_KEY>/integrations/heap/ \
      -H "Authorization: Token <TOKEN>" \
      -d '{"api_key": "test-key", "base_url": "https://eu.heapanalytics.com"}'

  4. Start the frontend ( cd frontend && ENV=local npm run dev ) and navigate to Integrations → Heap Analytics — confirm the "Base URL" field is visible alongside the API Key field.


Zaimwa9:
@gemini-code-assist review

api/integrations/heap/migrations/0004_add_default_base_url.py
```diff
@@ -0,0 +1,27 @@
...
+    operations = [
+        migrations.RunPython(
+            add_default_base_url, reverse_code=migrations.RunPython.noop
+        ),
+        migrations.AlterField(
+            model_name="heapconfiguration",
+            name="base_url",
+            field=models.URLField(default="https://heapanalytics.com"),
+        ),
+    ]
```
Zaimwa9: I think we should be able to avoid a migration and either set the default when saving the record or at runtime

Zaimwa9 (changes requested):
Overall we are able to drop the database migration. I guess this was inspired by the amplitude but with heap the url field is `null` for all the integrations and previous records will still be using the same url. 
So `self.url = "{config.base_url or DEFAULT_HEAP_API_URL}/api/track"` should do the job.

We are still discussing internally how to address the frontend so a first step would be to focus on the backend. We will discuss and come back to you

api/tests/unit/integrations/heap/test_unit_heap_views.py
```diff
@@ -103,10 +105,37 @@ def test_list_heap_config__config_exists__returns_config_list(
...
+        url,
+        data=json.dumps(data),
+        content_type="application/json",
+    )
+
+    # Then
+    assert response.status_code == status.HTTP_201_CREATED
+    config = HeapConfiguration.objects.get(environment=environment)
+    assert config.base_url == "https://eu.heapanalytics.com"
+
```
Zaimwa9, line 137: Would it be possible to add a test that spies on the request and assert the url is the right one ?

Something like (to be verified):
```
@pytest.mark.parametrize(
    "base_url, expected_url",
    [
        (None, "https://heapanalytics.com/api/track"),
        ("https://eu.heapanalytics.com", "https://eu.heapanalytics.com/api/track"),
    ],
)
def test_identify_integrations__heap_configured__posts_to_expected_url(
    mocker, environment, identity, base_url, expected_url,
):
    # Given
    HeapConfiguration.objects.create(
        environment=environment, api_key="abc-123", **({"base_url": base_url} if base_url else {}),
    )
    mocked_post = mocker.patch("integrations.heap.heap.requests.post")

    # When
    identify_integrations(identity, identity.get_all_feature_states())

    # Then
    assert mocked_post.call_args.args[0] == expected_url
```

Zaimwa9 (dismissed):
Thanks. Ideally just adding a coverage test asserting the request is made with the right url

matthewelwell (approved):
Approving based on @Zaimwa9 's review, and we will add his suggested test in a separate PR. Thanks for the contribution @SahilJat .

PR #7740, opened 2026-06-10
feat(experimentation): add experiment exposures model, warehouse query and summary builder

gagantrivedi:
- [ ] I have added information to `docs/` if required so people know about the feature. _(deferred — flag-gated and not user-visible yet; docs land with the UI.)_

## Changes

Contributes to the experimentation results layer (v0.1: experiment exposures panel). First backend slice — tasks, API and UI follow in stacked PRs.

- **`ExperimentExposures`** — one row per experiment (OneToOne), updated in place via `record_refresh(summary, as_of)` / `record_failure()`. Failure stamps `last_error_at` and preserves the last good `payload`/`as_of`. Every refresh recomputes the full window (`started_at` → `as_of`): identity dedup, multi-variant quarantine and late event delivery make incremental windows unsound.
- **`compute_exposures_summary(...)`** — single entry point: selects bucket granularity (hourly ≤ 72h windows, daily beyond), queries ClickHouse, builds the summary. Query and payload cannot disagree on granularity.
- **Query** — first-exposed identities per variant per time bucket from `$flag_exposure` events. Identities count once, in the bucket of their first exposure (immune to at-least-once duplicate delivery). Identities seen in >1 variant are flagged via an out-of-band `quarantined` column — a `$multiple` sentinel would collide with a real variant of that name. Bucket boundaries pinned to UTC in the column type, so the driver returns aware datetimes.
- **`ExposuresSummary`** — frozen dataclass tree, serialised once via `dataclasses.asdict()` at the model boundary. Increments-only payload: per-bucket `new_identities` per variant plus the `excluded_identities` count (quarantined identities appear nowhere else). No derivable fields — cumulative series, per-variant totals, headline total, share %, control badge and ordering are all computed client-side, using the reserved `"control"` key (enforced by the multivariate serializer).

- Unit tests: summary builder (bucket grouping, chronological ordering of sparse/unordered buckets, quarantine exclusion, empty data, granularity selection via the orchestrator), query service (row mapping, UTC bucket typing, parameters), model methods (refresh clears the error marker; failure preserves the payload; stored JSON shape pinned exactly).
- `pytest tests/unit/experimentation/` — 247 passed; `ruff` + `mypy` strict clean.


api/experimentation/mappers.py
```diff
@@ -0,0 +1,115 @@
...
+        for b in buckets:
+            if b.bucket == bucket_start:
+                running[b.variant] += b.new_units
+        points.append(
+            ExposureTimeseriesPoint(
+                bucket=_isoformat_utc(bucket_start),
+                cumulative_units=dict(running),
+            )
+        )
+    return points
```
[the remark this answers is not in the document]
gagantrivedi: Applied in dfd2b7264 (both here and in `_build_variants`, which had the same shape). Worth noting the practical bounds, though: granularity selection caps buckets at ~72 hourly points (windows ≤72h) or ~90 daily points (90-day events TTL), and this mapper runs in the task processor when a snapshot is computed — never in the request path — so the quadratic scan couldn't reach API latency. Taking the grouped version anyway since it's no less readable and stays safe if the granularity rules ever loosen.

api/experimentation/services.py
```diff
@@ -88,6 +103,130 @@ def get_warehouse_event_stats(environment_key: str) -> WarehouseEventStats:
...
+    WHERE environment_key = %(environment_key)s
+        AND event = %(exposure_event)s
+        AND feature_name = %(feature_name)s
+        AND timestamp >= %(window_start)s
+        AND timestamp <= %(window_end)s
+    GROUP BY identifier
+)
+SELECT
+    quarantined,
+    if(quarantined, '', variant) AS variant,
```
Zaimwa9: I'm reading that aliasing is version sensitive and it could cause Cyclic aliases.
Sounds cheap to rename to something like `observed_variant` and avoid the risk

api/experimentation/models.py
```diff
@@ -129,6 +132,29 @@ class Meta:
         ]
 
 
+class ExperimentExposures(models.Model):
+    experiment = models.OneToOneField(
+        Experiment,
+        on_delete=models.CASCADE,
+        related_name="exposures",
+    )
+    as_of = models.DateTimeField(null=True)
```
Zaimwa9:
```suggestion
    as_of = models.DateTimeField(null=True, blank=True)
```
?

api/experimentation/services.py
```diff
@@ -88,6 +103,130 @@ def get_warehouse_event_stats(environment_key: str) -> WarehouseEventStats:
...
+        identifier,
+        any(value) AS variant,
+        uniqExact(value) > 1 AS quarantined,
+        min(timestamp) AS first_exposure
+    FROM events
+    WHERE environment_key = %(environment_key)s
+        AND event = %(exposure_event)s
+        AND feature_name = %(feature_name)s
+        AND timestamp >= %(window_start)s
+        AND timestamp <= %(window_end)s
```
Zaimwa9: NIT: one or the other (most likely end) should be exclusive `<`

api/experimentation/models.py
```diff
@@ -129,6 +132,29 @@ class Meta:
...
+    )
+    last_error_at = models.DateTimeField(null=True, blank=True)
+
+    def record_refresh(self, summary: "ExposuresSummary", as_of: datetime) -> None:
+        self.payload = asdict(summary)
+        self.as_of = as_of
+        self.last_error_at = None
+        self.save(update_fields=["payload", "as_of", "last_error_at"])
+
+    def record_failure(self) -> None:
```
Zaimwa9, line 153: We could add a log here, seems overkill to store the error itself

Zaimwa9 (changes requested):
Strong stuff, really neat architecture. I think the alias is the one to resolve and maybe the upper window exclusion in the query

api/experimentation/models.py
Zaimwa9: Ok just saw it was addressed in follow up https://github.com/Flagsmith/flagsmith/pull/7751/changes#diff-43ee72054ed1c3140630b7c406e76cd834b5bd250376bdf9f57e5272c5a81a61R55-R58

api/experimentation/services.py
gagantrivedi: Good catch — fixed in 6d336b956, though by restructuring rather than renaming: the blanking moved into the CTE (`if(uniqExact(value) > 1, '', any(value)) AS variant`), so no expression references a column it shadows and the outer SELECT is a plain column list. Avoids carrying two names for the same concept.

api/experimentation/models.py
gagantrivedi: Added in 6d336b956 (migration regenerated in place — the PR is unmerged).

api/experimentation/services.py
gagantrivedi: Made `window_end` exclusive in 6d336b956 — half-open `[start, end)`, pinned in the SQL assertions. Nothing depended on the inclusive end (every refresh recomputes the full window), but the convention removes boundary-instant ambiguity if windows are ever chained or compared.

Zaimwa9 (approved):
👍

PR #7745, opened 2026-06-10
feat: label variant key UI

Zaimwa9:

## Changes

Adds UI to set a **label** on multivariate variants (maps to the backend `key` field) and revamps the variants section of the feature modal's Value tab.

**Variant labels**

- Each variant shows an inline-editable label (slug characters only, max 255, `control` reserved, unique per feature).
- New variants get a default label (`Variant_n`, first free number) that is persisted on save; existing variants with no key fall back to a display-only placeholder.
- No save-flow changes — the key persists through the existing writable `/mv-options/` endpoint.

**Improvements**
- Added error management at the MV option level (returning invalid key or duplicated one)
- Error message appears in card
- Success response from MV options update only are also showing a success toast and updating the edition colorswatch

**Variants UI revamp**

- Each variant renders in a bordered, shadowed card: label header with a "Not saved" chip + delete icon.
- "Variants" section header with a primary "+ Add variant" button (outline "Create A/B/n Test" when none exist), and a tooltip explaining labels are required for experiments (behind `experimental_flags`).
- Control Value title shows the control percentage as a chip.
- `Input` gains reusable `underline` and `centered` props; aligned the value editor height with adjacent inputs.

**Experiment setup page**

- The variations table shows the variant label/key as the name and drops the unused Description column; `ContentCard` gains a `description` prop used by the setup step.

- Extended the E2E flag test: `editVariantLabel` edits a variant's label via the new UI, saves, reopens the modal, and asserts persistence. Audited existing selectors (`add-variation`, `featureVariationValue*`, `featureVariationWeight*`, `delete-multivariate`) for compatibility with the new layout.
- Added Storybook stories for the new `Input` variants (`Underline`, `UnderlineCentered`).
- Manually, against a local API:
  1. Create a multivariate feature; add variants — they default to `Variant_1`, `Variant_2`…
  2. Rename a label; verify validation (spaces, `control`, duplicates rejected with inline errors) and the "Not saved" chip appearing only on edited variants.
  3. Save, reopen — labels persist; clear a label — it saves as `null` and the placeholder returns.

image
image

### DIFF
--- a/frontend/web/components/modals/create-feature/tabs/FeatureValueTab.tsx
+++ b/frontend/web/components/modals/create-feature/tabs/FeatureValueTab.tsx
@@ -98,18 +98,18 @@ const FeatureValueTab: FC<FeatureValueTabProps> = ({
   const addVariation = () => {
     // Default the label to the first free Variant_n so new variants are
     // saved with a key even if the user never edits it. Variants with no
-    // key display as Variant_{index + 1}, so avoid those names too.
+    // key display (and persist) their fallback, so avoid those names too.
     const existingNames = multivariate_options.map(
-      (option, index) => option.key || `Variant_${index + 1}`,
+      (option, index) => option.key || Utils.getDefaultVariantKey(index),
     )
-    let variantNumber = multivariate_options.length + 1
-    while (existingNames.includes(`Variant_${variantNumber}`)) {
-      variantNumber += 1
+    let nextIndex = multivariate_options.length
+    while (existingNames.includes(Utils.getDefaultVariantKey(nextIndex))) {
+      nextIndex += 1
     }
     const newVariation = {
       ...Utils.valueToFeatureState(''),
       default_percentage_allocation: 0,
-      key: `Variant_${variantNumber}`,
+      key: Utils.getDefaultVariantKey(nextIndex),
     }
     onProjectFlagChange({
       multivariate_options: [...multivariate_options, newVariation],
--- a/frontend/web/components/modals/create-feature/tabs/FeatureValueTab.tsx
+++ b/frontend/web/components/modals/create-feature/tabs/FeatureValueTab.tsx
@@ -5,6 +5,7 @@ import Constants from 'common/constants'
 import { VariationOptions } from 'components/mv/VariationOptions'
 import { AddVariationButton } from 'components/mv/AddVariationButton'
 import ErrorMessage from 'components/ErrorMessage'
+import InfoMessage from 'components/InfoMessage'
 import WarningMessage from 'components/WarningMessage'
 import Tooltip from 'components/Tooltip'
 import Icon from 'components/icons/Icon'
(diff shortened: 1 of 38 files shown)


Zaimwa9:
@gemini can you please have a thorough look of the data flow from the creation of a multivariate option, it's initial creation (POST) request and a direct update (PUT) without any full refresh

Zaimwa9:
@gemini-code-assist review

talissoncosta (changes requested):
Looks pretty @Zaimwa9 . 

Thanks for addressing the new design and writing the stories. 
I caught this on storybook:
1. The input underlined variant does not work well on dark mode yet. Do you mind double check that ?

Image

Overall looks great!

frontend/web/styles/components/_input.scss
```diff
@@ -28,6 +28,56 @@ textarea {
 
 @import '~react-datepicker/dist/react-datepicker.css';
 
+// Borderless input with a bottom underline only; the underline turns
+// primary while focused. Used for inline edits (variant label, weight).
+.input-container.input-underline {
+  input.input {
+    border: none;
+    border-bottom: 1px solid $input-border-color;
```
talissoncosta, line 36: Nit (non-blocking): these underline colours map cleanly onto our semantic tokens, which already flip per theme:

- `$input-border-color` / `$white-alpha-16` → `var(--color-border-default)`
- `$primary` (focus) → `var(--color-border-action)`
- hover `$basic-alpha-48` / `$white-alpha-48` → closest is `var(--color-border-strong)` (0.24, a touch subtler than 0.48)

Because the tokens are theme-aware, this would also let the separate `.dark .input-container.input-underline` block below collapse into the base rule.

No need to change anything here — we can fold this into #7758 (Migrate Input to TS), which tokenises the base input properly. Flagging just so the new CSS lines up with where we're heading. 🙏

frontend/web/components/modals/create-feature/tabs/FeatureValueTab.tsx
```diff
@@ -155,13 +174,77 @@ const FeatureValueTab: FC<FeatureValueTabProps> = ({
...
+      'Failed to save this variation.'
+    )
+  })
+  const valueTitle = hasVariations ? (
+    <span className='d-inline-flex align-items-center'>
+      Control Value
+      <span className='ml-1'>
+        <Icon name='info-outlined' />
+      </span>
+      <span className='chip chip--xs ml-2'>{controlPercentage}%</span>
```
talissoncosta: I think would be good if we standardise the Badge/Chip component. I see that this is a recurring pattern and we still don't have a proper component to handle that. 
No action to be done in here, just checking if you agree on that, so I can create an issue to handle that separately.

talissoncosta (dismissed):

talissoncosta (dismissed):

talissoncosta (approved):

Issue #7035, opened 2026-03-24
Spike: Map and enforce scopes based on permissions


Zaimwa9:
### Context

We need a scope layer for OAuth tokens without duplicating the permission system.

We map scopes to existing RBAC permissions.

A new `OAuthScopePermission` class checks whether the token's scopes cover the view's required RBAC permission. Unmapped permissions are denied by default

### Enforcement rule
Effective permissions is the intersection between the user permissions and scopes granted

### What to do
- Define the `SCOPE_PERMISSION_MAP`:
  ```python
  SCOPE_PERMISSION_MAP = {
      "flag:read": {VIEW_PROJECT, VIEW_FEATURE_STATE, VIEW_ENVIRONMENT},
      "flag:write": {CREATE_FEATURE, UPDATE_FEATURE_STATE, DELETE_FEATURE},
      # ...
  }
  ```
- Build `OAuthScopePermission` class:
  - If request is not an OAuth token → return `True` (no-op for session users)
  - If the view has no declared RBAC permission → deny (default deny)
  - Look up which scopes cover the required RBAC permission → check the token has one
- Add `OAuthScopePermission` to the permission classes:
  ```python
  permission_classes = [IsAuthenticated, OAuthScopePermission, ProjectPermission]
  ```
- Define scope list in DOT settings (`OAUTH2_PROVIDER["SCOPES"]`)

### Scope-to-permission mapping

| Scope | RBAC Permissions | MCP Endpoints |
|---|---|---|
| `organisation:read` | *(org membership)* | `GET /organisations/`, `/organisations/{id}/groups/`, `/organisations/{id}/invites/` |
| `organisation:write` | `CREATE_PROJECT`, `MANAGE_USER_GROUPS` | `POST /organisations/{id}/invites/` |
| `project:read` | `VIEW_PROJECT`, `MANAGE_TAGS` | `GET /organisations/{id}/projects/`, `/projects/{id}/`, `/projects/{id}/environments/` |
| `project:write` | `CREATE_ENVIRONMENT` | `PUT /projects/{id}/` |
| `environment:read` | `VIEW_ENVIRONMENT` | `GET /environments/` |
| `environment:write` | `MANAGE_SEGMENT_OVERRIDES` | |
[...]

PR #7746, opened 2026-06-10
feat(MCP): Push MCP usage via OTel

emyller:

## Changes

Contributes to https://github.com/Flagsmith/flagsmith-private/issues/152

Filter incoming requests from MCP servers and emit an `mcp.tool.called` log to OTel containing 1. the original W3C baggage from the request, and 2. the organisation ID.

Telemetry can then be exported so we can track usage.

New unit + integration tests, and full local harness setup using an InfluxDB data storage as telemetry export target.

### DIFF
--- a/api/telemetry/middleware.py
+++ b/api/telemetry/middleware.py
@@ -33,12 +33,11 @@ class MCPUsageLoggerMiddleware:
             # - flagsmith.mcp.client.name
             # - flagsmith.mcp.client.version
             "status": "error" if response.status_code >= 400 else "success",
-            "organisation__id": (org_id := self._get_organisation_id(request)),
         }
-        if org_id is not None:
-            logger.info("tool.called", **event)
+        if (org_id := self._get_organisation_id(request)) is not None:
+            logger.info("tool.called", organisation__id=org_id, **event)
         else:
-            logger.warning("tool.called", **event)
+            logger.warning("tool.called", organisation__id=None, **event)
 
         return response
 
--- a/api/telemetry/middleware.py
+++ b/api/telemetry/middleware.py
@@ -8,10 +8,6 @@ from opentelemetry import baggage
 from telemetry.spans import get_span_attribute
 
 
-class UndefinedOrganisationError(Exception):
-    """The organisation can't be probed from context."""
-
-
 class MCPUsageLoggerMiddleware:
     """Emit telemetry events for MCP usage"""
 
@@ -37,16 +33,16 @@ class MCPUsageLoggerMiddleware:
             # - flagsmith.mcp.client.name
             # - flagsmith.mcp.client.version
             "status": "error" if response.status_code >= 400 else "success",
+            "organisation__id": (org_id := self._get_organisation_id(request)),
         }
-        try:
-            org_id = self._get_organisation_id(request)
-            logger.info("tool.called", organisation__id=org_id, **event)
-        except UndefinedOrganisationError:
-            logger.warning("tool.called", organisation__id=None, **event)
+        if org_id is not None:
+            logger.info("tool.called", **event)
+        else:
+            logger.warning("tool.called", **event)
 
         return response
 
-    def _get_organisation_id(self, request: HttpRequest) -> int:
+    def _get_organisation_id(self, request: HttpRequest) -> int | None:
         """Obtain the organisation ID from the request context."""
         from organisations.models import Organisation
 
@@ -59,6 +55,6 @@ class MCPUsageLoggerMiddleware:
             return request.user.organisations.get().id
         except (
             Organisation.DoesNotExist,
-            Organisation.MultipleObjectsReturned,
-        ) as error:
-            raise UndefinedOrganisationError from error
+            Organisation.MultipleObjectsReturned,  # Don't guess
+        ):
+            return None
--- a/api/telemetry/middleware.py
+++ b/api/telemetry/middleware.py
@@ -1,0 +1,94 @@
+from collections.abc import Callable
+
+import structlog
+from django.http.request import HttpRequest
+from django.http.response import HttpResponse
+from opentelemetry import baggage
+
+
+class UndefinedOrganisationError(Exception):
+    """The organisation can't be probed from context."""
+
+
+class MCPUsageLoggerMiddleware:
+    """Emit telemetry events for MCP usage"""
+
+    _url_param_to_organisation_lookup = {
+        "organisation_pk": "pk",
+        "project_pk": "projects__pk",
+        "environment_pk": "projects__environments__pk",
+        "environment_api_key": "projects__environments__api_key",
+    }
+
+    _pk_url_name_to_organisation_lookup = {
+        "organisation-projects": "pk",
+        "project-detail": "projects__pk",
+        "project-environments": "projects__pk",
+        "featurestates-detail": "projects__features__feature_states__pk",
+        "feature-segment-detail": "projects__features__feature_segments__pk",
+    }
+
+    def __init__(
+        self,
+        get_response: Callable[[HttpRequest], HttpResponse],
+    ) -> None:
+        self.get_response = get_response
+
+    def __call__(self, request: HttpRequest) -> HttpResponse:
+        response = self.get_response(request)
+
+        if not request.user.is_authenticated:
+            return response
+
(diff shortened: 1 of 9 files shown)


api/telemetry/middleware.py
```diff
@@ -0,0 +1,95 @@
...
+        }
+        try:
+            org_id = self._get_organisation_id(request)
+            logger.info("tool.called", organisation__id=org_id, **event)
+        except UndefinedOrganisationError:
+            logger.warning("tool.called", organisation__id=None, **event)
+
+        return response
+
+    def _get_organisation_id(self, request: HttpRequest) -> int:
```
khvn26: I'd find this easier to swallow if this was its own `get_organisation_id_from_request` service.

api/telemetry/middleware.py
```diff
@@ -0,0 +1,95 @@
...
+        from organisations.models import Organisation
+
+        assert request.user.is_authenticated  # NOTE: protected upstream
+        user_orgs = request.user.organisations.all()
+        try:  # Most of the time, the user belongs to one organisation
+            return user_orgs.get().id
+        except Organisation.MultipleObjectsReturned:
+            pass
+        except Organisation.DoesNotExist as error:
+            raise UndefinedOrganisationError from error
```
khvn26: I can't justify using an exception for flow control here. The middleware has to expect it now only to set a null attribute anyway. `int | None` return type for the org id getter seems much better to me. Everything becomes clearer and simpler to maintain.

khvn26 (commented):
I'd rather not have user.organisations.first() heuristics at all than allow faulty data. I think it's important to provide exact data.

I would still much prefer setting the org id to a span attribute in the actual code that retrieves it for permission handling. It's the idiomatic OTel way IMO, with less potential failure modes and overall LoC to maintain.

I understand the need to provide org id explicitly so it's documented in the catalogue. However, I'd consider spending time on codegen + log tooling to allow adding select span attributes as OTLP event attributes (using e.g. a lazy getter callable or a sentinel object as a value). This would allow us to pave a way to provide common Flagsmith logging context (e.g. org id, user uuid, plan name, persona type etc) exactly from the source without having to maintain specific plumbing for each case.

Feel free to dismiss my review if you get another one while I'm OOO.

emyller:
> I'd rather not have user.organisations.first() heuristics at all than allow faulty data. I think it's important to provide exact data.

I agree. Would you please point out, in this PR, why you believe the code is not producing exact data?

> I would still much prefer setting the org id to a span attribute in the actual code that retrieves it for permission handling. It's the idiomatic OTel way IMO, with less potential failure modes and overall LoC to maintain.

[a15009a](https://github.com/Flagsmith/flagsmith/pull/7746/commits/a15009a251ee8cd1edf585194f66efcddc47ad50) — note that the permission layer does **not** cover all endpoints used by MCP, leading us to the single-org user fallback. Fixing that has a large blast radius, which I experimented with, and prefer leaving to a separate issue.

> I understand the need to provide org id explicitly so it's documented in the catalogue. However, I'd consider spending time on codegen + log tooling to allow adding select span attributes as OTLP event attributes (using e.g. a lazy getter callable or a sentinel object as a value). This would allow us to pave a way to provide common Flagsmith logging context (e.g. org id, user uuid, plan name, persona type etc) exactly from the source without having to maintain specific plumbing for each case.

I understand the potential of span enrichment, and I think it warrants proper scoping.

api/telemetry/middleware.py
emyller: > I can't justify using an exception for flow control here.

Sorry, I firmly disagree. While I think using exceptions rather than interface for flow control is but a design choice, I don't mind going either way, erring to what seems to be the choice in this code base? b139c3ca682fbf1460b96f7fff9ce886c7e7f3fa

api/telemetry/middleware.py
emyller: ~Try with mayonnaise.~ Agreeing with your previous rationale, I'd rather see the span-based approach standardised in our API machinery (see [note](https://github.com/Flagsmith/flagsmith/pull/7746#issuecomment-4676963746)) than seeing this _evolving_ into a service. Let me know if you have a strong preference here.

api/telemetry/middleware.py
emyller: Arguably, the exception should have been private. FWIW, it does read better now given the body of `_get_organisation_id` has diminished.

api/telemetry/middleware.py
```diff
@@ -0,0 +1,59 @@
...
+        self.get_response = get_response
+
+    def __call__(self, request: HttpRequest) -> HttpResponse:
+        response = self.get_response(request)
+
+        if not request.user or not request.user.is_authenticated:
+            return response
+
+        if baggage.get_baggage("flagsmith.client.name") != "flagsmith-mcp":
+            return response
```
Zaimwa9: NIT: maybe we could invert the 2 conditions to avoid triggering the user lazy loading if the request is not from the MCP

Zaimwa9 (dismissed):
Small NIT if ever there are additional back and forth but from my point of view the implementation is good and opened topics have been resolved or deferred.
Approving to unblock, leaving to your appreciation whether to wait on Kim' confirmation on the points he raised

khvn26:
> Would you please point out, in this PR, why you believe the code is not producing exact data?

Good point, I misread it initially. Yeah, optimising for the common case seems like the right call to me.

> the permission layer does not cover all endpoints used by MCP

Aw shucks, I assumed otherwise. Thanks for uncovering this. Looks like a good bit of info for https://github.com/Flagsmith/flagsmith/issues/7035.

> and I think it warrants proper scoping

Yes, agreed, there are decisions to make - for example, we might want to implement our own context layer instead of using spans directly.

api/telemetry/middleware.py
khvn26: Agreed.

api/telemetry/middleware.py
khvn26: Thanks for the change despite the disagreement. Yes, a private exception would've worked better, but I'd still avoid using exceptions in middleware unless they are meant for Django to return the appropriate HTTP response.

khvn26 (dismissed):

api/telemetry/middleware.py
emyller: Good call — 033f911ade3e049475fae8796b2407c94db6db91

matthewelwell (approved):
Approving based on Kim's prior review and minor additional change.

emyller:
> Looks like a good bit of info for https://github.com/Flagsmith/flagsmith/issues/7035.

For reference:

(LLM content)
| Tool | Org id |
|---|---|
| `list_organization_groups` | span-originated |
| `list_organization_invites` | span-originated |
| `create_organization_invite` | span-originated |
| `get_project` | span-originated |
| `update_project` | span-originated |
| `list_project_environments` | span-originated |
| `list_environments` | span-originated |
| `list_project_features` | single-org fallback |
| `get_feature_flag` | span-originated |
| `update_feature` | span-originated |
| `create_feature` | span-originated |
| `get_feature_evaluation_data` | single-org fallback |
| `get_feature_code_references` | span-originated |
| `get_feature_external_resources` | single-org fallback |
| `get_feature_health_events` | span-originated |
| `list_feature_multivariate_options` | span-originated |
| `create_feature_multivariate_option` | span-originated |
| `update_feature_multivariate_option` | span-originated |
| `delete_feature_multivariate_option` | span-originated |
| `list_project_segments` | span-originated |
| `get_project_segment` | span-originated |
| `update_project_segment` | span-originated |
| `create_project_segment` | span-originated |
| `list_feature_segments` | single-org fallback |
| `delete_feature_segment` | span-originated |
| `create_segment_override` | span-originated |
| `update_environment_feature_state` | span-originated |
| `update_feature_state` | span-originated |
| `get_environment_feature_versions` | span-originated |
| `create_environment_feature_version` | span-originated |
| `get_environment_feature_version_states` | span-originated |
| `create_environment_feature_version_state` | span-originated |
| `update_environment_feature_version_state` | span-originated |
| `publish_environment_feature_version` | span-originated |
| `list_organizations` | single-org fallback |
| `list_projects_in_organization` | single-org fallback |

30 of 36 (83%) span-originated
(End of LLM content)

Tools from private modules aren't included.

PR #7751, opened 2026-06-11
feat(experimentation): refresh experiment exposures via task processor

gagantrivedi:
- [ ] I have added information to `docs/` if required so people know about the feature. _(deferred — flag-gated and not user-visible yet; docs land with the UI. Events catalogue regenerated here.)_

## Changes

Contributes to the experimentation results layer (v0.1: experiment exposures panel). Second backend slice, stacked on #7740 — the API endpoint and UI follow.

- **`compute_experiment_exposures(experiment_id)`** — computes the full-window exposures summary and stores it via `record_refresh`. Window is `started_at` → `ended_at or now`, so a completed experiment's final refresh is frozen at its end. Warehouse errors are absorbed into `record_failure` (the last good payload survives; nothing raises into the task processor). Never-started experiments are skipped.
- **`refresh_running_experiment_exposures`** — recurring sweep every `EXPERIMENTATION_EXPOSURES_REFRESH_MINUTES` (new setting, default 60) enqueueing one compute per RUNNING experiment; no-op when `EXPERIMENTATION_CLICKHOUSE_URL` is unset. No staleness filter: the sweep cadence is the refresh interval, and an `as_of`-older-than-interval check would skip every other sweep (task latency keeps `as_of` slightly younger than the interval).
- **Completing an experiment enqueues the final refresh** from `transition_experiment_status`, making the stored row the durable record once events pass the 90-day ClickHouse TTL.
- **structlog events** — `experimentation.exposures.computed` / `exposures.compute_failed`; events catalogue regenerated.

- Unit tests: compute task (stored payload/`as_of`, completed-experiment frozen window, failure preserves last good payload, never-started skip, log events asserted), sweep (enqueues running experiments only; skips without a warehouse DSN), complete action enqueues the final refresh / start does not.
- `pytest tests/unit/experimentation/` — 257 passed; `ruff` + `mypy` strict clean.


gagantrivedi:
Superseded: the task-processor approach is cut in favour of an explicit refresh endpoint, landing with the GET endpoint in the next PR.

PR #7754, opened 2026-06-11
feat(experimentation): experiment exposures read and refresh endpoints

gagantrivedi:
- [ ] I have added information to `docs/` if required so people know about the feature. _(deferred — flag-gated; docs land with the UI.)_

## Changes

Contributes to the experimentation exposures panel (v0.1). Follows #7740; UI lands next.

- **`GET …/experiments/{id}/exposures/`** — returns the stored row (`{"exposures": null}` before the first compute); never hits ClickHouse.
- **`POST …/experiments/{id}/exposures/refresh/`** — stamps `refresh_requested_at` (migration 0007) and enqueues the compute task; 202. Guards: 400 before start, 400 once final, 429 + `Retry-After` within `EXPOSURES_REFRESH_MIN_INTERVAL` (5 min).
- **`compute_experiment_exposures(experiment_id)`** — full-window recompute (`started_at` → `ended_at or now`); warehouse errors → `record_failure`, preserving the last good payload; skips deleted, never-started and final rows.
- **`ExperimentExposures.is_final`** — finality (`as_of >= ended_at`) defined once on the model and enforced in the task, so no caller can recompute a final row after the warehouse's 90-day event TTL and lose data.
- **`tasks` now imports `models`/`services`**, not the reverse; the `WarehouseConnection` hooks defer their task imports to the enqueue site.

Both endpoints sit behind the existing `ExperimentPermission`.

- Endpoint and task tests covering: row present/absent/stale-with-error, 202 + enqueue + stamp, 429 with exact `Retry-After`, finalising refresh allowed vs final row rejected, failure preserves payload, deleted/never-started skips, 403s.
- `pytest tests/unit/experimentation/` — 275 passed; `ruff` + `mypy` strict clean.


api/experimentation/views.py
```diff
@@ -268,6 +273,49 @@ def pause(self, request: Request, **kwargs: object) -> Response:
...
+            exposures.refresh_requested_at is not None
+            and timezone.now() - exposures.refresh_requested_at
+            < EXPOSURES_REFRESH_MIN_INTERVAL
+        ):
+            return Response(
+                {"detail": "A refresh was requested recently. Try again later."},
+                status=status.HTTP_429_TOO_MANY_REQUESTS,
+            )
+        exposures.record_refresh_request()
+        compute_experiment_exposures.delay(kwargs={"experiment_id": experiment.id})
```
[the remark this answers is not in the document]
gagantrivedi: bad bot! no ATOMIC_REQUESTS in our project

api/experimentation/tasks.py
```diff
@@ -11,3 +14,32 @@ def add_environment_key_to_ingestion(environment_api_key: str) -> None:
...
+    try:
+        summary = compute_exposures_summary(
+            environment_key=experiment.environment.api_key,
+            feature_name=experiment.feature.name,
+            window_start=experiment.started_at,
+            window_end=as_of,
+        )
+    except Exception:
+        exposures.record_failure()
+        return
```
Zaimwa9: Can we log an error here or within `record_failure` ?

api/experimentation/models.py
```diff
@@ -143,6 +143,15 @@ class ExperimentExposures(models.Model):
         models.JSONField(null=True, blank=True)
     )
     last_error_at = models.DateTimeField(null=True, blank=True)
+    refresh_requested_at = models.DateTimeField(null=True, blank=True)
+
+    @property
+    def is_final(self) -> bool:
+        # Recomputing a final row can only lose data: warehouse events expire.
+        ended_at = self.experiment.ended_at
+        return (
```
Zaimwa9, line 151: Can we also return true when an experiment has been running for more than 90 days and the first buckets data have expired ?

Zaimwa9 (changes requested):
Looks good but i'd appreciate logging refreshing errors and potentially a guard against refreshing the exposure data if an experiment has been running past the 90 days TTL?

api/experimentation/tasks.py
gagantrivedi: Done in 85e3a9fe0

api/experimentation/models.py
gagantrivedi: 90 day ttl doesn't really exist yet on the ClickHouse side. Let's think about this when we add it back (we'll probably make sure the experiment can't run for more than whatever ttl we decide makes sense).

Zaimwa9 (approved):

PR #7756, opened 2026-06-11
chore(api): Remove uv override-dependencies block

matthewelwell:

## Changes

Removes the `[tool.uv].override-dependencies` block, added during the Poetry to uv migration to keep resolution identical to the old `poetry.lock`. Nothing in it is load-bearing any more: main has since dropped the `clickhouse-driver` override, and `flagsmith-private` 0.13.0 lifted its `cryptography<44` cap.

Lockfile effect is three packages — `datamodel-code-generator` 0.25.8 → 0.64.1 (the override was contradicting the declared `>=0.64,<0.65`), `deprecation` 2.0.7 → 2.1.0, and `tomli` added. `cryptography`, `pyOpenSSL` and `pysaml2` are unchanged.

Also drops the `email-validator` dev dependency and its stale comment, both of which only existed for lockfile parity — nothing imports it.

`uv lock` and `uv sync` resolve cleanly, typecheck passes, and the API unit tests are green on 3.11, 3.12 and 3.13 in CI.


api/pyproject.toml
```diff
@@ -132,10 +132,8 @@ dev = [
     "diff-cover>=10.1.0,<11.0.0",
     "django-debug-toolbar",
     "ipython>=9.10.0,<10.0.0",
-    # datamodel-code-generator requires `pydantic[email]`, but our
-    # override-dependencies on pydantic strips the email extra. Re-add
-    # the transitive deps so poetry.lock parity is preserved.
-    "email-validator>=2.0.0",
+    # imported directly by trust_relationships, so declared rather than
+    # relied upon transitively
```
khvn26: nit: slop?

khvn26 (dismissed):

api/pyproject.toml
matthewelwell: yep, agreed

matthewelwell: 4634e6e

khvn26 (approved):

khvn26 (approved):

PR #7760, opened 2026-06-11
refactor(multivariate): move-mv-option-saves-to-rtk-query

Zaimwa9:

## Changes

Migrates the multivariate option save flow from the legacy Flux store to a single RTK Query mutation.

- New `saveMultivariateOptions` mutation diffs submitted options against the server's current state before upserting/deleting, so stale client state can no longer create duplicate variants.
- `editFeatureMv` becomes a thin adapter preserving the existing contracts: store events, per-card error attribution, and the positional option ordering consumed by both v1 and v2 feature-state saves. All failing options now surface errors, not just the first.
- Variants are now sorted by id at store ingress and created sequentially, so their order no longer reshuffles after a save (the API returns them unordered; a backend `ordering` fix will follow separately).

Out of scope (follow-up): retiring the store's event system and converting the modal to mutation hooks.

- Unit tests and E2E tests (including new mv-option E2E coverage for repeated saves, add/remove in one save, and v2 versioning).

### DIFF
--- a/frontend/common/services/useMultivariateOption.ts
+++ b/frontend/common/services/useMultivariateOption.ts
@@ -35,41 +35,42 @@ export const multivariateOptionService = service.injectEndpoints({
           (flagRes.data as ProjectFlag)?.multivariate_options || []
         const errors: Record<number, any> = {}
         // Results are written back by input index — downstream feature
-        // state saves map weights to option ids positionally.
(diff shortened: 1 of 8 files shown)


Zaimwa9:
@gemini-code-assist review

talissoncosta (approved):

PR #7769, opened 2026-06-12
feat(experimentation): Bayesian stats kernel

gagantrivedi:

## Changes

Contributes to the experimentation results scorecard (v0.2). Pure stats module — no Django, ClickHouse or schema changes; the query, model and endpoints follow in stacked PRs.

- **`VariantStats`** — per-variant sufficient statistics (`n`, `sum`, `sum_squares`) with derived mean and clamped unbiased variance; the only data the warehouse needs to return per metric × variant.
- **`compare_to_control(control, treatment)`** — relative lift with delta-method variance, Normal posterior under a flat prior: 95% credible interval and `chance_to_win = P(lift > 0)`; direction flips for "lower is better" metrics belong to the payload builder. Returns `None` when inference is undefined (an arm below two observations, or zero control mean); zero variance degenerates to certainty without dividing by zero.
- **`docs/docs/experiment-statistics.md`** — a plain-language guide to the engine for non-statisticians: a glossary of every UI term, plus exposures (available now) and the Bayesian scorecard/SRM (coming soon), clearly marked.
- **`srm_p_value(observed, expected_shares)`** — chi-squared goodness-of-fit survival, exact for integer degrees of freedom using only the stdlib (`erfc` base case + the incomplete-gamma recurrence; no scipy). `None` when not computable.

- 19 example-based tests: hand-derived proportions case (lift/CI/chance to 1e-4), identical arms, zero-variance degenerate certainty, undefined-inference inputs, chi-squared reference values for 1, 2 and 3 dof (0.0455, e⁻¹, 0.04601), heavy imbalance below the 0.001 threshold, non-computable SRM inputs, variance clamping.
- `pytest tests/unit/experimentation/` — 265 passed; `ruff` + `mypy` strict clean.

### DIFF
--- a/api/experimentation/stats.py
+++ b/api/experimentation/stats.py
@@ -55,8 +55,9 @@ def compare_to_control(
     treatment: VariantStats,
 ) -> Inference | None:
     # Inference is undefined without two observations per arm (no spread to
-    # measure) or a zero control mean (relative lift divides by it).
-    if control.n < 2 or treatment.n < 2 or control.mean == 0:
+    # measure) or a non-positive control mean (relative lift against it is
+    # meaningless, and divides by zero when the mean is exactly zero).
+    if control.n < 2 or treatment.n < 2 or control.mean <= 0:
         return None
 
     lift = (treatment.mean - control.mean) / control.mean
--- a/api/experimentation/stats.py
+++ b/api/experimentation/stats.py
@@ -1,0 +1,90 @@
+import math
+from collections.abc import Sequence
+from dataclasses import dataclass
+from statistics import NormalDist
+
+_STANDARD_NORMAL = NormalDist()
+_Z_95 = 1.959963984540054
+
+
+@dataclass(frozen=True)
+class VariantStats:
+    n: int
+    sum: float
+    sum_squares: float
+
+    @property
+    def mean(self) -> float:
+        return self.sum / self.n
+
+    @property
+    def variance(self) -> float:
+        return max(0.0, (self.sum_squares - self.sum**2 / self.n) / (self.n - 1))
+
+
+@dataclass(frozen=True)
+class Inference:
+    lift: float
+    ci_low: float
+    ci_high: float
+    chance_to_win: float
+
+
+def compare_to_control(
+    control: VariantStats,
+    treatment: VariantStats,
+) -> Inference | None:
+    if control.n < 2 or treatment.n < 2 or control.mean == 0:
+        return None
+
+    lift = (treatment.mean - control.mean) / control.mean
+    # Delta-method variance of the relative lift; the arms are independent.
+    variance = treatment.variance / (
+        treatment.n * control.mean**2
+    ) + treatment.mean**2 * control.variance / (control.n * control.mean**4)
+    standard_error = math.sqrt(variance)
+    if standard_error == 0:
+        certainty = 0.5 if lift == 0 else float(lift > 0)
+        return Inference(lift=lift, ci_low=lift, ci_high=lift, chance_to_win=certainty)
+
+    return Inference(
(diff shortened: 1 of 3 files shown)


api/experimentation/stats.py
```diff
@@ -0,0 +1,129 @@
...
+    chance_to_win: float  # probability (0–1) the treatment really beats control
+
+
+def compare_to_control(
+    control: VariantStats,
+    treatment: VariantStats,
+) -> Inference | None:
+    # Inference is undefined without two observations per arm (no spread to
+    # measure) or a zero control mean (relative lift divides by it).
+    if control.n < 2 or treatment.n < 2 or control.mean == 0:
```
Zaimwa9:
```suggestion
    if control.n < 2 or treatment.n < 2 or control.mean <= 0:
```

Zaimwa9 (dismissed):
Approving with a edge case guard, leaving it up to you

api/experimentation/stats.py
gagantrivedi: Good call — done in f33a43805

docs/docs/experiment-statistics.md
```diff
@@ -0,0 +1,155 @@
+---
+title: Experiment Statistics
+sidebar_label: Experiment Statistics
```
Zaimwa9, line 3: We might want to make it clear it's not yet available. "Coming Soon" ?

docs/docs/experiment-statistics.md
```diff
@@ -0,0 +1,155 @@
...
+sidebar_position: 5
+---
+
```
Zaimwa9, line 6: Discussed offline with Asaph, we could add a mention that the beta will soon be available and they can sign-up for it by contacting us

And that it is enterprise only

docs/docs/experiment-statistics.md
```diff
@@ -0,0 +1,155 @@
...
+Flagsmith's statistics engine answers three questions about an experiment: **Am I winning?** (is the variant better than
+control), **by how much?** (the lift), and **can I trust it?** (is the difference real, and was traffic split fairly).
+This page explains the terms you'll see, in plain language — no statistics background needed.
+
+:::info Availability
+
+- **Available now:** the **Exposures panel** — who was bucketed into each variant, over time.
```
Zaimwa9: Actually coming soon for users

docs/docs/experiment-statistics.md
```diff
@@ -0,0 +1,155 @@
...
+
+**Collecting data** — Not enough data yet to report a result, so numbers are withheld rather than shown when they'd be
+meaningless.
+
+**Last computed (as-of)** — Results are computed periodically, not on every page load. This timestamp tells you how
+fresh the figures are.
+
+## Exposures
+
+**Available now.**
```
Zaimwa9: Same

Zaimwa9 (changes requested):
Just some feedback to emphasize it's not available at all yet

docs/docs/experiment-statistics.md
gagantrivedi: Done in 32c858220.

docs/docs/experiment-statistics.md
gagantrivedi: Done in 32c858220.

docs/docs/experiment-statistics.md
gagantrivedi: Done in 32c858220.

docs/docs/experiment-statistics.md
gagantrivedi: Done in 32c858220.

Zaimwa9 (approved):
awesome

PR #6023, opened 2025-09-05
feat(CI): Upload e2e error videos to GitHub

emyller:
Experiment made with Claude Sonnet 4.

This is an attempt to improve our CI debugging workflow by moving e2e failure videos and/or screenshots as [artifacts to each pull request](https://docs.github.com/en/actions/tutorials/store-and-share-data). As of the creation of this issue, the team relies on a Slack channel that receives e2e video artifacts.

PR #7774, opened 2026-06-12
ci(frontend): upload Playwright artifacts when E2E tests fail

talissoncosta:

## Changes

When the E2E gate fails on the production deploy (or staging/ECS runs), the Playwright traces, screenshots and DOM snapshots die with the runner, so failures like the ones blocking today's deploy can only be debugged by reproducing locally.

This uploads `e2e/test-results` and the Playwright report as a run artifact when the E2E step fails:

- 7-day retention, since traces capture network payloads including the (throwaway, reseeded-per-run) e2e users' auth tokens.
- Artifact name includes `github.run_attempt` to avoid the immutable-name collision in `upload-artifact@v4` when a failed job is re-run.
- No-op on green runs (`if: failure()`), and `if-no-files-found: ignore` covers failures before Playwright starts.

Validated the action YAML parses. The step only runs on E2E failure in CI; the paths match Playwright's configured `outputDir` (`e2e/test-results`) and report directory, verified against a local failing run.


emyller (approved):
Thanks! I'm finally happier since https://github.com/Flagsmith/flagsmith/pull/6023 was closed.

PR #7779, opened 2026-06-14
fix(Utils): add guard against null rule value in validateRule

afsuyadi:

- [✔] I have read the [Contributing Guide](/Flagsmith/flagsmith/blob/main/CONTRIBUTING.md).
- [✔] I have added information to `docs/` if required so people know about the feature.
- [✔] I have filled in the "Changes" section below.
- [✔] I have filled in the "How did you test this code" section below.

## Changes

Contributes to #7536 

1. Error happens when `CreateSegment` component is trying to call method `validateRule()` by passing an object.
2. The method crashes when `split()` method is executed on a variable that isn't string-type. 
3. Inside method `validateRule()`, only _value_ prop that is accessed and doesn't have its own guard, so the method crashes when it's empty (either _null_ or _undefined_).
4. I add _loose equality guard_, to catch only _null_ and _undefined_ value.

1. I try to simulate this on my local by manually creating the Segment via UI, but it seems no error happens, because the button Create Segment itself is already been greyed out.
2. I deduce that this error happens via API backend hit.

[Screencast from 14-06-26 11:46:58.webm](https://github.com/user-attachments/assets/d6d8d390-aac1-45ce-94eb-c60e15422cf9)

### DIFF
--- a/frontend/common/utils/utils.tsx
+++ b/frontend/common/utils/utils.tsx
@@ -816,6 +816,9 @@ const Utils = Object.assign({}, BaseUtils, {
     }
 
     if (operatorObj?.value?.toLowerCase?.().includes('semver')) {
+      if (rule.value == null) {
+        return false
+      }
       return !!semver.valid(`${rule.value.split(':')[0]}`)
     }
 
@@ -826,7 +829,7 @@ const Utils = Object.assign({}, BaseUtils, {
       }
       case 'REGEX': {
         try {
-          if (!rule.value) {
+          if (rule.value == null) {
             throw new Error('')
           }
           new RegExp(`${rule.value}`)
@@ -836,6 +839,9 @@ const Utils = Object.assign({}, BaseUtils, {
         }
       }
       case 'MODULO': {
+        if (rule.value == null) {
+          return false
+        }
         const valueSplit = rule.value.split('|')
         if (valueSplit.length === 2) {
           const [divisor, remainder] = [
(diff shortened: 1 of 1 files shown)


manos-saratsis:

**0** critical · **1** warnings across 3 agents on `fix-error-create-segment` (3662bc2)

|---|---|---:|---:|---:|
| 🧪 Code Quality | ✅ done | 0 | 1 | 97 |
| 🛡️ Security | ✅ done | 0 | 0 | 100 |
| 📋 Compliance | ✅ done | 0 | 0 | 100 |

### Top findings
- 🟠 **Code Quality** — Near-identical value-mapping logic duplicated in valueToFeatureState and valueToTrait `frontend/common/utils/utils.tsx:821`

[View full report on Dromeas →](https://dromeas.ai/product/pr-review/2cfc2cbe-0e51-4461-b7c0-2e7f3ee2bc77)

frontend/common/utils/utils.tsx
```diff
@@ -830,6 +830,9 @@ const Utils = Object.assign({}, BaseUtils, {
         }
       }
       case 'MODULO': {
+        if (rule.value == null) {
```
Zaimwa9: Actually, let's avoid the loose comparison and keep the `!rule.value`. From the above (Regex), it hasn't generated an error so there is no reason for now to make a change there

Zaimwa9 (changes requested):
Thanks for the contribution. 
It would be preferable to avoid loosy comparisons if you don't mind.
We have been ramping up on using unit tests on the frontend. `validateRule` would be a perfect candidate if we could add a couple of parametrized tests around it, including ones validating the fix.
👍

frontend/common/utils/utils.tsx
afsuyadi: Okay I'll revert the change. Thanks for the review

Zaimwa9 (approved):
Thanks

PR #7781, opened 2026-06-15
feat(experimentation): results aggregation query and payload builder

gagantrivedi:
- [ ] I have added information to `docs/` if required so people know about the feature. _(deferred — internal; docs land with the results UI.)_

## Changes

Contributes to the experimentation results scorecard (v0.2): the ClickHouse aggregation and pure payload builder that feed the stats kernel (#7769). Model, endpoints, task and the ORM orchestrator (metric specs + per-environment expected shares) follow in the next PR.

- **Results query** — one pass: a shared exposures CTE (first-exposure dedup, quarantine, half-open window) joined to post-exposure metric events, conditionally aggregated per metric into `(n, sum, sum_squares)` plus per-variant identity counts for SRM. Aggregations: `occurrence` / `count` / `sum` / `mean`; the join is window-bounded so ClickHouse range-scans on the sort key.
- **`build_results_summary`** — pure: `compare_to_control` per treatment + `srm_p_value` on the counts. Inference is withheld below the data floor (n ≥ 50/arm, ≥ 5 conversions/arm for occurrence) and SRM below 100 identities; chance-to-win is flipped for lower-is-better metrics. Raw per-variant stats are kept; means/status/ordering are derived client-side.
- **Dataclasses** `MetricSpec` / `ResultsAggregates` / `MetricResult` / `ResultsSummary` (reusing the kernel's `VariantStats` / `Inference`); `asdict` is the wire shape.

- Unit (faked client): row mapping, per-aggregation expressions, windowed post-exposure join, no-metrics path; builder data floors, lower-is-better flip, SRM balanced/imbalanced/not-computable, exact wire shape via `asdict`.
- Ran the query against a local ClickHouse with seeded data — confirmed the numbers and caught a real JOIN-ON error the substring tests couldn't.
- `pytest tests/unit/experimentation/` — 285 passed; `ruff` + `mypy` strict clean.

> [!NOTE]
> CI has no ClickHouse (parked), so the query's CH-specific semantics are asserted by SQL substring and were checked manually against a local instance; the `mean`/`avgIf` path stays the least-covered until ClickHouse-in-CI lands.


api/experimentation/services.py
```diff
@@ -227,6 +250,178 @@ def get_exposure_buckets(
...
+unit_values AS (
+    SELECT
+        e.variant AS variant,
+        {value_expressions}
+    FROM exposures AS e
+    LEFT JOIN events AS m
+        ON m.identifier = e.identifier
+        AND m.environment_key = %(environment_key)s
+        AND m.event IN %(metric_events)s
+        AND m.timestamp < %(window_end)s
```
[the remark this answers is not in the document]
gagantrivedi: Done

api/experimentation/services.py
```diff
@@ -227,6 +250,178 @@ def get_exposure_buckets(
...
+    *,
+    expected_shares: dict[str, float],
+) -> ResultsSummary:
+    exposure_counts = aggregates.exposure_counts
+    total = sum(exposure_counts.values())
+    if expected_shares and total >= SRM_MIN_TOTAL_IDENTITIES:
+        srm = srm_p_value(
+            [exposure_counts.get(variant, 0) for variant in expected_shares],
+            list(expected_shares.values()),
+        )
```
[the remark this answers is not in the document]
gagantrivedi: Not needed here: the only producer builds shares with control as the remainder ((100 − Σ allocations)/100), so they sum to 1.0 by construction — percentages or 0.99-drift can't arise. If we ever
  wanted a guard it'd live in srm_p_value, not at this call site. Leaving as-is for now

api/experimentation/services.py
```diff
@@ -227,6 +248,179 @@ def get_exposure_buckets(
...
+GROUP BY variant"""
+    )
+
+
+def get_metric_variant_stats(
+    *,
+    environment_key: str,
+    feature_name: str,
+    window_start: datetime,
+    window_end: datetime,
```
Zaimwa9, line 239: I'm wondering how to deal with paused experiments. Still early to have any conclusions, throwing an idea seeing this, which would be to pass in an array of active periods (or paused periods to exclude).

No action from this comment

api/tests/unit/experimentation/test_services.py
```diff
@@ -538,3 +543,359 @@ def test_refresh_warehouse_connection_status__already_connected__is_noop(
...
+    metric_stats: dict[int, dict[str, VariantStats]],
+) -> ResultsAggregates:
+    return ResultsAggregates(
+        specs=specs,
+        exposure_counts=exposure_counts,
+        metric_stats=metric_stats,
+    )
+
+
+def test_get_metric_variant_stats__metrics__queries_and_maps_rows(
```
Zaimwa9, line 585: Can we add coverage for multi-variant scenarios please ?

api/experimentation/services.py
```diff
@@ -227,6 +248,179 @@ def get_exposure_buckets(
...
+        # per-identity average, zero when the identity has no matching events
+        return f"if(countIf({condition}) > 0, avgIf({value}, {condition}), 0)"
+    raise ValueError(f"Unsupported metric aggregation: {aggregation}")
+
+
+def _build_results_query(specs: Sequence[MetricSpec]) -> str:
+    if not specs:
+        return (
+            _EXPOSURES_CTE
+            + "\nSELECT variant, count() AS n"
```
Zaimwa9: NIT: looks like this one is not using the triple quoted string pattern

api/experimentation/services.py
```diff
@@ -227,6 +248,179 @@ def get_exposure_buckets(
...
+    value = "toFloat64OrZero(m.value)"
+    if aggregation == MetricAggregation.OCCURRENCE:
+        return f"countIf({condition}) > 0"
+    if aggregation == MetricAggregation.COUNT:
+        return f"countIf({condition})"
+    if aggregation == MetricAggregation.SUM:
+        return f"sumIf({value}, {condition})"
+    if aggregation == MetricAggregation.MEAN:
+        # per-identity average, zero when the identity has no matching events
+        return f"if(countIf({condition}) > 0, avgIf({value}, {condition}), 0)"
```
Zaimwa9: Sorry for the maybe dumb question, is this standard in A/B testing to treat a non converted identity value as a 0 when calculating the average?
As long as all the calculations are harmonized, I'm fine with it but it feels this approach is going to drag the value down, which we might not want for all type of metrics

api/tests/unit/experimentation/test_services.py
```diff
@@ -538,3 +543,359 @@ def test_refresh_warehouse_connection_status__already_connected__is_noop(
...
+    ]
+    mock_client = mocker.Mock()
+    mock_client.execute.return_value = rows
+    mocker.patch(
+        "experimentation.services._get_clickhouse_client",
+        return_value=mock_client,
+    )
+    specs = [
+        _spec(metric_id=7, event="purchase", aggregation=MetricAggregation.OCCURRENCE),
+        _spec(metric_id=9, event="revenue", aggregation=MetricAggregation.SUM),
```
Zaimwa9, line 612: Can we add mean and count ?

Zaimwa9 (changes requested):
I really like the way you architectured building the results from query to payload. I have some open questions and mostly coverage comments that would be better to add sooner than later

api/experimentation/services.py
gagantrivedi: Yes, deliberate — this is intent-to-treat: non-participants get unit value 0 so the mean is over all exposed identities. This is the standard approach: Statsig ["imputes 0s: the calculation is the total value over the count of units exposed"](https://docs.statsig.com/statsig-warehouse-native/features/understanding-experiments), GrowthBook says ["the denominator for mean metrics is always the number of users in an experiment variation"](https://docs.growthbook.io/app/metrics), and Amplitude ["computes that metric on all exposed users"](https://amplitude.com/docs/feature-experiment/advanced-techniques/advanced-metric-use-cases). It avoids self-selection bias and measures the real population-level impact.

api/tests/unit/experimentation/test_services.py
gagantrivedi: done

Zaimwa9 (approved):
Nice. One nit is the import of private suffixed `_EXPOSURES_CTE` but that can be changed in an incoming PR

PR #7782, opened 2026-06-15
test(e2e): fix permission write-race in environment-permission-test

talissoncosta:

## Changes

One of the intermittent E2E failures on the production deploy gate, distinct from the boot-loader deadlock (#7776) — here the app boots fine but `environment-permission-test` times out waiting for `switch-environment-production` after the permission to view it was just granted.

Two changes:

- **`setUserPermission` awaits the save.** The helper toggled the permission switch and immediately closed the modal + logged out. The toggle saves async (POST/PUT to `.../user-permissions/`), so logout could abort the in-flight request and the grant never persisted — the next user then couldn't see the entity. The backend has no permission read-cache, so the lost write is the most likely cause. Now we await the `/user-permissions/` response before closing the modal.
- **`reloadUntilVisible` helper.** Belt-and-braces for the read side: the environment list is fetched once on navigation, so a just-granted `VIEW_ENVIRONMENT` can't appear by waiting on the same page — it needs a reload. Used for the 'User with permissions can see environment' assertion.

`environment-permission-test` reproduced the `switch-environment-production` timeout (app booted, env absent). Traced the grant to an un-awaited `/user-permissions/` POST in `setUserPermission`. Locally re-ran the spec against production with the fix; the grant now persists and the env appears. As a timing race against shared prod state, the real signal is the deploy-gate pass rate.

Note: separate from the non-idempotent-seed failure (createEnvironment on a dirty seed) seen in project-permission-test — that's a teardown/seed-lifecycle issue, not addressed here.

### DIFF
--- a/frontend/e2e/helpers/e2e-helpers.playwright.ts
+++ b/frontend/e2e/helpers/e2e-helpers.playwright.ts
@@ -872,12 +872,36 @@ export class E2EHelpers {
     if (entityName) {
       await this.click(byId(`permissions-${entityName.toLowerCase()}`));
     }
+    // The toggle saves async via POST/PUT to .../user-permissions/. Await it
+    // before closing the modal — otherwise logout can abort the in-flight
+    // save and the grant never lands, so the next user can't see the entity.
+    const saved = this.page.waitForResponse(
+      (res) =>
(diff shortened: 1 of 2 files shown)


talissoncosta:

PR #7784, opened 2026-06-15
test(e2e): drop waitForLoadState('networkidle') from helpers

talissoncosta:

## Changes

Another intermittent E2E failure on the production deploy gate, distinct from the boot deadlock (#7776) and the permission write-race (#7782). Here the app is fully booted and on the right page, but `page.waitForLoadState('networkidle')` times out after 20s.

The dashboard loads HubSpot, LinkedIn, Google Ads and Chargebee scripts that fire tracking pixels and polling indefinitely, so the network never goes idle for 500ms and `networkidle` can't resolve regardless of whether the page is ready. Playwright explicitly discourages `networkidle` for this reason.

All four uses in the helpers (`gotoTraits`, the role-assignment helper, `createEnvironment`, `createTag`) were immediately followed by a concrete element wait — the real readiness signal — so the `networkidle` calls are redundant as well as flaky. Removed all four.

Observed the timeout in two captured failures (`gotoTraits` line 231 in environment-permission-test, and `createTag` line 915 in flag-tests), both with a fully-rendered page in the snapshot. Confirmed each removed call is followed by a deterministic `waitForElementVisible`/`waitFor` that already gates readiness.

### DIFF
--- a/frontend/e2e/helpers/e2e-helpers.playwright.ts
+++ b/frontend/e2e/helpers/e2e-helpers.playwright.ts
@@ -228,7 +228,6 @@ export class E2EHelpers {
     const userRow = this.page.locator('[data-test^="user-item-"]').filter({ hasText: identifier }).first();
     await userRow.waitFor({ state: 'visible', timeout: LONG_TIMEOUT });
     await userRow.locator('a').first().click();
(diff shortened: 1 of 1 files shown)


talissoncosta:

PR #7789, opened 2026-06-16
feat(Feature Lifecycle): Update API with feature lifecycle info

### LINKED ISSUE #7433: Improve API support for feature lifecycle dashboard

We want to make feature lifecycle GA, the current solution stitched endpoints together in a way that's not viable for production 

Image

There are 2 data points we need to solve here: 

 1. The summary information for the counts against each lifecycle status (e.g. `/api/v1/projects/:id/features/lifecycle-status-counts`
 2. The list of features within each status. 

I suggest that these should be 2 separate endpoints to simplify the querying on the database. 

For the second piece, I suggest that we add a `lifecycle_stage` query parameter to the `/api/v1/projects/:id/features` endpoint, which should take the values `NEW`, `PERMANENT`, etc. 

This approach (versus those defined in the alternatives below) means that the functionality is kept isolated from the rest of the application, and that we get a realtime view of the statuses when viewing the page. 

- Using current tagging system (see history of this issue for more context). While this seems like a good solution, and perhaps a future option, the concern is that it requires us to catch all events that should move a flag between lifecycles. It could result in flags becoming out of sync with their actual state. 
- We could have dedicated fields against the flags that represent lifecycle but it feels like tags make way more sense since we can filter and see those in a way that makes the most sense for the solution. This also suffers from the same issue as tags with regards to synchronisation. 

### Additional context

_No response_

emyller:

## Changes

Closes https://github.com/Flagsmith/flagsmith/issues/7433

1. Introduces a new endpoint (also to MCP) `/api/v1/environments/{environment}/feature-lifecycle-counts/`
    ```json
    {
      "NEW": 475,
      "LIVE": 734,
      "PERMANENT": 190,
      "STALE": 374,
      "NEEDS_MONITORING": 30,
      "TO_REMOVE": 197
    }
    ```
1. Adds new query params to the existing `GET /api/v1/projects/{project_pk}/features/` (list):
    - `lifecycle_stage`: one of _stages_ below; requires `environment`.
1. Adds a new attribute `lifecycle_stage ` to the above list endpoint, and also to `.../{feature_id}` (retrieve), also carrying one of _stages_ below.

**Stages** logical propositions are as follows:

| State | Condition |
| --- | --- |
| `PERMANENT` | `has_permanent_tag` |
| `LIVE` | `~has_permanent_tag AND ~has_stale_tag AND  has_code_references` |
| `NEW` | `~has_permanent_tag AND ~has_stale_tag AND ~has_code_references` |
| `STALE` | `~has_permanent_tag AND has_stale_tag AND has_code_references` |
| `NEEDS_MONITORING` | `~has_permanent_tag AND has_stale_tag AND ~has_code_references AND evaluated_recently` |
| `TO_REMOVE` | `~has_permanent_tag AND has_stale_tag AND ~has_code_references AND ~evaluated_recently` |

- New tests
- Will gradually roll out in production and collect perf data
- LLM-assisted benchmarking (does **not** represent production):

```yaml
method: synthetic data, InfluxDB analytics backend, local Docker (indicative)
baseline: main's list/detail query path (already annotates code-refs), no lifecycle

small_project:        # 1K flags, 200 in use
  influxdb_query: ~80ms
  feature_list:
    baseline_main: ~2ms
    pr_no_stage_filter: ~40ms
    pr_filtered_by_stage: ~85ms
  feature_detail:
    baseline_main: ~1ms
    pr_with_lifecycle: ~35ms
  stage_counts_endpoint: ~55ms   # new in PR, no baseline

stress_test:          # 30K flags, 600 in use — deliberate worst-case
  influxdb_query: ~120ms
  feature_list:
    baseline_main: ~8ms
    pr_no_stage_filter: ~0.12s
    pr_filtered_by_stage: ~0.45s
  feature_detail:
    baseline_main: ~1ms
    pr_with_lifecycle: ~70ms
  stage_counts_endpoint: ~0.35s
```

## How to review

Reviewing is easier if you split:

- [Part 1](https://github.com/Flagsmith/flagsmith/pull/7789/changes/fa6c72b9041485e2062e90c072b9ad5144a4cde8): InfluxDB API refactor (to enable tests) — it revealed we need to rewrite it, or maybe invest in something else.
- [Part 2](https://github.com/Flagsmith/flagsmith/pull/7789/changes/9326c3d0b9e4be5062ffa5c55d17dbe043f2d261..450a2c65f82de60aaa01e3cc4f35b044c3fe8342#diff-67dca7deb92a89f854a5a6643d1ec13127bac6c3b0d70609242e3379015ed74e): Feature Lifecycle stage counts API (to feed the count badges in the sidebar.
- [Part 3](https://github.com/Flagsmith/flagsmith/pull/7789/changes/5ed1952c0224f3b381b3dda12576d317c5ea7c8b..bc4570dbc540017c0d19b6ffd55fc99e7b1f1822): Extend the feature list and retrieve endpoints with lifecycle stage information.
- [Part 4](https://github.com/Flagsmith/flagsmith/pull/7789/changes/c54b9fdc1fa869d35da676d76091bc794b4415bb): Add filter by lifecycle stage capability to the feature list endpoint.


manos-saratsis:

**0** critical · **1** warnings across 3 agents on `feat/feature-lifecycle-api-endpoints` (76de9b7)

|---|---|---:|---:|---:|
| 🧪 Code Quality | ✅ done | 0 | 1 | 97 |
| 🛡️ Security | ✅ done | 0 | 0 | 97 |
| 📋 Compliance | ✅ done | 0 | 0 | 88 |

### Top findings
- 🔴 **Security** — Flux Query Injection via Unsanitized String Interpolation `api/app_analytics/influxdb_wrapper.py:125`
- 🟠 **Code Quality** — Hardcoded retention days constant with unclear value `api/projects/code_references/services.py:104`
- 🟠 **Compliance** — No structured audit logging for feature flag data access and mutation operations `api/features/views.py:179`
- 🟠 **Compliance** — No Audit Trail for Sensitive Data Access in Analytics Module `api/app_analytics/influxdb_wrapper.py:141`
- 🟠 **Compliance** — Sensitive InfluxDB Query Logged in Plain Text at Debug Level `api/app_analytics/influxdb_wrapper.py:132`

[View full report on Dromeas →](https://dromeas.ai/product/pr-review/8322355b-98ef-4073-8f4b-8ece22268a21)

docs/docs/integrating-with-flagsmith/_mcp-tool-catalogue.md
```diff
@@ -16,6 +16,7 @@
 | `get_feature_external_resources`           | Retrieves external resources linked to the feature flag.                                                                                                                                                                                     |
 | `get_feature_flag`                         | Retrieves detailed information about a specific feature flag.                                                                                                                                                                                |
 | `get_feature_health_events`                | Retrieves feature health monitoring events and metrics for the project.                                                                                                                                                                      |
+| `get_feature_lifecycle_counts`             | Retrieves the count of features in each lifecycle stage (new, live, stale, permanent, needs_monitoring, to_remove) for the specified environment.                                                                                            |
```
khvn26: nit: Slop

```suggestion
| `get_feature_lifecycle_counts`             | Retrieves the count of features in each lifecycle stage for the specified environment.                                                                                            |
```

api/app_analytics/types.py
```diff
@@ -27,6 +28,11 @@
 ]
 
 
+class DownsampleSize(StrEnum):
+    SHORT_TERM = "15m"
+    LONG_TERM = "1h"
```
khvn26: nit: I don't think the neat naming warrants the indirection, unless we're set to use "short term" and "long term" elsewhere, i.e. the UI. 

```suggestion
    15M = "15m"
    1H = "1h"
```

Also, I'd prefer `Literal` oven an enum.

api/features/feature_lifecycle/views.py
```diff
@@ -0,0 +1,59 @@
...
+        environment = get_object_or_404(Environment, pk=environment_pk)
+        if not request.user.has_environment_permission(VIEW_ENVIRONMENT, environment):  # type: ignore[union-attr]
+            return Response(status=403)
+
+        features = annotate_feature_queryset_with_lifecycle_stage(
+            environment.project.features.all(),
+            environment,
+        )
+
+        counts = features.values("lifecycle_stage").annotate(count=Count("pk"))  # type: ignore[misc]
```
khvn26, line 47: Do you think it would make sense to add a Prometheus histogram here to expose the query latency measurements?

api/features/serializers.py
```diff
@@ -456,10 +463,17 @@ class FeatureSerializerWithMetadata(MetadataSerializerMixin, CreateFeatureSerial
         read_only=True,
     )
 
+    # NOTE: This field is populated by `features.feature_lifecycle.services.annotate_feature_queryset_with_lifecycle_stage`.
+    lifecycle_stage = serializers.ChoiceField(
+        choices=list(LifecycleStage),
+        read_only=True,
+    )
```
khvn26, line 470: 1. Do we have an idea of how this will affect influx usage in SaaS prod?
2. Do we have an idea of the latency hit for the dashboard?

api/app_analytics/services.py
```diff
@@ -31,3 +38,63 @@ def track_usage_by_resource_host_and_environment(
...
+    environment: Environment,
+    since: datetime | None = None,
+) -> list[str]:
+    # NOTE: Neighbour buckets may bleed depending on `since`
+    buckets = FeatureEvaluationBucket.objects.filter(
+        environment_id=environment.pk,
+        bucket_size=constants.ANALYTICS_READ_BUCKET_SIZE,
+        created_at__gte=since,
+        total_count__gt=0,
+    )
```
khvn26, line 71: This is currently a seq scan. Do we need to add an index to `FeatureEvaluationBucket`?

api/features/views.py
```diff
@@ -261,8 +265,14 @@ def get_queryset(self):  # type: ignore[no-untyped-def]
...
 
         if environment_id:
-            page = self.paginate_queryset(queryset)
             self.environment = Environment.objects.get(id=environment_id)
+            if is_feature_lifecycle_enabled(project.organisation):
+                queryset = annotate_feature_queryset_with_lifecycle_stage(
+                    queryset, self.environment
+                )
+                if lifecycle_stage := query_data.get("lifecycle_stage"):
+                    queryset = queryset.filter(lifecycle_stage=lifecycle_stage)
```
khvn26, line 274: I think ideally this needs to happen _after_ `paginate_queryset`.

khvn26 (commented):
The code was pretty easy to read through, good job. Just a handful of questions/comments.

api/app_analytics/types.py
emyller: Honestly I invented short and long _terms_ because `15M =` spawns a syntax error. Perhaps too ambitious to give old things new names... Tried to improve in a9273d0b5cdbcaaf11cf647bef99da611a8ec3b0.

> Also, I'd prefer Literal oven an enum.

I'd rather keep the enum to favour typing across this patch. Let me know why you'd prefer `Literal` though — I'm curious. 🤔

api/features/serializers.py
emyller: No I don't have numbers for SaaS prod. The plan is to gradually roll this out to a few who're interested in trying it out, and collect real metrics.

On my local benchmarking, added latency is negligible (30K features + simulated usage + code references).

khvn26: Can you add the results to the PR description? This is a fairly critical path I expect us to include in our SLOs someday.

api/app_analytics/services.py
emyller: Maybe... A bit of scope creep but I added 2d75b64f326549825e8e83249c234c808d1be4a8 since it's low-cost in _effort_ — both evaluation and usage buckets are often queried for by `created_at`.

I wonder though if adding either index here may cause a sudden spike in db storage for certain self-hosted customers? 🤔 

Please cast a vote.

api/features/views.py
emyller: I think the order must be filter then paginate, or am I missing your point? 👀

api/features/feature_lifecycle/views.py
emyller: Good question, I am not Prometheus-savvy enough to share a solid opinion.

If anything, I think I'd be interested in seeing how each _portion_ contributes to the end result, in **each** of the three scenarios: count (this view), feature list, and feature detail:

1. How long it takes to obtain usage info (from Postgres or Influx)
3. How long it takes to evaluate the coderefs-annotaged + stage-annotated queryset

I'd assume HTTP metrics (URL-based) + existing telemetry in Sentry would allow looking at performance data but please advise.

api/app_analytics/types.py
khvn26: It doesn't matter as much in the AI age, but I always liked how `Literal` doesn't force you to import it in order to use it, and reap (most of) the same benefits enums provide. It's one of the cases where strict typing allows us to enforce guarantees typecheck-time while keeping runtime lean. These values are not part of any schema and don't cross domain boundaries so `Literal` works well here.

I'm happy with a9273d0 anyway.

api/features/feature_lifecycle/views.py
khvn26: > If anything, I think I'd be interested in seeing how each portion contributes to the end result

That's fair — perhaps we should invest in setting up proper datastore for traces, or figure out acceptable Sentry sample rates to make sure Sentry-based tracing is always on for us.

> I'd assume HTTP metrics (URL-based) + existing telemetry in Sentry would allow looking at performance data

In current Sentry setup, it'll be a bit hard, if at all possible, to filter out dashboard requests that include feature lifecycle enrichment. However, the p95/p99 HTTP route metrics will be a good proxy for this, IMO. I'm happy to proceed without a dedicated metric for now, as long as we keep an eye on the performance proactively 👍

api/app_analytics/services.py
khvn26: I don't think this index will pose an improvement — `(environment_id, bucket_size, created_at)` oughta do it, IMO.

It slipped my mind though that we have a default retention of 90 days so neither my perf concern nor your storage concern are pressing. My vote is to revert 2d75b64, and revisit this when (if) this becomes a real problem.

api/features/views.py
khvn26: Right, good call — we'd get uneven pages otherwise.

api/features/feature_lifecycle/views.py
emyller: Please do suggest any interesting metrics if you think of any — I'd be happy to introduce them. :)

Also, setting sample rate to 1.0 on a specific URL seems possible.

api/features/serializers.py
emyller: Done!

emyller: I suppose it's obvious to expect we'll need to learn from observation and tweak as needed. I deem current numbers are not perfect, but I'd rather not try to optimise early.

api/app_analytics/services.py
emyller: fc975b18d87faf6b9644db39418c5e88614e9b84 ✅

khvn26 (approved):

PR #7800, opened 2026-06-16
docs(MCP): Document self-hosting the MCP server

khvn26:

## Changes

Closes https://github.com/Flagsmith/flagsmith-private/issues/155.

Preview: [Self-hosting the MCP Server](https://docs-git-docs-mcp-self-hosting-flagsmith.vercel.app/deployment-self-hosting/mcp-server)

Adds a self-hosting guide for the MCP server under Deployment and Self-hosting, modelled on the Edge Proxy page. It covers:

- Running the `flagsmith/flagsmith-mcp` container over HTTP (Docker CLI and Docker Compose).
- Running it over stdio as a local subprocess via `uvx`, pointed at a self-hosted API.
- The full set of configuration environment variables.
- The three authentication modes (OAuth, per-request API key, static API key).
- Bare-URL routing, the health check, and observability hooks.

Also links to the new page from the self-hosting note on the MCP Server integration page.

Docs-only change. Built the Docusaurus site locally, confirmed the page renders and all cross-linked pages resolve, and ran `npx prettier --write` over both files.


gagantrivedi (commented):
## Code Review — docs(MCP): Document self-hosting the MCP server

Good addition overall. The structure mirrors the Edge Proxy page well and all three auth modes are clearly explained. A few things worth addressing before merge:

---

### Issues

**`uvx` command pins to `@main` (moving target)**

```bash
uvx --from "git+https://github.com/Flagsmith/flagsmith.git@main#subdirectory=mcp" flagsmith-mcp
```

`@main` will silently pull breaking changes. This should pin to a versioned tag (e.g. `@v2.151.0`) or, preferably, a published PyPI package once one is available. The same issue applies to the `mcp.json` snippet below it. If there is intentionally no published package yet, a note explaining that — and advising users to pin to the tag that matches their Flagsmith version — would prevent confusion.

---

**Docker Compose YAML uses single-space indentation**

The compose block uses 1-space indentation throughout, which is valid YAML but non-standard and inconsistent with every other compose snippet in the docs. Standard 2-space indentation would be clearer and match what users expect to copy-paste.

---

**No mention of reverse proxy / TLS for the HTTP transport**

The Docker examples expose port 8000 over plain HTTP, but `MCP_SERVER_URL` is set to `https://mcp.flagsmith.example.com`. There is no guidance on how TLS termination is achieved (e.g. "place a reverse proxy in front; the container itself serves plain HTTP"). Without this, the gap between the example and reality is likely to confuse self-hosters.

---

**Docker Compose: internal vs external `FLAGSMITH_API_URL`**

In a Compose deployment where the MCP server sits alongside the Flagsmith API, `FLAGSMITH_API_URL` should point at the internal service name (e.g. `http://api:8000`), not the public domain. The current example uses the external URL in both the CLI and Compose tabs, which will cause DNS failures for users following it literally. Consider either adding a note or showing the internal URL in the Compose tab.

---

**OAuth section is vague on prerequisites**

> OAuth requires a Flagsmith deployment that exposes the MCP authorisation-server endpoints. If yours does not, use an API key instead.

A self-hoster reading this has no way to know whether their deployment "exposes" those endpoints or how to enable them. A one-liner pointing to the relevant configuration option, minimum version, or feature flag would make this actionable.

---

### Minor / Nits

- `import Tabs from '@theme/Tabs'; import TabItem from '@theme/TabItem';` — two imports on one line. Other MDX pages in this repo put each import on its own line; worth keeping consistent.
- The `sidebar_position: 4` is fine as long as it doesn't collide with another page — worth a quick check in the self-hosting sidebar.

---

### What looks good

- Structure and section ordering are clear and logical.
- The env-var reference table is comprehensive and the default/description columns are useful.
- The cross-link update in `mcp-server.mdx` is clean — the new wording is more precise than the old.
- Health-check and observability sections are a nice touch that most docs skip.


---
_Generated by [Claude Code](https://claude.ai/code)_

docs/docs/deployment-self-hosting/mcp-server.md
```diff
@@ -0,0 +1,158 @@
...
+Authorization: Api-Key <your-api-key>
+```
+
+The server forwards this header to the Flagsmith API unchanged. This suits CI jobs and headless agents.
+
+### Static API key (single shared credential)
+
+Setting `FLAGSMITH_API_TOKEN` makes the server use that key for every request, regardless of which client connects. This
+disables OAuth and is the only option for the `stdio` transport, which has no inbound request to forward a credential
+from. Avoid it for shared HTTP deployments, since every client then acts with the same permissions.
```
matthewelwell: nit: arguably the lack of audit is more of an issue than the permissions.

matthewelwell (dismissed):
Approved with a minor nit.

docs/docs/deployment-self-hosting/mcp-server.md
khvn26: Addressed in 6d3c52941.

matthewelwell (approved):

emyller (approved):
Incomplete non-blocking review.
Looks great overall assuming all MCP installation examples were tested.

PR #7805, opened 2026-06-17
fix(experiments): resolve environment-level variant allocations

Zaimwa9:

## Changes

Fix experiment list serialiser to return environment-level variant allocations instead of feature-level defaults. Order experiments by newest first.

- `ExperimentFeatureSerializer.get_multivariate_options` looks up the environment's feature state and overlays the actual `percentage_allocation` per multivariate option
- `order_by("-created_at")` on the experiments queryset

Manually — created an experiment whose environment-level allocation differs from the feature default and confirmed the serialiser returns the correct percentages. Verified the list endpoint returns newest experiments first.


api/tests/unit/experimentation/test_experiment_views.py
```diff
@@ -1373,3 +1375,85 @@ def test_post__concurrent_create_race__returns_409(
...
+    response = admin_client_new.get(_detail_url(environment, experiment))
+
+    # Then
+    assert response.status_code == status.HTTP_200_OK
+    options = response.json()["feature"]["multivariate_options"]
+    returned_allocs = sorted(o["default_percentage_allocation"] for o in options)
+    assert returned_allocs == sorted(env_allocations)
+
+
+@pytest.mark.django_db()
```
gagantrivedi: nit: don't think we need `django_db mark`

api/experimentation/serializers.py
```diff
@@ -285,15 +285,45 @@ def create(self, validated_data: dict[str, Any]) -> Experiment:
...
         read_only_fields = fields
 
+    def get_multivariate_options(self, feature: Feature) -> list[dict[str, Any]]:
+        options = NestedMultivariateFeatureOptionSerializer(
+            feature.multivariate_options.all(), many=True
+        ).data
+
+        environment: Environment | None = self.context.get("environment")
+        if not environment:
+            return options  # type: ignore[return-value]
```
gagantrivedi: should we raise if environmetn is not part of context? or is this a valid use case?

api/experimentation/serializers.py
```diff
@@ -285,15 +285,45 @@ def create(self, validated_data: dict[str, Any]) -> Experiment:
...
+        env_state = (
+            feature.feature_states.filter(
+                environment=environment,
+                identity__isnull=True,
+                feature_segment__isnull=True,
+            )
+            .order_by("-live_from", "-version")
+            .first()
+        )
+        if not env_state:
```
gagantrivedi, line 315: can this be none ever?

api/experimentation/serializers.py
```diff
@@ -285,15 +285,45 @@ def create(self, validated_data: dict[str, Any]) -> Experiment:
...
+
+        environment: Environment | None = self.context.get("environment")
+        if not environment:
+            return options  # type: ignore[return-value]
+
+        env_state = (
+            feature.feature_states.filter(
+                environment=environment,
+                identity__isnull=True,
+                feature_segment__isnull=True,
```
gagantrivedi, line 310: this also will need updating based on our plan of using segment override?

gagantrivedi (changes requested):

api/experimentation/serializers.py
Zaimwa9: Yes 👍

api/experimentation/serializers.py
Zaimwa9: In production it shouldn't happen. It was defensive. I'll raise instead just in case

gagantrivedi (approved):

PR #7807, opened 2026-06-17
fix(features): open feature slideout when deep link targets an off-page feature

### LINKED ISSUE #7652: Deep link to feature slideout fails when feature isn't on page 1 of features list
## Summary

Opening a URL of the form `/project/<project>/environment/<environment>/features?feature=<feature-id>&tab=<tab>` does not open the feature slideout if the deep-linked feature is not present in the currently rendered page of the features list. The page renders the paginated feature list as if no deep-link were supplied, and no error is surfaced (nothing reaches Sentry).

This is most easily reproduced on projects with more than 50 features, since the list defaults to `page_size=50` sorted by name ascending, and any feature alphabetically past page 1 is invisible to the deep-link handler.

## Steps to reproduce

1. Open a project with more than 50 features.
2. Pick a feature whose name sorts past the first 50 entries.
3. Open `/project/<project>/environment/<environment>/features?feature=<feature-id>&tab=segment-overrides` directly (e.g. paste in the address bar or follow a shared link).

## Expected

The features list opens and the side modal for the targeted feature opens, scrolled to the requested tab (e.g. Segment Overrides).

## Actual

The features list opens but the side modal never opens. No error is logged. The only network activity related to the feature is the standard `GET /api/v1/projects/<project>/features/?page=1&page_size=50&sort_direction=ASC&sort_field=name&...` call; there is no direct fetch of the target feature.

## Root cause

The deep-link is handled per-row inside `FeatureRow`, not at the page level. Each `FeatureRow` instance runs:

```tsx
// frontend/web/components/feature-summary/FeatureRow.tsx:113-124
useEffect(() => {
  const { feature } = Utils.fromParam()
  const { id } = projectFlag
  const isModalOpen =
    !!document?.getElementsByClassName('create-feature-modal')?.length ||
    !!document?.getElementsByClassName('create-experiment-modal')?.length
  if (`${id}` === feature && !isModalOpen) {
    editFeature()
  }
}, [environmentFlags, projectFlag])
```

If the targeted feature is not in the loaded page, no `FeatureRow` instance ever mounts for it, so the effect never fires and the slideout is never opened. There is no top-level handler in `FeaturesPage` that fetches the targeted feature directly when it is missing from the loaded page.

## Claude Suggested fix

Hoist the deep-link handling into `FeaturesPage` (`frontend/web/components/pages/features/FeaturesPage.tsx`). On mount, if `Utils.fromParam().feature` is set:

1. If the feature is already present in the loaded list, open the slideout via the existing path.
2. Otherwise, fetch the feature directly (`GET /api/v1/projects/<project>/features/<feature>/`) and open the slideout with the result, honouring the `tab` query param.

This makes the deep-link behaviour independent of pagination, sort order, and tag/archive filters. The per-row `useEffect` in `FeatureRow` can be removed once the page-level handler covers both cases.

## Notes

- No Sentry error is produced, so this failure is silent for users and invisible in error monitoring.
- The same bug likely affects any future deep-link that targets a single feature regardless of which tab is requested; this is not specific to `tab=segment-overrides`.

adamvialpando:
### Problem

Opening `/project/<p>/environment/<e>/features?feature=<id>&tab=<tab>` did not open the feature slideout when the targeted feature was not on the currently rendered page of the paginated list. Deep-linking was handled per-row inside `FeatureRow`, so if no row mounted for the feature (e.g. it sorts past `page_size=50`), the effect never fired and the slideout never opened. The failure was silent, with nothing reaching Sentry.

### Fix

Detect, at the page level, when the `?feature=` target is missing from the loaded page and fetch it directly, then render a hidden `FeatureRow` so the existing deep-link effect opens the slideout (honouring the `tab` param).

- `shouldDeepFetchFeature` decides whether a direct fetch is needed (no-ops when there is no param, the list has not loaded, the param is not a valid id, or the feature is already on the page).
- `useDeepLinkedFeature` fetches the project flag and its environment feature state when required.
- `FeaturesPage` renders a hidden row for the fetched feature; the on-page path is unchanged, so there is no double-open.

Behavior is now independent of pagination, sort order, and tag/archive filters.

### Testing

- Unit tests for `shouldDeepFetchFeature` and `pickEnvironmentFlag` (8 cases) — passing.
- E2E test covering a deep link to an off-page feature.
- Typecheck: the changed files introduce no new type errors.

### DIFF
--- a/frontend/e2e/tests/deep-link-test.pw.ts
+++ b/frontend/e2e/tests/deep-link-test.pw.ts
@@ -44,51 +44,71 @@ test.describe('Deep link to feature slideout', () => {
     expect(environment).toBeTruthy()
 
     log(`Create ${FEATURE_COUNT} features`)
-    for (let i = 0; i < FEATURE_COUNT; i++) {
-      const res = await request.post(
-        `${api}projects/${project.id}/features/`,
-        {
-          data: { name: `${FEATURE_PREFIX}${String(i).padStart(3, '0')}` },
-          headers,
-        },
+    // Unique names per run so the flakiness check (`E2E_REPEAT` / `/e2e N`) does
+    // not collide with features created by a previous iteration.
+    const runId = Date.now()
+    const featureName = (i: number) =>
+      `${FEATURE_PREFIX}${runId}_${String(i).padStart(3, '0')}`
+    const createdFeatureIds: number[] = []
+
+    try {
+      const created = await Promise.all(
+        Array.from({ length: FEATURE_COUNT }, (_, i) =>
+          request.post(`${api}projects/${project.id}/features/`, {
+            data: { name: featureName(i) },
+            headers,
+          }),
+        ),
       )
-      expect(res.ok()).toBeTruthy()
+      for (const res of created) {
+        expect(res.ok()).toBeTruthy()
+        createdFeatureIds.push((await res.json()).id)
+      }
+
+      // The list renders sorted by name ascending, so page 1 holds the first
+      // PAGE_SIZE features and a feature on page 2 never mounts a row on page 1.
+      const listUrl = (pageNumber: number) =>
+        `${api}projects/${project.id}/features/?environment=${environment.id}&page=${pageNumber}&page_size=${PAGE_SIZE}&sort_field=name&sort_direction=ASC`
+      const page1 = await (await request.get(listUrl(1), { headers })).json()
+      const page2 = await (await request.get(listUrl(2), { headers })).json()
+      const onPageFeature = page1.results[0] as { id: number; name: string }
+      const offPageFeature = page2.results[0] as { id: number; name: string }
+      expect(onPageFeature).toBeTruthy()
+      expect(offPageFeature).toBeTruthy()
+      log(`On-page: ${onPageFeature.name}, off-page: ${offPageFeature.name}`)
+
+      await login(E2E_USER, PASSWORD)
+      const featuresPath = `/project/${project.id}/environment/${environment.api_key}/features`
+      const slideout = page.locator('.create-feature-modal')
+
+      // When/Then - this is the #7652 regression: a deep link to a feature that
+      // is NOT on the first page previously rendered the list without opening
+      // any modal, because the deep-link handler only fired for mounted rows.
+      await page.goto(`${featuresPath}?feature=${offPageFeature.id}&tab=value`)
+      await expect(slideout).toBeVisible({ timeout: LONG_TIMEOUT })
+      await expect(slideout).toContainText(offPageFeature.name)
+
+      // And - the existing on-page deep link still works (no regression). A
+      // fresh navigation reloads the page, dismissing the previous slideout.
+      await page.goto(`${featuresPath}?feature=${onPageFeature.id}&tab=value`)
+      await expect(slideout).toBeVisible({ timeout: LONG_TIMEOUT })
+      await expect(slideout).toContainText(onPageFeature.name)
+
+      // And - an unknown feature id degrades gracefully (no slideout, no crash).
+      await page.goto(`${featuresPath}?feature=999999999&tab=value`)
+      await expect(page.locator('[data-test="features-page"]')).toBeVisible({
+        timeout: LONG_TIMEOUT,
+      })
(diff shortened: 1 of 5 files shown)


frontend/e2e/tests/deep-link-test.pw.ts
```diff
@@ -0,0 +1,94 @@
...
+    for (let i = 0; i < FEATURE_COUNT; i++) {
+      const res = await request.post(
+        `${api}projects/${project.id}/features/`,
+        {
+          data: { name: `${FEATURE_PREFIX}${String(i).padStart(3, '0')}` },
+          headers,
+        },
+      )
+      expect(res.ok()).toBeTruthy()
+    }
```
talissoncosta: Fixed names make this fail on iteration 2 under `E2E_REPEAT` / `/e2e N` (our flakiness check) with duplicate names. Could we make them unique per run, and create in parallel while here?

```ts
const runId = Date.now()
const featureName = (i: number) =>
  `${FEATURE_PREFIX}${runId}_${String(i).padStart(3, '0')}`

await Promise.all(
  Array.from({ length: FEATURE_COUNT }, (_, i) =>
    request
      .post(`${api}projects/${project.id}/features/`, {
        data: { name: featureName(i) },
        headers,
      })
      .then((res) => expect(res.ok()).toBeTruthy()),
  ),
)
```

frontend/web/components/pages/features/FeaturesPage.tsx
```diff
@@ -378,6 +403,21 @@ const FeaturesPage: FC<FeaturesPageProps> = ({
 
             <FormGroup className='mb-4'>{renderFeaturesList()}</FormGroup>
 
+            {deepLinkedFeature && (
+              <div style={{ display: 'none' }} data-test='deep-linked-feature'>
```
talissoncosta: Can you use the utility class instead of inline style?

````suggestion
              <div className='d-none' data-test='deep-linked-feature'>
````

talissoncosta (commented):
Looks good @adamvialpando! Thanks for handling this. Left a few minor suggestions, nothing blocking, but it'd be great if you could address them before merging.

manos-saratsis:

**0** critical · **4** warnings across 3 agents on `fix/7652-deep-link-offpage-feature` (aa9fe68)

|---|---|---:|---:|---:|
| 🧪 Code Quality | ✅ done | 0 | 4 | 88 |
| 🛡️ Security | ✅ done | 0 | 0 | 99 |
| 📋 Compliance | ✅ done | 0 | 0 | 98 |

### Top findings
- 🟠 **Code Quality** — Legacy Flux store coupling creates fragile two-way synchronization `frontend/web/components/pages/features/FeaturesPage.tsx:99`
- 🟠 **Code Quality** — FeaturesPage mixes data-fetching, legacy store sync, and rendering responsibilities `frontend/web/components/pages/features/FeaturesPage.tsx:99`
- 🟠 **Code Quality** — Component relies on implicit global `openModal` and `AppActions` singletons with no injection seam `frontend/web/components/pages/features/FeaturesPage.tsx:187`
- 🟠 **Code Quality** — Inline mutation event handlers use async callbacks without error handling `frontend/web/components/pages/features/FeaturesPage.tsx:293`
- 🟠 **Compliance** — Page tracking with persistent storage invoked without visible consent gate `frontend/web/components/pages/features/FeaturesPage.tsx:177`

[View full report on Dromeas →](https://dromeas.ai/product/pr-review/c33fc822-3fdf-454f-baa3-ba30c4a471aa)

Zaimwa9:
Is #7839  a duplicate ?

frontend/e2e/tests/deep-link-test.pw.ts
adamvialpando: Ahh great catch! Updated that and added cleanup in a finally so reruns delete the features as well.

frontend/web/components/pages/features/FeaturesPage.tsx
adamvialpando: Looks like a better idea to me. swapped it for <div className='d-none'>

talissoncosta (approved):
Looks good to me! Thanks for handling that @adamvialpando !

PR #7808, opened 2026-06-18
fix(Local): Fix `make install`

emyller:

## Changes

Require uv, and fix `make install` in the api directory:

- `python` unavailable in CLI: on MacOS, only `python3` is linked by default.
- `python -m pip` is broken os OS-level Python installation. See [why](https://github.com/pypa/pip/pull/11780).

```sh
$ git clone git@github.com:Flagsmith/flagsmith.git
$ cd flagsmith/api
$ make install  # works
```

On MacOS, and in Linux too.


khvn26 (dismissed):

.github/actions/install-uv/action.yml
```diff
@@ -5,16 +5,19 @@ description: |
...
 
 runs:
   using: composite
   steps:
-    - uses: astral-sh/setup-uv@v6
+    - name: Install uv
+      uses: astral-sh/setup-uv@v6
       with:
-        version-file: api/pyproject.toml
+        version-file: ${{ inputs.working-directory }}/pyproject.toml
```
khvn26, line 20: Is `${{ defaults.run.working-directory }}` accessible from here?

mcp/pyproject.toml
```diff
@@ -34,6 +34,9 @@ dev = [
 requires = ["uv_build>=0.11.0,<0.12.0"]
 build-backend = "uv_build"
 
+[tool.uv]
+required-version = ">=0.11.18"  # Ensure this matches the version in .pre-commit-config.yaml
```
khvn26, line 38: I wonder if we can exclude uv hook from pre-commit.ci autoupdate and write a Renovate config to bump all three versions at the same time.

Dockerfile
```diff
@@ -85,6 +85,7 @@ RUN cd frontend && npm run bundle
 
 # * build-python
 FROM wolfi-base AS build-python
+COPY --from=ghcr.io/astral-sh/uv /uv /uvx /bin/
```
khvn26, line 88: nit: This pulls latest uv, but we do have version constraints for it defined. Given that we don't define an upper bound, I think we can live with it.

Ideally, I'd love to see a renovate-managed tag reference like [here](https://github.com/Flagsmith/github-webhook-handler/blob/7dd460794f354fe794d7eb59eb1f1bd54678fa19/Dockerfile#L35-L36).

api/README.md
```diff
@@ -3,11 +3,11 @@
...
 The project assumes the following tools installed:
-- [Python](https://www.python.org/downloads/). Any version allowed by `requires-python` in `pyproject.toml` is supported.
-- [GNU Make](https://www.gnu.org/software/make/).
-- Docker or a compatible tool like [Podman](https://podman.io/). We recommend [OrbStack](https://orbstack.dev/) for macOS.
+- [uv](https://github.com/astral-sh/uv)
+- [GNU Make](https://www.gnu.org/software/make/)
+- [Docker](https://www.docker.com/) — or compatible software such as [OrbStack](https://orbstack.dev/) or [Podman](https://podman.io/).
 
-To install dev dependencies, run `make install`. Only Flagsmith maintainers can run `uv lock` due to private dependencies.
+To install dev dependencies, run `make install opts='--extra dev'`. Only Flagsmith maintainers can run `uv lock` due to private dependencies.
```
khvn26, line 10: Nice 👍

khvn26 (commented):
A couple questions.

mcp/pyproject.toml
emyller: Yup, been there, warrants a separate PR. 😉

.github/actions/install-uv/action.yml
emyller: That'd be nice, though it looks like it's not present in GHA [context](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context).

```suggestion
        version-file: ./pyproject.toml
```

Would you think the above look good? It think it looks a bit cryptic but don't have a strong opinion.

khvn26: I'm just a bit sad we have to pass cwd to an action explicitly.

It seems [the design](https://github.com/actions/runner/blob/main/docs/adrs/0549-composite-run-steps.md#shell-and-working-directory) suggests we don't have to do that.

khvn26: Or maybe I'm misreading:

> Note, `defaults` in a workflow only apply to run steps not "uses" steps

khvn26: Yep, looks like this is the canonical way to pass cwd.

mcp/pyproject.toml
khvn26: Agreed.

khvn26 (approved):
LGTM

Dockerfile
emyller: Agreed. I intentionally left this unbounded as `pyproject.toml` now uses a `>=` constraint.

Added uv version management to https://github.com/Flagsmith/flagsmith/issues/7826, if you don't mind.

PR #7831, opened 2026-06-19
feat(usage): push hourly usage snapshots to Control Plane

### LINKED ISSUE #7673: Push hourly usage snapshots to Control Plane
## Summary

Each Flagsmith Instance (Private Cloud or Self-Hosted) should report its own usage to the central Control Plane once an hour. The Control Plane already accepts these reports at `POST /v1/usage`; this issue is the sending half that
lives inside Flagsmith.

The report is a "usage snapshot": a point-in-time count of seats, projects, and API calls. Flagsmith staff use it to compare what a customer actually uses against what their licence entitles them to.

## How it works

A recurring task on the **task processor** runs hourly. For each licensed organisation on this Instance, the run:

1. Builds one usage snapshot from that organisation's own data.
2. Reads the licence signature (`OrganisationLicence.signature`).
3. POSTs the snapshot to the Control Plane with the signature as a bearer token.

## What goes in a snapshot

The Control Plane accepts **one unified payload shape regardless of deployment type** (`PushRequest` in `apps/ingestion/schemas.py`). Every field below is **required on every snapshot** — unknown extra fields are rejected (`extra="forbid"`). 

- `timestamp` — the hour the snapshot covers, timezone-aware (UTC). Send the most recent *complete* hour. Rejected if it is more than 5 minutes in the future (clock-skew tolerance).
- `seat_count` — total users in the organisation, dormant included (`organisation.num_seats`). Range `0`–`10^9`.
- `api_call_total` — total SDK API calls this period. Range `0`–`10^15`.
- `api_call_breakdown` — per-type counts, all required: `flags`, `identities`, `traits`, `environment_documents`. Each `0`–`10^15`.
- `project_count` — non-deleted projects (`organisation.projects.filter(deleted_at__isnull=True).count()`). Range `0`–`10^6`.
- `instance_version` — the Flagsmith version running, as plain semver (e.g. `2.142.3`). No `v` prefix, no `latest`/`unknown`. Length `1`–`64`.
- `project_usage` — per-project breakdown: a list of `{ project_id, api_call_count }`. **No `project_name`.** `project_id` is `1`–`10^12`, `api_call_count` is `0`–`10^12`. Max 5000 rows, and `project_id` must be unique within the list.

A Self-Hosted Instance with no API-call analytics still sends the same shape — it just reports zeros for `api_call_total`, every `api_call_breakdown` entry, and an empty `project_usage`.

All API-call counts come from `app_analytics` — reuse `analytics_db_service.get_usage_data()`, which already supports per-type and per-project filtering (the same source as the `GET /organisations/{id}/usage-data/` endpoint).

## Sending it

- **Endpoint:** `POST /v1/usage` on the Control Plane (base URL from config).
- **Auth:** `Authorization: Bearer <base64url(OrganisationLicence.signature)>`.
- **Rate limit:** 20 requests/hour per authenticated licence.
- **Expected responses:**
  - `201` — created (new snapshot stored).
  - `200` — idempotent duplicate replay for the same `(customer, timestamp)`. Safe to ignore.
  - `400` — payload validation failed. Log it; the snapshot was malformed.
  - `401` — unknown or invalid signature. Log it.
  - `403` — licence revoked, licence expired, or customer deactivated. Log it; do not retry.
  - `429` — rate limited. Back off; do not retry aggressively.
  - `500` — internal/server error. Retry on the next scheduled run, not in a tight loop.

Re-sending the same hour is harmless — the Control Plane drops duplicates by `(customer, timestamp)` and answers `200`. A missed run that catches up next hour is fine.

## Configuration

New setting, optional so existing deployments stay silent until turned on:

- `CONTROL_PLANE_URL` — base URL. If unset, the task is a no-op.

The licence signature is read from the database (`OrganisationLicence`), so it needs no new setting.

If the Control Plane URL is missing, skip the run quietly.

## Acceptance criteria

- [x] A recurring task is registered on the task processor and runs hourly.
- [x] It builds a correct snapshot per licensed organisation.
- [x] It authenticates with the licence signature and POSTs to `/v1/usage`.
- [x] It reuses `app_analytics` for API-call counts (no new counting logic).
- [x] It handles `201`/`200`/`400`/`401`/`403`/`429`/`500` without crashing and logs failures.
- [x] It is a no-op when the Control Plane URL is not configured.
- [x] `instance_version` is sent as clean semver.

germangarces:

## Changes

Closes #7673 

**refactor(app_analytics): extract bucket query and aggregation helpers**

Pull the shared queryset build and per-resource aggregation out of `get_usage_data_from_local_db` so a windowed reader can reuse them. No behaviour change.

**feat(usage): push hourly usage snapshots to Control Plane**

Recurring task reports each licensed organisation's data usage to the Control Plane once an hour, authenticated with the licence signature. No-op until `CONTROL_PLANE_URL` is set and a licensed organisation exists.

Reads API-call counts at hourly resolution via a new `app_analytics` window selector; the daily billing path is left untouched.

Tested end-to-end locally.

Quick check (No control plane needed):

1. `make docker-up django-migrate`
2. `LOG_LEVEL=DEBUG make django-shell`
3. In the shell:
```python
 from unittest.mock import patch, Mock
 from organisations.usage_reporting.services import push_snapshot
 from organisations.usage_reporting.mappers import map_organisation_to_usage_snapshot
 org = Organisation.objects.first() or Organisation.objects.create(name="Test")
 snap = map_organisation_to_usage_snapshot(org)
 with patch("organisations.usage_reporting.services.requests.post",
            return_value=Mock(ok=True, status_code=201)):
     push_snapshot(base_url="http://x", snapshot=snap, signature="dummy")
```

You should see something like:
```
[info] snapshot.pushed [usage_reporting] status_code=201
```

### DIFF
--- a/api/app_analytics/analytics_db_service.py
+++ b/api/app_analytics/analytics_db_service.py
@@ -5,7 +5,7 @@ import structlog
 from common.core.utils import is_saas, using_database_replica
 from dateutil.relativedelta import relativedelta
 from django.conf import settings
-from django.db.models import Q, Sum
+from django.db.models import Q, QuerySet, Sum
 from django.utils import timezone
 from rest_framework.exceptions import NotFound
 
@@ -78,19 +78,12 @@ def get_usage_data(
     return []
 
 
-def get_usage_data_from_local_db(
+def _get_api_usage_bucket_qs(
     organisation: Organisation,
     environment_id: int | None = None,
     project_id: int | None = None,
-    date_start: datetime | None = None,
-    date_stop: datetime | None = None,
     labels_filter: Labels | None = None,
-) -> list[UsageData]:
-    if date_start is None:
-        date_start = timezone.now() - timedelta(days=30)
-    if date_stop is None:
-        date_stop = timezone.now()
-
+) -> QuerySet[APIUsageBucket]:
     qs = APIUsageBucket.objects.filter(
         environment_id__in=_get_environment_ids_for_org(organisation),
         bucket_size=constants.ANALYTICS_READ_BUCKET_SIZE,
@@ -110,17 +103,96 @@ def get_usage_data_from_local_db(
     if labels_filter:
         qs = qs.filter(labels__contains=labels_filter)
 
-    qs = (
-        qs.filter(  # type: ignore[assignment]
-            created_at__date__lte=date_stop,
-            created_at__date__gt=date_start,
-        )
-        .order_by("created_at__date")
+    return qs
+
+
+def _aggregate_buckets(qs: QuerySet[APIUsageBucket]) -> list[UsageData]:
+    annotated = (
+        qs.order_by("created_at__date")
         .values("created_at__date", "resource", "labels")
         .annotate(count=Sum("total_count"))
     )
+    return map_annotated_api_usage_buckets_to_usage_data(annotated)  # type: ignore[arg-type]
 
-    return map_annotated_api_usage_buckets_to_usage_data(qs)
+
+def get_usage_data_from_local_db(
+    organisation: Organisation,
+    environment_id: int | None = None,
+    project_id: int | None = None,
+    date_start: datetime | None = None,
+    date_stop: datetime | None = None,
+    labels_filter: Labels | None = None,
+) -> list[UsageData]:
+    if date_start is None:
+        date_start = timezone.now() - timedelta(days=30)
+    if date_stop is None:
+        date_stop = timezone.now()
+
+    qs = _get_api_usage_bucket_qs(
+        organisation,
+        environment_id=environment_id,
+        project_id=project_id,
+        labels_filter=labels_filter,
+    ).filter(
+        created_at__date__lte=date_stop,
+        created_at__date__gt=date_start,
+    )
+    return _aggregate_buckets(qs)
+
+
+def get_usage_data_from_local_db_for_window(
+    organisation: Organisation,
+    date_start: datetime,
+    date_stop: datetime,
+    project_id: int | None = None,
+) -> list[UsageData]:
+    # Filters on the full timestamp (not ``created_at__date``) so that sub-day
+    # windows such as a single hour are honoured. The daily ``get_usage_data``
+    # path keeps its date-truncated filter to leave the billing endpoint
+    # unchanged.
+    qs = _get_api_usage_bucket_qs(
+        organisation,
+        project_id=project_id,
+    ).filter(
+        created_at__gte=date_start,
+        created_at__lt=date_stop,
+    )
+    return _aggregate_buckets(qs)
+
+
+def get_usage_data_for_window(
+    organisation: Organisation,
+    date_start: datetime,
+    date_stop: datetime,
+    project_id: int | None = None,
+) -> list[UsageData]:
+    """
+    Return per-resource API-call counts for an explicit datetime window.
+
+    Mirrors ``get_usage_data`` but at full datetime
+    resolution, so callers can request a window finer than a day.
+    """
+    if settings.USE_POSTGRES_FOR_ANALYTICS:
+        return get_usage_data_from_local_db_for_window(
+            organisation,
+            date_start=date_start,
+            date_stop=date_stop,
+            project_id=project_id,
+        )
+
+    if settings.INFLUXDB_TOKEN:
+        return get_usage_data_from_influxdb(
+            organisation_id=organisation.id,
+            project_id=project_id,
+            date_start=date_start,
+            date_stop=date_stop,
+        )
+
+    logger.warning(
+        "no-analytics-database-configured",
+        details=constants.NO_ANALYTICS_DATABASE_CONFIGURED_WARNING,
+    )
+    return []
 
 
 def get_top_organisations_from_local_db(
--- a/api/app_analytics/analytics_db_service.py
+++ b/api/app_analytics/analytics_db_service.py
@@ -146,10 +146,6 @@ def get_usage_data_from_local_db_for_window(
     date_stop: datetime,
     project_id: int | None = None,
 ) -> list[UsageData]:
-    # Filters on the full timestamp (not ``created_at__date``) so that sub-day
-    # windows such as a single hour are honoured. The daily ``get_usage_data``
-    # path keeps its date-truncated filter to leave the billing endpoint
-    # unchanged.
     qs = _get_api_usage_bucket_qs(
         organisation,
         project_id=project_id,
--- a/api/organisations/usage_reporting/mappers.py
+++ b/api/organisations/usage_reporting/mappers.py
@@ -18,12 +18,6 @@ from projects.models import Project
 MAX_PROJECT_USAGE_ROWS = 5_000
 
 
-def _complete_hour_window() -> tuple[datetime, datetime]:
-    hour_end = timezone.now().replace(minute=0, second=0, microsecond=0)
-    hour_start = hour_end - timedelta(hours=1)
-    return hour_start, hour_end
-
-
 def _total_api_calls(usage_data: list[UsageData]) -> int:
     return sum(
         data.flags + data.identities + data.traits + data.environment_document
@@ -74,7 +68,8 @@ def _project_usage(
 
 
 def map_organisation_to_usage_snapshot(organisation: Organisation) -> UsageSnapshot:
-    hour_start, hour_end = _complete_hour_window()
+    hour_end = timezone.now().replace(minute=0, second=0, microsecond=0)
+    hour_start = hour_end - timedelta(hours=1)
     api_call_total, api_call_breakdown = _aggregate(
         get_usage_data_for_window(organisation, hour_start, hour_end)
     )
--- a/api/organisations/usage_reporting/mappers.py
+++ b/api/organisations/usage_reporting/mappers.py
@@ -32,13 +32,7 @@ def _aggregate(usage_data: list[UsageData]) -> tuple[int, ApiCallBreakdown]:
         traits=sum(data.traits for data in usage_data),
         environment_documents=sum(data.environment_document for data in usage_data),
     )
-    api_call_total = (
-        breakdown.flags
-        + breakdown.identities
-        + breakdown.traits
-        + breakdown.environment_documents
-    )
-    return api_call_total, breakdown
+    return _total_api_calls(usage_data), breakdown
 
 
 def _project_usage(
--- a/api/organisations/usage_reporting/mappers.py
+++ b/api/organisations/usage_reporting/mappers.py
@@ -1,0 +1,95 @@
+from datetime import datetime, timedelta
+
+from common.core.utils import get_version
+from django.db.models import QuerySet
+from django.utils import timezone
+
+from app_analytics.analytics_db_service import get_usage_data_for_window
+from app_analytics.dataclasses import UsageData
+from organisations.models import Organisation
+from organisations.usage_reporting.dataclasses import (
+    ApiCallBreakdown,
+    ProjectUsage,
+    UsageSnapshot,
+)
+from projects.models import Project
+
+# The Control Plane rejects payloads with more than this many project_usage rows.
+MAX_PROJECT_USAGE_ROWS = 5_000
+
+
+def _complete_hour_window() -> tuple[datetime, datetime]:
+    hour_end = timezone.now().replace(minute=0, second=0, microsecond=0)
+    hour_start = hour_end - timedelta(hours=1)
+    return hour_start, hour_end
+
+
+def _total_api_calls(usage_data: list[UsageData]) -> int:
+    return sum(
+        data.flags + data.identities + data.traits + data.environment_document
+        for data in usage_data
+    )
+
+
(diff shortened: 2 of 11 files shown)


manos-saratsis:
## ⚠️ Dromeas PR Review — Fixes required

**1** critical · **4** warnings across 3 agents on `feat/push-usage` (d31427c)

|---|---|---:|---:|---:|
| 🧪 Code Quality | ✅ done | 1 | 4 | 89 |
| 🛡️ Security | ✅ done | 0 | 0 | 97 |
| 📋 Compliance | ✅ done | 0 | 0 | 97 |

### Top findings
- 🔴 **Code Quality** — Unguarded None from aggregate can crash billing/quota callers `api/app_analytics/analytics_db_service.py:178`
- 🔴 **Security** — Potential ReDoS via unsanitized labels_filter in ORM contains lookup `api/app_analytics/analytics_db_service.py:111`
- 🟠 **Code Quality** — Direct ORM and settings access in analytics functions prevents unit testing `api/app_analytics/analytics_db_service.py:81`
- 🟠 **Code Quality** — Hardcoded magic literal `30` for default analytics period used in multiple places `api/app_analytics/analytics_db_service.py:90`
- 🟠 **Code Quality** — Deprecated env variable fallback adds complexity and potential confusion `api/app/settings/common.py:396`

[View full report on Dromeas →](https://dromeas.ai/product/pr-review/eeb6a511-4e55-4e66-a8f0-99efca04ed5f)

api/tests/unit/organisations/usage_reporting/test_services.py
```diff
@@ -0,0 +1,216 @@
...
+from pytest_mock import MockerFixture
+from pytest_structlog import StructuredLogCapture
+
+from organisations.usage_reporting.dataclasses import (
+    ApiCallBreakdown,
+    ProjectUsage,
+    UsageSnapshot,
+)
+from organisations.usage_reporting.services import (
+    _build_auth_token,
```
khvn26: Private imports in tests are a code smell. Usually, if you have to import private code to maintain 100% coverage, this means dead code.

api/tests/unit/organisations/usage_reporting/test_services.py
```diff
@@ -0,0 +1,216 @@
...
+    get_licensed_organisations,
+    push_snapshot,
+    push_usage_snapshots,
+)
+
+SERVICES = "organisations.usage_reporting.services"
```
khvn26: inline this

api/tests/unit/organisations/usage_reporting/test_services.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+def _snapshot() -> UsageSnapshot:
```
khvn26: should be a fixture

api/tests/unit/organisations/usage_reporting/test_mappers.py
```diff
@@ -0,0 +1,103 @@
...
+    assert snapshot.instance_version == "2.142.3"
+    assert snapshot.api_call_total == 10
+    assert snapshot.api_call_breakdown.flags == 1
+    assert snapshot.api_call_breakdown.identities == 2
+    assert snapshot.api_call_breakdown.traits == 3
+    assert snapshot.api_call_breakdown.environment_documents == 4
+    assert snapshot.project_count == 1
+    assert snapshot.project_usage == [
+        ProjectUsage(project_id=project.id, api_call_count=10)
+    ]
```
khvn26: nit: full object comparison assertions are preferred

api/tests/unit/organisations/usage_reporting/test_mappers.py
```diff
@@ -0,0 +1,103 @@
...
+    snapshot = map_organisation_to_usage_snapshot(organisation)
+
+    # Then
+    assert snapshot.api_call_total == 0
+    assert snapshot.api_call_breakdown.flags == 0
+    assert snapshot.api_call_breakdown.environment_documents == 0
+    assert snapshot.project_count == 1
+    assert snapshot.project_usage == [
+        ProjectUsage(project_id=project.id, api_call_count=0)
+    ]
```
khvn26: same as [here](https://github.com/Flagsmith/flagsmith/pull/7831/changes#r3491357257)

api/organisations/usage_reporting/mappers.py
```diff
@@ -0,0 +1,84 @@
...
+    ProjectUsage,
+    UsageSnapshot,
+)
+from projects.models import Project
+
+# The Control Plane rejects payloads with more than this many project_usage rows.
+MAX_PROJECT_USAGE_ROWS = 5_000
+
+
+def _total_api_calls(usage_data: list[UsageData]) -> int:
```
khvn26:
```suggestion
def map_usage_data_list_to_total_api_calls(usage_data: list[UsageData]) -> int:
```

api/organisations/usage_reporting/mappers.py
```diff
@@ -0,0 +1,84 @@
...
+    return sum(
+        data.flags + data.identities + data.traits + data.environment_document
+        for data in usage_data
+    )
+
+
+def _aggregate(usage_data: list[UsageData]) -> tuple[int, ApiCallBreakdown]:
```
khvn26: IMO this should be

```suggestion
def map_usage_data_list_to_api_call_breakdown(usage_data: list[UsageData]) -> ApiCallBreakdown:
```

I can't justify the tuple here

api/app_analytics/analytics_db_service.py
```diff
@@ -110,17 +103,96 @@ def get_usage_data_from_local_db(
...
+def get_usage_data_from_local_db_for_window(
+    organisation: Organisation,
+    date_start: datetime,
+    date_stop: datetime,
+    project_id: int | None = None,
+) -> list[UsageData]:
+    # Filters on the full timestamp (not ``created_at__date``) so that sub-day
+    # windows such as a single hour are honoured. The daily ``get_usage_data``
+    # path keeps its date-truncated filter to leave the billing endpoint
+    # unchanged.
```
khvn26: low-value slop, remove

api/organisations/usage_reporting/mappers.py
```diff
@@ -0,0 +1,84 @@
...
+    organisation: Organisation,
+    projects: QuerySet[Project],
+    hour_start: datetime,
+    hour_end: datetime,
+) -> list[ProjectUsage]:
+    rows = [
+        ProjectUsage(
+            project_id=project.id,
+            api_call_count=_total_api_calls(
+                get_usage_data_for_window(
```
khvn26: Ideally, mappers operate on input data only, and don't pull in any data on their own.

I suggest either making data access explicit, and documenting it, like [here](https://github.com/Flagsmith/flagsmith/blob/420344691551aaf465a5c668516bd4b66de131f8/api/util/mappers/engine.py#L79-L82), or move it outside of the mapper.

api/organisations/usage_reporting/mappers.py
```diff
@@ -0,0 +1,84 @@
...
+    return rows[:MAX_PROJECT_USAGE_ROWS]
+
+
+def map_organisation_to_usage_snapshot(organisation: Organisation) -> UsageSnapshot:
+    hour_end = timezone.now().replace(minute=0, second=0, microsecond=0)
+    hour_start = hour_end - timedelta(hours=1)
+    api_call_total, api_call_breakdown = _aggregate(
+        get_usage_data_for_window(organisation, hour_start, hour_end)
+    )
+    projects = organisation.projects.all()
```
khvn26: same as [here](https://github.com/Flagsmith/flagsmith/pull/7831/changes#r3492970539).

api/organisations/usage_reporting/services.py
```diff
@@ -0,0 +1,79 @@
...
+logger = structlog.get_logger("usage_reporting")
+
+USAGE_ENDPOINT_PATH = "/v1/public/usage"
+REQUEST_TIMEOUT_SECONDS = 30
+
+
+def _build_auth_token(signature: str) -> str:
+    return (
+        base64.urlsafe_b64encode(signature.encode("utf-8")).decode("ascii").rstrip("=")
+    )
```
khvn26: nit: move to `mappers.py` as `map_signature_to_control_plane_auth_token` or even `map_licence_to_control_plane_auth_token`

api/organisations/usage_reporting/services.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def get_licensed_organisations() -> list[Organisation]:
+    if not settings.LICENSING_INSTALLED:
+        return []
+    return list(  # pragma: no cover - requires the optional licensing package
```
khvn26: I don't quite understand how having no `licensing` installed prevents us from covering this code.

api/tests/unit/organisations/usage_reporting/test_services.py
```diff
@@ -0,0 +1,216 @@
...
+def test_build_auth_token__base64url_token__decodes_back_to_signature() -> None:
+    # Given
+    signature = "abc+/=def=="
+
+    # When
+    token = _build_auth_token(signature)
+
+    # Then - the Control Plane recovers the raw signature this way
+    recovered = base64.urlsafe_b64decode(token + "=" * (-len(token) % 4))
+    assert recovered == signature.encode("utf-8")
```
khvn26: Are we testing for outside behaviour here? 🤔

api/organisations/usage_reporting/services.py
```diff
@@ -0,0 +1,79 @@
...
+def push_usage_snapshots() -> None:
+    if not (base_url := settings.CONTROL_PLANE_URL):
+        logger.debug("run.skipped", reason="control_plane_url_unset")
+        return
+    if not (organisations := get_licensed_organisations()):
+        logger.debug("run.skipped", reason="no_licensed_organisations")
+        return
+
+    for organisation in organisations:
+        with structlog.contextvars.bound_contextvars(organisation__id=organisation.id):
```
khvn26, line 67:
```suggestion
        with logger.bind(organisation__id=organisation.id) as log:
```

api/tests/unit/organisations/usage_reporting/test_services.py
germangarces: ae73dd25e

`_build_auth_token` stays fully covered via `push_snapshot`. No private import remains.

api/tests/unit/organisations/usage_reporting/test_services.py
germangarces: 4bb44791b

api/tests/unit/organisations/usage_reporting/test_services.py
germangarces: 635441b62

api/tests/unit/organisations/usage_reporting/test_mappers.py
germangarces: 9125b2192

api/tests/unit/organisations/usage_reporting/test_mappers.py
germangarces: 9125b2192

api/organisations/usage_reporting/mappers.py
germangarces: 86d4210c0

api/organisations/usage_reporting/mappers.py
germangarces: 496e3fb59

api/app_analytics/analytics_db_service.py
germangarces: 32c5ac0e6

api/organisations/usage_reporting/mappers.py
germangarces: 90b08ef3c

api/organisations/usage_reporting/mappers.py
germangarces: https://github.com/Flagsmith/flagsmith/commit/90b08ef3c1eda5086fdc7f02350cc33e5c3b43bf

api/organisations/usage_reporting/services.py
germangarces: fc200c827

api/organisations/usage_reporting/services.py
germangarces: I didn't fully understand this, let me know if this is not acceptable: 1285ba839
I replaced the comment with a guarded test

api/tests/unit/organisations/usage_reporting/test_services.py
germangarces: abe0b43f2

Good point, now I'm just asserting the exact value instead of "re-running" the implementation

api/organisations/usage_reporting/services.py
germangarces: My intent here was to propagate `organisation.id` into `push_snapshot`'s own logs too. `logger.bind()` wouldn't because as far as I know it only tags log lines in this function.

Happy to switch to `bind` if you'd prefer we thread the bound logger into `push_snapshot` instead.

khvn26 (approved):
LGTM 👍

Issue #7408, opened 2026-05-04
Implement new endpoint to return current members of a given segment


khvn26:
### Scope

One endpoint on the existing project segment router:

```
GET /api/v1/projects/{project_id}/segments/{id}/members?env={env_id}&cursor={identifier}&limit={N}
```

Runs the same SQL `WHERE` predicate the count refresh uses, ordered by `identifier` ASC for stable cursor pagination. ClickHouse returns the matched `(environment_id, identifier, identity_key, traits)` tuples; each row gets enriched with `identity_features` from Dynamo so the frontend can render per-flag override badges alongside the matched identity.

Pagination cap: `limit ≤ 200`. Final number gated on product sign-off.

RBAC: `MANAGE_SEGMENTS` on the project plus `VIEW_IDENTITIES` on the environment. Same gate the existing identity-list endpoint uses.

### Pagination

`identifier` is used as the cursor.

The CH query is `WHERE environment_id = :env AND identifier > :cursor AND (<sql engine predicate>) ORDER BY identifier ASC LIMIT N` against `FROM IDENTITIES AS i FINAL`. Cursor is the identifier of the last row of the previous page; first page omits the `cursor` param.

`FINAL` defeats the primary-key sort for skip-index purposes, but pagination (`limit ≤ 200`) bounds the per-call cost. Beta is SaaS-only and we accept the trade-off for accurate reads against the most-recent backfill.

### Acceptance criteria

- Returns identities matching the segment's predicate, env-scoped, paginated by `identifier` cursor; `limit > 200` rejected as 400.
- Each row carries the identity identifier and composite key.
- Emits one Prometheus histogram per call (`flagsmith_segment_membership_read_duration_seconds`) so per-request component latencies are correlatable during incidents.
- Respects the `segment_membership_inspection` FoF flag — returns 404 when the org has the flag off.
- Every CH query carries a `log_comment` of `flagsmith:segment_membership:read:org_<id>:project_<id>` for spend attribution via `system.query_log`.
- API documented in OpenAPI spec.

PR #7848, opened 2026-06-22
ci(api): Add ClickHouse to the local and CI test stack

khvn26:

## Changes

Contributes to #7408

Stands up a real ClickHouse in the test stack so segment membership code can be tested against actual ClickHouse execution instead of a mocked cursor.

- Add a `clickhouse` service to `docker-compose.local.yml`, so it comes up for `make test` both locally and in CI.
- Wire `CLICKHOUSE_*` into `.env-ci`/`.env-local`, and wait for it in `make wait-for-db`. The suite targets the always-present `default` database; the test runner creates its own `test_*` database on top, so we don't depend on first-run DB init.
- Add a `clickhouse_db` fixture and a `clickhouse` marker. ClickHouse has no transactional rollback, so the fixture truncates `IDENTITIES` on teardown for per-test isolation.
- Move the segment-membership tests that stubbed out ClickHouse — compute, refresh, backfill, and `log_comment` attribution — to live-ClickHouse integration tests under `tests/integration/segment_membership/`. Skip/defensive branches that never reach ClickHouse stay mocked.
- Add a thin ClickHouse test backend that implements `_clone_test_db`. Without it, `pytest --ci` results in `NotImplementedError`.

1. `make docker-up` brings up the new `clickhouse` service alongside Postgres.
2. `make test opts='tests/integration/segment_membership/test_segment_membership_clickhouse.py tests/unit/segment_membership/'` — all pass.

Verified specifically:

- The four migrated integration tests pass against real ClickHouse.
- Run under xdist (parallel test-DB cloning works) and with `-n0` (two tests sharing one test DB both see the correct counts, confirming truncation-based isolation).
- pytest-django creates the CH test database and runs the existing `0001_create_identities` migration (JSON-column DDL accepted under `allow_experimental_json_type`); `make wait-for-db` waits on ClickHouse.
- The remaining mocked segment-membership tests still pass with the alias present.

### DIFF
--- a/api/tests/conftest.py
+++ b/api/tests/conftest.py
@@ -1361,7 +1361,7 @@ def clickhouse_db(
     if "clickhouse" not in settings.DATABASES:  # pragma: no cover
         pytest.skip("No ClickHouse database configured, skipping")
     request.applymarker(pytest.mark.django_db(databases=["default", "clickhouse"]))
-    request.getfixturevalue("db")
+    request.getfixturevalue("db")  # Resolve `db` only after injecting the clickhouse db
     yield
     connection = connections["clickhouse"]
     with connection.cursor() as cursor:
--- a/api/tests/conftest.py
+++ b/api/tests/conftest.py
@@ -43,6 +43,7 @@ from common.projects.permissions import CREATE_ENVIRONMENT, DELETE_FEATURE, VIEW
 from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
 from django.core.cache import caches
+from django.db import connections
 from django.db.backends.base.creation import TEST_DATABASE_PREFIX
 from django.test.utils import setup_databases
 from django_test_migrations.migrator import Migrator
@@ -1342,6 +1343,28 @@ def enable_features(
 def clear_content_type_cache() -> typing.Generator[None, None, None]:
     yield
     ContentType.objects.clear_cache()
+
+
+@pytest.fixture
+def clickhouse_db(
+    request: pytest.FixtureRequest, settings: SettingsWrapper
+) -> typing.Generator[None, None, None]:
+    """
+    Opt a test into a live ClickHouse database.
+
+    Skips when no `clickhouse` alias is configured (i.e. ClickHouse isn't
+    running). ClickHouse has no transactional rollback, so -- unlike the
+    Postgres-backed `db` fixture -- we can't rely on Django wrapping the test
+    in a transaction. We truncate IDENTITIES on teardown instead to isolate
+    tests from one another.
+    """
+    if "clickhouse" not in settings.DATABASES:  # pragma: no cover
+        pytest.skip("No ClickHouse database configured, skipping")
+    request.applymarker(pytest.mark.django_db(databases=["default", "clickhouse"]))
+    request.getfixturevalue("db")
+    yield
+    with connections["clickhouse"].cursor() as cursor:
+        cursor.execute("TRUNCATE TABLE IDENTITIES")
 
 
 @pytest.fixture
--- a/api/tests/conftest.py
+++ b/api/tests/conftest.py
@@ -1355,7 +1355,7 @@ def clickhouse_db(
     Skips when no `clickhouse` alias is configured (i.e. ClickHouse isn't
     running). ClickHouse has no transactional rollback, so -- unlike the
     Postgres-backed `db` fixture -- we can't rely on Django wrapping the test
-    in a transaction. We truncate IDENTITIES on teardown instead to isolate
+    in a transaction. We truncate every table on teardown instead to isolate
     tests from one another.
     """
     if "clickhouse" not in settings.DATABASES:  # pragma: no cover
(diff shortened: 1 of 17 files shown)


api/tests/conftest.py
```diff
@@ -1344,6 +1345,28 @@ def clear_content_type_cache() -> typing.Generator[None, None, None]:
...
+    in a transaction. We truncate IDENTITIES on teardown instead to isolate
+    tests from one another.
+    """
+    if "clickhouse" not in settings.DATABASES:  # pragma: no cover
+        pytest.skip("No ClickHouse database configured, skipping")
+    request.applymarker(pytest.mark.django_db(databases=["default", "clickhouse"]))
+    request.getfixturevalue("db")
+    yield
+    with connections["clickhouse"].cursor() as cursor:
+        cursor.execute("TRUNCATE TABLE IDENTITIES")
```
Zaimwa9: NIT: Since they don't natively support transactionall rollback, could we truncate directly all tables so we don't miss any new addition in the future?
Or did you have in mind to keep the teardowns explicit per table?

Zaimwa9 (dismissed):
Non blocking comment on my side, anticipating new tests around experimentation coming soon. I leave it up to your appreciation, the rest makes sense to me (once confirmed by CI) and thanks for the issue in clickhouse

api/tests/unit/segment_membership/test_unit_segment_membership_services.py
```diff
@@ -38,6 +38,10 @@ def test_is_membership_enabled__flag_on__returns_true(
     assert is_membership_enabled(organisation) is True
 
 
+# NOTE: that a `log_comment` actually lands on the session and is attributable
+# in `system.query_log` is verified end-to-end against a live ClickHouse in
+# tests/integration/segment_membership/test_segment_membership_clickhouse.py.
+# This case covers the no-comment branch, where no session setting is applied.
 def test_open_clickhouse_cursor__no_log_comment__yields_cursor(
```
emyller: note: My brain can't make much sense out of this comment, or the test title, sorry.

api/tests/unit/segment_membership/test_unit_segment_membership_tasks.py
```diff
@@ -54,68 +54,10 @@ def test_backfill_identities_to_clickhouse__dynamo_disabled__skips(
...
-    assert {row[1] for row in rows_arg} == {"a", "b"}
-    assert any(
-        e["event"] == "backfill.environment.completed" and e["rows__count"] == 2
-        for e in log.events
-    )
-    refresh_dispatch.delay.assert_called_once_with(args=(project.id,))
+# NOTE: the happy path (rows actually landing in IDENTITIES, refresh dispatched)
+# is covered end-to-end against a live ClickHouse in
+# tests/integration/segment_membership/test_segment_membership_clickhouse.py.
+# The cases below cover the skip/error branches with the cursor mocked.
```
emyller: Looks like slop residue.

api/tests/unit/segment_membership/test_unit_segment_membership_tasks.py
```diff
@@ -254,48 +196,11 @@ def test_refresh_project_segment_counts__compute_fails__logs(
...
-            f"flagsmith:segment_membership:refresh"
-            f":org_{project.organisation_id}"
-            f":project_{project.id}"
-        )
-    )
+# NOTE: the happy path (real predicate over real IDENTITIES rows upserting a
+# real count) is covered end-to-end against a live ClickHouse in
+# tests/integration/segment_membership/test_segment_membership_clickhouse.py.
+# The cases below mock `compute_segment_counts_for_project` to cover the
+# upsert/delete bookkeeping in isolation.
```
emyller: Looks like slop residue.

api/tests/conftest.py
```diff
@@ -1344,6 +1345,28 @@ def clear_content_type_cache() -> typing.Generator[None, None, None]:
...
+    Skips when no `clickhouse` alias is configured (i.e. ClickHouse isn't
+    running). ClickHouse has no transactional rollback, so -- unlike the
+    Postgres-backed `db` fixture -- we can't rely on Django wrapping the test
+    in a transaction. We truncate IDENTITIES on teardown instead to isolate
+    tests from one another.
+    """
+    if "clickhouse" not in settings.DATABASES:  # pragma: no cover
+        pytest.skip("No ClickHouse database configured, skipping")
+    request.applymarker(pytest.mark.django_db(databases=["default", "clickhouse"]))
+    request.getfixturevalue("db")
```
emyller: Why?

emyller (commented):
Looks fine; few non-blocking questions.

api/tests/unit/segment_membership/test_unit_segment_membership_tasks.py
khvn26: It is, sorry. 2de66db

api/tests/unit/segment_membership/test_unit_segment_membership_tasks.py
khvn26: Removed in 2de66db

api/tests/unit/segment_membership/test_unit_segment_membership_services.py
khvn26: Fair enough. Test removed in 2de66db.

api/tests/conftest.py
khvn26: `request.getfixturevalue("db")` and the pytest-django marker makes sure the ClickHouse migrations are run.

api/tests/conftest.py
khvn26: Good point. Added in 40cb5a209c0fe8f4b0e6228dfaeac4907f46a074.

api/tests/conftest.py
emyller: The reasoning sounds a bit fragile, as if there is risk of a race condition if another fixture will load `db` first.

If this is safe otherwise, please add a comment to make this less obscure, e.g:

```suggestion
    request.getfixturevalue("db")  # Resolve `db` only after injecting the clickhouse db
```

emyller (approved):

PR #7851, opened 2026-06-23
feat(experimentation): experiment rollout with multivariate segment override

gagantrivedi:

## Changes

Adds an **experiment rollout** to experiment creation. Posting an `experiment_rollout` block alongside a new experiment provisions, in one atomic create:

- a `PERCENTAGE_SPLIT` **system segment** (`is_system_segment=True`, so it's hidden from the segments API/UI and audit log) gating the experiment cohort by `rollout_percentage`;
- a **segment override** on the experiment's multivariate feature carrying the control value and the per-variant `multivariate_feature_state_values` split.

Supporting change in the versioning layer: `FlagChangeSet` / `SegmentOverrideChangeSet` and `update_flag` / `update_flag_v2` now carry optional multivariate allocations (upsert-style), so segment-override writes set the variant split correctly for **both v1 and v2** feature-versioning environments. This also benefits the existing `update-flag-v2` endpoint.

### Not yet included (intentional)

- `experiment_rollout` is **write-only** — no read representation for the UI yet.
- No teardown of the rollout segment/override when an experiment completes.

Unit tests (all green locally), plus `mypy`, `ruff`, and the test linter:

- Versioning service: `update_flag` / `update_flag_v2` set/upsert/remove multivariate allocations on segment overrides, for v1 and v2 environments.
- Rollout services: `create_experiment_rollout` builds the system %-split segment + override; `update_experiment_rollout` updates it and is blocked while running/completed and when no rollout exists.
- API: creating an experiment with `experiment_rollout` (incl. invalid MV option → 400), the `rollout` PATCH action (happy path, running → 400, invalid MV → 400), and rejecting a rollout sent to the experiment update endpoint.


api/experimentation/services.py
```diff
@@ -512,6 +518,86 @@ def transition_experiment_status(
...
+        experiment.feature,
+        FlagChangeSet(
+            author=author,
+            enabled=enabled,
+            feature_state_value=feature_state_value,
+            type_=value_type,
+            segment_id=segment.id,
+            multivariate_values=multivariate_values,
+        ),
+    )
```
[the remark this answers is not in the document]
gagantrivedi: Atomic blocks have caused us more pain than they're worth. We can add them back if we actually need it.

api/features/feature_states/serializers.py
```diff
@@ -172,12 +209,24 @@ def change_set_v2(self) -> FlagChangeSetV2:
...
+                        multivariate_feature_option_id=mv[
+                            "multivariate_feature_option"
+                        ],
+                        percentage_allocation=mv["percentage_allocation"],
+                    )
+                    for mv in multivariate_data
+                ]
+                if multivariate_data
+                else None,
             )
```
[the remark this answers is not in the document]
gagantrivedi: Intentional — we no longer delete unpassed MV values, so `[]` and `None` are both no-ops ("nothing to upsert"). Clearing via an empty list isn't a supported operation here, so the distinction doesn't matter. Leaving as-is.

api/experimentation/services.py
```diff
@@ -10,9 +10,11 @@
 from django.conf import settings
 from django.db.models import Q
 from django.utils import timezone
+from flag_engine.segments.constants import PERCENTAGE_SPLIT
```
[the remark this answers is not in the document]
gagantrivedi: Not needed — we deliberately don't wrap the rollout update in a transaction (see the thread on `update_experiment_rollout`). Skipping.

api/features/versioning/versioning_service.py
```diff
@@ -252,6 +256,39 @@ def _update_feature_state_value(
...
+            mv.percentage_allocation = value.percentage_allocation
+            mv.save()
+
+    stale_option_ids = set(existing) - {
+        value.multivariate_feature_option_id for value in values
+    }
+    if stale_option_ids:
+        feature_state.multivariate_feature_state_values.filter(
+            multivariate_feature_option_id__in=stale_option_ids
+        ).delete()
```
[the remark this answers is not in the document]
gagantrivedi: Good catch. Handled upstream instead: duplicate `multivariate_feature_option`s are now rejected in serializer validation ("must be unique"), mirroring the existing duplicate-`segment_id` check, so the service never receives dupes. Preferred that over silently last-write-win.

api/experimentation/views.py
```diff
@@ -290,6 +295,28 @@
...
+                for mv in data.get("multivariate_feature_state_values", [])
+            },
+        )
+        try:
+            update_experiment_rollout(
+                experiment,
+                **ExperimentRolloutSerializer.to_service_kwargs(data, request),
+            )
+        except ValueError as exc:
+            return Response({"detail": str(exc)}, status=status.HTTP_400_BAD_REQUEST)
```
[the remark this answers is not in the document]
gagantrivedi: Fixed — `update_experiment_rollout` now raises `ValidationError` and the manual `str(exc)` response was removed, so DRF renders the 400 without exposing exception text.

api/experimentation/serializers.py
```diff
@@ -207,6 +215,35 @@ class ExperimentMetricInlineSerializer(serializers.Serializer):  # type: ignore[
...
+    rollout_percentage = serializers.FloatField(
+        required=True, min_value=0, max_value=100
+    )
+    feature_state_value = FeatureValueSerializer(required=True)
+    multivariate_feature_state_values = MultivariateValueSerializer(
+        many=True, required=False
+    )
+
+    @staticmethod
+    def to_service_kwargs(data: dict[str, Any], request: Any) -> dict[str, Any]:
```
Zaimwa9: I'd rename this to have a more business oriented definition, which looks more like a mapper to experiment_rollout. Suggestion like `map_to_rollout_create`

api/experimentation/views.py
```diff
@@ -290,6 +295,21 @@ def pause(self, request: Request, **kwargs: object) -> Response:
...
+    @action(detail=True, methods=["patch"])
+    def rollout(self, request: Request, **kwargs: object) -> Response:
+        experiment: Experiment = self.get_object()
+        serializer = ExperimentRolloutSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        data = serializer.validated_data
+        validate_multivariate_state_values(
+            experiment.feature, data.get("multivariate_feature_state_values", [])
+        )
+        update_experiment_rollout(
```
Zaimwa9: We might want to wrap it in a transaction here

api/features/feature_states/serializers.py
```diff
@@ -128,11 +131,36 @@ class EnvironmentDefaultSerializer(serializers.Serializer):  # type: ignore[type
...
 
 
+class MultivariateValueSerializer(serializers.Serializer):  # type: ignore[type-arg]
+    multivariate_feature_option = serializers.IntegerField(required=True)
+    percentage_allocation = serializers.FloatField(
+        required=True, min_value=0, max_value=100
+    )
+
+
+def validate_multivariate_state_values(
```
Zaimwa9, line 141: This could also be the right place to validate the total weights

api/tests/unit/features/versioning/test_unit_versioning_versioning_service.py
```diff
@@ -621,3 +632,298 @@ def test_get_environment_flags_list__from_replica__returns_feature_states(
...
+) -> FlagChangeSetV2:
+    return FlagChangeSetV2(
+        author=author,
+        environment_default_enabled=True,
+        environment_default_value="control",
+        environment_default_type="string",
+        segment_overrides=[
+            SegmentOverrideChangeSet(
+                segment_id=segment.id,
+                enabled=True,
```
Zaimwa9, line 651: We should set priority here ?

Zaimwa9 (changes requested):
Couple of comments, most important one to me is the transaction in the update path. And maybe be intentional on how we set the priority for these segments

api/experimentation/serializers.py
gagantrivedi: does it look better now?

api/experimentation/views.py
gagantrivedi: Done

api/features/feature_states/serializers.py
gagantrivedi: weight validation lives at `_update_multivariate_values` because it's the only layer with the retained allocations

api/tests/unit/features/versioning/test_unit_versioning_versioning_service.py
Zaimwa9: Discussed offline, going with this version until we have more visibility

api/experimentation/serializers.py
Zaimwa9: Even better than in my dreams

Zaimwa9 (approved):
👍

PR #7857, opened 2026-06-23
feat(experiments): wire up results refresh with a shared RefreshControl

Zaimwa9:
The "Refresh results" action was never wired up. This adds it as a primary button below the status button, at parity with exposures refresh.

- Extracts the exposures-only `AsOfRefreshControl` into a shared `base/forms/RefreshControl` (button + message slot + `theme`); adds a Storybook story.
- New `ExperimentResultsRefreshControl`: 429/`Retry-After` throttling, polling until recompute lands, view-state + final/availability gating. Converts the Bayesian refresh mutation to a `queryFn` for `Retry-After`.
- Token hygiene: phantom `--font-*-size` → literals; `#e53e3e` → `colorIconDanger`.

- `resultsViewState` unit tests + existing results suites: 37/37 pass; lint + tsc clean.
- Manual: refresh on a running experiment updates scorecards; rapid click → countdown; completed → disabled.

image
image


frontend/web/components/experiments/results/RefreshControl.tsx
```diff
@@ -0,0 +1,39 @@
...
+    >
+      {isRefreshing ? 'Refreshing…' : children ?? 'Refresh'}
+    </Button>
+    {label ? (
+      <div className='text-muted fs-caption mt-1 text-end'>{label}</div>
+    ) : null}
+  </div>
+)
+
+export default RefreshControl
```
talissoncosta, line 40: I'd keep this one out of `base/forms` and move it next to its consumers in `experiments/results/components`.

 Looking at what it actually does over `Button`: swap the label to "Refreshing…", fold `isRefreshing` into `disabled`, map `disabledReason` to `title`, and a caption slot underneath. So it's really "a Button with a loading state" plus a refresh-specific caption, and both its importers are experiment panels. I don't think it earns a spot as a shared primitive yet.

By the way, I like the idea of  loading state, we should add a `isLoading` prop that can provide this behaviour easily.

(if it moves local, the Storybook story can be moved under `Experiments` or be dropped, your call)

Image

frontend/web/components/experiments/results/ExperimentResultsRefreshControl.tsx
```diff
@@ -0,0 +1,142 @@
...
+} from './resultsViewState'
+
+const parseRetryAfter = (err: unknown): number | null => {
+  const fetchErr = err as {
+    status?: number
+    retryAfter?: number | null
+  }
+  if (fetchErr.status !== 429) return null
+  if (fetchErr.retryAfter) return fetchErr.retryAfter
+  return 300
```
talissoncosta: non blocking.

as we have costs for most of _MS. What about adding his to a const too, like: `DEFAULT_RETRY_AFTER_S`

frontend/web/components/experiments/results/ExperimentResultsRefreshControl.tsx
```diff
@@ -0,0 +1,142 @@
...
+  useEffect(() => {
+    if (retryAfter === null || retryAfter <= 0) return
+    const timer = setInterval(() => {
+      setRetryAfter((prev) => {
+        if (prev === null || prev <= 1) return null
+        return prev - 1
+      })
+    }, 1000)
+    return () => clearInterval(timer)
+  }, [retryAfter !== null]) // eslint-disable-line react-hooks/exhaustive-deps
```
talissoncosta: Optional: this countdown is generic timer logic and exposures has the same one, so it could become a small `useCountdown` hook. Using `setTimeout`-per-tick with `[remaining]` as the dep also drops the `eslint-disable`:

```ts
export const useCountdown = (): [number | null, (seconds: number) => void] => {
  const [remaining, setRemaining] = useState<number | null>(null)
  useEffect(() => {
    if (remaining === null || remaining <= 0) return
    const timer = setTimeout(
      () => setRemaining((prev) => (prev && prev > 1 ? prev - 1 : null)),
      1000,
    )
    return () => clearTimeout(timer)
  }, [remaining])
  const start = useCallback((s: number) => setRemaining(s > 0 ? s : null), [])
  return [remaining, start]
}
```

Then it replaces the `retryAfter` state + the countdown effect:

```tsx
const [retryAfter, startRetryCountdown] = useCountdown()

const handleRefresh = useCallback(async () => {
  setRefreshRequested(true)
  setPollStartedAt(Date.now())
  const result = await refresh({ environmentId, experimentId })
  if ('error' in result && result.error) {
    setRefreshRequested(false)
    setPollStartedAt(null)
    const seconds = parseRetryAfter(result.error)
    if (seconds !== null) {
      startRetryCountdown(seconds) // was setRetryAfter(seconds)
    } else {
      toast('Failed to refresh results', 'danger')
    }
  }
}, [refresh, environmentId, experimentId])
```

Totally fine to skip if you'd rather keep the PR focused.

frontend/web/components/experiments/results/ExperimentResultsRefreshControl.tsx
```diff
@@ -0,0 +1,142 @@
...
+  let label: ReactNode = undefined
+  if (retryAfter !== null) {
+    label = `Computing, retry in ${formatCountdown(retryAfter)}`
+  } else if (isRefreshing) {
+    label = 'Computing… results will update automatically.'
+  } else if (viewState.kind === 'error') {
+    label = (
+      <span className='text-danger'>The last results computation failed.</span>
+    )
+  }
```
talissoncosta: Suggestion:

What if we extract this to a function to `resultsViewState.ts` and unit test it ?

```ts
export const getResultsRefreshLabel = (
  retryAfter: number | null,
  isRefreshing: boolean,
  viewState: ResultsViewState,
): { message: string; tone: 'muted' | 'danger' } | null => {
  if (retryAfter !== null)
    return { message: `Computing, retry in ${formatCountdown(retryAfter)}`, tone: 'muted' }
  if (isRefreshing)
    return { message: 'Computing… results will update automatically.', tone: 'muted' }
  if (viewState.kind === 'error')
    return { message: 'The last results computation failed.', tone: 'danger' }
  return null
}
```

```tsx
  const label = getResultsRefreshLabel(retryAfter, isRefreshing, viewState)

  return (  
    <RefreshControl
      disabled={!availability.canRefresh || isRefreshing || retryAfter !== null}
      disabledReason={
        availability.reason
          ? REFRESH_DISABLED_COPY[availability.reason]
          : undefined
      }
      isRefreshing={isRefreshing && hasData}
      label={
        label && (
          <span className={label.tone === 'danger' ? 'text-danger' : undefined}>
            {label.message}
          </span>
        )
      }
      onRefresh={handleRefresh}
    >
      Refresh results 
    </RefreshControl>
  )

```

talissoncosta (changes requested):
Really nice work @Zaimwa9. 

One thing I'd like to sort before merge: RefreshControl in base/forms. Could we move it into `experiments/results/components` next to the panels?

I just wanna call for this component flickering when you click the refresh button. The text appear on the right then it appears below the button. This is not a blocker.

https://www.loom.com/share/9e2ab09f71744c718a5b44714d4fded5

Everything else inline is optional.


Thanks for this!

talissoncosta (approved):
Looks good, thanks for addressing the changes @Zaimwa9 !

PR #7858, opened 2026-06-24
feat(experimentation): expose experiment rollout in API responses

gagantrivedi:

## Changes

Contributes to 

Follow-up to #7851, which left `experiment_rollout` write-only. This exposes it on read so the UI can display the configured rollout.

The rollout isn't stored as a single field — on read it's reassembled from its two sources of truth:

- `rollout_percentage` ← the rollout segment's `PERCENTAGE_SPLIT` condition value
- `enabled` / `feature_state_value` / `multivariate_feature_state_values` ← the live segment-override feature state (resolved the same v1/v2-aware way as `ExperimentFeatureSerializer` does for env-level allocations)

A `get_experiment_rollout(experiment)` service returns this shape (mirroring the write input), and `ExperimentListSerializer` exposes it via a read-only `experiment_rollout` field — so it appears on list, retrieve, and the `rollout` action responses. The write-only input field stays on the base serializer (mirrors the `metrics` pattern); returns `null` when no rollout is configured.

Unit tests (all green locally) + `mypy`, `ruff`, the test linter, and docgen:

- Service: `get_experiment_rollout` reassembles the representation for v1 and v2 environments, and returns `None` when no rollout exists.
- API: `GET` experiment detail returns the populated `experiment_rollout`, and `null` when absent.
- Full experimentation suite passes — no regression from adding the field to the read representation.

### DIFF
--- a/api/experimentation/serializers.py
+++ b/api/experimentation/serializers.py
@@ -19,7 +19,10 @@ from experimentation.models import (
     WarehouseConnection,
     WarehouseType,
 )
-from experimentation.services import apply_experiment_rollout
+from experimentation.services import (
+    apply_experiment_rollout,
+    get_experiment_rollout,
+)
 from experimentation.types import (
     SNOWFLAKE_DEFAULTS,
     MetricExperimentResult,
@@ -251,7 +254,9 @@ class ExperimentSerializer(serializers.ModelSerializer):  # type: ignore[type-ar
         required=False,
         write_only=True,
     )
-    experiment_rollout = ExperimentRolloutSerializer(required=False, write_only=True)
+    experiment_rollout: Any = ExperimentRolloutSerializer(
+        required=False, write_only=True
+    )
 
     class Meta:
         model = Experiment
@@ -393,6 +398,10 @@ class ExperimentListSerializer(ExperimentSerializer):
         many=True,
         read_only=True,
     )
+    experiment_rollout = serializers.SerializerMethodField()
+
+    def get_experiment_rollout(self, experiment: Experiment) -> dict[str, Any] | None:
+        return get_experiment_rollout(experiment)
 
 
 class ExperimentExposuresSerializer(serializers.ModelSerializer):  # type: ignore[type-arg]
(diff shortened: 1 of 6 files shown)


api/experimentation/services.py
```diff
@@ -592,6 +595,40 @@ def apply_experiment_rollout(experiment: Experiment, spec: RolloutSpec) -> None:
...
+    condition = Condition.objects.get(
+        rule__segment_id=segment_id, operator=PERCENTAGE_SPLIT
+    )
+    value = feature_state.feature_state_value
+    return {
+        "enabled": feature_state.enabled,
+        "rollout_percentage": float(condition.value or 0),
+        "feature_state_value": {
+            "type": _ROLLOUT_VALUE_TYPE.get(value.type or STRING, "string"),
+            "value": str(value.value),
```
Zaimwa9: Won't this return `"True"/"False"` not as pure natively parseable boolean ? Can we add a test with a boolean value please ?

api/experimentation/serializers.py
```diff
@@ -393,6 +398,10 @@ class ExperimentListSerializer(ExperimentSerializer):
         many=True,
         read_only=True,
     )
+    experiment_rollout = serializers.SerializerMethodField()
+
+    def get_experiment_rollout(self, experiment: Experiment) -> dict[str, Any] | None:
```
Zaimwa9, line 406: it smells like a N+1 here as it is going to run on every row. The endpoint is paginated so the consequences are contained but it could be worth fixing it sooner than later

Zaimwa9 (commented):
I'm concerned about the boolean api contract and there is a hint of a N+1 query that could be worth a second look

api/experimentation/services.py
gagantrivedi: fixed now

api/experimentation/serializers.py
gagantrivedi: Yep, moved it to detail serialzier instead

Zaimwa9 (approved):
Thanks!

PR #7859, opened 2026-06-24
feat(experimentation): rollout configuration step in experiment wizard

Zaimwa9:

## Changes

Adds the **Rollout configuration** step to the Create Experiment wizard (before Metrics). Frontend only, pairs with the backend `experiment_rollout` contract in #7851.

- **Sample Size**: a centred slider setting the percentage of eligible identities that enter the experiment.
- **Variation Split**: editable per-variant weights (control takes the remainder, plus a "Split evenly" action). Weights default to the feature's live environment values, falling back to project defaults.
- **Rollout summary**: a distribution bar (scaled variant arms plus a hatched "not released" area) with a legend, shown in the step and the Review step.
- Wires the `experiment_rollout` payload into create: maps the control value to the backend's `{ type, value }` shape; variant splits pass through as `multivariate_feature_state_values`.
- Adds a reusable `DistributionBar` component and a pure, unit-tested `rollout.ts` util module.

- **Unit**: `rollout.test.ts` covers split defaults, even split, control percentage, summary rows, traffic segments, and the control-value mapper.
- **Manual**: walked the wizard end-to-end (slider, split editor with even split and weight cap enforced, summary and distribution bar, and the Review step), and confirmed `experiment_rollout` is posted on create.

image


frontend/web/components/base/grid/ContentCard/ContentCard.tsx
```diff
@@ -18,12 +19,14 @@ const ContentCard: FC<ContentCardProps> = ({
   compact,
   description,
   title,
+  white,
```
talissoncosta: Do you mind giving me a bit more context on this prop ?

talissoncosta: Do you mind giving me a bit more context on this prop ?

frontend/web/components/base/grid/ContentCard/ContentCard.scss
```diff
@@ -12,9 +12,14 @@
...
   border-radius: var(--radius-lg);
 
+  &--white {
+    background: var(--color-surface-default);
+  }
+
   &__heading {
     display: flex;
     flex-direction: column;
+    gap: 6px;
```
talissoncosta: nit:
`display: flex; flex-direction: column; gap: 6px` is `d-flex flex-column gap-2` (8px) as utilities on the element. `6px` is off the spacing scale, but `gap-2` might close enough and lets the rule go.

frontend/web/components/experiments/RolloutSummary/RolloutSummary.tsx
```diff
@@ -0,0 +1,90 @@
...
+      <div className='rollout-summary__note'>
+        <Icon name='people' width={20} />
+        <span>
+          {rolloutPercentage}% of eligible identities enter the experiment.
+          <br />
+          Actual time-to-significance depends on traffic, baseline rate, and the
+          lift you're trying to detect.
+        </span>
+      </div>
+    </div>
```
talissoncosta, line 86: Suggestion:
What if we add the legends below instead of following the spaces on bar.

Before:
Image

After:
Image


I just thought that it could be hard to read or identify the variant when the percentage is low

talissoncosta (dismissed):
Looks pretty good, solid!!! Thanks for that @Zaimwa9 !


My main question is around the `white` prob. As it is not a blocker I will approve it.

frontend/web/components/base/grid/ContentCard/ContentCard.tsx
Zaimwa9: I'll rename this prop. Most of the card have the default gray background, for the chart sample size and variation split ones it needed to be white

Zaimwa9: https://github.com/Flagsmith/flagsmith/pull/7859#discussion_r3474869400

frontend/web/components/experiments/RolloutSummary/RolloutSummary.tsx
Zaimwa9: Yeah I hesitated to have them at least overflow, knowing it would be a limitation. I'll try with the legend

Zaimwa9: image

As suggested

talissoncosta: Looks great! Thanks!

talissoncosta (approved):

PR #7861, opened 2026-06-24
feat(Segment membership inspection): Add segment members read endpoint

### LINKED ISSUE #7408: Implement new endpoint to return current members of a given segment

khvn26:

## Changes

Closes #7408.

Adds a read endpoint that lists the identities currently matching a segment, cursor-paginated.

`GET /api/v1/projects/{project_id}/segments/{id}/members?environment={env_id}&cursor={identifier}&limit={N}`

Added integration and unit tests.

Will test extensively in staging.


api/segments/serializers.py
```diff
@@ -246,3 +247,38 @@ class CloneSegmentSerializer(serializers.ModelSerializer[Segment]):
...
+        help_text="The id of the environment to list segment members for.",
+    )
+    cursor = serializers.CharField(
+        required=False,
+        help_text="The identifier of the previous page's last row; omit for the first page.",
+    )
+    q = serializers.CharField(
+        required=False,
+        help_text="Case-insensitive substring to filter members by identifier.",
+    )
```
emyller, line 263: question: Is filtering by the identifier string — which is [preferably a GUID](https://docs.flagsmith.com/flagsmith-concepts/identities) — a use case as relevant as filtering by other attributes, e.g. traits?

emyller (approved):
Looks great. One product-related question.

api/segments/serializers.py
khvn26: We'll cross that river when we get there. Tech wise, I'd say we're extremely close, but I'd like us to figure out the UX first.

emyller: What use case does `q` mean to serve? The reason I'm asking is to validate:
1. If the use case is "find **one** identity by `identifier_guid`", 👍 
2. If the use case is "find N identities by `arbitrary_str`", then it suggests a disagreement to the core concept mentioned above.

khvn26: We're mainly supporting already existing functionality here. Besides, for some customers, identifiers are not guids, and there's no guideline against prefixing/suffixing them — the idea behind using a GUID is simply making identifiers less guessable.

emyller: I still believe having identifier search around sounds inconsistent to how we _strongly recommend_ against it.

The compatibility argument makes sense, but I wonder: would this discussion exist if we [hypothetically] started doing trait search instead?

Sorry I don't mean to be annoying. I'm fine with shipping this, but would be happier with a comment pointing to the trait-based search issue as the intended behaviour in v1.

khvn26: > I still believe having identifier search around sounds inconsistent to how we _strongly recommend_ against it.

I don't understand the relevance. Hard to guess by a third party != not searchable.

> would this discussion exist if we [hypothetically] started doing trait search instead?

Again, we'll cross that bridge when we come to it. It's not relevant to the scope of this PR. It will be trivial to modify the backend code if/when it's needed for trait search. I don't see how any additional documentation should help.

PR #7865, opened 2026-06-24
feat(experimentation): whitelist server-side keys for warehouse ingestion

Zaimwa9:

## Changes

- Redis ingestion whitelist entries now store the environment's client API key as the value (was empty), enabling the pipeline to canonicalise `ser.…` keys to the correct `environment_key`
- Warehouse connection create/delete syncs all active server-side keys alongside the client key
- Server-side keys with `expires_at` get a matching Redis TTL
- `post_save`/`post_delete` signals on `EnvironmentAPIKey` keep the whitelist in sync on key rotation/deactivation/deletion (scoped to environments with a warehouse connection)
- Requires a companion pipeline PR (deploy pipeline first) to switch from `EXISTS` to `GET`

- Added 25 unit tests in `test_ingestion_sync_service.py`, `test_tasks.py`, `test_models.py`, `test_signals.py`


api/experimentation/ingestion_sync_service.py
```diff
@@ -20,11 +24,24 @@ def _get_client() -> RedisCluster:
...
-def set_environment_key(environment_api_key: str) -> None:
-    key = f"{INGESTION_ENVIRONMENT_KEY_PREFIX}{environment_api_key}"
-    _get_client().set(key, "")
+def set_ingestion_key(
+    key: str,
+    *,
+    environment_key: str,
+    expires_at: datetime | None = None,
+) -> None:
+    """Whitelist ``key`` for warehouse ingestion, mapping it to the canonical
```
gagantrivedi: nit: do we need this docstring?

api/experimentation/tasks.py
```diff
@@ -10,13 +11,74 @@
...
+        )
+    else:
+        ingestion_sync_service.delete_ingestion_key(api_key.key)
 
 
 @register_task_handler()
-def delete_environment_key_from_ingestion(environment_api_key: str) -> None:
-    ingestion_sync_service.delete_environment_key(environment_api_key)
+def remove_server_side_key_from_ingestion(key: str) -> None:
+    """Remove a deleted server-side key from the ingestion whitelist."""
```
gagantrivedi: are we overdoing with the docstring?

api/experimentation/signals.py
```diff
@@ -0,0 +1,32 @@
+from typing import Any
```
gagantrivedi: We're using hooks on the warehouse connection model but a signal here? Is there a reason we can't use hooks here too? Better to stick with one

api/experimentation/tasks.py
```diff
@@ -10,13 +11,74 @@
 
 
 @register_task_handler()
-def add_environment_key_to_ingestion(environment_api_key: str) -> None:
-    ingestion_sync_service.set_environment_key(environment_api_key)
+def sync_environment_ingestion_keys(environment_id: int) -> None:
```
gagantrivedi: nit: sync implies that this is handling delete as well?
maybe this can just be `write_environment_ingestion_keys` considering we also have `remove_environment_ingestion_keys`

api/experimentation/tasks.py
```diff
@@ -10,13 +11,74 @@
...
+    if environment is None:
+        return
+
+    ingestion_sync_service.delete_ingestion_key(environment.api_key)
+    for api_key in environment.api_keys.all():
+        ingestion_sync_service.delete_ingestion_key(api_key.key)
+
+
+@register_task_handler()
+def reconcile_server_side_key_ingestion(environment_api_key_id: int) -> None:
```
gagantrivedi: This method name doesnt make much sense to me? This looks like just a singlular form of the above methods?
`write_environment_ingestion_key` ?

api/experimentation/tasks.py
```diff
@@ -10,13 +11,74 @@
...
+            expires_at=api_key.expires_at,
+        )
+    else:
+        ingestion_sync_service.delete_ingestion_key(api_key.key)
 
 
 @register_task_handler()
-def delete_environment_key_from_ingestion(environment_api_key: str) -> None:
-    ingestion_sync_service.delete_environment_key(environment_api_key)
+def remove_server_side_key_from_ingestion(key: str) -> None:
```
gagantrivedi:
```suggestion
def remove_environment_ingestion_key(key: str) -> None:
```

gagantrivedi (changes requested):

api/experimentation/signals.py
```diff
@@ -0,0 +1,32 @@
...
+)
+
+
+@receiver(post_save, sender=EnvironmentAPIKey)
+def sync_server_side_key_to_ingestion(
+    sender: type[EnvironmentAPIKey],
+    instance: EnvironmentAPIKey,
+    **kwargs: Any,
+) -> None:
+    reconcile_server_side_key_ingestion.delay(
```
gagantrivedi: we also need to make sure we only trigger these if wearehouse connection is live

api/experimentation/signals.py
Zaimwa9: Let's go with hooks, there was a question of coupling the 2 apps but as we used the pattern already let's go with it.

Addressed in [a34982d](https://github.com/Flagsmith/flagsmith/pull/7865/commits/a34982d8fab173cea1c93e4007e1310c76b246a9)

api/experimentation/tasks.py
Zaimwa9: Addressed in [04f3941](https://github.com/Flagsmith/flagsmith/pull/7865/commits/04f3941a3a9f3413b626394a7f06147ddd2c6f17)

api/experimentation/tasks.py
Zaimwa9: Addressed in [04f3941](https://github.com/Flagsmith/flagsmith/pull/7865/commits/04f3941a3a9f3413b626394a7f06147ddd2c6f17)

api/experimentation/tasks.py
Zaimwa9: Addressed in [04f3941](https://github.com/Flagsmith/flagsmith/pull/7865/commits/04f3941a3a9f3413b626394a7f06147ddd2c6f17)

gagantrivedi (approved):

gagantrivedi (approved):

PR #7864, opened 2026-06-24
feat: Use new segment members api

kyle-ssg:
Thanks for submitting a PR! Please check the boxes below:

- [ ] I have read the [Contributing Guide](/Flagsmith/flagsmith/blob/main/CONTRIBUTING.md).
- [ ] I have added information to `docs/` if required so people know about the feature.
- [ ] I have filled in the "Changes" section below.
- [ ] I have filled in the "How did you test this code" section below.

## Changes

Frontend for https://github.com/Flagsmith/flagsmith/pull/7861

Replaces current view of identities that belong to a certain segment

## How did you test this code?

_Please describe._

PR #7866, opened 2026-06-24
feat(Segment membership inspection): Add identifier search to the segment members endpoint

khvn26:

## Changes

Follow-up to #7861. Adds an optional `q` query param to the segment members endpoint so the matched identities can be searched by identifier.

`GET /api/v1/projects/{id}/segments/{id}/members?environment={env_id}&q={term}`

Added integration/unit tests. Will be tested along with #7864.

### DIFF
--- a/api/segment_membership/services.py
+++ b/api/segment_membership/services.py
@@ -160,11 +160,13 @@ def get_segment_members_page(
     *,
     cursor: str | None,
     limit: int,
+    q: str | None = None,
 ) -> list[SegmentMember]:
     """Return one page of identities matching `segment` in `environment`,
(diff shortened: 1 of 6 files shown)


Zaimwa9 (approved):
This looks good to me. One non-blocking comment would be that I wonder whether we could find a more elegant way to deal with [this workaround](https://github.com/Flagsmith/flagsmith/pull/7866/changes#diff-092860cddb9eb4694a09d3c5e7b71ec85fcc2d4036548b36a38bd97876adb122R181-R189) but that's really NIT and not worth spending too much time on it imo

PR #7856, opened 2026-06-23
feat(onboarding): connection terminal and flags table

talissoncosta:
- [x] I have read the [Contributing Guide](/Flagsmith/flagsmith/blob/main/CONTRIBUTING.md).
- [ ] I have added information to `docs/` if required so people know about the feature.
- [x] I have filled in the "Changes" section below.
- [x] I have filled in the "How did you test this code" section below.

## Changes

Contributes to #7766. Builds on the single-page onboarding flow (#7765, now merged): adds the verify console and the flags table below the connect panel.

- **Verify terminal** — a dark SDK console whose checklist ticks as you act: copy the install command, copy the wire snippet, then the first evaluation flips the badge to LIVE. The connection step is **stubbed** behind `useOnboardingConnection` (a `?connected` query param exercises the connected UI) pending the real first-evaluation signal in #7767, so it won't go LIVE on its own yet.
- **Flags table** — the pre-created flag in a real table reusing the product `FeatureName` / `Tag` / `Switch`, with a real persisted Development toggle. Bootstrap attaches an "Onboarding" tag so the badge shows.
- **New-user redirect** — after signup a new user has no organisation, so `onLogin` sent them to the legacy `/create` page before the getting-started check ran. With the flow enabled they now land on `/getting-started` (which creates the organisation itself).
- **Rename keeps tags** — feature names are immutable, so the inline "rename" is a delete + recreate; it now carries over the old flag's tags/type/description so the Onboarding badge survives.
- **Accessibility** — distinct accessible names for the two copy buttons, a per-flag name on the toggle, and the flags table as a labelled region.
- Browser autofill opt-out on the shared `GhostInput` (the contact icon was overlapping the inline edit pencil).
- **SCSS → utilities** — layout (flex/gap/margin) moved to Bootstrap utilities and the flags-table surface/radius to token utilities; the SCSS keeps only what has no utility equivalent (the accent glow, the waiting state, non-standard metrics, and the terminal's literal console palette).
- **E2E** — new-user happy path behind `onboarding_quickstart_flow`: sign up, land on the flow, copy the install/wire snippets (checklist ticks), toggle the flag, rename it (asserting the tag survives), and the `?connected` state. Accessibility-first selectors (roles / labels / text).

## How did you test this code?

[...]

PR #7867, opened 2026-06-24
Onboarding: e2e coverage, rename keeps tags, accessible names

talissoncosta:

## Changes

Stacked on #7856. Follow-up from Wadii's review of the onboarding flow: e2e coverage plus a couple of robustness fixes.

- **Rename preserves tags.** Feature names are immutable, so the onboarding "rename" is a delete + recreate. The recreate started a bare flag, dropping the Onboarding tag attached in bootstrap (the badge vanished). It now carries over the old flag's tags/type/description.
- **Accessible names.** The two copy buttons were both "Copy Code" (ambiguous for assistive tech), the flags toggle had no accessible name, and the flags table wasn't a labelled region. Fixed all three; this also gives the e2e a stable, a11y-based handle (no test-ids).
- **E2E.** New-user happy path behind `onboarding_quickstart_flow`: sign up, land on the flow, copy the install/wire snippets (checklist ticks), toggle the Development flag, rename it (asserting the tag survives), and the `?connected` state. Selectors are accessibility-first (roles / labels / text).

Components type-check and lint clean. The e2e spec runs with `npm run test -- tests/onboarding-tests.pw.ts` (with `onboarding_quickstart_flow` on); staying in draft until it has a green run.

### DIFF
--- a/frontend/e2e/tests/onboarding-tests.pw.ts
+++ b/frontend/e2e/tests/onboarding-tests.pw.ts
@@ -1,16 +1,14 @@
 import { test, expect } from '../test-setup';
-import {
-  byId,
-  createHelpers,
-  getFlagsmith,
-  log,
-  visualSnapshot,
-} from '../helpers';
+import { byId, createHelpers, getFlagsmith, log, visualSnapshot } from '../helpers';
 import { E2E_SIGN_UP_USER, PASSWORD } from '../config';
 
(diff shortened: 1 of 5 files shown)


talissoncosta:
Folded into #7856 — the accessible names, the rename tag-preservation fix, and the e2e spec all landed there, so the stack is collapsed into one PR.

PR #7877, opened 2026-06-25
fix(Tagging): TagType evaluated incorrectly in database queries

emyller:

## Changes

Contributes to https://github.com/Flagsmith/flagsmith/issues/7433

Fixes a bug in which tag-based querying won't bring results because SQL is evaluated with non-str object representation, e.g. `WHERE type = '<TagType.STALE>'`.

Normalise `TagType` interfacing:
- `TextChoices` provides reliable ergonomics
- Tests are dumbimised and stick to bare value (ensures contract)

Running tests.

### DIFF
--- a/api/features/feature_external_resources/models.py
+++ b/api/features/feature_external_resources/models.py
@@ -110,7 +110,7 @@ class FeatureExternalResource(LifecycleModelMixin, models.Model):  # type: ignor
                     label=tag_by_type_and_state[self.type][state],
                     project=self.feature.project,
                     is_system_tag=True,
-                    type=TagType.GITHUB.value,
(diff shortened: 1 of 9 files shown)


khvn26 (approved):
LGTM!

Any other enums we should take care of? (doesn't have to be this PR, but having an issue would be nice if anything's left)

emyller:
> Any other enums we should take care of? (doesn't have to be this PR, but having an issue would be nice if anything's left)

Good call. Found a couple: https://github.com/Flagsmith/flagsmith/issues/7878

There's a bunch of enums — not `Choices` — that could be refactored, but I intentionally left those out of right now.

Issue #7771, opened 2026-06-12
MCP is missing error reporting


emyller:
We need Sentry!

khvn26:
This should cover adding the Sentry SDK, and initialising it for traceback/error capture.

Tracing push for new services should happen via OTLP (therefore, our otel collector).

PR #7879, opened 2026-06-25
feat(MCP): Set up Sentry error reporting

emyller:

## Changes

Contributes to https://github.com/flagsmith/flagsmith/issues/7771

Set up Sentry in the MCP server for error reporting.

New tests.


mcp/src/flagsmith_mcp/config.py
```diff
@@ -9,6 +9,10 @@
 class Settings(BaseSettings):
     model_config = {"use_attribute_docstrings": True}
 
+    environment: Literal["local", "dev", "staging", "production"] = Field(
```
khvn26: I don't think we need to constrain this

```suggestion
    environment: str = Field(
```

mcp/src/flagsmith_mcp/config.py
```diff
@@ -47,6 +51,10 @@ class Settings(BaseSettings):
     )
     """Public base URL of this MCP server, advertised in OAuth protected-resource
     metadata. Override for HTTP deployments behind a proxy/public hostname."""
+    sentry_dsn: str | None = Field(
+        default=None,
+    )
```
khvn26: nit: This can use [HttpUrl](https://pydantic.dev/docs/validation/2.0/api/pydantic/networks/#pydantic.networks.HttpUrl)

mcp/src/flagsmith_mcp/config.py
```diff
@@ -9,6 +9,10 @@
 class Settings(BaseSettings):
     model_config = {"use_attribute_docstrings": True}
 
+    environment: Literal["local", "dev", "staging", "production"] = Field(
+        default="local",
```
khvn26: If we keep this default, I expect the Docker build to set it to something else. But IMO we simply shouldn't set it at all.

khvn26 (commented):
Just a couple of comments.

mcp/src/flagsmith_mcp/config.py
emyller: I can't see why not, but can't say we _need_. 91d5ceb1c58ed641397fc956bdaeb857fca7bce3

mcp/src/flagsmith_mcp/config.py
emyller: Nice! 96bc1bd4ba9b326bbc5bb522a002a254f795a0cd

mcp/src/flagsmith_mcp/config.py
emyller: I think I intended `environment` as first-class interface citizen, but I realise it's not, for now. Thanks! 5eff3b268

mcp/src/flagsmith_mcp/config.py
khvn26: > I can't see why not

we use special env names for private cloud installs.

thanks!

mcp/src/flagsmith_mcp/config.py
khvn26: I don't think it needs to be. Thanks!

khvn26 (dismissed):
LGTM

mcp/src/flagsmith_mcp/config.py
khvn26: Wait, sorry, is there a reason you didn't use it for the DSN field?

emyller: None other than post-EOD LLM floppiness. acfe484ea

khvn26 (approved):

PR #7893, opened 2026-06-26
feat(experimentation): support multi-status filter on experiments list

Zaimwa9:
## Changes

Contributes to experimentation feature

The experiments list endpoint now accepts multiple `status` values as repeated query params (`?status=running&status=paused`). Previously only a single status string was accepted.

- `get_queryset` uses `getlist("status")` and `filter(status__in=...)` — fully backwards compatible with single-value usage
- Validation reports all invalid values at once

- Existing parametrised single-status filter tests pass unchanged (backwards compatible)
- Added `test_get_list__filter_by_multiple_statuses__returns_matching`
- Added `test_get_list__mixed_valid_and_invalid_status__returns_400`
- Backend mypy passes

### DIFF
--- a/api/experimentation/views.py
+++ b/api/experimentation/views.py
@@ -43,6 +43,7 @@ from experimentation.serializers import (
     ExperimentExposuresSerializer,
     ExperimentListSerializer,
     ExperimentMetricSerializer,
+    ExperimentQueryParamSerializer,
     ExperimentResultsSerializer,
     ExperimentRolloutSerializer,
     ExperimentSerializer,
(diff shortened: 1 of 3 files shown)


api/experimentation/views.py
```diff
@@ -245,13 +195,14 @@ def get_queryset(self) -> "QuerySet[Experiment]":
                 "feature__feature_states__multivariate_feature_state_values",
                 "experiment_metrics__metric",
             )
-        status_filter = self.request.query_params.get("status")
+        status_filter = self.request.query_params.getlist("status")
```
gagantrivedi: nit: we can move this to a query param serializer

gagantrivedi (dismissed):

gagantrivedi (approved):

PR #7895, opened 2026-06-26
feat(experiments): display rollout on detail page with inline edit

Zaimwa9:

## Changes

Frontend changes to the experiment detail page rollout configuration:

- Rename "Sample Size" to "Rollout" in the experiment creation wizard
- Display rollout percentage in the Variation Split card on the experiment detail page (reads from existing `experiment_rollout` response)
- Add inline edit mode (pen icon) for rollout and variation split with underline inputs and Save/Cancel controls
- Add `updateExperimentRollout` RTK Query mutation targeting the existing `PATCH /experiments/{id}/rollout/` endpoint
- Add `ExperimentRollout` type to match the backend response shape

1. Open the experiment creation wizard — verify the step is now labelled "Rollout" instead of "Sample Size"
2. Navigate to an experiment detail page — verify the Variation Split card shows a "Rollout" row with the percentage above the variant rows
3. Click the pen icon on the Variation Split card — verify inputs switch to underline style for rollout % and each variation %
4. Edit values and click Save — verify the PATCH call fires to `/rollout/` and the card updates
5. Click Cancel — verify inputs revert to read-only with original values
6. Verify control percentage auto-calculates and Save is disabled when split exceeds 100%

image
image
image


Zaimwa9:
@gemini-code-assist review

frontend/web/components/experiments/results/ExperimentConfiguration.tsx
```diff
@@ -65,22 +122,127 @@ const ExperimentConfiguration: FC<ExperimentConfigurationProps> = ({
...
-          <div className='d-flex flex-column gap-2'>
-            {identities.map((v, i) => (
-              <div
-                key={v.key}
-                className='d-flex align-items-center justify-content-between'
+        <ContentCard
+          compact
+          title='Variation Split'
+          action={
+            !isEditing ? (
```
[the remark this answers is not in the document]
Zaimwa9: already handled. In the (edge) case a rollout doesn't exist. The user will be able to edit as he wants and it will create the record

frontend/web/components/experiments/CreateExperimentWizard.tsx
```diff
@@ -100,7 +100,7 @@ const CreateExperimentWizard: FC<CreateExperimentWizardProps> = ({
 
   const controlPercentage = getControlPercentage(variationSplit)
   const isRolloutValid =
-    rolloutPercentage > 0 && controlPercentage >= 0 && controlPercentage <= 100
+    rolloutPercentage >= 0 && controlPercentage >= 0 && controlPercentage <= 100
```
talissoncosta: Can we use the `isValidPercentage` helper ?

frontend/web/components/experiments/results/ExperimentConfiguration.tsx
```diff
@@ -65,22 +146,131 @@ const ExperimentConfiguration: FC<ExperimentConfigurationProps> = ({
         </ContentCard>
       </div>
       <div className='col-md-4'>
-        <ContentCard compact title='Variation Split'>
-          <div className='d-flex flex-column gap-2'>
-            {identities.map((v, i) => (
-              <div
-                key={v.key}
-                className='d-flex align-items-center justify-content-between'
+        <ContentCard
```
talissoncosta: Suggestion

Have you consider extracting this card and all the complexity related to it to its own file, e.g. `ExperimentRolloutCard` ?

frontend/web/components/experiments/results/ExperimentConfiguration.tsx
```diff
@@ -65,22 +146,131 @@ const ExperimentConfiguration: FC<ExperimentConfigurationProps> = ({
...
+              if (!isEditing) {
+                currentPct = Math.round(getAllocation(i))
+              } else if (isControl) {
+                currentPct = Math.max(0, draftControlPct)
+              } else {
+                currentPct =
+                  draftSplit.find(
+                    (s) => s.multivariate_feature_option === optionId,
+                  )?.percentage_allocation ?? 0
+              }
```
talissoncosta: Do you think would worth to extract that logic into a function and test it to make sure the behaviour?

frontend/web/components/experiments/results/ExperimentConfiguration.tsx
```diff
@@ -65,22 +146,131 @@ const ExperimentConfiguration: FC<ExperimentConfigurationProps> = ({
...
+                    <span>{v.name}</span>
+                  </span>
+                  {isEditing && !isControl ? (
+                    <span className='d-flex align-items-center gap-1'>
+                      <Input
+                        type='number'
+                        size='xSmall'
+                        underline
+                        centered
+                        style={{ width: 56 }}
```
talissoncosta: Suggestion

Can we avoid this inline style ?

I've tested with `w-25` and it looks good, and using `justify-content-end` on the span will keep the same behaviour.

```suggestion
                    <span className='d-flex align-items-center gap-1 justify-content-end'>
                      <Input
                        type='number'
                        size='xSmall'
                        underline
                        centered
                        className='w-25'
```

This is not a blocker. I will leave at you to decided

talissoncosta (dismissed):
Looks good @Zaimwa9 !
None of comments are blocker. However, if we can extract the `ExperimentRolloutCard` it will reduce the complexity of `ExperimentConfiguration` and make the maintenance easier in the future.

Zaimwa9:
Addressed the feedback thanks! @talissoncosta

talissoncosta (approved):
Thanks for addressing the changes!

PR #7905, opened 2026-06-29
feat(experimentation): default to control when rollout segment is created

gagantrivedi:

## Changes

When a feature is first added to an experiment's rollout — the moment the rollout system segment is created — zero every variant's allocation on the environment-default feature state, so identities outside the rollout cohort receive control (the unallocated remainder) while the experiment runs.

- The rollout segment override is unchanged: it keeps the experiment's own split.
- Only runs when the rollout segment is first created; later rollout updates leave the default allocations alone, so manual edits to the default survive.
- Goes through the version-aware `update_flag` path, working under both v1 and v2 versioning.

Unit tests (`make test`): service tests covering first-rollout zeroing and later updates leaving the default alone; view tests asserting create-with-rollout zeroes the default and preserves the override split under both v1 and v2 versioning. `mypy`, `ruff`, and `flagsmith-lint-tests` clean.

### DIFF
--- a/api/experimentation/services.py
+++ b/api/experimentation/services.py
@@ -709,6 +709,8 @@ def _serialize_feature_state_value(
 ) -> tuple[str, FeatureValueType]:
     """Render a stored feature state value as the (string, API type) pair that
     a `FlagChangeSet` expects."""
+    if value.value is None:
+        return "", "string"
     return (
         str(value.value).lower() if value.type == BOOLEAN else str(value.value),
         _ROLLOUT_VALUE_TYPE.get(value.type or STRING, "string"),
--- a/api/experimentation/services.py
+++ b/api/experimentation/services.py
@@ -42,6 +42,7 @@ from experimentation.dataclasses import (
 )
 from experimentation.models import (
     VALID_STATUS_TRANSITIONS,
+    Experiment,
     ExperimentStatus,
     MetricAggregation,
     MetricDirection,
@@ -75,7 +76,7 @@ if typing.TYPE_CHECKING:
     from collections.abc import Sequence
     from datetime import datetime
 
-    from experimentation.models import Experiment, Metric, WarehouseConnection
+    from experimentation.models import Metric, WarehouseConnection
     from experimentation.types import ExposureGranularity
     from features.feature_states.models import FeatureValueType
     from features.models import FeatureStateValue
@@ -678,8 +679,9 @@ def apply_experiment_rollout(experiment: Experiment, spec: RolloutSpec) -> None:
         )
     validate_rollout_spec(experiment, spec)
     environment_id = experiment.environment_id
-    is_first_rollout = experiment.rollout_segment_id is None
     with transaction.atomic():
+        locked_experiment = Experiment.objects.select_for_update().get(pk=experiment.pk)
+        is_first_rollout = locked_experiment.rollout_segment_id is None
         segment = _sync_rollout_segment(experiment, spec.rollout_percentage)
         if is_first_rollout:
             _reset_default_allocations_to_control(experiment, spec.author)
--- a/api/experimentation/services.py
+++ b/api/experimentation/services.py
@@ -680,8 +680,8 @@ def apply_experiment_rollout(experiment: Experiment, spec: RolloutSpec) -> None:
     validate_rollout_spec(experiment, spec)
     environment_id = experiment.environment_id
     with transaction.atomic():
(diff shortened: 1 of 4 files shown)


api/experimentation/services.py
```diff
@@ -627,15 +633,56 @@ def _update_rollout_in_place(experiment: Experiment, change_set: FlagChangeSet)
...
         raise ValidationError(
             f"Cannot change the rollout of a {experiment.status} experiment."
         )
     validate_rollout_spec(experiment, spec)
     environment_id = experiment.environment_id
+    is_first_rollout = experiment.rollout_segment_id is None
     with transaction.atomic():
         segment = _sync_rollout_segment(experiment, spec.rollout_percentage)
+        if is_first_rollout:
+            _reset_default_allocations_to_control(experiment, spec.author)
```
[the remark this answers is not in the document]
gagantrivedi: The reset is idempotent

api/tests/unit/experimentation/test_services.py
```diff
@@ -1359,6 +1359,99 @@ def test_apply_experiment_rollout__no_segment__creates_segment_and_override(
...
+            ],
+            author=AuthorData(user=admin_user),
+        ),
+    )
+
+    # Then
+    allocation = default_state.multivariate_feature_state_values.get(
+        multivariate_feature_option=option_a
+    ).percentage_allocation
+    assert allocation == 25.0
```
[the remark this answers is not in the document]
gagantrivedi: Yeah, good call

api/experimentation/services.py
```diff
@@ -635,7 +680,11 @@ def apply_experiment_rollout(experiment: Experiment, spec: RolloutSpec) -> None:
     validate_rollout_spec(experiment, spec)
     environment_id = experiment.environment_id
     with transaction.atomic():
+        locked_experiment = Experiment.objects.select_for_update().get(pk=experiment.pk)
+        is_first_rollout = locked_experiment.rollout_segment_id is None
         segment = _sync_rollout_segment(experiment, spec.rollout_percentage)
+        if is_first_rollout:
+            _reset_default_allocations_to_control(experiment, spec.author)
```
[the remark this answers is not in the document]
gagantrivedi: Done

gagantrivedi:
@coderabbitai review

api/experimentation/services.py
```diff
@@ -655,6 +704,17 @@ def apply_experiment_rollout(experiment: Experiment, spec: RolloutSpec) -> None:
         )
 
 
+def _serialize_feature_state_value(
+    value: FeatureStateValue,
+) -> tuple[str, FeatureValueType]:
+    """Render a stored feature state value as the (string, API type) pair that
+    a `FlagChangeSet` expects."""
+    return (
+        str(value.value).lower() if value.type == BOOLEAN else str(value.value),
```
Zaimwa9, line 716: This might be return a `"None"` for null values no ?

api/experimentation/services.py
```diff
@@ -627,6 +634,44 @@ def _update_rollout_in_place(experiment: Experiment, change_set: FlagChangeSet)
...
+    feature state, leaving control (the unallocated remainder) at 100%.
+
+    Run once, when the rollout segment is first created: identities outside the
+    rollout cohort should all receive control while the experiment runs.
+    """
+    default_state = FeatureState.objects.get_live_feature_states(
+        environment=experiment.environment,
+        additional_filters=Q(feature_segment__isnull=True, identity__isnull=True),
+        feature_id=experiment.feature_id,
+    ).latest("id")
```
Zaimwa9: Mmh, it's a bit edgy so let's NIT it (and already existing).
The `latest("id"`) would drift if ever there are unordered scheduled changes:
- CR A scheduled in 30 minutes
- CR B scheduled in 15 minutes
- They both go live
==> We would get CR B instead of A

I honestly don't think it would happen in real life so if the fix is too complicated we can leave a comment and not block the PR

Zaimwa9 (changes requested):
Looks good overall but we should fix the `null => "None"` before moving forward

api/experimentation/services.py
gagantrivedi: That's a good shout! I should've used` get_environment_flags_list`

api/experimentation/services.py
gagantrivedi: The value will never be None(should not be at the very least) for a feature connected to an experiment... I have added a small guard in the meantime

Zaimwa9 (approved):
👍

PR #7912, opened 2026-06-30
fix(experimentation): keep variant assignment stable across rollout updates

gagantrivedi:

## Changes

Under v2 versioning, updating an experiment's rollout went through `update_flag`, which clones the segment override into a new feature state each time. The multivariate split is salted on the feature state id, so every update re-randomised control/variant for already-enrolled identities, corrupting the experiment data.

Now, once the override exists, it's updated in place (id preserved → stable split) and the environment document is rebuilt directly. Creating the override, and all v1 versioning, still go through `update_flag`.

Added `test_apply_experiment_rollout__reapplied_under_v2__keeps_variant_assignment` (fails on `main`, passes here). Full experimentation + versioning unit suites and `make typecheck` pass.

### DIFF
--- a/api/experimentation/services.py
+++ b/api/experimentation/services.py
@@ -17,6 +17,7 @@ from rest_framework.exceptions import ValidationError
 from audit.models import AuditLog
 from audit.related_object_type import RelatedObjectType
 from core.dataclasses import AuthorData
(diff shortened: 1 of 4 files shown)


Zaimwa9 (approved):
Good as a temporary solution

PR #7914, opened 2026-06-30
fix(features): preserve multivariate bucketing salt across feature state recreation

### LINKED ISSUE #7913: MV variant bucketing reshuffles when feature state is recreated (v2 versioning)
MV variant bucketing is salted on the **feature-state pk**: `FeatureState.get_multivariate_feature_state_value` hashes `[self.id, identity_hash_key]` (`api/features/models.py:771`). Any flow that creates a *new* feature state changes the salt, so every identity re-buckets across variants.

Two triggers:

1. **Enabling versioning**
2. **Editing MV weights under v2**

**Repro:** v2 env, MV flag 50/50 → 60/40: some identities that received `variant_1` under 50% switch to control, despite variant_1's allocation increasing.

gagantrivedi:

## Changes

Closes #7913

Multivariate bucketing is salted on the feature state id, so any flow that
recreates a feature state (version publish, weight edits, change request
commits) re-bucketed already-enrolled identities.

- Add `FeatureState.mv_hashing_salt` (nullable, internal — not exposed via API
  or audit logs). Bucketing seeds on the new `mv_hashing_seed` property
  (`salt or id`); no backfill needed.
- A `BEFORE_CREATE` hook makes any new multivariate feature state inherit the
  seed from the live state it supersedes, so every creation path (change
  request drafts, direct recreates, version edits) keeps bucketing stable
  without knowing the salt exists. `clone()` copies it like any other field.
- Engine mapper emits the seed as `django_id`, so Core, Edge and local-eval
  SDKs bucket identically with no document schema change.

Known gap: v1 change request drafts already open when this deploys predate the
salt and will re-bucket once when committed — that is today's behaviour, one
last time; not worth a backfill migration.

Unit tests: seed fallback, clone preservation (50 identities keep their
variant), engine mapper output, seed inheritance for v1/v2 direct recreates
and fresh lineages, and change request commits keeping variant assignment
stable end to end.


Zaimwa9:
@coderabbit review

api/features/models.py
```diff
@@ -818,6 +863,23 @@ def check_for_duplicate_feature_state(self):  # type: ignore[no-untyped-def]
...
             )
 
+    @hook(BEFORE_CREATE)
+    def inherit_mv_hashing_salt(self):  # type: ignore[no-untyped-def]
+        """Keep multivariate bucketing stable across recreation: a new row
+        superseding a live state in its lineage adopts that state's seed
+        (#7913). Identity overrides have no versioned lineage; the feature type
+        check confines the lineage query to rows where bucketing matters.
+        """
+        if (
```
Zaimwa9, line 873: NIT: We can add a cheap check on `environment_id is None` even though that shouldn't happen it could raise

api/util/mappers/engine.py
```diff
@@ -152,7 +152,11 @@ def map_feature_state_to_engine(
 
     return FeatureStateModel(
         enabled=feature_state.enabled,
-        django_id=feature_state.pk,
+        # The engine and SDKs seed multivariate variant allocation on django_id,
+        # so feeding it the bucketing seed keeps variant assignment stable when
+        # a feature state is recreated, without changing the engine model or
+        # environment document schema. See issue #7913.
+        django_id=feature_state.mv_hashing_seed,
```
Zaimwa9, line 159: 👍

Zaimwa9 (approved):
Ok great, good call to have pushed back and rethink the implementation.

I couldn't find flaws in it and it simplified _**a lot**_ the mental model. Approving with one nit that shouldn't happen normally so feel free to resolve and ignore

api/features/models.py
gagantrivedi: Nah, Let's let it crash if that happens

PR #7919, opened 2026-06-30
fix:  gate experiment freeze behind feature flag and unblock inputs during loading

Zaimwa9:

## Changes
- Inputs were frozen on experiment loading first load (even without flag)
- Skip entirely freeze hook if `experimental_flags` is disabled
- Removed `isLoading` from inputs

- Added tests
- Tested in parallel prod and local against prod


matthewelwell (approved):
Tested specific issue this sets out to resolve, and the code looks reasonable.

Issue #7593, opened 2026-05-25
Handle `is_deleted` attribute when querying for identities in ClickHouse


matthewelwell:
We need segment membership inspection to properly support identity deletes.

### Acceptance Criteria

- [x] Rows can be written to the identities table with `is_deleted` attribute
- [x] Rows with `is_deleted: true` do not appear in segment membership views
- [x] Identities UI deletes write ClickHouse rows with `is_deleted: true`, and trigger segment membership recounts

PR #7924, opened 2026-07-01
perf(Segment Membership): Members list improvements

khvn26:

## Changes

Contributes to #7593.

In this PR, we make two changes to the Segment Membership read path:

1. Drop `FINAL` in the list query. Adopt a two-stage `LIMIT 1 BY identifier` dedup and limit trait reads to a page.
2. Filter `is_deleted = false` in both read queries. Neither excluded soft-deleted identities before, so deleted identities appeared in the members list and were counted. 

Added new unit tests.

Verified the queries against production. A 52k-identity environment, default first page (no search term), `LIMIT 11`, returning 11 rows:

| Query shape | Warm latency | Warm bytes | Cold | Cold S3 GETs |
| --- | --- | --- | --- | --- |
| `FINAL` (before) | 2.7–4.2s | ~370 MB | 7.3s | 6,211 |
| No `FINAL` | 0.12s | 22 MB | 1.5s | 816 |
| Single-stage `LIMIT 1 BY` | 0.13–0.22s | 160 MB | 1.8s | 1,601 |
| Two-stage `LIMIT 1 BY` (this PR) | ~0.10s | 43 MB | 1.9s | 1,602 |

### DIFF
--- a/api/segment_membership/services.py
+++ b/api/segment_membership/services.py
@@ -129,7 +129,8 @@ def compute_segment_counts_for_project(
             f"SELECT {seg.id} AS segment_id, "
             f"i.environment_id AS env_key, count() AS c "
             f"FROM IDENTITIES AS i FINAL "
-            f"WHERE i.environment_id IN %(env_keys)s AND ({predicate}) "
+            f"WHERE i.environment_id IN %(env_keys)s "
+            f"AND i.is_deleted = false AND ({predicate}) "
             f"GROUP BY i.environment_id"
         )
 
@@ -186,7 +187,10 @@ def get_segment_members_page(
         )
         return []
 
-    conditions = ["i.environment_id = %(env_key)s"]
+    conditions = [
+        "i.environment_id = %(env_key)s",
+        "i.is_deleted = false",
+    ]
     params: dict[str, Any] = {"env_key": environment.api_key, "limit": limit}
     if cursor:
         conditions.append("i.identifier > %(cursor)s")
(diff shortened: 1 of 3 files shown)


api/segment_membership/services.py
```diff
@@ -196,13 +200,29 @@ def get_segment_members_page(
...
-    sql = (
-        "SELECT i.identifier, i.identity_key, i.traits "
-        "FROM IDENTITIES AS i FINAL "
+    # Why two queries? `traits` is a wide JSON column split over many per-path files.
+    # That fans out into thousands of object-storage requests, which is what
+    # makes a cold read slow. Delegating the `traits` read to an outer query
+    # limits them to the page.
+    #
+    # Besides that, bypass FINAL, which is heavy, in favour of LIMIT 1 BY i.identifier,
+    # which is (tolerably) less correct but substantially faster.
```
matthewelwell, line 209: Sorry, I'd like a bit more information here about what the trade-off is?

matthewelwell (changes requested):
Marking as requested changes, but I'm happy for the change to be a response to my comment for documentation purposes, although ideally I'd like to see an elaboration on the comment.

api/segment_membership/services.py
khvn26: `FINAL` forces query-time deduplication for ReplacingMergeTree, which is otherwise asynchronous. ClickHouse is optimised for high volume writes this way. The trade-off is having occasional out-of-date rows in the results, but they still will be bounded by last recount, clearly stated in the UI. Potential duplicate rows are remedied by `LIMIT 1 by identifier`.

Resources:
- https://clickhouse.com/docs/engines/table-engines/mergetree-family/replacingmergetree
- https://clickhouse.com/docs/guides/replacing-merge-tree

matthewelwell (approved):

PR #7933, opened 2026-07-01
feat(Segment Membership): Recount after identity delete

### LINKED ISSUE #7593: Handle `is_deleted` attribute when querying for identities in ClickHouse

khvn26:

## Changes

Closes #7593.

In this PR, we trigger segment count refresh when an identity is deleted in UI.

Added unit tests. Will test E2E in staging.

### DIFF
--- a/api/app/settings/common.py
+++ b/api/app/settings/common.py
@@ -1484,6 +1484,12 @@ SEGMENT_MEMBERSHIP_REFRESH_INTERVAL_HOURS = env.int(
     "SEGMENT_MEMBERSHIP_REFRESH_INTERVAL_HOURS", default=6
 )
 
+# Edge CDC writes the `is_deleted` tombstone to ClickHouse asynchronously; delay
+# the post-delete recount by this long so it reads state after the tombstone lands.
+SEGMENT_MEMBERSHIP_DELETE_REFRESH_DELAY_SECONDS = env.int(
+    "SEGMENT_MEMBERSHIP_DELETE_REFRESH_DELAY_SECONDS", default=120
+)
+
 # Always installed: the router fences the `clickhouse` app's migrations off
 # the default Postgres database whether or not a CH alias is configured.
 DATABASE_ROUTERS.append("app.routers.ClickHouseRouter")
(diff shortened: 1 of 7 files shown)


api/app/settings/common.py
```diff
@@ -1483,6 +1483,9 @@
 SEGMENT_MEMBERSHIP_REFRESH_INTERVAL_HOURS = env.int(
     "SEGMENT_MEMBERSHIP_REFRESH_INTERVAL_HOURS", default=6
 )
+SEGMENT_MEMBERSHIP_DELETE_REFRESH_DELAY_SECONDS = env.int(
+    "SEGMENT_MEMBERSHIP_DELETE_REFRESH_DELAY_SECONDS", default=120
+)
```
emyller, line 1489: question: Considering we already have a debounce mechanism, why the delay here?

emyller (commented):
One comment, otherwise LGTM.

api/app/settings/common.py
emyller: Actually I think I favour this logic if this reading is correct: given I delete a few identities in sequence, I only want one recount after I'm done, with 120s of grace.

khvn26: 120s is Edge CDC write-to-visible SLO. I believe the debounce logic should work as you expect with the scheduled task invocations.

emyller: Gotcha, thanks. Can we please:

1. Make the reason why the number matches the SLO explicit in a comment, and/or use another name directly, e.g. `WRITE_TO_VISIBLE_SLO`, instead of reading from an environment variable. Looking at the code alone, I wouldn't know why `SEGMENT_MEMBERSHIP_DELETE_REFRESH_DELAY_SECONDS` exists and how/when to tweak it.
2. Add a test case ensuring the behaviour we expect.

khvn26: Actually, I revisited the debounce logic to make sure that we account for the latest deletion, not the first: b34d5eb4fa176ea43adbd6ee328f83119b3a6870

This includes the setting doc (I don't agree with `SLO` naming and chose adding a comment instead), and the tests.

Let me know if that makes sense to you.

emyller: > Actually, I revisited the debounce logic to make sure that we account for the latest deletion, not the first: [b34d5eb](https://github.com/Flagsmith/flagsmith/commit/b34d5eb4fa176ea43adbd6ee328f83119b3a6870)

I like the solution, thanks.

> This includes the setting doc (I don't agree with `SLO` naming and chose adding a comment instead), and the tests.

The comment changes read much better!

emyller (approved):
LGTM

PR #7936, opened 2026-07-02
fix(usage): Send raw licence signature bytes in Control Plane auth token

germangarces:
The usage-push auth token was encoded twice, so the Control Plane never recognised it and rejected every push. This decodes the licence signature back to its raw bytes before encoding it for the request header, so the token matches what the Control Plane expects.

### DIFF
--- a/api/organisations/usage_reporting/mappers.py
+++ b/api/organisations/usage_reporting/mappers.py
@@ -18,9 +18,8 @@ MAX_PROJECT_USAGE_ROWS = 5_000
 
 
 def map_signature_to_control_plane_auth_token(signature: str) -> str:
-    return (
-        base64.urlsafe_b64encode(signature.encode("utf-8")).decode("ascii").rstrip("=")
-    )
+    raw_signature = base64.b64decode(signature)
+    return base64.urlsafe_b64encode(raw_signature).decode("ascii").rstrip("=")
 
 
 def map_usage_data_to_total_api_calls(usage_data: list[UsageData]) -> int:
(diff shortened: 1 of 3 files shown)


api/organisations/usage_reporting/mappers.py
```diff
@@ -18,9 +18,8 @@
 
 
 def map_signature_to_control_plane_auth_token(signature: str) -> str:
-    return (
-        base64.urlsafe_b64encode(signature.encode("utf-8")).decode("ascii").rstrip("=")
-    )
+    raw_signature = base64.b64decode(signature)
```
matthewelwell: This could perhaps have been avoided (and improved for the future) by better variable naming. If the signature being passed into this function is `b64` encoded, then it should probably be named `b64_signature`. Alternatively, could we decode it _before_ passing it into this function?

matthewelwell (changes requested):

api/organisations/usage_reporting/mappers.py
germangarces: I've renamed it in c99614c1c.
I don't really like decoding it before passing it. Current approach keeps the logic in one spot, instead of leaking it out into the calling method

matthewelwell: Ok, but it doesn't really, right? If we follow the code back, it's named as `signature` everywhere, but actually it's the base64 encoded signature. 

It looks like it originally comes from the `OrganisationLicence` model [here](https://github.com/Flagsmith/flagsmith/blob/f5b627301af6ef1860a310f415a18e153a80e2b6/api/organisations/usage_reporting/services.py#L72). 

I think the _proper_ solution would be to rename the field on the model, and wherever the variable is used, but that expands the scope into flagsmith-private as well. 

I think we can probably limit the scope to this repository, however, but at least change it throughout the calling path here? i.e. update the argument in `push_snapshot` as well?

germangarces: good call. done in d32fcb09e

matthewelwell (approved):

PR #7941, opened 2026-07-02
feat(onboarding): "next quest" cards

### LINKED ISSUE #7739: Onboarding: make the "next quest" cards actionable
Part of #7737. The bottom of the single-page flow.

After the first flag works, the flow offers "next quest" cards (kill switch / gradual rollout / experiment / remote config). Today they're illustrative previews.

## Scope
- Build the next-quest-card components (prop-driven) with stories.
- Wire each card to its real next step for the flag (`show_demo_button`) — % rollout, multivariate, remote-config value, etc.

## Done when
- Cards are standalone prop-driven components rendered by the page, with stories.
- Each card navigates to the real configuration for the flag; nothing faked.

talissoncosta:

## Changes

Closes #7739

The three "next quest" cards on the onboarding flow (gradual rollout, experiment, remote config) are now actionable. Each one deep-links to the flag's real configuration, nothing faked:

- **Gradual rollout** → the flag's segment overrides tab
- **Experiment** → the environment's experiments page
- **Remote config** → the flag's value tab

The cards stay locked and dimmed until the app connects, matching the flags table above them.

Also folded in two small fixes for the cards: the rollout deep-link now uses the slugified tab label so it resolves to the right tab, and the cards' secondary text no longer renders gold (a Bootstrap `.text-secondary` collision).

Review follow-ups: the `cn` import alignment (Wadii) lands here; the connect-panel, spacing and inline-edit polish (Kyle, Wadii, Matt) is in the stacked #7989.

Added e2e coverage in `onboarding-tests.pw.ts`: it asserts the cards are locked before the first evaluation, and that once connected each card deep-links to the right destination (segment overrides tab, value tab, experiments page).

Manually:

1. Enable `onboarding_quickstart_flow`, sign up, land on `/getting-started`.
2. Before connecting, confirm the next-quest cards are dimmed with an "Unlocks after your first evaluation" note.
3. Force the connected state via `/getting-started?connected`.
4. Click each card and confirm it lands on the flag's segment overrides tab, the flag's value tab, and the experiments page respectively.

### DIFF
--- a/frontend/web/components/pages/onboarding/OnboardingNextSteps/OnboardingNextSteps.tsx
+++ b/frontend/web/components/pages/onboarding/OnboardingNextSteps/OnboardingNextSteps.tsx
@@ -1,0 +1,120 @@
+import React, { FC, ReactNode } from 'react'
+import classNames from 'classnames'
+import Icon from 'components/icons/Icon'
+import SelectableCard from 'components/base/SelectableCard/SelectableCard'
+import './OnboardingNextSteps.scss'
+
+export type OnboardingNextStep = 'rollout' | 'experiment' | 'remote-config'
+
+export type OnboardingNextStepsProps = {
+  // Locked (dimmed, non-interactive) until the app connects, like the flags
+  // table - the steps only make sense once the flag is live.
+  locked: boolean
+  onSelect: (step: OnboardingNextStep) => void
+}
+
+// A 25%-filled track: "ship to a % of your users".
+const RolloutPreview: FC = () => (
+  <span className='onboarding-next-steps__track'>
+    <span className='onboarding-next-steps__track-fill' />
+  </span>
+)
+
+// An A/B segmented control, A selected.
+const ExperimentPreview: FC = () => (
+  <span className='onboarding-next-steps__segmented d-inline-flex'>
+    <span className='onboarding-next-steps__segment onboarding-next-steps__segment--active'>
+      A
+    </span>
+    <span className='onboarding-next-steps__segment'>B</span>
+  </span>
+)
+
+// A value chip: "serve a value, not just on/off".
+const RemoteConfigPreview: FC = () => (
+  <span className='onboarding-next-steps__value-pill bg-surface-action-subtle text-action'>
+    {'"theme": "dark"'}
+  </span>
+)
+
+const STEPS: {
+  key: OnboardingNextStep
+  title: string
+  description: string
+  preview: ReactNode
+  caption?: string
+}[] = [
+  {
+    caption: '25% of users',
+    description: 'Ship to a % of your users.',
+    key: 'rollout',
+    preview: <RolloutPreview />,
+    title: 'Gradual rollout',
+  },
+  {
+    description: 'A/B test which variant wins.',
+    key: 'experiment',
+    preview: <ExperimentPreview />,
+    title: 'Experiment',
+  },
+  {
+    description: 'Serve a value, not just on/off.',
+    key: 'remote-config',
+    preview: <RemoteConfigPreview />,
+    title: 'Remote config',
+  },
+]
+
+// "Choose your next quest": the three ways the demo flag can level up, each
+// linking to its real config (the page owns where). Locked until the app
+// connects.
+const OnboardingNextSteps: FC<OnboardingNextStepsProps> = ({
+  locked,
+  onSelect,
+}) => (
+  <section className='onboarding-next-steps d-flex flex-column gap-2'>
+    {locked && (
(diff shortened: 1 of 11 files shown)


kyle-ssg:
[WIP Feedback] Some of this might just be generic onboarding feedback, I think it looks so good!

Minor: I think this means it's incorrectly always overflow scroll?

image

Let's just replace this with the selected badges, they look really good, though maybe make them fw-semibold

image

image

The code here doesnt use useFlagsmith or flagsmith imports

image


i.e. 

```
// App root
import flagsmith from "@flagsmith/flagsmith";
import { FlagsmithProvider, useFlags } from '@flagsmith/flagsmith/react';

export function HomePage() {
  const flags = useFlags(['show_demo_button','banner_size']); // only causes re-render if specified flag values / traits change
  const show_demo_button = flags.show_demo_button.enabled
  const banner_size = flags.banner_size.value
  return (
      <>
        {`show_demo_button: ${show_demo_button}`}
        {`banner_size: ${banner_size}`}
      </>
  );
}
export default function App() {
  return (
      <FlagsmithProvider
          options={{
            environmentID: '2ULH8Sq9WxofQsJQc2Cbeg',
          }}
          flagsmith={flagsmith}>
          <HomePage/>
      </FlagsmithProvider>
  );
}
```

I think we should reuse what you had and remove codehelp language select

image

frontend/web/components/pages/onboarding/OnboardingNextSteps/OnboardingNextSteps.tsx
```diff
@@ -0,0 +1,120 @@
...
+  <span className='onboarding-next-steps__segmented d-inline-flex'>
+    <span className='onboarding-next-steps__segment onboarding-next-steps__segment--active'>
+      A
+    </span>
+    <span className='onboarding-next-steps__segment'>B</span>
+  </span>
+)
+
+// A value chip: "serve a value, not just on/off".
+const RemoteConfigPreview: FC = () => (
```
Zaimwa9, line 34: I'm not 100% sure that's where it belongs but i'd add some more margin/padding top here

Image

Zaimwa9: Confirmed it's not here :D

frontend/web/components/pages/onboarding/OnboardingNextSteps/OnboardingNextSteps.tsx
```diff
@@ -0,0 +1,120 @@
+import React, { FC, ReactNode } from 'react'
+import classNames from 'classnames'
```
Zaimwa9: NIT: it's imported as cn in `SelectableCard`

Zaimwa9 (commented):
I have some nits, but I think Kyle's comments are worth addressing, although the last one would be a NIT to me that I let you decide what do with 👍

talissoncosta:
Thanks both 🙏

This PR
- classnames as cn: aligned it with SelectableCard (3292d244a).

Follow-up PR (keeping this one focused): the connect step plus the spacing nit.
- Top spacing on OnboardingNextSteps: will tidy it there.
- Always scrolling: switched to overflow-x: auto so it only scrolls when it needs to.
- Language labels: replaced with the selected badges, fw-semibold.
- React snippet: now uses FlagsmithProvider + useFlags, matching your example.
- Reuse CodeHelp + drop its language select: the meaty one. CodeHelp is dark-only today, so it needs the shared code theme to go light/dark first (spike in #7969). Happy to treat it as a follow-up, as Wadii suggested.

Shout if you'd rather I fold these into [this PR](https://github.com/Flagsmith/flagsmith/pull/7989) instead.

frontend/web/components/pages/onboarding/OnboardingNextSteps/OnboardingNextSteps.tsx
talissoncosta: Good call. Fixed in 3292d244a0423dc521e82faaf522d8946b8395be

frontend/web/components/pages/onboarding/OnboardingNextSteps/OnboardingNextSteps.tsx
talissoncosta: I will handle that on a follow up pr

talissoncosta: I am handling it on this [PR](https://github.com/Flagsmith/flagsmith/pull/7989)

Zaimwa9 (approved):
Nothing to add there on my side here, happy to approve, will review the follow up

PR #7945, opened 2026-07-06
fix(API): Nested viewsets expose data across projects and organisations

khvn26:

## Changes

In this PR, we fix four vulnerabilites:

1. `MultivariateFeatureOptionViewSet` resolved the feature by bare `feature_pk`, so a user could read another project's multivariate options. Scoped the lookup to the URL's project, mirroring `create`.
2. `FeatureExternalResourceViewSet` filtered the list by bare `feature_pk`, and `FeaturePermissions` applied no project check on `list`. Scoped the lookup to the caller's permitted projects, matching `FeatureViewSet`.
3. `UserPermissionGroupEnvironmentPermissionsViewSet` gated only by `IsAuthenticated`, unlike its sibling `UserEnvironmentPermissionsViewSet`. Because the environment client-side api key is not a secret, any authenticated user could read and delete another organisation's environment group permission assignments. Added the missing `NestedEnvironmentPermissions`.
4. `GithubRepositoryViewSet` filtered by bare `github_pk` while the permission only checked membership of the URL's `organisation_pk`, neither tying the configuration to that organisation. Validated the configuration belongs to the organisation in `HasPermissionToGithubConfiguration`, covering every action.

Added regression tests.

### DIFF
--- a/api/integrations/github/permissions.py
+++ b/api/integrations/github/permissions.py
@@ -1,5 +1,7 @@
 from rest_framework.permissions import BasePermission
 
+from integrations.github.models import GithubConfiguration
+
 
 class HasPermissionToGithubConfiguration(BasePermission):
     """
@@ -10,4 +12,17 @@ class HasPermissionToGithubConfiguration(BasePermission):
     def has_permission(self, request, view):  # type: ignore[no-untyped-def]
         organisation_id = view.kwargs.get("organisation_pk")
 
-        return request.user.belongs_to(organisation_id=int(organisation_id))
+        if not request.user.belongs_to(organisation_id=int(organisation_id)):
+            return False
+
+        # For nested routes, ensure the GitHub configuration in the URL belongs
+        # to the organisation the caller is authorised for
+        if (github_pk := view.kwargs.get("github_pk")) is not None and str(
+            github_pk
+        ).isdigit():
+            return GithubConfiguration.objects.filter(
+                id=github_pk,
+                organisation_id=organisation_id,
+            ).exists()
+
+        return True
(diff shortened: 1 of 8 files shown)


api/integrations/github/permissions.py
```diff
@@ -10,4 +12,17 @@ class HasPermissionToGithubConfiguration(BasePermission):
...
+        # to the organisation the caller is authorised for
+        if (github_pk := view.kwargs.get("github_pk")) is not None and str(
+            github_pk
+        ).isdigit():
+            return GithubConfiguration.objects.filter(
+                id=github_pk,
+                organisation_id=organisation_id,
+            ).exists()
+
+        return True
```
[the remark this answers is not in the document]
khvn26: It'll raise `ValidationError` later, resulting in a 400, which is covered by `test_get_github_repository__github_pk_not_a_number__returns_400`.

api/integrations/github/permissions.py
```diff
@@ -10,4 +12,17 @@ class HasPermissionToGithubConfiguration(BasePermission):
...
+
+        # For nested routes, ensure the GitHub configuration in the URL belongs
+        # to the organisation the caller is authorised for
+        if (github_pk := view.kwargs.get("github_pk")) is not None and str(
+            github_pk
+        ).isdigit():
+            return GithubConfiguration.objects.filter(
+                id=github_pk,
+                organisation_id=organisation_id,
+            ).exists()
```
emyller, line 26:
```suggestion
        if (github_pk := view.kwargs.get("github_pk")):
            return GithubConfiguration.objects.filter(
                id=github_pk,
                organisation_id=organisation_id,
            ).exists()
```

emyller (approved):
LGTM

api/integrations/github/permissions.py
khvn26: It's nicer code, but it'll cause a 500 for non-digit inputs. Sucks to duplicate validation logic, but there's really no way around it. We can thank Django REST for that, I guess.

emyller: Sad times.

khvn26:
Thanks for reporting, @dmitrymaranik!

PR #7957, opened 2026-07-07
fix: return 400 when environment is null in featurestate update

### LINKED ISSUE #6597: 500 when calling /api/v1/features/featurestates/{pk}/ with empty environment
Sentry Issue: [FLAGSMITH-API-5GQ](https://flagsmith.sentry.io/issues/7219818703/?referrer=github_integration)

```
AttributeError: 'NoneType' object has no attribute 'id'
(8 additional frame(s) were not displayed)
...
  File "features/serializers.py", line 578, in validate_environment
    if self.instance and self.instance.environment_id != environment.id:  # type: ignore[union-attr]
```

## Acceptance criteria

- `/api/v1/features/featurestates/{pk}/` responds with HTTP 400 if `environment` is null in the payload.

mangodxd:
## Description

Fixes a 500 crash in `/api/v1/features/featurestates/{pk}/` when the `environment` field is `null` in the request payload.

**Root cause:** `validate_environment` in `FeatureStateSerializerBasic` calls `environment.id` without checking if `environment` is `None` first, causing an `AttributeError: 'NoneType' object has no attribute 'id'`.

**Fix:** Added a null check at the beginning of `validate_environment` that returns a 400 validation error with the message `"Environment may not be null."` when the environment field is null.

## Changes

- `api/features/serializers.py`: Added null check in `validate_environment` method
- `api/tests/unit/features/test_unit_features_views.py`: Added regression test `test_update_feature_state__null_environment__returns_400`

## Related

- Fixes #6597
- Sentry: FLAGSMITH-API-5GQ

### DIFF
--- a/api/features/serializers.py
+++ b/api/features/serializers.py
@@ -728,6 +728,10 @@ class FeatureStateSerializerBasic(WritableNestedModelSerializer):
         return feature
 
     def validate_environment(self, environment):  # type: ignore[no-untyped-def]
+        if environment is None:
+            raise serializers.ValidationError(
+                "Environment may not be null."
+            )
         if self.instance and self.instance.environment_id != environment.id:  # type: ignore[union-attr]
             raise serializers.ValidationError(
                 "Cannot change the environment of a feature state"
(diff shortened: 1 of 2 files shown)


Zaimwa9 (approved):
Looking good to me. Thanks for your first contribution!

Issue #7958, opened 2026-07-07
OpenAPI schema omits OAuth 2.0 authentication


khvn26:
### Is your feature request related to a problem?

`OAuth2BearerTokenAuthentication` and `JWTCookieAuthentication` are in the global `DEFAULT_AUTHENTICATION_CLASSES`, so every authenticated endpoint accepts an OAuth2 bearer token and a JWT session cookie. Neither appears in the generated OpenAPI schema.

The reason is that drf-spectacular can only render an authenticator it has an `OpenApiAuthenticationExtension` for. We register extensions for the environment key and master API key, and the built-in token/basic auth resolve on their own, but there is none for `OAuth2BearerTokenAuthentication` or `JWTCookieAuthentication`. drf-spectacular warns and drops them at generation time:

```
Warning [WebhookViewSet]: could not resolve authenticator
<class 'oauth2_metadata.authentication.OAuth2BearerTokenAuthentication'>.
There was no OpenApiAuthenticationExtension registered for that class. Ignoring for now.
```

So the committed `openapi.yaml` under-documents auth: consumers can't see that OAuth2 auth is accepted, and the generation logs are noisy.

### Describe the solution you'd like.

Register an `OpenApiAuthenticationExtension` for `OAuth2BearerTokenAuthentication` (emitting an `oauth2` scheme with the authorisation-code flow — authorisation/token URLs and scopes derived from settings) and one for `JWTCookieAuthentication`, mirroring the existing `EnvironmentKeyAuthenticationExtension` / `MasterAPIKeyAuthenticationExtension`. The schema then documents these globally-accepted methods and the resolution warnings disappear.

### Describe alternatives you've considered

Leaving it as-is: the schema stays incomplete for OAuth2/cookie auth and the warnings persist. Documenting the schemes by hand risks drift from the real authenticator configuration.

### Additional context

N/A

PR #7959, opened 2026-07-07
feat(API): Bundle the Core API OpenAPI schema

### LINKED ISSUE #7669: Bundle the Core API OpenAPI schema
In order to:
- have a single source of truth for the MCP server and other API consumers
- have better visibility into contract changes we deliver

we should check the Core API OpenAPI JSON in our version control.

This should immediately allow us to utilise FastMCP's tool generation capabilities without restricting self-hosted MCPs in air-gapped environments (the current implication is that FastMCP tool generator should rely on SaaS swagger endpoint, which might not be accessible).

The existing MCP docgen introduced in #7668 can also be rewired to consume the static schema, and `MCPSchemaGenerator` can be retired completely.

khvn26:

## Changes

Closes #7669

In this PR, we bundle the full OpenAPI schema for Flagsmith's management API so:
1. It's a revieable, single source of truth, and
2. the MCP server generate its tools without reaching the SaaS swagger endpoint, so it works offline and in air-gapped deployments.

We introduce version controlled `openapi.yaml` that describes the entirety of Flagsmith's management API, including the scale-up/enterprise gated endpoints. It's currently generated in CI by `drf-spectacular`. A subset of this schema, `mcp/openapi.json`, is generated via the `generate-docs` Makefile target.

Plan-gated endpoints are now marked in the schema with the `x-flagsmith-minimum-plan` extension.

`MCPSchemaGenerator` and the `?mcp=true` swagger view are retired. Since we no longer push to Gram, the custom OAuth scheme wiring is removed. I've opened #7958 to properly address covering remaining auth schemes, which is out of scope for this PR.

Verified that the bundled schema-based MCP tool list is identical to current mcp.flagsmith.com one.


.github/workflows/api-tests-with-private-packages.yml
```diff
@@ -69,6 +69,13 @@ jobs:
             > ../docs/docs/integrating-with-flagsmith/_mcp-tool-catalogue-enterprise.md
           git diff --exit-code ../docs/docs/integrating-with-flagsmith/_mcp-tool-catalogue-enterprise.md
 
+      - name: Check full OpenAPI schema is up to date
+        env:
+          DOTENV_OVERRIDE_FILE: .env-ci
+        run: |
+          make generate-full-openapi
+          git diff --exit-code ../openapi.yaml
```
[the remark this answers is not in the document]
khvn26: Well, the other workflow checks for that, does it not?

api/Makefile
```diff
@@ -172,8 +172,23 @@ integrate-private-tests:
...
+
+.PHONY: generate-mcp-openapi
+generate-mcp-openapi:
+	npx openapi-format@${OPENAPI_FORMAT_VERSION} ../openapi.yaml \
+	  --filterFile openapi-filter-mcp.yaml \
+	  --output ../mcp/src/flagsmith_mcp/openapi.json
+
 .PHONY: generate-docs
-generate-docs: generate-flagsmith-sdk-openapi
+generate-docs: generate-flagsmith-sdk-openapi generate-mcp-openapi
```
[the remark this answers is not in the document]
khvn26: I wanted to respond that it won't work, but it actually can now since CodeArtifact auth is OIDC-based now, but that is probably a bigger side-quest to merge private and OSS API CI runs which is way out of scope of this PR.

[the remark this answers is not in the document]
khvn26: Well the full schema is generated in the other workflow that includes the private dependency.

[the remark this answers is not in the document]
khvn26: Meta-comment: I expected CodeRabbit to resolve its own discussion here. Resolving it myself.

api/app/urls.py
```diff
@@ -103,7 +103,10 @@
     from organisations.subscriptions.permissions import require_minimum_plan
 
     scale_up_permission = require_minimum_plan(SubscriptionPlanFamily.SCALE_UP)
-    SamlConfigurationViewSet.permission_classes += [scale_up_permission]
+    SamlConfigurationViewSet.permission_classes = [
+        *SamlConfigurationViewSet.permission_classes,
+        scale_up_permission,
+    ]
```
emyller, line 109: No action needed, but curious to why this change.

api/tests/unit/api/test_openapi_plan_gating.py
```diff
@@ -0,0 +1,61 @@
...
+) -> None:
+    # Given
+    mocker.patch(
+        "drf_spectacular.openapi.AutoSchema.get_operation",
+        return_value={"operationId": "op"},
+    )
+    schema = AutoSchema()
+    schema.view = SimpleNamespace(
+        permission_classes=[require_minimum_plan(SubscriptionPlanFamily.SCALE_UP)]
+    )
```
emyller, line 21: No action needed: I noticed `SimpleNamespace` use increasing, perhaps thanks to LLMs, but I wonder if there's any benefit in tests opposed to using a mock. 🤔

mcp/tests/integration/conftest.py
```diff
@@ -69,11 +70,14 @@ def openapi_spec() -> openapi.OpenAPI:
...
-    respx_mock.get(constants.OPENAPI_SPEC_URL).respond(
-        json=openapi_spec.model_dump(by_alias=True, exclude_none=True, mode="json")
+def openapi_spec_mock(
+    fs: FakeFilesystem,
+    openapi_spec: openapi.OpenAPI,
+) -> None:
+    path = resources.files("flagsmith_mcp").joinpath(constants.OPENAPI_SPEC_FILENAME)
+    fs.create_file(
+        str(path),
+        contents=openapi_spec.model_dump_json(),
```
emyller, line 80: question: Sorry, not clear... are we still mocking MCP spec contents? We could use the real one, or maybe we already are?

emyller (approved):
Looks great! Thanks for the extra mile in the refactor.

api/app/urls.py
khvn26: It was a bug — see ea27c49b5985db28bbf43ab098b7e0f95bf94765 commit message.

api/tests/unit/api/test_openapi_plan_gating.py
khvn26: No benefits or downsides, IMO both would be legit here.

mcp/tests/integration/conftest.py
khvn26: Yes, because it's easier to test against a pre-determined test schema than against the real one. If we decide to test agaisnt real endpoints, we're creating an unnecessary dependency. It's safer and cleaner to have a fixture expressing requirements in a minimal manner.

api/app/urls.py
emyller: Thanks. I read the message now, and my brain still can't parse what's the practical difference between A and B. I'll get to it if I get to it. :)

mcp/tests/integration/conftest.py
emyller: Makes sense!

api/app/urls.py
khvn26: Mutable class attribute + MRO.

Issue #7623, opened 2026-05-28
Onboarding: track first SDK evaluation received (Core)


talissoncosta:
Expose an API that can be polled to determine if an environment has completed onboarding. 

For SaaS, expose an API that allows setting onboarding state for an environment.

### Acceptance criteria
- The environment document Dynamo schema is updated with `onboarding_pending: bool | null` field.
- `GET /api/v1/environments/{environment_api_key}/onboarding-status/` returns `first_evaluated_at: datetime | null, first_evaluated_sdk_label: str | null`.
- `PUT /api/v1/environments/{environment_api_key}/onboarding-status/` persists `first_evaluated_at` and `first_evaluated_sdk_label`. It's authenticated via environment API key, or equivalent auth sufficient to be called from inside an Edge API lambda.
- No new queries fired during SDK API operation for onboarded environments (including pre-existing environments).

Issue #7767, opened 2026-06-11
Onboarding: wire the real first-evaluation signal


talissoncosta:
Part of #7737. The cross-cutting plumbing the parts depend on.

## Scope
- Wire the real first-evaluation signal (#7623) into the terminal's connected state.

## Done when
- Connected state fires from the real #7623 signal, never faked.

Depends on #7623, plus #7765 & #7766 for the components it assembles.

PR #7960, opened 2026-07-07
feat(onboarding): funnel analytics for the new flow

### LINKED ISSUE #7738: Onboarding: funnel analytics + variant tagging for the new flow
Part of #7540. Follows #7541, which mapped the current flow's analytics and agreed an activation metric set, explicitly leaving "instrument the new flow" to a separate ticket — this one.

## Goal

Make the new single-page flow (#7737) comparable to the current onboarding **in one funnel, broken down by variant** — not a parallel set of events.

## Work

- Both flows already emit the same milestone events from the shared stores (`REGISTER`, `CREATE_FIRST_ORGANISATION`, `CREATE_FIRST_PROJECT`, `CREATE_FIRST_FEATURE`), so the comparable skeleton exists.
- Tag those events with an `onboarding_variant` property (`control` | `single_page`), sourced from the FoF flag the flow is gated behind, so the funnel can be split by variant and compared directly.
- Ensure the activation milestone — first SDK evaluation received (#7623) — is captured and variant-tagged. It's the moment that matters.
- Add new-flow-only diagnostics (`snippet.copied`, AI-connect used, in-page toggle) for the single-page flow's internal drop-off — kept separate from the comparison funnel, since the control has no equivalent.
- Include org / project / environment IDs; no PII.

## Sequencing

Instrumentation over the finished UI — wire it once the flow (#7737) has landed. One of the last tasks.

talissoncosta:

## Changes

Contributes to #7738.

Funnel analytics for the new single-page onboarding flow. It now emits the same milestone events as the current flow, tags the variant it served, and adds new-flow-only diagnostics. Funnel analytics only. The multivariate flag and a true A/B split (tagging the control arm) are a follow-up.

- **`onboarding_variant` user property** (`control` | `single_page`), set at the getting-started gate when a user reaches onboarding (not at identify), from the `onboarding_quickstart_flow` flag. Tags which flow the user saw, so onboarding events can be grouped by it.
- **Milestone events from the new flow.** It auto-creates the project and flag via RTK, which emitted nothing, so the funnel had no project/feature step. It now fires `First Project created` / `First Feature created` on a real create (org already fired). Milestone events only, not the generic `Project/Feature created` actions, which stay reserved for resources a user makes by hand.
- **New-flow-only diagnostics** for internal drop-off: `Onboarding snippet copied` (install / wire), `Onboarding AI connect used`, `Onboarding flag toggled`.

Follow-ups: multivariate flag plus control-arm tagging for a true A/B split; activation (first SDK evaluation, pending #7623); shared-analytics hardening (`setOnce`, `trackEvent` try/catch) in #8061.

Manually, with `onboarding_quickstart_flow` on and analytics logging enabled (or by spying on `API.trackEvent` / `API.trackTraits` in the console):

1. Sign up and land on the flow; confirm `onboarding_variant = single_page` is set once at the gate.
2. Confirm `First Project created` / `First Feature created` fire once on the auto-bootstrap, and not again on revisit.
3. Copy a snippet, copy the AI prompt, toggle the flag; confirm the three diagnostics events (the toggle event only after it saves).

### DIFF
--- a/frontend/common/utils/getOnboardingVariant.ts
+++ b/frontend/common/utils/getOnboardingVariant.ts
@@ -2,10 +2,12 @@ import Utils from './utils'
 
 export type OnboardingVariant = 'control' | 'single_page'
 
-export const getOnboardingVariant = (): OnboardingVariant =>
-  Utils.getFlagsmithHasFeature('onboarding_quickstart_flow')
-    ? 'single_page'
-    : 'control'
-
+// The served variant name picks the arm ('control' is the reserved key the
+// API reports for the default arm). Enabled with no variant is the
+// pre-conversion boolean flag, which means the new flow.
 export const isSinglePageOnboarding = (): boolean =>
-  getOnboardingVariant() === 'single_page'
+  Utils.getFlagsmithHasFeature('onboarding_quickstart_flow') &&
+  Utils.getFlagsmithVariant('onboarding_quickstart_flow') !== 'control'
+
+export const getOnboardingVariant = (): OnboardingVariant =>
+  isSinglePageOnboarding() ? 'single_page' : 'control'
--- a/frontend/common/utils/getOnboardingVariant.ts
+++ b/frontend/common/utils/getOnboardingVariant.ts
@@ -1,0 +1,11 @@
+import Utils from './utils'
+
+export type OnboardingVariant = 'control' | 'single_page'
+
+export const getOnboardingVariant = (): OnboardingVariant =>
+  Utils.getFlagsmithHasFeature('onboarding_quickstart_flow')
+    ? 'single_page'
+    : 'control'
+
+export const isSinglePageOnboarding = (): boolean =>
+  getOnboardingVariant() === 'single_page'
--- a/frontend/web/components/pages/onboarding/GettingStartedGate.tsx
+++ b/frontend/web/components/pages/onboarding/GettingStartedGate.tsx
@@ -1,19 +1,20 @@
-import React, { FC } from 'react'
-import Utils from 'common/utils/utils'
+import React, { FC, useEffect } from 'react'
+import {
+  getOnboardingVariant,
+  isSinglePageOnboarding,
+} from 'common/utils/getOnboardingVariant'
+import API from 'project/api'
 import GettingStartedPage from 'components/pages/GettingStartedPage'
 import OnboardingFlow from './OnboardingFlow'
 
-// Gates /getting-started: renders the new onboarding flow when the
-// `onboarding_quickstart_flow` flag is on, otherwise the existing page.
-//
-// Deliberately a flat file, not a folder + barrel - a one-line route gate with
-// no styles or sub-parts, the documented exception to the component-folder
-// convention (frontend/CLAUDE.md rule 8).
-const GettingStartedGate: FC = () =>
-  Utils.getFlagsmithHasFeature('onboarding_quickstart_flow') ? (
-    <OnboardingFlow />
-  ) : (
-    <GettingStartedPage />
-  )
+const GettingStartedGate: FC = () => {
+  const variant = getOnboardingVariant()
+
+  useEffect(() => {
+    API.trackTraits({ onboarding_variant: variant })
+  }, [variant])
+
+  return isSinglePageOnboarding() ? <OnboardingFlow /> : <GettingStartedPage />
+}
 
 export default GettingStartedGate
--- a/frontend/web/components/pages/onboarding/OnboardingFlow/OnboardingFlow.tsx
+++ b/frontend/web/components/pages/onboarding/OnboardingFlow/OnboardingFlow.tsx
@@ -16,7 +16,11 @@ import { useOnboardingFlag } from 'components/pages/onboarding/hooks/useOnboardi
 import { useOnboardingConnection } from 'components/pages/onboarding/hooks/useOnboardingConnection'
 import { useUpdateOrganisationMutation } from 'common/services/useOrganisation'
 import { useUpdateProjectMutation } from 'common/services/useProject'
+import API from 'project/api'
+import Constants from 'common/constants'
 import './OnboardingFlow.scss'
+
+type OnboardingSnippet = 'install' | 'wire'
 
 // The single-page onboarding flow, rendered at /getting-started when
 // onboarding_quickstart_flow is on (see GettingStartedGate).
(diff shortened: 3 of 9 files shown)


frontend/web/components/pages/onboarding/hooks/bootstrapOnboarding.ts
```diff
@@ -93,6 +95,10 @@ async function ensureProject(
       }),
     )
     .unwrap()
+  // Auto-created setup is a milestone reached, not a user action: fire First
+  // Project created, not the generic Project created (reserved for projects a
+  // user makes by hand). First is safe here - we only get here with no project.
+  API.trackEvent(Constants.events.CREATE_FIRST_PROJECT)
```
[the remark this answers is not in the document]
kyle-ssg: Agree with this - maybe trackEvent itself should just be try catch though not the callers.

frontend/web/components/pages/onboarding/OnboardingFlow/OnboardingFlow.tsx
```diff
@@ -139,6 +141,37 @@ const OnboardingFlow: FC = () => {
...
+  const diagnosticIds = {
+    environment_id: environment?.id,
+    organisation_id: organisationId,
+    project_id: projectId,
+  }
+  const trackSnippetCopied = (snippet: 'install' | 'wire') =>
+    API.trackEvent({
+      ...Constants.events.ONBOARDING_SNIPPET_COPIED,
+      extra: { ...diagnosticIds, snippet },
+    })
```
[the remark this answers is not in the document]
kyle-ssg: Yeah we should have a type defined

frontend/common/utils/getOnboardingVariant.ts
```diff
@@ -0,0 +1,10 @@
+import Utils from './utils'
+
+export type OnboardingVariant = 'control' | 'single_page'
+
+// Which onboarding flow the user is in, from the onboarding_quickstart_flow
+// flag the new flow is gated behind (#7738).
```
Zaimwa9: We can remove this comment

frontend/web/components/pages/onboarding/hooks/bootstrapOnboarding.ts
```diff
@@ -182,6 +189,11 @@ async function ensureFlag(
       }),
     )
     .unwrap()
+  // Milestone, not action (see the project create above): First only.
```
Zaimwa9: Same. Can you go over the comments and only leave the ones clearly stating what is happening?

frontend/common/utils/getOnboardingVariant.ts
```diff
@@ -0,0 +1,10 @@
...
+export const getOnboardingVariant = (): OnboardingVariant =>
+  Utils.getFlagsmithHasFeature('onboarding_quickstart_flow')
+    ? 'single_page'
+    : 'control'
```
Zaimwa9: How do we plan to release this one ? Should we directly have 2 variants in the feature and grab the variant name from the flag ?

frontend/web/project/api.ts
```diff
@@ -247,7 +248,12 @@ const API = {
         role: selectedRole,
         tasks: user.onboarding?.tasks?.map((t) => t.name) || [],
       })
-      API.flagsmithIdentify()
+      // Pin the onboarding variant as a user property once the flag has
+      // resolved, so every event (incl. autocapture pageviews) is attributable
+      // to control vs single_page in one funnel (#7738).
+      API.flagsmithIdentify()?.then(() =>
```
Zaimwa9: we could have it in a finally maybe to also grab the default value ? 

This means we are setting it each time we identify a user ?

All users are getting an onboardingVariant trait while they haven't gone through the onboarding ?

Zaimwa9: Just checked, we have this amplitude method to `Identify.setOnce` to avoid mutating it

Zaimwa9 (changes requested):
Mmh, sorry overall I feel that the whole flag/variant/tracking management is done at the wrong place and should be using a multivariate flag (if not an experimentation but that can be done later on).

I don't think the tracking should happen in identify:
1. `API.identify` runs on every user model load and the variant could change over time
2. The real tracking should happen when we consume the variant (`GettingStartedGate.tsx` seems like the right place)
3. We would track it on 3 years old accounts for example and/or on users that never reach the onboarding, meaning we would corrupt the data

Let me know if i'm missing something

frontend/common/utils/getOnboardingVariant.ts
talissoncosta: That's a good question @Zaimwa9. I think this concern will be addressed on next [issue](https://github.com/Flagsmith/flagsmith/issues/7767). I will keep it simple now, ff will be boolean for now until we decided the path forward.

frontend/web/project/api.ts
talissoncosta: Good call @Zaimwa9. I will revert the changes on `api.ts` in order to keep this PR clean with the funnel analytics and will create follow up PR's to address your concerns.

frontend/common/utils/getOnboardingVariant.ts
talissoncosta: Done, removed in c67bc327

frontend/web/components/pages/onboarding/hooks/bootstrapOnboarding.ts
talissoncosta: Done removed in 30ffd6ca. Trimmed the narrative comments, left only the two flagging
non-obvious behaviour.

frontend/web/project/api.ts
talissoncosta: I did it on a follow-up #8061, to keep this PR onboarding-only.

frontend/web/components/pages/onboarding/GettingStartedGate.tsx
```diff
@@ -1,19 +1,20 @@
...
-  Utils.getFlagsmithHasFeature('onboarding_quickstart_flow') ? (
-    <OnboardingFlow />
-  ) : (
-    <GettingStartedPage />
-  )
+const GettingStartedGate: FC = () => {
+  const variant = getOnboardingVariant()
+
+  useEffect(() => {
+    API.trackTraits({ onboarding_variant: variant })
```
talissoncosta, line 14: Tagged at the gate (only users who reach onboarding). `set` for now; immutable `setOnce` in #8061.

frontend/common/utils/getOnboardingVariant.ts
```diff
@@ -2,10 +2,18 @@ import Utils from './utils'
...
+
+// The MV flag's served value names the variant; enabled with no or unknown
+// value falls back to the boolean behaviour (single_page).
+export const getOnboardingVariant = (): OnboardingVariant => {
+  if (!Utils.getFlagsmithHasFeature('onboarding_quickstart_flow')) {
+    return 'control'
+  }
+  const value = Utils.getFlagsmithValue('onboarding_quickstart_flow')
+  return isOnboardingVariant(value) ? value : 'single_page'
+}
```
Zaimwa9: Sorry, see here https://github.com/Flagsmith/flagsmith/pull/7979/changes#diff-da830800eab2c18085ef679517e14d18836f9d0db7940a925117f4bd8c5c565bR27-R28

We should have access to the variant **NAME** directly, so no need to use the value at all normally

Zaimwa9 (changes requested):
One change to use the right property

frontend/common/utils/getOnboardingVariant.ts
talissoncosta: Pushed df6090f1e, reading the arm from flag.variant. Is that what you meant? Left getExperimentFlag for the experimentation follow-up, glad to hear your thoughts.

Zaimwa9 (approved):
thanks for the changes!

PR #7965, opened 2026-07-08
feat: Redesign Compare Environments page

kyle-ssg:

## Changes

- Convert CompareEnvironments from JS to TypeScript
- Add expandable rows with full diff view (segments, variations)
- Add standard table filters (search, tags, owners, groups, sort)
- Split results into "Changed features" and "Unchanged features" tables
- Add column headers for source/target environments and segment changes
- Show environment names in diff view dropdown instead of "Old/New Value"
- Add environment name legend with +/- indicators in diff colours
- Style improvements: proper table header styling, badge counts
_Please describe._

Compared features in production


frontend/web/components/CompareEnvironments.tsx
```diff
@@ -0,0 +1,607 @@
...
+  }
+
+  const fetch = useCallback(async () => {
+    if (!environmentLeft || !environmentRight) {
+      return
+    }
+    setIsLoading(true)
+    try {
+      const [
+        environmentLeftProjectFlags,
```
talissoncosta: Since this is a rewrite, could the four fetches go through the RTK services instead (`useGetProjectFlagsQuery` / `useGetFeatureStatesQuery`)? That would also remove the `ProjectStore` import from a brand-new file.

```tsx
const { data: leftFlags } = useGetProjectFlagsQuery({ project: projectId, environment: envLeftId, page_size: 999 })
```

frontend/web/components/CompareEnvironments.tsx
```diff
@@ -0,0 +1,532 @@
+import React, { FC, useCallback, useMemo, useState } from 'react'
```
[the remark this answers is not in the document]
talissoncosta: I agree with this comment. We already have a few components (`ExpandedRow` is defined in-file, `renderRow` is a component in disguise). Could it move to its own folder with a barrel and split by concern in order to improve maintenance ?

```
web/components/compare-environments/
  index.ts                     // re-exports default
  CompareEnvironments.tsx      // state + layout
  CompareFeatureRow.tsx        // the renderRow content
  ExpandedRow.tsx
  useEnvironmentComparison.ts  // fetch + diff derivation
  types.ts                     // FeatureChange
```

Pulling the fetch/diff logic into a `useEnvironmentComparison` hook would also make it unit-testable.

talissoncosta (dismissed):
Looks good @kyle-ssg ! Thanks for that! 

I just left some suggestions.

If we can break it into small components and get rid of the `_data` import would be amazing.

talissoncosta (dismissed):
Thanks for addressing the changes.

It looks good!

frontend/common/utils/recursivePageGet.ts
```diff
@@ -6,8 +6,10 @@ import { PagedResponse } from 'common/types/responses'
 export function recursivePageGet<T>(
   url: string,
   parentRes: null | PagedResponse<T>,
-  baseQuery: (arg: unknown) => any, // matches rtk types,
-): Promise<{ data: PagedResponse<T> } | { error: any }> {
+  baseQuery: (
+    arg: unknown,
+  ) => MaybePromise<QueryReturnValue<PagedResponse<T>, unknown>>,
```
talissoncosta, line 17: It seems that `MaybePromise` and `QueryReturnValue` are used but never imported.

frontend/web/styles/3rdParty/_react-diff.scss
```diff
@@ -95,3 +95,27 @@
...
+
+.diff-removed-header {
+  background-color: $danger-alfa-16;
+  color: $danger;
+}
+
+.diff-added-header {
+  background-color: $success-alfa-16;
+  color: $success;
+}
```
talissoncosta, line 116: Can you check if these colours can use our semantic tokens? We have dark-mode-aware equivalents:

```css
.diff-removed-header {
  background-color: var(--color-surface-danger);  // was $danger-alfa-16
  color: var(--color-text-danger);                // was $danger
}
.diff-added-header {
  background-color: var(--color-surface-success); // was $success-alfa-16
  color: var(--color-text-success);               // was $success
}
```
The $…-alfa-16 values are static light-mode rgba, so they won't adapt in dark mode, whereas these tokens have dedicated dark values.

frontend/web/components/CompareEnvironments/useEnvironmentComparison.ts
```diff
@@ -0,0 +1,167 @@
...
+    refetchLeftFlags,
+    refetchRightFlags,
+  ])
+
+  const leftEnvironment = ProjectStore.getEnvironment(
+    leftEnvironmentKey,
+  ) as StoredEnvironment
+  const rightEnvironment = ProjectStore.getEnvironment(
+    rightEnvironmentKey,
+  ) as StoredEnvironment
```
talissoncosta: Can we use `useGetEnvironmentsQuery({ projectId })` and match on `api_key` here instead of `ProjectStore`? That'd give both the env object and its numeric id (replacing the getEnvironmentIdFromKey calls too) and drop the Flux dependency.

frontend/web/styles/3rdParty/_react-diff.scss
kyle-ssg: Done 👍

talissoncosta (approved):

PR #7966, opened 2026-07-08
fix(Segment Membership): Enabled for different organisations in backend and frontend

khvn26:

## Changes

The dashboard decided whether to show the segment membership UI by evaluating its own Flagsmith flag in the Flagsmith Website project, separately from the backend's `is_membership_enabled` check. The two evaluations target different projects, and their `PERCENTAGE_SPLIT` rules are salted by segment id, so an org falls into one segment's 10% bucket but not the other's. The result is that the frontend and backend disagree about whether the feature is on for a given org.

In this PR, we drop the separate frontend flag and derive availability strictly from backend output.

- Manual: with an org membership-enabled and counts present, the Identities tab uses the members endpoint; disabling the org (removing it from `membership-enabled-orgs`) clears the counts on the next refresh, and the tab falls back to the legacy identities list.

### DIFF
--- a/api/tests/unit/segment_membership/test_unit_segment_membership_tasks.py
+++ b/api/tests/unit/segment_membership/test_unit_segment_membership_tasks.py
@@ -449,23 +449,37 @@ def test_refresh_project_segment_counts__no_clickhouse_creds__skips(
     )
 
 
-def test_refresh_project_segment_counts__ff_disabled__skips(
+def test_refresh_project_segment_counts__ff_disabled__skips_and_purges_stale_counts(
     mocker: MockerFixture,
     settings: SettingsWrapper,
     project: Project,
+    environment: Environment,
+    segment: Segment,
     log: StructuredLogCapture,
 ) -> None:
-    # Given
+    # Given a count left over from when the org was still enabled
     settings.CLICKHOUSE_ENABLED = True
     spy = mocker.patch.object(tasks, "open_clickhouse_cursor")
+    SegmentMembershipCount.objects.create(
+        segment=segment,
+        environment=environment,
+        count=15,
+        last_synced_at=timezone.now(),
+    )
 
     # When
     refresh_project_segment_counts(project.id)
 
-    # Then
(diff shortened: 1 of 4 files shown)


api/tests/unit/segment_membership/test_unit_segment_membership_tasks.py
```diff
@@ -449,23 +449,37 @@ def test_refresh_project_segment_counts__no_clickhouse_creds__skips(
...
+        count=15,
+        last_synced_at=timezone.now(),
+    )
 
     # When
     refresh_project_segment_counts(project.id)
 
-    # Then
+    # Then the compute path is skipped and the stale row is purged, so the
+    # dashboard stops showing the feature for the now-disabled org
```
emyller:
```suggestion
    # Then
```

emyller (dismissed):
One nit.

api/tests/unit/segment_membership/test_unit_segment_membership_tasks.py
khvn26: 03d1173

api/tests/unit/segment_membership/test_unit_segment_membership_tasks.py
```diff
@@ -449,23 +449,37 @@ def test_refresh_project_segment_counts__no_clickhouse_creds__skips(
...
+def test_refresh_project_segment_counts__ff_disabled__skips_and_purges_stale_counts(
     mocker: MockerFixture,
     settings: SettingsWrapper,
     project: Project,
+    environment: Environment,
+    segment: Segment,
     log: StructuredLogCapture,
 ) -> None:
-    # Given
+    # Given a count left over from when the org was still enabled
```
emyller:
```suggestion
    # Given
```

khvn26: 1c3f6a1

emyller (approved):

PR #7979, opened 2026-07-10
feat: signup corporate-only experiment with exposure and new_signup events

Zaimwa9:

## Changes

Sets up our first dogfood experiment, **Signup corporate domains only**.

The experiment tests whether requiring a work email address at signup affects conversion. Control keeps today's behaviour (an informational message for free email domains); the `signup_corporate_only` variant disables the Create Account button for free email domains and shows "Please use your work email address to create your account."

- Upgrade `@flagsmith/flagsmith` to 12.1.0 for `trackEvent`, `trackExposureEvent`, `getExperimentFlag` and the `variant` key on flags.
- New `useSignupExperiment` hook: identifies anonymous signup visitors with a transient identity (persistent `signup_anonymous_id` so returning visitors keep their variant), resolves the `signup_corporate_only` flag's variant, and records a `$flag_exposure`. Takes a boolean to fall back to environment flags — no identify, no exposure, control experience — for non-participants (invitees, login views, logged-in users, `preventEmailPassword`).
- `new_signup` conversion event on successful registration, attributed to the same anonymous identity, with `free_email_domain`, `invite`, `signup_method` and `utm_source` metadata.
- Signup button now also requires non-blank first and last names.

Manually, against the staging configuration with the flag and experiment set up:

1. Open `/signup` logged out — the network tab shows `identities/?identifier=<uuid>&transient=true`, followed (≤10s) by a `POST /v1/events` containing a `$flag_exposure` with the variant as `value` (verified 202).
2. In the treatment variant, entering a gmail.com address disables Create Account and shows the work-email message; a work domain enables it. Control behaviour is unchanged.
3. Completing a signup sends `new_signup` with the same identifier and metadata.
4. Invite links and the login view trigger no identify/exposure and always get control.
5. Variant bucketing verified ~50/50 across identifiers against the identities API; unit tests (236) and typecheck pass.

### DIFF
--- a/frontend/common/stores/account-store.js
+++ b/frontend/common/stores/account-store.js
@@ -246,6 +246,18 @@ const controller = {
         }
 
         data.setToken(Project.cookieAuthEnabled ? 'true' : res.key)
+        if (res.is_new_user) {
+          try {
+            flagsmith.trackEvent('new_signup', {
+              metadata: {
+                invite: !!API.getInvite(),
+                signup_method: type,
+              },
+            })
+          } catch (e) {
+            // never let analytics break the login flow
+          }
+        }
         return controller.onLogin()
       })
       .catch((e) => API.ajaxHandler(store, e))
@@ -268,15 +280,19 @@ const controller = {
         data.setToken(Project.cookieAuthEnabled ? 'true' : res.key)
         API.trackEvent(Constants.events.REGISTER)
         const freeEmailDomain = isFreeEmailDomain(user.email)
-        flagsmith.trackEvent('new_signup', {
-          metadata: {
-            ...(freeEmailDomain && { domain: user.email.split('@')[1] }),
-            free_email_domain: freeEmailDomain,
-            invite: !!API.getInvite(),
-            signup_method: 'email',
-            utm_source: user.utm_data?.utm_source,
-          },
-        })
+        try {
+          flagsmith.trackEvent('new_signup', {
+            metadata: {
+              ...(freeEmailDomain && { domain: user.email.split('@')[1] }),
+              free_email_domain: freeEmailDomain,
+              invite: !!API.getInvite(),
+              signup_method: 'email',
+              utm_source: user.utm_data?.utm_source,
+            },
+          })
+        } catch (e) {
+          // never let analytics break the signup flow
+        }
         if (API.getReferrer()) {
           API.trackEvent(
             Constants.events.REFERRER_REGISTERED(API.getReferrer().utm_source),
--- a/frontend/common/stores/account-store.js
+++ b/frontend/common/stores/account-store.js
@@ -267,9 +267,11 @@ const controller = {
       .then(async (res) => {
         data.setToken(Project.cookieAuthEnabled ? 'true' : res.key)
         API.trackEvent(Constants.events.REGISTER)
+        const freeEmailDomain = isFreeEmailDomain(user.email)
         flagsmith.trackEvent('new_signup', {
           metadata: {
-            free_email_domain: isFreeEmailDomain(user.email),
+            ...(freeEmailDomain && { domain: user.email.split('@')[1] }),
+            free_email_domain: freeEmailDomain,
             invite: !!API.getInvite(),
             signup_method: 'email',
             utm_source: user.utm_data?.utm_source,
--- a/frontend/common/stores/account-store.js
+++ b/frontend/common/stores/account-store.js
@@ -17,6 +17,8 @@ import { hidePylon, identifyChatUser } from 'common/loadChat'
 import { service } from 'common/service'
 import { getBuildVersion } from 'common/services/useBuildVersion'
 import { createOnboardingSupportOptIn } from 'common/services/useOnboardingSupportOptIn'
+import flagsmith from '@flagsmith/flagsmith'
+import isFreeEmailDomain from 'common/utils/isFreeEmailDomain'
 
 const controller = {
   acceptInvite: (id) => {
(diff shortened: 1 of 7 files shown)


frontend/web/components/pages/HomePage.tsx
```diff
@@ -663,7 +674,10 @@ const HomePage: React.FC = () => {
                                     disabled={
                                       isLoading ||
                                       isSaving ||
-                                      !allRequirementsMet
+                                      !allRequirementsMet ||
+                                      !firstName.trim() ||
+                                      !lastName.trim() ||
+                                      blockGenericEmailDomain
```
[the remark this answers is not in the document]
Zaimwa9: Per spec. Safe to ignore

frontend/web/components/pages/HomePage.tsx
```diff
@@ -180,6 +181,12 @@ const HomePage: React.FC = () => {
       currentLocation.indexOf('signup') !== -1)
   const disableSignup = preventSignup && isSignup
   const preventEmailPassword = Project.preventEmailPassword
+
+  const signupVariant = useSignupExperiment(
+    !isSignup || isInvite || !!preventEmailPassword || !!AccountStore.getUser(),
+  )
+  const blockGenericEmailDomain =
+    signupVariant === 'signup_corporate_only' && isFreeEmailDomain(email)
```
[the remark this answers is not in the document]
Zaimwa9: Ignoring

frontend/common/useSignupExperiment.ts
```diff
@@ -0,0 +1,24 @@
...
+    const identifyAndExpose = async () => {
+      const id = storageGet('signup_anonymous_id') || crypto.randomUUID()
+      storageSet('signup_anonymous_id', id)
+      // @ts-expect-error transient is missing from the SDK's identify type
+      await flagsmith.identify(id, {}, true)
+      const flag = flagsmith.getExperimentFlag('signup_corporate_only')
+      setVariant(flag?.enabled ? flag.variant : undefined)
+    }
+    identifyAndExpose()
+  }, [useEnvironmentFlag])
```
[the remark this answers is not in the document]
Zaimwa9: ignoring

frontend/common/useSignupExperiment.ts
```diff
@@ -0,0 +1,24 @@
+import { useEffect, useState } from 'react'
+import flagsmith from '@flagsmith/flagsmith'
+import { storageGet, storageSet } from './safeLocalStorage'
```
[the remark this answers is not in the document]
Zaimwa9: This is enforced by the repo's linter: `@dword-design/import-alias/prefer-alias` rewrites imports to the `common/` alias except for same-directory imports, which it keeps relative — `eslint --fix` converts `common/safeLocalStorage` back to `./safeLocalStorage` here. Keeping the linter's form.

frontend/common/useSignupExperiment.ts
```diff
@@ -0,0 +1,24 @@
...
+      // @ts-expect-error transient is missing from the SDK's identify type
+      await flagsmith.identify(id, {}, true)
+      const flag = flagsmith.getExperimentFlag('signup_corporate_only')
+      setVariant(flag?.enabled ? flag.variant : undefined)
+    }
+    identifyAndExpose()
+  }, [useEnvironmentFlag])
+
+  return variant
+}
```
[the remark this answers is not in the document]
Zaimwa9: overkill for a single consumer

frontend/common/stores/account-store.js
```diff
@@ -265,6 +267,16 @@ const controller = {
       .then(async (res) => {
         data.setToken(Project.cookieAuthEnabled ? 'true' : res.key)
         API.trackEvent(Constants.events.REGISTER)
+        const freeEmailDomain = isFreeEmailDomain(user.email)
+        flagsmith.trackEvent('new_signup', {
```
talissoncosta: OAuth signups skip `register`, so they get an exposure but never a new_signup. In the variant, blocked free-email users are the ones most likely to switch to Google/GitHub, so this undercounts variant conversions. Intended, or should the OAuth path also fire this event?

frontend/common/useSignupExperiment.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export default function useSignupExperiment(useEnvironmentFlag: boolean) {
+  const [variant, setVariant] = useState<string>()
+
+  useEffect(() => {
+    if (useEnvironmentFlag) {
+      return
+    }
+    const identifyAndExpose = async () => {
+      const id = storageGet('signup_anonymous_id') || crypto.randomUUID()
```
talissoncosta: This runs on self-hosted too, and on plain-http installs crypto.randomUUID is undefined( according to claude) so it throws on every signup view. Can you please double check that? Worth gating to SaaS?

talissoncosta (commented):
Looks good, just two small comments. Do you mind checking them?

frontend/common/stores/account-store.js
Zaimwa9: There's no OAuth on the signup formn it only appears on login. The exposure is triggered based on the URL so the user already landed on the signup form without oauth 
image

frontend/common/useSignupExperiment.ts
Zaimwa9: Addressed in [5fc7532](https://github.com/Flagsmith/flagsmith/pull/7979/commits/5fc7532516c07f7a84b68dba12ca82453a6a76d8)

frontend/common/stores/account-store.js
Zaimwa9: I agree that he could navigate back but I think that's a limitation we have right now. Or we could trigger the exposure when a user starts typing in the form.

Although, the advantage is that the flaw in the flow affects both variants because the real A/B test appears when he typed his email

talissoncosta: Thanks for clarifying @Zaimwa9

talissoncosta (dismissed):
Thanks for checking the comments!

Zaimwa9:
@coderabbitai review

talissoncosta (approved):

PR #7980, opened 2026-07-10
fix(Experiments): Dark mode visibility in results page

Zaimwa9:
## Changes

Fix dark mode visibility issues on the experiment results page:
- Variant names invisible in recommendation card
- Hypothesis text too faint
- Stat card titles barely visible
- Chip/tag backgrounds blending into background
- "Baseline" text invisible in scorecard table

Toggled dark mode and verified all affected elements are visible.

### DIFF
--- a/frontend/web/components/experiments/results/ExperimentDetailHeader.tsx
+++ b/frontend/web/components/experiments/results/ExperimentDetailHeader.tsx
@@ -200,7 +200,7 @@ const ExperimentDetailHeader: FC<ExperimentDetailHeaderProps> = ({
     if (isEditingHypothesis) {
       return (
         <div className='mt-3' style={{ maxWidth: 640 }}>
-          <span className='fs-caption text-muted fw-bold'>Hypothesis</span>
+          <span className='fs-caption text-secondary fw-bold'>Hypothesis</span>
           <div className='d-flex align-items-start gap-2 mt-1'>
             <textarea
               autoFocus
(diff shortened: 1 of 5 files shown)


frontend/web/components/experiments/results/ExperimentDetailHeader.tsx
```diff
@@ -252,9 +252,9 @@ const ExperimentDetailHeader: FC<ExperimentDetailHeaderProps> = ({
 
     return (
       <div className='mt-3' style={{ maxWidth: 640 }}>
-        <span className='fs-caption text-muted fw-bold'>Hypothesis</span>
+        <span className='fs-caption text-secondary fw-bold'>Hypothesis</span>
         <div className='d-flex align-items-start gap-1'>
-          <p className='text-muted mb-0'>
+          <p className='text-secondary mb-0'>
```
talissoncosta, line 257: Hey @Zaimwa9 

I just wanna flag that `text-secondary` actually renders as: it's not our token,
it's Bootstrap's `.text-secondary` utility, which maps to `$secondary` (the
brand yellow #fae392) and wins the cascade with `!important`. Readable on dark,
but ~1.3:1 on white in light mode (light vs dark screenshots below).

For muted text like this the semantic colour we want is `--color-text-secondary`
(5.22:1 light / 7.16:1 dark). The catch: the `.text-secondary` utility class is
currently shadowed by Bootstrap, so the token version never applies. Until we fix
that at the DS level, the clean way is to use the token via a component class,
same as the rest of `results.scss` already does:

```scss
.experiment-results__muted { color: var(--color-text-secondary); }
```
I'm picking up the Bootstrap-colour-vs-token collision as separate DS work, so this PR is just about getting these labels onto the token. Wdyt?

Image

Image

talissoncosta: I wrote this small [PR](https://github.com/Flagsmith/flagsmith/pull/7981) that will allow us to take ownership of text-secondary. My idea is to drop the bootstrap ownership on colors classes utilities.

talissoncosta (approved):
Good job @Zaimwa9 !

PR #7981, opened 2026-07-10
fix(dark mode): resolve .text-secondary to the token, not Bootstrap

talissoncosta:
## Changes

Contributes to #6606

`.text-secondary` was generated twice: by our token utilities
(`--color-text-secondary`) and by Bootstrap's colour utility (`$secondary`, the
brand yellow, with `!important`). Bootstrap won the cascade, so muted text rendered
yellow, fine on dark but ~1.3:1 on white (fails WCAG AA).

This drops the `secondary` key from Bootstrap's text-colour utility map so
`.text-secondary` resolves to the design-system token (`#656d7b` light / `#9da4ae`
dark, AA in both). The one intentional brand-yellow usage (`PricingPanel`) moves to
a local `.pricing-accent` class so it keeps `$secondary`.

Compiled the full stylesheet and confirmed the only `.text-secondary` rule in the
output now resolves to `var(--color-text-secondary)` (Bootstrap's yellow copy is
gone). `PricingPanel` is the sole component that used `.text-secondary`, migrated to
`.pricing-accent`; its enterprise panel (subtitle, "All from" text, feature icons)
should stay yellow, worth a quick visual confirm on review.

### DIFF
--- a/frontend/web/components/modals/payment/PricingPanel.tsx
+++ b/frontend/web/components/modals/payment/PricingPanel.tsx
@@ -85,7 +85,7 @@ export const PricingPanel = ({
 
               {isEnterprise && (
                 <Row className='pt-3 justify-content-center'>
-                  <div className='pricing-type text-secondary'>
+                  <div className='pricing-type pricing-accent'>
                     Maximum security and control
                   </div>
                 </Row>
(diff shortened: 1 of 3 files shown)


Zaimwa9 (approved):
Thanks!!

PR #7984, opened 2026-07-10
fix(dark mode): migrate text-muted usages to the text-secondary token

talissoncosta:
## Changes

Contributes to #6606.

Migrates all 173 `text-muted` classNames to `text-secondary`. They are all text
(no icon usages), so it is a straight swap onto the design-system token.

`text-muted` was a static `#656d7b` with no dark variant (3.45:1 on the dark
surface, fails WCAG AA). `--color-text-secondary` is `#656d7b` in light (unchanged)
and `#9da4ae` in dark (7.16:1, passes AA/AAA). So: **light is a visual no-op, dark
mode gets the contrast fix.**

Also updates the one scoped override (`.table-header .text-muted` in `_lists.scss`)
to key on `.text-secondary`, preserving the table-header muted colour exactly.

Compiled the full stylesheet (no errors). Verified no `text-muted` classNames
remain and all `text-muted` usages were text (no icons needing `icon-secondary`).
CI lints only changed lines.

Recommend a **dark-mode QA sweep**: muted/secondary text is used widely (tables,
forms, labels, empty states), and this is the change that lightens it from
`#656d7b` to `#9da4ae` in dark. Worth eyeballing a few dense screens.

### DIFF
--- a/frontend/web/components/Breadcrumb.tsx
+++ b/frontend/web/components/Breadcrumb.tsx
@@ -19,12 +19,12 @@ const Breadcrumb: FC<BreadcrumbType> = ({
           <Link className='text-primary h6 mb-0' to={item.url}>
             {item.title}
           </Link>
-          <div className='text-muted mx-2 h6 mb-0'>/</div>
+          <div className='text-secondary mx-2 h6 mb-0'>/</div>
         </>
       ))}
       {isCurrentPageMuted ? (
         <div
-          className='active h6 text-muted mb-0'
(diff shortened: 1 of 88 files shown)


talissoncosta:
Superseded: `text-muted` doesn't collide with the token utilities, so migrating all 173 usages is disproportionate. Fixing the dark-mode contrast at the value level instead (pointing $text-muted at --color-text-secondary), in a one-line PR. Full removal of Bootstrap's .text-muted can wait for the eventual cleanup.

PR #7982, opened 2026-07-10
fix(dark mode): move colour and shadow utilities from Bootstrap to tokens

talissoncosta:
## Changes

Contributes to #6606. Consolidates the dark-mode colour work into one PR (supersedes #7985, #7986, #7987) so it can be reviewed and QA'd in a single pass.

Moves ownership of colour and shadow **utilities** from Bootstrap to the design-system token layer. Bootstrap's utilities load later and carry `!important`, so wherever a token utility shared a name it was silently dead and Bootstrap's value won. Commit-by-commit:

- **Reclaim semantic text colours** — drop `success` / `danger` / `warning` / `info` from Bootstrap's text-colour utility (values identical, no-op) + the unused `light` / `black` / `white-50` / `black-50`. (`secondary` already landed in #7981.)
- **Migrate `text-primary` / `text-dark` / `text-body`** to token classes: `text-action` / `icon-action` for the purple, `text-default` / `icon-default` for body text (~43 files).
- **`text-muted` theme-aware** — point `$text-muted` at `--color-text-secondary` (one line): `#656d7b` light unchanged, `#9da4ae` dark (3.45:1 → 7.16:1, fixes AA). No usage churn.
- **Reclaim border-colour + shadow** — `border-*` danger/info/success/warning (identical, no-op); `shadow-sm`/`shadow-lg` now use the DS `--shadow-*` tokens (were silently using Bootstrap's heavier shadows).
- **Remove the global `.text-info` purple override** — an unscoped `!important` rule forced all `.text-info` to purple; moved the two webhook doc links to `.text-action` and restored `.text-info` to info-blue.

## How did you test this code?

Compiled the full stylesheet and confirmed every previously-colliding utility now resolves to a single token rule: `text-secondary/success/danger/warning/info`, `border-success/warning`, `shadow-sm/lg` → `var(--...)`; `text-muted` → `var(--color-text-secondary)`; `text-info` → info-blue. No stray `text-primary/dark/body/muted/info` classNames remain.

**Single dark-mode QA pass** covers everything that actually changes appearance:
- purple `text-action` elements (tabs, breadcrumbs, links, segment names) — near-black → readable purple in dark
- muted text (tables, forms, labels) — lighter grey in dark
- the one `shadow-sm` element — subtler DS shadow
- the two webhook "Secret" doc links — still purple

PR #7985, opened 2026-07-10
fix(dark mode): make text-muted theme-aware via the secondary token

talissoncosta:
## Changes

Contributes to #6606.

`.text-muted` is a static `#656d7b` with no dark variant, so muted text is 3.45:1 on the dark surface (fails WCAG AA). Point `$text-muted` at the `--color-text-secondary` token so the utility becomes theme-aware: `#656d7b` in light (unchanged, 5.22:1) and `#9da4ae` in dark (7.16:1).

One line, no usage churn. Unlike `text-secondary`, `.text-muted` doesn't collide with the token utilities, so its ~170 usages don't need migrating; fixing the value is enough. Full removal of Bootstrap's `.text-muted` can come with the later cleanup.

Compiled the stylesheet: `.text-muted` now emits `color: var(--color-text-secondary) !important`. Light mode unchanged; recommend a dark-mode eyeball on muted text (tables, forms, labels).

### DIFF
--- a/frontend/web/styles/_variables.scss
+++ b/frontend/web/styles/_variables.scss
@@ -31,7 +31,7 @@ $body-color: #1a2634;
 $text-icon-light: #ffffff;
 $text-icon-grey: #656d7b;
 $text-icon-light-grey: rgba(157, 164, 174, 1);
-$text-muted: $text-icon-grey;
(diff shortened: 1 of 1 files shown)


talissoncosta:

PR #7986, opened 2026-07-10
fix(dark mode): reclaim border-colour and shadow utilities for tokens

talissoncosta:
## Changes

Contributes to #6606. Same `map-remove` pattern as the text-colour reclaim, extended to two more utility families found by auditing token-vs-Bootstrap collisions.

- **border-colour** (`danger`/`info`/`success`/`warning`): token `.border-*` was shadowed by Bootstrap. Values are **identical**, so this is a no-op; it just moves ownership to the token layer.
- **shadow** (`sm`/`lg`/`none`): `.shadow-sm` and `.shadow-lg` were silently using **Bootstrap's** shadows, while `.shadow-md`/`.shadow-xl` (token-only) used the DS tokens, so the scale was half-and-half. Now all of sm/md/lg/xl use `--shadow-*`. Suffixless `.shadow` stays Bootstrap.

Compiled the stylesheet: `.border-*` now resolve to `var(--color-border-*)`, and `.shadow-sm`/`.shadow-lg` to `var(--shadow-*)`. Border is a no-op (identical values). Shadow is a real change but tiny footprint (`shadow-sm`: 1 usage, `shadow-lg`: 0) — the DS shadows are subtler; worth a quick eyeball on that one `shadow-sm` element.

### DIFF
--- a/frontend/web/styles/3rdParty/_bootstrap.scss
+++ b/frontend/web/styles/3rdParty/_bootstrap.scss
@@ -29,6 +29,26 @@ $_color-util: map-merge(
 );
 $utilities: map-merge($utilities, ("color": $_color-util));
 
+// Border colour: token border utilities take over. Values are identical to
+// Bootstrap's, so a no-op; it just moves ownership to the token layer.
+$_border-util: map-get($utilities, "border-color");
+$_border-util: map-merge(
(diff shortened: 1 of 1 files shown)


talissoncosta:

PR #7987, opened 2026-07-10
fix(dark mode): stop .text-info being hijacked to purple

talissoncosta:
## Changes

Contributes to #6606.

`_alert.scss` had an unscoped `.text-info { color: $primary !important }` (added in 2022 for two webhook secret-field doc links). It forced **every** `.text-info` to purple globally, overriding the info-blue token, and `!important` meant it survived the text-colour reclaim too.

- Move the two webhook doc links (`CreateWebhook`, `CreateAuditLogWebhook`) to `.text-action` — they're links, so the interactive/purple colour is the correct semantic, and appearance is preserved.
- Remove the global override, so `.text-info` resolves to `--color-text-info` (info-blue) again.

Compiled the stylesheet: the purple `.text-info` rule is gone; `.text-info` now resolves to the info-blue token. No `text-info` classNames remain (both moved to `text-action`). The two links keep their purple.

### DIFF
--- a/frontend/web/components/modals/CreateAuditLogWebhook.tsx
+++ b/frontend/web/components/modals/CreateAuditLogWebhook.tsx
@@ -104,7 +104,7 @@ const CreateAuditLogWebhook: React.FC<Props> = ({
             <label>
               Secret (Optional) -{' '}
               <a
-                className='text-info'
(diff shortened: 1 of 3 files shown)


talissoncosta:

PR #7991, opened 2026-07-11
fix(sdk): return 400 when identifier is missing from SDK identity endpoint

Marnie0415:
## Changes

The SDK identities endpoint (`GET /api/v1/environments/{api_key}/identities/`) currently returns HTTP 200 when the required `identifier` query parameter is missing. This is incorrect — a missing required parameter should return HTTP 400 Bad Request.

The existing code even has a TODO comment acknowledging this: `# TODO: add 400 status - will this break the clients?`

**What this PR does:**
- Adds `status=status.HTTP_400_BAD_REQUEST` to the error response
- Removes the TODO comment

**Will this break SDK clients?**

No. SDK clients check `response.ok` (which maps to `200 <= status < 300`). Currently, when the identifier is missing, the SDK receives a 200 status with an error body — clients silently treat this as success and display empty/corrupted data. After this change, the SDK will correctly detect the error via the 400 status code.

Verified that:
1. The `status` import already exists at line 17 of the file (`from rest_framework import status, viewsets`)
2. No other callers depend on the 200 status for this error case
3. All SDK clients (Python, JS, Go, Ruby) check `response.ok` which correctly handles 400 as an error

Manual test: `GET /api/v1/environments/{key}/identities/` without `?identifier=` should return 400 with `{"detail": "Missing identifier"}`.

### DIFF
--- a/api/environments/identities/views.py
+++ b/api/environments/identities/views.py
@@ -169,8 +169,9 @@ class SDKIdentities(SDKAPIView):
         identifier = request.query_params.get("identifier")
         if not identifier:
             return Response(
-                {"detail": "Missing identifier"}
-            )  # TODO: add 400 status - will this break the clients?
+                {"detail": "Missing identifier"},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
 
         if request.query_params.get("transient"):
             identity = Identity(


Marnie0415:
Re: SDK compatibility — this change is safe because all official Flagsmith SDKs (Python, JavaScript, Go, Ruby, .NET) use 
esponse.ok or equivalent to check for success, which only returns true for 200-299 status codes. The current 200 response on error causes SDKs to silently treat the missing-identifier case as success. With this fix, SDKs will correctly detect the error via the 400 status and surface it to the caller.

PR #7955, opened 2026-07-06
feat(__future__): Add MV capabilities to the experimental update-flag endpoints

emyller:
Thanks for submitting a PR! Please check the boxes below:

- [x] I have read the [Contributing Guide](/Flagsmith/flagsmith/blob/main/CONTRIBUTING.md).
- [x] I have added information to `docs/` if required so people know about the feature.
- [x] I have filled in the "Changes" section below.
- [x] I have filled in the "How did you test this code" section below.

## Timeline for context

1. New — and experimental — [update-flag endpoints](https://docs.flagsmith.com/integrating-with-flagsmith/flagsmith-api-overview/admin-api/updating-flags#roll-out-a-feature-to-a-segment) were introduced as Options A and B (https://github.com/Flagsmith/flagsmith/pull/6305).
2. The _experimentation_ team has eventually made use of the _experimental_ endpoints, and implemented support to, specifically, _**re-weighting segment override multivariate flags in the Option B endpoint**_ (https://github.com/Flagsmith/flagsmith/pull/7851).
3. This PR adds full MV support to both endpoints.

## Changes

Contributes to https://github.com/Flagsmith/flagsmith/issues/7642

- Introduces full MV (multi variate) support for both experimental update-flags endpoints.
- Behaviour is consistent to UX in the dashboard.
- Makes all keys except `feature` in the update-flag schema optional, to allow for isolated changes to MVs — and other properties too, why not.
- Sorry it adds review effort, but helps sanity: disambiguation of v1/v2 used in both experimental update-flags and feature versioning; standardised to _Option A/B_ taxonomy already adopted by docs. Plus strict-typing of all interfaces touched in the diff.

## How did you test this code?

Unit tests.

## How to review

Reading new docs should help: https://docs-lpoi8k37l-flagsmith.vercel.app/integrating-with-flagsmith/flagsmith-api-overview/admin-api/updating-flags

Reading commit-by-commit helps a ton.

3/5 effort.

Issue #7642, opened 2026-05-29
Update experimental endpoints to support MV features and null value features


matthewelwell:
Currently, the 'experimental' endpoints as defined [here](https://docs.flagsmith.com/integrating-with-flagsmith/flagsmith-api-overview/admin-api/updating-flags) do not support multivariate flags. 

In order to have the FE dashboard migrate to using one or both of these endpoints, we'll need to add support for multivariate flags. 

Blocks #7641

kyle-ssg:
Mentioned [here](https://github.com/Flagsmith/flagsmith/issues/7641#issuecomment-4670427114)

gagantrivedi:
Related: https://github.com/Flagsmith/flagsmith/pull/7851

emyller:
Superseded by https://github.com/Flagsmith/flagsmith/issues/8088

PR #8000, opened 2026-07-13
feat(__future__): Identify MV options by key in experimental update-flags endpoint

emyller:

## Timeline for context

1. New — and experimental — [update-flag endpoints](https://docs.flagsmith.com/integrating-with-flagsmith/flagsmith-api-overview/admin-api/updating-flags#roll-out-a-feature-to-a-segment) were introduced as Options A and B (https://github.com/Flagsmith/flagsmith/pull/6305).
2. The _experimentation_ team added a new `"key"` attribute to identify multivariate options (https://github.com/Flagsmith/flagsmith/pull/7698).
3. The update-flag endpoints got full MV support (https://github.com/Flagsmith/flagsmith/pull/7955).
4. This PR adds the ability to use MV keys in the update-flag endpoints.

## Changes

Contributes to https://github.com/Flagsmith/flagsmith/issues/7642

Similarly to how a feature can be identified in the experimental update-flag endpoints, i.e. by either its ID or name, now we allow multivariate options to be identified by either their ID or [the new] _key_.

This means to make it easier to use update-flag, eliminating retrieving an MV ID as an extra step to achieving the use case of updating or re-weighting multivariates.

New tests.

### DIFF
--- a/api/experimentation/serializers.py
+++ b/api/experimentation/serializers.py
@@ -26,12 +26,10 @@ from experimentation.services import (
 from experimentation.types import (
     SNOWFLAKE_DEFAULTS,
     MetricExperimentResult,
+    RolloutMultivariateValuePayload,
     SnowflakeConfig,
 )
-from features.feature_states.serializers import (
-    FeatureValueSerializer,
-    SegmentOverrideMultivariateValueSerializer,
-)
+from features.feature_states.serializers import FeatureValueSerializer
 from features.feature_types import MULTIVARIATE
 from features.models import Feature
 from features.multivariate.serializers import NestedMultivariateFeatureOptionSerializer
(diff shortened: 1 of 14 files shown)


api/features/feature_states/serializers.py
```diff
@@ -177,6 +219,91 @@ def validate_multivariate_state_values(
...
+                    or option_id_by_key.get(mv["key"]) in allowed_option_ids
+                )
+            if not allowed:
+                raise serializers.ValidationError(
+                    {
+                        "multivariate_feature_state_values": [
+                            "Segment overrides can only re-weight existing variants."
+                        ]
+                    }
+                )
```
khvn26, line 280: Can you explain why we need this together with `_validate_multivariate_keys_exist`?

api/experimentation/serializers.py
```diff
@@ -217,13 +215,22 @@ class ExperimentMetricInlineSerializer(serializers.Serializer):  # type: ignore[
     expected_direction = serializers.ChoiceField(choices=ExpectedDirection.choices)
 
 
+class RolloutMultivariateValueSerializer(
```
matthewelwell, line 218: Where did this name come from? I don't quite understand how `Rollout` and `MultivariateValue` are related?

api/features/versioning/versioning_service.py
```diff
@@ -410,6 +455,30 @@ def update_environment_multivariate_options(
     update_multivariate_values(feature_state, reweighted_values)
 
 
+def _upsert_multivariate_option_by_key(
```
matthewelwell, line 458: I'm a bit unsure why we're adding upsert functionality here? Do we need to? 

Without having the full context my thoughts here are that it would complicate things unnecessarily. If we restrict the creation of multivariate values to a single place, then we can more easily implement restrictions such as not being able to create new variations when an experiment is taking place.

api/tests/integration/features/versioning/test_update_flag_endpoints.py
```diff
@@ -889,3 +1155,140 @@ def test_update_flag__unknown_feature_with_multivariate_options__responds_400(
     # Then
     assert response.status_code == 400
     assert "not found in project" in str(response.json())
+
+
+def test_update_flag_option_b__keyed_variants_in_single_request__configures_environment_and_overrides_segment(
```
matthewelwell, line 1160: Why are we still referencing `option_b` here - I thought we were removing the different options?

matthewelwell (commented):
Added a few high-level comments. I think the key question is whether we should allow the creation of new multivariate options as part of these endpoints. IMO we shouldn't...

api/tests/integration/features/versioning/test_update_flag_endpoints.py
emyller: This patch pre-dates that decision, and will likely be closed / re-shaped. Thanks!

emyller:
Superseded by https://github.com/Flagsmith/flagsmith/pull/8102.

Issue #7862, opened 2026-06-24
Self-hosted Enterprise (licence): cannot invite a 2nd user - seat limit enforced against max_seats DB column instead of licence seat count


Holmus:
## Summary

On self-hosted **Enterprise (licence-activated)** instances, inviting any user beyond the first organisation admin fails, even though the licence grants more seats and the UI correctly shows them as available (e.g. "1 of 20 seats used").

Two different, misleading errors surface depending on the invite method, but both have the **same root cause**.

## Affected versions

Present from **v2.217.1** (first release containing #6663) through **latest (v2.245.0)**. Confirmed by reading source at `v2.233.0` and `main`, and reproduced on a self-hosted Enterprise instance running **v2.233.0**. Setting `max_seats` manually (see workaround) resolves it, confirming the mechanism.

## Symptoms

**A. "Invite members" (email invite):** invitee registers, then accepting fails with:
> The organisation you have been invited to has no seats available. Please contact the organisation administrator to resolve this before trying again.

**B. "Copy Invite Link" (shareable link):** invitee registers, then sees:
> We could not validate your invite, please check the invite URL and email address you have entered is correct.

In both cases the org shows e.g. **"1 of 20 seats"** used and the licence grants 20.

## Root cause

`Organisation.over_plan_seats_limit()` only consults subscription metadata (which correctly reads the licence) when `has_paid_subscription()` is true — i.e. when a `subscription_id` is set. A licence-activated Enterprise org has **no `subscription_id`** (that's a Chargebee/SaaS field), so it falls through to the raw `Subscription.max_seats` DB column, which is never populated from the licence and defaults to `MAX_SEATS_IN_FREE_PLAN = 1`.

```python
# api/organisations/models.py
def over_plan_seats_limit(self, additional_seats=0):
    if self.has_paid_subscription():   # bool(subscription_id) -> False for licence-only enterprise
        return self.num_seats + additional_seats > self.subscription.get_subscription_metadata().seats
    # licence-only enterprise lands here:
    return self.num_seats + additional_seats > getattr(self.subscription, "max_seats", MAX_SEATS_IN_FREE_PLAN)
    #                                                   ^ defaults to 1, never set from the licence
```

This contradicts the **display** and **invite-creation** paths, which both go through `get_subscription_metadata()` → `_get_subscription_metadata_for_self_hosted()` → `licence.get_licence_information().num_seats` (= 20). So:

[...]

PR #8014, opened 2026-07-15
fix(organisations): Enforce seat limit against licence-derived metadata

matthewelwell:

## Changes

Contributes to #7862

Route `Organisation.over_plan_seats_limit` through `get_subscription_metadata` unconditionally so licence-only Enterprise orgs enforce the licence's seat count, not the unpopulated `Subscription.max_seats` default of 1.

New unit tests cover both branches of the licence path; `make test opts="tests/unit/organisations/ tests/unit/users/"` and `make typecheck` are clean.

### DIFF
--- a/api/organisations/models.py
+++ b/api/organisations/models.py
@@ -134,14 +134,9 @@ class Organisation(LifecycleModelMixin, SoftDeleteExportableModel):  # type: ign
             },
         )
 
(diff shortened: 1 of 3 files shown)


khvn26 (approved):
Good job on improving coverage.

PR #8018, opened 2026-07-15
perf(Segment Membership): Count all segments in one scan per environment

### LINKED ISSUE #8015: Segment Membership: Segment fanout spending unnecessary compute
Currently, we run (n environments) x (n segments) queries to refresh the segment membership counters.

We can, and should, perform one query per environment via `GROUP BY`.

khvn26:

## Changes

Closes #8015.

In this PR, we optimise the segment membership count calculation by compressing querying all relevant segments for a given environment into a single query, as opposed to scanning an environment once per each segment.

We replace the fan-out with a single `GROUP BY environment_id` scan that counts every segment in one pass via `countIf(<predicate>)` per segment, then maps the columns back to per-segment counts. Each environment is scanned and de-duplicated once regardless of segment count.

Added a test asserting that three segments counted in one query map to the right counts and that a zero-match segment is absent; updated the existing mock-based test for the new row shape. Existing single-segment count tests (including the deleted-identity and percent-regex cases) still pass.

Measured the query shapes against a production environment with 20 segments/~16M-identities:

| Shape | rows read | data | elapsed |
| --- | --- | --- | --- |
| `UNION ALL` (before) | 331M | 25.7 GB | 8.65s |
| single-scan `countIf` (this PR) | 16.6M | 1.29 GB | 0.49s |

The heaviest production projects run 80–159 segments, so the reduction there is proportionally larger, and peak memory falls too.

### DIFF
--- a/api/segment_membership/services.py
+++ b/api/segment_membership/services.py
@@ -159,8 +159,6 @@ def compute_segment_counts_for_project(
     if not count_columns:
         return []
 
-    # One scan per environment: dedupe with FINAL once, then count every segment
-    # in the same pass via `countIf`, rather than re-scanning per segment.
     sql = (
         f"SELECT i.environment_id AS env_key, {', '.join(count_columns)} "
         f"FROM IDENTITIES AS i FINAL "
--- a/api/segment_membership/services.py
+++ b/api/segment_membership/services.py
@@ -113,8 +113,7 @@ def compute_segment_counts_for_project(
     `project`, scanning each environment once.
 
     A single `GROUP BY environment_id` over `IDENTITIES FINAL` counts every
-    segment in one pass via `countIf(<predicate>)` per segment, instead of
-    re-scanning (and re-deduping) the environment once per segment.
+    segment in one pass via `countIf(<predicate>)` per segment.
 
     Returns unsaved `SegmentMembershipCount` instances with `count` and
     keys populated; the caller stamps `last_synced_at` consistently
--- a/api/segment_membership/services.py
+++ b/api/segment_membership/services.py
@@ -110,7 +110,11 @@ def compute_segment_counts_for_project(
     project: Project, cursor: CursorWrapper
 ) -> list[SegmentMembershipCount]:
     """Count identity matches per (canonical-segment, environment) for
-    `project` in one `UNION ALL` query.
+    `project`, scanning each environment once.
+
+    A single `GROUP BY environment_id` over `IDENTITIES FINAL` counts every
+    segment in one pass via `countIf(<predicate>)` per segment, instead of
+    re-scanning (and re-deduping) the environment once per segment.
 
     Returns unsaved `SegmentMembershipCount` instances with `count` and
     keys populated; the caller stamps `last_synced_at` consistently
@@ -128,7 +132,8 @@ def compute_segment_counts_for_project(
 
     dialect = ClickHouseDialect()
     binder = Binder(PyformatParamStyle())
-    select_clauses: list[str] = []
+    count_columns: list[str] = []
+    counted_segment_ids: list[int] = []
     for seg in segments:
         translate_ctx = TranslateContext(
             evaluation_context=EvaluationContext(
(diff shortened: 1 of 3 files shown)


api/segment_membership/services.py
```diff
@@ -149,33 +154,37 @@ def compute_segment_counts_for_project(
...
+        # Columns line up with counted_segment_ids; a zero-match pair is absent.
+        for segment_id, count in zip(counted_segment_ids, row[1:]):
+            if count:
+                membership_counts.append(
+                    SegmentMembershipCount(
+                        segment_id=segment_id,
+                        environment_id=env_id,
+                        count=int(count),
+                    )
+                )
```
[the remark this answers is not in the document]
khvn26: A trade-off I'm willing to live with. Failing queries should be fixed as bugs anyway.

api/segment_membership/services.py
```diff
@@ -149,33 +154,37 @@ def compute_segment_counts_for_project(
...
-    sql = "\nUNION ALL\n".join(select_clauses)
+    # One scan per environment: dedupe with FINAL once, then count every segment
+    # in the same pass via `countIf`, rather than re-scanning per segment.
+    sql = (
+        f"SELECT i.environment_id AS env_key, {', '.join(count_columns)} "
+        f"FROM IDENTITIES AS i FINAL "
+        f"WHERE i.environment_id IN %(env_keys)s AND i.is_deleted = false "
+        f"GROUP BY i.environment_id"
+    )
     cursor.execute(sql, {"env_keys": tuple(env_id_by_key), **binder.params})
```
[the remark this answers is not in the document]
khvn26: Doubt we'll hit it anytime soon: 
- To approach the 256 KB default you'd need both a high segment count and complex predicates at once — roughly ~180+ complex-predicate segments in one project.
- The actual heavy projects from SaaS don't combine those. Nobody's near 256 KB today.

api/tests/unit/segment_membership/test_unit_segment_membership_services.py
```diff
@@ -131,7 +131,8 @@ def test_compute_segment_counts_for_project__unknown_env_key_in_row__skips(
         return_value="TRUE",
     )
     cursor = MagicMock()
-    cursor.fetchall.return_value = [(segment.id, "ghost-env", 99)]
+    # New row shape: (env_key, count-per-segment...); env_key is unknown here
```
emyller:
```suggestion
```

Deixis?

api/segment_membership/services.py
```diff
@@ -149,33 +153,37 @@ def compute_segment_counts_for_project(
...
+        count_columns.append(f"countIf({predicate}) AS c{seg.id}")
+        counted_segment_ids.append(seg.id)
 
-    if not select_clauses:
+    if not count_columns:
         return []
 
-    sql = "\nUNION ALL\n".join(select_clauses)
+    # One scan per environment: dedupe with FINAL once, then count every segment
+    # in the same pass via `countIf`, rather than re-scanning per segment.
```
emyller:
```suggestion
```

Already explained in docstring. Also, the "rather than" portion is deixis — it only makes sense from the perspective of this pull request.

api/tests/unit/segment_membership/test_unit_segment_membership_services.py
```diff
@@ -288,6 +310,44 @@ def test_compute_segment_counts_for_project__deleted_identity__excluded_from_cou
...
+    segment_membership_identities: None,
+    three_segments: dict[str, Segment],
+    project: Project,
+) -> None:
+    # Given / When
+    with connections["clickhouse"].cursor() as cursor:
+        counts = compute_segment_counts_for_project(project, cursor)
+
+    # Then
+    # each column maps to its own segment and the zero-match one is absent
```
emyller:
```suggestion
```

emyller (dismissed):
LGTM with notes on over-commenting slop, and a question: would expanding the query scope to a full organisation — currently it's project-scoped — improve this even further?

khvn26:
> would expanding the query scope to a full organisation — currently it's project-scoped — improve this even further?

I don't think so. There are no more redundant reads to shave off, and this would only save on the round-trip overhead, while substantially increasing the max_query_size risk.

api/tests/unit/segment_membership/test_unit_segment_membership_services.py
khvn26: 116988486

api/segment_membership/services.py
khvn26: 116988486

api/tests/unit/segment_membership/test_unit_segment_membership_services.py
khvn26: 116988486

emyller (approved):
👌

PR #8020, opened 2026-07-15
feat: ClickHouse bring-your-own-warehouse connections

Zaimwa9:

## Changes

Adds the `clickhouse` warehouse connection type so customers can bring their own ClickHouse instance for experimentation. This PR covers connection setup and credentials management; event routing is a follow-up in `flagsmith-analytics-pipeline`, and the frontend ships separately behind the `clickhouse_warehouse` flag.

- New `EncryptedJSONField` stores connection credentials encrypted at rest, keyed off `SECRET_KEY` — no new configuration required.
- The serializer accepts `config` (`host` required, sensible defaults for port/database/username/TLS) and a write-only `credentials: {password}` that is never returned by the API. Omitting credentials on PATCH keeps the stored password; switching the connection to another type clears it.
- Connections are verified with `SELECT 1` on create, on config or credential change, and via the existing test-connection endpoint, setting the status to `connected` or `errored`.
- New verification counter metric and structlog events.

Unit tests (100% diff coverage), `make lint`, `make typecheck`.

Manually: POST `/api/v1/environments/<api_key>/warehouse-connections/` with a `clickhouse` payload returns 201 with `"status": "connected"` or `"errored"`; re-verify via POST `…/test-warehouse-connection/`.


openapi.yaml
```diff
@@ -24887,13 +24887,23 @@ components:
           allOf:
             - $ref: '#/components/schemas/WarehouseConnectionStatusEnum'
           readOnly: true
+        status_detail:
```
Zaimwa9, line 24892: This is nice to have, I believe especially with other warehouses that we don't know so well, it will be helpful to keep track of the error

api/core/fields.py
```diff
@@ -0,0 +1,45 @@
...
+from cryptography.fernet import Fernet, InvalidToken
+from django.conf import settings
+from django.db import models
+
+logger = structlog.get_logger("core")
+
+
+def _get_fernet() -> Fernet:
+    digest = hashlib.sha256(settings.SECRET_KEY.encode()).digest()
+    return Fernet(base64.urlsafe_b64encode(digest))
```
Zaimwa9, line 17: As pointed out by CodeRabbit, it would be better to use a dedicated secret for encryption. I first want to confirm we are happy with fernet to encrypt before over-engineering

Zaimwa9:
@themis-blindfold review

Zaimwa9:
@themis-blindfold what are possible alternatives to using fernet? Pick 3 alternatives and provide pros/cons in 2 lines maximum for each.
Then write a haiku about them

api/core/fields.py
```diff
@@ -0,0 +1,45 @@
...
+import structlog
+from cryptography.fernet import Fernet, InvalidToken
+from django.conf import settings
+from django.db import models
+
+logger = structlog.get_logger("core")
+
+
+def _get_fernet() -> Fernet:
+    digest = hashlib.sha256(settings.SECRET_KEY.encode()).digest()
```
gagantrivedi: we should create a new secret key just for dealing with warehouse connections, it's better to keep them separate

gagantrivedi (changes requested):

api/core/fields.py
Zaimwa9: Done; added `WAREHOUSE_CREDENTIALS_SECRET` env var (defaults to `SECRET_KEY` for backwards compatibility). `_get_fernet()` now derives the Fernet key from the dedicated secret.

api/webhooks/fields.py
```diff
@@ -27,26 +27,5 @@ def run_validators(self, value: str) -> None:
...
-            if (
-                ip.is_loopback
-                or ip.is_private
-                or ip.is_link_local
-                or ip.is_reserved
-                or ip.is_multicast
-            ):
-                self.fail("internal_address")
+        if is_internal_address(hostname):
+            self.fail("internal_address")
```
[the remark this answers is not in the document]
Zaimwa9: out of scope voluntarily

openapi.yaml
```diff
@@ -24889,13 +24889,23 @@ components:
...
           maxLength: 255
         config:
           oneOf:
             - {}
             - type: 'null'
+        credentials:
+          oneOf:
+            - {}
+            - type: 'null'
+          writeOnly: true
```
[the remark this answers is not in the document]
Zaimwa9: this is automatically generated

gagantrivedi (approved):

PR #8027, opened 2026-07-16
fix(e2e): staging E2E failures from shared-project pollution and onboarding flag race

Zaimwa9:

## Changes

Fixes the staging E2E failures red since 15 Jul. Two causes:

1. The deep-link test seeds 60 `e2e_deeplink_*` features in the shared project, filling page 1 of the feature list. Features created by concurrent tests end up on page 2 and their waits time out. The test now uses its own per-run project, cleaned up by the e2etests teardown.
2. `App.js` routes new signups to `/getting-started` or `/create` before `flagsmithIdentify()` settles, so the flag check can miss user-level values. It now waits for identify, capped at 2s so a degraded flags API falls back to routing with the already-loaded flags (today's behaviour) instead of blocking the redirect. The onboarding test also retries its `?connected` navigation, which the features page's URL sync can abort.

Both tests pass against staging, where CI has been failing.

### DIFF
--- a/frontend/e2e/tests/deep-link-test.pw.ts
+++ b/frontend/e2e/tests/deep-link-test.pw.ts
@@ -20,10 +20,6 @@ test.describe('Deep link to feature slideout', () => {
     const { login } = createHelpers(page)
     const api = Project.api
 
-    // Given - an authenticated API session and a dedicated per-run project
-    // with > PAGE_SIZE features. Seeding a shared project would push other
-    // tests' features off page 1 of the list; the e2etests teardown deletes
-    // the E2E user's organisations, so no cleanup is needed here.
     log('Authenticate against the API')
     const loginRes = await request.post(`${api}auth/login/`, {
       data: { email: E2E_USER, password: PASSWORD },
--- a/frontend/e2e/tests/deep-link-test.pw.ts
+++ b/frontend/e2e/tests/deep-link-test.pw.ts
@@ -20,12 +20,10 @@ test.describe('Deep link to feature slideout', () => {
     const { login } = createHelpers(page)
     const api = Project.api
 
-    // Given - an authenticated API session and a dedicated project with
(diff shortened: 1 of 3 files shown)


frontend/e2e/tests/deep-link-test.pw.ts
```diff
@@ -1,6 +1,6 @@
 import { test, expect } from '../test-setup'
 import { createHelpers, LONG_TIMEOUT, log } from '../helpers'
-import { E2E_USER, PASSWORD, E2E_TEST_PROJECT } from '../config'
+import { E2E_USER, PASSWORD } from '../config'
```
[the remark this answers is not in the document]
Zaimwa9: e2e sits outside the webpack alias build. Safe to ignore

khvn26 (dismissed):

frontend/e2e/tests/deep-link-test.pw.ts
```diff
@@ -20,7 +20,12 @@ test.describe('Deep link to feature slideout', () => {
     const { login } = createHelpers(page)
     const api = Project.api
 
-    // Given - an authenticated API session and a project with > PAGE_SIZE features
+    // Given - an authenticated API session and a dedicated project with
+    // > PAGE_SIZE features. The project is created per run: seeding the shared
+    // "My Test Project" fills page 1 of its feature list with e2e_deeplink_*
+    // rows, which hides features that concurrently running tests create (they
+    // sort after "e" and land on page 2). The e2etests teardown deletes all of
+    // the E2E user's organisations, so nothing here needs explicit cleanup.
```
khvn26: nit: I don't think this deictic comment is justified here.

Zaimwa9: Removing this slop in a min

khvn26 (approved):

PR #8028, opened 2026-07-16
feat: clickhouse warehouse setup UI

Zaimwa9:

## Changes

Frontend for the `clickhouse` warehouse connection type, gated behind the `clickhouse_warehouse` flag (off by default). Builds on the backend in #8020 and the stateless test endpoint in #8060.

- New ClickHouse config form, visually identical to the Snowflake one: host, name, port, database, username, password and a TLS toggle, with defaults matching the backend. The host is locked when editing; leaving the password blank on edit keeps the stored credentials.
- Test-before-save flow: Save starts disabled and enables after a connection test against the stateless endpoint (#8060). A successful test confirms inline; a failed test still allows saving, with a warning that events won't be delivered until the connection succeeds. Changing any connection field resets the test; name-only edits save without one.
- No Cancel button on create (the warehouse type cards handle switching); Cancel is kept on edit to return to the connection card.
- The warehouse setup flow offers ClickHouse alongside Flagsmith and Snowflake when the flag is enabled.
- The connection card wires the Test connection button for ClickHouse and surfaces the `status_detail` field: the status tooltip shows the failure reason when the connection is errored, and the test-connection toast reports it instead of a generic message.
- API layer types updated: write-only `credentials`, read-only `status_detail`, new `testWarehouseConnectionConfig` mutation with a `WarehouseConnectionTestResult` response type.

Unit tests for the form validation, payload and test-flow helpers (`npm run test:unit -- --testPathPatterns=clickhouseConfig`), `npm run lint`, `npm run typecheck` (no new errors against the baseline).

Manually: enable the `clickhouse_warehouse` flag, go to Environment Settings → Warehouse, test and connect a ClickHouse instance (both reachable and unreachable), save after a failed test to check the warning, then edit it and re-test with changed connection details.

### DIFF
--- a/frontend/web/components/pages/environment-settings/tabs/warehouse-tab/ClickHouseConfigForm.tsx
+++ b/frontend/web/components/pages/environment-settings/tabs/warehouse-tab/ClickHouseConfigForm.tsx
@@ -90,11 +90,22 @@ const ClickHouseConfigForm: FC<ClickHouseConfigFormProps> = ({
         environmentId,
         warehouse_type: 'clickhouse',
       }).unwrap()
-      setTestState(result.status === 'connected' ? 'connected' : 'errored')
-      setTestDetail(result.status_detail)
+      if (result.status === 'connected') {
+        setTestState('connected')
+        setTestDetail(null)
+        toast('Connection verified')
+      } else {
+        setTestState('errored')
+        setTestDetail(result.status_detail)
+        toast(
+          result.status_detail || 'Connection failed — check your credentials',
+          'danger',
+        )
+      }
     } catch {
       setTestState('errored')
       setTestDetail(null)
+      toast('Failed to test connection', 'danger')
     }
   }
 
@@ -243,12 +254,19 @@ const ClickHouseConfigForm: FC<ClickHouseConfigFormProps> = ({
         </div>
 
         {testState === 'connected' && (
-          <span className='wh-config-form__hint text-success'>
-            Connection verified. You can now save.
-          </span>
+          <div className='d-flex justify-content-end'>
+            <span className='wh-config-form__hint text-success'>
+              Connection verified. You can now save.
+            </span>
+          </div>
         )}
         {testState === 'errored' && (
-          <WarningMessage warningMessage={getTestFailureWarning(testDetail)} />
+          <div className='d-flex justify-content-end'>
+            <WarningMessage
+              warningMessage={getTestFailureWarning(testDetail)}
+              warningMessageClass='mb-0'
+            />
+          </div>
         )}
       </div>
     </form>
--- a/frontend/web/components/pages/environment-settings/tabs/warehouse-tab/ClickHouseConfigForm.tsx
+++ b/frontend/web/components/pages/environment-settings/tabs/warehouse-tab/ClickHouseConfigForm.tsx
@@ -3,6 +3,7 @@ import Button from 'components/base/forms/Button'
 import Input from 'components/base/forms/Input'
 import Switch from 'components/Switch'
 import ErrorMessage from 'components/ErrorMessage'
+import WarningMessage from 'components/WarningMessage'
 import { ClickHouseConfig } from 'common/types/responses'
 import { useTestWarehouseConnectionConfigMutation } from 'common/services/useWarehouseConnection'
 import {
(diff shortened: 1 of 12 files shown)


Zaimwa9:
@themis-blindfold review

Zaimwa9:
@themis-blindfold review once again but this time focus on the frontend design patterns to make sure the used components follow the snowflake designs. Let's flag whether some scss or classes are duplicated, or could be better implemented.
Can I know why you scored 3/5 in product impact while this is clearly a product-focused new addition?

frontend/web/components/pages/environment-settings/tabs/warehouse-tab/ClickHouseConfigForm.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+type ClickHouseConfigFormProps = {
+  onSave: (data: ClickHouseFormData) => Promise<unknown>
+  onCancel: () => void
+  isEdit?: boolean
+  initialConfig?: ClickHouseConfig
+  initialName?: string
+}
+
+const getButtonLabel = (isEdit: boolean, isSaving: boolean): string => {
```
Zaimwa9: *🧹 Nit · ⚡ Quick win*

**`getButtonLabel` is duplicated verbatim from `ConfigForm.tsx`**

Both warehouse config forms define an identical `getButtonLabel` helper. Extract it into a shared module (e.g. alongside `clickhouseConfig.ts` or a new `warehouseFormUtils.ts`) so the next warehouse type doesn't copy it a third time. The same applies to the error message template at line 164.

frontend/web/components/pages/environment-settings/tabs/warehouse-tab/ClickHouseConfigForm.tsx
```diff
@@ -0,0 +1,190 @@
...
+        {error && (
+          <ErrorMessage
+            error={`Failed to ${
+              isEdit ? 'update' : 'create'
+            } warehouse connection. Please try again.`}
+          />
+        )}
+
+        <div className='wh-config-form__actions'>
+          <Button theme='outline' size='small' onClick={onCancel} type='button'>
```
Zaimwa9: *🧹 Nit · ⚡ Quick win*

**Missing `id` attributes on the form buttons**

The Snowflake `ConfigForm.tsx` gives these buttons `id='warehouse-config-cancel'` / `id='warehouse-config-save'` for E2E selectors; the ClickHouse form omits them.

Zaimwa9 (commented):
**Verdict: 🧹 Ship it, nits inside** — no correctness issues, two nits filed inline:

1. **`getButtonLabel` duplication** — identical helper copied from the Snowflake form; should be shared to avoid a third copy when the next warehouse type arrives.
2. **Missing `id` attributes on form buttons** — the Snowflake form has them for E2E selectors; the ClickHouse form omits them.

The PR is well-structured overall: clean type separation for credentials vs config, solid unit tests on the pure logic, correct polling scoping, and consistent use of the shared SCSS and BEM conventions.

_Recovered Themis review, posted manually (bot post failed on an output-format validation error)._

frontend/web/components/pages/environment-settings/tabs/warehouse-tab/index.tsx
```diff
@@ -200,6 +266,11 @@ const WarehouseTab: FC<WarehouseTabProps> = ({ environmentId }) => {
             : undefined
         }
         onSendTestEvent={handleSendTestEvent}
+        onTestConnection={
+          connection.warehouse_type === 'clickhouse'
+            ? handleTestConnection
+            : undefined
+        }
```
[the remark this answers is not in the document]
Zaimwa9: Right now on purpose as snowflake connections are disabled until fully implemented

Zaimwa9:
@themis-blindfold review

Zaimwa9:
@coderabbitai full review

talissoncosta (approved):
Looks pretty good. I've tested that with real ClickHouse Cloud instance.

I've added some notes but not for this PR, but worth a follow-up. 

**1. Trim inputs before building the payload.**
Trailing spaces/newlines from a paste are a classic cause of "auth failed / won't connect". A `.trim()` in `buildClickHousePayload` removes a whole class of confusion for near-zero cost.

**2. Show why Test/Save is disabled.**
An invalid port or missing field silently disables the buttons with no explanation. Inline field-level validation (e.g. "Port must be 1–65535") pairs with the host validation above.

**3. Don't double up failure messaging.**
A failed inline test fires both a danger toast and the inline warning banner for the same action. The banner is the better one (carries the "save anyway / events won't be delivered" context), so drop the toast on inline-test failure.

**4. `autoComplete="new-password"` on the password input** so the browser doesn't offer saved logins.

PR #8035, opened 2026-07-17
feat(experimentation): add per-organisation ingestion infrastructure service

gagantrivedi:

## Changes

Adds a service layer for provisioning per-organisation experiment events ingestion infrastructure, to be consumed by the events ingestion server and hooked up to an API later.

- New `experimentation/ingestion_infra_service.py`: `provision_ingestion_infrastructure(organisation_id)` idempotently creates an S3 events bucket and a DirectPut Firehose delivery stream per organisation.
- The stream uses dynamic partitioning on each record's `environment_key`, so objects land under per-environment S3 namespaces (`events/env_key=.../year=.../...`) and are batched per environment.
- New settings: `INGESTION_EVENTS_BUCKET_PREFIX`, `INGESTION_FIREHOSE_DELIVERY_ROLE_ARN` (a single shared Firehose delivery role, provisioned out-of-band, scoped by bucket naming convention).

Note: with a shared delivery role, per-stream/bucket isolation is enforced by the control plane (only this service sets stream destinations), not IAM; per-organisation roles are a possible v2 hardening.

Unit tests with moto covering fresh provisioning (full destination config), idempotent re-runs, missing-settings errors, and AWS client error propagation. `make lint`, `make typecheck`, and the full experimentation unit suite pass locally.

### DIFF
--- a/api/experimentation/ingestion_infra_service.py
+++ b/api/experimentation/ingestion_infra_service.py
@@ -5,7 +5,6 @@ from functools import lru_cache
 
 import boto3
 import structlog
-from botocore.exceptions import ClientError
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 
@@ -83,22 +82,17 @@ def get_stream_name(organisation_id: int) -> str:
     return f"{STREAM_NAME_PREFIX}{organisation_id}"
 
 
-def _ensure_events_bucket(bucket_name: str, *, organisation_id: int) -> None:
+def _create_events_bucket(bucket_name: str, *, organisation_id: int) -> None:
     s3 = _get_s3_client()
-    try:
-        s3.create_bucket(
-            Bucket=bucket_name,
-            CreateBucketConfiguration={"LocationConstraint": AWS_REGION},
-        )
-    except ClientError as exc:
-        if exc.response["Error"]["Code"] != "BucketAlreadyOwnedByYou":
-            raise
-    else:
-        logger.info(
-            "ingestion_infra.bucket_created",
-            organisation__id=organisation_id,
-            bucket__name=bucket_name,
-        )
+    s3.create_bucket(
+        Bucket=bucket_name,
+        CreateBucketConfiguration={"LocationConstraint": AWS_REGION},
+    )
+    logger.info(
+        "ingestion_infra.bucket_created",
+        organisation__id=organisation_id,
+        bucket__name=bucket_name,
+    )
     s3.put_public_access_block(
         Bucket=bucket_name,
         PublicAccessBlockConfiguration={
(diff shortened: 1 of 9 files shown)


Zaimwa9:
@themis-blindfold review

api/experimentation/ingestion_infra_service.py
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+
+def _create_delivery_stream(
+    stream_name: str,
+    *,
+    bucket_name: str,
+    organisation_id: int,
+) -> None:
+    _get_firehose_client().create_delivery_stream(
```
Zaimwa9, line 178: I saw that we can tag the streams https://docs.aws.amazon.com/firehose/latest/APIReference/API_CreateDeliveryStream.html

and it would directly reconcile the costs with the given organisation if my understanding is correct. Would it make sense to add them sooner than later ?
It needs to be activated in the console

api/experimentation/ingestion_infra_service.py
```diff
@@ -0,0 +1,201 @@
...
+
+def get_bucket_name(organisation_id: int) -> str:
+    return f"{BUCKET_NAME_PREFIX}{organisation_id}-{_get_account_id()}-{AWS_REGION}-an"
+
+
+def get_stream_name(organisation_id: int) -> str:
+    return f"{STREAM_NAME_PREFIX}{organisation_id}"
+
+
+def _create_events_bucket(bucket_name: str, *, organisation_id: int) -> None:
```
Zaimwa9, line 89: Actually I don't think the comment raised by CodeRabbit has been addressed: https://github.com/Flagsmith/flagsmith/pull/8035#pullrequestreview-4734607135

If indeed `create_bucket` is create only and not upsert, we could at least skip a `BucketAlreadyOwnedByYou` error and resume the put operations if there had been failure during the previous creation cycle.

But I agree that we'd also need a retry mechanism somewhere and this can be a follow up. Can we in this case describe the accepted risk as a comment to not forget ?

Zaimwa9 (commented):
2 comments more than requesting changes. For the create flow, a comment would be satisfying imo.
For the tags, it looks interesting to keep track of the costs precisely

api/experimentation/ingestion_infra_service.py
gagantrivedi: Yeah, good shout! done now

api/experimentation/ingestion_infra_service.py
gagantrivedi: I have kept it as create only for now with no reconciliation whatsoever. Handling BucketAlreadyOwnedByYou makes things less clear for now imo when we only handle one case for S3 and don't handle others, and don't do anything for streams. I would like it to be explicit for now.

This whole solution itself is also v0.1. Once this matures we will move to a single stream with a lambda to route traffic between data lakes or external warehouses.

Zaimwa9 (approved):

PR #8059, opened 2026-07-21
Fix #7983: Add EDGE_IDENTITY to allowed integration audit events

### LINKED ISSUE #7983: Slack integration doesn't notify on identity flag overrides (Edge / SaaS)
Flip a flag for a specific identity and nothing shows up in Slack, even though flipping the same flag at the environment or segment level does. The per-identity change just goes unannounced.

This came up from a customer using the Slack integration for observability. Every project and segment level change lands in their channel, but identity overrides are invisible, so they have no way to see when a flag was changed for an individual user.

**What's going on**

The event itself isn't lost. When you override a flag for an identity, Flagsmith does write an audit log entry for it. Things break down one step later, where audit events get handed off to the integrations.

The Slack handler only forwards a fixed set of event types: `FEATURE`, `FEATURE_STATE`, `SEGMENT`, and `EF_VERSION`. On SaaS, identities live on the Edge, and an Edge identity override is recorded as `EDGE_IDENTITY`, which isn't on that list. So it gets dropped before it ever reaches Slack.

Segment and environment changes come through as `SEGMENT` / `FEATURE_STATE`, both on the list, which is exactly why those notifications work and identity ones don't. (One nuance: self-hosted without Edge logs identity overrides as `FEATURE_STATE`, so they do notify there. The gap is specific to Edge, i.e. SaaS.)

**The fix**

Add `EDGE_IDENTITY` to the event types the messaging integrations forward, so identity overrides notify the same way segment and environment changes already do. Same gap affects the other messaging integrations that share that filter, not just Slack.

srijantrpth:

## Changes

Closes #7983

_Please describe._
Added `RelatedObjectType.EDGE_IDENTITY` to the `track_only_feature_related_events` signal filter decorator in `api/audit/signals.py`. This ensures audit events triggered by identity flag overrides in Edge/SaaS environments are properly processed and sent to messaging integrations like Slack, Grafana, Datadog, and Webhooks.

_Please describe._
- Added a new unit test (`test_send_audit_log_event_to_grafana__edge_identity__calls_expected`) in `api/tests/unit/audit/test_unit_audit_signals.py` to cover `EDGE_IDENTITY` events.
- Ran the specific test file locally using `uv run pytest` to verify that the audit signal properly triggers integration events for edge identities.

### DIFF
--- a/api/audit/signals.py
+++ b/api/audit/signals.py
@@ -121,6 +121,7 @@ def track_only_feature_related_events(signal_function):  # type: ignore[no-untyp
         RelatedObjectType.FEATURE_STATE,
         RelatedObjectType.SEGMENT,
         RelatedObjectType.EF_VERSION,
+        RelatedObjectType.EDGE_IDENTITY,
     ]
     return track_only(allowed_types)(signal_function)
 
--- a/api/tests/unit/audit/test_unit_audit_signals.py
+++ b/api/tests/unit/audit/test_unit_audit_signals.py
@@ -596,3 +596,33 @@ def test_send_feature_flag_went_live_signal__non_feature_state_instance__skips_s
 
     # Then
     mock_signal_send.assert_not_called()
+
+def test_send_audit_log_event_to_grafana__edge_identity__calls_expected(
+    mocker: MockerFixture,
+    project: Project,
+) -> None:
+    # Given
+    audit_log_record = AuditLog.objects.create(
+        project=project,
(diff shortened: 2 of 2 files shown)


Zaimwa9:
@themis-blindfold review

srijantrpth:
Hey @Zaimwa9 ! Just checking in here. I've addressed the feedback from the previous review and pushed the updates. Let me know if the new changes look good to you or if you'd like me to tweak anything else. Thanks

PR #8061, opened 2026-07-21
refactor(analytics): make trackEvent/trackTraits error-safe, tag onboarding variant when flags are ready

talissoncosta:
**Reviewing this:** Start with `GettingStartedGate.tsx`, the only behaviour change: the onboarding variant is now written to analytics only after the flag service has answered for the logged-in user, so a stale or default value is never recorded. `api.ts` is mechanical, both tracking helpers wrapped in try/catch. The timing guard is the bit to scrutinise; the rest can be skimmed.

Follow-up to #7960 (rebased on main now that it merged). Two robustness fixes to the shared analytics helpers.

## Changes

- `trackEvent` and `trackTraits` wrapped in try/catch so a failing tracker can't break the caller, e.g. the onboarding bootstrap firing milestone events (kyle's note).
- The getting-started gate writes the `onboarding_variant` user property only once flags have loaded from the server for the current user, closing the race where a stale or default variant could be recorded (Wadii's note; replaces the earlier `setOnce` approach).

- [x] Reproduced the race by delaying the flag-service identify request 4s: without the guard the variant was written mid-fetch (and twice); with the guard, exactly once after the server answered.
- [x] Revisited `/getting-started`: property unchanged (a rewrite with the same value is expected, it's a mutable trait now).
- [ ] Fresh signup on the deploy preview: `onboarding_variant` appears in Amplitude with the served variant (localhost has no Amplitude key).
- [ ] Ad-blocker on (tracker requests fail): onboarding still loads, no uncaught console errors.

### DIFF
--- a/frontend/web/components/pages/onboarding/GettingStartedGate.tsx
+++ b/frontend/web/components/pages/onboarding/GettingStartedGate.tsx
@@ -1,4 +1,6 @@
 import React, { FC, useEffect } from 'react'
+import flagsmith from '@flagsmith/flagsmith'
+import ConfigProvider from 'common/providers/ConfigProvider'
 import {
   getOnboardingVariant,
   isSinglePageOnboarding,
@@ -8,13 +10,21 @@ import GettingStartedPage from 'components/pages/GettingStartedPage'
 import OnboardingFlow from './OnboardingFlow'
 
 const GettingStartedGate: FC = () => {
+  // ConfigProvider re-renders the gate on every SDK fetch; only tag the
+  // variant once the server has answered for this identity.
+  const trustworthy =
+    !flagsmith.loadingState?.isFetching &&
+    flagsmith.loadingState?.source === 'SERVER' &&
+    !!flagsmith.getContext().identity
+
   const variant = getOnboardingVariant()
 
   useEffect(() => {
+    if (!trustworthy) return
     API.trackTraits({ onboarding_variant: variant })
-  }, [variant])
+  }, [trustworthy, variant])
 
   return isSinglePageOnboarding() ? <OnboardingFlow /> : <GettingStartedPage />
 }
 
-export default GettingStartedGate
+export default ConfigProvider(GettingStartedGate)
--- a/frontend/web/project/api.ts
+++ b/frontend/web/project/api.ts
@@ -309,27 +309,33 @@ const API = {
     label?: string
     extra?: Record<string, any>
   }): void {
-    if (Project.ga) {
-      if (Project.logAnalytics) console.log('ANALYTICS EVENT', data)
-      if (!data || !data.category || !data.event)
-        return console.error('Invalid event provided', data)
-      if (data.category === 'First')
-        API.postEvent(
-          `${data.event}${data.extra ? ` ${JSON.stringify(data.extra)}` : ''}`,
-          'first_events',
-        )
-      ga('send', {
-        eventAction: data.event,
-        eventCategory: data.category,
-        eventLabel: data.label,
-        hitType: 'event',
-      })
-    }
-    if (Project.amplitude) {
-      amplitude.track(data.event, {
-        category: data.category,
-        ...(data.extra || {}),
-      })
+    try {
+      if (Project.ga) {
+        if (Project.logAnalytics) console.log('ANALYTICS EVENT', data)
+        if (!data || !data.category || !data.event)
+          return console.error('Invalid event provided', data)
+        if (data.category === 'First')
+          API.postEvent(
+            `${data.event}${
+              data.extra ? ` ${JSON.stringify(data.extra)}` : ''
+            }`,
+            'first_events',
+          )
+        ga('send', {
+          eventAction: data.event,
+          eventCategory: data.category,
+          eventLabel: data.label,
+          hitType: 'event',
+        })
+      }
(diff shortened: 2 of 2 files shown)


Zaimwa9:
@themis-blindfold review this pull request. Take into account that this a follow up from PR #7960. Use the review from #7960 to contextualize this PR

Zaimwa9:
@themis-blindfold  is there still anything left to address from the #7960 pull request, related to the use of the feature flag ?

frontend/web/project/api.ts
```diff
@@ -309,27 +309,33 @@ const API = {
...
+        })
+      }
+      if (Project.amplitude) {
+        amplitude.track(data.event, {
+          category: data.category,
+          ...(data.extra || {}),
+        })
+      }
+    } catch (err) {
+      console.error('Error tracking event', err)
```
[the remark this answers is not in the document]
talissoncosta: I will leave as it is as they aren't new console calls.

frontend/web/project/api.ts
```diff
@@ -352,6 +358,16 @@ const API = {
...
   },
+
+  trackTraitsOnce(traits: Record<string, any>): void {
+    if (Project.amplitude && traits) {
+      const identifyObj = new amplitude.Identify()
+      Object.entries(traits).forEach(([key, value]) =>
+        identifyObj.setOnce(key, value),
+      )
+      amplitude.identify(identifyObj)
+    }
```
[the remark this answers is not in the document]
talissoncosta: Good catch, done in 09cb98287. Wrapped both trackTraitsOnce and trackTraits (same shape, same useEffect risk) in try/catch, matching the existing identify helper.

frontend/web/project/api.ts
```diff
@@ -344,12 +350,30 @@ const API = {
...
+          identifyObj.set(key, value),
+        )
+        amplitude.identify(identifyObj)
+      }
+    } catch (err) {
+      console.error('Error tracking traits', err)
+    }
+  },
+
+  trackTraitsOnce(traits: Record<string, any>): void {
```
Zaimwa9: My bad as I am the one pushing towards `setOnce` but I end up thinking it wasn't the right call as we might set a wrong default variant in an immutable way.

I looked around. I think we should get rid of this one and use the flagsmith instance + useEffect to determine whether we have the right variant and dispatch it with `extra: variant` in the constants 
I poked around with claude and I agree with most of it (the flagsmith part etc) but something I didn't see was the store could help so I let you check whether it's worth exploring

```
const GettingStartedGate: FC = () => {
  const [, bump] = useReducer((x) => x + 1, 0)
  useEffect(() => {
    ConfigStore.on('change', bump)
    return () => ConfigStore.off('change', bump)
  }, []) // => this part with the useReducer

  const ready = flagsmith.initialised && !flagsmith.loadingState?.isFetching
  const variant = getOnboardingVariant()

  useEffect(() => {
    if (!ready) return
    API.trackTraits({ onboarding_variant: variant })
  }, [ready, variant])

  return isSinglePageOnboarding() ? <OnboardingFlow /> : <GettingStartedPage />
}
```

Zaimwa9 (changes requested):
See my comment (sorry for the back and forth) let me know what you think about it

frontend/web/project/api.ts
talissoncosta: Done in 2d67667. Dropped trackTraitsOnce, the gate tags with plain trackTraits once the identify fetch has settled.

frontend/web/components/pages/onboarding/GettingStartedGate.tsx
```diff
@@ -8,13 +10,21 @@ import GettingStartedPage from 'components/pages/GettingStartedPage'
 import OnboardingFlow from './OnboardingFlow'
 
 const GettingStartedGate: FC = () => {
+  // ConfigProvider re-renders the gate on every SDK fetch; only tag the
+  // variant once the server has answered for this identity.
+  const trustworthy =
+    !flagsmith.loadingState?.isFetching &&
+    flagsmith.loadingState?.source === 'SERVER' &&
+    !!flagsmith.getContext().identity
```
talissoncosta, line 18: Checking `initialised` alone isn't enough here: with flag caching on, the SDK reports ready from the previous session's flags before the current user's fetch settles. All three checks are load-bearing, verified by artificially delaying the fetch 4s.

frontend/web/components/pages/onboarding/GettingStartedGate.tsx
```diff
@@ -8,13 +10,21 @@ import GettingStartedPage from 'components/pages/GettingStartedPage'
...
+    if (!trustworthy) return
     API.trackTraits({ onboarding_variant: variant })
-  }, [variant])
+  }, [trustworthy, variant])
 
   return isSinglePageOnboarding() ? <OnboardingFlow /> : <GettingStartedPage />
 }
 
-export default GettingStartedGate
+export default ConfigProvider(GettingStartedGate)
```
talissoncosta, line 30: App is already wrapped in ConfigProvider, but that doesn't help here: the route tree is built once at module load, so App re-renders never reach route components. The gate needs its own subscription to re-run the effect when flags land.

Zaimwa9 (approved):
Thanks! Lgtm

PR #8023, opened 2026-07-15
refactor(styles): theme-adaptive shared code theme

talissoncosta:
### Reviewing this

Code snippets in the app (SDK examples, flag values) now switch with light/dark mode instead of always being light. Almost all of it is CSS in `_tokens.scss` + `_hljs.scss`; the other files just use it. The only fiddly part: making our colours win over the support-chat widget (Pylon), which injects its own code CSS, and only on deployed envs. The accordion header above code blocks is left unchanged.

Now also carries the rest of the code-snippet work (moved off #7989 plus Slack review feedback): markup tag colours, the corrected React snippet, and Copy Code no longer pasting `&lt;` entities.

- [x] I have read the [Contributing Guide](/Flagsmith/flagsmith/blob/main/CONTRIBUTING.md).
- [ ] I have added information to `docs/` if required so people know about the feature.
- [x] I have filled in the "Changes" section below.
- [x] I have filled in the "How did you test this code" section below.

## Changes

Code snippets in the app (SDK examples, flag values, etc.) now follow light/dark mode instead of always being dark.

- New `--color-code-*` colour tokens for syntax, separate from the UI text colours
- Shared code styling now lives in one place; the onboarding card reuses it
- Our colours override the support-chat widget's (Pylon) own code CSS, deployed envs only
- Markup tag names render blue, not red (red reads as an error); the `< >` brackets stay grey, like an IDE
- Corrected the React init snippet to use `FlagsmithProvider` + `useFlags`
- Copy Code now copies real `< >`: snippet sources were pre-escaped HTML entities, so pasting into an editor produced `&lt;FlagsmithProvider`
- Embedded code scrollbar only shows on overflow
- Adds Storybook stories; removes dead variables.

## How did you test this code?

Toggle light/dark and check code snippets are readable in both (not always dark):

- [ ] Onboarding code cards - /getting-started  ( behind ff - `onboarding_quickstart_flow` ); scrollbar only when the snippet overflows
- [ ] A modal with a code example (e.g. Create Webhook)
- [ ] Flag value editor (edit a feature's value)
- [ ] An SDK snippet (CodeHelp, e.g. features page): React snippet reads `FlagsmithProvider` + `useFlags`; tag names blue, `< >` grey
- [ ] Copy Code on the React snippet, paste into an editor: real `< >`, no `&lt;` entities
- [ ] Accordion header above code blocks looks the same as before
[...]

PR #8062, opened 2026-07-21
refactor(code-help): escape snippets centrally instead of in source strings

talissoncosta:
> **Reviewing this:** The bug: Copy Code put HTML entities (`&lt;`) in the clipboard, because snippet sources were pre-escaped and rendered as raw HTML. Now `Highlight` escapes everything it renders, once, and the sources are real code. Start with `Highlight.js` (one line); the snippet files are a mechanical entity-to-bracket swap, safe to skim. The two installs (Java, .NET) are the subtle ones: they escaped at runtime, not in the source.

## Changes

Split out of #8023 to keep that PR to theming only.

Copy Code was shipping HTML entities (`&lt;FlagsmithProvider`) to the clipboard: snippet sources were pre-escaped because `CodeHelp` rendered them as raw HTML. Escape centrally in `Highlight` instead.

- Snippet sources store real `< >` (literal entities removed; Java/.NET drop their runtime `Utils.escapeHtml`); `Highlight` always escapes at render, `preventEscape` is gone.
- Copy Code now pastes working code; rendering is unchanged from main.
- The onboarding card's unescape workaround is dead and removed.
- The prose install notes (curl, Flutter, Rust) drop their `<a>` tags: the links were already dead on prod (the highlighter strips them) and would have shown as raw markup here; same text, no markup.

- [ ] Features page → SDK integration snippet: React shows `<FlagsmithProvider>`, Java shows the Maven `<dependency>` XML, .NET shows `<PackageReference>` (not `&lt;`)
- [ ] Copy Code on the React and Java snippets, paste into an editor: real brackets, working code
- [ ] Onboarding `/getting-started` (behind ff `onboarding_quickstart_flow`): code cards render and copy clean
- [ ] MCP integration page: config snippet renders unchanged
- [ ] Install snippets for curl, Flutter and Rust: prose renders without `<a href>` markup, same text as prod
- [ ] Flag value editor: a value containing `< >` still edits and renders (path unchanged, quick sanity)

`lint` + `typecheck` pass.

## Screenshots

| | Before (main) | After |
|---|---|---|
| Java SDK snippet on screen | | |
| React snippet pasted into an editor | | |

### DIFF
--- a/frontend/common/code-help/create-user/create-user-react.js
+++ b/frontend/common/code-help/create-user/create-user-react.js
@@ -10,7 +10,7 @@ import { FlagsmithProvider } from '${NPM_CLIENT}/react';
 
 export default function App() {
   return (
-    &lt;FlagsmithProvider
+    <FlagsmithProvider
       options={{
         environmentID: '${envId}',${
   Constants.isCustomFlagsmithUrl()
@@ -20,9 +20,9 @@ export default function App() {
         identity: '${userId || USER_ID}',
         traits: {${TRAIT_NAME}: 21},
       }}
-      flagsmith={flagsmith}&gt;
+      flagsmith={flagsmith}>
       {...Your app}
-    &lt;/FlagsmithProvider>
+    </FlagsmithProvider>
   );
 }
 
--- a/frontend/common/code-help/init/init-react.js
+++ b/frontend/common/code-help/init/init-react.js
@@ -3,13 +3,24 @@ import Constants from 'common/constants'
 export default (
   envId,
   { FEATURE_NAME, FEATURE_NAME_ALT, LIB_NAME, NPM_CLIENT },
-) => `// App root
-import ${LIB_NAME} from "${NPM_CLIENT}";
-import { FlagsmithProvider } from '@flagsmith/flagsmith/react';
+) => `import ${LIB_NAME} from "${NPM_CLIENT}";
+import { FlagsmithProvider, useFlags } from '${NPM_CLIENT}/react';
+
+export function HomePage() {
+  const flags = useFlags(['${FEATURE_NAME}','${FEATURE_NAME_ALT}']); // only causes re-render if specified flag values / traits change
+  const ${FEATURE_NAME} = flags.${FEATURE_NAME}.enabled
+  const ${FEATURE_NAME_ALT} = flags.${FEATURE_NAME_ALT}.value
+  return (
+    <>
+      {\`${FEATURE_NAME}: \${${FEATURE_NAME}}\`}
+      {\`${FEATURE_NAME_ALT}: \${${FEATURE_NAME_ALT}}\`}
+    </>
+  );
+}
 
 export default function App() {
   return (
-    &lt;FlagsmithProvider
+    <FlagsmithProvider
       options={{
         environmentID: '${envId}',${
   Constants.isCustomFlagsmithUrl()
@@ -17,21 +28,8 @@ export default function App() {
     : ''
 }
       }}
-      flagsmith={flagsmith}&gt;
-      {...Your app}
-    &lt;/FlagsmithProvider>
-  );
-}
-
-// Home Page
-import ${LIB_NAME} from '${NPM_CLIENT}';
-import { useFlags, useFlagsmith } from '${NPM_CLIENT}/react';
-
-export default function HomePage() {
-  const flags = useFlags(['${FEATURE_NAME}','${FEATURE_NAME_ALT}']); // only causes re-render if specified flag values / traits change
-  const ${FEATURE_NAME} = flags.${FEATURE_NAME}.enabled
-  const ${FEATURE_NAME_ALT} = flags.${FEATURE_NAME_ALT}.value
-  return (
-    &lt;>{...}&lt;/>
+      flagsmith={${LIB_NAME}}>
+      <HomePage />
+    </FlagsmithProvider>
   );
 }`
--- a/frontend/common/code-help/install/install-curl.js
+++ b/frontend/common/code-help/install/install-curl.js
@@ -1,5 +1,5 @@
-export default () => `You can access the API directly with tools like <a href="https://curl.haxx.se/">curl</a> or <a href='https://httpie.org/'>httpie</a>, 
+export default () => `You can access the API directly with tools like curl or httpie,
 or with clients for languages that we do not currently have SDKs for.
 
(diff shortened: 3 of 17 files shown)


Zaimwa9 (dismissed):

talissoncosta:
@Zaimwa9 can I get a new stamp? I had to rebase it

frontend/web/components/Highlight.js
```diff
@@ -134,7 +134,7 @@ class Highlight extends React.Component {
     }
 
     const raw = this.getRawHtml()
-    const html = this.props.preventEscape ? raw : escapeHtml(raw)
+    const html = escapeHtml(raw)
```
talissoncosta, line 137: Always escaping is safe: the only two consumers that opted out (`CodeHelp`, `MCPSnippet`) did so because their sources were pre-escaped, and those sources are now plain text. Every other consumer already went through this path.

frontend/common/code-help/install/install-java.js
```diff
@@ -1,11 +1,9 @@
-import Utils from 'common/utils/utils'
-
 export default () => `// Maven
-${Utils.escapeHtml('<dependency>')}
-    ${Utils.escapeHtml('<groupId>com.flagsmith</groupId>')}
-    ${Utils.escapeHtml('<artifactId>flagsmith-java-client</artifactId>')}
-    ${Utils.escapeHtml('<version>7.4.1</version>')}
-${Utils.escapeHtml('</dependency>')}
+<dependency>
```
talissoncosta, line 2: Java and .NET didn't have literal entities in source; they escaped at runtime via `Utils.escapeHtml`, so the first sweep missed them and they double-escaped (showed `&lt;dependency&gt;` on screen), plus the same clipboard bug. After this, `Utils.escapeHtml` had no callers left, so a later commit on this branch removes it.

frontend/web/components/pages/onboarding/OnboardingConnectPanel/sdkSnippets.ts
```diff
@@ -14,14 +11,6 @@ import { SdkLang } from './sdkLangs'
 // exported); if that placeholder ever changes, update this too.
 const PLACEHOLDER_FLAG = 'my_cool_feature'
```
talissoncosta, line 12: The removed `unescapeHtml` helper existed only to undo the pre-escaped sources before the onboarding card rendered them. Dead once the sources are clean.

frontend/common/code-help/create-user/create-user-react.js
```diff
@@ -20,9 +20,9 @@ export default function App() {
         identity: '${userId || USER_ID}',
         traits: {${TRAIT_NAME}: 21},
       }}
-      flagsmith={flagsmith}&gt;
+      flagsmith={flagsmith}>
       {...Your app}
-    &lt;/FlagsmithProvider>
+    </FlagsmithProvider>
```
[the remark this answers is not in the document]
talissoncosta: Pre-existing (this PR only swaps the entities), but real. A fix exists but is parked with the wider snippet-accuracy sweep (#8086 is closed for now pending that discussion); the missing client import will land there.

frontend/common/code-help/init/init-react.js
```diff
@@ -3,35 +3,33 @@ import Constants from 'common/constants'
...
+export function HomePage() {
+  const flags = useFlags(['${FEATURE_NAME}','${FEATURE_NAME_ALT}']); // only causes re-render if specified flag values / traits change
+  const ${FEATURE_NAME} = flags.${FEATURE_NAME}.enabled
+  const ${FEATURE_NAME_ALT} = flags.${FEATURE_NAME_ALT}.value
+  return (
+    <>
+      {\`${FEATURE_NAME}: \${${FEATURE_NAME}}\`}
+      {\`${FEATURE_NAME_ALT}: \${${FEATURE_NAME_ALT}}\`}
+    </>
+  );
```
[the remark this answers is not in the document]
talissoncosta: A fix exists but is parked with the wider snippet-accuracy sweep (#8086 is closed for now pending that discussion).

frontend/common/code-help/init/init-react.js
```diff
@@ -3,35 +3,33 @@ import Constants from 'common/constants'
...
-) => `// App root
-import ${LIB_NAME} from "${NPM_CLIENT}";
-import { FlagsmithProvider } from '@flagsmith/flagsmith/react';
+) => `import ${LIB_NAME} from "${NPM_CLIENT}";
+import { FlagsmithProvider, useFlags } from '${NPM_CLIENT}/react';
+
+export function HomePage() {
+  const flags = useFlags(['${FEATURE_NAME}','${FEATURE_NAME_ALT}']); // only causes re-render if specified flag values / traits change
+  const ${FEATURE_NAME} = flags.${FEATURE_NAME}.enabled
+  const ${FEATURE_NAME_ALT} = flags.${FEATURE_NAME_ALT}.value
```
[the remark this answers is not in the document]
talissoncosta: Valid but pre-existing and systemic across all the templates, not this PR. Whether dashed flag names are supported needs a call first; tracking it separately rather than rewriting snippet style here.

frontend/common/code-help/install/install-curl.js
```diff
@@ -1,5 +1,5 @@
-export default () => `You can access the API directly with tools like <a href="https://curl.haxx.se/">curl</a> or <a href='https://httpie.org/'>httpie</a>, 
+export default () => `You can access the API directly with tools like curl or httpie,
```
talissoncosta, line 1: The dropped `<a>` tags have been dead since at least 2021: the highlighter replaces the block content with highlighted plain text, stripping anchor elements, so prod shows exactly this text with no working links (the old `.hljs a` CSS was styling nothing). Copy Code also stops shipping the markup. @kyle-ssg , as the original author: sanity-check welcome that nothing else relied on these.

frontend/common/utils/utils.tsx
```diff
@@ -180,12 +180,6 @@ const Utils = Object.assign({}, BaseUtils, {
     }
     return null
   },
-  escapeHtml(html: string) {
```
talissoncosta, line 183: Zero callers after this branch (Highlight has its own local escapeHtml for rendering). Removed so the next snippet author doesn't reach for it to pre-escape content again, which is the pattern that caused the Java/.NET bug.

Zaimwa9 (approved):

PR #8072, opened 2026-07-22
fix(workflows): block stale change request commits that conflict with published changes

### LINKED ISSUE #7931: Change Requests can silently overwrite concurrent changes on commit
### Summary
On v2 feature versioning a Change Request captures the overrides it touches when created. Another CR can then change those overrides and publish. That makes the first CR stale but it can still be committed. On commit it reverts the newer changes back to its captured values. The only feedback is a passive notice that a change request was published since this one was created. 

### Reproduction
1. Create a CR that only reorders segment override positions on a feature.
2. Have another CR change the value or enabled state of one of those overrides and publish it.
3. Commit the reorder CR.
4. The concurrent change is reverted. Only a passive warning shows.

### Notes
- Stale detection already exists. It runs on scheduled publishes and blocks them on conflict. It looks like a manual commit skips it.
- A reorder captures each moved override's full state, not just its position. That is why it reverts unrelated fields.

### Requested outcome
- Run the existing conflict check on manual commits too. Flag the CR as out of date. Notify the author. Block the commit or require a refresh first.
- A reorder should only change ordering.
- The diff should separate the author's edits from values that were only carried along.

JSap0914:

## Changes

Closes #7931

On v2 feature versioning, a Change Request captures the overrides it touches when it is created. If another Change Request then modifies and publishes those same overrides, the first CR becomes stale, yet it could still be committed manually — silently reverting the newer values. The only feedback was a passive notice.

Stale detection already existed, but it only ran for **scheduled** publishes inside the task processor (`publish_version_change_set(..., is_scheduled=True)`). A **manual commit** publishes its change sets immediately (`VersionChangeSet.publish` calls `publish_version_change_set` synchronously with `is_scheduled=False`), so the conflict check was skipped entirely.

This PR runs the existing conflict check on manual commits too:

- `ChangeRequestCommitService.commit` now calls a new `_check_for_conflicts()` **before** publishing anything. For every change set that would publish immediately, it reuses `VersionChangeSet.get_conflicts()` and raises a new `ChangeRequestConflictError` (HTTP 409) when the change request is out of date. The commit is blocked atomically, before any state is written.
- Scheduled change sets are left untouched — they are still conflict-checked when their task fires (further conflicts can appear before they go live).
- Change requests with `ignore_conflicts` set continue to commit as before.

`ChangeRequestConflictError` is added alongside the other change-request exceptions in `features/workflows/core/exceptions.py`.

Scope note: this addresses the core of the report — running the existing conflict check on manual commits and blocking the stale commit. The reorder-only diff separation described in the issue's "Requested outcome" is a larger, separate change and is intentionally out of scope here.

Added focused regression tests in `api/tests/unit/features/versioning/test_unit_versioning_tasks.py`:

- `test_manual_commit__conflict_with_published_change__blocks_commit` — a stale CR committed immediately after a conflicting CR is published now raises `ChangeRequestConflictError`, `committed_at` stays `None`, and the newer value survives. Verified this test **fails** (`DID NOT RAISE`) without the fix and passes with it.
- `test_manual_commit__conflict_with_ignore_conflicts__commits` — a CR with `ignore_conflicts=True` still commits and its values win.

Gates run locally (Postgres 15.5, `uv run pytest`):

- `tests/unit/features/versioning/test_unit_versioning_tasks.py` and `tests/unit/features/workflows/` — 56 passed, including the pre-existing scheduled-conflict email test (unchanged behaviour).
- `ruff check` and `ruff format --check` — clean.
- `mypy` on the changed source modules — no issues.

### DIFF
--- a/api/core/workflows_services.py
+++ b/api/core/workflows_services.py
@@ -8,7 +8,10 @@ from environments.tasks import rebuild_environment_document
 from features.versioning.models import EnvironmentFeatureVersion
 from features.versioning.signals import environment_feature_version_published
 from features.versioning.tasks import trigger_update_version_webhooks
-from features.workflows.core.exceptions import ChangeRequestNotApprovedError
+from features.workflows.core.exceptions import (
+    ChangeRequestConflictError,
+    ChangeRequestNotApprovedError,
+)
 
 if TYPE_CHECKING:
     from features.workflows.core.models import ChangeRequest
@@ -27,6 +30,7 @@ class ChangeRequestCommitService:
                 "Change request has not been approved by all required approvers."
             )
 
+        self._check_for_conflicts()
         self._publish_feature_states()
         self._publish_environment_feature_versions(committed_by)
         self._publish_change_sets(committed_by)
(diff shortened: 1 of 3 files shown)


khvn26:
Closing seeing no follow-ups on the automated review.

PR #8080, opened 2026-07-23
feat(experimentation): seed per-environment ingestion destination in Redis

gagantrivedi:

## Changes

The Django side of [flagsmith-analytics-pipeline#21](https://github.com/Flagsmith/flagsmith-analytics-pipeline/pull/21): the pipeline now routes each environment's events to a per-destination Firehose stream, resolved from a new Redis key. This seeds that key.

- Adds `set_ingestion_destination` / `delete_ingestion_destination` to `ingestion_sync_service`, writing `experimentation:environment_destinations:<client_api_key>` = the org's stream name.
- Writes the destination when provisioning external-warehouse ingestion, keyed on the environment's client key so `ser.*` keys follow their environment; the value is the stream name actually recorded on the infrastructure row.
- Clears it alongside the auth keys on connection delete (a no-op for Flagsmith connections, which have no custom stream).

Flagsmith-warehouse connections write no destination key, so the pipeline falls back to its default stream.

Unit tests: assert the destination key is written on external-warehouse provisioning (keyed on the client key, valued with the provisioned stream) and removed on teardown, plus the `set`/`delete` Redis calls. Both changed modules stay at 100% coverage.

### DIFF
--- a/api/experimentation/tasks.py
+++ b/api/experimentation/tasks.py
@@ -64,14 +64,16 @@ def provision_external_warehouse_ingestion_infrastructure(environment_id: int) -
         return
 
     infrastructure = enable_ingestion_for_organisation(environment.project.organisation)
-    write_environment_ingestion_keys(environment_id)
-    # Route the environment's events to the org's stream; absent this key the
-    # pipeline falls back to its default stream.
-    assert infrastructure.stream_name is not None  # set once provisioned
+    if infrastructure.stream_name is None:
+        raise RuntimeError("Provisioned ingestion infrastructure has no stream name")
+    # Set the destination before publishing the ingestion keys: the keys gate
+    # the pipeline, so a key without a destination would route events to the
+    # default stream until this write lands.
     ingestion_sync_service.set_ingestion_destination(
         environment.api_key,
         stream_name=infrastructure.stream_name,
     )
+    write_environment_ingestion_keys(environment_id)
 
 
 @register_task_handler()
--- a/api/experimentation/tasks.py
+++ b/api/experimentation/tasks.py
@@ -50,6 +50,7 @@ def remove_environment_ingestion_keys(environment_id: int) -> None:
     ingestion_sync_service.delete_ingestion_key(environment.api_key)
     for api_key in environment.api_keys.all():
         ingestion_sync_service.delete_ingestion_key(api_key.key)
+    ingestion_sync_service.delete_ingestion_destination(environment.api_key)
 
 
 @register_task_handler()
@@ -62,8 +63,15 @@ def provision_external_warehouse_ingestion_infrastructure(environment_id: int) -
     if environment is None:
         return
 
-    enable_ingestion_for_organisation(environment.project.organisation)
+    infrastructure = enable_ingestion_for_organisation(environment.project.organisation)
     write_environment_ingestion_keys(environment_id)
+    # Route the environment's events to the org's stream; absent this key the
+    # pipeline falls back to its default stream.
+    assert infrastructure.stream_name is not None  # set once provisioned
+    ingestion_sync_service.set_ingestion_destination(
+        environment.api_key,
+        stream_name=infrastructure.stream_name,
+    )
 
 
 @register_task_handler()
--- a/api/experimentation/tasks.py
+++ b/api/experimentation/tasks.py
@@ -64,7 +64,7 @@ def provision_external_warehouse_ingestion_infrastructure(environment_id: int) -
         return
 
     infrastructure = enable_ingestion_for_organisation(environment.project.organisation)
-    if infrastructure.stream_name is None:
+    if not infrastructure.stream_name:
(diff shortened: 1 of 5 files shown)


gagantrivedi:
@themis-blindfold review

api/experimentation/tasks.py
```diff
@@ -62,8 +63,15 @@ def provision_external_warehouse_ingestion_infrastructure(environment_id: int) -
...
-    enable_ingestion_for_organisation(environment.project.organisation)
+    infrastructure = enable_ingestion_for_organisation(environment.project.organisation)
     write_environment_ingestion_keys(environment_id)
+    # Route the environment's events to the org's stream; absent this key the
+    # pipeline falls back to its default stream.
+    assert infrastructure.stream_name is not None  # set once provisioned
+    ingestion_sync_service.set_ingestion_destination(
+        environment.api_key,
+        stream_name=infrastructure.stream_name,
+    )
```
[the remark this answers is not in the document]
gagantrivedi: Good catch on both — applied in d85d1608e. Destination is now written before `write_environment_ingestion_keys`, since the ingestion keys are what gate the pipeline: publishing them first could route events to the default stream until the destination write lands. Also swapped the `assert` for an explicit `raise RuntimeError` (so `-O` can't strip it) and added a test covering that guard.

api/experimentation/tasks.py
```diff
@@ -62,7 +63,16 @@ def provision_external_warehouse_ingestion_infrastructure(environment_id: int) -
     if environment is None:
         return
 
-    enable_ingestion_for_organisation(environment.project.organisation)
+    infrastructure = enable_ingestion_for_organisation(environment.project.organisation)
+    if infrastructure.stream_name is None:
+        raise RuntimeError("Provisioned ingestion infrastructure has no stream name")
```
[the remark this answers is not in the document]
gagantrivedi: Fixed in 98bc9e429 — swapped the guard to `if not infrastructure.stream_name:` so blank names are rejected alongside `None`, and parametrised the error-path test over both `None` and `""`.

Zaimwa9:
@themis-blindfold review

api/tests/unit/experimentation/test_ingestion_sync_service.py
```diff
@@ -99,6 +99,48 @@ def test_delete_ingestion_key__valid_key__deletes_from_redis(
...
+    mock_client = mocker.Mock()
+    mocker.patch(
+        "experimentation.ingestion_sync_service._get_client",
+        return_value=mock_client,
+    )
+
+    # When
+    ingestion_sync_service.set_ingestion_destination(
+        "client-env-key",
+        stream_name="events-ingestion-org-1",
```
Zaimwa9, line 115: NIT: Just to avoid any confusion as it lives in a different repo. Can we use `events-dest` in the mocks

api/experimentation/tasks.py
```diff
@@ -62,7 +63,16 @@ def provision_external_warehouse_ingestion_infrastructure(environment_id: int) -
...
-    enable_ingestion_for_organisation(environment.project.organisation)
+    infrastructure = enable_ingestion_for_organisation(environment.project.organisation)
+    if not infrastructure.stream_name:
+        raise RuntimeError("Provisioned ingestion infrastructure has no stream name")
+    # Set the destination before publishing the ingestion keys: the keys gate
+    # the pipeline, so a key without a destination would route events to the
+    # default stream until this write lands.
+    ingestion_sync_service.set_ingestion_destination(
+        environment.api_key,
+        stream_name=infrastructure.stream_name,
```
Zaimwa9, line 74: Sorry if I misunderstood, but we are prefixing `STREAM_NAME_PREFIX = "events-ingestion-org-"` ([here](https://github.com/Flagsmith/flagsmith/pull/8035/changes#diff-7a58f7175465d1e13e3907068d7ac45264e9f0efe3e1c73d6019e6f566e5bd85R19-R20)

which would fail analytics pipeline validation no?

Zaimwa9:
yeah it's a bit useless after coderabbit

Zaimwa9 (commented):
Can we check the prefix name ? Otherwise lgtm

api/experimentation/tasks.py
gagantrivedi: it won't fail. The prefix check was dropped from the pipeline PR: https://github.com/Flagsmith/flagsmith-analytics-pipeline/pull/21#discussion_r3636665965

api/tests/unit/experimentation/test_ingestion_sync_service.py
gagantrivedi: The mocks use `events-ingestion-org-1` deliberately — that's our actual stream name (`STREAM_NAME_PREFIX = "events-ingestion-org-"` in `ingestion_infra_service` + pulumi), so the test mirrors what we really write to Redis

Zaimwa9 (approved):

PR #8084, opened 2026-07-23
perf(features): Decouple feature list query cost from project version history

khvn26:

## Changes

`GET /api/v1/projects/<id>/features/?environment=<id>` was slow even for modestly sized projects (741ms for 134 flags against production), and got slower as version history accumulated. Profiling against a seeded project (200 features, 5 v2 environments, 20k versions) showed 2.7s per request, 2.45s of it in two queries:

- The `last_modified_in_any_environment` and `last_modified_in_current_environment` annotations were `Max()` aggregations joining `features × feature_states × environment_feature_version` across every environment and every historical version, forcing a `GROUP BY` over the whole join. The joined row count grows with total version history, so the endpoint degrades forever. The resulting statement was also large enough (planner cost ~3.2M, ~30 subplans) to trip Postgres JIT compilation, which accounted for 1.15s of a 1.2s execution on its own — actual row work was 45ms.
- The paginator's `COUNT` re-ran the same aggregated query.
- `get_overrides_data()` fetched every latest feature state, including all segment and identity overrides, for the whole environment into Python — while the serialiser only reads the current page.

This PR decouples the request cost from project size and version history:

- The two `last_modified` annotations are now correlated subqueries against `EnvironmentFeatureVersion` directly (`ORDER BY created_at DESC LIMIT 1`). We get rid of joins and `GROUP BY`; Postgres evaluates them only for rows that survive pagination, and Django strips them from the `COUNT` query, which is now a plain `SELECT COUNT(*) ... WHERE project_id = %s`.
- Since the tag and owners filters relied on the aggregation's `GROUP BY` for join dedup, they are rewritten as `Exists` subqueries.
- `get_overrides_data()` and its core/edge variants take a `feature_ids` argument and only fetch rows for the requested page. The view passes the paginated feature IDs and reuses the `Environment` instance it already fetched, saving a query.
- A new migration adds two partial indexes on `EnvironmentFeatureVersion` — `(feature, -created_at)` and `(environment, feature, -created_at)` where published and not deleted — so the latest-published-version lookups stay `O(log n)` as history grows. Indexes are created via `PostgresOnlyRunSQL` (partial indexes aren't supported on Oracle) with `CREATE INDEX CONCURRENTLY` to avoid locking the table during deployment.

Results on the seeded project: 2.73s → 0.25s end to end; the sum of SQL time went from 2.49s to 0.035s. The remaining time is DRF serialisation.

Behavioural note: soft-deleted versions no longer count towards `last_modified_*` values (the old join ignored `deleted_at` on `EnvironmentFeatureVersion`; the subquery goes through the default soft-delete manager).

Follow-up candidates, out of scope here: the `get_latest_versions` raw SQL still aggregates over an environment's full version history per call (index-scan-backed, ~7ms locally), `value_search`/`is_enabled` filtering still materialises the whole environment's flags, and serialisation now dominates response time.

- Seeded a local project with 200 features, 5 v2-versioned environments, 20 published versions per feature per environment, plus segment and identity overrides; profiled the endpoint with captured queries and `EXPLAIN (ANALYZE, BUFFERS)` before and after.
- Verified tag `UNION`/`INTERSECTION` and owners filters return deduplicated results and correct counts against the seeded data.


api/tests/unit/features/test_unit_features_views.py
```diff
@@ -4335,7 +4336,7 @@ def test_list_features__last_modified_without_rbac__returns_expected(
         feature,
         with_project_permissions,
         django_assert_num_queries,
-        num_queries=20,
+        num_queries=19,
```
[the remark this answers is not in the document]
khvn26: Probably makes sense, even though I don't know what's even the legitimate use case for soft deleting versions.

api/features/views.py
```diff
@@ -186,14 +187,20 @@ def get_queryset(self):  # type: ignore[no-untyped-def]  # noqa: C901
 
         project = get_object_or_404(accessible_projects, pk=self.kwargs["project_pk"])
 
+        # Correlated subqueries (as opposed to joined aggregations) keep the
+        # main query free of joins and GROUP BY, so they are only evaluated
+        # for the rows on the requested page and are stripped from the
+        # paginator's COUNT query.
```
emyller:
```suggestion
```

Only nice to know while reading this pull request.

api/features/views.py
```diff
@@ -564,16 +580,28 @@ def filter_owners_and_group_owners(
         queryset: QuerySet[Feature],
         query_data: dict[str, typing.Any],
     ) -> QuerySet[Feature]:
+        # Use subqueries rather than joins so that features matching multiple
+        # owners or group owners don't produce duplicate rows.
```
emyller:
```suggestion
```

api/features/views.py
```diff
@@ -589,10 +617,17 @@ def _filter_queryset(
             queryset = queryset.filter(name__icontains=query_data["search"])
 
         if "tags" in query_serializer.initial_data:
+            # Use subqueries rather than joins so that features matching
+            # multiple tags don't produce duplicate rows.
```
emyller:
```suggestion
```

emyller (dismissed):
LGTM bar slop comments.

api/tests/unit/features/test_unit_features_views.py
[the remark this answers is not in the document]
khvn26: I think I'll hold this off until we actually understand what function deleting an EFV serves.

emyller (approved):

PR #8086, opened 2026-07-23
fix(code-help): make the React identity snippets runnable

talissoncosta:

## Changes

Fixes two snippet bugs CodeRabbit flagged on #8062 (both pre-existing there):

- Option 1 of the create-user and traits React snippets passed `flagsmith={flagsmith}` without importing the client, so the copied code didn't run. Import it the way Option 2 already does.
- The React init snippet's two demo output lines rendered concatenated (`my_cool_feature: truebanner_size: 42`); each gets its own line now.

Not addressed: interpolating flag names as JS identifiers breaks for dashed names (`flags.checkout-beta.enabled`). Systemic across templates and needs a call on whether dashed names are supported; left out deliberately.

Stacked on #8062; will retarget `main` once it merges.

- [ ] Features page → SDK snippet → React, "Create user" and "Traits" tabs: Option 1 imports `flagsmith` and passes it to the provider
- [ ] Copy Option 1, paste into an editor: no undefined `flagsmith`
- [ ] Init snippet: the two example output lines are separate

`lint` passes; snippet templates only, no component changes.

### DIFF
--- a/frontend/common/code-help/create-user/create-user-react.js
+++ b/frontend/common/code-help/create-user/create-user-react.js
@@ -6,6 +6,7 @@ export default (
   userId,
 ) => `
 // Option 1: Initialise with an identity
+import flagsmith from '${NPM_CLIENT}';
 import { FlagsmithProvider } from '${NPM_CLIENT}/react';
 
 export default function App() {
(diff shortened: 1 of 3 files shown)


talissoncosta:
Parking this: the snippet review grew into a much bigger accuracy sweep, so we're taking it to a team discussion on how to land it (issue to follow). The fixes are preserved and will come back per whatever split we agree.

PR #8096, opened 2026-07-24
perf(SDK): Decouple SDK identity request cost from version history

khvn26:

## Changes

The `GET`/`POST /api/v1/identities/` hot path had several costs, some of which grow with environment size and age:

1. In v1-versioning environments, flag evaluation fetched every committed version of every feature state. Version selection now happens in SQL with a `ROW_NUMBER()` window over `(feature, feature_segment, identity)` ordered by `(live_from, version)`, replicating `FeatureState.__gt__`. The Python dedup still runs on the reduced set to make sure the logic is unchanged. The `additional_filters` logic is preserved, and v2-versioning environments are unaffected.

2. Every `POST` with traits wrote unchanged traits. `update_traits` compared the stored scalar against the SDK-deserialised dict, which never matches. This is now fixed.

3. The wide `Environment` join happened multiple times. We're now making sure it's done once per request. This also benefits `GET /api/v1/flags`. There were also some unused joins: `identity` and `feature_segment__segment`, that are now removed.

4. No `.distinct()` on the environment segments queryset meant a more expensive dedup on the Python side. Implemented as a semi-join to keep Oracle compat.

Benchmarked with [this script](https://gist.github.com/khvn26/18ff3c56b5f2cdc0f3a51f67f16950ce) (100 features, 20 segments overriding 5 features each, 5 committed versions per override, 10 traits):

| | before | after |
|---|---|---|
| `GET` mean latency | 174 ms | 60 ms |
| `POST` mean latency | 179 ms | 57 ms |
| `POST` queries (writes) | 19 (1) | 10 (0) |

Real deployments should see a larger gain as latency climbed along with version counts.

- Added `test_update_traits__unchanged_sdk_trait_value_data__makes_single_query` pinning the no-write behaviour for SDK-shaped trait values.
- Updated three deprecated-endpoint query count assertions downwards (8→6, 8→6, 7→5).
- Benchmark linked above.

### DIFF
--- a/api/environments/identities/managers.py
+++ b/api/environments/identities/managers.py
@@ -2,6 +2,8 @@ from typing import TYPE_CHECKING
 
 from django.db.models import Manager
 
+from environments.identities.services import replace_identity_environment
+
 if TYPE_CHECKING:
     from typing import Iterable
 
@@ -28,7 +30,7 @@ class IdentityManager(Manager["Identity"]):
             identifier=identifier,
             environment=environment,
         )
-        identity.environment = environment
+        replace_identity_environment(identity, environment)
         return identity, created
 
     def with_traits(
--- a/api/environments/identities/managers.py
+++ b/api/environments/identities/managers.py
@@ -9,7 +9,6 @@ if TYPE_CHECKING:
 
     from environments.identities.models import Identity
     from environments.models import Environment
-    from integrations.integration import IntegrationConfig
 
 
 class IdentityManager(Manager["Identity"]):
@@ -25,28 +24,18 @@ class IdentityManager(Manager["Identity"]):
         identifier: str,
         environment: "Environment",
     ) -> "tuple[Identity, bool]":
-        return self.with_context().get_or_create(
+        identity, created = self.with_traits().get_or_create(
             identifier=identifier,
             environment=environment,
         )
+        identity.environment = environment
+        return identity, created
 
-    def with_context(
+    def with_traits(
         self,
-        integrations: "Iterable[IntegrationConfig] | None" = None,
-        extra_select_related: "Iterable[str] | None" = None,
         extra_prefetch_related: "Iterable[str | Prefetch] | None" = None,  # type: ignore[type-arg]
     ) -> "QuerySet[Identity]":
-        from integrations.integration import IDENTITY_INTEGRATIONS
-
-        return self.select_related(
-            "environment",
-            "environment__project",
-            *(
-                f"environment__{integration['relation_name']}"
-                for integration in (integrations or IDENTITY_INTEGRATIONS)
-            ),
-            *(extra_select_related or []),
-        ).prefetch_related(
+        return self.prefetch_related(
             "identity_traits",
             *(extra_prefetch_related or []),
         )
--- a/api/environments/models.py
+++ b/api/environments/models.py
@@ -535,10 +535,10 @@ class Environment(
         if not segments:
             segments = list(
                 Segment.live_objects.filter(
-                    feature_segments__feature_states__environment=self
-                )
-                .distinct()
-                .prefetch_related(
+                    id__in=FeatureSegment.objects.filter(
+                        feature_states__environment=self
+                    ).values("segment_id")
+                ).prefetch_related(
                     "rules",
                     "rules__conditions",
                     "rules__rules",
--- a/api/environments/models.py
+++ b/api/environments/models.py
@@ -536,7 +536,9 @@ class Environment(
             segments = list(
                 Segment.live_objects.filter(
                     feature_segments__feature_states__environment=self
-                ).prefetch_related(
+                )
+                .distinct()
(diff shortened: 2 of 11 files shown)


api/tests/unit/features/versioning/test_unit_versioning_versioning_service.py
```diff
@@ -136,6 +136,114 @@ def test_get_environment_flags_list__multiple_versions_and_identities__returns_l
...
+    )
+
+    # When
+    environment_feature_states = get_environment_flags_list(environment=environment)
+
+    # Then
+    assert set(environment_feature_states) == {
+        environment_default_feature_state,
+        segment_override_v2_feature_state,
+    }
```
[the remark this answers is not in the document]
khvn26: Sorry, it's moot. I only care about `get_environment_flags_list` returning the correct result; this test asserts that. If anything, the fact that it passes on both branches reaffirms us that the perf solution is correct.

api/features/versioning/versioning_service.py
```diff
@@ -584,26 +589,44 @@ def _get_feature_states_queryset(
...
+        version_rank=Window(
+            expression=RowNumber(),
+            partition_by=[
+                F("feature_id"),
+                Coalesce("feature_segment_id", Value(0)),
+                Coalesce("identity_id", Value(0)),
+            ],
+            order_by=[F("live_from").desc(), F("version").desc()],
+        ),
+    ).filter(version_rank=1)
```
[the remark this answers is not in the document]
khvn26: 9843055cc4a9d087c15d4298a356529e6ee3108e bar "v2-versioning behaviour remain unchanged" — this is evident from the diff, and green CI, alone.

api/environments/identities/managers.py
```diff
@@ -25,28 +24,18 @@ def get_or_create_for_sdk(
         identifier: str,
         environment: "Environment",
     ) -> "tuple[Identity, bool]":
-        return self.with_context().get_or_create(
+        identity, created = self.with_traits().get_or_create(
             identifier=identifier,
             environment=environment,
         )
+        identity.environment = environment
```
matthewelwell: I don't follow the need for this? Shouldn't the returned ORM model already have the environment set?

matthewelwell: Oh, I think I just answered my own question. It's probably because the environment included as an argument already has the prefetch/select_related caches in the object? 

If so, would it be better to move this to a mutation function so we can document it? 

Something like: 

```python
def replace_identity_environment(identity: "Identity", environment: "Environment") -> None:
    """
    Replace the environment attribute on the identity model.
    Used for optimisation in scenarios where the environment already has
    ORM caches (select|prefetch_related) on it, while the identity environment
    doesn't.  
    """
    identity.environment = environment
```

api/environments/models.py
```diff
@@ -536,7 +536,9 @@ def get_segments_from_cache(self) -> typing.List[Segment]:
             segments = list(
                 Segment.live_objects.filter(
                     feature_segments__feature_states__environment=self
-                ).prefetch_related(
+                )
+                .distinct()
```
matthewelwell: > No .distinct() on the environment segments queryset meant a more expensive dedup on the Python side. Added

Do we know how much load this adds to the DB though? I might suggest that this one gets pulled into it's own PR. `DISTINCT` can be a fickle mistress in my experience. I suspect it also might not play nicely with Oracle...

api/features/versioning/versioning_service.py
```diff
@@ -584,26 +589,44 @@ def _get_feature_states_queryset(
...
+    return queryset.select_related(
+        "feature",
+        "feature_state_value",
+        "environment_feature_version",
+        "feature_segment",
+        *additional_select_related_args,
+    ).prefetch_related(*additional_prefetch_related_args)
+
+
+def _exclude_superseded_versions(
```
matthewelwell: Possibly very annoying comment, but isn't this what `get_live_feature_states` is supposed to do... so I'd consider it a bug that it doesn't. Can we move this logic into that method, or does it break a load of other things?

matthewelwell (changes requested):
I've marked as 'Request Changes' which feels too strong, but I don't want to approve without at least a discussion about extracting the method, and 'Comment' is just too nothing-y.

api/features/versioning/versioning_service.py
khvn26: Looks like we have a bug on our hands. Created #8127 and referred in a64aa84.

api/environments/models.py
khvn26: Good spot — replaced with a semi-jon in 2223fc21c735c890e89df167503f19ad0e7a1a23. Perf improvement still stands.

api/environments/identities/managers.py
khvn26: Done in b7fd31d.

matthewelwell (approved):

PR #8099, opened 2026-07-24
feat(experiments): colour lift by metric direction

Zaimwa9:

## Changes

Depends on #8098.

Colours experiment lift values by the primary metric's direction instead of sign only (follow-up to #8091):

- Green when the lift moves with the metric ("higher is better" + positive, "lower is better" + negative), red when against, neutral grey for informational metrics.
- Applies to the analysis chart, the delta bar/value in the results table, and the "Lift vs control" scorecard (now driven by a `liftTone` of success/danger/neutral).
- Until #8098 is deployed, `direction` is absent from the API response and metrics fall back to "higher is better" — identical to current behaviour.

Parametrised unit tests for direction → tone mapping in `derive.test.ts`; manually checked chart/table/scorecard colours for up and down metrics.

### DIFF
--- a/frontend/web/components/experiments/results/ExperimentSummaryScorecard.tsx
+++ b/frontend/web/components/experiments/results/ExperimentSummaryScorecard.tsx
@@ -21,7 +21,7 @@ const ExperimentSummaryScorecard: FC<ExperimentSummaryScorecardProps> = ({
   )
   const hasResults = !!results
 
-  let liftClassName: string | undefined
+  let liftClassName = 'text-secondary' // neutral
   if (summary?.liftTone === 'success') liftClassName = 'text-success'
   if (summary?.liftTone === 'danger') liftClassName = 'text-danger'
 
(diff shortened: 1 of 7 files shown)


frontend/web/components/experiments/results/ExperimentSummaryScorecard.tsx
```diff
@@ -22,9 +22,8 @@ const ExperimentSummaryScorecard: FC<ExperimentSummaryScorecardProps> = ({
   const hasResults = !!results
 
   let liftClassName: string | undefined
```
talissoncosta: Would be better to default it to neutral like:

```tsx
let liftClassName = 'text-secondary' // neutral
```

talissoncosta (changes requested):
just one minor comment. Do you mind checking it before move forward?

talissoncosta (approved):

PR #8102, opened 2026-07-25
feat(__future__): Experimental update-flag

### LINKED ISSUE #8088: update-flag endpoint needs a stable API
We experimented with Options A and B (issue1), but implementing multi-variate support to both options (issue2, superseded by this) proved quite difficult in both work and reviewing.

Since we seemed to have chosen Option B as the winner, we'll pick it up and re-shape it into a new Option C from scratch

## Acceptance criteria

- [x] A new `/api/__future__/environments/{key}/features/{id}/` supports PATCH and PUT methods for updating a flag's environment default and segment overrides in one request.
- [x] Multi-variate support is present.
- [x] Code lives in a `future` namespace, isolated so we can easily delete or canonise it down the road.

emyller:

## Changes

Contributes to https://github.com/Flagsmith/flagsmith/issues/7642
Closes https://github.com/Flagsmith/flagsmith/issues/8088

Stabilises experimental `update-flag` endpoints to one option, still experimental. It supports the verbs `PATCH` and `PUT`.

- Avoids touching Options A + B — cleanup + interface migration issue will follow.
- Includes multivariate flag support.
- Improves API and module design.

Updated documentation: https://docs-git-feat-update-flag-option-c-flagsmith.vercel.app/managing-flags/updating-flags

Integration tests included.


docs/docs/managing-flags/updating-flags.md
```diff
@@ -0,0 +1,211 @@
...
+    ]
+  }'
+```
+
+The `priority` field in segment overrides is optional. The lowest number has the highest priority. Omit it to add at the
+lowest priority.
+
+### Configure multiple segment overrides
+
+Set different values per segment — for example, pricing tiers.
```
khvn26: Can I omit priorities in this case?

docs/docs/managing-flags/updating-flags.md
```diff
@@ -0,0 +1,211 @@
...
+```bash
+curl -X POST 'https://api.flagsmith.com/api/experiments/environments/{environment_key}/update-flag/' \
+  -H 'Authorization: Api-Key <your_token>' \
+  -H 'Content-Type: application/json' \
+  -d '{
+    "feature": {"name": "new_checkout"},
+    "segment_overrides": [
+      {
+        "segment_id": 456,
+        "delete": true
```
khvn26: Not loving the API here. I'd prefer to have full replace PUT semantics instead — demand the full overrides list, delete omitted elements.

IMO, individual override operations work well with individual APIs, e.g.

- `DELETE /api/experiments/environments/{environment_key}/update-flag/{feature_id}/segment/{segment_id}`
- `PUT /api/experiments/environments/{environment_key}/update-flag/{feature_id}/identity/{identifier}`

This option especially makes sense in the CLI V2 context.

khvn26 (commented):
Looks good overall. My main gripe is combining full-replace and `delete: true` semantics for lists of different entities in a single API.

docs/docs/managing-flags/updating-flags.md
```diff
@@ -0,0 +1,211 @@
...
+
+The `priority` field in segment overrides is optional. The lowest number has the highest priority. Omit it to add at the
+lowest priority.
+
+### Configure multiple segment overrides
+
+Set different values per segment — for example, pricing tiers.
+
+```bash
+curl -X POST 'https://api.flagsmith.com/api/experiments/environments/{environment_key}/update-flag/' \
```
khvn26: nit: Since we're shipping experiments, I think it's best to avoid confusion:

```suggestion
curl -X POST 'https://api.flagsmith.com/api/experimental/environments/{environment_key}/update-flag/' \
```

But maybe we should bite the bullet already:

```suggestion
curl -X POST 'https://api.flagsmith.com/api/v2/environments/{environment_key}/update-flag/' \
```

docs/docs/managing-flags/updating-flags.md
```diff
@@ -0,0 +1,211 @@
...
+Set up features with weighted variants and customise weights per segment.
+
+The `variants` list in `environment_default` defines the available variants for the feature, and their default weights
+in the environment. Each variant is identified by a `key` (slug format). Omitting a variant from the list deletes it in
+the environment, and in any segment overrides.
+
+A `weight` is a fraction between 0 and 1. Any weight not allocated to variants serves the flag's default `value`.
+
+The `variants` list in `segment_overrides` can only re-weight existing variants. Variants omitted from it keep their
+current weights for that segment.
```
khvn26: > The `variants` list in `segment_overrides` can only re-weight existing variants.

👍 

> Variants omitted from it keep their current weights for that segment.

I am not exactly sure how that's supposed to work. For example, if we send 1 variant with 100% weight, the backend ought to re-weight all other variants to 0?

docs/docs/managing-flags/updating-flags.md
matthewelwell: We don't need `v2` though because we're not changing an endpoint - this is a new endpoint. It's under the `experiments` namespace because it's subject to change, not because it's any sort of breaking change. If we go with `v2` and need to change it, we'll have to go to `v3` ... 

I do agree with the comment on `experiments` vs `experimental` though. It might be a PITA to change _that_ now though.

docs/docs/managing-flags/updating-flags.md
matthewelwell: I'd say that we should infer priorities from the ordering in the payload if not provided.

docs/docs/managing-flags/updating-flags.md
```diff
@@ -0,0 +1,211 @@
...
+      }
+    ]
+  }'
+```
+
+### Configure A/B/n experiments (variants)
+
+Set up features with weighted variants and customise weights per segment.
+
+The `variants` list in `environment_default` defines the available variants for the feature, and their default weights
```
matthewelwell: I'm not sure I agree that modifying the `variants` themselves should be part of this endpoint at all. `variants` (or `MultivariateFeatureOption`s) are project level concepts. This endpoint should be concerned only with environment level entities. 

IMO we should raise an error if a variant `key` is provided that doesn't already exist.

docs/docs/managing-flags/updating-flags.md
matthewelwell: Should we instead enforce that the client has to provide all variants?

docs/docs/managing-flags/updating-flags.md
matthewelwell: > I'd prefer to have full replace PUT semantics instead — demand the full overrides list, delete omitted elements.

This sounds like a can of worms that leads back to needing both option_a and _b. I do agree that I don't like the idea of adding the new `"delete": true` attribute to the payload, however. I prefer a separate call for this. We had (I think) a `/remove-segment-override` endpoint previously alongside the `/update-flag` one, right? I tend to think that Kim's suggestion is a good one though of having `DELETE /api/experiments/environments/{environment_key}/update-flag/{feature_id}/segment/{segment_id}` .

docs/docs/managing-flags/updating-flags.md
khvn26: I actually thought this was implied by

> The variants list in segment_overrides can only re-weight existing variants.

matthewelwell: That's not how I understood it - I understood that there was a difference in behaviour between the environment_default section of the payload, and the segment_overrides on this matter. TBC by @emyller .

docs/docs/managing-flags/updating-flags.md
emyller: > IMO, individual override operations work well with individual APIs, e.g.
> 
> * `DELETE /api/experiments/environments/{environment_key}/update-flag/{feature_id}/segment/{segment_id}`
> * `PUT /api/experiments/environments/{environment_key}/update-flag/{feature_id}/identity/{identifier}`

Love it. For some reason, I thought we were constrained to one swiss-army-knifey `/update-flag/` endpoint only.

docs/docs/managing-flags/updating-flags.md
emyller: Which priority numbers could represent _ordering in the payload_? Are we thinking of something close to `INF - n`, or a regular 1-based enumeration?

khvn26: Is there any difference?

emyller: My initial instinct was to steer away from user input space, but I guess it mostly doesn't matter since we reorder in the UI using drag and drop.

docs/docs/managing-flags/updating-flags.md
emyller: My honest understanding of adding MVs support to `update-flag` was that it could be used to _configure a multivariate experiment_. If it's not, I think I need to learn more about the use case.

docs/docs/managing-flags/updating-flags.md
emyller: Should be improved in 98593e00abd7caaae4e7852cd527c1b3a9abc468.

docs/docs/managing-flags/updating-flags.md
emyller: > demand the full overrides list, delete omitted elements.

I think this would break the intended ergonomics of this endpoint — _change only these things about this feature, leave out all the rest_. I think this intent is only inheritance of Option B's nature, which we chose.

On a second thought, the suggested `/update-flag/{feature_id}/` path also breaks ergonomics since it relies on the feature ID, while this endpoint assumes the user may choose to identify features given their name.

So, should we:

1. Support either a feature name or ID in the URL?
2. Choose to support only `feature_name`?
3. Restore Gagan's separate `POST .../delete-segment-override/` but with Option B's style?

docs/docs/managing-flags/updating-flags.md
emyller: Arguably, this is the first time this `experiments/` namespace is used, and we're updating the URLs anyway, so it should be an easy change.

```
/api/__future__/environments/{environment_key}/update-flag/{feature_name}
```

I've borrowed Python's dunder-future name in the pull request title. Not sure if a good disambiguation suggestion for an API path, but it is definitely one.

docs/docs/managing-flags/updating-flags.md
emyller: Improved in e11139461.

docs/docs/managing-flags/updating-flags.md
emyller: Leaning towards knowledge I'm yet to learn: e11139461

docs/docs/managing-flags/updating-flags.md
khvn26: > change only these things about this feature, leave out all the rest

My reasoning is: we can't cleanly express this logic for all input lists (overrides, variants); supporting `delete: true` for all lists doesn't sound like a good idea either; full-replace logic is the only one we can support consistently for both variants and overrides. 

> On a second thought, the suggested `/update-flag/{feature_id}/` path also breaks ergonomics since it relies on the feature ID

`{feature_name}`, or using either name or id, sounds good to me.

docs/docs/managing-flags/updating-flags.md
khvn26: Works for me.

docs/docs/managing-flags/updating-flags.md
khvn26: Current experiments UI does not allow to add, delete or modify variants — unless I'm missing something, it consumes existing ones.

docs/docs/managing-flags/updating-flags.md
emyller: cfe48faacb858e5b3872fa2df949ac3c1b890b00 — wild, let's go!

emyller:
This should be ready for a new review: since 81077d9eab8eb471396fc038d2d1ee7b4bb55254, I've updated the proposed documentation to make it consistent to our [offline] discussion. Will update tests and move right on to implementing it once this looks good.

I've opted to let go of a separate `DELETE` endpoint, as it seemingly has no use in the UI, which is satisfied with a `PUT` path. Overall I think all the `PATCH` path isn't much useful to our dashboard, but it definitely helps ergonomics _if we need it_, e.g. in the CLI.

api/tests/integration/features/experiments/test_update_flag_endpoint.py
```diff
@@ -0,0 +1,640 @@
...
+    assert override.get_feature_state_value() == "enterprise"
+
+
+@pytest.mark.parametrize(
+    "variants, expected_allocations",
+    [
+        pytest.param(
+            [
+                {"key": "variant_a", "weight": 0.25},
+                {"key": "variant_b", "weight": 0.25},
```
khvn26: nit: as example values these weights make no sense, unless you want the API to accept 0-1 floats (as opposed to current 0-100).

emyller: _You're absolutely right!_ Seriously though, I've intentionally not updated tests since my last changes to docs. Please refer to my note above:

> Will update tests and move right on to implementing it once this looks good.

docs/docs/managing-flags/updating-flags.md
emyller: The spec was updated as per [this comment](https://github.com/Flagsmith/flagsmith/pull/8102#issuecomment-5198229214).

docs/docs/managing-flags/updating-flags.md
```diff
@@ -0,0 +1,177 @@
...
+
+| Type      | Example                                |
+| --------- | -------------------------------------- |
+| `string`  | `{"type": "string", "value": "hello"}` |
+| `integer` | `{"type": "integer", "value": "42"}`   |
+| `boolean` | `{"type": "boolean", "value": "true"}` |
+
+### Toggle a flag on or off
+
+The simplest case — flip a feature flag in an environment:
```
matthewelwell: nit

```suggestion
The simplest case — enable a feature flag in an environment:
```

(because 'flip' might imply that sending the same request multiple times would toggle it on,off,on,off,etc)

docs/docs/managing-flags/updating-flags.md
```diff
@@ -0,0 +1,177 @@
...
+          {"key": "variant_a", "weight": 25},
+          {"key": "variant_b", "weight": 25}
+        ]
+      }
+    ]
+  }'
+```
+
+In both `environment_default` and `segment_overrides`, the `variants` list **must** include all variants for the
+feature, even if their weight is zero.
```
matthewelwell:
```suggestion
In both `environment_default` and `segment_overrides`, the `variants` list **must** include all variants for the
feature, even if their weight is zero, and regardless of the HTTP verb used (`PUT` or `PATCH`).
```

Not sure if this is correct, but it should probably be added if it is.

khvn26:
I guess my only question at this point is whether we stick with `feature_id/feature_name` addressing on the API side, or only accept feature ids — this is not apparent from the docs currently.

emyller:
> I guess my only question at this point is whether we stick with `feature_id/feature_name` addressing on the API side, or only accept feature ids — this is not apparent from the docs currently.

I decided not to support feature names in this endpoint, for now, because we know no consumer that would benefit from it yet. The proposed docs assumes this discussion never existed, and includes `{feature_id}` in the example URLs. Let me know if you think anything should be improved towards that.

api/features/future/services.py
```diff
@@ -0,0 +1,315 @@
...
+from api_keys.user import APIKeyUser
+from environments.models import Environment
+from features.models import Feature, FeatureSegment, FeatureState, FeatureStateValue
+from features.multivariate.models import MultivariateFeatureStateValue
+from features.versioning.models import EnvironmentFeatureVersion
+from features.versioning.versioning_service import get_environment_flags_list
+from users.models import FFAdminUser
+
+from .mappers import map_environment_default, map_segment_override, map_variants
+from .types import (
```
khvn26: nit: can we avoid local imports please

api/features/future/mappers.py
```diff
@@ -0,0 +1,64 @@
+"""https://docs.flagsmith.com/managing-flags/updating-flags"""
+
+from features.feature_states.models import API_VALUE_TYPES
+from features.models import FeatureState, FeatureStateValue
+
+from .types import (
```
khvn26: nit: avoid local imports please

api/features/future/serializers.py
```diff
@@ -0,0 +1,113 @@
+"""https://docs.flagsmith.com/managing-flags/updating-flags"""
+
+from rest_framework import serializers
+
+from features.feature_types import MULTIVARIATE
+from features.models import Feature, FeatureStateValue
+from segments.models import Segment
+
+from .types import (
```
khvn26: nit: avoid local imports please

api/features/future/services.py
emyller: 14c3324fe.

api/features/future/mappers.py
emyller: 14c3324fe.

api/features/future/serializers.py
emyller: 14c3324fe.

docs/docs/managing-flags/updating-flags.md
emyller: This seems resolved. Please reopen otherwise.

api/features/future/serializers.py
```diff
@@ -0,0 +1,113 @@
...
+
+    enabled = serializers.BooleanField(required=False)
+    value = FlagValueSerializer(required=False)
+    variants = VariantSerializer(many=True, required=False)
+
+    def validate_value(self, value: FlagValue) -> FlagValue:
+        try:
+            FeatureStateValue().set_value(value["value"], value["type"])
+        except ValueError as exc:
+            raise serializers.ValidationError(str(exc)) from exc
```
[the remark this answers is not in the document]
emyller: This was coded consciously of the errors produced downstream, yet this is a valid concern. Attempted improvement: a9b59ec51d3cf2dd577850021a3a0592c4cd12da.

api/features/future/services.py
```diff
@@ -0,0 +1,318 @@
...
+        segment_id = change["segment"]["id"]
+        if feature_state := overrides.get(segment_id):
+            written.updated.append(segment_id)
+        else:
+            feature_state = _create_segment_override(
+                environment=environment,
+                feature=feature,
+                version=version,
+                segment_id=segment_id,
+                priority=change.get("priority", position),
```
khvn26, line 252: I think we should avoid inserting duplicate priorities.

For instance, adding two overrides without a priority will currently result in duplicate `0` priorities, since each is position 0 of its own list:

```
// three overrides already at 10, 20, 30
PATCH {"segment_overrides": [{"segment": {"id": 1}, "enabled": true}]}
→ [(1, 0), (3, 10), (4, 20), (5, 30)]

PATCH {"segment_overrides": [{"segment": {"id": 2}, "enabled": true}]}
→ [(1, 0), (2, 0), (3, 10), (4, 20), (5, 30)]
```

I think we should attempt to infer from position, and fail hard if it results in duplicate priorities.

I don't think we should plainly validate `incoming_priority == any_existing_priority` either, rather make sure that the _resulting_ priority list doesn't have any duplicates. This way `flagsmith flag reorder` should still work.

api/features/future/serializers.py
```diff
@@ -0,0 +1,113 @@
...
+                {"variants": ["Must include all feature's variants."]}
+            )
+        return attrs
+
+
+class SegmentOverrideSerializer(FlagStateSerializer):
+    """What the flag serves to the identities a segment matches."""
+
+    segment = SegmentReferenceSerializer()
+    priority = serializers.IntegerField(min_value=0, required=False)
```
khvn26, line 98: The docs say:

> In both `environment_default` and `segment_overrides`, the `variants` list **must** include all variants for the
feature, even if their weight is zero, and regardless of the HTTP verb used (`PUT` or `PATCH`).

...which is currently not true: 

```
PUT {"environment_default": {"enabled": true}}                       // 400 with "Must include all feature's variants."
PUT {"segment_overrides": [{"segment": {"id": 1}, "enabled": true}]} // 200, and the variants silently overwritten with environment defaults.
```

I think `SegmentOverrideSerializer` should reuse `EnvironmentDefaultSerializer`'s validation.

api/features/future/services.py
```diff
@@ -0,0 +1,318 @@
...
+            )
+            written.created.append(segment_id)
+        _write_segment_override(
+            feature_state,
+            change,
+            replace=replace or segment_id in written.created,
```
khvn26, line 258: Consider this example:

```
// environment default: enabled=true, value="default-on"
PATCH {"segment_overrides": [{"segment": {"id": 2}, "value": {...}}]}
// override: enabled=false, value from the request
```

Can we either inherit the environment default state as well as the value, or document the "implicit off" behaviour?

api/api/urls/future.py
```diff
@@ -0,0 +1,20 @@
...
+from django.urls import path
+
+from features.future.views import UpdateFlagAPIView
+
+app_name = "future"
+
+urlpatterns = [
+    path(
+        "environments/<str:environment_key>/features/<int:feature_id>/",
+        UpdateFlagAPIView.as_view(),
```
khvn26: Question: Could this path take a `GET`? I think `get_flag()` already returns what a client needs, and every write already calls it.

It's a little awkward to display mutation results without it in https://github.com/Flagsmith/flagsmith-cli/pull/92.

docs/docs/managing-flags/updating-flags.md
```diff
@@ -0,0 +1,318 @@
...
+```
+
+Overrides listed in a `PATCH` payload are added or updated by segment; overrides not listed are left unchanged. When
+adding a new segment override, if `priority` is omitted, it defaults to the override's position in the
+`segment_overrides` list. The lowest number has the highest priority.
+
+A new segment override serves whatever the environment default serves, until you give it a `value` of its own. An
+existing override keeps its priority unless you send a new one.
+
+### Remove a segment override
```
khvn26, line 209: Note: While implementing https://github.com/Flagsmith/flagsmith-cli/pull/92 my agent really wished that `DELETE .../features/{feature_id}/segment-overrides/{segment_id}/` existed. It didn't like the wholesale PUT semantics at all, and rightfully so since race risks are greater.

If the dashboard integration trips on this too, perhaps we should consider following up.

api/features/future/services.py
```diff
@@ -0,0 +1,318 @@
...
+    feature: Feature,
+    changes: UpdateFlagRequest,
+    replace: bool,
+    author: FFAdminUser | APIKeyUser,
+) -> UpdateFlagResponse:
+    """Write the given parts of a flag, whichever versioning the environment uses."""
+    written = WrittenSegmentOverrides([], [], [])
+
+    with transaction.atomic():
+        version = _create_draft_version(environment, feature)
```
khvn26, line 283: An empty body still creates and publishes a version:

```
PATCH {} // 200, versions for the feature go from 7 to 8
```

Can we avoid this?

api/features/future/exceptions.py
```diff
@@ -0,0 +1,13 @@
+"""https://docs.flagsmith.com/managing-flags/updating-flags"""
+
+from rest_framework import status
+from rest_framework.exceptions import APIException
+
+
+class ChangeRequestsEnabledError(APIException):
+    """Raised where a flag can only be changed by going through a change request."""
+
+    status_code = status.HTTP_400_BAD_REQUEST
```
khvn26: Can we add something machine-distinguishable? I.e. a `code` in the body, or a 409.

api/features/future/serializers.py
```diff
@@ -0,0 +1,113 @@
...
+)
+from features.models import Feature, FeatureStateValue
+from segments.models import Segment
+
+
+class FlagValueSerializer(serializers.Serializer[FlagValue]):
+    """A flag value, typed by the caller so it survives the round trip as a string."""
+
+    type = serializers.ChoiceField(choices=["string", "integer", "boolean"])
+    value = serializers.CharField(allow_blank=True)
```
khvn26, line 25: So, no `null`s. Just want to make sure it's a conscious decision here.

api/features/future/services.py
```diff
@@ -0,0 +1,318 @@
...
+
+def _write_variants(feature_state: FeatureState, variants: Sequence[Variant]) -> None:
+    weighted = {
+        multivariate_value.multivariate_feature_option_id: multivariate_value
+        for multivariate_value in feature_state.multivariate_feature_state_values.all()
+    }
+    for variant in variants:
+        if multivariate_value := weighted.get(variant["id"]):
+            multivariate_value.percentage_allocation = variant["weight"]
+            multivariate_value.save(update_fields=["percentage_allocation"])
```
khvn26, line 81: nit: Can we avoid N+1 by switching to `bulk_update`?

api/features/future/services.py
```diff
@@ -0,0 +1,315 @@
...
+                segment_id=segment_id,
+                priority=change.get("priority", position),
+            )
+            written.created.append(segment_id)
+        _write_segment_override(
+            feature_state,
+            change,
+            replace=replace or segment_id in written.created,
+            environment_default=environment_default,
+        )
```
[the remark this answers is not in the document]
emyller: Other than keeping whatever `OrderedModelBase` does, and our custom index-if-omitted behaviour, I'm not interested in pursuing any other logic. Current code works as expected and is properly tested.

docs/docs/managing-flags/feature-versioning.md
```diff
@@ -27,7 +27,7 @@ Enabling Feature Versioning v2 on an environment is irreversible.
 
 To produce a new published version on a v2 environment, use one of:
 
-- **The experimental [update-flag endpoints](/integrating-with-flagsmith/flagsmith-api-overview/admin-api/updating-flags)** (`update-flag-v1`, `update-flag-v2`, `delete-segment-override`). These accept the same payloads as on v1 environments and publish a new version per call on v2 environments.
+- **The experimental [update-flag endpoints](/managing-flags/updating-flags)** (`update-flag-v1`, `update-flag-v2`, `delete-segment-override`). These accept the same payloads as on v1 environments and publish a new version per call on v2 environments.
```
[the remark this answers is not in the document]
emyller: Fixed: 7cb40d412

api/features/future/serializers.py
```diff
@@ -0,0 +1,113 @@
...
+
+    enabled = serializers.BooleanField(required=False)
+    value = FlagValueSerializer(required=False)
+    variants = VariantSerializer(many=True, required=False)
+
+    def validate_value(self, value: FlagValue) -> FlagValue:
+        try:
+            FeatureStateValue().set_value(value["value"], value["type"])
+        except FeatureValueError as error:
+            raise serializers.ValidationError(str(error)) from error
```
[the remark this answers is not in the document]
emyller: This is conscious.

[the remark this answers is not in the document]
emyller: Again??

api/features/future/permissions.py
```diff
@@ -0,0 +1,35 @@
...
+
+PROPERTY_PERMISSIONS = {
+    "environment_default": UPDATE_FEATURE_STATE,
+    "segment_overrides": MANAGE_SEGMENT_OVERRIDES,
+}
+
+
+def check_update_permissions(
+    user: UserABC, environment: Environment, properties: Container[str]
+) -> None:
```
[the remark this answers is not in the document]
emyller: Improved in 89b7b85bfa84f185622f10020f75f7ebcee86b3b.

api/features/future/views.py
```diff
@@ -0,0 +1,98 @@
...
+        *,
+        replace: bool,
+    ) -> Response:
+        assert not isinstance(request.user, AnonymousUser)
+
+        try:
+            environment = Environment.objects.get(api_key=environment_key)
+        except Environment.DoesNotExist:
+            raise NotFound() from None
+        check_update_permissions(request.user, environment, request.data)
```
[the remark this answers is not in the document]
emyller: Improved in 89b7b85bfa84f185622f10020f75f7ebcee86b3b.

api/tests/integration/conftest.py
```diff
@@ -218,36 +219,41 @@ def server_side_sdk_client(
...
+) -> int:
+    response = admin_client.post(
+        f"/api/v1/projects/{project}/features/",
+        data={
+            "name": feature_name,
+            "initial_value": default_feature_value,
+            "project": project,
+        },
+    )
+    return int(response.json()["id"])
```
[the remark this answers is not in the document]
emyller: Silly but why not. ee2c22fca71d4583177db927b4d49de4e503bb78

docs/docs/integrating-with-flagsmith/flagsmith-api-overview/admin-api/updating-flags.md
khvn26, line 1: Please update `vercel.json` with a redirect.

khvn26 (commented):
Some comments to resolve before I approve.

api/features/future/serializers.py
```diff
@@ -0,0 +1,113 @@
...
+    SegmentOverrideRequest,
+    UpdateFlagRequest,
+    Variant,
+)
+
+
+class FlagValueSerializer(serializers.Serializer[FlagValue]):
+    """A flag value, typed by the caller so it survives the round trip as a string."""
+
+    type = serializers.ChoiceField(choices=["string", "integer", "boolean"])
```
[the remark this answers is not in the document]
emyller: Good catch, e49f75720c96be6edc6fddcdd0baf512daf60121. 💅

docs/docs/integrating-with-flagsmith/flagsmith-api-overview/admin-api/updating-flags.md
emyller: I chose not to do it because it's an experimental page. You think we need the redirect regardless?

khvn26: Documentation is still documentation. We expect the URLs to it to always work for everyone.

api/features/future/serializers.py
emyller: Correct. `FlagValueSerializer.value` cannot be null. However, `FlagStateSerializer.value` can.

After checking out the current behaviour in the UI, I think allowing a null FSV is better than creating an empty string — at least from the backend perspective.

api/tests/integration/features/future/test_flag_endpoint.py
```diff
@@ -0,0 +1,1727 @@
...
+        {
+            "level": "warning",
+            "event": "flag.update_rejected",
+            "organisation__id": versioned_environment.project.organisation_id,
+            "project__id": versioned_environment.project_id,
+            "environment__id": versioned_environment.id,
+            "feature__id": feature,
+            "reason": "change_requests_enabled",
+        },
+    ]
```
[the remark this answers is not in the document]
emyller: Can you see 9ebb0c341c005776ae534fb529a7a8a3d612a71f?

docs/docs/managing-flags/updating-flags.md
```diff
@@ -0,0 +1,312 @@
...
+
+- They cannot be used when [change requests](/administration-and-security/governance-and-compliance/change-requests) are
+  enabled.
+- They do not support identity overrides.
+
+These may be lifted in the future.
+
+:::
+
+Learn more in the [API specification](link TODO).
```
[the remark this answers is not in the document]
emyller: Good catch, fixed: 36b9208770464e763f61a6c61570da437d332c8a

docs/docs/managing-flags/updating-flags.md
```diff
@@ -0,0 +1,318 @@
...
+    }
+  ]
+}
+```
+
+In both `environment_default` and `segment_overrides`, the `variants` list **must** include all variants for the
+feature, even if their weight is zero, and regardless of the HTTP verb used (`PUT` or `PATCH`).
+
+Because `PUT` replaces `environment_default` in full, it **must** carry `variants` for a multivariate feature. A segment
+override that omits `variants` inherits the weights of the environment default instead.
```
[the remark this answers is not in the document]
emyller: Should be improved in ce77df32f9da54351f7a2326fda9e1301ebf73ba.

api/features/future/serializers.py
emyller: Similar to this [other thread](https://github.com/Flagsmith/flagsmith/pull/8102#discussion_r3785981794). Should be improved in ce77df32f9da54351f7a2326fda9e1301ebf73ba, but the design stays for better ergonomics IMO. Let me know if you'd disagree.

api/tests/integration/features/future/test_flag_endpoint.py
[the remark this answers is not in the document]
emyller: Excellent! Now please read the contents of that commit, and see it helps addressing your concern. Thank you in advance for being a proactive and smart agent.

api/features/future/services.py
emyller: I kind of [intentionally](https://github.com/Flagsmith/flagsmith/pull/8102#discussion_r3786080798) allowed duplicate priorities to make sure we wouldn't engage in priority math, but I think I like a duplicate validation better now: 2f070b6ed14f17cf6187789e9604d61d95af9600.

api/features/future/services.py
emyller: Great catch, thank you. The correct behaviour should be as of 7da1e6bf23f957dbd03207539f6bfdf70e0b3198.

api/api/urls/future.py
emyller: Heck yeah why not: 1304243efd10f00995876f52a12335dc26251ddd

docs/docs/managing-flags/updating-flags.md
emyller: I decided not to add a `DELETE` endpoint solely to squeeze all I could out of scope, since the dashboard will already make a `PUT` request. I'm not against the idea though — would be happy to see an issue.

api/features/future/services.py
emyller: Another great catch. Fix here: 9d640b545803f543b7d5a4db27430eb97794fc4d.

docs/docs/managing-flags/updating-flags.md
khvn26: https://github.com/Flagsmith/flagsmith/issues/8300

api/features/future/exceptions.py
emyller: b2efd4efafdacad377c49a988361abdb5ff2d61e — hope we can actually support CRs later.

docs/docs/integrating-with-flagsmith/flagsmith-api-overview/admin-api/updating-flags.md
emyller: Agreed: 27dcc94a2536a836e444e6916fa1b79c4796813b.

api/features/future/services.py
emyller: I think we need this to keep audit logging.

api/features/future/services.py
khvn26: Yeah I think this, ultimately, makes sense even for use cases outside the dashboard.

One small nit left IMO: I'd make the error message actionable so users understand why we can't accept ambiguous priorities. Something among the lines of `The overrides for seg1 and seg2 are in conflict; provide explicit priority values.`

khvn26: Happy to bring that nit as a follow-up; resolving.

khvn26 (approved):
Great work 🙌

PR #8104, opened 2026-07-26
fix: send falsy variant values to Mixpanel and Rudderstack integrations

### LINKED ISSUE #7701: Mixpanel integration: falsy variant values collapse into enabled-state

- [ ] Self Hosted with Docker
- [x] SaaS at flagsmith.com

### Describe the bug

When a flag is enabled and the variant value is falsy, the mixpanel integration incorrectly fallbacks the value to the flag state `true` instead of `false`.

The documentation is ambiguous, only stating `if the flag has no remote config value, pass the boolean state`
```
# Current mapping

  Flag state | Variant value              | Sent to Mixpanel
  -----------+----------------------------+----------------------
  ON         | True                       | true (value)
  ON         | False / None / 0 / ""      | true (enabled state)
  ON         | "hello" / 42               | the value
  OFF        | anything                   | false (enabled state)
```

### Steps To Reproduce

1. Activate Mixpanel integration in your environment
2. Create a multivariate feature
3. Enable the flag
4. Set a variant value to be `false` 
5. Mixpanel will show the variant as `true`

### Expected behavior

The mapping should be:
```
  Flag state | Variant value              | Sent to Mixpanel
  -----------+----------------------------+----------------------
  ON         | True                       | true (value)
  ON         | False / "" / 0             | false
  ON         | None / null                | true (enabled state)
  ON         | "hello" / 42               | the value
  OFF        | anything                   | false (enabled state)
```

The resolution of the issue includes:
- Updating the doc to be exhaustive in relation to the behavior
- Communicate with impacted customers before releasing

### Screenshots

_No response_

genrichez:
### Describe the pull request
This PR fixes a bug in the Mixpanel and Rudderstack integrations where falsy variant values (like `False`, `0`, or `""`) were being incorrectly dropped and replaced with the feature's `enabled` state (usually `True`).

**Root cause:**
The integrations used the ternary expression:
value if (feature_state.enabled and value) else feature_state.enabled
Because `False`, `0`, and `""` are falsy in Python, the `and value` condition failed, causing the code to fall back to `feature_state.enabled`.

**Fix:**
Updated the logic to check for `value is not None`, matching the correct pattern already used in the Amplitude and Heap integrations:
value if (feature_state.enabled and value is not None) else feature_state.enabled

I also updated the existing unit tests and added new parametrized tests (matching the Heap integration test pattern) to explicitly verify that falsy values (`False`, `0`, `""`) are sent correctly to both Mixpanel and Rudderstack.

### Link to the issue
Closes #7701

### DIFF
--- a/api/integrations/mixpanel/mixpanel.py
+++ b/api/integrations/mixpanel/mixpanel.py
@@ -46,6 +46,9 @@ class MixpanelWrapper(AbstractBaseIdentityIntegrationWrapper[MixpanelUserData]):
 
         for feature_state in feature_states:
             value = feature_state.get_feature_state_value(identity=identity)
+            # enabled + value set (incl. falsy like 0 or "") -> value
+            # enabled + value is None -> True
+            # disabled -> False
             feature_properties[feature_state.feature.name] = (
                 value
                 if (feature_state.enabled and value is not None)
--- a/api/integrations/mixpanel/mixpanel.py
+++ b/api/integrations/mixpanel/mixpanel.py
@@ -47,7 +47,9 @@ class MixpanelWrapper(AbstractBaseIdentityIntegrationWrapper[MixpanelUserData]):
         for feature_state in feature_states:
             value = feature_state.get_feature_state_value(identity=identity)
             feature_properties[feature_state.feature.name] = (
-                value if (feature_state.enabled and value) else feature_state.enabled
+                value
+                if (feature_state.enabled and value is not None)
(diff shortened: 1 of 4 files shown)


khvn26:
Hi @genrichez, thanks for your contribution. While it does solve the original issue, there's more context to consider before we can produce a true solution (see [here](https://github.com/Flagsmith/flagsmith/issues/7701#issuecomment-4649759645)).

PR #8107, opened 2026-07-27
feat: push delivery of buffered events to external ClickHouse warehouses

gagantrivedi:

## Changes

Delivers buffered event objects to customers' own ClickHouse warehouses.

- Every 10 minutes, a recurring task fans out one task per active ClickHouse connection.
- Each run streams the environment's pending S3 objects (gzipped JSONEachRow from Firehose) into the customer's ClickHouse over HTTPS via `clickhouse-connect`'s `raw_insert` — no decompression or parsing in the app. `clickhouse-driver` cannot do this: the native protocol requires the client to build columnar blocks, which its maintainer confirms is [the only way data can be sent](https://github.com/mymarilyn/clickhouse-driver/issues/34) over that protocol.
- Delivered objects move to `archive/`; objects the warehouse rejects (bad data) move to `failed/` and the run continues; connection-level failures (auth, network, missing table) abort the run, leave everything in place, and surface on the connection's `status`/`status_detail`. A later successful run resets the status.
- A run that delivers nothing and rejects everything is treated as a failure, since that points at the table's schema rather than the objects.
- Each run is bounded by a time budget so it finishes inside its task timeout, deferring the rest of a backlog to the next tick.
- Redirects are refused: the internal-address guard validates the host being dialled, and following a redirect would let a permitted host bounce the request, and its event payload, to an address that was never checked.
- The HTTPS port is inferred from the stored native port (9440→8443, 9000→8123); the SSRF host re-check from verification is repeated before connecting.

## Follow-ups

- #8149 — per-object delivery tracking, for retries of transient failures and claim/lease mutual exclusion.
- #8150 — ask for the HTTPS port instead of inferring it, and move verification onto the interface delivery uses. Until then, a connection on a non-standard native port verifies healthy but cannot deliver.

Customer-facing docs (the required `events` DDL) are deferred until the schema-ownership question in #8150 is settled, hence the unticked docs box.

- Unit tests with moto S3 and a mocked ClickHouse boundary; 100% coverage on new code.
- The gzip passthrough, epoch-millis timestamps and the object-vs-connection error codes were verified against a local ClickHouse 24.8 with the real `flagsmith_exp.events` DDL.


gagantrivedi:
@themis-blindfold review

api/experimentation/warehouse_delivery_service.py
```diff
@@ -0,0 +1,212 @@
...
+    client = clickhouse_connect.get_client(
+        host=host,
+        port=http_port,
+        username=username,
+        password=password,
+        database=database,
+        secure=secure,
+        connect_timeout=CONNECT_TIMEOUT_SECONDS,
+        send_receive_timeout=INSERT_TIMEOUT_SECONDS,
+    )
```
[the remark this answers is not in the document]
gagantrivedi: Fixed in 25148a759. Confirmed the mechanism independently: the request path passes `retries=self.http_retries` to `PoolManager.urlopen` without `redirect=False` (`_backend/http_sync.py:304`), and on urllib3 2.7.0 `Retry.from_int(1)` yields `redirect=None`, so one redirect is followed from the total budget.

Delivery now builds its client with a `_NoRedirectPoolManager` whose `urlopen` forces `redirect=False`, passed via `get_client(pool_mgr=...)` so it covers the version handshake as you noted. It is constructed from `httputil.get_pool_manager_options()`, keeping the library's own TLS defaults (`CERT_REQUIRED`) rather than a bare manager. Covered by a unit test asserting the manager refuses redirects, plus an `isinstance` assertion that the client is actually given it.

api/experimentation/services.py
```diff
@@ -795,6 +800,117 @@ def mark_warehouse_pending_connection(
...
+            connection,
+            detail=warehouse_delivery_service.describe_delivery_error(exc),
+        )
+        flagsmith_experimentation_warehouse_delivery_runs_total.labels(
+            result="failure"
+        ).inc()
+        log.error("delivery.failed", exc_info=exc)
+        return
+
+    mark_warehouse_delivery_succeeded(connection)
```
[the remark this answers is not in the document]
gagantrivedi: Fixed in 25148a759. Agreed that a systematically wrong DDL is the likelier cause than N independent bad objects, given every record comes from the same pipeline.

A run with `delivered_count == 0 and rejected_count` now marks the connection errored with an actionable detail naming the `events` table, increments `..._delivery_runs_total{result="failure"}`, and logs `warehouse.delivery.all_objects_rejected`.

The narrower point you raise — that a *transient* rejection sends an object to `failed/` permanently, where the 30-day lifecycle rule then deletes it — is not fixed here, because it needs per-object attempt tracking. Raised as #8149.

api/experimentation/services.py
```diff
@@ -795,6 +800,117 @@ def mark_warehouse_pending_connection(
...
+                client,
+                bucket_name=bucket_name,
+                pending=pending,
+                log=log,
+            )
+    except Exception as exc:
+        # The warehouse itself is unusable; deliver nothing, leave every
+        # remaining object in place for the next run, and surface the
+        # breakage on the connection.
+        mark_warehouse_delivery_failed(
```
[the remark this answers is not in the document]
gagantrivedi: Fixed in 25148a759. The handler is now `except (DeliveryConfigError, ClickHouseError)`, so S3 failures propagate and fail the task for retry instead of marking the connection errored — matching `list_pending_objects`, which you correctly note already behaves that way.

Worth adding to your reasoning: because the old handler swallowed and returned, the task was also recorded as SUCCESS, so an S3 failure produced no retry and no failure signal on our side at all.

api/experimentation/warehouse_delivery_service.py
```diff
@@ -0,0 +1,212 @@
...
+    if is_internal_address(host, include_shared=True):
+        raise DeliveryConfigError(
+            "Host must not target internal or private network addresses."
+        )
+
+    if (http_port := NATIVE_TO_HTTP_PORT.get(port)) is None:
+        raise DeliveryConfigError(
+            f"No HTTP port is known for ClickHouse port {port}; "
+            f"expected one of {sorted(NATIVE_TO_HTTP_PORT)}."
+        )
```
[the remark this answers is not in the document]
gagantrivedi: Correct, and deliberately not fixed in this PR. Rather than reject unmapped ports at validation time, we are removing the native port from the config entirely and asking only for the HTTPS port — which also lets verification exercise the interface delivery actually uses, and check the `events` table exists. Raised as #8150 and called out in the PR description as a known limitation until it lands.

api/experimentation/warehouse_delivery_service.py
```diff
@@ -0,0 +1,212 @@
...
+
+if typing.TYPE_CHECKING:
+    from collections.abc import Iterator
+    from typing import Any
+
+    from clickhouse_connect.driver.client import Client
+
+    from experimentation.models import WarehouseConnection
+
+logger = structlog.get_logger("experimentation")
```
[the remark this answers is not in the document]
gagantrivedi: Fixed in 25148a759 — the `logger` and its `structlog` import are gone. Left behind when the delivery orchestration moved to `services.py`.

api/experimentation/services.py
```diff
@@ -795,6 +800,117 @@ def mark_warehouse_pending_connection(
...
+            warehouse_delivery_service.move_object(
+                bucket_name,
+                s3_key,
+                to_prefix=warehouse_delivery_service.FAILED_PREFIX,
+            )
+            rejected_count += 1
+            flagsmith_experimentation_warehouse_delivery_objects_total.labels(
+                result="rejected"
+            ).inc()
+            log.warning("delivery.object_rejected", exc_info=True)
```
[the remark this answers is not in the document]
gagantrivedi: Fixed in 25148a759. The bind now happens inside `_deliver_pending_objects` from the connection rather than arriving as a parameter, which is the form docgen can resolve. All five delivery events are in the regenerated catalogue, and the rejection event carries `connection.id` and `s3.key` so a specific object can be traced.

api/tests/unit/experimentation/test_warehouse_delivery_service.py
```diff
@@ -0,0 +1,331 @@
...
+        # Redirects must not be followed: the internal-address guard only
+        # validates the host being dialled.
+        pool_manager = get_client.call_args.kwargs["pool_mgr"]
+        assert isinstance(
+            pool_manager,
+            warehouse_delivery_service._NoRedirectPoolManager,
+        )
+        get_client.return_value.close.assert_not_called()
+
+    get_client.return_value.close.assert_called_once_with()
```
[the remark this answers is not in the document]
gagantrivedi: Added in 818213059 — `test_delivery_client__body_raises__still_closes_client` pins that the pooled HTTP connection is released when the block raises.

api/experimentation/tasks.py
```diff
@@ -106,6 +122,38 @@ def remove_environment_ingestion_key(key: str) -> None:
     ingestion_sync_service.delete_ingestion_key(key)
 
 
+@register_recurring_task(run_every=DELIVERY_INTERVAL)
+def deliver_events_to_external_warehouses() -> None:
+    connection_ids = WarehouseConnection.objects.filter(
+        warehouse_type=WarehouseType.CLICKHOUSE,
+    ).values_list("id", flat=True)
+    for connection_id in connection_ids:
+        deliver_events_for_connection.delay(kwargs={"connection_id": connection_id})
```
Zaimwa9, line 131: NIT: Should this be behind the `is_warehouse_feature_enabled` flag ?

api/experimentation/tasks.py
```diff
@@ -106,6 +122,38 @@ def remove_environment_ingestion_key(key: str) -> None:
     ingestion_sync_service.delete_ingestion_key(key)
 
 
+@register_recurring_task(run_every=DELIVERY_INTERVAL)
+def deliver_events_to_external_warehouses() -> None:
+    connection_ids = WarehouseConnection.objects.filter(
+        warehouse_type=WarehouseType.CLICKHOUSE,
```
Zaimwa9, line 128: NIT: we could have != Flagsmith, really not important at this stage

api/experimentation/warehouse_delivery_service.py
```diff
@@ -0,0 +1,233 @@
...
+def deliver_object(client: "Client", bucket_name: str, s3_key: str) -> int:
+    """Stream one gzipped events object into the customer's warehouse and
+    return the number of rows written.
+
+    The body is passed through untouched — ClickHouse decompresses and parses
+    it, so neither happens in this process. Delivery is at-least-once, so a
+    redelivered object duplicates rows; the distinct-aware aggregates in the
+    results queries account for that.
+    """
+    body = _get_s3_client().get_object(Bucket=bucket_name, Key=s3_key)["Body"]
```
Zaimwa9, line 221: Do we need to close the connection ?

api/experimentation/warehouse_delivery_service.py
```diff
@@ -0,0 +1,233 @@
...
+        Bucket=bucket_name,
+        Key=destination_key,
+        CopySource={"Bucket": bucket_name, "Key": s3_key},
+    )
+    s3.delete_object(Bucket=bucket_name, Key=s3_key)
+    return destination_key
+
+
+@contextmanager
+def delivery_client(connection: "WarehouseConnection") -> "Iterator[Client]":
```
Zaimwa9, line 162: I feel like the boundary between this new `warehouse_delivery_service` and the previous `services` is a bit unclear.
E.g this one is declared here and used in services.

As of I'm a bit unsure about being able to describe what `services` responsibility truly is versus this new one.

api/experimentation/services.py
```diff
@@ -795,6 +806,159 @@ def mark_warehouse_pending_connection(
...
+        )
+        flagsmith_experimentation_warehouse_delivery_runs_total.labels(
+            result="failure"
+        ).inc()
+        log.error("delivery.failed", exc_info=exc)
+        return
+
+    if delivered_count == 0 and rejected_count:
+        # Records all come from one ingestion pipeline, so every object being
+        # rejected points at the table's schema rather than the objects.
```
Zaimwa9, line 940: Will we be able to replay them in that case or safe to assume it's coming from a non-setup database ?

Zaimwa9 (dismissed):
Approving with couple of nits and questions

api/experimentation/tasks.py
gagantrivedi: It's effectively gated already: the permission check on the creation flow means a connection row only ever exists for an org with the flag on, so the loop can't pick up anyone who hasn't opted in

api/experimentation/warehouse_delivery_service.py
gagantrivedi: The CH client yes — closed by the context manager. The S3 body: drained to EOF on success (auto-released), but a failed insert left it to GC. Added an explicit close in b1ef98f10.

api/experimentation/services.py
gagantrivedi: A non-setup database (missing table, auth) doesn't reject — it aborts the run, objects stay in `events/` and replay automatically once fixed. Rejection is only content-vs-schema errors; those land in `failed/` and can be replayed manually (move back to `events/`) within the bucket's 30-day TTL.

api/experimentation/tasks.py
Zaimwa9: yes but does that stop delivering events? If we switch it back off, wouldn't it still keep pushing events ?

api/experimentation/warehouse_delivery_service.py
gagantrivedi: services.py is the business-logic entry point for the module; the individual service files deal with specific layers underneath (ClickHouse plumbing, S3, etc.). It'll take better shape once we add a non-ClickHouse warehouse

api/experimentation/tasks.py
gagantrivedi: I don't think that feature should control this but Let's defer that for now

Zaimwa9 (approved):

api/experimentation/services.py
```diff
@@ -803,6 +814,159 @@ def mark_warehouse_pending_connection(
...
+            ),
+        )
+        flagsmith_experimentation_warehouse_delivery_runs_total.labels(
+            result="failure"
+        ).inc()
+        log.error(
+            "delivery.all_objects_rejected",
+            objects__rejected_count=rejected_count,
+        )
+        return
```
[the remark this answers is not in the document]
gagantrivedi: as a follow up

PR #8129, opened 2026-07-28
fix(api): DCR rejects registrations without a client_name

khvn26:

## Changes

Claude Desktop still cannot register with `POST /o/register/` after #8114. The failure is now a clean 400, which leaves no Sentry trace, so we cannot tell which validation trips it.

In this PR, we:

1. Make `client_name` optional with a default of "MCP client". RFC 7591 §2 makes it optional and some MCP clients omit it. This is one of the candidate causes of the Claude Desktop failure.

2. Log every rejected registration as a `oauth2_metadata.registration.rejected` event carrying the error code, description, and user agent. The next rejected attempt will identify the failing validation and client from logs.

Added unit tests.

### DIFF
--- a/api/oauth2_metadata/serializers.py
+++ b/api/oauth2_metadata/serializers.py
@@ -23,7 +23,12 @@ _CLIENT_NAME_RE = re.compile(r"^[\w\s.\-()]+$", re.ASCII)
 
 
 class DCRRequestSerializer(serializers.Serializer[None]):
-    client_name = serializers.CharField(max_length=255, required=True)
+    # Optional per RFC 7591 §2; some MCP clients register without a name.
+    client_name = serializers.CharField(
+        max_length=255,
+        required=False,
+        default="MCP client",
+    )
     redirect_uris = serializers.ListField(
         child=serializers.CharField(max_length=2000),
         min_length=1,
--- a/api/oauth2_metadata/serializers.py
+++ b/api/oauth2_metadata/serializers.py
@@ -21,13 +21,17 @@ class OAuthConsentSerializer(serializers.Serializer):  # type: ignore[type-arg]
 # ASCII-only to prevent Unicode homoglyph spoofing on the consent screen.
 _CLIENT_NAME_RE = re.compile(r"^[\w\s.\-()]+$", re.ASCII)
 
+DEFAULT_CLIENT_NAME = "MCP client"
+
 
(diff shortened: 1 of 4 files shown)


api/oauth2_metadata/serializers.py
```diff
@@ -23,7 +23,12 @@ class OAuthConsentSerializer(serializers.Serializer):  # type: ignore[type-arg]
 
 
 class DCRRequestSerializer(serializers.Serializer[None]):
-    client_name = serializers.CharField(max_length=255, required=True)
+    # Optional per RFC 7591 §2; some MCP clients register without a name.
+    client_name = serializers.CharField(
```
Zaimwa9, line 29: We might want to `allow_blank` and `allow_null`. It looks like absent `client_,ame` is accepted but not null and empty values

Zaimwa9 (changes requested):
looks good overall but i'd like a second confirmation that empty values / null are deliberate before approving

api/oauth2_metadata/serializers.py
khvn26: 2e61d2035

We now assume absence of value for:
- absent `client_name`
- null `client_name`
- blank `client_name`

...and replace all three with our default name.

api/oauth2_metadata/serializers.py
```diff
@@ -21,9 +21,18 @@ class OAuthConsentSerializer(serializers.Serializer):  # type: ignore[type-arg]
...
 
 class DCRRequestSerializer(serializers.Serializer[None]):
-    client_name = serializers.CharField(max_length=255, required=True)
+    # Optional per RFC 7591 §2; null, blank and absent all mean "unnamed".
+    client_name = serializers.CharField(
+        max_length=255,
+        required=False,
+        allow_blank=True,
+        allow_null=True,
+        default=DEFAULT_CLIENT_NAME,
```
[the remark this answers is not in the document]
khvn26: Not reproducible.

I ran your verification script against DRF 3.15.2 and every case returns `'MCP client'`.

`allow_null` skips field-level validators, but serializer-level `validate_client_name` is always invoked by `Serializer.to_internal_value` after `field.run_validation`, which is where the normalisation lives. The parametrised cases in `test_dcr.py` assert this end-to-end through the endpoint.

Zaimwa9 (approved):
👍

PR #8132, opened 2026-07-28
feat(onboarding): wire the verify console to the real first-evaluation signal

### LINKED ISSUE #7767: Onboarding: wire the real first-evaluation signal

talissoncosta:

## Changes

Contributes to #7767, consuming the first-evaluation signal from #7623.

The verify console was stubbed: it read a `?connected` query param, so the connected state could be faked from the URL. It now reflects the real signal.

- New RTK Query service `useEnvironmentOnboardingStatus`, calling `GET environments/{key}/onboarding-status/`. Unauthenticated, keyed by the environment's client-side API key, served by Core rather than the edge host.
- `useOnboardingConnection` takes the environment key and polls that endpoint every 3s, flipping to `connected` once `first_evaluated_at` is no longer null. Edge reports the first evaluation asynchronously after the SDK's first request, so the flip is not instant; the value never reverts, so polling stops once it lands.
- The `?connected` hatch is gone.

The e2e test (`onboarding-tests.pw.ts`) mocks `onboarding-status/` and flips `first_evaluated_at` mid-test, so it now asserts through the real polling path rather than the query param.

Manually verified with the `onboarding_quickstart_flow` flag on:

- [x] `/getting-started` holds on "listening" while nothing has evaluated
- [x] Running the snippet in a scratch app flips the console to LIVE within a few seconds, no reload needed
- [x] `?connected` on the URL no longer does anything

`tsc` and `eslint` are clean on the changed files.

### DIFF
--- a/frontend/common/types/responses.ts
+++ b/frontend/common/types/responses.ts
@@ -1441,6 +1441,12 @@ export type Res = {
       rank: number
     }[]
   }
+  environmentOnboardingStatus: {
+    // Null until the environment's first SDK evaluation is reported by Edge.
+    first_evaluated_at: string | null
+    // A Core `KnownSDK` label (e.g. 'flagsmith-js-sdk') or 'unknown'.
+    first_evaluated_sdk_label: string | null
+  }
   profile: User
   onboarding: {}
   userPermissions: UserPermission[]
--- a/frontend/web/components/pages/onboarding/hooks/useOnboardingConnection.ts
+++ b/frontend/web/components/pages/onboarding/hooks/useOnboardingConnection.ts
@@ -1,13 +1,34 @@
-import { useLocation } from 'react-router-dom'
+import { useEffect, useState } from 'react'
+import { useGetEnvironmentOnboardingStatusQuery } from 'common/services/useEnvironmentOnboardingStatus'
 
 export type OnboardingConnectionStatus = 'listening' | 'connected'
 
-// Connection status for the verify console. Reports the pre-connection state
-// with a `?connected` query-param hatch to exercise the connected UI.
-// TODO(#7767): replace with the real first-evaluation signal.
-export const useOnboardingConnection = (): OnboardingConnectionStatus => {
-  const { search } = useLocation()
-  return new URLSearchParams(search).has('connected')
-    ? 'connected'
-    : 'listening'
+// Edge reports the first evaluation asynchronously after the SDK's first
+// request, so the flip isn't instant — poll until it lands.
+const POLL_INTERVAL_MS = 3000
+
+// Connection status for the verify console, driven by the real first-evaluation
+// signal: Edge reports the first SDK evaluation to Core, exposed at
+// GET environments/{key}/onboarding-status/. `first_evaluated_at` flips from
+// null to a timestamp once received; it never reverts, so we stop polling then.
+export const useOnboardingConnection = (
+  environmentKey: string,
+): OnboardingConnectionStatus => {
+  const [firstEvaluated, setFirstEvaluated] = useState(false)
+
+  const { data } = useGetEnvironmentOnboardingStatusQuery(
+    { environmentKey },
+    {
+      pollingInterval: POLL_INTERVAL_MS,
+      skip: firstEvaluated || !environmentKey,
+    },
+  )
+
+  useEffect(() => {
+    if (data?.first_evaluated_at) {
+      setFirstEvaluated(true)
+    }
+  }, [data])
+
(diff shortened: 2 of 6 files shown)


Zaimwa9:
@themis-blindfold review

talissoncosta:
Nit 1 fixed in 00f0f1ce8: moved the wait ahead of the snapshot and dropped the later one, which became redundant once the latch is held. Small correction on the mechanism, RTK Query fires its initial fetch on mount, so it wasn't a 3s gap, just a race with the first response.

frontend/web/components/pages/onboarding/hooks/useOnboardingConnection.ts
```diff
@@ -1,13 +1,34 @@
...
+export const useOnboardingConnection = (
+  environmentKey: string,
+): OnboardingConnectionStatus => {
+  const [firstEvaluated, setFirstEvaluated] = useState(false)
+
+  const { data } = useGetEnvironmentOnboardingStatusQuery(
+    { environmentKey },
+    {
+      pollingInterval: POLL_INTERVAL_MS,
+      skip: firstEvaluated || !environmentKey,
```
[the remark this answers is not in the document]
Zaimwa9: This can be ignored I believe

frontend/common/types/responses.ts
```diff
@@ -1441,6 +1441,12 @@ export type Res = {
       rank: number
     }[]
   }
+  environmentOnboardingStatus: {
+    // Null until the environment's first SDK evaluation is reported by Edge.
+    first_evaluated_at: string | null
+    // A Core `KnownSDK` label (e.g. 'flagsmith-js-sdk') or 'unknown'.
+    first_evaluated_sdk_label: string | null
```
Zaimwa9, line 1448: Non blocking / Followup: It's not used but maybe we could show from which sdk it's been evaluated as a nice to have

frontend/common/types/responses.ts
```diff
@@ -1441,6 +1441,12 @@ export type Res = {
...
+  }
```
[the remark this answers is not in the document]
Zaimwa9: can skip i believe

frontend/web/components/pages/onboarding/hooks/useOnboardingConnection.ts
```diff
@@ -1,13 +1,34 @@
...
+// null to a timestamp once received; it never reverts, so we stop polling then.
+export const useOnboardingConnection = (
+  environmentKey: string,
+): OnboardingConnectionStatus => {
+  const [firstEvaluated, setFirstEvaluated] = useState(false)
+
+  const { data } = useGetEnvironmentOnboardingStatusQuery(
+    { environmentKey },
+    {
+      pollingInterval: POLL_INTERVAL_MS,
```
Zaimwa9, line 22: Actually we could either increase a bit the polling (5-10sec) or have a backoff. It's true the user might go setup the sdk and could leave it there for a while and it's one DB read every 3sec (negligeable but still could be lowered without impacting too much the UX I believe)

Zaimwa9 (dismissed):
Couple of non-blocking comments that could be left for later

frontend/common/types/responses.ts
talissoncosta: Good call! I will explore this scenario.

frontend/web/components/pages/onboarding/hooks/useOnboardingConnection.ts
talissoncosta: Agreed on lowering it, though rather than hand rolling a backoff: RTK 2 ships `skipPollingIfUnfocused`, which pauses polling while the tab isn't visible and resumes when the user comes back. That covers exactly the case you describe, an abandoned tab costing reads for hours, and it's one option rather than custom timer logic. We're on RTK 1.9.1 so it doesn't exist for us yet.

So I'm putting up a separate PR for the RTK 2 + react-redux 9 bump. Worth doing anyway, we're currently running react-redux 8 under React 19 and its peer range only goes to 18. 

The surface seems to be small: no extraReducers anywhere, configureStore already uses the middleware callback form, and no connect() calls. 

Once that lands I'll stack the polling change on it: skipPollingIfUnfocused plus your interval bump to 5s.

Zaimwa9 (approved):

PR #8134, opened 2026-07-28
feat(onboarding): add lazy Lottie loading splash

talissoncosta:
Part of #7767.

Replaces the plain loader shown while the onboarding resources (org / project / envs / first flag) are created with a Lottie "setting up your workspace" splash. Ported from the onboarding prototype (`feat/onboarding-quickstart-7182`).

Lazy-loaded via `React.lazy` + `Suspense` so `lottie-web` and the ~180KB animation JSON stay out of the page's initial bundle (the `Loader` fallback shows meanwhile).

## Test plan
- `tsc` and `eslint` clean on changed files.
- Renders in the rspack build (same bundler + `lottie-react` import the prototype used).


kyle-ssg:
Whilst I like this @talissoncosta, it does add another form of loading, if we were to do this would we replace other areas of the app?

talissoncosta:
@kyle-ssg this was part of the new onboarding proposal, but I decided to hold off and validate it before moving forward.

talissoncosta:
I agree with you @kyle-ssg , we can close this PR for now!

PR #8137, opened 2026-07-28
fix(API): _validate_owner_removal counts non-existent group IDs as removals

bardock-2393:

## Changes

When `enforce_feature_owners` is on, removing group owners from a feature could be wrongly rejected with a 400 if the request included a stale or non-existent group ID alongside a real one. The validation counted every ID in the request as a removal, instead of only the IDs that were actually current owners, so a harmless extra ID could make it look like the last owner was being removed.

Contributes to https://github.com/Flagsmith/flagsmith/issues/8038

- [x] `_validate_owner_removal` now counts only the requested IDs that are actually current owners/group owners of the feature, instead of trusting the request's ID count.

Added a regression test (`test_remove_group_owners__enforce_owners_nonexistent_group_id_in_request__returns_200`) reproducing the issue's exact scenario: a feature with one user owner and one group owner, removing the group owner while also passing a non-existent group ID. Verified it fails with `400` against the old code and passes with the fix. Ran the full `test_unit_features_views.py` suite (298 passed, 2 pre-existing skips), plus `mypy` and the project's lint/format pre-commit hooks, all clean.

Review effort: 1/5

### DIFF
--- a/api/features/views.py
+++ b/api/features/views.py
@@ -448,8 +448,7 @@ class FeatureViewSet(viewsets.ModelViewSet):  # type: ignore[type-arg]
         serializer.is_valid(raise_exception=True)
         self._validate_owner_removal(
             feature,
-            owners_to_remove=0,
-            group_owners_to_remove=len(serializer.validated_data["group_ids"]),
+            group_ids_to_remove=serializer.validated_data["group_ids"],
         )
         serializer.remove_group_owners(feature)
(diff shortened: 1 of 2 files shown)


khvn26:
Hello @bardock-2393. The change looks sound to me. We'll approve if tests green.

bardock-2393:
@khvn26 Tests are green now — 3.11/3.12/3.13, E2E, and codecov all pass.

PR #8145, opened 2026-07-28
fix(code-help): make the SDK snippets runnable

talissoncosta:

## Changes

Follow-up to #8132. The onboarding's Python snippet failed on copy-paste, so I ran every snippet through its real toolchain. Most were broken:

- **Didn't compile:** .NET (`static` on a top-level declaration, and the constructor takes a `FlagsmithConfiguration`), Go (rune-literal key, `GetEnvironmentFlags` missing its `ctx`, no `package`/`main`), Next.js (missing paren, `<>{...}</>`).
- **Didn't run:** Python (`json.loads` with no import), Node (imports a default export the SDK lacks), PHP (undefined `$flags`, no `<?php`, install omitted the required PSR-18 client), Flutter (undefined `config`, `baseURI` passed to the wrong class), React Native (rendered `<div>`/`<p>`, which don't exist in RN), JavaScript (called a function the user doesn't have).
- **Wrong output:** Ruby commented with `//`, .NET with `#`, iOS printed the literal string `false`. The Rust and Flutter installs were prose, not commands.

Snippets now print their result, and code blocks get `tab-size` so Go's tabs don't render 8 columns wide.

`codeHelp.INIT` takes an optional feature name. The onboarding passes the flag it created, so the snippet stops referencing `banner_size`, a flag nobody has; the Features page passes nothing and keeps both placeholders as today. Flag names are no longer interpolated into variable names, since a user-chosen name with a space or hyphen broke the JS snippets.

Identity and segment snippets are in #8146.

A test covers all 15 init snippets in both shapes (368 unit tests). It needed `.js` in the Jest transform, with `node_modules` excluded again; the suite still runs in ~1s.

Then each snippet through its toolchain, in Docker where needed:

| Snippet | Result |
| --- | --- |
| Python, Node.js | Executed |
| Go | `go build`, `go run`, gofmt-clean |
| .NET | `dotnet build` succeeded, ran |
| Java | `mvn compile` succeeded |
| Rust | `cargo build`, no warnings |
| PHP | `php -l`, composer, executed |
| Flutter | `dart analyze`, no issues |
| React, JavaScript | Bundled and loaded in a browser; React renders the flag values |
| React Native, Next.js (both routers) | Parsed |
| Ruby | `ruby -c` only, no gem installed |
| curl | Executed, clean 404 for a placeholder key |
| iOS | Not tested, needs Xcode |

All runs used a placeholder key, so this shows the code is valid and the SDK calls exist, not that a real flag value came back. Java, Rust and Flutter are fragments, compiled inside a minimal wrapper; Go and .NET are now complete programs because their compilers reject fragments.

- [ ] Features page → SDK integration: two placeholder flags as today, plus the fixes
- [ ] Onboarding → Connect your code: every language references only the created flag

### DIFF
--- a/frontend/common/code-help/__tests__/init-snippets.test.ts
+++ b/frontend/common/code-help/__tests__/init-snippets.test.ts
@@ -1,0 +1,173 @@
+// The templates only need `Constants` for the custom-URL branch. Mocking it
+// keeps the legacy Flux stores (which touch `window`) out of this test.
+jest.mock('common/constants', () => ({
+  __esModule: true,
+  default: {
+    getFlagsmithSDKUrl: () => 'https://edge.api.flagsmith.com/api/v1/',
+    isCustomFlagsmithUrl: () => false,
+  },
+}))
+
+import initDotnet from 'common/code-help/init/init-dotnet'
+import initFlutter from 'common/code-help/init/init-flutter'
+import initGo from 'common/code-help/init/init-go'
+import initIos from 'common/code-help/init/init-ios'
+import initJava from 'common/code-help/init/init-java'
+import initJs from 'common/code-help/init/init-js'
+import initNextAppRouter from 'common/code-help/init/init-next-app-router'
+import initNextPagesRouter from 'common/code-help/init/init-next-pages-router'
+import initNode from 'common/code-help/init/init-node'
+import initPhp from 'common/code-help/init/init-php'
+import initPython from 'common/code-help/init/init-python'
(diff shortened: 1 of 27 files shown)


kyle-ssg:
@talissoncosta great job, I've gone through and run validation of every language against sample data and it all passes. I think that's as good as a review I can give, one thing's for sure is that it's an improvement - I think we should go ahead and merge this

kyle-ssg (approved):

Zaimwa9 (approved):

PR #8146, opened 2026-07-28
fix(code-help): correct the identity and segment snippets

talissoncosta:

## Changes

Split out of #8145, which fixes the same class of bug in the snippets the onboarding renders. These eleven render on the Identity, Identities and Segments pages instead.

Two bugs:

- **A stray `false` in the output.** Ten templates used `${cond && \`…\`}` inside a template literal, which renders the string `false` when `cond` is false, i.e. on SaaS. The iOS snippet read `Flagsmith.shared.apiKey = "KEY"false`. Now a ternary with an empty fallback.
- **The Node snippets don't run.** They import a default export `@flagsmith/nodejs` doesn't have, so the first line throws `SyntaxError: The requested module '@flagsmith/nodejs' does not provide an export named 'default'`. It's `import { Flagsmith }`, which is what `docs/integrating-with-flagsmith/sdks/server-side.mdx:257` already shows.

The Node import is verified in #8145, where the same fix to `init-node.js` was executed against `@flagsmith/nodejs@9`: it throws before the change and reaches the network call after.

The `false` fix is verified by rendering: every template was rendered with a non-custom API URL and checked for a boolean glued onto the preceding value. #8145 adds a unit test asserting exactly that for the init snippets.

- [ ] Identity page → the traits snippets for Java, Node, PHP, Ruby and Rust contain no stray `false`
- [ ] Identities page → the create-user snippets for iOS, Java, Node, PHP, Ruby and Rust contain no stray `false`
- [ ] Segments page → the traits snippets render the same
- [ ] Node snippet on either page starts with `import { Flagsmith } from "@flagsmith/nodejs"`

### DIFF
--- a/frontend/common/code-help/create-user/create-user-ios.js
+++ b/frontend/common/code-help/create-user/create-user-ios.js
@@ -11,8 +11,9 @@ func application(_ application: UIApplication,
   [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
 
   Flagsmith.shared.apiKey = "${envId}"${
-  Constants.isCustomFlagsmithUrl() &&
-  `\n  Flagsmith.shared.baseURL = "${Constants.getFlagsmithSDKUrl()}"\n`
+  Constants.isCustomFlagsmithUrl()
+    ? `\n  Flagsmith.shared.baseURL = "${Constants.getFlagsmithSDKUrl()}"\n`
+    : ''
 }
 
   // This will create a user in the dashboard if they don't already exist
--- a/frontend/common/code-help/create-user/create-user-java.js
+++ b/frontend/common/code-help/create-user/create-user-java.js
@@ -7,10 +7,11 @@ export default (
 ) => `${LIB_NAME_JAVA} ${LIB_NAME} = ${LIB_NAME_JAVA}
     .newBuilder()
     .setApiKey("${envId}")${
-  Constants.isCustomFlagsmithUrl() &&
-  `\n    .withConfiguration(FlagsmithConfig.builder()
+  Constants.isCustomFlagsmithUrl()
+    ? `\n    .withConfiguration(FlagsmithConfig.builder()
         .baseUri("${Constants.getFlagsmithSDKUrl()}")
         .build())`
+    : ''
 }
     .build();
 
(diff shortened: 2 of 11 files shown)


frontend/common/code-help/create-user/create-user-ios.js
```diff
@@ -11,8 +11,9 @@ func application(_ application: UIApplication,
   [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
 
   Flagsmith.shared.apiKey = "${envId}"${
-  Constants.isCustomFlagsmithUrl() &&
-  `\n  Flagsmith.shared.baseURL = "${Constants.getFlagsmithSDKUrl()}"\n`
+  Constants.isCustomFlagsmithUrl()
```
talissoncosta, line 14: Image

frontend/common/code-help/create-user/create-user-java.js
```diff
@@ -7,10 +7,11 @@ export default (
 ) => `${LIB_NAME_JAVA} ${LIB_NAME} = ${LIB_NAME_JAVA}
     .newBuilder()
     .setApiKey("${envId}")${
-  Constants.isCustomFlagsmithUrl() &&
-  `\n    .withConfiguration(FlagsmithConfig.builder()
+  Constants.isCustomFlagsmithUrl()
```
talissoncosta, line 10: Image

frontend/common/code-help/traits/traits-node.js
```diff
@@ -4,11 +4,12 @@ export default (
   envId,
   { FEATURE_NAME, LIB_NAME, NPM_NODE_CLIENT, TRAIT_NAME, USER_ID },
   userId,
-) => `import Flagsmith from "${NPM_NODE_CLIENT}"; // Add this line if you're using ${LIB_NAME} via npm
+) => `import { Flagsmith } from "${NPM_NODE_CLIENT}"; // Add this line if you're using ${LIB_NAME} via npm
 
 const ${LIB_NAME} = new Flagsmith({${
-  Constants.isCustomFlagsmithUrl() &&
```
talissoncosta, line 10: Image

frontend/common/code-help/create-user/create-user-java.js
talissoncosta: This is the old version, it was rendering false

frontend/common/code-help/create-user/create-user-ios.js
talissoncosta: This is the old version, it was rendering false

Zaimwa9 (approved):

kyle-ssg (approved):

PR #8147, opened 2026-07-28
feat(onboarding): name the SDK that reported the first evaluation

talissoncosta:

## Changes

Stacked on #8145, so the snippets a reviewer copies from the preview are the fixed ones. Follows @Zaimwa9's suggestion on #8132: `first_evaluated_sdk_label` arrives in the onboarding-status response and nothing used it.

The console's receipt line was invented, and asserted a flag value nothing had checked:

```
SDK initialized · flags loaded · show_demo_flag: true
```

It now reports what actually happened:

```
flagsmith-python-sdk · first evaluation received
```

That's the one thing the console couldn't tell you before: whether the evaluation came from the user's own app rather than something else hitting the environment.

- Core sends `'unknown'` when it can't identify the SDK from the user agent. The hook normalises that to null and the line falls back to `first evaluation received`.
- `useOnboardingConnection` returns `{ status, sdkLabel }` instead of a bare status string. The label is latched alongside the status, since skipping the query drops `data`.

Not in scope, still open from the same review: the poll ignores `isError`, so a 404 or 500 renders identically to "still listening".

The e2e mock already returned `flagsmith-js-sdk`, so `onboarding-tests.pw.ts` now asserts the console names it after the flip.

- [x] `typecheck` and `lint` clean on the changed files
- [x] Storybook: added a `ConnectedWithSdkLabel` variant next to `Connected`
- [ ] Onboarding → run an SDK against the environment: the receipt names that SDK, e.g. `flagsmith-python-sdk · first evaluation received`
- [x] Verified end to end on staging: ran the Python snippet, then `GET environments/{key}/onboarding-status/` returned `first_evaluated_sdk_label: "flagsmith-python-sdk"`

Note on the null-label path: `features/views.py:1040` only records the evaluation when `map_request_to_sdk_label` recognises the user agent, so `'unknown'` never reaches this endpoint from the SDK flow. The normalisation is defensive, and not reachable by hitting the API with plain `curl`.

## Screenshots

| | Before (main) | After |
|---|---|---|
| Console receipt, connected | | |
| Console receipt, SDK not identified | n/a | |

### DIFF
--- a/frontend/documentation/pages/onboarding/OnboardingTerminal.stories.tsx
+++ b/frontend/documentation/pages/onboarding/OnboardingTerminal.stories.tsx
@@ -14,7 +14,7 @@ const meta: Meta<typeof OnboardingTerminal> = {
     docs: {
       description: {
         component:
-          'The onboarding verify console. The checklist ticks as the user acts (copy install, copy snippet), and the first evaluation flips the badge to LIVE and prints the connection receipt. Always dark, since a terminal reads the same in light and dark mode.',
+          'The onboarding verify console. The checklist ticks as the user acts (copy install, copy snippet), and the first evaluation flips the badge to LIVE and prints the connection receipt, naming the SDK that reported it when Core could identify it. Always dark, since a terminal reads the same in light and dark mode.',
       },
     },
     layout: 'padded',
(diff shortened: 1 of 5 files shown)


Zaimwa9 (approved):
👍

Issue #7369, opened 2026-04-29
API accepts multivariate percentage splits over 100%


emyller:
### How are you running Flagsmith

- [x] SaaS at flagsmith.com

### Describe the bug

When a feature state ends up with `MultivariateFeatureStateValue.percentage_allocation` rows summing to more than 100%, [`process_environment_update`](https://github.com/Flagsmith/flagsmith/blob/dca4942a23acb60df3bb0d2764d266152727be6d/api/environments/tasks.py#L31-L44) fails when [`map_feature_state_to_engine`](https://github.com/Flagsmith/flagsmith/blob/dca4942a23acb60df3bb0d2764d266152727be6d/api/util/mappers/engine.py#L153) raises a Pydantic `ValidationError` from `FeatureStateModel`. While that state persists, the environment document on DynamoDB is never rewritten, so the Edge API keeps serving the previous value of every flag in that environment regardless of what the dashboard shows.

### Acceptance criteria

- It is not possible to leave a feature state's `MultivariateFeatureStateValue.percentage_allocation` rows summing to more than 100% in the database, by any means.

### How to reproduce

There is one [observed case](https://flagsmith.sentry.io/issues/7015694923/) in production, with unclear reproduction steps. I was able to reproduce it in the API via `mv-options`:

1. Pick a feature whose env-default feature state already has multivariate percentage allocations summing to 100%.
2. `POST /api/v1/projects/<project_id>/features/<feature_id>/mv-options/` with a body whose `default_percentage_allocation` is greater than zero.
3. `GET /api/v1/environments/<environment_api_key>/featurestates/?feature=<feature_id>` and observe the env-default feature state's `multivariate_feature_state_values` now summing to more than 100%.

<img width="793" height="1010" alt="Image" src="https://github.com/user-attachments/assets/459f1a55-3200-4e53-ba80-c49c8fa4a035" />

In the above screenshot, the dashboard displays the faulty MV setup, with an error message, when it already exists.

PR #8151, opened 2026-07-29
fix(Multivariate): API accepts multivariate percentage splits over 100%

bardock-2393:
Creating a multivariate option (`POST .../mv-options/`) only validated the new option's `default_percentage_allocation` against the option-level defaults of its sibling options. It never checked the actual `percentage_allocation` values already stored per environment, which can drift from those option-level defaults after a per-environment edit (e.g. rebalancing two variants for one environment via the feature state endpoint). A new option could therefore pass validation and still push one environment's feature state values over 100%, which then fails when the environment document is rebuilt, so the Edge API keeps serving stale flag values for that environment while the dashboard shows something else.

## Changes
- [x] Creating a multivariate option now also checks each environment's actual live feature-state allocations, not just sibling options' defaults, and rejects the request with a 400 if any environment would be pushed over 100%.
- [x] Added a regression test reproducing the drifted-allocation scenario from the issue: bump one environment's live allocation via the feature state endpoint, then add a new option that only overflows that one environment.

Contributes to #7369

Added an integration test in `tests/integration/features/multivariate/test_integration_multivariate.py` that reproduces the exact repro steps from the issue against a real Postgres database, and confirms the option is rejected with a 400 and not left partially created. Ran the full `features` unit + integration suite (1015 passed, unrelated pre-existing InfluxDB-dependent tests excluded), `mypy`, and the repo's `pre-commit` hooks (ruff lint/format).

Review effort: 2/5

### DIFF
--- a/api/features/multivariate/serializers.py
+++ b/api/features/multivariate/serializers.py
@@ -7,7 +7,10 @@ from rest_framework import serializers
 from core.constants import BOOLEAN, INTEGER
 from features.constants import CONTROL_VARIANT_KEY, RESERVED_VARIANT_KEY_MESSAGE
 from features.models import Feature, FeatureState
-from features.multivariate.models import MultivariateFeatureOption
+from features.multivariate.models import (
+    MultivariateFeatureOption,
+    MultivariateFeatureStateValue,
+)
 
 
 class NestedMultivariateFeatureOptionSerializer(serializers.ModelSerializer):  # type: ignore[type-arg]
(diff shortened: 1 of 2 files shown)


khvn26:
Hello @bardock-2393 and thanks for your contribution. It looks good to me; we'll take a look at CI results and approve if it passes.

bardock-2393:
Hi @khvn26 — just flagging that CI workflows are sitting at "action_required" and haven't run yet on the latest commit. Could you approve the run when you get a chance?

khvn26 (approved):
LGTM

PR #8166, opened 2026-07-30
fix: experiment results polish

Zaimwa9:

## Changes

UI/UX fixes for the experiment results page:

- When control wins, the summary card becomes "Control vs best variant" and shows the best treatment's lift negated (e.g. +65.1%) instead of "Baseline".
- Control's win probability is inferred from the treatments' chances instead of showing an em dash.
- Long variant names in the axis chart truncate with an ellipsis instead of overflowing the chart edge.
- Narrower Delta column, wider Variant column; delta track hidden below 1264px.
- Delta and win-probability tracks use a clearer surface colour in dark mode.
- Summary card values use the same typography as usage stats; winning variation shows its colour swatch; "Chance to be best" is green above 90%.
- `VariantName` takes a `fontSize` prop and vertically centres its swatch without disturbing the text baseline.
- Header metric name bumped by 2px.

Manually.

**Swatch + variant alignment**
image

**Increased text + ellipsis + reduced Delta column width + winning control version**
image


Zaimwa9:
@themis-blindfold review

frontend/web/components/experiments/results/VariantName.tsx
```diff
@@ -1,21 +1,60 @@
...
+      let size = fontSize
+      el.style.fontSize = `${size}px`
+      while (size > FIT_MIN_FONT_SIZE && el.scrollWidth > el.clientWidth) {
+        size -= 1
+        el.style.fontSize = `${size}px`
+      }
+    }
+    fitText()
+    const observer = new ResizeObserver(fitText)
+    observer.observe(el)
```
talissoncosta: the observer watches el, but fitText also changes el's font size, so it can end up triggering itself (that RO loop warning you sometimes get). could it watch the parent instead, or only refit when the width actually changes?

frontend/web/components/experiments/results/VariantName.tsx
```diff
@@ -1,21 +1,60 @@
...
-  </span>
-)
+// fit shrinks the font (down to FIT_MIN_FONT_SIZE, then ellipsis) until the
+// name matches the container width; it renders as a block, which drops the
+// prose baseline alignment — leave it off inside a sentence.
+const VariantName: FC<VariantNameProps> = ({ colour, fit, fontSize, name }) => {
+  const large = !!fontSize && fontSize >= 20
+  const ref = useRef<HTMLSpanElement>(null)
+
+  useLayoutEffect(() => {
```
talissoncosta: What about pulling this into a `useFitText` hook? `VariantName` goes back to presentational, and it also clears up the two things this brings with it: the observer watching the same element it mutates (can loop on itself), and the `fontSize` guard on the `style` prop.

```ts
function useFitText(maxFontSize, deps = []) {
  const ref = useRef(null)
  const [fontSize, setFontSize] = useState(maxFontSize)
  useLayoutEffect(() => {
    const el = ref.current
    if (!el) return
    const fit = () => { /* measure, shrink until it fits */ setFontSize(size) }
    const ro = new ResizeObserver(([e]) => { /* refit only when width changes */ fit() })
    ro.observe(el)
    return () => ro.disconnect()
  }, [maxFontSize, ...deps])
  return { ref, fontSize }
}
```

Then the component is just:

```tsx
const { ref, fontSize } = useFitText(24, [name])
return <span ref={ref} style={{ fontSize }}>…</span>
```

frontend/web/components/experiments/results/StatCard.tsx
```diff
@@ -10,9 +10,9 @@ type StatCardProps = {
 const StatCard: FC<StatCardProps> = ({ label, loading, value }) => (
   <ContentCard compact>
     <div className='text-secondary fs-caption'>{label}</div>
-    <div className='fs-3 mt-1'>
+    <h4 className='mb-0 mt-1'>
```
talissoncosta: The value renders inside `<h4>` but the label stays a `<div>`, so screen
readers announce the stat numbers as headings and they show up in the document
outline. Was the heading intended, or could a class give the same typography
without the heading semantics?

frontend/web/components/experiments/results/derive.ts
```diff
@@ -228,21 +253,27 @@ export const deriveSummary = (
   const controlIdentity = identities.find((v) => v.isControl)
 
   const direction = metric.direction ?? 'up'
+  let lift: number | null = winner.inference?.lift ?? null
+  if (winner.isControl) {
+    // Control's lead, kept in control's terms: the best treatment's lift
+    // negated, so the magnitude matches the delta column and axis chart.
+    const bestTreatment = getBestTreatment(metricResult, identities)
```
talissoncosta: `getBestTreatment` already runs inside `getWinningVariant` up on L209, so for a
control-win this recomputes it. Could `winner` carry it through instead?

talissoncosta (commented):
Looks good, just some suggestions.

frontend/common/hooks/useFitText.ts
```diff
@@ -0,0 +1,41 @@
+import { useLayoutEffect, useRef, useState } from 'react'
```
talissoncosta, line 1: What about moving it to `common/hooks/`? That's where most of hooks are current living  `components/hooks` only one

talissoncosta (dismissed):
Much cleaner, comments addressed. Just the hook location nit.

kyle-ssg (approved):

kyle-ssg:
@talissoncosta I've resolved these

talissoncosta (approved):

Issue #5814, opened 2025-07-23
Prevent inconsistent replica reads


emyller:
Context: https://github.com/Flagsmith/flagsmith/issues/5632#issuecomment-3104661634

Goals:
1. Implement an utility to help explicitly using replicas for certain operations.
2. Dismiss the replica router, as we're no longer going to implicitly read from replicas.

emyller:
We ran a thorough test in staging SaaS to verify the behavior.

## Given

1. A database replica is set up and available.
2. App settings (`REPLICA_DATABASE_URLS`) is modified to point to the new replica.

## When

1. 100x `GET https://api-staging.flagsmith.com/api/v1/flags/`
2. 100x `GET https://api-staging.flagsmith.com/api/v1/identities/?identifier=xuxa`

## Then

1. HTTP responses look the same as if only with a writer instance.
2. No substantial performance difference was observed.
3. The replica database had the expected activity and behavior (connections + queries).

<img width="1087" height="208" alt="Observed SQL queries" src="https://github.com/user-attachments/assets/936a104f-f66f-4dfb-9c61-4e7fdd75103a" />

---

This has been released as of [v2.196.0](https://github.com/Flagsmith/flagsmith/releases/tag/v2.196.0).

PR #8168, opened 2026-07-30
fix(SDK): Flag fetches fail when replica databases are configured

### LINKED ISSUE #8167: Replica reads break GET /api/v1/flags/ since 2.257.0: "the current database router prevents this relation"

- [ ] Self Hosted with Docker
- [ ] SaaS at flagsmith.com

### Describe the bug

Since **2.257.0**, `GET /api/v1/flags/` returns a 500 for any deployment that has
`REPLICA_DATABASE_URLS` configured. The SDK flag-fetch endpoint fails
unconditionally on the replica read path, so SDK clients cannot retrieve flags.

Introduced by commit
[`bea1512`](https://github.com/flagsmith/flagsmith/commit/bea1512a643956c9325f198b8fcc01a5352a5a35)
("perf(SDK): Decouple SDK identity request cost from version history", #8096).

## Affected versions

- First broken release: **2.257.0**
- Still present in **2.260.0** (current latest)
- Last good release: **2.256.1**

Verified by ancestry: `bea1512` is an ancestor of `v2.257.0` (`behind_by: 0`) and
is not contained in `v2.256.1` (`behind_by: 3`).

Corroborated empirically: our deployment was erroring at a steady rate on 2.260.0
across three clusters; rolling back to 2.256.1 stopped the 500s immediately, with
no occurrences on 2.256.1 and no other change to configuration.

## Root cause

`bea1512` added FK-cache priming inside `get_environment_flags_dict`:

```python
for feature_state in feature_states:
    # Every feature state in the queryset belongs to `environment`;
    # populating the relation up front keeps `feature_state.environment`
    # accesses from querying the database.
    feature_state.environment = environment
```

When called with `from_replica=True` (as `SDKFeatureStates.get` does at
`features/views.py:1067`), `feature_state._state.db` is `replica_N` while
`environment._state.db` is `default`.

Django's `ForeignKey.__set__` calls `router.allow_relation(...)` before allowing
the assignment. `app/routers.py` only defines `AnalyticsRouter` and
`ClickHouseRouter`, both of which return `None` for core models, so Django falls
back to its default `allow_relation` — which permits the relation only when both
objects live in the same database. Hence the `ValueError`.

The assignment is a no-op-by-intent optimisation, but it is only safe when the
router set declares primary and replicas to be relationally compatible.

## Suggested fix

Add a router that treats the primary and its replicas as one logical database,
following the `PrimaryReplicaRouter` example in the Django docs
(https://docs.djangoproject.com/en/stable/topics/db/multi-db/#topics-db-multi-db-routing):

```python
class ReplicaRelationRouter:
    def allow_relation(self, obj1, obj2, **hints):
        dbs = {obj1._state.db, obj2._state.db} - {None}
        if all(
            db == "default"
            or db.startswith(("replica_", "cross_region_replica_"))
            for db in dbs
        ):
            return True
        return None
```

Appending it **last** in `DATABASE_ROUTERS` preserves the existing
`AnalyticsRouter` / `ClickHouseRouter` restrictions, since Django uses the first
non-`None` response.

Alternatively, guard the assignment so it is skipped when the queryset was read
from a replica.

## Workarounds

- Pin to 2.256.1 or earlier.
- Unset `REPLICA_DATABASE_URLS`, which moves SDK reads onto the primary.

Note that `DATABASE_ROUTERS` is built as a hardcoded list in
`app/settings/common.py` with no environment-variable hook, so self-hosted
deployments cannot add the router above without overriding
`DJANGO_SETTINGS_MODULE`. Exposing a way to append routers would help here
independently of this bug.

### Expected behavior

--

### Screenshots

_No response_

khvn26:

## Changes

Closes #8167.

In this PR, we allow relations between all read objects by adding a custom `ReplicaRouter`.

Added unit tests:

- Parametrised coverage of `ReplicaRouter.allow_relation` across primary/replica, replica/replica, cross-region replica, unsaved, `analytics` and `task_processor` combinations.
- A regression test that loads a `FeatureState`, marks it as replica-loaded, and assigns the primary-loaded `Environment`.

### DIFF
--- a/api/app/routers.py
+++ b/api/app/routers.py
@@ -1,9 +1,13 @@
-from typing import Any, Literal
+from typing import Any, Literal, get_args
 
+from common.core.utils import ReplicaNamePrefix
 from django.db.models import Model
 
 AnalyticsDatabaseName = Literal["analytics"]
 ClickHouseDatabaseName = Literal["clickhouse"]
+
+PRIMARY_DATABASE_NAME = "default"
+REPLICA_NAME_PREFIXES: tuple[str, ...] = get_args(ReplicaNamePrefix)
 
 
 class AnalyticsRouter:
@@ -73,4 +77,33 @@ class ClickHouseRouter:
             return app_label in self.route_app_labels
         if app_label in self.route_app_labels:
             return False
+        return None
+
+
+class ReplicaRouter:
+    """
+    Treat the primary database and its read replicas as one logical database.
+
+    Replicas hold copies of the primary's data, so an object read from a replica
+    can legitimately be related to an object read from the primary.
+
+    Only `allow_relation` is implemented.
+
+    This router must be registered last so that the restrictions of other routers
+    keep taking precedence.
+    """
+
+    def allow_relation(self, obj1: Model, obj2: Model, **hints: Any) -> bool | None:
+        databases = (obj1._state.db, obj2._state.db)
+        if all(
+            database is not None
+            and (
+                database == PRIMARY_DATABASE_NAME
+                or database.startswith(REPLICA_NAME_PREFIXES)
+            )
+            for database in databases
+        ):
+            return True
+        # Defer to the remaining routers for anything
+        # that isn't the primary or one of its replicas.
         return None
--- a/api/app/settings/common.py
+++ b/api/app/settings/common.py
@@ -1543,6 +1543,9 @@ SEGMENT_MEMBERSHIP_DELETE_REFRESH_DELAY_SECONDS = env.int(
 # the default Postgres database whether or not a CH alias is configured.
 DATABASE_ROUTERS.append("app.routers.ClickHouseRouter")
 
+# Should be registered last.
+DATABASE_ROUTERS.append("app.routers.ReplicaRouter")
+
 if CLICKHOUSE_ENABLED:
     _clickhouse_db: dict[str, Any] = {
         "ENGINE": "core.db_backends.clickhouse",
--- a/api/tests/unit/app/test_unit_app_routers.py
+++ b/api/tests/unit/app/test_unit_app_routers.py
@@ -5,6 +5,8 @@ from django.apps import apps
 from django.db import models
 
 from app import routers
+from environments.models import Environment
+from features.models import Feature, FeatureState
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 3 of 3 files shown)


matthewelwell (commented):
I can't say that I fully understand the implementation here yet, but I'd like to understand how it interacts with `using_database_replica()` (see [here](https://github.com/Flagsmith/flagsmith-common/blob/main/src/common/core/utils.py#L153)), and whether it could see a return of the issue that `using_database_replica()` was implemented to fix (see [here](https://github.com/Flagsmith/flagsmith/issues/5814)).

khvn26:
I would say it interacts with `using_database_replica()` well, neither overlapping nor contradicting it precisely because we only implement `allow_relation` without touching `db_for_read`.

matthewelwell:
> I would say it interacts with `using_database_replica()` well, neither overlapping nor contradicting it precisely because we only implement `allow_relation` without touching `db_for_read`.

Ok, so if I understand correctly then, the new code in `ReplicaRouter.allow_relation` only gets executed for a replica database if the queryset is annotated with `using_database_replica()`, right? That sounds safe enough then I think. 

The answer to the second question is therefore that, since we're being explicit with where we're using replicas already, we shouldn't reintroduce the bug that was caused by allowing replicas across all querysets.

khvn26:
> the new code in ReplicaRouter.allow_relation only gets executed for a replica database if the queryset is annotated with using_database_replica()

That is correct.

> we shouldn't reintroduce the bug that was caused by allowing replicas across all querysets

I believe we are not doing that, because the original bug was caused by the `db_for_read` implementation, which we do not introduce.

matthewelwell:
Sorry, then one final thing that I'm not certain if it matters given our usage and implementation, but it definitely _could_ do: I think we only want to allow relations to be checked from replica -> primary. If we allow relations to be checked from primary -> replica, the relation might not exist due to replication delay?

khvn26:
Django passes the assigned object first: `feature_state.environment = environment` reaches the router as `allow_relation(environment, feature_state)`, primary before replica. Theoretical absence of relation would've manifested as `ObjectNotFound` before that.

Writing stale related objects is not a risk either: only the FK column is written if we call `obj.save()` after assigning a relation via `__setattr__`.

matthewelwell:
Ok, spoke to Kim to confirm that my lasting concern about whether this could be an issue in other execution paths is not a real concern. As the comment above states (somewhat implicitly), the objects are already valid objects and the `allow_relation` is only checking whether we should be able to assign one object as a relation of the other.

matthewelwell (approved):

PR #8169, opened 2026-07-30
fix: consolidate copy field pattern usage

### LINKED ISSUE #8125: Consolidate inline copy-to-clipboard fields on CopyField
CopyField (read-only input + icon-only copy button, optional wired label) shipped with the SCIM work, but the same pattern is still hand-rolled elsewhere: no label wiring, ad-hoc layouts, inconsistent buttons.

First target: CreateSAML's ACS URL. It needs a small `tooltip` passthrough on CopyField's `title` first, and the site's dead `w-full` class goes with it.

Field-shaped candidates from a `copyToClipboard` grep (a starting map, confirm per site):

- `CreateSAML.tsx` (ACS URL)
- `SDKKeysPage.tsx`, `ServerSideKeyRow.tsx`, `Token.js`
- `UsersAndPermissionsPage.tsx` (invite link)
- `EditHealthProvider.tsx`

Out of scope: copy-on-click text and snippet blocks (`FeatureName`, `JSONReference`, `CodeHelp`, `MCPSnippet`).

SahilJat:

## Changes

Closes #8125 

This PR standardizes the "read-only input + icon-only copy button" pattern by leveraging the `CopyField` component that was originally shipped with the SCIM work.

- Enhanced `<CopyField>` to support an optional wired label via the `title`, `tooltip`, and `tooltipPlace` props, passing through to `<FieldLabel>`.
- Generates stable, unique DOM ids using `useId()` for accessibility (`htmlFor`).
- Migrated the ACS URL field in `CreateSAML.tsx` from an ad-hoc layout to `<CopyField>`, preserving its tooltip and `stopPropagation` behaviour.
- Migrated the Environment Key field in `SDKKeysPage.tsx` to use `<CopyField>`.
- Removed the dead `w-full` CSS class usage from `CreateSAML.tsx` (the class had no corresponding rule).
- Added a `WithLabel` story to `CopyField.stories.tsx` to document the new props.

*Note: Other candidates like `ServerSideKeyRow`, `Token`, `UsersAndPermissionsPage`, and `EditHealthProvider` were evaluated but explicitly excluded because they represent different patterns (e.g. table rows with multiple actions, masked secret tokens, non-input text display).*

- **Manually via Storybook**: Verified the new `WithLabel` story renders correctly with the tooltip on the info icon.
- **Manually**: Navigated to Organisation Settings -> SSO -> SAML (Edit Mode) to ensure the ACS URL field renders correctly, the tooltip works, and the input value can be copied.
- **Manually**: Navigated to an Environment's SDK Keys page to ensure the Client-side Environment Key field renders as expected and copies correctly.
- **TypeScript**: Ran a full `tsc --noEmit` check to ensure the changes are fully type-safe.

### DIFF
--- a/frontend/documentation/components/CopyField.stories.tsx
+++ b/frontend/documentation/components/CopyField.stories.tsx
@@ -57,3 +57,15 @@ export const LongValue: Story = {
     </Container>
   ),
 }
+
+export const WithLabel: Story = {
+  render: () => (
+    <Container>
+      <CopyField
+        value='https://app.flagsmith.com/api/v1/auth/saml/my-org/response/'
+        title='Assertion Consumer Service (ACS) URL'
+        tooltip='Your identity provider posts SAML responses to this URL.'
+      />
+    </Container>
+  ),
+}
--- a/frontend/web/components/CopyField.tsx
+++ b/frontend/web/components/CopyField.tsx
@@ -1,33 +1,41 @@
-import React, { FC } from 'react'
+import React, { FC, useId } from 'react'
 import Button from './base/forms/Button'
 import Flex from './base/grid/Flex'
+import FieldLabel from './base/forms/FieldLabel'
 import Icon from './icons/Icon'
 import Input from './base/forms/Input'
 import Row from './base/grid/Row'
+import { TooltipProps } from './Tooltip'
 import Utils from 'common/utils/utils'
 
-// Minimal read-only input + icon-only Copy button pattern. The codebase has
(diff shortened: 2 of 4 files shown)


talissoncosta (approved):
Thanks @SahilJat! Really nice addition, this looks great. 🙌

Nice touch wiring the label with useId so the htmlFor is actually linked, and the note on the components you deliberately left out shows you scoped it carefully rather than sweeping everything in.

PR #8174, opened 2026-07-31
fix(invite-flow): invite flow asks signed-in users to create an account

### LINKED ISSUE #8164: Invite flow asks signed-in users to create an account
Click an invite while signed in and you get the signup form, headed "Create an account to accept your invite". If you already have an account, signup fails with "Email already exists" and you are stuck.

`main.js` skips restoring the token whenever the URL contains "invite", so you are treated as logged out. The guard predates the 2021 repo merge and nobody wrote down why. Best guess is that it stops one person's session accepting someone else's invite link, since links carry no email check.

Keep the session and let the invite page decide instead: no session goes to sign in, a live session asks before joining and offers to switch account. That keeps the protection and names the account, so joining as the wrong person becomes a choice rather than an accident.

Not worth doing: opening straight on the login form when the invited address already has an account. It needs a new public endpoint exposing the invited email and whether it is registered, only works for email invites, and saves one click.

Being stuck in the dead end is #8176. Fixing that does not fix this.

talissoncosta:

## Changes

Closes #8164

`main.js` skipped restoring the auth token whenever the URL contained "invite", so clicking an invite while signed in showed you the signup form headed "Create an account to accept your invite". Existing users then failed with "Email already exists".

The session is kept now, and `InvitePage` decides: no session sends you to sign in as before, a live one asks whether to join as the account you're already using, with the option to switch.

Invite routes are public because `InvitePage` handles both cases. Bouncing from `main.js` would flash the signup form at someone already signed in, since restoring a session is async.

`InvitePage` no longer accepts the invite on mount, and moved to TSX.

**Not yet run. Draft for that reason.** Lint and typecheck are clean, nothing else.

Needs walking by hand, both paths:

1. Signed in, open an invite link for an org you're not in. Expect the confirm screen naming your account, not the signup form.
2. Accept. Expect to join and land on the org.
3. "Use a different account". Expect logout, then the sign-in page for that invite.
4. Signed out, open an invite. Expect today's behaviour, unchanged.
5. Signed in as someone other than the invited address, open an **email** invite. Expect the existing mismatch error.

E2E has a new case asserting the confirm screen appears with a live session. That has not been run either.

Worth review attention: this touches auth bootstrap, so getting it wrong locks people out. The guard being removed predates the 2021 repo merge and nobody recorded why. My reading is that it stopped one person's session accepting another's invite link, since links carry no email check, and confirming before joining keeps that while naming the account.


talissoncosta:
Closing for now. The UI changes in #8177 cover the reported problem, and this one rewires how a session is restored on invite URLs, which is more risk than the remaining benefit justifies.

What stays: clicking an invite while signed in still shows the signup form. #8177 makes that recoverable rather than a dead end.

Branch kept if we want to pick it up.

PR #8196, opened 2026-07-31
fix(integrations): notify messaging integrations on identity flag overrides

### LINKED ISSUE #7983: Slack integration doesn't notify on identity flag overrides (Edge / SaaS)

bardock-2393:

## Changes

Closes #7983

Flipping a flag for a specific identity didn't notify Slack (or Datadog, New Relic, Dynatrace, Grafana), even though the same change at the project, environment, or segment level did. On SaaS, identity overrides live on the Edge and are recorded as an `EDGE_IDENTITY` audit log entry. The shared filter that decides which audit log events get forwarded to these integrations only allowed `FEATURE`, `FEATURE_STATE`, `SEGMENT`, and `EF_VERSION` events through, so `EDGE_IDENTITY` events were silently dropped before ever reaching the integration.

This adds `EDGE_IDENTITY` to that shared allow-list, so identity overrides on SaaS now notify the same way project, environment, and segment changes already do.

Added unit tests in `api/tests/unit/audit/test_unit_audit_signals.py` that create an `AuditLog` with `related_object_type=EDGE_IDENTITY` and assert it now reaches the Datadog and Slack signal handlers (previously it would have been filtered out before this fix). Also verified the full `api/tests/unit/audit/`, `api/tests/unit/integrations/slack`, `api/tests/unit/integrations/datadog`, and `api/tests/unit/edge_api/identities` suites pass (157 tests), plus `mypy` and the repo's lint hooks.

### DIFF
--- a/api/audit/signals.py
+++ b/api/audit/signals.py
@@ -121,6 +121,7 @@ def track_only_feature_related_events(signal_function):  # type: ignore[no-untyp
         RelatedObjectType.FEATURE_STATE,
         RelatedObjectType.SEGMENT,
         RelatedObjectType.EF_VERSION,
+        RelatedObjectType.EDGE_IDENTITY,
     ]
     return track_only(allowed_types)(signal_function)
 
--- a/api/tests/unit/audit/test_unit_audit_signals.py
+++ b/api/tests/unit/audit/test_unit_audit_signals.py
@@ -10,14 +10,17 @@ from audit.models import AuditLog
 from audit.related_object_type import RelatedObjectType
 from audit.signals import (
     call_webhooks,
+    send_audit_log_event_to_datadog,
     send_audit_log_event_to_dynatrace,
     send_audit_log_event_to_grafana,
+    send_audit_log_event_to_slack,
     send_feature_flag_went_live_signal,
     trigger_feature_state_change_webhooks,
 )
 from environments.models import Environment
 from features.models import Feature, FeatureState
 from features.versioning.models import EnvironmentFeatureVersion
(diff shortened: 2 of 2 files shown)


Zaimwa9:
@themis-blindfold review

PR #8198, opened 2026-07-31
fix(Utils): coerce non-string values before split in validateRule

### LINKED ISSUE #7536: TypeError: Cannot read properties of undefined (reading 'split')
Sentry Issue: [FLAGSMITH-FRONTEND-4NB](https://flagsmith.sentry.io/issues/7489683482/?referrer=github_integration)

```
TypeError: Cannot read properties of undefined (reading 'split')
  at Utils.validateRule (./common/utils/utils.tsx:798:39)
  at <anonymous> (./web/components/modals/CreateSegment.tsx:361:33)
  at Array.find (<anonymous>)
  at <anonymous> (./web/components/modals/CreateSegment.tsx:361:20)
  at Array.find (<anonymous>)
...
(15 additional frame(s) were not displayed)
```

bardock-2393:

## Changes

Closes #7536

`Utils.validateRule` (used by both the create-segment and edit-segment forms to decide whether a condition row is valid) crashes with `TypeError: Cannot read properties of undefined (reading 'split')` for some segment conditions.

#7779 previously added a guard so a missing (`null`/`undefined`) condition value returns `false` instead of crashing for the semver and Modulo operators. That guard only covers a *missing* value, though — it still calls `.split()` directly on `rule.value` once the value is present. If a condition reaches `validateRule` with a truthy value that isn't a string (for example a boolean, which can happen while a condition row's value is mid-update), `.split` doesn't exist on it and the same class of crash occurs, which is why this issue stayed open after #7779 merged.

This PR coerces the value to a string (`` `${rule.value}` ``) before calling `.split()` in both the semver and Modulo branches, so `validateRule` never throws regardless of what type the value happens to be, while keeping the existing pass/fail validation behaviour for well-formed input unchanged.

Added unit tests in `frontend/common/utils/__tests__/utils.test.ts` covering `validateRule` for the Modulo and semver operators with missing (`undefined`/`null`/empty string) and non-string (boolean) values, plus the `IS_SET`/`IS_NOT_SET` "hideValue" operator, and existing valid/invalid value cases. The two non-string-value tests fail on the code prior to this fix (confirmed by reverting the fix locally and re-running) and pass afterwards.

Ran the full frontend unit test suite (`npm run test:unit`): 368 tests passed across 28 suites. Also ran `npx eslint` and `npm run typecheck` against the changed files — no new lint or type errors introduced.

### DIFF
--- a/frontend/common/utils/__tests__/utils.test.ts
+++ b/frontend/common/utils/__tests__/utils.test.ts
@@ -1,0 +1,96 @@
+// These stores pull in `window` globals (via the flux dispatcher) that
+// aren't available under the jest `node` test environment, so they're
+// mocked out to allow `common/utils/utils` to load for real.
+jest.mock('common/stores/account-store', () => ({}))
+jest.mock('common/stores/project-store', () => ({}))
+jest.mock('common/store', () => ({ getStore: () => ({}) }))
+jest.mock('@flagsmith/flagsmith', () => ({
+  getValue: (_key: string, opts: { fallback: unknown }) => opts.fallback,
+}))
+
+import Utils from 'common/utils/utils'
+import { SegmentCondition } from 'common/types/responses'
+
+const buildRule = (overrides: Partial<SegmentCondition>): SegmentCondition => ({
+  operator: 'EQUAL',
+  property: 'trait',
+  value: '',
+  ...overrides,
+})
+
+describe('validateRule', () => {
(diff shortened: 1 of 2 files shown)


kyle-ssg:
I'm not sure there's a real case that will hit this but it looks safe to me

kyle-ssg (approved):

PR #8199, opened 2026-07-31
fix(api): update-traits returns 400 for malformed request body

### LINKED ISSUE #7951: update-traits crash when traits malformed
Sentry Issue: [FLAGSMITH-API-68J](https://flagsmith.sentry.io/issues/7596271009/?referrer=github_integration)

```
TypeError: util.engine_models.identities.traits.models.TraitModel() argument after ** must be a mapping, not list
(4 additional frame(s) were not displayed)
...
  File "edge_api/identities/views.py", line 180, in update_traits
    trait = TraitModel(**request.data)
```

## Acceptance criteria

- update-traits data validation broadens and responds 400 when parameters are malformed, e.g. `traits` is given a JSON array instead of an object.

bardock-2393:

## Changes

Closes #7951

The edge identity `update-traits` endpoint (`PUT /api/v1/environments/<api_key>/edge-identities/<uuid>/update-traits/`) crashed with a 500 error instead of returning a 400 when the request body wasn't a JSON object — for example, sending a JSON array as the payload.

This happened because the view unpacked the request body directly into a pydantic model (`TraitModel(**request.data)`). When `request.data` was a list, Python raised a `TypeError` on the `**` unpacking itself, before pydantic ever got a chance to validate anything, so the existing `except pydantic.ValidationError` handler never caught it.

The fix adds an explicit check that the request body is a JSON object before unpacking it, returning a standard 400 validation error otherwise — consistent with how the endpoint already reports other validation failures.

Added a test (`test_edge_identities_update_trait__malformed_payload__returns_400`) that sends a JSON array as the request body to the update-traits endpoint and asserts a 400 response. I confirmed this test reproduces the original crash on the unpatched code (fails with the exact `TypeError` from the linked Sentry issue) and passes with the fix applied.

Ran the full test module for this view (`api/tests/integration/edge_api/identities/test_edge_identity_viewset.py`, 22 tests) against a local Postgres database — all passing. Also ran `ruff` (lint + format) and `mypy` on the changed files with no issues.

### DIFF
--- a/api/edge_api/identities/views.py
+++ b/api/edge_api/identities/views.py
@@ -176,6 +176,8 @@ class EdgeIdentityViewSet(
         if not environment.project.organisation.persist_trait_data:
             raise TraitPersistenceError()
         edge_identity = self.get_object()
+        if not isinstance(request.data, dict):
+            raise ValidationError({"detail": "Request data must be a JSON object."})
         try:
             trait = TraitModel(**request.data)
         except pydantic.ValidationError as validation_error:
--- a/api/tests/integration/edge_api/identities/test_edge_identity_viewset.py
+++ b/api/tests/integration/edge_api/identities/test_edge_identity_viewset.py
@@ -696,3 +696,30 @@ def test_edge_identities_update_trait__persist_trait_data_disabled__returns_400(
     )
     # Then
     assert response.status_code == status.HTTP_400_BAD_REQUEST
+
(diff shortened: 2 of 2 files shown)


Zaimwa9:
@themis-blindfold review

Zaimwa9 (approved):
Thanks for the contribution

Issue #8218, opened 2026-08-04
Security: SSRF via unvalidated webhook and integration URLs in the management API


adamvialpando:
### Summary
Multiple management API endpoints accept an attacker-controlled absolute URL (e.g. `url`, `base_url` on webhooks and third party integrations such as Dynatrace / Datadog / New Relic) and persist it with no allow list, scheme restriction, DNS resolution check, or internal network deny list at the application layer. When a feature flag state later changes, the task processor issues an outbound request to that stored URL from inside the deployment's network. This is a server-side request forgery (SSRF) vector.

Combined with open self registration (`ALLOW_REGISTRATION_WITHOUT_INVITE=True`, which mints an active session), the chain can be driven end to end by an unauthenticated external actor.

### Steps to reproduce
1. `POST /api/v1/auth/users/` to create an account (open registration mints an active session).
2. Create an organisation, project and environment.
3. Register a webhook or a Dynatrace / Datadog / New Relic integration whose URL points at an internal or attacker controlled host.
4. Toggle a feature flag to trigger the server-side request.

PR #8220, opened 2026-08-05
fix(webhooks): validate organisation and integration URLs against SSRF

bardock-2393:
## Changes

Extends the existing SSRF-safe URL field (previously added for environment webhooks only) to `OrganisationWebhookSerializer`, the GitLab/Sentry/generic webhook integration serializers, and every integration built on the shared `IntegrationsModel.base_url` (Dynatrace, Datadog, New Relic, Grafana, Amplitude, Heap, Mixpanel, Rudderstack). These previously accepted any syntactically valid URL, including ones resolving to private/internal network ranges. Also tightens the shared validator to reject non-http(s) schemes (it previously allowed `ftp`/`ftps` via Django's default `URLValidator`).

Contributes to #8218 — coverage is limited to fields confirmed to be used for outbound server-side requests (verified via call-site grep against every app under `api/integrations/`). Fields that only build display links (not fetched server-side) were left untouched.

Note: this removes the ability for self-hosted deployments to point webhooks/integrations at `localhost` or internal Docker-network hosts, even intentionally in dev/test setups — consistent with the existing environment-webhook precedent, now applied repo-wide. Happy to follow up with a settings-driven allowlist if that's needed for dev workflows.

Added private-IP-rejection, DNS-rebind-rejection, and scheme-rejection tests across the touched serializers' existing test files. All 302 tests in the touched integration/webhook/organisation suites pass against real Postgres. Confirmed old-vs-new by reverting just the validator change: 2 tests fail (ftp/ftps accepted) on old code, pass on new code. Ran mypy (full project, clean) and the project's pre-commit hooks (all clean).

Review effort: 4/5


matthewelwell:
Hi @bardock-2393 , thanks for the contribution here. 

One general design question I would have is that it looks like we're applying the serializer field wherever we're using `URLField` at the model level. Is there a more generic approach we can take to link the `NoSSRFURLField` to the model level `URLField` instead to avoid having to manually define it on all of the serializers?

bardock-2393:
Widening the existing `_BaseIntegrationModelSerializer` hook from `field_name == "base_url"` to `isinstance(model_field, models.URLField)` would auto-cover `gitlab_instance_url` and Sentry's `webhook_url` — both are real `URLField`s on serializers already using that base.

It won't reach `WebhookConfigurationSerializer.url` or `OrganisationWebhookSerializer.url` though — their model field is a plain `CharField`, not `URLField`, so there's nothing to key off. Those stay manual.

Happy to widen the check for the two that qualify.

matthewelwell:
@bardock-2393 nope, that's not what I asked for. Please think about this outside of the context of the change you have already implemented. What about if we have other URLs elsewhere in the app in the future - I want to make sure we have a generic solution that won'y mean we forget to add this guarrail in the future. 

How about using [this](https://www.django-rest-framework.org/api-guide/serializers/#serializer_url_field)? Could we change the base `ModelSerializer` class perhaps? 

Side note: I'd love it to feel like I'm talking to a human here, not an agent. Please bear in mind that this is a human responding to you, and show us the same respect.

bardock-2393:
@matthewelwell I think  we need is [serializer_field_mapping](https://www.django-rest-framework.org/api-guide/serializers/#serializer_field_mapping), it controls how DRF picks a field type per model field. we can use that so every url field gets protected automatically, now and in the future.

sorry for drafting the earlier reply with an agent, that's on me. should I implement it ?

matthewelwell:
@bardock-2393 what's the benefit of using `serializer_field_mapping` over `serializer_url_field`?

bardock-2393:
@matthewelwell `serializer_url_field` just builds DRF's hyperlink url field (like "url": "/api/users/5/"), not real url data. `serializer_field_mapping` decides the actual field type per model field, e.g. mapping gitlab_instance_url to the safe url field, that's the one we need.

matthewelwell:
@bardock-2393 ok, yep, looks like you're right. Let's see how it looks with `serializer_field_mapping` then 👍

api/webhooks/apps.py
```diff
@@ -3,3 +3,14 @@
...
+    def ready(self) -> None:
+        from django.db import models
+        from rest_framework.serializers import ModelSerializer
+
+        from webhooks.fields import NoSSRFURLField
+
+        # Any `ModelSerializer` field built from a `models.URLField` gets the
+        # SSRF-safe field instead, for every current and future serializer,
+        # without each one having to opt in.
+        ModelSerializer.serializer_field_mapping[models.URLField] = NoSSRFURLField
```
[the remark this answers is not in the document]
bardock-2393: Checked `CodeReferencesRepository.url` (`projects/code_references/models.py`) specifically: it's never exposed through a `ModelSerializer`. It's only validated by a hand-declared `serializers.URLField()` on a plain `serializers.Serializer` (`FeatureFlagCodeReferencesScanSerializer` / `_CodeReferenceDetailSerializer` in `projects/code_references/serializers.py`), and the model itself is only ever written via `VCSRepository.objects.update_or_create(...)` directly, not through a serializer. So the `serializer_field_mapping` override doesn't reach it at all, that part of the finding doesn't apply here.

`FeatureExternalResource.url` was a real hit though (it IS a `models.URLField` exposed via `FeatureExternalResourceSerializer(ModelSerializer)`), fixed in 9c3e501 by declaring it explicitly as a plain `URLField()` again, since it's never fetched server-side (only regex-matched and sent as GitHub API payload data).

matthewelwell: Thanks for going ahead and implementing this based on our conversation, but seeing it implemented I actually don't think this is the right approach. 

Thinking about it some more, I think this should live at the model layer instead. Let's create a custom field in `core/fields.py`, built on a shared validator in `core/validators.py`:

```python
# core/fields.py
class NoSSRFURLField(models.URLField):  # type: ignore[type-arg]
    """
    A URL field restricted to http(s) URLs that do not resolve to internal
    or private network addresses.
    """

    default_validators = [validate_no_internal_address]
 
# integrations/common/models.py
base_url = NoSSRFURLField(blank=False, null=True)

# core/validators.py
def validate_no_internal_address(value: str) -> None:
    if is_internal_address(urlparse(value).hostname or ""):
        raise ValidationError("URLs must not target internal or private network addresses.")
```

Admittedly this also somewhat contradicts the approach that I took [here](https://github.com/Flagsmith/flagsmith/pull/7550), although I think we will still need the serializer field for any non-ORM built serializers. 

We should also migrate `AbstractBaseWebhookModel.url` to this new field (this will likely generate a bunch of migration fiels but is the right approach imo. 

Sorry @bardock-2393 , this has become a bit more involved based on this. Let me know if you're happy to take on this refactor, otherwise we can close the PR and take it on internally.

matthewelwell (changes requested):

bardock-2393:
@matthewelwell done, pushed.

moved it all to the model layer like you suggested. new validator in
`core/validators.py`, the `NoSSRFURLField` in `core/fields.py`, and the webhook
url field now uses it. dropped the serializer mapping thing. kept the old
serializer field just for the couple of serializers that aren't model backed.

one thing to note, it only checks on input, so urls already in the db aren't touched.

i've had a go at the migration side of it too. if it's not what you had in mind,
no problem closing this and picking it up internally. and if it just needs small
changes, happy to make them.

api/tests/unit/organisations/test_unit_organisations_serializers.py
matthewelwell, line 1: These additional tests seem like overkill now?

matthewelwell (changes requested):
Thanks @bardock-2393 , this looks mostly good to me now. One remaining comment, and I'd also like to see the outputt from `python manage.py sqlmigrate` for at least one of the migrations.

api/tests/unit/organisations/test_unit_organisations_serializers.py
bardock-2393: agreed, dropped them. the field level tests already cover this.

api/webhooks/apps.py
bardock-2393: moved to the model layer as you suggested, the mapping override is gone.

bardock-2393:
@matthewelwell here's the sqlmigrate output:

```
$ python manage.py sqlmigrate gitlab 0004_use_no_ssrf_url_field
BEGIN;
--
-- Alter field gitlab_instance_url on gitlabconfiguration
--
-- (no-op)
COMMIT;
```

matthewelwell (approved):

matthewelwell:
Thanks @bardock-2393 !

PR #8229, opened 2026-08-05
feat(onboarding): run the tour against the project named in the URL

talissoncosta:

## Changes

Stacked on #8217. Suggested by @Zaimwa9.

`/getting-started` carries no project, so the tour ran against the org's first one. You could be inside project B, click Getting Started, and see project A.

- The nav link now names the project: the one you're in (`TopNavbar` already has `projectId`), else the one you were last in.
- The page runs against that project, but only if it's in this organisation's list. Otherwise it falls back, so a link carried across orgs or an edited URL can't point the tour somewhere it shouldn't.
- A bare `/getting-started`, which is what a fresh signup gets, behaves as before.

A query param rather than a nested route: `/getting-started` has to work before any project exists. `/project/:id/environment/:key/getting-started` is the better shape and would delete this selection code, but it needs an entry that provisions and redirects, plus a decision on whether the tour keeps its chromeless layout. Worth doing with the drawer.

Environment choice is unchanged, still `Development` or the first one. `lastEnv` holds an environment too, but using it would mean telling someone to wire up production during a tour.

- [x] 12 unit tests: URL project wins, id matched as a string, id outside the org ignored, no param, empty org, plus `parseLastEnv` against absent and malformed storage
- [x] `test:unit` (406), `typecheck` and `lint` clean

- [ ] Inside project B: link carries `?project=<B>` and the tour shows B
- [ ] Outside a project: falls back to the one you were last in
- [ ] After switching org: no stale project from the previous one
- [ ] URL edited to another org's project: falls back
- [ ] New signup: unchanged

### DIFF
--- a/frontend/common/hooks/__tests__/useLastEnv.test.ts
+++ b/frontend/common/hooks/__tests__/useLastEnv.test.ts
@@ -1,0 +1,16 @@
+import { parseLastEnv } from 'common/hooks/useLastEnv'
+
+describe('parseLastEnv', () => {
+  it('reads what usePageTracking wrote', () => {
+    const raw = JSON.stringify({ environmentId: 'abc', orgId: 1, projectId: 2 })
+    expect(parseLastEnv(raw)).toEqual({
+      environmentId: 'abc',
+      orgId: 1,
+      projectId: 2,
+    })
+  })
+
+  it.each([null, '', 'not json', '{'])('survives %p', (raw) => {
+    expect(parseLastEnv(raw)).toBeNull()
+  })
+})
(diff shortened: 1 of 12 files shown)


talissoncosta:
Closing this one. Walking the cases, it buys less than it looked like:

- A fresh signup has no `?project=` on the URL and one project, so `projects[0]` is already the right answer.
- An established customer clicking the nav link lands on a project that almost certainly has flags, so #8217 shows them the already-set-up state either way. More predictable, same destination.
- It only genuinely helps with two or more projects while you're inside a new empty one and `projects[0]` is an established one.

The change that actually fixes "established customers shouldn't be in the tour" is gating the nav link on whether the account has already evaluated a flag. Once that lands, everyone who can reach /getting-started is a new user with a single project, and `?project=` equals `projects[0]` for all of them. Worth revisiting if the onboarding drawer re-opens the entry point to established accounts.

PR #8231, opened 2026-08-06
feat: expose distinct warehouse event names per connection

Zaimwa9:
- [x] I have read the [Contributing Guide](/Flagsmith/flagsmith/blob/main/CONTRIBUTING.md).
- [x] I have added information to `docs/` if required so people know about the feature.
- [x] I have filled in the "Changes" section below.
- [x] I have filled in the "How did you test this code" section below.

## Changes

Adds `GET /environments/{key}/warehouse-connections/{id}/events/`, returning the distinct event names recorded in the connection's warehouse so metric creation can offer them for selection.

- Dispatches on warehouse type: managed ClickHouse queried directly, BYO ClickHouse via `delivery_client` with a 5-minute per-connection cache (successes and failures).
- Most recently seen first, capped at 500, `is_truncated` flag as the hook for future pagination.
- Pagination deferred: brand-new feature, and a single cached query contains warehouse costs.
- `503` when unreachable, `400` for unsupported types; read throttle scope applies.

## How did you test this code?

Parametrised unit tests (dispatch, truncation, caching, response mapping); manual check against a ClickHouse Cloud instance via a BYO connection.

PR #8234, opened 2026-08-06
feat: metric event name combobox fed by warehouse events

Zaimwa9:

## Changes

Stacked on #8231.

Turns the Event name field in metric creation into a combo box listing the distinct event names from the environment's connected warehouse, most recently seen first.

- New `EventNameSelect` component wrapping react-select's `CreatableSelect`: type-to-filter, free text kept via a `Use "xyz"` option, and an inline warning when the chosen event hasn't been received by the warehouse yet.
- Degrades to a plain creatable input when there is no connection, the warehouse type is unsupported, or the fetch fails.
- Dropdown overlays the page instead of growing it, adapts its height to the viewport (flips above when space is short), and keeps a visible scrollbar plus a partially cut row as scroll affordances.
- New `getWarehouseConnectionEvents` RTK Query endpoint and response types; the edit-metric flow inherits the combo box.
- Pagination is deferred (see #8231): the API caps at the 500 most recent events and filtering is client-side; the `is_truncated` flag is the hook for server-side narrowing later.

- Parametrised Jest tests on the option-building and unknown-event helpers.
- Manual create and edit flows against a ClickHouse Cloud BYO connection.
- Small-viewport dropdown behaviour (flip, adaptive height, scrollbar) and the free-text warning checked manually.

### DIFF
--- a/frontend/web/components/experiments/EventNameSelect/EventNameSelect.tsx
+++ b/frontend/web/components/experiments/EventNameSelect/EventNameSelect.tsx
@@ -1,0 +1,65 @@
+import { FC, useMemo } from 'react'
+import CreatableSelect from 'react-select/creatable'
+import {
+  useGetWarehouseConnectionEventsQuery,
+  useGetWarehouseConnectionsQuery,
+} from 'common/services/useWarehouseConnection'
+import { WarehouseType } from 'common/types/responses'
+import { useRouteContext } from 'components/providers/RouteContext'
+import Icon from 'components/icons/Icon'
+import { buildEventOptions, EventOption, isUnknownEvent } from './utils'
+import './EventNameSelect.scss'
+
+type EventNameSelectProps = {
+  value: string
+  onChange: (value: string) => void
+}
+
+const SUPPORTED_WAREHOUSE_TYPES: WarehouseType[] = ['flagsmith', 'clickhouse']
+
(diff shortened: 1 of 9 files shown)


frontend/web/components/experiments/EventNameSelect/EventNameSelect.tsx
```diff
@@ -0,0 +1,65 @@
...
+        options={options}
+        value={value ? { label: value, value } : null}
+        onChange={(option: EventOption | null) => onChange(option?.value ?? '')}
+        placeholder='e.g. checkout_completed'
+        formatCreateLabel={(input: string) => `Use "${input}"`}
+        noOptionsMessage={() => 'Type to add a new event'}
+      />
+      {showWarning && (
+        <span className='d-flex align-items-center gap-1 text-warning fs-small mt-1'>
+          <Icon name='warning' width={14} className='text-warning' />
```
talissoncosta, line 56: Do you mind checking if the className is valid here? I think it will inherit from parent

talissoncosta (approved):
Looks good! Do you mind just checking the open comments from code rabbit ?

frontend/web/components/experiments/EventNameSelect/EventNameSelect.tsx
Zaimwa9: it is 👍

Issue #7818, opened 2026-06-18
Stop using the SegmentRule and Condition models


khvn26:
Migrate the last segment rule/condition readers (release pipeline, organisation export) onto `Segment.rules_data` and stop writing rows. Leave the models so historical segments can still be read, and mark them as deprecated.

### Acceptance criteria

- `SegmentRule` and `Condition` are clearly marked as deprecated. 
- Creating or updating a segment writes no `SegmentRule` or `Condition` rows.
- The release-pipeline phased-rollout percentage is read correctly from `rules_data`.
- An organisation import produces `rules_data` and no `SegmentRule` and `Condition` rows.
- The OpenAPI schema does not rely on the `SegmentRule` model.
- The ordering settings are gone and no test depends on them.
- The newly introduced `Segment.rules_data` is renamed to `Segment.rules` — rules naturally flow from the serialiser into the model.

Issue #7819, opened 2026-06-18
Streamline segment mutations in the frontend


khvn26:
Move the normal segment edit flow to full-replace semantics. Drop the `delete: true` convention both from frontend and backend.

### Acceptance criteria

- Removing a rule or condition in the editor omits it from the request body; no entry carries `delete: true`.
- `SegmentRule` and `SegmentCondition` types no longer carry `id`.
- Editing a segment work against the id-free API.
- API does not expect `delete: true` anymore.
- The change request segment diff renders unchanged.

matthewelwell:
> Both sides ship in one PR since the frontend and API live in the same repo and deploy together.

@khvn26 this isn't true for SaaS or a lot of self-hosted deployments. We'll have to have either the FE or the API be able to support both paths.

khvn26:
@matthewelwell Not a concern; the backend is backwards-compatible up to this point, and we guarantee minor-version compat.

PR #8245, opened 2026-08-08
feat(Segments): Write rules as plain JSON

### LINKED ISSUE #7815: Store segment rule JSON alongside existing models
We want to start storing segment rule data as JSON, but not abandon the existing relational model yet. This bounds the scope nicely for easier review, and gives us enough wiggle room to smoke test in prod and/or rollback if needed.

This issue owns the entire write side: every segment write goes through one service that stores `rules_data`, so no later issue migrates a write site. Reads are unchanged here; the engine and API still read rows.

The writes are switched over completely to `rules_data` in #7818 — we stop writing the rules and conditions models once that's merged. 

### Acceptance criteria

- A segment created or updated through the serializer stores JSON data equal to its reconstructed rule and condition tree, with segment rules and condition rows still written. The nested (n = 2) structure is maintained.
- The segment versioning history for JSON data is maintained as expected.
- A cloned segment via has JSON data matching the source segment.
- A LaunchDarkly import populates both JSON and relational data for rules/conditions.
- A request body for `/api/v1/projects/{project_pk}/segments/{id}/` containing a rule or condition `id` is accepted; the `id` is ignored and not persisted in the JSON data.
- A request flagging a rule or condition with `delete: true` omits that entry from both the JSON and relational data.
- The condition-count limit, value-length limit, and `WhitelistedSegment` exemption are enforced both for JSON and relational data.
- The backfill migration sets JSON data for every pre-existing live segment. Historical segments remain as they are.

emyller:
- [x] ~I have added information to `docs/` if required so people know about the feature.~

## Changes

Closes https://github.com/Flagsmith/flagsmith/issues/7815

This is the first step towards getting rid of `SegmentRule` and `Condition` models, and the tree complexity they have brought upon us thus far.

We achieve:

- [x] The `Segment` model earns a new `rules_data` attribute
  - [x] Creating / updating segments will snapshot clean JSON rules to `rules_data`
  - [x] Cloning / evolving segments carry JSON rules along with other attributes.
  - [x] Importing segments (e.g. from LaunchDarkly) also stores `rules_data`.
  - [x] All **active** segments are backfilled with `rules_data`.
- [x] Existing tests interacting with old models are _marked for deletion_ — most are replaced with new surviving tests.
- [x] Code paths interacting with old models are _marked for deletion_.

Unplanned things we also achieve:

- [x] Overall test code and coverage improvements.
- [x] Segments API will now **reject** rules nested more than two levels deep.
  - We used to accept them, before we started to ignore them.
  - Unfortunately, active segments with more than two levels of rules exist, and all conditions are evaluated recursively, despite the front-end being unable to represent them visually in full.
- [x] Fix some leaking writable segment fields, now read-only.

Chose not to achieve:

- **A service to centralise writing JSON rules**, because it would be scaffolding that we'd only use until #7818. By then, `rules` should need no special treatment as a JSON field, properly validated by the interface.

Deleted LoC will look great in #7818!

Exhaustive unit testing. The migration was also benchmarked and optimised.

> [!IMPORTANT]
> **This includes a data migration** to backfill active segments with JSON rules. The migration code was exhaustively tested and tweaked. On a very limited Postgres local container (1 CPU, 2 GB memory), it took ~16s to rasterise rules of 100K segments with varying depth of rules.


api/tests/unit/segments/test_unit_segments_views.py
```diff
@@ -1883,9 +2123,33 @@ def test_clone_segment__valid_name__returns_cloned_segment(
...
+            "updated_at": mocker.ANY,
+            "name": new_segment_name,
+            "description": segment.description,
+            "project": project.id,
+            "feature": None,
+            "version_of": cloned_segment.id,
+            "metadata": [],
+            "membership_counts": [],
+            "rules": [],  # TODO: Should contain rules as per https://github.com/Flagsmith/flagsmith/issues/7818
+        }
```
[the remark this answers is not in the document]
emyller: This assessment is myopic.

Right now, and also before this pull request, this test is cloning a segment that has no rules [rows] in it. This patch improves it today by:

- Ensuring `rules_data` is carried along and persisted, though not present in the API response yet. This is a clear goal of the issue this PR contributes to.
- Marking `"rules"` in the JSON response to contain actual rules after #7818 is merged. The test is expected to automatically fail since the `segment` fixture earned `rules_data`.

api/segments/serializers.py
```diff
@@ -139,19 +150,20 @@ def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
...
         metadata_data = validated_data.pop("metadata", [])
+        self._set_rules_data(validated_data)
         segment = super().create(validated_data)  # type: ignore[no-untyped-call]
         self._update_metadata(segment, metadata_data)
         enqueue_membership_refresh(segment.project)
         return segment
 
     def update(self, segment: Segment, validated_data: dict[str, Any]):  # type: ignore[no-untyped-def]
         metadata = validated_data.pop("metadata", [])
+        self._set_rules_data(validated_data)
```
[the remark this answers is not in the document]
emyller: Good catch, fixed in 9b5e9e73b0f45e4a7c166bb0c0e27f99812523c2.

api/segments/validators.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def __call__(
+        self,
+        attrs: dict[str, Any],
+        serializer: serializers.BaseSerializer,  # type: ignore[type-arg]
+    ) -> None:
+        rules_data: list[LegacySegmentRule] = serializer.initial_data.get("rules", [])
+        self._validate_depth(rules_data)
+        if not self._is_whitelisted(serializer):
+            self._validate_condition_count(rules_data)
```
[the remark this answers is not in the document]
emyller: Fixed in c32aea733a2d4280c5603bc2a3ca59516d3d3bc7. You deserve a carrot.

api/segments/validators.py
```diff
@@ -0,0 +1,71 @@
...
+            len(rule_data.get("conditions", []))
+            + self._count_conditions(rule_data.get("rules", []))
+            for rule_data in rules_data
+        )
+
+    @staticmethod
+    def _is_whitelisted(serializer: serializers.BaseSerializer) -> bool:
+        return bool(
+            (segment := serializer.instance)
+            and WhitelistedSegment.objects.filter(segment=segment).exists()
```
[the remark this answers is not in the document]
emyller: This seems outdated?

api/tests/unit/segments/conftest.py
```diff
@@ -0,0 +1,77 @@
...
+                    ],
+                ),
+                {"segment": ["Rules must not be nested more than 2 levels deep."]},
+            ),
+            id="rules-nested-too-deep",
+        ),
+    ],
+)
+def invalid_rules_case(request: FixtureRequest) -> InvalidSegmentRulesCase:
+    return request.param  # type: ignore[no-any-return]
```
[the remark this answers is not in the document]
emyller: Agreed! 2d70da3fb0f99f47851cbf631f46c87b71af26d1

api/integrations/launch_darkly/services.py
```diff
@@ -409,7 +503,20 @@ def _create_feature_segment_from_clauses(
...
+        import_request=import_request,
+        segment_name=segment.name,
+        clauses=clauses,
+    )
+    rules_data = segment.rules_data or [  # LaunchDarkly environments share the segment
+        {"type": constants.ALL_RULE, "conditions": [], "rules": []}
+    ]
+    rules_data[0]["rules"] += subrules
+    segment.rules_data = rules_data
+    segment.save(update_fields=["rules_data"])
```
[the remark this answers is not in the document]
emyller: A second import is not supported. If the new behaviour [JSON] is consistent to the current behaviour [rows], we're fine.

api/segments/migrations/0031_add_segment_rules_data.py
```diff
@@ -0,0 +1,93 @@
...
+
+    segments = Segment.objects.filter(
+        id=models.F("version_of"),  # Means "current version"
+        deleted_at__isnull=True,
+    ).only("id").prefetch_related(
+        models.Prefetch("rules", rules, to_attr="live_rules"),
+        models.Prefetch("live_rules__conditions", conditions, to_attr="live_conditions"),
+        models.Prefetch("live_rules__rules", rules, to_attr="live_rules"),
+        models.Prefetch("live_rules__live_rules__conditions", conditions, to_attr="live_conditions"),
+        models.Prefetch("live_rules__live_rules__rules", rules, to_attr="live_rules"),  # rasterise recurses one level deeper
```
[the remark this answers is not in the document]
emyller: Good catch. While we only intend to support two levels of segment rules nesting, I found history of existing segments with 3~4 layers. 04b05a6aaf4d592a8eea5783b737b5c20601f30e

api/tests/unit/segments/test_unit_segments_views.py
```diff
@@ -1059,7 +1049,173 @@ def test_update_segment__system_segment__returns_404(
...
+                    },
+                ],
+                "rules": [],
+            },
+        ],
+    }
+
+
+def test_update_segment__rules_and_conditions_with_ids__ignores_ids(
+    admin_client: APIClient,
```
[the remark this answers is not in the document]
emyller: Nah we're keeping it for now.

api/integrations/launch_darkly/services.py
```diff
@@ -409,7 +503,20 @@ def _create_feature_segment_from_clauses(
...
+        import_request=import_request,
+        segment_name=segment.name,
+        clauses=clauses,
+    )
+    rules = segment.rules_data or [  # LaunchDarkly environments share the segment
+        {"type": constants.ALL_RULE, "conditions": [], "rules": []}
+    ]
+    rules[0]["rules"] += subrules
+    segment.rules_data = rules
+    segment.save(update_fields=["rules_data"])
```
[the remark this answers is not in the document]
emyller: 6f0394ba3976f1feeb55f268973f206f606f21c2

api/integrations/launch_darkly/services.py
```diff
@@ -1152,41 +1274,42 @@ def process_import_request(
...
-        )
+            # Count deprecated flags for reporting
+            import_request.status["deprecated_flag_count"] = sum(
+                1 for ld_flag in ld_flags if ld_flag["deprecated"]
+            )
 
-        # Refresh membership counts for the segments the import just created.
-        enqueue_membership_refresh(import_request.project)
+            # Refresh membership counts for the segments the import just created.
+            enqueue_membership_refresh(import_request.project)
```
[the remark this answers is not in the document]
emyller: 080cde018cb659dffdf7e81f47c3287d6db6a13c

🥕

emyller:
@coderabbitai review

api/tests/unit/integrations/launch_darkly/test_services.py
```diff
@@ -103,6 +107,30 @@ def test_process_import_request__api_error__expected_status(
...
+    # When
+    with pytest.raises(RuntimeError):
+        process_import_request(import_request)
+
+    # Then
+    assert import_request.completed_at
+    assert import_request.status["result"] == "failure"
+    assert not Environment.objects.filter(project=project).exists()
+    assert not Feature.objects.filter(project=project).exists()
+    assert not Segment.objects.filter(project=project).exists()
```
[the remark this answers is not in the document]
emyller: 06f13bb900ec1a7420d60ae2a6b69ebe99e8e4f0

openapi.yaml
```diff
@@ -18537,6 +18537,7 @@ components:
             - 'null'
         project:
           type: integer
+          readOnly: true
```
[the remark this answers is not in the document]
emyller: Are you sure? 🤔

api/segments/migrations/0032_add_segment_rules_data.py
```diff
@@ -0,0 +1,108 @@
...
+                "description": condition.description,
+            }
+            for condition in (
+                rule.live_conditions
+                if hasattr(rule, "live_conditions")  # We only prefetch two levels deep
+                else rule.conditions.filter(deleted_at__isnull=True)
+            )
+        ],
+    }
+    # Top-level rules always carry "rules"; nested ones only for legacy deeper nesting.
```
khvn26: question: Have we checked SaaS for top-level conditions?

api/segments/serializers.py
```diff
@@ -166,6 +191,102 @@ def update(self, segment: Segment, validated_data: dict[str, Any]):  # type: ign
...
+            for nested_rule in rule.get("rules", []):
+                if nested_rule.get("rules"):
+                    raise ValidationError(
+                        {
+                            "segment": [
+                                f"Rules must not be nested more than "
+                                f"{SEGMENT_RULES_MAX_DEPTH} levels deep."
+                            ]
+                        }
+                    )
```
khvn26: nit: for extra class, can we make the loop depend on `SEGMENT_RULES_MAX_DEPTH`

api/segments/types.py
```diff
@@ -1,5 +1,46 @@
...
+    type: RuleType
+    conditions: list[LegacySegmentCondition]
+
+
+class LegacyNestedSegmentRule(_BaseLegacySegmentRule):
+    pass
+
+
+class LegacySegmentRule(_BaseLegacySegmentRule):
+    rules: list[LegacyNestedSegmentRule]
```
khvn26: question: Do we expect to keep these after https://github.com/Flagsmith/flagsmith/issues/7818?

api/segments/serializers.py
```diff
@@ -166,6 +191,102 @@ def update(self, segment: Segment, validated_data: dict[str, Any]):  # type: ign
         enqueue_membership_refresh(segment.project)
         return segment
 
+    def _validate_rules_depth(self, rules: list[LegacySegmentRule]) -> None:
+        # Raise loudly because the interface ignores rules nested too deep
```
khvn26: nit: I don't quite understand this comment here.

api/segments/serializers.py
```diff
@@ -166,6 +191,102 @@ def update(self, segment: Segment, validated_data: dict[str, Any]):  # type: ign
...
+                            "segment": [
+                                f"Rules must not be nested more than "
+                                f"{SEGMENT_RULES_MAX_DEPTH} levels deep."
+                            ]
+                        }
+                    )
+
+    def _validate_rules_condition_count(self, rules: list[LegacySegmentRule]) -> None:
+        if self._can_segment_own_more_conditions_than_limit():
+            return
```
khvn26: nit: can be moved under the if condition

api/segments/serializers.py
```diff
@@ -166,6 +191,102 @@ def update(self, segment: Segment, validated_data: dict[str, Any]):  # type: ign
...
+        """Set the .rules_data attribute
+        TODO: Delete this as per https://github.com/Flagsmith/flagsmith/issues/7818
+        """
+        if "rules" not in validated_data:
+            return  # PATCH support
+        validated_data["rules_data"] = self._cleanup_rules_and_conditions(
+            validated_data["rules"]
+        )
+
+    def _cleanup_rules_and_conditions(
```
khvn26: I feel weird about this method:

1. Its name implies mutation, but it doesn't change anything in-place (thank God!)
1. Its docstring is strangely specific about which data it's supposed to remove, which, to me, reads like typical Claude deixis/redundancy.
1. I have 0 idea where the input type comes from. Its sole call site exists in an untyped context.

If we're supposed to remove this method entirely in #7818, I think this comment can be safely ignored.

khvn26 (commented):
Looks quite good to me! Just a handful nits/questions to clear up, and I think we can 🚢

api/segments/migrations/0032_add_segment_rules_data.py
emyller: Yeah, a few active orgs have conditions in the top-level rule. Our types and migration should cater for that, despite my worry about the UI representation being incorrect — also for historical segments having more than two layers of rules.

khvn26: In that case I don't quite see the purpose of the comment here.

api/segments/types.py
emyller: Yup, because we still need to support `id` and `delete`.

api/segments/serializers.py
emyller: Improved in 96daa3390472fa05894dd45b594b5187ee0683e8

api/segments/serializers.py
emyller: Should be improved in 96daa3390472fa05894dd45b594b5187ee0683e8.

api/segments/types.py
khvn26: Hard disagree on `id` — dropping support for it is IMO inconsequential.

For `delete`, perhaps we should for courtesy reasons, but after https://github.com/Flagsmith/flagsmith/issues/7819 I'd love us to deprecate it entirely.

emyller: Are you sure you want to respond `400 Bad Request` if we can just ignore `id`? I think both are immediately **deprecated**, and should be deleted in our next major version.

Sorry I don't follow if there's any action here.

khvn26: We seem to ignore `id` just as we ignore any extra fields, so I don't see why keep documenting `id` specifically unless we aim to keep it in the API schema.

> Sorry I don't follow if there's any action here.

I would expect either a TODO if we're to remove these, or a tighter relationship with the API schema — our drf-spectacular can generate one from TypedDicts directly. The latter would also solve point 3 from [my other comment](https://github.com/Flagsmith/flagsmith/pull/8245#discussion_r3767725655).

api/segments/migrations/0032_add_segment_rules_data.py
emyller: This was to try and explain the `if` condition right below, because it's not obvious, but allowed `rules_data` to always match the `SegmentRule` type.

In c80e05fcee3afb884a5391d7666837b4bdd434d0, we drop `"rules"` when it's empty even if they're at top-level, assuming that segments can survive with only a top-level rule having conditions.

We discussed offline that, in the future, we should lax the rules schema and fix the frontend so it can manage `rules_data` structures it still can't. Filed in https://github.com/Flagsmith/flagsmith/issues/8280.

api/segments/types.py
emyller: > We seem to ignore `id` just as we ignore any extra fields

Good point.

> I would expect either a TODO if we're to remove these, or a tighter relationship with the API schema — our drf-spectacular can generate one from TypedDicts directly. The latter would also solve point 3 from [my other comment](https://github.com/Flagsmith/flagsmith/pull/8245#discussion_r3767725655).

I put some thought to this, which ended up in 424e836454d5fd908829a6ca8b38d2069a69c8d9.

The added `@extend_schema_field` plays no effect because, thanks to the combination of custom serialiser machinery and `many=True`, it's nullified. Once we lose `WritableNestedModelSerializer` in #7818, and the TODO is addressed, the new `rules` field should pick up in the OpenAI schema — it unfortunately won't pick it up from the model field definition yet, tested.

api/segments/serializers.py
emyller: Agreed! Improved in c6e9507c6d57015b173782b22455b94f3bf585d6 — also this is not a nit in my book: we save one database call for most requests now. Thank you!

api/segments/serializers.py
emyller: > I feel weird about this method:
> 
> 1. Its name implies mutation, but it doesn't change anything in-place (thank God!)

Amen! Please see if 30ee857bd6ebde25df73145c7de8a935b4569dc7 improves that.

> 2. Its docstring is strangely specific about which data it's supposed to remove, which, to me, reads like typical Claude deixis/redundancy.

Between being complimented and offended, I'll pick the one that makes me happiest tonight — I wrote it in the keyboard, using my biological finger tips. 😆

Jokes apart, I do understand the sentiment. No, I wrote it intentionally specific to the things we're dismissing, because they wouldn't really change in 2.x timeline I think.

Regardless, I hope you like the refactor in the above commit.

> 3. I have 0 idea where the input type comes from. Its sole call site exists in an untyped context.

I hope this was improved in 424e836454d5fd908829a6ca8b38d2069a69c8d9.

> 
> If we're supposed to remove this method entirely in #7818, I think this comment can be safely ignored.

Nope, this survives. The second portion of the docstring aims to make that clear:

```
In https://github.com/Flagsmith/flagsmith/issues/7814, we moved from a
SegmentRule and Condition tree to a JSON field. This cleanup exists to
keep the interface compatible.
```

khvn26: If this survives, I'd convert it to a mapper... but we'll have a chance to revisit this in #7818, so not a blocker by any means. Resolving.

khvn26 (approved):
LGTM

Issue #8074, opened 2026-07-22
Scale-Up revival left outdated plan checks and plan descriptions behind


emyller:
Scale-Up left the self-serve payment options in September 2024 (https://github.com/Flagsmith/flagsmith/pull/4583) and returned as v4 in April 2026 (https://github.com/Flagsmith/flagsmith/pull/7216). The v4 Chargebee plan ids are [mixed-case](https://github.com/Flagsmith/flagsmith/blob/46ae7a06bc52dca1582faf623d3b97fb7f8b6f6c/frontend/env/project_prod.js#L33-L36) (`Scale-Up-v4-USD-Monthly` / `Scale-Up-v4-USD-Yearly`) while all earlier plan ids were lowercase, and some plan checks and in-app plan descriptions still assume the old ids or the reduced plan lineup.

## Acceptance criteria

- [ ] Plan id checks match case-insensitively — Chargebee plan ids cannot be renamed, and v4 introduced mixed-case ids:
    - [`get_scaleup_plan_version`](https://github.com/Flagsmith/flagsmith/blob/46ae7a06bc52dca1582faf623d3b97fb7f8b6f6c/api/organisations/models.py#L314-L317) resolves the version from the v4 plan ids.
    - [Plan-gated project fields](https://github.com/Flagsmith/flagsmith/blob/46ae7a06bc52dca1582faf623d3b97fb7f8b6f6c/api/projects/serializers.py#L74-L76) treat Scale-Up v4 as Scale-Up.
    - [Stale flags tagging](https://github.com/Flagsmith/flagsmith-private/blob/139aa16579c6b1169f1205dc3bb7a04d2333f5c3/src/workflows_logic/stale_flags/tasks.py#L20-L26) includes Scale-Up v4 organisations.
- [ ] The plan descriptions on the in-app billing page agree with the pricing page on which plan includes governance features — today the [Enterprise panel](https://github.com/Flagsmith/flagsmith/blob/46ae7a06bc52dca1582faf623d3b97fb7f8b6f6c/frontend/web/components/modals/payment/pricingFeatures.tsx#L91) lists roles, permissions, change requests, and audit logs that Scale-Up already includes.

afsuyadi:
Hi @emyller , am I able to work on this issue?

PR #8249, opened 2026-08-10
fix: match Scale-Up plan ids case-insensitively

Holmus:
## Summary

Scale-Up returned as v4 with mixed-case Chargebee plan ids (`Scale-Up-v4-USD-Yearly` / `Scale-Up-v4-USD-Monthly`), but two plan-id checks still match case-sensitively:

- `Subscription.get_scaleup_plan_version()` fails to match the v4 ids and falls back to version 1, so v4 Scale-Up organisations are treated as pre-v4 and keep **unlimited audit log visibility** (`audit_log_visibility_days = None`).
- `ReadOnlyIfNotValidPlanMixin` compiles `invalid_plans_regex` case-sensitively, so `ProjectCreateSerializer`'s `scale-up.*` gate misses v4 plans and the plan-gated project fields (`stale_flags_limit_days`, `enable_realtime_updates`) become **writable for Scale-Up v4 organisations**.

Both now match case-insensitively. Chargebee plan ids cannot be renamed, and `SubscriptionPlanFamily.get_by_plan_id` already normalises its input for exactly this reason — this brings the remaining two checks in line.

Part of #8074 (the two backend acceptance criteria in this repo; the stale-flags tagging item lives in flagsmith-private and the billing-page plan descriptions are a separate frontend change).

## Test plan

- Extended `test_get_scaleup_plan_version__given_plan__returns_expected` with the production v4 plan ids — fails without the fix.
- Extended the `ReadOnlyIfNotValidPlanMixin` read-only parametrisation with a mixed-case plan id against the production regex from `ProjectCreateSerializer`.
- Ran locally: 9/9 plan-version tests and 5/5 mixin tests pass.

### DIFF
--- a/api/organisations/models.py
+++ b/api/organisations/models.py
@@ -312,7 +312,7 @@ class Subscription(LifecycleModelMixin, SoftDeleteExportableModel):  # type: ign
         return self.subscription_plan_family == SubscriptionPlanFamily.ENTERPRISE
 
     def get_scaleup_plan_version(self) -> int:
-        if match := re.match(r"scale-up-v(\d+)", self.plan or ""):
+        if match := re.match(r"scale-up-v(\d+)", self.plan or "", re.IGNORECASE):
             return int(match.group(1))
         return 1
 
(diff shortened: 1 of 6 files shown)


Zaimwa9 (approved):
Thanks!

PR #8251, opened 2026-08-10
fix: experiment form UX fixes

Zaimwa9:

## Changes

Small experiment UX fixes:

- Refetch multivariate flags when landing on the experiment setup step, so flags made multivariate in another tab appear without a full refresh.
- Keep the typed event name when the metric event combobox loses focus. The reactstrap modal wrapper steals focus on any click inside it, which made react-select discard the draft; the value is now committed on blur.
- Winning variation stat now renders in the default text colour, matching the other result cards.
- Removed a redundant `ms-lg-2` on the `PageTitle` CTA that indented the button when it wraps below the description (e.g. Segments page).

Manually.

image

image

### DIFF
--- a/frontend/web/components/PageTitle.tsx
+++ b/frontend/web/components/PageTitle.tsx
@@ -20,7 +20,7 @@ const PageTitle: FC<PageTitleType> = ({ children, className, cta, title }) => {
             </Row>
           )}
         </div>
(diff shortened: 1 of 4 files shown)


talissoncosta (approved):
Looks good!

PR #8273, opened 2026-08-12
fix(experimentation): retry experiment computes on transient ClickHouse errors

Zaimwa9:

## Changes

Contributes to https://github.com/Flagsmith/flagsmith/issues/8274

Experiment computes fail with `TimeoutError` when a ClickHouse read exceeds the 30s `send_receive_timeout`, and the failure was swallowed, so refreshes silently produced nothing.

- Raise `TaskBackoffError` on `OSError` in `compute_experiment_exposures`/`compute_experiment_results` so the task processor retries.
- Use a 120s timeout for background warehouse queries (interactive paths keep 30s), with a matching 3-minute task timeout. One client is cached per timeout.

Unit tests covering the retry behaviour, per-timeout client caching, and background timeout usage; full experimentation suite, lint, and mypy pass.


api/experimentation/services.py
```diff
@@ -148,19 +149,30 @@ def is_experiment_feature_enabled(organisation: Organisation) -> bool:
...
-    kwargs.setdefault("client_name", settings.CLICKHOUSE_CONNECTION_CLIENT_NAME)
-    return Client(host, **kwargs)
+    clients: dict[int, Client] = getattr(_clickhouse_clients, "clients", None) or {}
+    _clickhouse_clients.clients = clients
+    if (client := clients.get(send_receive_timeout)) is None:
+        host, kwargs = parse_url(settings.EXPERIMENTATION_CLICKHOUSE_URL)
+        kwargs.setdefault("connect_timeout", CLICKHOUSE_CONNECT_TIMEOUT_SECONDS)
+        kwargs.setdefault("send_receive_timeout", send_receive_timeout)
+        kwargs.setdefault("client_name", settings.CLICKHOUSE_CONNECTION_CLIENT_NAME)
+        client = clients[send_receive_timeout] = Client(host, **kwargs)
```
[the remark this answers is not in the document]
Zaimwa9: Intentional. the DSN overriding code defaults is a documented operator escape hatch (pre-existing this PR btw), and our DSN doesn't set send_receive_timeout. 
duplicated cache client could be addressed in a follow-up

api/experimentation/services.py
```diff
@@ -148,19 +149,30 @@ def is_experiment_feature_enabled(organisation: Organisation) -> bool:
     )
 
 
-@lru_cache(maxsize=1)
-def _get_clickhouse_client() -> Client:
+_clickhouse_clients = threading.local()
```
khvn26: Can you describe what is the exact problem we're solving by using thread-local storage? I.e. are there leaking connections? Should other clickhouse-driver consumers accommodate?

api/experimentation/services.py
```diff
@@ -148,19 +149,30 @@ def is_experiment_feature_enabled(organisation: Organisation) -> bool:
...
-    host, kwargs = parse_url(settings.EXPERIMENTATION_CLICKHOUSE_URL)
-    kwargs.setdefault("connect_timeout", CLICKHOUSE_CONNECT_TIMEOUT_SECONDS)
-    kwargs.setdefault("send_receive_timeout", CLICKHOUSE_QUERY_TIMEOUT_SECONDS)
-    kwargs.setdefault("client_name", settings.CLICKHOUSE_CONNECTION_CLIENT_NAME)
-    return Client(host, **kwargs)
+    clients: dict[int, Client] = getattr(_clickhouse_clients, "clients", None) or {}
+    _clickhouse_clients.clients = clients
+    if (client := clients.get(send_receive_timeout)) is None:
+        host, kwargs = parse_url(settings.EXPERIMENTATION_CLICKHOUSE_URL)
+        kwargs.setdefault("connect_timeout", CLICKHOUSE_CONNECT_TIMEOUT_SECONDS)
```
khvn26: I don't think connect_timeout realistically should matter here? Even when a DWH is idle, the cloud's proxy still is expected to handle the connection in time?

api/experimentation/services.py
```diff
@@ -344,7 +356,9 @@ def get_exposure_buckets(
     window_end: datetime,
     granularity: ExposureGranularity,
 ) -> list[ExposureBucket]:
-    rows = _get_clickhouse_client().execute(
+    rows = _get_clickhouse_client(
+        send_receive_timeout=CLICKHOUSE_BACKGROUND_QUERY_TIMEOUT_SECONDS,
```
khvn26, line 354: So, just to clear up my own understanding: we have "background" queries, ran by task processor, and "live" queries, driven from API/UI. Both are affected by sleeping warehouses taking time to wake up, but we want the "live" queries to fail faster to make the UI more responsive.

I guess my question is, are we 100% clear on where the snappy live query requirement comes from? Don't we want a higher query success rate at the expense of latencies, and accommodate the UX to higher latencies instead?

api/experimentation/tasks.py
```diff
@@ -28,6 +29,11 @@
     deliver_warehouse_events,
 )
 
+# Warehouse computes may legitimately run for up to
+# `CLICKHOUSE_BACKGROUND_QUERY_TIMEOUT_SECONDS`, e.g. while an idled
+# ClickHouse Cloud service wakes up.
```
khvn26: I don't think I quite understand this comment. 3 minutes is more than `CLICKHOUSE_BACKGROUND_QUERY_TIMEOUT_SECONDS`, which is currently at 2 minutes.

khvn26 (commented):
Questions.

api/experimentation/tasks.py
Zaimwa9: Yep I removed this comment just before your review (`feat: remove slop)`, it's not the case and also the prod clickhouse shouldn't be sleeping

api/experimentation/services.py
Zaimwa9: That's more of an opportunistic improvement. The `lru_cache` singleton was shared across threads within each process and the client isn't thread safe. Now each thread gets its own client and we keep one client per thread per timeout max

api/experimentation/services.py
Zaimwa9: No it wasn't the problem, but this one was pre-existing to the PR, the fix is only around `send_receive_timout`

api/experimentation/services.py
khvn26: > the client isn't thread safe

Do we know how this affects us specifically? I haven't encountered thread-related issues on the segment membership side of things yet.

I'd like to do a deeper dive on that (mainly to see if we can improve anything in clickhouse-driver) — can we factor this out in its own issue/PR?

api/experimentation/services.py
Zaimwa9: We don't have a snappy live query requirements. So, at first all warehouses could go idle. We reconfigured prod to be always awake.
This defualt is inherited from this time and I don't think we'd gain anything raising the timeouts. The live queries are synchronous requests and we don't have success rate problems there (besides the previous idle connection).

To be honest, that's my understanding of your question, so let me know if you needed another answer

khvn26: What's the rationale behind having distinct `CLICKHOUSE_BACKGROUND_QUERY_TIMEOUT_SECONDS` and `CLICKHOUSE_QUERY_TIMEOUT_SECONDS`, then?

Zaimwa9: They are not comparable in terms of computation. The UI ones are direct http requests with simple compositions (`select distinct count`; `select count`) so there's no need to allocate a high timeout (> 30s would be more of a transient or permanent issue). This distinct constant comes from the first idle scenarios.

While the background ones are full aggregations in the task processor where a longer wait is legitimate and where we can easily retry.

If there is a consolidation or cleanup to do it would be out of the scope of this fix

api/experimentation/services.py
Zaimwa9: reverting this then

khvn26 (dismissed):

khvn26 (approved):

PR #8279, opened 2026-08-12
fix(identities): multivariate override editor hides the identity's override value

### LINKED ISSUE #8271: Multivariate override editor hides the identity's override value (shows the flag's control value instead)
## Description

When an identity has a single-value override on a feature that is later made multivariate, the "Edit User Feature" modal stops showing the identity's override value. In multivariate mode the editor renders only the flag's options as radios (control + each variant), so an override whose value is not one of those options has nowhere to appear. The control row shows the _environment's_ control value, so the identity looks like it is on the default.

The override is intact and the SDK serves it correctly, so this is a display issue in the override editor. If the user saves that modal, the displayed control value is written over the override, so the value can be lost on save.

## Steps to reproduce

1. Create a standard flag with control value `ENV_DEFAULT`.
2. Set an identity override on it to `MY_OVERRIDE`.
3. On the Features page, make the feature multivariate: update the control value and add a variant `VARIANT_A`.
4. Open that identity and open the "Edit User Feature" modal for the flag.

## Expected

The editor shows the identity's override value (`MY_OVERRIDE`).

## Actual

The editor shows the flag's control value selected. `MY_OVERRIDE` is not shown anywhere in the modal. The identities list still shows `MY_OVERRIDE`, and the SDK still serves it.

## Screenshots

- **Identities list (correct):** the feature row shows the override `MY_OVERRIDE`.
- **Edit User Feature modal (the bug):** only `ENV_DEFAULT` (selected) and `VARIANT_A` are shown; `MY_OVERRIDE` is absent.
- **Features page:** the flag's control value and variation as configured.

## Impact

Display only, no data lost while viewing. Risk: saving the modal overwrites the override with the displayed control value.

## Root cause (initial)

`frontend/web/components/modals/create-feature/tabs/FeatureValueTab.tsx` passes the environment control value as `controlValue` to `VariationOptions` in select mode (changed in #6585, which fixed #6582). The multivariate override editor only models "control or a variant", so a free-form override value has no representation.

## Proposed behaviour

Show the identity's own override value in the editor, **read-only**, with a warning like _"This identity override contains a value that is not one of this flag's variations. We recommend changing it."_, and offer the variations for selection so the user can migrate it. Guard against a save silently overwriting the override.

## References

- Confirmed on staging.
- Related: #4599, #5263, #6582, #6585, #7254

bardock-2393:
When a flag with an existing identity override is later made multivariate, the _Edit User Feature_ modal stops showing what that identity is actually being served. The editor offers only the environment's control value and each variation as radios, so an override holding anything else has nowhere to appear — the control row reads as selected and the identity looks like it is on the environment default. The override is intact and the SDK keeps serving it, but saving the modal replaced it with the control value, so the value could be lost by simply opening and saving.

Rather than hiding it, the editor now shows the override's own value as a read-only, already-selected row alongside the variations, and warns that the value is not one of them so the user can move the identity onto a variation deliberately. Saving keeps the value unless the user picks the control or a variation instead.

## Changes

- [x] The _Edit User Feature_ modal shows an identity's override value even when it is not one of the flag's variations.
- [x] A warning explains that the value is not a variation, and recommends changing it.
- [x] Saving the modal no longer replaces such an override with the environment's control value.
- [x] Tests covering when an override counts as unrepresentable by the variation radios.

Closes https://github.com/Flagsmith/flagsmith/issues/8271

Review effort: 2/5

Manually, end to end, against a local API and dashboard, following the reproduction steps in the issue: a flag with control value `ENV_DEFAULT`, an identity override of `MY_OVERRIDE`, then a variation `VARIANT_A` added to make the flag multivariate.

- Opening _Edit User Feature_ for that identity shows the warning, `MY_OVERRIDE` selected and read-only, `ENV_DEFAULT` and `VARIANT_A` unselected. On `main` the same modal shows `ENV_DEFAULT` selected and `MY_OVERRIDE` nowhere.
- Pressing _Update Feature_ without changing the selection leaves the stored value as `MY_OVERRIDE` (confirmed directly in the database). On `main` this is where the override was replaced with the control value.

Automated: the rule deciding whether an override is representable by the variation radios is unit tested in `common/utils/__tests__/multivariate.test.ts`, including the plan-unchanged case, an identity assigned a variation, a partially weighted override, and null/undefined/empty edge cases. `npm run test:unit` passes in full (402 tests), `npm run lint` reports nothing on the changed files, and `npm run typecheck` produces an identical error set to `main` for them.

## A note on behaviour

The same guard applies when a multivariate identity override's stored value has drifted from the environment's control value for any other reason — for example the environment control being edited after the override was made. Previously a save silently re-synced the identity onto the new control value, changing what that identity is served; now the value is shown, flagged, and left alone until the user chooses. That seemed the safer default given the issue is about a value disappearing, but say the word if you would rather keep re-syncing in that case.

### DIFF
--- a/frontend/common/utils/__tests__/multivariate.test.ts
+++ b/frontend/common/utils/__tests__/multivariate.test.ts
@@ -95,6 +95,26 @@ describe('multivariate', () => {
       ).toBe(true)
     })
 
+    // An unloaded feature state must not read as an override: the caller
+    // latches this answer, so a verdict from absent data would stick.
+    it.each`
+      controlValue     | overrideValue    | scenario
+      ${'ENV_DEFAULT'} | ${undefined}     | ${'the override has not loaded'}
+      ${undefined}     | ${'MY_OVERRIDE'} | ${'the control value has not loaded'}
+      ${undefined}     | ${undefined}     | ${'neither has loaded'}
+    `(
+      'withholds a verdict while $scenario',
+      ({ controlValue, overrideValue }) => {
+        expect(
+          hasUnmatchedIdentityOverride({
+            controlValue,
+            overrideValue,
+            variationOverrides: [],
+          }),
+        ).toBe(false)
+      },
+    )
+
     it.each`
       controlValue | overrideValue | expected
       ${null}      | ${undefined}  | ${false}
@@ -169,6 +189,16 @@ describe('multivariate', () => {
           overrideValue: 'MY_OVERRIDE',
         }),
       ).toEqual({ selected: true, value: 'MY_OVERRIDE' })
+    })
+
+    it('does not latch while the override value is unavailable', () => {
+      expect(
+        resolveUnmatchedOverride({
+          isSelected: true,
+          latchedValue: undefined,
+          overrideValue: undefined,
+        }),
+      ).toBeUndefined()
     })
 
     // `null` is a real override value, so it has to latch like any other —
--- a/frontend/common/utils/__tests__/multivariate.test.ts
+++ b/frontend/common/utils/__tests__/multivariate.test.ts
@@ -1,5 +1,6 @@
 import {
   getDefaultVariantKey,
+  hasUnmatchedIdentityOverride,
   sortMultivariateOptions,
 } from 'common/utils/multivariate'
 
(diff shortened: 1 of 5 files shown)


Holmus:
Hey @bardock-2393, looks like there are some automated review comments to address.

Thanks for your contribution!

bardock-2393:
I will resolved it today

kyle-ssg:
hey @bardock-2393 thanks for raising this use case, what you've done makes sense however I think one thing could do with fixing:

image

Here as soon as I click the second option the first (the identity override) vanishes. I think it should only be removed once the feature has been saved.

Thanks!

bardock-2393:
@kyle-ssg Thanks — resolved in 4ee1e53f1. The row was rendered off the same state as its selected status, so picking a variation unmounted it. It now stays listed until save and just deselects, and clicking it restores the original value. Verified against the component directly; I couldn't run a full instance locally, so worth a quick check on your end.

talissoncosta (approved):
Thanks for sticking with this through several rounds, and for the detailed write up.

Kyle's point looks addressed. The row stays listed and just deselects, and reselecting it restores the value.

One small thing left over, not worth another round from you: the warning is gated on `unmatchedOverrideSelected` rather than `unmatchedOverride`, so it disappears when the row deselects and leaves that value unlabelled. I'll pick it up in a follow-up.

Merging, thanks again.

PR #8281, opened 2026-08-13
fix(OAuth): `flagsmith login` never completes for a new signup

### LINKED ISSUE #8165: `flagsmith login` experience is broken for signups

- [ ] Self Hosted with Docker
- [x] SaaS at flagsmith.com

### Describe the bug

Early CLI testers report the following defects when running `flagsmith init`/`flagsmith login` without having a registered account at Flagsmith:
1. CLI OAuth callback waiting indefinitely
2. Infinite spinner when signing up via SSO (e.g. Google)

I suspect this applies to Flagsmith MCP as well.

### Steps To Reproduce

1. Log out Flagsmith
2. `curl -fsSL https://raw.githubusercontent.com/Flagsmith/flagsmith-cli/main/install.sh | sh`
3. `flagsmith init`
4. Instead of logging in, sign up a new account
5. Repeat the above and sign up via Google SSO to witness the infinite spinner

### Expected behavior

Getting redirected to OAuth consent screen and finishing CLI login successfully for all authentication methods (email/password, Google, Github, SAML, LDAP).

We should also make sure that onboarding is not getting skipped by this flow.

### Screenshots

_No response_

khvn26:

## Changes

Closes #8165.

In this PR, we:

1. Route OAuth-driven signups to onboarding.
1. Redirect to OAuth consent screen once onboarding (both control and single-page) has provisioned the organisation.
1. Stop two things moving the browser out from under that request.
1. Serve the consent screen's scope descriptions from the API.

Added unit tests.

Manually, against a local API and dashboard, with `flagsmith-cli` as the client and `scope=admin-api`. Open `/oauth/authorize/?client_id=flagsmith-cli&response_type=code&redirect_uri=http%3A%2F%2F127.0.0.1%2Fcallback&code_challenge=…&code_challenge_method=S256&scope=admin-api&state=x` while logged out, then sign up.

Will test the Google path in production.

### DIFF
--- a/api/oauth2_metadata/constants.py
+++ b/api/oauth2_metadata/constants.py
@@ -6,3 +6,20 @@ SCOPE_ADMIN_API = "admin-api"
 FIRST_PARTY_CLIENT_IDS = frozenset({FLAGSMITH_CLI_CLIENT_ID})
 FIRST_PARTY_SCOPES = frozenset({SCOPE_ADMIN_API})
 THIRD_PARTY_SCOPES = frozenset({SCOPE_MCP})
+
+SCOPE_GRANTS: dict[str, tuple[str, ...]] = {
+    SCOPE_MCP: (
+        "Manage feature flags, toggle states, and update values",
+        "Create and manage audience targeting segments",
+        "View and configure environments",
+        "View and update project settings",
+        "Create and review change requests",
+        "View organisation details, roles, and groups",
+        "Act with your own permissions, in every organisation you belong to",
+    ),
+    SCOPE_ADMIN_API: (
+        "Act with your own permissions, in every organisation you belong to",
+        "Manage feature flags, segments, environments and projects you can access",
+        "Manage organisation settings, members and roles you can access",
+    ),
+}
--- a/api/tests/unit/oauth2_metadata/test_authorize_view.py
+++ b/api/tests/unit/oauth2_metadata/test_authorize_view.py
@@ -9,7 +9,12 @@ from oauth2_provider.models import Application
 from rest_framework import status
 from rest_framework.test import APIClient
 
-from oauth2_metadata.constants import FLAGSMITH_CLI_CLIENT_ID
+from oauth2_metadata.constants import (
+    FLAGSMITH_CLI_CLIENT_ID,
+    SCOPE_ADMIN_API,
+    SCOPE_GRANTS,
+    SCOPE_MCP,
+)
 
 
 def _pkce_pair() -> tuple[str, str]:
@@ -82,7 +87,10 @@ def test_get__valid_params__returns_application_info(
     data = response.json()
     assert data["application"]["name"] == "Test App"
     assert data["application"]["client_id"] == oauth_application.client_id
-    assert "mcp" in data["scopes"]
+    assert data["scopes"]["mcp"] == {
+        "label": "MCP access",
+        "grants": list(SCOPE_GRANTS[SCOPE_MCP]),
+    }
     assert data["redirect_uri"] == "https://example.com/callback"
     assert data["is_verified"] is False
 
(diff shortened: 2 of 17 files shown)


frontend/common/utils/pendingAuthorisation.ts
```diff
@@ -0,0 +1,18 @@
...
+export const AUTHORISE_PATH = '/oauth/authorize'
+
+/**
+ * Whether a stored post-login redirect is a consent request waiting to be
+ * answered. Compares the whole path so an absolute URL never matches: the
+ * redirect is read from a cookie, and this decides where we send the browser.
+ */
+export const isPendingAuthorisation = (
+  redirect?: string | null,
+): redirect is string => {
```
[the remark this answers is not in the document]
khvn26: Took the cookie one — `CookieExpiry` now says what the two forms mean, in ebb86f19b.

Left the other: `redirect?: string | null` is just an optional string, not a set of allowed values, so naming it wouldn't tell a reader anything the type guard doesn't.

api/oauth2_metadata/constants.py
```diff
@@ -6,3 +6,20 @@
...
+        "Create and review change requests",
+        "View organisation details, roles, and groups",
+        "Act with your own permissions, in every organisation you belong to",
+    ),
+    SCOPE_ADMIN_API: (
+        "Act with your own permissions, in every organisation you belong to",
+        "Manage feature flags, segments, environments and projects you can access",
+        "Manage organisation settings, members and roles you can access",
+    ),
+}
```
emyller, line 25:
```suggestion
SCOPE_GRANTS: dict[str, frozenset[str, ...]] = {
    SCOPE_MCP: frozenset([
        "Manage feature flags, toggle states, and update values",
        "Create and manage audience targeting segments",
        "View and configure environments",
        "View and update project settings",
        "Create and review change requests",
        "View organisation details, roles, and groups",
        "Act with your own permissions, in every organisation you belong to",
    ]),
    SCOPE_ADMIN_API: frozenset([
        "Act with your own permissions, in every organisation you belong to",
        "Manage feature flags, segments, environments and projects you can access",
        "Manage organisation settings, members and roles you can access",
    ]),
}
```

nit: The immutability of tuples does not justify using them as _immutable collections_. Please, if you agree, prefer to use tuples when they could be interchanged with their verbose brother, `NamedTuple`.

api/tests/unit/oauth2_metadata/test_authorize_view.py
```diff
@@ -82,7 +87,10 @@ def test_get__valid_params__returns_application_info(
     data = response.json()
     assert data["application"]["name"] == "Test App"
     assert data["application"]["client_id"] == oauth_application.client_id
-    assert "mcp" in data["scopes"]
+    assert data["scopes"]["mcp"] == {
+        "label": "MCP access",
+        "grants": list(SCOPE_GRANTS[SCOPE_MCP]),
+    }
```
emyller, line 93: In here, and in other occurrences in this patch, can we hardcode the expected list of grants? I'd expect any change in it to break relevant tests, and the actual grants to be verified in tests.

emyller (changes requested):
Backend changes look good, bar one test tweak.

api/tests/unit/oauth2_metadata/test_authorize_view.py
khvn26: I would agree, but this is all temporary anyway until #7035 lands.

emyller: Got it. Consider adding TODOs if you think these could have any chance slipping through eternity.

api/oauth2_metadata/constants.py
khvn26: It's hard to justify `frozenset` here considering we don't need O(1) membership checks here; and using a mutable type like `list` makes things geniunely worse in my opinion.

I'm sorry, I'd rather you come to terms with the fact that `tuple[str, ...]` is a legitimate type.

emyller: > I'm sorry, I'd rather you come to terms with the fact that `tuple[str, ...]` is a legitimate type.

I continue to disagree, and I'd be glad to argument, though this isn't worth it. I will try not to pick these from now on.

emyller (approved):
Approved from **backend** standpoint.

talissoncosta (approved):
Looks good on FE side! 💯

PR #8283, opened 2026-08-13
feat: create segment from CSV drawer

Zaimwa9:

## Changes

Adds the "New Segment — from a CSV list" drawer behind the segment sources modal, and makes the sources modal config-driven instead of hardcoded.

**Flag-driven sources.** The modal now reads its cards from the JSON value of the `create_segment_with_external_sources` flag:

```json
[
  { "active": true, "name": "csv", "visible": true },
  { "active": false, "name": "amplitude", "visible": true },
  { "active": false, "name": "mixpanel", "visible": true },
  { "active": false, "name": "adobe_journey_manager", "visible": true }
]
```

`visible: false` hides a card, `active: false` renders it as a Beta fake door, `active: true` makes it real (currently only CSV has an implementation — an active source without one safely stays a fake door). Array order controls display order. Copy and icons stay in the frontend. Flag enabled with an empty or invalid value falls back to the manual drawer.

**CSV drawer.**

- Client-side CSV parsing (quoted fields, CRLF, BOM)
- Optional header row
- Identifier column selection with a highlighted 5-row preview
- Ignored-row warnings (empty/duplicate) and a blocking error when a column yields no identifiers
- Submission is wired in the follow-up PR

Unit tests for the CSV parser/extractor (`common/utils/__tests__/csv.test.ts`). Manually:

1. Set the JSON value above on the `create_segment_with_external_sources` flag and enable it.
2. Segments → Create Segment: modal shows the cards; non-active sources show the Beta chip and the request-access flow.
3. Click "From a CSV list" (active): the drawer opens; upload a CSV with/without headers, switch the identifier column, and check the preview, warnings, and the empty-column blocking error.
4. Disable the flag or set the value to `[]`: the button opens the manual drawer directly.

## Screens
**Upload screen**
image

**Uploaded but unselected**
image

**Uploaded selected**
image
image

**Error states**
image
image
image

### DIFF
--- a/frontend/web/components/EnvironmentSelect.tsx
+++ b/frontend/web/components/EnvironmentSelect.tsx
@@ -18,6 +18,7 @@ type EnvironmentSelectType = Partial<Omit<Props, 'value'>> & {
   readOnly?: boolean
   idField?: 'id' | 'api_key'
   ignore?: string[]
+  size?: 'default' | 'select-sm' | 'select-xsm'
   dataTest?: (value: { label: string }) => string
 }
 
@@ -30,6 +31,7 @@ const EnvironmentSelect: FC<EnvironmentSelectType> = ({
   projectId,
   readOnly,
   showAll,
+  size = 'select-xsm',
   value,
   ...rest
 }) => {
@@ -64,7 +66,7 @@ const EnvironmentSelect: FC<EnvironmentSelectType> = ({
     <div data-test={dataTestProp}>
       <Select
         {...rest}
-        className='react-select select-xsm'
+        className={size === 'default' ? 'react-select' : `react-select ${size}`}
         value={
           foundValue
             ? foundValue
--- a/frontend/web/components/EnvironmentSelect.tsx
+++ b/frontend/web/components/EnvironmentSelect.tsx
@@ -66,7 +66,7 @@ const EnvironmentSelect: FC<EnvironmentSelectType> = ({
     <div data-test={dataTestProp}>
       <Select
         {...rest}
-        className={size === 'default' ? 'react-select' : `react-select ${size}`}
+        size={size === 'default' ? undefined : size}
         value={
           foundValue
             ? foundValue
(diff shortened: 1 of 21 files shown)


Zaimwa9:
@themis-blindfold review

frontend/web/components/modals/CreateSegmentFromCsv/CreateSegmentFromCsv.tsx
```diff
@@ -0,0 +1,292 @@
...
+        inputProps={{ className: 'full-width', name: 'featureDesc' }}
+        onChange={(e: React.ChangeEvent<HTMLInputElement>) => {
+          setDescription(Utils.safeParseEventValue(e))
+        }}
+        type='text'
+        title='Description'
+        placeholder="e.g. 'People who have spent over $100' "
+      />
+      <div className='mb-4'>
+        <label>Environment</label>
```
talissoncosta: This one and `Identifier column` on line 154 have no `htmlFor`, so they aren't announced with their control. `FieldLabel` wires that up:

```tsx
   <FieldLabel htmlFor='environment-select'>Environment</FieldLabel>
```

Related: Name on line 94 is a raw label plus `<Flex><Input/></Flex>` while Description right below it uses `InputGroup`. Any reason the two differ?

frontend/web/components/EnvironmentSelect.tsx
```diff
@@ -64,7 +66,7 @@ const EnvironmentSelect: FC<EnvironmentSelectType> = ({
     <div data-test={dataTestProp}>
       <Select
         {...rest}
-        className='react-select select-xsm'
+        className={size === 'default' ? 'react-select' : `react-select ${size}`}
```
talissoncosta: Could you please double check if `Select` already take a `size` prop?

talissoncosta (dismissed):
Tested locally, looks pretty good. A few comments inline, nothing blocking.

talissoncosta (approved):

PR #8286, opened 2026-08-13
feat(onboarding): gradual rollout quest screen

### LINKED ISSUE #8036: Onboarding: gradual rollout fake door next quest
The "Roll out gradually" quest currently links straight to the feature's segment overrides tab. This drops users into the UI without explaining the process, and gradual rollout today requires several manual steps.

Replace it with a dedicated quest screen:

1. **How to do it today**, from [Rollout by Percentage](https://docs.flagsmith.com/managing-flags/rollout/rollout-by-percentage): create a segment with a `% Split` rule, add a segment override on the flag, raise the percentage over time. Framed as "a few manual steps", never as "not possible".
2. **Coming soon**: we're making that one action instead of several. Notify-me CTA.

Draft design:

Image

### Framing

The fake door is about **the simplification, not the capability**. 

Copy rules that follow:

- Never imply you can't roll out gradually today.
- No metrics claim: release pipelines aren't connected to experimentation metrics.
- Call out the prerequisite the docs flag, since it will otherwise silently waste their time: percentage splits only apply when you identify users, and the onboarding snippet doesn't call `flagsmith.identify(...)`.

### Open questions

- What does "soon" refer to: Release Pipelines reaching lower plans, or a simpler flow in core? Release Pipelines is Enterprise-gated (`utils.tsx:554`), so if it's the former the block is an upgrade signal rather than a roadmap tease, and it shouldn't say "coming soon" to an Enterprise user who already has it.

talissoncosta:

## Changes

Closes #8036.

"Gradual rollout" deep-linked straight to the segment overrides tab, which explains none of the steps a rollout takes. It now opens a modal first: the three steps from [Rollout by Percentage](https://docs.flagsmith.com/managing-flags/rollout/rollout-by-percentage), the `flagsmith.identify(...)` prerequisite, and a fake door for doing all three in one action.

"Create a rollout segment" still goes where the card used to. Gated on `onboarding_rollout_quest`, default off, so off is today's behaviour exactly.

"Notify me" records a `rollout_beta_requested` trait, so the asked state survives a reload and the beta can later be handed out by targeting it.

### Screenshot

image

image

- [x] `typecheck` and `lint` clean
- [x] Gradual rollout opens the modal, not the overrides tab
- [x] With `onboarding_rollout_quest` off, it deep-links as before
- [ ] Maybe later and close both return to onboarding
- [ ] Create a rollout segment lands on the same tab the card used to
- [ ] `rollout_beta_requested` carries the right email and organisation, and the trait persists across a reload
- [ ] Light and dark


frontend/web/components/pages/onboarding/OnboardingFlow/OnboardingFlow.tsx
```diff
@@ -149,6 +153,15 @@ const OnboardingFlow: FC = () => {
...
   }
+  const openRolloutQuest = useRolloutQuest({
+    diagnosticIds,
+    featureName,
+    onContinue: () => goToFlagConfig('rollout'),
+    who: { email: profile?.email, organisation: organisationDisplayName },
+  })
+
+  const goToNextStep = (step: OnboardingNextStep) =>
+    step === 'rollout' ? openRolloutQuest() : goToFlagConfig(step)
```
Zaimwa9: Can we gate it with a flag please ?

frontend/web/components/pages/onboarding/OnboardingRolloutQuest/RolloutComingSoonCard.tsx
```diff
@@ -0,0 +1,59 @@
...
+  onFeedback: () => void
+}
+
+// The one thing we don't ship yet: the three steps above in a single action. It
+// promises a simpler flow, never the capability.
+const RolloutComingSoonCard: FC<RolloutComingSoonCardProps> = ({
+  onFeedback,
+  onNotifyMe,
+}) => {
+  const [notified, setNotified] = useState(false)
```
Zaimwa9, line 26: I haven't done it with the sources fake door but I believe setting a trait and using the trait as a persisted state is actually better, especially if we want to use it to access the beta in the future

frontend/web/components/pages/onboarding/OnboardingRolloutQuest/rolloutSteps.tsx
```diff
@@ -0,0 +1,35 @@
...
+    body: 'Raise the percentage as your confidence grows: 10% → 25% → 50% → 100%.',
+    title: 'Increase over time',
+  },
+]
+
+// Sits with the steps rather than in docs: without it the override reaches
+// nobody and the user has no way to tell.
+export const RolloutPrerequisite: FC = () => (
+  <p className='onboarding-rollout-quest__prerequisite text-secondary d-flex gap-2 m-0'>
+    <Icon name='info' width={14} className='icon-secondary' />
```
Zaimwa9: isn't `icon-secondary` dead when combined with `info` that hardcodes the fill ?

frontend/web/components/pages/onboarding/OnboardingRolloutQuest/OnboardingRolloutQuest.scss
```diff
@@ -0,0 +1,24 @@
+.onboarding-rollout-quest {
+  // 1.5 rather than .lh-base, which is 1.375 here.
+  &__body {
+    font-size: 13px;
```
Zaimwa9: don't we have a token we can use ?

frontend/web/components/pages/onboarding/OnboardingRolloutQuest/OnboardingRolloutQuest.scss
```diff
@@ -0,0 +1,24 @@
...
+    border: 1px solid var(--color-border-default);
+  }
+
+  &__step-number {
+    width: 24px;
+    height: 24px;
+  }
+
+  &__prerequisite {
+    font-size: 12px;
```
Zaimwa9: Same, don't we have a token we can use ?

frontend/web/components/pages/onboarding/OnboardingRolloutQuest/trackRolloutInterest.ts
```diff
@@ -0,0 +1,26 @@
...
+// reply to. Sent through flagsmith.trackEvent, same as the segment sources door.
+const trackRolloutInterest = (
+  event: string,
+  { email, organisation }: RolloutInterest,
+): void => {
+  flagsmith.trackEvent(event, {
+    metadata: {
+      email,
+      organisation,
+      source: 'onboarding-rollout-quest',
```
Zaimwa9: NIT: there's a bit of a conflict in the meaning of `source` especially because the segment one is brand new :). Maybe this could be `origin` more than `source`

Zaimwa9 (changes requested):
Overall good thanks with some comments, i think the trait one could be a good improvement

frontend/web/components/pages/onboarding/OnboardingRolloutQuest/rolloutSteps.tsx
talissoncosta: Good catch! No idea how I missed it! 😢

frontend/web/components/pages/onboarding/OnboardingRolloutQuest/trackRolloutInterest.ts
talissoncosta: Fair, origin in 963519c42

frontend/web/components/pages/onboarding/OnboardingRolloutQuest/OnboardingRolloutQuest.scss
talissoncosta: Yep, they were not exactly matches, but I does not make too much diff. Replacing it with tokens, thanks for raising this. Fixed in 963519c42

frontend/web/components/pages/onboarding/OnboardingRolloutQuest/rolloutSteps.tsx
talissoncosta: Fixed in 963519c42

frontend/web/components/pages/onboarding/OnboardingRolloutQuest/RolloutComingSoonCard.tsx
talissoncosta: Good call, done in 963519c42. Reused the event name as the trait key so there's one name for the thing:

```tsx 
    const alreadyAsked = notified || !!flagsmith.getTrait(ROLLOUT_BETA_REQUESTED) 
```

frontend/web/components/pages/onboarding/OnboardingFlow/OnboardingFlow.tsx
talissoncosta: Done in 963519c42. Created `onboarding_rollout_quest`.

frontend/web/components/pages/onboarding/OnboardingFlow/OnboardingFlow.tsx
```diff
@@ -167,6 +171,20 @@ const OnboardingFlow: FC = () => {
     organisation_id: organisationId,
     project_id: projectId,
   }
+
+  // Off, rollout deep-links to the overrides tab like every other next step.
+  const rolloutQuestEnabled = Utils.getFlagsmithHasFeature(
+    'onboarding_rollout_quest',
+  )
+  const goToNextStep = (step: OnboardingNextStep) =>
```
Zaimwa9, line 178: I think we need to branch the flag in the e2e tests too no ?

frontend/web/components/pages/onboarding/OnboardingRolloutQuest/RolloutComingSoonCard.tsx
```diff
@@ -0,0 +1,66 @@
...
+// promises a simpler flow, never the capability.
+const RolloutComingSoonCard: FC<RolloutComingSoonCardProps> = ({
+  onFeedback,
+  onNotifyMe,
+}) => {
+  // A trait rather than local state, so asking survives a reload and the beta
+  // can later be handed out by targeting it. Read at render, as Announcement
+  // does, so it still resolves if traits land after mount.
+  const [notified, setNotified] = useState(false)
+  const alreadyAsked = notified || !!flagsmith.getTrait(ROLLOUT_BETA_REQUESTED)
```
Zaimwa9, line 27: Non blocking: I had this issue with the experiment fake door. It might not be reactive and wouldn't properly re-render when a user returns to the fakedoor.

frontend/web/components/pages/onboarding/OnboardingFlow/OnboardingFlow.tsx
```diff
@@ -167,6 +171,20 @@ const OnboardingFlow: FC = () => {
     organisation_id: organisationId,
     project_id: projectId,
   }
+
+  // Off, rollout deep-links to the overrides tab like every other next step.
+  const rolloutQuestEnabled = Utils.getFlagsmithHasFeature(
+    'onboarding_rollout_quest',
```
Zaimwa9: Wouldn't it save some invokes by moving it within the `goToNextStep` function?

Zaimwa9 (commented):
Couple of extra comments to confirm please 👍

frontend/web/components/pages/onboarding/OnboardingFlow/OnboardingFlow.tsx
talissoncosta: Yep, it only passed with the flag off. Branched in b0c2dc498:

```ts
if (flagsmith.hasFeature('onboarding_rollout_quest')) {
  await expect(
    page.getByRole('heading', { name: 'How to roll out gradually today' }),
  ).toBeVisible();
  await page.getByRole('button', { name: 'Create a rollout segment' }).click();
}
await expect(page).toHaveURL(/\/features\?feature=\d+&tab=segment-overrides/);
```

Not run with the flag on yet.

frontend/web/components/pages/onboarding/OnboardingRolloutQuest/RolloutComingSoonCard.tsx
talissoncosta: It doesn't subscribe, agreed. The quest is a modal though, so it remounts on each open and re-reads the trait. My comment was the wrong bit, fixed in 6f0ff9523.

Real reactivity needs `onChange` in `config-store.js`, we don't mount `FlagsmithProvider`. Want it here?

frontend/web/components/pages/onboarding/OnboardingFlow/OnboardingFlow.tsx
talissoncosta: Worse than invokes: `hasFeature` counts an evaluation, and this page re-renders on every connection poll. Moved in 4271727da:

```tsx
const goToNextStep = (step: OnboardingNextStep) =>
  step === 'rollout' &&
  Utils.getFlagsmithHasFeature('onboarding_rollout_quest')
```

frontend/web/components/pages/onboarding/OnboardingRolloutQuest/RolloutComingSoonCard.tsx
Zaimwa9: Ah ok, no onChange might be overkilled, you are right on the re-mounting

Zaimwa9 (approved):
Thanks for addressing the changes 👍

Issue #7540, opened 2026-05-19
Epic: Onboarding Flow Improvements


talissoncosta:
The current onboarding stops at "you have created a flag" — users never reach the SDK install or a first evaluation in-app.

Following the audit and team review, we're collapsing onboarding to a fast path from signup to a first working flag, with a role question at the start so different personas get a path that fits their job.

**Done when**

The new flow is live behind a Flagsmith-on-Flagsmith flag, measured against the existing flow, with a recorded decision to ramp to 100% or roll back.

PR #8288, opened 2026-08-13
fix(onboarding): let segment overrides match the entry decision

talissoncosta:
## Changes

Contributes to #7540

Signing up with a `@flagsmith.com` email gave me the legacy flow, even though the `flagsmith_team` segment on `onboarding_quickstart_flow` should have put me into the new one.

The segment was fine. The entry decision identifies with no traits (`flagsmith.identify('', {}, true)`), and segment overrides match on traits, so none could ever apply. This passes the user's email through as a transient trait.

Bucketing is unchanged: the split is keyed on the identifier the API assigns, not on traits.

Reproduced the symptom by signing up with a `@flagsmith.com` email. Unit tests and eslint pass.

Not confirmed end to end: sign up with an email in the `flagsmith_team` segment and check you land on the new onboarding.

### DIFF
--- a/frontend/common/utils/onboardingEntry.ts
+++ b/frontend/common/utils/onboardingEntry.ts
@@ -23,8 +23,8 @@ export type OnboardingEntryDecision = {
 export async function decideOnboardingEntry(
   email?: string,
 ): Promise<OnboardingEntryDecision> {
-  // No trait means no segment override can match, and email is the only trait
-  // we have before the organisation exists.
+  // Only used to match segment overrides, not for bucketing. An override on
+  // another trait needs that trait passed here too.
   // @ts-expect-error transient is missing from the SDK's identify type
   await flagsmith.identify('', email ? { email } : {}, true)
   const flag = flagsmith.getExperimentFlag('onboarding_quickstart_flow')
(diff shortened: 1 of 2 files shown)


frontend/common/utils/onboardingEntry.ts
```diff
@@ -20,9 +20,13 @@ export type OnboardingEntryDecision = {
  * the logged-in user, and the routing it should have driven has happened.
  * Call `persistOnboardingEntry` with an accepted decision.
  */
-export async function decideOnboardingEntry(): Promise<OnboardingEntryDecision> {
+export async function decideOnboardingEntry(
+  email?: string,
+): Promise<OnboardingEntryDecision> {
+  // No trait means no segment override can match, and email is the only trait
+  // we have before the organisation exists.
```
Zaimwa9: Ok I think this comment is misleading in the sense that it made me understand that we needed a trait to have a proper evaluation while it's more to identify flagsmith' team.

That being said, that also means we need to be careful (maybe adding it in the description of the flag in the app) that for any extra segment overrides for this specific flag, we need to pass it (if available) here

Zaimwa9 (dismissed):
Minor but let's not keep the misleading comment

frontend/common/utils/onboardingEntry.ts
talissoncosta: Thanks for the call. Reworded in 2b78a28.

```tsx
   // Only used to match segment overrides, not for bucketing. An override on
   // another trait needs that trait passed here too.
```
Does it make sense to you ?
I am planning to add similar note to the flag description in the app.

kyle-ssg (approved):

kyle-ssg (approved):

PR #8294, opened 2026-08-14
feat: cohort CSV sync endpoint and cohort summary on segments

Zaimwa9:
Thanks for submitting a PR! Please check the boxes below:

- [x] I have read the [Contributing Guide](/Flagsmith/flagsmith/blob/main/CONTRIBUTING.md).
- [x] I have added information to `docs/` if required so people know about the feature.
- [x] I have filled in the "Changes" section below.
- [x] I have filled in the "How did you test this code" section below.

## Changes

Adds the API surface the dashboard needs to create segments from a CSV of identifiers, on top of the cohort CRUD from #8248.

**CSV sync endpoint**

- `POST /api/v1/environments/{api_key}/cohorts/{id}/sync-csv/`: multipart upload with `file`, `identifier_column` (0-based, defaults to the first column) and `has_header` (defaults to true).
- Declarative replace: the uploaded identifiers become the cohort's desired membership. Responds `202` with `{version, added, removed, unchanged, ignored: {empty, duplicates, too_long}}` and queues the membership delta application inside the transaction, with a row lock on the cohort and a version bump.
- The file is parsed and discarded, never stored. Files over 10MB are rejected with `413`.

**Cohort creation**

- `POST .../cohorts/` now accepts segment metadata (custom fields), applied to the managed segment.

**Segments API**

- Segment payloads carry a read-only `cohort` summary (`id`, `environment`, `source_type`, `version`, `deletion_requested_at`), prefetched on the list view. This lets the dashboard tag CSV segments, route their deletion through the cohorts endpoint, and grey out segments awaiting drain.

**Observability**

- New metrics `flagsmith_cohorts_csv_syncs_total` and `flagsmith_cohorts_csv_sync_identifiers`, new `cohorts.csv.synced` event, both documented in the observability catalogues.

## How did you test this code?

- Unit tests for the CSV extraction and sync services (delta semantics, duplicates, empty rows, too-long identifiers, version bumps), the sync-csv view (validation, size cap, permissions), metadata on create, and the segment list cohort summary.
- Manually through the dashboard drawer (stacked frontend branch): created cohorts from CSV files, verified membership rows and delta counts in the database, re-synced and deleted cohorts.

PR #8295, opened 2026-08-14
feat: wire CSV segments to the cohorts API

Zaimwa9:

## Changes

Wires the CSV segment drawer (#8283) to the cohorts API (#8294).

**Create flow**

- New `useCohort` service: `createCohort`, `syncCohortCsv` (multipart) and `deleteCohort` mutations, invalidating the segment list.
- The drawer submits in two steps: create the cohort, then sync the trimmed, headerless, single-column CSV built from the selected identifier column. Only the identifier column leaves the browser. If the sync fails, retrying reuses the created cohort instead of creating a duplicate segment.
- Success toast reports the added count from the API and stays visible for 10 seconds.
- Custom Fields (metadata) are sent with the create call.

**Segment list**

- CSV-backed segments show a `CSV` chip and a chip with the targeted environment's name, driven by the `cohort` summary on the segment payload.
- Removing a CSV segment routes through the cohort delete endpoint (the segment endpoint rejects cohort-managed segments); plain segments are unchanged.
- Deletion is drain-then-delete on the API side, so while a cohort is awaiting drain the row is greyed out, cannot be opened, and has no actions.
- Fixed the row action dropdown staying open under the confirm modal (it only closed on outside clicks).

**Upload hardening**

- `CsvUpload` takes a `maxSizeBytes` prop; the drawer passes 10MB to mirror the API cap, so oversized files are rejected client-side with a clear message.
- `toParsedCsv` no longer spreads every row into `Math.max`, which overflowed the call stack on large files.

- Unit tests for the CSV utilities (17 passing), including quoting and column extraction.
- Manual QA against a local API: happy path with multi-column and single-column files, count parity between the preview and the toast, metadata on create, oversized file rejection, CSV and plain segment deletion, the deleting state, and membership rows verified in the database.

image

### DIFF
--- a/frontend/common/services/useCohort.ts
+++ b/frontend/common/services/useCohort.ts
@@ -1,0 +1,82 @@
+import { Res } from 'common/types/responses'
+import { Req } from 'common/types/requests'
+import { service } from 'common/service'
+import toFormData from 'common/utils/toFormData'
+
+export const cohortService = service
+  .enhanceEndpoints({ addTagTypes: ['Cohort', 'Segment'] })
+  .injectEndpoints({
+    endpoints: (builder) => ({
+      createCohort: builder.mutation<Res['cohort'], Req['createCohort']>({
+        invalidatesTags: (q, e, arg) => [
+          { id: 'LIST', type: 'Cohort' },
+          { id: `LIST${arg.projectId}`, type: 'Segment' },
+        ],
+        query: (query) => ({
+          body: {
+            description: query.description,
+            metadata: query.metadata,
+            name: query.name,
+          },
+          method: 'POST',
+          url: `environments/${query.environmentApiKey}/cohorts/`,
+        }),
+      }),
+      deleteCohort: builder.mutation<void, Req['deleteCohort']>({
+        invalidatesTags: (q, e, arg) => [
+          { id: 'LIST', type: 'Cohort' },
+          { id: `LIST${arg.projectId}`, type: 'Segment' },
+        ],
+        query: (query) => ({
+          method: 'DELETE',
+          url: `environments/${query.environmentApiKey}/cohorts/${query.cohortId}/`,
+        }),
+      }),
+      syncCohortCsv: builder.mutation<
+        Res['cohortCsvSync'],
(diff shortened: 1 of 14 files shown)


Zaimwa9:
@themis-blindfold review

Zaimwa9:
@themis-blindfold review

Zaimwa9:
@themis-blindfold review

Zaimwa9:
@themis-blindfold review

frontend/common/services/useCohort.ts
```diff
@@ -0,0 +1,82 @@
...
+          const formData = toFormData({ ...rest })
+          const { data, error } = await baseQuery({
+            body: formData,
+            method: 'POST',
+            url: `environments/${environmentApiKey}/cohorts/${cohortId}/sync-csv/`,
+          })
+          return { data, error } as {
+            data: Res['cohortCsvSync']
+            error: never
+          }
```
talissoncosta, line 55: Non blocking: is this assertion doing anything? Removing it typechecks, and `unwrap()` still resolves to `Res['cohortCsvSync']` from the mutation generic:

```ts
return { data, error }
```

frontend/common/types/responses.ts
```diff
@@ -174,6 +174,16 @@ export type SegmentMembersResponse = PagedResponse<SegmentMember> & {
   // Pass as `cursor` to fetch the next page; null when there are no more rows.
   next_cursor: string | null
 }
+export type SegmentCohort = {
+  id: number
+  environment: number
+  environment_api_key: string
+  environment_name: string
+  source_type: 'csv'
```
talissoncosta, line 182: Small and not blocking. What about adding it to a shared type ? when you add the next integrations it only needs a single change.

talissoncosta (approved):
Looks pretty good, just some non-blocking comments.

What about the name input, is it intentional to be narrow ?

Image

frontend/common/types/responses.ts
Zaimwa9: Yep thanks, i'll keep it for the next integrations that are coming quite soon

Issue #19, opened 2018-10-10
Feature Leakage


rawkode:
When calling `getFeatures`, the API returns **ALL** feature flags. I'm proposing we should only return those flags which are enabled for the given user.

I don't see the point in "leaking" feature flags to an inquisitive user that otherwise doesn't need to know they exist :+1: 

Proposing the API would return this, instead of the latter:
```json
{
    "gt8_feature_flag_test": {
        "enabled": true,
        "value": null
    },
    "floating_action_button": {
        "enabled": true,
        "value": null
    },
    "oauth_google": {
        "enabled": true,
        "value": null
    },
    "interactive_tutorial": {
        "enabled": true,
        "value": null
    },
    "breadcrumbs": {
        "enabled": true,
        "value": null
    },
    "tasks": {
        "enabled": true,
        "value": null
    }
}

```

instead of:

```json
{
    "gt8_feature_flag_test": {
        "enabled": true,
        "value": null
    },
    "floating_action_button": {
        "enabled": true,
        "value": null
    },
    "oauth_google": {
        "enabled": true,
        "value": null
    },
    "test_remote": {
        "enabled": false,
        "value": "1"
    },
    "interactive_tutorial": {
        "enabled": true,
        "value": null
    },
    "breadcrumbs": {
        "enabled": true,
        "value": null
    },
    "thought_from_template": {
        "enabled": false,
        "value": null
    },
    "tasks": {
        "enabled": true,
        "value": null
    }
}
```

dabeeeenster:
We've had a discussion about this and we're not sure hiding those disabled flags is the right approach. There will be cases where the default state for an app is having something enabled, and so the feature flag state being disabled would be relevant/required. 

We're not sure how we could get around this situation? I guess an option would be to add a second boolean ```hidden``` property to each flag that allowed you to prevent them from being sent back to clients?

rawkode:
> We're not sure how we could get around this situation? I guess an option would be to add a second boolean hidden property to each flag that allowed you to prevent them from being sent back to clients?

Seems like a sensible solution to me :+1:

dabeeeenster:
We now have the ability to hide disabled flags.

PR #8305, opened 2026-08-15
feat(edge-proxy): Wire the flagsmith-private edge_control_plane app into the api

gagantrivedi:

## Changes

Contributes to https://github.com/Flagsmith/edge-proxy/issues/128

Wires up the `edge_control_plane` private app (Flagsmith/flagsmith-private#265, renamed in #285, released in `flagsmith-private` 0.14.0, which `main` already locks via #8540). It provides the proxy key credential, the proxy config endpoint that lets a running Edge Proxy discover its environments, and the usage ingestion endpoint. Consumed by the proxy in Flagsmith/edge-proxy-rs#17, #18 and #19.

Three edits, following the `rbac`/`scim` pattern:

- `EDGE_CONTROL_PLANE_INSTALLED` detection and conditional `INSTALLED_APPS` append.
- Mounts for the two URL trees: proxy key management under `organisations/{id}/edge-proxy/`, and `api/v1/proxy/` for the config and usage endpoints.
- A mypy override so the no-private run passes.

Private cloud only: SaaS images ship the private wheel, so the mounts are gated on `is_saas()` at runtime rather than on the app being absent; the app's own views carry a matching permission as defence in depth.

With the 0.14.0 app installed: `EDGE_CONTROL_PLANE_INSTALLED` is `True`, `edge_control_plane` is in `INSTALLED_APPS`, and the three routes resolve (`/api/v1/proxy/config/`, `/api/v1/proxy/usage/`, `/api/v1/organisations/1/edge-proxy/keys/`). With a `SAAS_DEPLOYMENT` marker file present they 404. The app's own behaviour is covered by the integration tests in flagsmith-private.


api/app/settings/common.py
```diff
@@ -1160,6 +1160,10 @@
         "USER_FILTER_PARSER": "scim.filters.UserFilterQuery",
     }
 
+EDGE_PROXY_INSTALLED = importlib.util.find_spec("edge_proxy") is not None
+if EDGE_PROXY_INSTALLED:
+    INSTALLED_APPS.append("edge_proxy")
```
matthewelwell, line 1165: I'd like to come up with a better name here, but I'm struggling to think of one... Some suggestions: 

- `edge_connections`
- `edge_orchestration`
- `edge_proxy_connections`
- `edge_proxy_orchestation`

gagantrivedi: What do you think about `edge_control_plane`

matthewelwell: Nice, I like it!

matthewelwell (approved):

PR #8307, opened 2026-08-15
feat(__future__): Delete a flag's segment override

### LINKED ISSUE #8300: `DELETE` update-flag endpoint for individual overrides
> I decided not to add a `DELETE` endpoint solely to squeeze all I could out of scope, since the dashboard will already make a `PUT` request. I'm not against the idea though — would be happy to see an issue.

_Originally posted by @emyller in https://github.com/Flagsmith/flagsmith/pull/8102#discussion_r3787055503_

khvn26:

## Changes

Closes https://github.com/Flagsmith/flagsmith/issues/8300

Adds `DELETE /api/__future__/environments/{environment_key}/features/{feature_id}/segment-overrides/{segment_id}/`, so a caller can remove one segment override without sending the whole list back through `PUT`.

Behaviour follows the sibling verbs:

- `MANAGE_SEGMENT_OVERRIDES` gates it. A caller with no permission on the environment gets `404`; one holding only `UPDATE_FEATURE_STATE` gets `403`.
- `409` where change requests are enabled, logging `features.flag.update_rejected`.
- `404` for non-existing segment overrides.
- Versioning-agnostic.
- Remaining overrides keep their priorities.

Updated documentation: https://docs-git-feat-update-flag-delete-flagsmith.vercel.app/managing-flags/updating-flags

Added integration tests.


api/tests/integration/features/future/test_flag_endpoint.py
```diff
@@ -2060,3 +2060,349 @@ def test_update_flag__variants_on_standard_feature__responds_400(
         .multivariate_feature_state_values.exists()
     )
     assert log.events == []
+
+
+@pytest.fixture()
+def two_segment_overrides(
```
emyller: Do you mind moving this fixture to fixtureland? i.e. top of the file, or maybe it's time for a separate collection.

khvn26: 4b36fdd59

api/features/future/services.py
```diff
@@ -70,6 +73,27 @@ def _get_feature_states_to_write(
...
+    )
+
+
+def _get_overrides_by_segment_id(
+    feature_states: Sequence[FeatureState],
+) -> dict[int, FeatureState]:
+    return {
+        feature_segment.segment_id: feature_state
+        for feature_state in feature_states
+        if (feature_segment := feature_state.feature_segment) is not None
```
emyller, line 93: Do we need to prefetch `feature_segment` and avoid a N+1 by pre-hydrating these? If you think this is a concern, please add a queries test.

api/tests/integration/features/future/test_flag_endpoint.py
```diff
@@ -2060,3 +2096,313 @@ def test_update_flag__variants_on_standard_feature__responds_400(
...
+            "environment__id": versioned_environment.id,
+            "feature__id": feature,
+            "segment_overrides__created__segment__ids": [],
+            "segment_overrides__updated__segment__ids": [],
+            "segment_overrides__deleted__segment__ids": [segment],
+        },
+    ]
+
+
+def test_delete_segment_override__last_override__removes_override(
```
emyller: question: why is this one relevant compared to the test above it?

emyller (dismissed):

emyller:
Left a couple non-blocking comments.

api/tests/integration/features/future/test_flag_endpoint.py
khvn26: Good catch. 2ed87a3

api/features/future/services.py
khvn26: We already do. I tried adding unit tests with query count asserts, and it revealed that v2 versioning induces an N+1 anyway, so I reverted them for the time being, and opened #8312.

emyller (approved):
LGTM

PR #8309, opened 2026-08-15
fix: prevent 500 error on partial update of multivariate options

### LINKED ISSUE #8306: Partially updating a multivariate option returns 500
`PATCH /api/v1/projects/{project}/features/{feature}/mv-options/{id}/` returns a 500 when the payload leaves out `default_percentage_allocation`, which prevents updates to change only a variant key.

## Steps To Reproduce

1. Create a multivariate feature with at least one variant.
2. `PATCH .../mv-options/{id}/` with `{"feature": <feature_id>, "key": "hero"}`.
3. The response is 500.

## Expected behaviour

A partial update succeeds, leaving the allocation as it is.

srijantrpth:

## Changes

Closes #8306 

Resolved an issue where partially updating a multivariate option via a `PATCH` request returned a 500 Internal Server Error when omitting fields like `default_percentage_allocation` or `feature`. 

Updated `MultivariateFeatureOptionSerializer.validate()` and `_validate_key_is_unique()` in `api/features/multivariate/serializers.py` to safely extract these fields using `.get()` and fallback to `self.instance` values, preventing the unhandled `KeyError`.

Added a unit test (`test_partially_updating_multivariate_option_success`) in `api/tests/unit/features/multivariate/test_unit_multivariate_views.py` to verify that sending a `PATCH` request with only the `key` field correctly returns a 200 OK and preserves the existing `default_percentage_allocation` without throwing an error. Tested successfully via pytest.


khvn26:
Hello @srijantrpth, thanks for your contribution.

Can you fix the remaining linting error?

```
flagsmith-lint-tests.....................................................Failed
- hook id: flagsmith-lint-tests
- exit code: 1

api/tests/unit/features/multivariate/test_unit_multivariate_views.py:111:1: FT003 Test name `test_partially_updating_multivariate_option_success` doesn't match `test_{subject}__{condition}__{expected}` (found 1 parts, expected 3)
```

srijantrpth:
Hi @khvn26 , Thanks for the review! I just pushed a commit to fix that linting error by renaming the test to match the required test_{subject}underscore underscore{condition}underscore underscore {expected} convention. Let me know if everything looks good now!

api/tests/unit/features/multivariate/test_unit_multivariate_views.py
```diff
@@ -105,3 +106,36 @@ def test_list_mv_options__feature_in_other_project__returns_404(
...
+) -> None:
+    # Given
+    mv_option = multivariate_feature.multivariate_options.first()
+
+    assert mv_option is not None
+
+    url = reverse(
+        "api-v1:projects:feature-mv-options-detail",
+        args=[project.id, multivariate_feature.id, mv_option.id],
+    )
```
khvn26: Can we use a compiled url here? See [here](https://github.com/Flagsmith/flagsmith/pull/8309/changes#diff-772430d57aeb980845a1401224fb87d1263cca1602c00bd683cd26a0540a98a1R102).

khvn26: Sorry. See the `url` variable in `test_list_mv_options__feature_in_other_project__returns_404` as an example.

srijantrpth: Hey @khvn26, I think the link you added for 'here' might be broken it just loops me back to this exact comment! Could you share the link to the example you meant to show me so I can get this updated?

srijantrpth: Ah, got it! Thanks for pointing me to that example. I've updated it to use the f-string URL pattern.

khvn26 (approved):
This LGTM.

Would like @gagantrivedi's or @Zaimwa9's opinion before merging this to make sure the experimentation team is aligned with this change.

PR #8310, opened 2026-08-16
refactor(Versioning): Publish a feature version with an AuthorData

khvn26:

## Changes

In this PR, we accommodate `EnvironmentFeatureVersion.publish()` to the `AuthorData` dataclass.

`EnvironmentFeatureVersion.publish()` took `published_by` and `published_by_api_key`, so every caller had to categorise a request user. That split was written four ways across the codebase, and one of them was wrong: `EnvironmentFeatureVersionCreateSerializer` only handled `FFAdminUser`, so a version created with `publish_immediately` by a master API key recorded no publisher at all. Fixed here, with a test.

`publish()` now takes an `AuthorData`, which callers construct directly, or via `AuthorData.from_request(request)` where a request is available.

We also drop a dead `serializer.save(published_by=request.user)` kwarg in `EnvironmentFeatureVersionViewSet.publish` — the serializer reads the author from the request context and ignored it.

No behaviour change beyond the attribution fix.

Added unit tests, and modified existing tests asserting `EnvironmentFeatureVersion.publish()`.

### DIFF
--- a/api/features/future/services.py
+++ b/api/features/future/services.py
@@ -7,7 +7,7 @@ import structlog
 from django.db import transaction
 from django.db.models import Count, Q
 
-from api_keys.user import APIKeyUser
+from core.dataclasses import AuthorData
 from environments.models import Environment
 from features.future.exceptions import (
     DuplicatePriorityError,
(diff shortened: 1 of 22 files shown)


api/features/future/services.py
```diff
@@ -294,7 +282,7 @@ def update_flag(
     feature: Feature,
     changes: UpdateFlagRequest,
     replace: bool,
-    author: FFAdminUser | APIKeyUser,
+    author: AuthorData,
```
Zaimwa9, line 285: I feel that this could somehow give some leads to #5978

Zaimwa9 (approved):
Clean, thanks 👍 
Just commented a thought that you can resolve when read

PR #7440, opened 2026-05-06
refactor(button): migrate _buttons.scss to semantic colour tokens

talissoncosta:
Thanks for submitting a PR! Please check the boxes below:

- [x] I have read the [Contributing Guide](/Flagsmith/flagsmith/blob/main/CONTRIBUTING.md).
- [x] I have added information to \`docs/\` if required so people know about the feature.
- [x] I have filled in the "Changes" section below.
- [x] I have filled in the "How did you test this code" section below.

## Changes

**TL;DR.** Phase 2 of the design-system migration. Replaces SCSS theme primitives with semantic CSS custom properties in \`_buttons.scss\`. **Deletes the entire \`.dark { ... }\` block** (~110 lines) — semantic tokens auto-theme via the CSS cascade, the manual dark overrides are now redundant.

> **Depends on #7439** (Phase 1, the saturated token additions). Until that lands, this branch is based on it.

### What changed

\`_buttons.scss\`: **+47 lines, −160 lines** (net −113 lines).

Before:
\`\`\`scss
.btn:hover, .btn:focus { background-color: \$btn-hover-bg; }   // = \$primary600

.dark .btn:hover { background-color: \$btn-hover-bg-dark; }    // = \$primary400
\`\`\`

After:
\`\`\`scss
.btn:hover, .btn:focus { background-color: var(--color-surface-action-hover); }
// (no .dark block needed — token auto-themes)
\`\`\`

The entire \`.dark { ... }\` block is gone. Most of its rules existed only to override colours — semantic tokens handle that automatically because their values switch between \`:root\` and \`[data-bs-theme='dark']\` in \`_tokens.scss\`.

### Intentional visual diffs (Chromatic will surface)

- **Primary button hover/active** shifts one shade darker. The old \`\$primary600\`/\`\$primary700\` mapped to the *base* and *hover* in the semantic ramp; the new mapping gives a more progressive darken-on-interaction. More conventional.
- **Success button hover gets DARKER**, not lighter. The old pattern used \`\$success400\` on hover (lightening the button), which is unusual. The new mapping (\`success-strong-hover\` = green-600) follows the standard darken-on-hover convention.
- **Danger button hover/active** shift slightly — from custom hex shades (\`#cd384d\`, \`#ac2646\`) to the primitive \`red-600\`/\`red-700\` ramp. Same visual intent, more consistent across the palette.

### What's still using SCSS variables (deferred)

These don't have clean semantic-token mappings in the current catalogue:

[...]

PR #8314, opened 2026-08-17
fix(button): give buttons a visible keyboard focus indicator

talissoncosta:

## Changes

`.btn:focus-visible` set `box-shadow: none`, removing Bootstrap's focus ring without replacing it, so no button in the app shows keyboard focus. Icon-only buttons show nothing at all, which reads as Tab skipping them.

Two commits:

1. Replaces the suppressed ring with the outline `BareButton` already uses, so the two primitives match. WCAG 2.4.7.
2. Swaps the remaining `:focus` rules to `:focus-visible`, so a mouse click no longer leaves a button looking pressed. This is #7437, which was closed unreviewed; drop the commit if you'd rather keep it separate.

Not yet, hence the draft. It touches every button, so it needs a visual-regression run, particularly icon buttons in table cells and tight flex rows where `outline-offset` may clip.

Found while reviewing #8044.

### DIFF
--- a/frontend/web/styles/project/_buttons.scss
+++ b/frontend/web/styles/project/_buttons.scss
@@ -20,7 +20,7 @@ button.btn {
   &-link {
     color: $primary;
     &:hover,
-    &:focus {
+    &:focus-visible {
       background: transparent;
     }
     &.btn:active {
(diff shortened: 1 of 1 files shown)


talissoncosta:
Closing in favour of #7440, which migrates `_buttons.scss` to semantic tokens and deletes the `.dark` block, so most of the rules this PR edits disappear there.

Carrying over: #7440 still has `&:focus-visible { box-shadow: none; }` with nothing in its place, and still styles `:focus` rather than `:focus-visible` on the hover pairs. Those two fixes need to land there, or buttons keep showing no keyboard focus and looking pressed after a click.

Issue #8258, opened 2026-08-11
Display the billing period dates on the dashboard


talissoncosta:
**Goal.** Someone can see which billing period they are in, when their allowance resets, and how long is left.

Three things on the usage page describe the period itself, and none can be read from the API today:

- the period's date range
- when the allowance resets
- days remaining

Needs #8257. Free plans have no period, so they show nothing in their place, which the page already handles.

**Done when** someone on a paid plan can see all three, and someone on a rolling window sees none of them.

<img width="938" height="60" alt="Image" src="https://github.com/user-attachments/assets/92396cf9-e71d-42d7-b9f4-568f308d2463" />

<img width="473" height="218" alt="Image" src="https://github.com/user-attachments/assets/74a042bf-a5bc-40fd-be83-5488723de437" />

Issue #8188, opened 2026-07-31
Show projected end-of-period usage


talissoncosta:
**Goal.** See where usage is heading before the period ends, while there is still time to act. Today the page only shows what has already happened.

**Build.** One calculation, three places:

- a **Projected end-of-period** tile
- a **dashed line** on the chart, from today to the period end
- a **sentence under the meter**, like "On track to use ~1.9M (94% of your limit) by 17 Aug", or that it lands over the limit

**Hide it early in the period**, where an average means very little. Pick the point: a number of days, or a share of the period.

**Never on a rolling window.** Free plans have no period end to project to, and their total can fall as old days drop out, so a projection there is misleading rather than merely early.

**Call it an estimate.** A straight-line average of usage so far, not a forecast.

**Needs #8257.** The projection divides by the period length, which the API does not return today.

**Done when**

- the projection appears in all three places once there is enough of the period to mean anything, and in none of them before
- a projection that lands over the limit is obvious without working it out

<img width="535" height="164" alt="Image" src="https://github.com/user-attachments/assets/116f9655-b6e0-4f09-8030-73f0f74b42db" />

Issue #8187, opened 2026-07-31
Build the over-limit and restricted states for the usage page


talissoncosta:
**Goal.** Build the states shown when an organisation goes past its plan limit, so it can see what happened, what it means and what restores normal service.

**Problem.** Neither state exists. The restricted one is unreachable anyway: a restricted organisation is sent to the blocked screen on every route except the organisations list, so the page that would explain it is the one page they cannot open.

**Scope**

- **paid, over limit**: the banner, and the line under the meter saying how far over they are
- **free, over limit**: that they have 7 days to bring usage down before serving pauses
- **free, restricted**: what is paused, what restores it, and the upgrade path
- usage, limit and the over-limit figures readable in every state
- allow the usage route while the organisation is blocked from admin, otherwise the restricted state cannot be reached

**Correction: the 7 days is consumable.** `OrganisationBreachedGracePeriod` is written on the first restriction and never deleted, and `restrict_use_due_to_api_limit_grace_period_over` drops the wait once it exists, so a repeat breach is restricted on the next task run. The notification emails already conditionalise on this. The API does not say which case an organisation is in, so the page cannot promise the deadline truthfully.

**Status**

#8432 ships the route fix and both over-limit states, and a restricted organisation now gets its own copy telling it what restores access. Still open:

- **what is paused.** `stop_serving_flags` is on the model but on no serializer, so the page can say an organisation is restricted but not whether flag serving stopped or only admin access. Requested on #8256
- **the 7 day warning for a first free breach.** Needs the API to distinguish a first breach from a repeat, also #8256
- **a regression test for the route exception.** Needs an e2e test that can set organisation-level flags; nothing in the Playwright suite does that today

**Not here**

- whether a paid customer is covered by grace or being charged. That is #8264, which needs #8256
- estimated overage cost. The page can say whether they are charged, not how much
- notification configuration

**Done when**

- a paid organisation over its limit sees the over-limit state
- a free organisation over its limit is told it has 7 days to act
- a restricted organisation loads the usage page and sees the restricted state
- every other route still blocks as it does today, and an unrestricted organisation sees no change

<img width="808" height="643" alt="Image" src="https://github.com/user-attachments/assets/73dfd664-0562-4f73-a3bb-e1a8d6fe79d9" />

<img width="935" height="750" alt="Image" src="https://github.com/user-attachments/assets/82a85ff2-fc70-4283-a4dc-fc35978c31c5" />

Issue #8186, opened 2026-07-31
Break usage down by request type or SDK


talissoncosta:
**Goal.** Someone who has seen how much they used can find out where it came from.

**Problem.** The breakdown is a daily volume chart with By Endpoint and By SDK as separate tabs, so the four billable request types are never named as such.

**Scope**

- the breakdown gets its own section, with no plan limit on it, since an individual request type has no limit of its own
- rename By Endpoint to By request type. The rows are the four billable types
- a dimension menu: request type, SDK

Both dimensions derive from the response the charts already fetch, so they cost no extra requests. Project and environment are split out into their own issue: each costs one request per key, and `usage-data` is throttled at 5/min per user, so a multi-project org gets 429s and a list that ranks the rest wrongly.

**Decide here.** The menu covers the same ground as the existing tabs. If it replaces them, removing the old By SDK view and its sidebar is part of this work rather than left behind as dead code.

**Done when**

- the breakdown renders per dimension, and carries no limit line
- either the old tabs are gone, or there is a written reason they stayed

<img width="903" height="222" alt="Image" src="https://github.com/user-attachments/assets/636872ce-9be1-4df8-9a70-5a9a97604c33" />

<img width="355" height="292" alt="Image" src="https://github.com/user-attachments/assets/9d67c8bb-853b-4de0-8d66-4faa6d272f4d" />

PR #8320, opened 2026-08-18
feat(usage): Show usage against the plan limit for the current billing period

### LINKED ISSUE #8185: Show usage against the plan limit for the current billing period
**Goal.** Someone opening Usage can tell how much of their plan they have used, without doing any arithmetic.

**Problem.** The page reports totals for whichever window you picked. It never says how close you are to the plan limit.

**Scope**

- default to the current billing period
- "% of plan consumed" as the headline, and as a tile
- the usage bar, with the notification markers on it
- cumulative usage against the plan limit, with the limit drawn as a ceiling. Billing periods only, see below
- the slot under the meter for the line that explains the numbers. Its wording is a projection (#8188) or an overage consequence (#8187), so this issue builds the slot rather than the sentences

Covers the healthy and approaching states. Over limit and restricted are #8187.

**Not here**

- the projected end-of-period figure, in the tile, the chart line and the slot. All #8188
- the period's dates, its reset date and days remaining. All #8258, because none of them can be read from the API today

**Free plans**

Measured over a trailing 30 days, so there is no period start and nothing to project to. They get the meter and daily volume rather than cumulative against a ceiling, because a rolling total can fall as old days drop out of the window.

**No plan limit**

Where there is no limit to divide by, show the total and no percentage.

**Done when** someone on any plan can see how much of their allowance they have used.

Image

talissoncosta:
### Reviewing this

Every state is in [Storybook under Pages > Usage Dashboard](https://69aedd804464bb3078b3ee78-suwpccilxq.chromatic.com/?path=/story/pages-usage-dashboard-page--paid-with-a-billing-period) if you want to scan them quickly.

## Changes

Closes #8185

Shows usage against the plan allowance. Behind `usage_dashboard`, off everywhere. Replaces the old page rather than extending it, so nothing changes until the flag is on.

- a meter: percentage consumed, used against limit, notification thresholds
- a chart: cumulative against the limit where there is a billing period, daily bars where there is not, because a rolling window's total falls as old days leave it
- loading, failure and empty states

Worth knowing:

- free plans now default to 30 days rather than 90, which is the window enforcement actually measures
- `UsageBar` gains `thresholds`, `warnAt` and `ariaLabel`. `warnAt` defaults to 85, so the two other pages using it are unchanged
- also fixes `Format.shortenNumber`, which returned `NaN` for `0`. `OrganisationLimit` had the same exposure

Not here: period dates (#8258), projection (#8188), over-limit and restricted states (#8187), breakdown (#8186). The flag should stay internal until #8186, since swapping the page loses the request-type totals and the SDK view.

99 unit tests over the period rules, the daily aggregation, the bar maths and the meter copy.

To QA, with `usage_dashboard` on:

- [ ] paid org: meter shows % of plan, chart is a rising line with a red limit ceiling
- [ ] free org: daily bars, period defaults to "Last 30 days"
- [ ] project filter on a project with no traffic: "No usage recorded", not blank space
- [ ] flag off: Organisation Settings > Usage behaves as today
- [ ] Project Settings > Usage bars unchanged


frontend/web/components/pages/OrganisationUsagePage.tsx
```diff
@@ -125,6 +128,11 @@ const OrganisationUsagePage: FC = () => {
     }
   }
 
+  // The billing-aligned dashboard replaces this page rather than extending it.
+  if (isUsageDashboardEnabled) {
+    return <UsageDashboard organisationId={organisationId || 0} />
+  }
```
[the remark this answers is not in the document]
talissoncosta: Went with `skip` on both queries instead of splitting the page (3b2963e):

```ts
{ skip: !organisationId || isUsageDashboardEnabled }
```

Same effect, and this file gets deleted when the flag comes out, so the split would be short-lived.

frontend/web/components/organisation-settings/usage/UsageDashboard/UsageDashboard.tsx
```diff
@@ -0,0 +1,90 @@
...
+    !!subscription?.has_active_billing_periods
+
+  // The default depends on the plan, which arrives after the first render, so
+  // only an explicit choice is stored. 'default' resolves once it is known.
+  const [chosenPeriod, setChosenPeriod] = useState<
+    Req['getOrganisationUsage']['billing_period'] | 'default'
+  >('default')
+  const defaultPeriod = hasBillingPeriod ? 'current_billing_period' : undefined
+  const billingPeriod =
+    chosenPeriod === 'default' ? defaultPeriod : chosenPeriod
```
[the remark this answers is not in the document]
talissoncosta: `freePeriods` already carries the rolling window as its value:

```ts
export const freePeriods = [
  { label: 'Last 90 days', value: '90_day_period' },
  { label: 'Last 30 days', value: undefined },
]
```

So `.find((period) => period.value === billingPeriod)` matches and the select shows "Last 30 days". `undefined` is the value the API expects for that window, not a missing selection.

frontend/web/components/organisation-settings/usage/UsageOverTime/UsageOverTime.tsx
```diff
@@ -0,0 +1,103 @@
...
+  [...(data?.events_list ?? [])]
+    .sort((a, b) => (a.day < b.day ? -1 : 1))
+    .map((event) => ({
+      day: moment(event.day).format('D MMM'),
+      total:
+        (event.flags ?? 0) +
+        (event.identities ?? 0) +
+        (event.traits ?? 0) +
+        (event.environment_document ?? 0),
+    }))
```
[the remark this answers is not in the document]
talissoncosta: Good catch, this was a real bug. The API returns a row per day and client type, so any day with more than one drew duplicate bars. Sums by raw date before formatting now (6a2de77).

frontend/web/components/shared/UsageBar/UsageBar.tsx
```diff
@@ -0,0 +1,84 @@
...
+          role='progressbar'
+          aria-label={label ?? ariaLabel}
+          aria-valuenow={percent}
+          aria-valuemin={0}
+          aria-valuemax={100}
+          aria-valuetext={`${percent}% of ${Format.shortenNumber(limit)}`}
+        >
+          <div
+            className={`usage-bar__fill usage-bar__fill--${tone}`}
+            style={{ width: `${Math.min(percent, 100)}%` }}
```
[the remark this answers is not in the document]
talissoncosta: Real bug too. An org at 118% reported `aria-valuenow={118}` against `aria-valuemax={100}`. Clamped in 6a2de77, with `aria-valuetext` keeping the true percentage.

frontend/common/utils/format.ts
```diff
@@ -42,6 +42,11 @@ const Format = {
   shortenNumber(number: number): string {
     // Converts a float number into a short literal with suffix for the magnitude:
     // 1523125 > 1.5M
+    // Guarded because the maths below takes log10 of the value, so zero and
+    // anything missing come back as NaN.
+    if (!number || !Number.isFinite(number)) {
+      return '0'
```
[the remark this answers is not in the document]
talissoncosta: Widened to a named type in 6a2de77:

```ts
type NullableNumber = number | null | undefined
```

frontend/common/utils/__tests__/format.test.ts
```diff
@@ -12,6 +12,10 @@ describe('Format', () => {
       ${1234}          | ${'1.2K'}
       ${12345}         | ${'12.3K'}
       ${123456}        | ${'123.5K'}
+      ${0}             | ${'0'}
+      ${undefined}     | ${'0'}
+      ${null}          | ${'0'}
+      ${NaN}           | ${'0'}
```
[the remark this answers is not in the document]
talissoncosta: Added both infinities in 6a2de77.

frontend/web/components/organisation-settings/usage/UsageMeter/UsageMeter.tsx
```diff
@@ -0,0 +1,71 @@
...
+const UsageMeter: FC<UsageMeterProps> = ({ limit, note, total }) => {
+  const hasLimit = !!limit
+  const percent = hasLimit ? Math.round((total / limit) * 100) : 0
+  const tone = toneFor(percent)
+
+  return (
+    <div className='usage-meter'>
+      <div className='usage-meter__headline'>
+        <div>
+          <p className='usage-meter__label'>Plan usage this period</p>
```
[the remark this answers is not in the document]
talissoncosta: Right, "this period" was wrong on a rolling window. Dropped to "Plan usage" in 6a2de77. The period wording comes back with the dates in #8258.

frontend/web/components/organisation-settings/usage/UsageDashboard/UsageDashboard.tsx
```diff
@@ -0,0 +1,90 @@
...
+  // without one, free and most enterprise agreements, run on rolling windows.
+  const hasBillingPeriod =
+    planNames.free !== Utils.getPlanName(subscription?.plan ?? '') &&
+    !!subscription?.has_active_billing_periods
+
+  // The default depends on the plan, which arrives after the first render, so
+  // only an explicit choice is stored. 'default' resolves once it is known.
+  const [chosenPeriod, setChosenPeriod] = useState<
+    Req['getOrganisationUsage']['billing_period'] | 'default'
+  >('default')
```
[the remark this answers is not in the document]
talissoncosta: Named both in 6a2de77: `UsageTone` exported from `UsageBar`, and `PeriodOption` typed off `Req['getOrganisationUsage']['billing_period']`.

frontend/documentation/components/UsageDashboard.stories.tsx
```diff
@@ -0,0 +1,119 @@
+import type { Meta, StoryObj } from 'storybook'
+import UsageDashboardView from 'components/organisation-settings/usage/UsageDashboard/UsageDashboardView'
```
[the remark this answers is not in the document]
talissoncosta: Exported the view from the barrel and imported through the folder in 6a2de77.

frontend/documentation/components/UsageDashboard.stories.tsx
```diff
@@ -0,0 +1,119 @@
...
+    total: 1830000,
+  },
+}
+
+export const PaidOverTheLimit: Story = {
+  args: {
+    data: usage(28, 92000),
+    hasBillingPeriod: true,
+    limit: 2000000,
+    total: 2560000,
```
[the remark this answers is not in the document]
talissoncosta: The hand-written totals contradicted the chart. Derived from the fixtures in 6a2de77, which lands them at 93.5% and 120.2%, so the story names still hold.

frontend/documentation/components/UsageMeter.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export const WithANote: Story = {
+  args: {
+    limit: 2000000,
+    note: (
+      <p className='mt-3 mb-0 text-muted fs-small'>
+        On track to use ~1.9M calls by 17 Aug.
+      </p>
+    ),
```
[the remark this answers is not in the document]
talissoncosta: Replaced with date-independent copy in 6a2de77.

frontend/web/components/organisation-settings/usage/UsageDashboard/UsageDashboardBody.tsx
```diff
@@ -0,0 +1,61 @@
...
+  hasBillingPeriod,
+  isError,
+  isLoading,
+  limit,
+  total,
+}) => {
+  if (isLoading) {
+    return (
+      <div className='text-center'>
+        <Loader />
```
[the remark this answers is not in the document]
talissoncosta: `Loader` is a global, declared in `global.d.ts:61` alongside `Row` and `Select`:

```ts
const Loader: typeof Component
```

So no import is needed, and `tsc --noEmit` is clean on this file. Roughly 40 other components render it the same way, e.g. `AppLoader.tsx` and `FeatureVersion.tsx`.

frontend/web/components/organisation-settings/usage/UsageDashboard/UsageDashboard.tsx
```diff
@@ -0,0 +1,95 @@
...
+ * it, so the old one stays untouched until the swap. Owns the fetching;
+ * UsageDashboardView draws the result.
+ */
+const UsageDashboard: FC<UsageDashboardProps> = ({ organisationId }) => {
+  const [project, setProject] = useState<string | undefined>()
+
+  const { data: organisation } = useGetOrganisationQuery(
+    { id: organisationId },
+    { skip: !organisationId },
+  )
```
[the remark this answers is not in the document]
talissoncosta: Good catch, and this one was not actually addressed by those commits. The usage query ran before the organisation loaded, so a billed org asked for the rolling window and then re-fired with `current_billing_period`. Gated on the organisation now (4b3973e6).

frontend/web/components/organisation-settings/usage/UsageDashboard/UsageDashboard.tsx
```diff
@@ -0,0 +1,95 @@
...
+      organisationId,
+      // ProjectFilter hands the id back as a string, the query wants the pk.
+      projectId: project ? Number(project) : undefined,
+    },
+    { skip: !organisationId },
+  )
+  const { data: subscriptionMeta } = useGetSubscriptionMetadataQuery(
+    { id: organisationId },
+    { skip: !organisationId },
+  )
```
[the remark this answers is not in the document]
talissoncosta: Fixed the loading half in 4b3973e6: while the metadata query was in flight the meter read as an org with no plan limit rather than as loading, so all three queries now feed one `isLoading`.

Left its error non-fatal on purpose. No subscription metadata is what self-hosted looks like, and falling back to a raw count is correct there, so making it fatal would break self-hosted.

frontend/web/components/organisation-settings/usage/UsageMeter/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+})
+
+/**
+ * Without a limit there is nothing to divide by, so the meter falls back to
+ * reporting the raw count. One branch, so the two readings cannot drift apart.
+ */
+export const meterCopy = (
+  total: number,
+  limit: number | null | undefined,
+): MeterCopy => (limit ? withLimit(total, limit) : withoutLimit(total))
```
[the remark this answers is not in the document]
talissoncosta: Named it `PlanLimit` and exported it from `UsageBar/utils.ts` (4b3973e6). It was repeated across six files.

frontend/web/components/shared/UsageBar/UsageBarThresholds.tsx
```diff
@@ -0,0 +1,41 @@
...
+          >
+            Notify {threshold}%
+          </span>
+        </span>
+      )
+    })}
+  </>
+)
+
+export default UsageBarThresholds
```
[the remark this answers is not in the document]
talissoncosta: `UsageBarThresholds` is a sub-component of `UsageBar`, and the guideline puts those inside the parent's folder: "`ComponentName/ComponentName.tsx`, co-located `ComponentName.scss`, any sub-components, and an `index.ts`". Only the folder is imported from, via `components/shared/UsageBar`.

frontend/web/components/organisation-settings/usage/UsageDashboard/UsageDashboard.tsx
```diff
@@ -0,0 +1,95 @@
+import { FC, useState } from 'react'
+import Utils, { planNames } from 'common/utils/utils'
+import { periodOptions, rollingPeriodOptions, Req } from 'common/types/requests'
+import { useGetOrganisationQuery } from 'common/services/useOrganisation'
+import { useGetOrganisationUsageQuery } from 'common/services/useOrganisationUsage'
+import { useGetSubscriptionMetadataQuery } from 'common/services/useSubscriptionMetadata'
+import ProjectFilter from 'components/ProjectFilter'
+import UsageDashboardView from './UsageDashboardView'
+import './UsageDashboard.scss'
```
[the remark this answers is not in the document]
talissoncosta: Rule 2 is about cross-module imports (`common/`, `components/`, `project/`). Within a component folder, relative is the existing pattern, e.g. `CreateMetricForm.tsx` imports `./utils` and `ExperimentConfiguration.tsx` imports `./ExperimentRolloutCard`. Leaving as is.

talissoncosta (commented):
Notes on the parts that are not self-evident.

frontend/web/components/pages/OrganisationUsagePage.tsx
```diff
@@ -53,19 +56,20 @@ const OrganisationUsagePage: FC = () => {
...
+  // Skipped behind the flag, otherwise this page and UsageDashboard both fetch.
   const { data, isError } = useGetOrganisationUsageQuery(
     {
       billing_period: billingPeriod,
       environmentId: environment,
       organisationId: organisationId || 0,
       projectId: project,
     },
-    { skip: !organisationId },
+    { skip: !organisationId || isUsageDashboardEnabled },
```
talissoncosta, line 66: The flag is in the skip condition so the old page stops fetching once the new one is on. Without it both pages request usage on every load.

frontend/web/components/pages/usage/components/UsageOverTime/UsageOverTime.tsx
```diff
@@ -0,0 +1,85 @@
...
+  isBillingPeriod,
+  limit,
+}) => {
+  const daily = useMemo(() => dailyTotals(data), [data])
+
+  const cumulative = useMemo(() => cumulativeTotals(daily), [daily])
+
+  const xAxisInterval = xAxisIntervalFor(daily.length)
+
+  const chart = isBillingPeriod ? (
```
talissoncosta, line 37: Two chart types on purpose. A rolling window's total drops as old days fall out of it, so drawing it as a running total would show usage going backwards.

frontend/web/components/pages/usage/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * A billing term is what makes usage accumulate towards a reset. Plans without
+ * one, free and most enterprise agreements, run on rolling windows instead.
+ *
+ * The plan lookup stays with the caller: resolving it pulls in the Flux stores,
+ * which cannot load outside a browser.
+ */
+export const hasBillingPeriod = (
+  subscription: Subscription | undefined,
+  isFreePlan: boolean,
```
talissoncosta, line 13: `isFreePlan` is passed in rather than worked out here: looking up the plan name loads the legacy stores, which need a browser, so the rule could not be tested.

frontend/web/components/shared/UsageBar/UsageBar.tsx
```diff
@@ -0,0 +1,82 @@
...
+import Format from 'common/utils/format'
+import UsageBarHeader from './UsageBarHeader'
+import UsageBarThresholds from './UsageBarThresholds'
+import { boundPercent, toneFor, usagePercent } from './utils'
+import './UsageBar.scss'
+
+// The border tokens hold the saturated status colours and stay the same in both
+// themes. The text ones are darkened for contrast and read muddy as a fill; the
+// surface ones are 8% tints and read as empty track.
+const FILL_COLOURS = {
```
talissoncosta, line 12: Border tokens rather than the text ones. The text colours are darkened for readability and shift between light and dark, which makes the bar muddy. We have no token yet meaning "solid status colour".

frontend/common/types/requests.ts
```diff
@@ -86,7 +89,7 @@ export const billingPeriods = [
   { label: 'Last 90 days', value: '90_day_period' },
   { label: 'Last 30 days', value: undefined },
```
talissoncosta, line 90: `undefined` is a real value here, meaning "last 30 days", not a missing selection. Worth knowing before reading the period logic.

frontend/web/components/shared/UsageBar/UsageBarThresholds.tsx
```diff
@@ -0,0 +1,41 @@
...
+              : 'usage-bar__marker'
+          }
+          style={{ left: `${threshold}%` }}
+        >
+          <span
+            className={`usage-bar__marker-label fs-captionXSmall fw-semibold text-${
+              atLimit ? 'danger' : 'warning'
+            }`}
+          >
+            Notify {threshold}%
```
Zaimwa9, line 28: Image 

Can we ensure a bit of margin on the right there please ?

And a bit of vertical space against the ticker 
Image

frontend/web/components/pages/usage/UsageDashboardPage.tsx
```diff
@@ -0,0 +1,113 @@
...
+  resolvePeriod,
+} from './utils'
+import './UsageDashboardPage.scss'
+
+type UsageDashboardPageProps = {
+  /** Absent until the route context resolves, so nothing is drawn yet. */
+  organisationId: number | undefined
+}
+
+/** Behind `usage_dashboard`. Owns the fetching; the view draws the result. */
```
Zaimwa9: Can we remove this kind of comments? Same goes for the one L20 that would otherwise be valid for half the props type definition

frontend/web/components/pages/usage/UsageDashboardPage.tsx
```diff
@@ -0,0 +1,113 @@
...
+import Utils, { planNames } from 'common/utils/utils'
+import { useGetOrganisationQuery } from 'common/services/useOrganisation'
+import { useGetOrganisationUsageQuery } from 'common/services/useOrganisationUsage'
+import { useGetSubscriptionMetadataQuery } from 'common/services/useSubscriptionMetadata'
+import FieldLabel from 'components/base/forms/FieldLabel'
+import ProjectFilter from 'components/ProjectFilter'
+import { PeriodOption } from 'common/types/requests'
+import UsageDashboard from './UsageDashboard'
+import {
+  hasBillingPeriod as deriveHasBillingPeriod,
```
Zaimwa9: IMO something is wrong with the naming if we need to always use an alias to use the function. In the downstream variable or in the function name

frontend/web/components/pages/usage/UsageDashboardPage.tsx
```diff
@@ -0,0 +1,113 @@
...
+  } = useGetOrganisationQuery(
+    organisationId ? { id: organisationId } : skipToken,
+  )
+  const subscription = organisation?.subscription
+  const isFreePlan =
+    planNames.free === Utils.getPlanName(subscription?.plan ?? '')
+  const hasBillingPeriod = deriveHasBillingPeriod(subscription, isFreePlan)
+
+  const [chosenPeriod, setChosenPeriod] = useState<PeriodSelection>('default')
+  const billingPeriod = resolvePeriod(chosenPeriod, hasBillingPeriod)
```
Zaimwa9: Something looks wrong to me here. The `hasBillingPeriod` is derived from the organisation subscription right ?
Meaning that as long as the organisation as a billing period, it will be shown the 4 period options, including last 30/90 days (rolling window).
So if you select one of those 2, the data will be drawn against the wrong plan limit (90 days usage against monthly limit)

It's actually also visible without a billing period:

Image

Image

frontend/web/components/pages/usage/UsageDashboardPage.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  // Waits for the plan. Asking earlier would request the rolling window first,
+  // so a billed organisation would show the wrong period and then correct
+  // itself, at the cost of an extra request.
+  const {
+    data,
+    isError: usageFailed,
+    isLoading: loadingUsage,
```
Zaimwa9: `isFetching` would allow to show the loader again on period updates while `isLoading` will be true only during the initial fetch if I recall well.
Is it intentional?

frontend/web/components/pages/usage/components/UsageOverTime/UsageOverTime.tsx
```diff
@@ -0,0 +1,85 @@
...
+      verticalGrid={false}
+      barSize={14}
+    />
+  )
+
+  return (
+    <div className='p-4 border border-default rounded-lg bg-surface-default'>
+      <div className='d-flex align-items-baseline justify-content-between gap-3 mb-3'>
+        <strong>
+          {isBillingPeriod ? 'Usage vs plan limit' : 'Daily usage'}
```
Zaimwa9: For self hosted there is no limit, so maybe we could fallback to plain `Usage` or `Cumulative usage` if limit is null ?

frontend/web/components/pages/OrganisationUsagePage.tsx
```diff
@@ -15,9 +15,12 @@ import { useGetOrganisationUsageQuery } from 'common/services/useOrganisationUsa
 import { useGetSubscriptionMetadataQuery } from 'common/services/useSubscriptionMetadata'
 import UsageChartFilters from 'components/organisation-settings/usage/components/UsageChartFilters'
 import UsageChartTotals from 'components/organisation-settings/usage/components/UsageChartTotals'
+import UsageDashboard from './usage'
```
Zaimwa9: Sorry to nitpick on the import naming but that is a bit confusing versus `UsageDashboardPage`

Zaimwa9 (changes requested):
Overall good, a couple of comments as I think there is a sort of conflict related to the billing periods.
If I understand correctly, this is just the first step of a ~5/6 other PRs right ?

frontend/web/components/pages/OrganisationUsagePage.tsx
talissoncosta: Good call. Renamed to UsageDashboardPage. 482f03a

frontend/web/components/pages/usage/components/UsageOverTime/UsageOverTime.tsx
talissoncosta: Fixed in 482f03a

talissoncosta: Falls back to Cumulative usage when there's no limit

frontend/web/components/pages/usage/UsageDashboardPage.tsx
talissoncosta: Not intentional. Switched to isFetching. 482f03a

frontend/web/components/pages/usage/UsageDashboardPage.tsx
talissoncosta: Good catch. Split in two: planHasBillingPeriod decides which options are offered, isBillingPeriodSelected decides whether the chart draws against the limit. Rolling windows no longer get a limit line. 482f03a

frontend/web/components/pages/usage/UsageDashboardPage.tsx
talissoncosta: Agreed. Renamed to planHasBillingPeriod. 482f03a

frontend/web/components/pages/usage/UsageDashboardPage.tsx
talissoncosta: Missed it. thanks. Removed. 482f03a

talissoncosta:
Thanks for the thoughtful review @Zaimwa9 
I've addressed the changes. Yes, this is 1/5 ~ .

frontend/web/components/pages/usage/UsageDashboardPage.tsx
Zaimwa9: Not blocking on this specific PR but here, I still have a usage compared to a monthly limit. Is there a product decision related to how we show this ? Specifying that the 50M is monthly e.g, or maybe we could just have a pro-rata and a mention that it is indicative ?

image

Zaimwa9 (approved):
Approving with a non-blocking comment that can be done in one of the follow-up or if there is a product decision

frontend/web/components/pages/usage/UsageDashboardPage.tsx
talissoncosta: Good call!  I will handle that on a follow up PR.

Issue #8040, opened 2026-07-17
Trust Relationships (OIDC)


khvn26:
The "API Keys" Organisation settings is renamed to "API Access" and expanded with Trust relationships.

When adding a Trust relationship, the user is prompted for:
- Trusted issuer URL.
- Expected audience.
- Claim matching rules — e.g. for GitHub: repo, environment, workflow filename.
- RBAC and/or admin toggle (same as with Master API Keys.)

Admin API exposes a `POST /api/v1/auth/oidc/token/` exchange endpoint which validates the signature, checks claim, and mints a short-lived access token.

PR #8324, opened 2026-08-19
feat(OIDC): stack the issuer and audience in one column

talissoncosta:

## Changes

Contributes to #8040

Follow-up to #8313. Issuer and audience each had their own column, which left both cramped and neither labelled. They now share an "OIDC configuration" column as a label/value stack, and the provider icon moves into the name cell as a tile so every row is identifiable at a glance.

An issuer we have no preset for falls back to the freeform provider's icon, so the column stays even.

| Before | After |
| --- | --- |
|image | image |

Manually, plus the existing unit tests in the trust relationships suite (29 passing).

Steps:

1. Enable the `trust_relationships` flag for your user, then go to Organisation Settings, API Access.
2. Confirm a GitHub issuer row renders the GitHub tile, and an issuer with no preset falls back to the shield tile.
3. Confirm issuer and audience read as a labelled stack in the single "OIDC configuration" column.
4. Check both light and dark mode.


khvn26 (approved):
I love it ❤️

PR #8329, opened 2026-08-20
feat: Add EMAIL_FRONTEND_DOMAIN override to Djoser

germangarces:

## Changes

Contributes to https://github.com/Flagsmith/infrastructure/issues/492

Changes were made in https://github.com/Flagsmith/flagsmith/pull/2120 to allow the domain to be retrieved from an environment variable instead of Django's sites framework, but it didn't account for emails from Djoser (and probably other third-party libs).

`FLAGSMITH_DOMAIN` is already used for `DOMAIN_OVERRIDE`, which has the same purpose.

Manually:

```bash
FLAGSMITH_DOMAIN=asdf.example.com DJANGO_SETTINGS_MODULE=app.settings.local \
  uv run python -c "from django.conf import settings; print(repr(settings.DJOSER['EMAIL_FRONTEND_DOMAIN']))"
```

Send the email: `make serve`, then POST `/api/v1/auth/users/reset_password/` with `{"email": "<your user>"}`. The email is printed in the terminal. The reset link's domain is properly overridden.

### DIFF
--- a/api/app/settings/common.py
+++ b/api/app/settings/common.py
@@ -917,6 +917,12 @@ TRENCH_AUTH = {
     "SECRET_KEY_LENGTH": 32,
 }
 
+# Controls the app domain used in emails (currently invites and change requests).
+# If set, domain stored with `django.contrib.sites` is disregarded.
+DOMAIN_OVERRIDE = env.str("FLAGSMITH_DOMAIN", "")
+# Used when no Django site is specified.
+DEFAULT_DOMAIN = "app.flagsmith.com"
+
 USER_CREATE_PERMISSIONS = env.list(
     "USER_CREATE_PERMISSIONS", default=["custom_auth.permissions.IsSignupAllowed"]
 )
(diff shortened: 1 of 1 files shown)


api/app/settings/common.py
```diff
@@ -928,6 +928,7 @@
     "PASSWORD_RESET_CONFIRM_URL": "password-reset/confirm/{uid}/{token}",
     # if True user required to click activation link in email to activate account
     "SEND_ACTIVATION_EMAIL": env.bool("ENABLE_EMAIL_ACTIVATION", default=False),
+    "EMAIL_FRONTEND_DOMAIN": env.str("FLAGSMITH_DOMAIN", default=""),
```
matthewelwell: Why not:

```suggestion
    "EMAIL_FRONTEND_DOMAIN": DOMAIN_OVERRIDE,
```

matthewelwell (changes requested):

api/app/settings/common.py
germangarces: Oh right. I've applied your suggestion. Thanks

matthewelwell: Yes, but you didn't check that my suggestion wasn't completely broken ;) 

You'll need to re-order the var assignments in order to make this work.

germangarces: 02d6849f9335dd78be324ae11804b20b1d8a84b4

matthewelwell (approved):

PR #8335, opened 2026-08-20
fix(e2e): decide the signup flow from the redirect

talissoncosta:

## Changes

The signup test decided which onboarding flow it would get before signing up, with `test.skip(flagsmith.hasFeature('onboarding_quickstart_flow'))`. That check runs in the test process, which initialises the SDK without an identity, so it only sees the environment default. The app evaluates the flag under a per-signup identity and production rolls it out by percentage, so the test drove the legacy flow while the app redirected to `/getting-started`, then timed out waiting for `/create`.

It now waits for whichever redirect happened and skips if it isn't the legacy one. That also stops the retries poisoning themselves: a failure left the signup user behind, so every retry then failed at `/login` for an unrelated reason.

This has blocked the production deploy gate since 19 Aug 11:57Z.

Ran the suite against production (`ENV=prod npm run test`), where the flag is rolled out.

- Before: failed all three attempts, the first on `/getting-started`, the retries on `/login`.
- After: skips when it lands in the single-page flow, passes when it lands on control. The gate is unblocked either way.

### DIFF
--- a/frontend/e2e/tests/initialise-tests.pw.ts
+++ b/frontend/e2e/tests/initialise-tests.pw.ts
@@ -30,13 +30,9 @@ test.describe('Signup', () => {
     await setText(byId('email'), E2E_SIGN_UP_USER);
     await setText(byId('password'), PASSWORD);
     await click(byId('signup-btn'));
-    // The app decides between the legacy /create page and the single-page flow
-    // at /getting-started (onboarding_quickstart_flow). It reads that flag
-    // under a per-signup identity, so a percentage rollout lands differently
-    // run to run and cannot be predicted here: `hasFeature` in this process
-    // has no identity and only ever sees the environment default. Wait for
-    // whichever redirect happened, then bow out if it isn't ours; the
-    // single-page flow is covered by onboarding-tests.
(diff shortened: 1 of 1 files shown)


frontend/e2e/tests/initialise-tests.pw.ts
```diff
@@ -39,8 +30,16 @@ test.describe('Signup', () => {
     await setText(byId('email'), E2E_SIGN_UP_USER);
     await setText(byId('password'), PASSWORD);
     await click(byId('signup-btn'));
-    // Wait for navigation and form to load after signup
-    await page.waitForURL(/\/create/, { timeout: 20000 });
+    // onboarding_quickstart_flow buckets on an identity the API assigns, while
```
khvn26: nit: Technically it's still the FE that's assigning the identity. I think the actually important distinction here is that the identifier used for `onboarding_quickstart_flow` is per-org, not per-user.

khvn26 (commented):
Approved with a nit.

khvn26 (dismissed):
Approved with a nit.

khvn26 (approved):

PR #8338, opened 2026-08-21
feat: Mixpanel cohort sync webhook

gagantrivedi:

## Changes

Contributes to https://github.com/Flagsmith/flagsmith-private/issues/261

Adds the receiving end of Mixpanel's [Custom Webhook cohort sync](https://docs.mixpanel.com/docs/cohort-sync/webhooks), on top of the Amplitude branch (#8290):

- One webhook endpoint handling Mixpanel's `members`, `add_members` and `remove_members` actions, answering in the `{action, status}` envelope Mixpanel expects.
- A `members` first sync creates the cohort on the fly, keyed by a new `Cohort.external_id`; snapshot pages are treated as adds only.
- The sync key can now also be sent as the password of Basic credentials, since Mixpanel's webhook form only offers username/password.
- Failure responses carry the validation errors and emit a `sync_webhook.rejected` event.

Unit tests for the webhook actions, cohort auto-creation (including the simultaneous-creation and deleted-cohort cases), Basic credential handling, and cross-environment isolation; 100% coverage on the cohorts app. The real integration test happens against Mixpanel once the endpoint is deployed.


gagantrivedi:
@coderabbitai review

api/cohorts/models.py
```diff
@@ -26,6 +29,11 @@ class Cohort(SoftDeleteExportableModel):
         choices=CohortSourceType.choices,
         default=CohortSourceType.CSV,
     )
+    # The cohort's identifier in the external source. Mixpanel pushes under
+    # its own cohort ID, so we store it to route later requests; Amplitude
+    # uses the ID we hand back at list creation, and CSV cohorts have no
+    # external system, so both leave this null.
+    external_id = models.CharField(max_length=255, null=True, blank=True)
```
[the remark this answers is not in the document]
gagantrivedi: Deliberately dropped to keep v1 lean — we'll add the constraint and conflict recovery if duplicate creation is observed in practice.

api/cohorts/serializers.py
```diff
@@ -60,6 +60,27 @@ class AmplitudeListSerializer(serializers.Serializer[None]):
     name = serializers.CharField(max_length=2000)
 
 
+class MixpanelMemberSerializer(serializers.Serializer[None]):
+    # Length mirrors CohortMembership.identifier.
+    mixpanel_distinct_id = serializers.CharField(max_length=2000)
```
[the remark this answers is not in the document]
gagantrivedi: Deferred for now — tracked as a TODO together with the same byte-length check for Amplitude user_ids and CSV uploads.

api/cohorts/sync_views.py
```diff
@@ -83,3 +109,116 @@ def _get_cohort(self, request: Request, pk: str) -> Cohort:
...
+    The receiving end of Mixpanel's Custom Webhook cohort destination:
+    https://docs.mixpanel.com/docs/cohort-sync/webhooks
+
+    Mixpanel POSTs every message to this one URL and reads the outcome from
+    the response body, which must repeat the action alongside a
+    success/failure status.
+    """
+
+    authentication_classes = [CohortSyncKeyAuthentication]
+    permission_classes = [HasCohortSyncKey]
```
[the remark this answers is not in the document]
Zaimwa9: Agreed. Also as in amplitude we should keep them in sync with the `max_segments_allowed` limit

api/cohorts/sync_views.py
```diff
@@ -83,3 +109,116 @@ def _get_cohort(self, request: Request, pk: str) -> Cohort:
...
+            404: _MIXPANEL_FAILURE_RESPONSE,
+        },
+    )
+    def post(self, request: Request) -> Response:
+        serializer = MixpanelWebhookSerializer(data=request.data)
+        if not serializer.is_valid():
+            return self._failure(
+                request,
+                message=(
+                    f"Invalid payload: {json.dumps(serializer.errors, default=str)}"
```
Zaimwa9, line 146: Should we be worried about the size of the dumps in case of completely wrong payload ? It would be for 1k member at most right ?

api/cohorts/sync_views.py
```diff
@@ -83,3 +109,116 @@ def _get_cohort(self, request: Request, pk: str) -> Cohort:
...
+        parameters = data["parameters"]
+        identifiers = [
+            member["mixpanel_distinct_id"] for member in parameters["members"]
+        ]
+        environment = typing.cast(CohortSyncKey, request.auth).environment
+
+        if webhook_action == "members":
+            # A large first sync arrives as several requests, each one page
+            # of members. Every page only adds; removals can't be detected
+            # without seeing all pages at once.
```
Zaimwa9, line 173: Hey, to improve the deletion lifecycle related to the dropped unique constraint discussion raised by coderabbit, I wonder if we could add a check here that raises a `Cohort is being deleted` and return a 404 if the cohort is processing the deletions.

My understanding is that it would pause the sync and send an email. So it would remind the user to completely shutdown his mixpanel sync or resume it intentionally.

Sounds like a cheap trade-off

Zaimwa9 (changes requested):
Looks good, just needs to mirror the plan gating and max allowed segment. 

Also shared my opinion on adding some robustness regarding the decision to drop the unique constraint

api/cohorts/sync_views.py
gagantrivedi: Yeah, done

api/cohorts/sync_views.py
gagantrivedi: yes, 1k at most

Zaimwa9 (approved):
👏

Issue #8337, opened 2026-08-20
Deep-linked feature value renders as [object Object]


talissoncosta:
Opening a flag via a **deep-link** (`…/features?feature=<id>`) renders its value as **`[object Object]`** in the editor. Opening the same flag by clicking into it from the Features list shows the correct value.

**Root cause:** the two paths read the value from different endpoints with different shapes.
- Feature-list (normal path): `environment_feature_state` is serialised by `FeatureStateSerializerSmall` → `get_feature_state_value()` → a **flat** value (e.g. a string).
- Deep-link path: `useDeepLinkedFeature` → `useGetFeatureStatesQuery` (`features/featurestates/`), whose `feature_state_value` is the **nested** object `{type, string_value, integer_value, boolean_value}`.

`pickEnvironmentFlag` (`web/components/pages/features/hooks/deepLinkedFeature.ts`) passes that nested object straight through, so the value editor stringifies it to `[object Object]`. Confirmed against the live `featurestates` response (value returned as `{type:"unicode", string_value:"…"}`).

**Not display-only:** the edit view shows the same bad value, so saving (even with no changes) would persist `[object Object]` over the real value. The stored value is currently correct; the risk is on save.

**Repro:** deep-link to a JSON-valued flag that sits **beyond the first page** of the list (`FEATURES_PAGE_SIZE = 50`) — that triggers the `useDeepLinkedFeature` fetch of `features/featurestates/`. Flags on page 1 use the list's flat value and render fine.

PR #8344, opened 2026-08-21
refactor(featureStateToValue): extract into a Flux-free module

talissoncosta:
## Changes

Contributes to #8337 (follow-up to #8341)

`Utils.featureStateToValue` is a pure function, but it lives in `utils.tsx`, which imports the Flux stores (`AccountStore`/`ProjectStore`) at module load. That import chain breaks any unit-tested consumer, which is why #8341's deep-link hook had to inline its own copy of the flatten logic.

This extracts it into `common/utils/featureStateToValue.ts` (types only, no Flux) and has `Utils` delegate to it via a shorthand re-export — so the ~22 existing `Utils.featureStateToValue(...)` callers are unchanged. The deep-link hook now imports the shared function and drops its duplicate.

The function is broadened while moving:
- accepts an already-flat value (returned untouched), so the hook doesn't need a cast;
- normalises a missing `int`/`float` to `null`.

This is the first step of decoupling the pure helpers from `utils.tsx` (which follows the existing `common/utils/<name>.ts` convention, e.g. `ensureTrailingSlash`, `multivariate`).

> **Note:** stacked on `fix/deeplink-feature-value-8337` (#8341). Retarget to `main` once that merges.

- New `common/utils/__tests__/featureStateToValue.test.ts`: covers every type branch (bool/float/int/unicode), `value_type` for core traits, missing int/float → `null`, null/undefined → `null`, and already-flat passthrough.
- `deepLinkedFeature.test.ts`: keeps the `pickEnvironmentFlag` behaviour tests plus one integration case proving the nested value is flattened via the shared function.
- 20 unit tests pass; eslint + typecheck clean on the changed files.

### DIFF
--- a/frontend/common/utils/featureStateToValue.ts
+++ b/frontend/common/utils/featureStateToValue.ts
@@ -1,0 +1,35 @@
+import type { FeatureStateValue, FlagsmithValue } from 'common/types/responses'
+
+/**
+ * Flattens a feature state (or core trait) value into its typed scalar.
+ *
+ * Accepts either the nested `{ type, string_value, ... }` shape returned by the
+ * featurestates endpoint or an already-flat value, and returns the flat value.
+ *
+ * Kept in its own module, free of `common/utils` imports, so consumers (and
+ * their unit tests) don't pull the Flux stores in through `utils.tsx`.
+ */
+export function featureStateToValue(
+  value: FlagsmithValue | FeatureStateValue | undefined,
+): FlagsmithValue {
+  if (value === null || value === undefined) {
+    return null
+  }
+  if (typeof value !== 'object') {
+    return value
+  }
+  // `value_type` is the type key on core traits; `type` on feature states.
+  const type =
+    (value as { value_type?: FeatureStateValue['type'] }).value_type ??
+    value.type
+  switch (type) {
+    case 'bool':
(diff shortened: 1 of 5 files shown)


Zaimwa9 (dismissed):
👍

frontend/common/utils/featureStateToValue.ts
```diff
@@ -0,0 +1,35 @@
...
+): FlagsmithValue {
+  if (value === null || value === undefined) {
+    return null
+  }
+  if (typeof value !== 'object') {
+    return value
+  }
+  // `value_type` is the type key on core traits; `type` on feature states.
+  const type =
+    (value as { value_type?: FeatureStateValue['type'] }).value_type ??
```
Zaimwa9, line 23: Non blocking: maybe we could have a dedicated type to deal with this cleanly and catch renames

Zaimwa9 (approved):
One non blocking comment otherwise good to go

Zaimwa9 (approved):
Giving the approval again as it stalled

frontend/common/utils/featureStateToValue.ts
talissoncosta: Yep, I agree, I will handle it as a follow up

PR #8351, opened 2026-08-25
fix: keep experiment refresh state until requests settle

Zaimwa9:

## Changes

Clicking Refresh on the experiment results page flashed the button: the settle effect cleared `refreshRequested` against stale cached data, dropping the spinner until the invalidation refetch landed (and breaking the poll timeout guard). "Settled" now also requires no in-flight requests, extracted as `hasRefreshSettled()`.

Unit tests for `hasRefreshSettled`, plus manually: Refresh now spins continuously until results update, no flash.

### DIFF
--- a/frontend/web/components/experiments/results/resultsViewState.ts
+++ b/frontend/web/components/experiments/results/resultsViewState.ts
@@ -3,6 +3,7 @@ import {
   ExperimentStatus,
 } from 'common/types/responses'
 import { formatCountdown } from 'common/hooks/useCountdown'
+import { ExposuresViewState } from './exposuresViewState'
 
 export type ResultsViewState =
   | { kind: 'empty' }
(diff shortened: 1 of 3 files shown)


Zaimwa9:
@themis-blindfold review

frontend/web/components/experiments/results/resultsViewState.ts
```diff
@@ -49,6 +44,15 @@ export const deriveResultsViewState = (
   return { kind: 'empty' }
 }
 
+export const hasRefreshSettled = (
+  resultsViewState: ResultsViewState,
+  exposuresViewState: { kind: string },
```
talissoncosta: Non blocking: any reason for `{ kind: string }` rather than the real type?

```ts
import { ExposuresViewState } from './exposuresViewState'


exposuresViewState: ExposuresViewState,
```

talissoncosta (dismissed):
Looks good, just one small non-blocking comment.

talissoncosta (approved):

Issue #8350, opened 2026-08-24
Segment change requests silently discard custom field edits


khvn26:
## Summary

Custom fields (metadata) on a segment can be edited while segment change requests are enabled, and the edit is carried into the change request, but it is never shown to the approver and never applied when the change request is committed. The edit is silently lost.

## Repro

1. Enable Segment Change Requests on a project, and configure a custom field for segments.
2. Edit a segment, change a value on the Custom Fields tab, and save. A change request is created, as expected.
3. Open the change request. The diff shows Name, Description and Rules only — the custom field change is not shown anywhere.
4. Approve and publish the change request.
5. The live segment's custom field is unchanged.

## Expected

The custom field edit is published with the rest of the change request, shown in the change request diff, and is correctly applied OR metadata is excluded from the change requests entirely.

Issue #8127, opened 2026-07-27
Enabled features can disagree with the flags actually served after a scheduled change


khvn26:
When a scheduled change request and a direct edit to the same flag overlap, parts of Flagsmith stop agreeing with what the SDKs actually serve.

To reproduce, in an environment on v1 versioning:

1. Create a change request that turns a flag off, scheduled for tomorrow.
2. Before it goes live, edit the same flag directly and leave it on.
3. Wait for the scheduled time to pass.

The SDKs now serves the flag as off, which is expected as the scheduled change is the most recent one to go live. But the dashboard still counts the flag as enabled.

The reason is that the flag evaluation code picks the version that went live most recently, while several other places pick the version that was *created* most recently. A change request is created before the direct edit but goes live after it, so the two disagree. Anywhere a flag has more than one version, the second kind of code can show you a superseded value.

What this means for the user, by feature:

| What you see                                                      | What can go wrong                                                                                     |
| ----------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------- |
| Flags served by SDKs                                              | Correct. This is the behaviour everything else should match.                                          |
| Environment metrics: enabled features count                       | Counts the flag above as enabled when it is being served as off.                                      |
| Experiment health: sample ratio mismatch warning                  | Can warn about a traffic split you have already replaced, or miss a real mismatch.                    |
| Importing flags into an environment                               | The import fails outright once the flag you are importing into has any version history.               |
| Audit log entries for new segment and identity overrides          | A new override can be missing from the audit log, or logged twice, after a scheduled change. |
| GitHub issue and PR comments on linked features                   | The comment can list old flag values alongside the current one.                                       |
| Multivariate options endpoint used by Vercel Flags Explorer                        | Serves an arbitrary version once the flag has history.                                                |
| GitLab merge request comments showing per-environment flag states | Shows an arbitrary version once the flag has history.                                                 |
| Experiment variant stickiness                                     | No impact — this deliberately looks at history and must keep doing so (#7913).                        |
| Environment metrics: identity and segment override counts         | No impact — these count how many overrides exist, not their values.                                   |
| Experiment rollout split                                          | No impact in practice.                                                                                |

Technical detail

[...]

PR #8358, opened 2026-08-25
fix(segments): Segment Change Requests bypassable via segment deletion

### LINKED ISSUE #7532: Segment Change Requests bypassable via segment create or delete
## Summary

The project level Segment Change Requests setting gates edits to a segment's definition but does not gate creation or deletion. Combined with `Manage segments` being a single project level permission (no separate create, edit, or delete grants), any user who can edit segments can bypass the change request gate.

## Repro

1. Enable Segment Change Requests on a project.
2. As a user with `Manage segments`, edit an existing segment's rules. A change request is required, as expected.
3. Bypass paths that do not require a change request:
   - Delete the segment and create a new one with the desired rules, then point feature segment overrides at the new segment.
   - Create a new segment with the desired rules and switch the feature segment override to it, leaving the original segment in place.

Neither path triggers a change request, so the gate is effectively optional for anyone with `Manage segments`.

## Expected

With Segment Change Requests enabled, segment creation and deletion should also be gated by change requests. At minimum, deleting a segment that is referenced by any feature segment override should require a change request.

## Notes

- `Manage segments` is a single project level permission covering create, edit, and delete, so there is no RBAC workaround to grant edit only access.
- Worth aligning the gate's scope with the intent: governance over segment behaviour, not just the literal edit API call.

khvn26:

## Changes

Closes #7532.

In this PR, we prevent the following when Segment Change Requests are enabled:

- Deleting a segment that has live or scheduled feature overrides. Overrides that exist only inside an uncommitted change request do not count.
- Editing a segment directly. `PUT` and `PATCH` on a segment are refused, including metadata-only writes. Drafts held by a change request are exempt, since editing those is the workflow rather than a way around it.

Supporting changes:

- `ChangeRequestsEnabledError` moves from `features/future/exceptions.py` to `core/exceptions.py` so both features and segments raise the same 409 and error code. Its detail message is now generic.
- `Project.is_workflow_enabled`, mirroring `Environment.is_workflow_enabled`.
- `EnvironmentFeatureVersionManager.get_superseded_versions()`, which answers "has a later version already replaced this one" as a correlated subquery. Ordering by go-live rather than creation is deliberate, per #8127.
- Segments now expose `has_overrides`, so the UI is able to gray out submission buttons based on it. 
- `segments/views.py` moves to structlog, emitting `segments.update_rejected` and `segments.delete_rejected`.

Deliberately out of scope:

- Segment creation stays ungated. A segment serves nothing until an override points at it, and adding overrides is governed by Feature Change Requests.
- Custom fields (metadata) are carried into a segment change request but never shown to the approver and dropped on commit (#8350). Because this PR gates metadata-only writes too, custom fields cannot be changed on a change-request-enabled project until #8350 is fixed.
- Override writes themselves are only gated on the newer flag endpoints. `FeatureSegmentViewSet` and `EnvironmentFeatureVersionViewSet.create`/`publish` have no change request check, so the API route around Feature Change Requests remains open. The documentation added here is careful not to claim otherwise.

Added unit tests, and benchmarked `EnvironmentFeatureVersionManager.get_superseded_versions()`.

### DIFF
--- a/api/features/versioning/managers.py
+++ b/api/features/versioning/managers.py
@@ -14,16 +14,28 @@ with open(Path(__file__).parent.resolve() / "sql/get_latest_versions.sql") as f:
 
 
 class EnvironmentFeatureVersionManager(SoftDeleteManager):  # type: ignore[misc]
-    def get_superseded_versions(
+    def get_superseding_versions(
         self,
         feature_id: typing.Any,
         environment_id: typing.Any,
         live_from: typing.Any,
     ) -> QuerySet["EnvironmentFeatureVersion"]:
         """
-        Get the published versions of a flag that a later version has replaced.
+        Get the versions of a flag that have gone live since the given time.
 
-        Versions are ordered by when they went live, matching what the SDKs serve.
+        A version is superseded once another has gone live after it, so
+        `~Exists(get_superseding_versions(..., live_from=<the version's
+        live_from>))` identifies the version currently being served. Versions
+        scheduled to go live later are excluded, since they are not serving
+        anything yet.
+
+        Ordering by when versions went live rather than when they were created
+        matches what the SDKs serve — see
+        https://github.com/Flagsmith/flagsmith/issues/8127.
+
+        The arguments are loosely typed so that callers can pass `OuterRef`
+        expressions and use this as a correlated subquery, which avoids
+        querying per environment.
         """
         return self.filter(  # type: ignore[no-any-return]
             feature_id=feature_id,
--- a/api/features/versioning/managers.py
+++ b/api/features/versioning/managers.py
@@ -23,7 +23,7 @@ class EnvironmentFeatureVersionManager(SoftDeleteManager):  # type: ignore[misc]
         live_from: datetime | OuterRef,
     ) -> QuerySet["EnvironmentFeatureVersion"]:
         """
-        Get the published versions of a flag that have gone live since a given time.
+        Get the published versions of a flag that went live between the provided `live_from` and now. 
         """
         return self.filter(  # type: ignore[no-any-return]
             feature_id=feature_id,
--- a/api/features/versioning/managers.py
+++ b/api/features/versioning/managers.py
@@ -1,6 +1,8 @@
 import typing
+from datetime import datetime
 from pathlib import Path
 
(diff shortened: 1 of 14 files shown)


api/features/versioning/managers.py
```diff
@@ -14,6 +14,25 @@
...
 
 class EnvironmentFeatureVersionManager(SoftDeleteManager):  # type: ignore[misc]
+    def get_superseded_versions(
+        self,
+        feature_id: typing.Any,
+        environment_id: typing.Any,
+        live_from: typing.Any,
+    ) -> QuerySet["EnvironmentFeatureVersion"]:
+        """
+        Get the published versions of a flag that a later version has replaced.
```
matthewelwell: I don't think this is very accurate - if we were filtering for published versions superseded by a specific version, I'd expect to see that version passed in and the filter to be something like `live_from__lt=version.live_from`, but here we're just passing in the `live_from` directly so it reads to me that this method should be called (and documented as) something like `get_versions_published_before`?

matthewelwell (changes requested):
Mostly LGTM - one minor comment on function naming.

api/features/versioning/managers.py
```diff
@@ -14,6 +16,23 @@
...
 
 class EnvironmentFeatureVersionManager(SoftDeleteManager):  # type: ignore[misc]
+    def get_versions_live_since(
+        self,
+        feature_id: int | OuterRef,
+        environment_id: int | OuterRef,
+        live_from: datetime | OuterRef,
+    ) -> QuerySet["EnvironmentFeatureVersion"]:
+        """
+        Get the published versions of a flag that have gone live since a given time.
```
matthewelwell: Sorry to be picky, but while `since` does kind of work in the method name itself, it doesn't really work here. I'd go with: 


```suggestion
        Get the published versions of a flag that went live between the provided `live_from` and now. 
```

api/features/versioning/managers.py
khvn26: I agree, the name was actually completely misleading. Renamed to `get_versions_live_since` in 1817a31ef405d76f552e065b5ec6b71e8eae37e6.

matthewelwell (approved):

Issue #8368, opened 2026-08-25
Support deactivated user membership


khvn26:
For the interest of Okta integration, we need to support deactivated user membership.

### Acceptance criteria

1. `UserOrganisation` model exposes an `is_active` boolean field.
2. Users with deactivated memberships are able to log in, but cannot access the organisation, or see it in the org picker.
3. Deactivated memberships do not count towards org seat limits.
4. Via SCIM API, deactivated members return 200 with `active: false`.
5. SCIM PUT/PATCH with `active: false` writes a `UserOrganisation.is_active = False`.
7. SCIM PUT/PATCH with `active: true` writes a `UserOrganisation.is_active = True`.
8. SCIM docs are updated to reflect the above.

khvn26:
Verified in prod via Okta's Runscope suite.

Issue #4993, opened 2025-01-13
Users are removed from organisation with no warning if subscription is cancelled


rolodato:
### How are you running Flagsmith

- [ ] Self Hosted with Docker
- [ ] Self Hosted with Kubernetes
- [x] SaaS at flagsmith.com
- [ ] Some other way (add details in description below)

### Describe the bug

In https://github.com/Flagsmith/flagsmith/issues/2197, we added behaviour for removing all but the first admin user from an organisation if its subscription is cancelled for any reason. This includes accidentally not paying/renewing the subscription and having it be automatically canceled.

This is not a good user experience for several reasons:

* The first admin is not necessarily the one actually using Flagsmith, and they might be unavailable to do anything.
* Even if the first admin is available, they might not have access to the billing portal.
* There is no warning or notification anywhere that this will happen or has happened.
* If a customer wants to re-instate their paid subscription, they have to renew the subscription AND also invite the additional admins. From the customer's perspective, this is data loss and does not inspire confidence.
* If an admin was removed and they log in to Flagsmith, they will not see any indication they were removed. They might see a forever-loading screen or a blank organisation switcher with no other context.

Slack discussion here: https://flagsmith.slack.com/archives/CTF0THS2D/p1736773771007339

### Steps To Reproduce

On a paid organisation with more than 1 admin user, cancel the subscription. All admins that joined after the first one are removed with no explanation or context.

### Expected behavior

Never automatically remove users from organisations without user interaction or consent.

One alternative might be introduce the following behaviour. If the following are true:

* Organisation is on a free plan
* Organisation has more than 1 admin which does not have `is_staff` (this allows support staff to add themselves to free organisations)

Then, block access for all users in this organisation to the whole dashboard except the following:

* Organisation page (to allow switching to another non-blocked org)
* Account page
* Organisation Users & Permissions page
* Usage
* Organisation Settings (specifically Billing page at a minimum)

All admins including staff users should be blocked from accessing this org, so that they can see exactly what the customer is seeing in this situation.

A message explaining the situation with suggested action items should be displayed to the user, so they can resolve the situation themselves.

The current blocking mechanism we have is at the organisation level and is more restrictive, since it does not allow access to the Users & Permissions page. The mechanism proposed here should be more lenient to allow users to unblock themselves after they run into this situation.

asaphko:
This is especially relevant for customers moving from Trial to Enterprise or other paid plans - that is the focus, more than free users.

PR #8370, opened 2026-08-25
feat(SCIM): Support deactivated user membership

khvn26:

## Changes

Contributes to #8368

In this PR, we add `UserOrganisation.is_active`, defaulting to `True`, so a membership can be suspended without being destroyed. 

A deactivated member:
- can still log in, but gets a `403` on the organisation and no longer sees it in the organisation picker;
- doesn't count towards the plan's seat limit;
- keeps their role, project/environment permissions and group memberships, so reactivation restores access exactly as it was.

Membership is checked in a lot of places, all of which previously traversed the `Organisation.users` M2M and so would have kept letting deactivated members in.

The traversals are now expressed as `permission_service.get_active_membership_filter`, which keeps the `user` and `is_active` lookups in a single `filter()` call.

`UserOrganisation.is_active` is exposed read-only on `UserOrganisationSerializer` so the frontend can surface deactivated members later. Nothing in the api writes it yet outside `set_organisation_membership_active`.

Added unit tests.


api/organisations/models.py
```diff
@@ -231,6 +231,13 @@ class UserOrganisation(LifecycleModelMixin, models.Model):  # type: ignore[misc]
     organisation = models.ForeignKey(Organisation, on_delete=models.CASCADE)
     date_joined = models.DateTimeField(auto_now_add=True)
     role = models.CharField(max_length=50, choices=OrganisationRole.choices)
+    is_active = models.BooleanField(
+        default=True,
+        help_text=(
+            "Inactive members can still log in, but cannot access the "
+            "organisation, and do not count towards its seat limit."
+        ),
+    )
```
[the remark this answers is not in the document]
khvn26: Fixed in 70993ec.

api/organisations/models.py
```diff
@@ -209,28 +209,44 @@ def rebuild_environments(self):  # type: ignore[no-untyped-def]
...
+        if remaining_seat_holder is None:
+            # No seat is in use, so there is nothing to cancel down to.
+            return
 
         UserOrganisation.objects.filter(
             organisation=self,
-        ).exclude(
-            id=remaining_seat_holder.id  # type: ignore[union-attr]
-        ).delete()
+        ).exclude(id=remaining_seat_holder.id).delete()
```
matthewelwell, line 235: Maybe it's outside the scope of this PR, but this feels like a great opportunity to resolve [this issue](https://github.com/Flagsmith/flagsmith/issues/4993) at the same time by just changing this to `update(is_active=False)` instead of `delete()` ?

api/users/models.py
```diff
@@ -254,12 +254,44 @@ def join_organisation_from_invite(self, invite: "AbstractBaseInviteModel"):  # t
...
-    def get_admin_organisations(self):  # type: ignore[no-untyped-def]
-        return Organisation.objects.filter(
+    def get_admin_organisations(self) -> QuerySet[Organisation]:
+        # NOTE: the lookups must stay in a single `filter()` call so that they
+        # apply to the same `UserOrganisation` row.
+        return Organisation.objects.filter(  # type: ignore[no-any-return]
             userorganisation__user=self,
             userorganisation__role=OrganisationRole.ADMIN.name,
+            userorganisation__is_active=True,
         )
```
matthewelwell: Now that there are multiple filters, maybe this is worth doing for DRY-ness? 

```suggestion
        return self.get_active_organisations().filter(
            userorganisation__role=Organisation.ADMIN.name
        )
```

api/users/models.py
```diff
@@ -254,12 +254,44 @@ def join_organisation_from_invite(self, invite: "AbstractBaseInviteModel"):  # t
...
+        """
+        Activate or deactivate this user's membership of `organisation`.
+
+        Deactivated members keep their roles, permissions and group memberships,
+        but lose access to the organisation and free up their seat.
+
+        Reactivation deliberately does not enforce the plan's seat limit: this is
+        driven by an external identity provider over SCIM, where failing the call
+        would leave the provider and Flagsmith out of sync. Going over the limit
+        is handled by the usual seat overage billing instead.
```
matthewelwell: I'm not sure that this is true.

As I understand it, the seat overage billing is called explicitly, not implicitly. See here: https://github.com/Flagsmith/flagsmith/blob/8b4d8466e14c9d5c08739dfe94eae6741f2cf56c/api/organisations/models.py#L481. 

I'm not saying that we should necessarily worry about it, but this comment is definitely wrong.

matthewelwell (changes requested):

api/users/models.py
khvn26: Agreed. 38dda0a3e

api/users/models.py
khvn26: No, that would result in a double join, and incorrect results. That's precisely what the `NOTE:` comment is warning against. See https://docs.djangoproject.com/en/5.2/topics/db/queries/#spanning-multi-valued-relationships

api/organisations/models.py
khvn26: My only gripe with this is that SCIM might start arguing with Chargebee. I think we should think this through. Worth to add a comment to #4993 mentioning that we can now suspend memberships, though.

matthewelwell (approved):
Approving, and I'll think on where we should document the possible seat overage gap reached when re-activating users through SCIM.

api/users/models.py
matthewelwell: Updated the title and description of [this issue](https://github.com/Flagsmith/flagsmith-private/issues/223) to cover this as well if/when needed.

PR #8371, opened 2026-08-25
ci(frontend-deploy): report every deploy outcome in one Slack message

talissoncosta:

## Changes

`#frontend-deploy` gets three messages per deploy: a start card and a terminal card from the GitHub Slack app, plus our own notification. The app's `✅ succeeded` card duplicates ours with less detail, no commit, no actor, no link to production. On 25 Aug that was ~18 messages for 6 runs.

Our notification only fired on success, so unsubscribing the app would have lost the failed and cancelled cases. This moves them here.

- The job now runs on `always()` and takes its outcome from `needs.deploy-production.result`, not the run's own conclusion. `deploy-demo` and `deploy-demo2` run alongside production, so one failed demo would otherwise report production as failed while it was serving the new release.
- Three outcomes: **deployed**, **failed**, **stopped**. Only a deploy gets the Open Production button and the smoke test ask.
- The card carries the run start time and duration, replacing the app's start ping.
- Cancelled gets its own wording. Job level concurrency on `run-tests` stops a run whenever the next commit lands on main, so most cancellations mean superseded, and the app rendered those with a red cross. The copy does not assert what production is serving, since a hand cancellation can land mid-deploy.
- The Block Kit payload moves to `payload.jq`, so it can be read and run on its own.

18 messages down to 6, one per run.

### Deployed

image

### Failed

image

### Stopped (superseded)

image

### Follow-up, not in this PR

Once a **superseded** run has been reported by the new card, the app subscription needs removing from the channel by hand, otherwise the duplicates continue. Waiting for that specific case is deliberate: it is the one path that depends on `always()` scheduling a job after a cancellation, which cannot be verified before merge.

```
/github unsubscribe Flagsmith/flagsmith workflows:{name:"Frontend Deploy to Production" event:"push" branch:"main"}
```

Rendered all three cards through Slack's Block Kit Builder using the action's real `payload.jq` (screenshots above). Ran `payload.jq` directly for every conclusion: `success` and `cancelled` get their own copy, `failure`, `timed_out` and `skipped` all read as failed, and with no timings supplied the Started and Duration fields drop out. Checked the timing logic in an `ubuntu:24.04` container since it needs GNU `date -d`, covering both timestamps, start only, neither, end before start, and unparseable input.

Checked the outcome mapping across the seven `needs.*.result` combinations that can occur, including production succeeding while a demo fails, which reports `success`.

Not exercised end to end, and it cannot be: this workflow only runs on push to `main`. The Slack step is `continue-on-error: true` and the job is downstream of the deploy, so nothing here can affect a deploy. The one behaviour that needs a live run is whether `always()` schedules the job after a concurrency cancellation, which is why the unsubscribe waits on seeing a supersede reported.

### DIFF
--- a/.github/actions/notify-slack-deploy/action.yml
+++ b/.github/actions/notify-slack-deploy/action.yml
@@ -1,5 +1,5 @@
 name: Notify Slack of a deploy
-description: Post a deploy notification to a Slack incoming webhook, prompting a smoke test.
+description: Post the outcome of a deploy to a Slack incoming webhook.
 
 inputs:
   webhook_url:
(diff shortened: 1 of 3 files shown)


Zaimwa9 (approved):
All good to me 👍

PR #8377, opened 2026-08-26
feat(api): auto-connect Flagsmith warehouse for new orgs via experimental_flags value

Zaimwa9:

## Changes

Adds auto-provisioning of Flagsmith hosted warehouse connections, controlled by the `experimental_flags` flag value.

- `get_experiment_flag_config()` reads and parses the `experimental_flags` JSON value (only when the flag is enabled for the org).
- `ensure_flagsmith_warehouse_connection()` auto-creates a Flagsmith warehouse connection on first access if `auto_connect_warehouse` is `true` and no connection exists. Race-safe via the existing unique constraint.
- Wired into the warehouse connection list endpoint for lazy provisioning.

Companion frontend PR: #8378

### Flag configuration

Set the `experimental_flags` segment override value to:
```json
{"auto_connect_warehouse": true}
```

Revert by setting `auto_connect_warehouse` to `false` or removing the value — no code deploy needed.

- `ensure_flagsmith_warehouse_connection` is a no-op when `auto_connect_warehouse` is absent/false or a connection already exists.
- Race condition handled by `IntegrityError` catch on the existing `unique_active_warehouse_per_env` constraint.
- Compile checks pass.


api/experimentation/views.py
```diff
@@ -142,6 +143,7 @@ def perform_destroy(self, instance: WarehouseConnection) -> None:
         instance.delete()
 
     def list(self, request: Request, *args: object, **kwargs: object) -> Response:
+        ensure_flagsmith_warehouse_connection(self._get_environment())
```
gagantrivedi: I wonder if we should move creation the front-end? I see following issue with this:
1. A user can't delete their warehouse connection — the next page load recreates it.

ensure_flagsmith_warehouse_connection checks for a live connection only, and delete() is a soft delete (SoftDeleteExportableModel, confirmed by test_delete__exists__returns_204 asserting deleted_at is not None). So: admin deletes the Flagsmith connection → remove_environment_ingestion_keys fires → frontend refetches the list → connection is recreated and write_environment_ingestion_keys fires again. There's no way to opt out short of turning the flag off org-wide.

2. A GET now writes to the DB and enqueues a task(which is odd)

gagantrivedi (changes requested):

api/experimentation/views.py
Zaimwa9: 1. True, thanks for catching that and actually my initial idea was wrong. I will actually move the logic to an environment hook. I re-checked and that shouldn't incur any extra clickhouse costs. Just redis writes for auth
2. True too, would not be the case anyways

api/experimentation/signals.py
```diff
@@ -0,0 +1,19 @@
+from __future__ import annotations
+
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from environments.models import Environment
+from experimentation.services import ensure_flagsmith_warehouse_connection
+
+
+@receiver(post_save, sender=Environment)
```
gagantrivedi: Sorry, can we move this to a hook instead of a signal? I think we use hook for other experimenation related stuff as well

gagantrivedi (changes requested):

gagantrivedi (approved):

PR #8379, opened 2026-08-26
feat(fe): flag-driven warehouse plan gating for free orgs

Zaimwa9:

## Changes

Makes warehouse plan gating flag-driven so free orgs can access experimentation when enabled.

- `getPlanPermission` checks `getRequiredPlan` before the free-plan bail-out — if a feature resolves to `'free'`, all plans pass. No regression risk: every other feature returns a hardcoded plan string that is never `'free'`.
- `getRequiredPlan('WAREHOUSE')` supports both the legacy array format and a new object format (`{ allowed_plans: [...] }`).
- `Plan` type and `planHierarchy` include `'free'`.

Companion API PR: #8377

### Flag configuration

Set the `experimentation_warehouse_connection` segment override value to:
```json
{"allowed_plans": ["free"]}
```
Or use the legacy array format: `["free"]`.

Revert by removing `"free"` from the array — warehouse reverts to enterprise-only plan gate. No code deploy needed.

- No regression for existing `getPlanPermission` callers: every other feature returns a hardcoded plan string, so the new `'free'` early-return never fires for them.
- Lint passes.

### DIFF
--- a/frontend/common/utils/utils.tsx
+++ b/frontend/common/utils/utils.tsx
@@ -4,7 +4,6 @@ import Project from 'common/project'
 import {
   ContentType,
   FeatureState,
-  FeatureStateValue,
   FlagsmithValue,
   MultivariateFeatureStateValue,
   MultivariateOption,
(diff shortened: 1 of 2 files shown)


Zaimwa9:
@themis-blindfold review

frontend/common/utils/utils.tsx
```diff
@@ -180,9 +180,23 @@ const Utils = Object.assign({}, BaseUtils, {
...
+      case 'bool':
+        return featureState.boolean_value
+      case 'float':
+        return featureState.float_value
+      case 'int':
+        return featureState.integer_value
+      default:
+        return featureState.string_value
+    }
+  },
```
talissoncosta: Did you mean to re-inline this function? It was extracted on #8344

talissoncosta (dismissed):
One question before approving it.

gagantrivedi (approved):

PR #8403, opened 2026-08-28
feat: allow uppercase characters in trait keys

Zaimwa9:

## Changes

- Stop lowercasing the trait key input — the API and SDKs are case-sensitive,
  so `appVersion` couldn't be created from the dashboard. Spaces still become underscores.

- Manually: the Trait ID field keeps typed casing and saves it; spaces still convert to underscores.


talissoncosta (approved):
LGTM. What about covering this with a test? createTrait already asserts the key renders exactly, so flipping the age pair in `segment-test.pw.ts` to ageAny would do it. 

Happy for it to be a follow-up if you'd rather keep this one to a single line.

PR #8407, opened 2026-08-29
feat: Exclude verified Edge Proxy requests from API usage tracking

gagantrivedi:
## Changes

Contributes to Flagsmith/flagsmith-private#256

An Edge Proxy reports the requests it serves through its own usage endpoint, so counting its `/environment-document/` fetches here would double-count. `APIUsageMiddleware` now skips a request when the private `edge_proxy` app verifies it as the proxy's own — `is_edge_proxy_request()` checks the `X-Proxy-Key` value and that the key's grants cover the presented environment. Bare header presence is never trusted, so the header can't be spoofed to dodge metering.

Wired only on non-SaaS deployments with the `edge_proxy` app installed; SaaS behaviour is untouched.

Stacked on #8305. **Before merge:** bump the `flagsmith-private` pin to the release that ships `is_edge_proxy_request`.

19 middleware unit tests: verifier wired only when installed + non-SaaS; verified requests untracked, unverified tracked; a proxy header with no verifier available still tracked (spoof regression); existing tracking cases unchanged. The verifier itself is tested against a real database in flagsmith-private.

### DIFF
--- a/api/app/settings/common.py
+++ b/api/app/settings/common.py
@@ -1160,6 +1160,10 @@ if SCIM_INSTALLED:
         "USER_FILTER_PARSER": "scim.filters.UserFilterQuery",
     }
 
+EDGE_PROXY_INSTALLED = importlib.util.find_spec("edge_proxy") is not None
+if EDGE_PROXY_INSTALLED:
+    INSTALLED_APPS.append("edge_proxy")
+
 DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
 
 # Used to keep edge identities in sync by forwarding the http requests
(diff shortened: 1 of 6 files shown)


gagantrivedi:
Companion PRs for Flagsmith/flagsmith-private#256:
- Proxy-side reporter: Flagsmith/edge-proxy-rs#19 (stacked on Flagsmith/edge-proxy-rs#18)
- Usage ingestion endpoint: Flagsmith/flagsmith-private#282
- Core middleware exclusion: Flagsmith/flagsmith#8407 (stacked on Flagsmith/flagsmith#8305)

Merge/deploy order: #282 and 8407 (with the flagsmith-private pin bump) before any proxy build with #19 is deployed, else document polls double-count during the overlap.

Issue #8442, opened 2026-09-02
Create feature / Value - Input type text enhacement


dragosbubu-tech:
### Is your feature request related to a problem?

To move copy icon inside input text field - user expected behaviour is different & we want a clear separation from file extension selection and copy action

### Describe the solution you'd like.

<img width="798" height="787" alt="Image" src="https://github.com/user-attachments/assets/35425c37-34c0-4f33-ad05-3aaeab0d50b9" />

### Describe alternatives you've considered

no alternatives to explore

PR #8448, opened 2026-09-02
test(e2e): drop the textarea fork and select the value editors by role

### LINKED ISSUE #8441: Accessibility feature - Input type text: placeholder

The current version of colour doesn't pass AA accessibility standards for commercial apps which means it has poor visibility for users leading to slowing down the experience / difficulty in reading / challenges to visual impairment issues.

Color change for placeholder from #9DA4AF to #636D7C
or in the neutral colors defined, from 300 to 500.

Refence:
https://69aedd804464bb3078b3ee78-gulesaeysh.chromatic.com/?path=/docs/design-system-palette--docs

no alternatives to explore

### Additional context

_No response_

talissoncosta:
> [!WARNING]
> **Not for this round.** #8446 is the PR that closes #8441 and #8442; this is
> follow-up work and should not merge before it. Kept as a draft against a
> non-main base for that reason.

## Changes

Stacked on #8447 — review that first, this targets its branch.

Split out of #8447 deliberately: it is the only part of that work CI can validate and the only part I cannot check locally, so it is worth being able to revert on its own.

**The fork.** `E2E ? <textarea> : <Highlight>` was added in 2022 (#1642) as a TestCafe workaround. Playwright landed in March (#6562) and drives a `contenteditable` with `fill()`; TestCafe is gone from `package.json` entirely. Until now every E2E assertion about this editor ran against a `<textarea>` that only exists during tests — the real editor was never exercised.

`placeholder` and `readOnly` only ever worked on that textarea, so they go with it, along with the three call sites passing them.

**The selectors.** `ValueEditor` no longer takes a `data-test`. Three accessors replace it, and `setText`/`waitForElementVisible` accept a `Locator`:

```ts
featureValueField()           // role=textbox, /^(Value|Control Value)/
variationValueField(i)        // role=textbox, "Variation Value", .nth(i)
segmentOverrideValueField(i)  // scoped to the override, then role+name
```

This is only possible because #8447 gave the editor `role="textbox"` and an accessible name.

The alternation in `featureValueField` is not defensive: the label becomes `Control Value <weight>%` once a feature has variations, and `editRemoteConfig` hits that path.

It also retires a selector that encoded its own value:

```jsx
data-test={`featureVariationValue${Utils.featureStateToValue(value) || index}`}
```

The id was `featureVariationValue1` only while the field was empty, and became `featureVariationValueadded` once you typed. The tests passed because they addressed it before typing.

`toHaveValue` becomes `toHaveText` in `change-request-test`: the editor is a `contenteditable`, so there is no `value` to assert on.

**CI is the test here, and that is the point of the split.** I cannot run the suite locally — no API on :8000 and no containers — so this PR is unverified until the E2E job runs.

What I could check: reproduced what `fill()` does to the live component in Storybook (focus, select all, `insertText`) and confirmed the edit reaches `onChange` and survives a blur. No new type errors in `e2e/` (3 before, 3 after, all pre-existing).

## Reviewing this

If the E2E job is red, the two commits are independently revertible: the fork removal and the selector swap fail differently — a selector problem fails as "element not found", a contenteditable problem as "value not set".

## Follow-up

The helper layer is the next thing this unlocks. `setText` wraps `fill()` with a redundant `waitFor` and `clear()` (both already done by `fill()`), plus `.first()`, which disables Playwright's strict mode. Across the file: 1246 lines, 75 helpers, 51 `.first()` calls, 47 redundant waits. As tests move to `getByRole`, they can call Playwright directly and the layer shrinks — no big-bang rewrite needed.

### DIFF
--- a/frontend/documentation/components/ValueEditor.stories.tsx
+++ b/frontend/documentation/components/ValueEditor.stories.tsx
@@ -21,7 +21,12 @@ const Interactive = ({
   const [value, setValue] = useState(initialValue)
   return (
     <div style={{ maxWidth: width, padding: 16 }}>
-      <ValueEditor {...props} value={value} onChange={setValue} />
+      <ValueEditor
+        data-test='valueEditor'
+        {...props}
+        value={value}
+        onChange={setValue}
+      />
     </div>
   )
 }
--- a/frontend/documentation/components/ValueEditor.stories.tsx
+++ b/frontend/documentation/components/ValueEditor.stories.tsx
@@ -21,12 +21,7 @@ const Interactive = ({
   const [value, setValue] = useState(initialValue)
   return (
     <div style={{ maxWidth: width, padding: 16 }}>
-      <ValueEditor
-        data-test='valueEditor'
-        {...props}
-        value={value}
-        onChange={setValue}
-      />
+      <ValueEditor {...props} value={value} onChange={setValue} />
     </div>
   )
 }
--- a/frontend/e2e/helpers/e2e-helpers.playwright.ts
+++ b/frontend/e2e/helpers/e2e-helpers.playwright.ts
@@ -1,4 +1,4 @@
-import { Page, expect } from '@playwright/test';
+import { Locator, Page, expect } from '@playwright/test';
 import { LONG_TIMEOUT, SHORT_TIMEOUT, byId, log, logUsingLastSection, getFlagsmith } from './utils.playwright';
 
 // Re-export for backwards compatibility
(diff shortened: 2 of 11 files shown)


talissoncosta:
Folded into #8447 rather than merged on its own: its two commits now sit at the top of that branch (`9eab949b8` removes the textarea fork, `4ee253ee4` moves the selectors to role and accessible name), so GitHub sees the base branch containing this head and closed it as merged.

The split left whichever PR merged first with the production contenteditable untested, which Themis raised as a blocking major on #8447.

PR #8451, opened 2026-09-03
feat(experimentation): add conversion-over-time rows to experiment results

gagantrivedi:

## Changes

Contributes to the conversion-rate-over-time chart on the experiment results page (backend half; frontend follows).

The results payload only carried headline statistics, so the dashboard had nothing to draw a per-variant conversion rate over time from.

- `ResultsSummary.exposures_timeseries`: per-bucket newly exposed identities per variant, computed in the same run as the metrics so both sides of the rate share one `as_of`.
- `MetricResult.conversions_timeseries`: per-bucket identities whose first post-exposure conversion landed in that bucket, per variant. Occurrence metrics only; `null` for value metrics.
- One gatherer, `get_results_aggregates`, runs the results, conversions and exposures reads with a single `ResultsQueryBuilder`. The conversions query reuses the exposures CTE and each slot's attribution condition, so bucket totals equal the metric's `sum`, and joins only the charted metrics' events.
- Stored payloads predating this lack the two keys; the frontend should treat them as optional.

Unit tests for the query builder, bucket decoding and summary assembly. Ran the gatherer against a local ClickHouse with seeded events: pre-exposure conversions ignored, duplicate events counted once, quarantined identities excluded, and day/hour bucket totals match each metric's `sum`.

### DIFF
--- a/api/experimentation/dataclasses.py
+++ b/api/experimentation/dataclasses.py
@@ -113,7 +113,10 @@ class ResultsSummary:
     srm_p_value: float | None
     metrics: list[MetricResult]
     # Denominator for the conversion charts, computed in the same run as the
-    # metrics so both sides of the rate share one as_of.
+    # metrics so both sides of the rate share one as_of. Exposures are bucketed
+    # by first exposure and conversions by first conversion, so only running
+    # totals divide: a bucket's own conversions over its own new identities
+    # compares different people and can exceed 100%.
     exposures_timeseries: ExposuresTimeseries
 
 
--- a/api/experimentation/dataclasses.py
+++ b/api/experimentation/dataclasses.py
@@ -1,4 +1,4 @@
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from datetime import datetime
 
 from core.dataclasses import AuthorData
@@ -57,6 +57,26 @@ class ExposuresSummary:
 
 
 @dataclass(frozen=True)
+class ConversionBucket:
+    metric_id: int
+    variant: str
+    bucket: datetime
+    converted_identities: int
+
+
+@dataclass(frozen=True)
+class ConversionsTimeseriesPoint:
+    bucket: str
+    converted_identities: dict[str, int]
+
+
+@dataclass(frozen=True)
+class ConversionsTimeseries:
+    granularity: ExposureGranularity
+    points: list[ConversionsTimeseriesPoint]
+
+
+@dataclass(frozen=True)
 class MetricSpec:
     metric_id: int
     event: str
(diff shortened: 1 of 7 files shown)


api/experimentation/services.py
```diff
@@ -482,15 +558,36 @@ def compute_results_summary(
...
+            granularity=granularity,
+        ),
+        conversion_buckets=get_conversion_buckets(
+            environment_key=environment_key,
+            feature_name=feature_name,
+            window_start=window_start,
+            window_end=window_end,
+            specs=specs,
+            granularity=granularity,
+        ),
```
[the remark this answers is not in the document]
gagantrivedi: not worth it for now

api/experimentation/dataclasses.py
```diff
@@ -66,26 +85,36 @@ class MetricSpec:
...
+    conversions_timeseries: ConversionsTimeseries | None
 
 
 @dataclass(frozen=True)
 class ResultsSummary:
     srm_p_value: float | None
     metrics: list[MetricResult]
+    # Denominator for the conversion charts, computed in the same run as the
+    # metrics so both sides of the rate share one as_of.
+    exposures_timeseries: ExposuresTimeseries
```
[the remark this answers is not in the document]
gagantrivedi: are you happy now?

Zaimwa9:
@themis-blindfold review

api/experimentation/services.py
```diff
@@ -413,37 +449,64 @@ def get_exposure_buckets(
...
+    )
     exposure_counts, metric_stats = builder.decode_rows(
         rows, [name for name, _type in columns]
     )
 
+    conversion_buckets: dict[int, list[ConversionBucket]] = {}
+    conversions_query = builder.build_conversions_query(
+        bucket_function=_EXPOSURE_BUCKET_FUNCTIONS[granularity]
+    )
+    if conversions_query is not None:
```
Zaimwa9, line 489: If I recall well, we have 120s timeouts for clickhouse queries. This one is the third in this task when it is an occurence metric, do we have an idea of the risk to timeout the task ?

api/experimentation/services.py
```diff
@@ -413,37 +449,64 @@ def get_exposure_buckets(
...
+        conversion_buckets = builder.decode_conversion_rows(
+            rows, [name for name, _type in columns]
+        )
+
     return ResultsAggregates(
         specs=list(specs),
         exposure_counts=exposure_counts,
         metric_stats=metric_stats,
+        granularity=granularity,
+        exposure_buckets=get_exposure_buckets(
```
Zaimwa9, line 502: Yeah, we are scanning many times the exposures. Not in this scope but maybe for a follow-up, it would be nice to find a way to re-use, merge or share the exposures between jobs

Zaimwa9 (approved):
Approving as I have non-blocking comments.

I wonder if we really need the hourly granularity though or if we are ok to tease with "incoming results" if the experiment is younger than 72h

api/experimentation/services.py
gagantrivedi: p99 over the last month for query latency in ClickHouse is under 800ms, so based on that I guess we're well covered.

api/experimentation/services.py
gagantrivedi: Yes, make sense for a follow up

PR #8455, opened 2026-09-03
feat(Segment Membership): Skip traitless identities when seeding

matthewelwell:

## Changes

Contributes to https://github.com/Flagsmith/edge-api/pull/714

Ignores empty<sup>1</sup> identities when seeding identity data for segment membership purposes. 

This comes with a slight compromise in that percentage split segments and segments containing only 'Is not set' operators may report inaccurate counts, but is an accepted compromise in order to reduce the amount of data in ClickHouse for now. 

<sup>1</sup> _an empty identity is an identity with `identity_traits == [] and system_traits == []`_

Added `test_seed_organisation_identities__traitless_identities__are_not_mirrored`, covering an identity with no traits (dropped) and one carrying only a system trait (kept).


api/segment_membership/tasks.py
```diff
@@ -61,6 +61,9 @@ def seed_organisation_identities(organisation_id: int) -> None:
 
     Rows are versioned at scan start via `inserted_at`
     so writes arriving mid-scan win ReplacingMergeTree dedup over the seeded row.
+
+    Identities carrying no traits at all are skipped, keeping the mirror off the
+    large populations of empty identities some environments accumulate.
```
emyller, line 66:
```suggestion
    Identities carrying no traits are skipped as they carry little to no value for segment membership.
```

Deixis. If we need to mention a temporary context, consider adding a TODO pointing to an issue in which we discuss stop persisting ephemeral identities.

api/segment_membership/tasks.py
```diff
@@ -104,7 +107,13 @@ def seed_organisation_identities(organisation_id: int) -> None:
                                     scan_started_at,
                                 )
                                 for doc in batch
+                                # An identity with nothing on it can only match
+                                # a segment by percentage split or `is not set`.
```
emyller, line 111: Sorry I don't see how this comment warrants _skipping_ identities. As I see, both percentage split and `is not set` are perfectly valid segment use case; we're skipping because we need to avoid ephemeral identities and checking for empty traits happens to match how a certain customer adds them.

emyller (changes requested):
> Ran `tests/unit/segment_membership` — 48 passed. The ClickHouse-marked tests were deselected; I don't have a local ClickHouse to run them against.
> 
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

I'm [pretty sure](https://github.com/Flagsmith/flagsmith/blob/main/docker/api/docker-compose.local.yml#L52-L77) you do. ClickHouse related tests seem to fail.

Also, a couple comments.

api/segment_membership/tasks.py
matthewelwell: Improved (I think) in dfdace5

api/segment_membership/tasks.py
matthewelwell: applied in dfdace5

matthewelwell:
> I'm [pretty sure](https://github.com/Flagsmith/flagsmith/blob/main/docker/api/docker-compose.local.yml#L52-L77) you do. ClickHouse related tests seem to fail.
> 
> Also, a couple comments.

Thanks for the review (and the slop tolerance!) 

I've fixed the test, and actioned the comments. I'll be honest, I'm still not 100% confident on the compromise, but I think it's the right decision. 

Given the implementation here though (where we're skipping empty identities for all environments), we might want to update the [CDC side](https://github.com/Flagsmith/edge-api/pull/714) to only send non-empty identities for all environments too? What do you think?

api/tests/unit/segment_membership/test_unit_segment_membership_tasks.py
```diff
@@ -151,6 +151,53 @@ def test_seed_organisation_identities__insert_fails__logs_and_continues(
...
+    # When
+    seed_organisation_identities(project.organisation_id)
+
+    # Then
+    # `dave` has nothing on him at all; `erin`'s cohort membership lives in
+    # `system_traits`, so she stays in the mirror.
+    payloads = [call.args[1] for call in cursor.executemany.call_args_list]
+    mirrored_identifiers = sorted(row[1] for payload in payloads for row in payload)
+    assert mirrored_identifiers == ["alice", "carol", "erin"]
+    # `dave`'s batch is skipped outright rather than written as an empty INSERT.
```
emyller: Sorry, can you clarify the evidence for this? e.g. should there be a certain mock call assert? Or is `dave`'s absence in the `mirrored_identifiers` the proof?

The extra prose here makes understanding a bit harder IMO, and we've been preferring not to include them, under the premise that the test title and body should be easy to understand.

api/segment_membership/tasks.py
```diff
@@ -104,7 +107,12 @@ def seed_organisation_identities(organisation_id: int) -> None:
                                     scan_started_at,
                                 )
                                 for doc in batch
+                                # 'empty' identities carry little to no value
+                                # so we skip them here to save on resources
```
emyller: Please choose either comment to keep, this or the one in the docstring. IMO we don't need both.

```suggestion
```

emyller (commented):
The code looks like what we need, but I'd suggest another anti-slop pass for the sake of our better understanding of this decision, and why we made it, in the future.

api/segment_membership/tasks.py
matthewelwell: Excellent point, sorry - this was actually my own real-life human mistake :(

emyller:
> Given the implementation here though (where we're skipping empty identities for all environments), we might want to update the [CDC side](https://github.com/Flagsmith/edge-api/pull/714) to only send non-empty identities for all environments too? What do you think?

If I understand all this correctly: this PR caters for historical identities, while https://github.com/Flagsmith/edge-api/pull/714 caters for live identities, so yes, I guess they **should** go together.

api/tests/unit/segment_membership/test_unit_segment_membership_tasks.py
matthewelwell: Yep, this test was uber hard to read. I used my own human brain to try and make it easier in 805558e1f9ee895805ed1358569e6513c0364398

api/segment_membership/tasks.py
matthewelwell: Removed in 5c26f9d

matthewelwell:
> If I understand all this correctly: this PR caters for historical identities, while [Flagsmith/edge-api#714](https://github.com/Flagsmith/edge-api/pull/714) caters for live identities, so yes, I guess they **should** go together.

I agree. I've added [this PR](https://github.com/Flagsmith/edge-api/pull/716) stacked on top of [Flagsmith/edge-api#714](https://github.com/Flagsmith/edge-api/pull/714) which adds support for `system_traits`, and makes the filter generic to all environments.

api/tests/unit/segment_membership/test_unit_segment_membership_tasks.py
emyller: I'm angry because the answer to my question was there already, it was just hard to unpack from the clutter. Thanks for improving it.

emyller (approved):
[Reviewing](https://github.com/Flagsmith/edge-api/pull/716#pullrequestreview-5145513610) the other pull request led me to think we're potentially missing on an opportunity to apply the filter at DDB query-level, and save on resources/cost/time.

Given we're already dragging this and the above is a late finding, I'm ready to approve as to unblock the goal, but would love a follow up.

PR #8463, opened 2026-09-04
feat: experiment conversion rate chart

Zaimwa9:

## Changes

Frontend for the conversion timeseries added in #8451 (branched off it).

Adds a "Conversion rate over time" card to the experiment Analysis section: stacked bars per variant (cumulative exposures, conversions filled in), with a Cumulative/Daily toggle. Daily shows raw per-bucket counts side by side — no per-day rate, per the backend contract.

Derivation lives in a tested `deriveConversionRate.ts`; `BarChart`/`ChartTooltip` gained optional props only (existing consumers unchanged). The card hides itself on payloads that predate the backend change.

- 10 Jest tests on the derive module (bucket union, carry-forward, div-by-zero, daily vs cumulative); 120 passing across experiments/charts. Typecheck delta zero.
- Manually against real payloads with fabricated timeseries: both toggle modes, tooltips, and the hidden state for value-metric/pre-deploy payloads.

image

image

### DIFF
--- a/frontend/web/components/charts/BarChart.tsx
+++ b/frontend/web/components/charts/BarChart.tsx
@@ -35,20 +35,90 @@ type BarChartProps = {
   barSize?: number
   /** Render vertical grid lines (one per x tick). Default `true`. */
   verticalGrid?: boolean
+  /** Chart height in pixels. Default 400. */
+  height?: number
+  /**
+   * dataKey → stack id, for part-of-whole bars: series sharing a stack id
+   * stack together, distinct ids sit side by side (e.g. converted/remainder
+   * segments stacked per variant, variants grouped). Default: all series
+   * share one stack.
+   */
+  stackMap?: Record<string, string>
+  /**
+   * dataKey → fill opacity (0–1). Colours are CSS `var()` strings, so
+   * transparency must come from SVG fill-opacity, not an alpha channel.
+   */
+  opacityMap?: Record<string, number>
+  /**
+   * Per-entry tooltip value renderer, threaded to ChartTooltip. Skipped for
+   * missing or non-numeric values, which render blank.
+   */
+  tooltipValueFormatter?: (
+    value: number,
+    seriesKey: string,
+    label: string,
+  ) => string
+  /**
+   * Hide the tooltip's total row — required when `tooltipValueFormatter`
+   * renders a non-additive unit such as a percentage.
+   */
+  tooltipHideTotal?: boolean
 }
+
+type LegendValue = string | number
+
+type FadedSwatchLegendProps = {
+  opacityMap: Record<string, number>
+  seriesLabels?: Record<string, string>
+  // Injected by recharts' <Legend content={...}>.
+  payload?: { value?: LegendValue; color?: string }[]
+}
+
+const FadedSwatchLegend: FC<FadedSwatchLegendProps> = ({
+  opacityMap,
+  payload,
+  seriesLabels,
+}) => (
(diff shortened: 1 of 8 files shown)


Zaimwa9:
@themis-blindfold review

Zaimwa9:
@themis-blindfold review

talissoncosta (commented):
BarChart picks up seven props here: five that only this card calls, and two (`grouped`, `yAxis`) with no caller anywhere.

Pushed a draft on top of your branch instead of a nit list: #8470. One `BarSeries[]` replaces the four dataKey-keyed maps, which makes grouped unnecessary (side by side is just distinct stack ids, as your daily mode already does) and collapses the two tooltip props into one. 15 props down to 8, your call site from six to two. A `PartOfWholeStacks` story comes with it, so the stacks and the faded legend are easy to check in Storybook.

Take it or close it, your call. I left `REST_SUFFIX` and the tooltip's key parsing alone, since they depend on whether buildConversionRateChartData stays.

It targets your branch: cherry-pick the two commits, or land yours first and I'll retarget.

frontend/web/components/charts/BarChart.tsx
```diff
@@ -35,20 +35,100 @@ type BarChartProps = {
   barSize?: number
   /** Render vertical grid lines (one per x tick). Default `true`. */
   verticalGrid?: boolean
+  /** Chart height in pixels. Default 400. */
+  height?: number
+  /**
+   * Render series side by side instead of stacked. Required for non-additive
+   * values (rates, percentages) where stacking would be meaningless.
+   */
+  grouped?: boolean
```
talissoncosta, line 44: It seems to have no caller for this one, and stackMap overrides it anyway. Same for yAxis on line 57. Do you mind checking it ?

frontend/web/components/experiments/results/ExperimentConversionRateCard/ExperimentConversionRateCard.tsx
```diff
@@ -0,0 +1,146 @@
...
+            <Select
+              isDisabled
+              size='select-sm'
+              value={{ label: metric.metric_name, value: metric.metric }}
+              options={[{ label: metric.metric_name, value: metric.metric }]}
+            />
+          </div>
+        ) : undefined
+      }
+      className='experiment-results__conversion-rate-card'
```
talissoncosta, line 102: Do you mind checking if this class is defined anywhere? I couldn't find a rule for it, and ContentCard has background='white' if that's the surface you were after.

talissoncosta (dismissed):
Approving. Just left two small comments, not blocking.

frontend/web/components/experiments/results/ExperimentConversionRateCard/ExperimentConversionRateCard.tsx
```diff
@@ -0,0 +1,145 @@
...
+  )?.conversions_timeseries
+  const hasConversions = !!conversions && conversions.points.length > 0
+
+  return (
+    <ContentCard
+      action={
+        hasConversions ? (
+          // Single metric today — disabled until multi-metric ships.
+          <div style={{ minWidth: 180 }}>
+            <Select
```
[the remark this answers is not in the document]
Zaimwa9: wrong, it's a global

Zaimwa9:
@talissoncosta, addressed the nits, thanks for the follow-up PR, I will pick it up during the week but keeping the scope to this chart in this one

talissoncosta (approved):

PR #8466, opened 2026-09-04
chore(renovate): enable lock file maintenance

matthewelwell:
Enable [lock file maintenance](https://docs.renovatebot.com/configuration-options/#lockfilemaintenance) in renovate which deletes and recreates all 4 lock files (`api/uv.lock`, `docs/package-lock.json`, `frontend/package-lock.json` and `mcp/uv.lock`) once a week. 

This exists in order to work around the lack of transitive dependency support in renovate. 

As I understand it, this will run as part of the self-hosted renovate flow that we are using, and hence should be correctly authenticated with code artifact, but that will need testing. 

Related PR to refresh (most of) `api/uv.lock` so we're not starting from scratch: #8469.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -4,7 +4,16 @@
     "security:only-security-updates",
     ":semanticCommitTypeAll(deps)"
   ],
+  "lockFileMaintenance": {
+    "enabled": true,
+    "schedule": ["before 4am on monday"]
+  },
   "packageRules": [
+    {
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -4,7 +4,16 @@
     "security:only-security-updates",
     ":semanticCommitTypeAll(deps)"
   ],
+  "lockFileMaintenance": {
+    "enabled": true,
+    "schedule": ["before 4am on monday"]
```
emyller, line 9: note: they turned crontab into an LLM prompt? lol

emyller (approved):

